PR #4, opened 2016-03-14
Add skip_quoted_offset variable

kdave:
Testing the neomutt workflow with a simple feature patch from my local stash.
---
Tunable for number of lines of quoted text that are displayed before the
unquoted text after 'skip to quoted' command (S).

Signed-off-by: David Sterba <dsterba@suse.cz>

### DIFF
--- a/globals.h
+++ b/globals.h
@@ -203,6 +203,7 @@ WHERE short ReflowWrap;
 WHERE short SaveHist;
 WHERE short SendmailWait;
 WHERE short SleepTime INITVAL (1);
+WHERE short SkipQuotedOffset;
 WHERE short TimeInc;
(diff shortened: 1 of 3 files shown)


flatcap:
@kdave 
Hurray!  NeoMutt's first pull-request.  The code looks fairly safe and useful.

However, I was put into a bit of a panic over your expression "Testing the neomutt workflow".  Reading that, I realised that I didn't know how all the branches should fit together.  Fortunately, I drew a new diagram and all's well.

Rather than merge the change directly, I'll put it in its own feature branch.  Some time soon, I want to be able to sell the Mutt-devs on each feature.

Thanks.

PR #6, opened 2016-03-18
build without any network module

257:
should fix the build failure with this run of configure:

``./configure --with-homespool=~/.maildir --enable-sidebar --enable-notmuch --with-bdb --disable-full-doc --enable-hcache``

### DIFF
--- a/version.c
+++ b/version.c
@@ -162,7 +162,9 @@ static struct compile_options comp_opts[] = {
 #else
 	{ "HAVE_CURS_SET", 0 },
 #endif
+#ifdef HAVE_GETADDRINFO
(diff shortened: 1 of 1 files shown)


flatcap:
@257 Thanks for that.
I checked configure and I was **certain** that some variables were **always** defined.  Never mind.

I've extended your change to **all** the definitions.
Also, I've pulled your change into the _ifdef_ branch and merged **that** into neomutt.

[Commit](https://github.com/neomutt/neomutt/commit/508c1da4b1151f5a1beb4bb776c64b61a2d0cc99)

Cheers,
Rich / FlatCap

257:
thanks.

PR #10, opened 2016-03-24
main: add command to limit to only current thread

kdave:
Add command to limit view on the thread from any of its message.
Predefined on command <ESC>L, though the lowercase 'l' would be better
(less keystrokes), but it's already occupied by a not-so-useful command
"show limit".

Tested on large folders, no performance problems observed.

Search in current thread is not implemented.

Signed-off-by: David Sterba <dsterba@suse.cz>


flatcap:
@kdave Ooh, that's nice.  But d'you know what'd be nicer?  If it were a toggle.  If you've some free time :-)

I've pushed it to [[neomutt/feature/limit-current-thread](https://github.com/neomutt/neomutt/tree/feature/limit-current-thread)].
I'm just doing some testing on Mutt, but I'll try to make a NeoMutt release, with your features, very soon.

kdave:
I'm not sure there's enough information about the previous state, as it modifies the "the is mail in limit view". You can go back by redoing the limit, that's what I do.

PR #22, opened 2016-05-03
Make README markdown so github will render properly

markhuge:
This is a really trivial OCD PR. I won't be butthurt if you reject it.


flatcap:
@markhuge 
All improvements are gratefully received.
Thanks.

Issue #33, opened 2016-05-26
Cleaning up trailing whitespaces and mixed indents


darnir:
Going through the source here, I've noticed a significant amount of trailing whitespace and mixed indentation (tabs + spaces) in the code.

If you're okay with it, I can submit a patch that reformats the entire code with a single style, either tabs or spaces. What does the official Mutt use? We should follow the same.

Trailing whitespaces should always be eliminated, so I'll send a PR with that in any case, as soon as I'm done with the current task

flatcap:
Hang on!

Mutt's code is a ****ing mess.  If offered to clean it up, but that just turned into a large argument on the dev mailing list: who, what, when, why, how...  Eventually, they decided they weren't ready for change.

All the trailing whitespace makes me mad, but I'm not sure I can do anything about it.
If I unify all the code in my repo, then all the upstream patches become a nightmare to apply.

While NeoMutt is **not** a fork, then I'm stuck.
If that ever changes, whitespace will be the first thing I change.

Looking though NeoMutt, you'll see four types of code:

- **Original Mutt code** -- 2 space indent, badly inconsistant
- **Maintained code (NNTP, NotMuch)** -- 2 space indent, like Mutt, but better
- **NeoMutt inclusions in Mutt code** - indented to match surrounding code
- **NeoMutt entire functions/files** - tab indent, kernel-style {}, comment blocks, strict adherence

Tab-indenting means different people can view the code at their preferred level of indentation.
sidebar.c is a good example of how I'd like the code to look.

flatcap:
Hmm... `--ignore-whitespace` sounds promising.
I've just checked the git-merge man page and it seem it can do the same.

I have a script (not yet public), that pulls from hg into a git repo, then merges into [mutt/*].
When I get a chance, I'll create [devel/mutt-whitespace] and we'll see how it goes.

flatcap:
Sorry about the wait.  I hadn't forgotten about this.  I just haven't come up with a clever solution.
The problem is that NeoMutt isn't the end of the line, it's the middle.

If we imagine that we take your pull-request and tidy NeoMutt.
By using the  `--ignore-whitespace` option we can merge new Mutt code.

But...  I need to create feature patches which will **apply to** clean Mutt source.
I can only do that if the features are **based on** clean Mutt source.

Sorry.  This request isn't practical, so I'm closing it.

PR #34, opened 2016-05-26
Remove trailing whitespaces

darnir:


flatcap:
Sorry, this is impractical.
Please see my comment in #33

PR #35, opened 2016-05-26
Cleanup travis script

darnir:
This commit introduces the following changes:
  * Use a git clone depth of 3. We don't need any more than that
  * Disable use of ccache. Since each instance is building only one
  version of Mutt, ccache gives us no speedup advantage at all
  * Use ./prepare instead of autoreconf -i && ./configure. Eliminate
  need for a before_script step.

EDIT: Also use sudo:no to enable the container based architecture which promises to be faster.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -5,7 +5,7 @@ notifications:
     on_success: change
     on_failure: always
 
-sudo: required
+sudo: no
 dist: trusty
 
 language: c


flatcap:
Hmm... broken.  I don't have the time to investigate this at the moment.
As you quickly spotted, my travis scripts have room for improvement.

- sudo : I read that I needed it for something; I forget what.
- ccache : it should be working.  travis stores the contents of `~/.ccache` .  If you take a [successful build](https://travis-ci.org/neomutt/neomutt/jobs/132692300), and search for "store build cache" and "ccache -s", you see cache hits.

I'm not sure how much, if any, time the cache is saving.  Travis seems to spend more time creating/destroying the containers.

If you're interested in fixing more, I can give you a list of requirements :-)

darnir:
I've updated the branch and commit to simply enable containerized builds now.

I think we should disable sudo and enable containerized builds which are faster unless you have some specific reason not to

flatcap:
I'd prefer the containers, too, but they break the notmuch build.

**Working Example**: https://travis-ci.org/neomutt/travis-test/builds/139061158

**Config:**
```
sudo: required
dist: trusty
```

**Output**
```
Operating System Details
Distributor ID: Ubuntu
Description:    Ubuntu 14.04.3 LTS
Release:        14.04
Codename:       trusty
```

**Broken Example**: https://travis-ci.org/neomutt/travis-test/builds/139061922

**Config:**
```
sudo: no
dist: trusty
```

**Output**
```
Operating System Details
Distributor ID: Ubuntu
Description:    Ubuntu 12.04.5 LTS
Release:        12.04
Codename:       precise
```

Disabling sudo means we get Ubuntu 12.04 (despite: `dist: trusty`).  This has a very old version of notmuch which doesn't compile.

Yes, I could find a PPA for a recent version of notmuch, but I'd rather not.
I'd rather use the newer version of Ubuntu and all the newer compile tools.

flatcap:
I've adopted the git clone depth into the config files and I've left the ccache.
Now lots of builds are happening in one travis run, we see even more of an improvement.

PR #59, opened 2016-06-22
Fix compiler warnings due to always true condition

darnir:
BUFFY declares path as a char array. Hence, p->path will always resolve
to a valid array location making the check redundant. It causes a
compiler warning. Hence, remove the redundant check to silence the
warning.


flatcap:
I thought I'd fixed this, but you found another one.  Thanks.
**Note:** I've merged this into [feature/notmuch].  It'll get merged into [neomutt] when I next do a batch of merges.

PR #86, opened 2016-07-21
Configure

darnir:


flatcap:
Committed:
- 1-7 into NeoMutt
- 8,9,10 into their branches, then merged into NeoMutt
- 11-13 into NeoMutt

PR #91, opened 2016-07-24
Add configure summary for Header Cache database

darnir:


flatcap:
merged

I suggest changing configure's internal variable names to be a bit more descriptive.
```
Header Cache:      tc
```

Please join me on IRC, if you're able.

PR #95, opened 2016-08-01
patch 1.5.24.aw.timeouthook.1 from http://wolfermann.org/mutt.html

ilintault:
Adds a timeout-hook to allow arbitrary commands to run when the timer expires.  Original credit to: http://wolfermann.org/mutt.html

-----------

This patch implements a new hook that is called periodically when Mutt checks for new mail. To call the hook every minute add set `timeout=60` to your .muttrc.

Example: `timeout-hook . "exec sync-mailbox"`

### DIFF
--- a/PATCHES
+++ b/PATCHES
@@ -1,3 +1,4 @@
+patch-1.5.24.aw.timeouthook.1
 patch-quasi-delete-neo-git
 patch-progress-neo-git
 patch-status-color-neo-git
--- a/curs_main.c
+++ b/curs_main.c
@@ -949,8 +949,10 @@ int mutt_index_menu (void)
 
       dprint(4, (debugfile, "mutt_index_menu[%d]: Got op %d\n", __LINE__, op));
 
-      if (op == -1)
-	continue; /* either user abort or timeout */
+      if (op == -1) {
+        mutt_timeout_hook(".");
+ 	continue; /* either user abort or timeout */
+      }
 
       mutt_curs_set (1);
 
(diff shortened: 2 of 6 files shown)


flatcap:
Thanks, @ilintault that patch is already in the system (but not NeoMutt).
Each distro has its own, small, set of patches on top of NeoMutt -- Debian has the timeout patch.
(The patches are all kept in the [Integration Repo](https://github.com/neomutt/integration)).

The plan (as yet unstarted) is to merge all these patches into NeoMutt and bring the distros together.

So far, I've been keeping  each of the features in a separate branch.  Each feature gets some basic documentation, example muttrc, etc.  Any updates would go into that branch which would then be merged into NeoMutt.  This is more work, but means that if upstream Mutt wants to adopt a feature, then it's easy to create a patch.

So...  If you've got some free time (and it wouldn't take long), you could help "fill-out" the feature.
This would get the feature accepted sooner ;-)

If not, please wait patiently.  Your feature will be along shortly -- I'm busy with other things, but it's high up the list.

ilintault:
Thanks, @flatcap - By fill out you mean docs and ?

flatcap:
Yes, some basic docs and that's about it.
Being a simple feature, there's little to do.

- Create a branch based at mutt-1.6.2 (tag)
- Apply the patch
- Copy the template docs from the [Management Repo](https://github.com/neomutt/management/tree/master/template-docs)
- Edit the docs (most of the sections won't be relevant and can be deleted).

flatcap:
Any questions, let me know.
I'm also in #neomutt on irc.freenode.net if you prefer.

flatcap:
merged

ilintault:
Thanks for doing this!

PR #96, opened 2016-08-07
Don't whine if getrandom() fails.

kilobyte:
Debian for example supports kernels down to 3.2, and getrandom() appeared
only on 3.16 on x86 or later on other archs.  Current stable has only 3.16
(ie, no getrandom() outside of x86), and ARM vendor kernels are for most
machines even older.

There is no security risk in falling back to /dev/urandom, as mutt quits
if that fails instead of using unsecure randomness.


flatcap:
Hi @kilobyte 
I saw your very thorough bug report.
Thanks for the patch.

flatcap:
Merged: 5831e48
Thanks.

PR #97, opened 2016-08-09
Contrib/keybase

JakkinStewart:
I finished updating the readme. It was incomplete at the time of the merge.


pickfire:
@JakkinStewart Is there any reason to not edit it directly?

flatcap:
> I finished updating the readme. It was incomplete at the time of the merge.

OK, sorry, that's my fault.  I didn't announce that I was making a release.
I'll give you more warning next time.

> Is there any reason to not edit it directly?

You have the privileges to merge things into NeoMutt, but... I would prefer that you don't.
I'd like to keep a little control because some users are updating directly from git.

---

Unless there's something urgent, there's no need for pull-requests.
Now that your feature is in NeoMutt, I'll merge in the latest version just before every release.

I'll merge this later today.  Thanks.

PR #102, opened 2016-08-14
Make distcheck passes. Need to fine tune now

### LINKED ISSUE #58: Clean up configure script
Taking a look at the build system, I saw lots of issues with the configure.ac script that can be cleaned up.

1. The various configure options often use the wrong terminology. --enable options should enable features in the codebase while --with options should enable support using 3rd party libraries. Hence, it should be --with-gpgme and --with-pgp.
2. There is no need to test for `x$enableval = yes` in the AC_ARG_ENABLE macros. In the 3rd argument, you can simply set the variable to the value of $enableval. Take a look at this configure script for ideas: https://github.com/rockdaboot/wget2/blob/master/configure.ac
Tim has spent considerable time making sure the configure.ac script is perfect and we've had enough long discussions about the entire autotools tree.
3. We can have configure print a summary of build options at the end. This is often useful debug information, especially for devs. Again, take a look at the above cofngure script for what it does.

I don't mind working on these points, but I'll do it only after discussion and if I know that the patches will be accepted. It would be nice to upstream these some of these patches as well for the cases where they are similar.

darnir:
This is the first commit in a series that should allow far more automation of making neomutt releases

In this commit, we essentially try to jsut get make distcheck working.
Next steps:
1. Find all files in Tarball that are superfluous and eliminate them
2. Find all files in repo that should be in tarball but are not
3. Ensure build works
4. Ensure build works without needing autotools
5. Fix the version numbering

At the end of these, we should have a nice mechanism for making releases and can close #58

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -27,6 +27,7 @@ doc/stamp-doc-xml
 
 po/POTFILES
 
+.version
 aclocal.m4
 compile
 config.guess
--- a/Makefile.am
+++ b/Makefile.am
@@ -22,7 +22,7 @@ if BUILD_HCACHE
 HCVERSION = hcversion.h
 endif
 
-BUILT_SOURCES = keymap_defs.h patchlist.c reldate.h conststrings.c $(HCVERSION)
+BUILT_SOURCES = keymap_defs.h patchlist.c reldate.h conststrings.c hcachever.sh $(HCVERSION)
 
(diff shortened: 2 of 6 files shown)


darnir:
If this works, please close it in favour of #104

flatcap:
Closed in favour of #104

PR #104, opened 2016-08-14
Allow creating Distribution Tarballs with autogen sources

darnir:
With this commit `make dist` works again. The following changes were
made:
  * Change how versioning is done. We now use the `git-version-gen`
  script from gnulib to generate a tarball version based on the latest
  git tag.
  * Ignored files in distribution: Some autogenerated files don't need to be in the distribution tarball:
    * hcachever.sh
    * mkchangelog.sh
    * prepare (Doesn't need to be since tarballs don't use it)
    * BEWARE (Is to scare developers, not users)
    * check_sec.sh (For devs to test)
    * README.notmuch (Installed as a generated file with other READMEs)
  * Added files to the tarball:
    * README.md
    * README.neomutt
    * ChangeLog.neomutt
    * .version
  * Fixed various issues preventing `make dist` from completing

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -27,6 +27,7 @@ doc/stamp-doc-xml
 
 po/POTFILES
 
+.version
 aclocal.m4
 compile
 config.guess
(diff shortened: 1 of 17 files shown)


flatcap:
Merged

PR #117, opened 2016-08-19
Fix `make dist` and add NeoMutt branding

darnir:


flatcap:
Fixed typos :-)
I also changed: compose_format, status_format, ts_status_format to display "NeoMutt"
Merged.

PR #137, opened 2016-09-11
Avoid segfault when listing mailboxes on startup.

jdswinbank:
When mutt is started with the -y flag, _mutt_select_file() is called while
CurrentFolder is a null pointer. If that's the case, we shouldn't try to use
it.


flatcap:
Thanks @jdswinbank

PR #152, opened 2016-09-17
Add Kyoto Cabinet header cache

neverpanic:
Tokyo Cabinet's website now "strongly recommends" using Kyoto Cabinet instead of Tokyo Cabinet because it claims to be superior. Add Kyoto Cabinet as an option.

In some preliminary testing, the performance seems to be on-par with Tokyo Cabinet, but not better. Kyoto Cabinet seems to need a little more space, but the difference is probably not significant.

Note that since this does not actually provide any improvement other than not using the older (and possibly unmaintained?) Tokyo Cabinet libraries, it might be a valid decision to not merge this pull request at all. Since there are a couple of attempts at implementing support for Kyoto Cabinet spread over Gists and patches, opening a PR here at least makes it more visible that the code has already been written.

### DIFF
--- a/PATCHES
+++ b/PATCHES
@@ -8,6 +8,7 @@ patch-ifdef-neomutt
 patch-index-color-neomutt
 patch-initials-neomutt
 patch-keywords-neomutt
+patch-kyoto-neomutt
 patch-limit-current-thread-neomutt
 patch-lmdb-neomutt
 patch-multiple-fcc-neomutt
--- a/configure.ac
+++ b/configure.ac
@@ -903,6 +903,7 @@ db_requested=auto
 hcache_db_used=no
 AC_ARG_ENABLE(hcache, AS_HELP_STRING([--enable-hcache],[Enable header caching]))
 AC_ARG_WITH(tokyocabinet, AS_HELP_STRING([--without-tokyocabinet],[Don't use tokyocabinet even if it is available]))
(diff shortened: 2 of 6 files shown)


flatcap:
Ooh!  This was a pleasant surprise, @neverpanic.

I've pulled your changes into the branch [[devel/kyoto](https://github.com/neomutt/neomutt/tree/devel/kyoto)]
The code looks OK, it compiles and runs OK.

Now, please can you write a quick few words of documentation for the feature.
I've copied our doc template into the branch and deleted the unnecessary sections.
You'll need to edit `doc/manual.xml.head` -- search for "FEATURE-NAME"

If you cut'n'paste from the [LMDB feature page](https://www.neomutt.org/feature/lmdb/) it won't take more than five minutes.

Many thanks,
Rich / FlatCap

neverpanic:
Rebased on top of the branch and fixed the docs. I guess you may want to pull this into the devel/kyoto branch for now.

flatcap:
Lovely docs (one minor tweak to fix the DocBook) -- pushed to the [website](https://www.neomutt.org/feature/kyoto-cabinet/)
I've made a couple of very small changes to tidy the code: a524270

neverpanic:
Thanks. I don't think this pull request makes sense anymore then, so let's abandon it and merge from the branch later.

Issue #138, opened 2016-09-11
attach_keyword change string to regex


mrajner:
I like new forgotten-attachment settings in .muttrc. However it would be nice if it could support also regex definition in `attach_keyword`, not only single string.
That functionality would be useful for multilingual uses.

flatcap:
That **would** be useful.
@darnir who merged the patch didn't have time to do anything more with it.
I'm sure we'll find some time to extend it.

mjk-gh:
I second the desire to support more than just one keyword,
but would like to propose a (maybe) simpler solution: a
space-separated list of keywords, like so:

> set attach_keywords="Anhang angeh&auml;ngt h&auml;ngt\ an"

(Note the use of backslash in the terminal phrase "h&auml;ngt an",
which is German for "is attached".)

Of course, I would not object to the above being written as

> set attach_keywords="(Anhang|angeh&auml;ngt|h&auml;ngt an)"

if you supported regex, just thought that a space-separated list
might be easier and thus faster to implement. Maybe amend this
patch with regex later and make both co-exist, e.g. by interpreting
"/pattern/" as regex pattern, everything else as string list.

flatcap:
## Help Us

There are plenty of examples, in Mutt, of parsing strings and checking regexes.
These could be copied to extend the "forgotten-attachment" feature.

If you know how to code in C, even a little, then I will be happy to help you make this enhancement become a reality.

Writing code is the guaranteed way to get the feature into NeoMutt :-)

flatcap:
#### This enhancement will be in the next release of NeoMutt.

PR #155, opened 2016-09-19
Change 'attach_keyword' option from string to regex

weisslj:
A regular expression is much more useful than a single word to check for forgotten attachments, especially when writing in multiple languages.

This is inspired by the [muttng patchset abort_noattach](https://github.com/codito/muttng-patchset/blob/master/patches/muttng.abort_noattach.diff), which I use for years, but newly implemented to make minimal changes to the neomutt source code.

The keyword is no longer mentioned in the error message as regular expressions tend to be longer.

The option 'attach_keyword' could be renamed 'attach_remind_regex' as in the muttng patch.

This pull request would solve #138.

### DIFF
--- a/README.neomutt
+++ b/README.neomutt
@@ -64,10 +64,11 @@ Here's a list of everyone who's helped NeoMutt:
 Alex Pearce, Ander Punnar, Antonio Radici, Baptiste Daroussin, Chris Salzberg,
 Christoph Berg, Darshit Shah, David Sterba, Elimar Riesebieter, Evgeni Golov,
 Fabian Groffen, Fabio Alessandro Locati, Faidon Liambotis, Guillaume Brogi,
-Ivan Tham, Johannes Frankenau, Joshua Jordi, Karel Zak, Kurt Jaeger,
-Marco Hinz, Matteo Vescovi, Pierre-Elliott Bécue, Richard Hartmann,
-Richard Russon, Santiago Torres, Sven Guckes, Thomas Klausner, Udo Schweigert,
-Vsevolod Volkov, Werner Fink, Yoshiki Vázquez Baeza.
+Ivan Tham, Johannes Frankenau, Johannes Weißl, Joshua Jordi, Karel Zak,
+Kurt Jaeger, Marco Hinz, Matteo Vescovi, Pierre-Elliott Bécue,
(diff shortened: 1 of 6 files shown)


flatcap:
Hi @weisslj 
Thank you very much for this.

The code looks good.
Please could you add a few regex examples to the docs.
Also, feel free to add yourself to the credits section :-)

weisslj:
Thanks for the review and also for inviting me to the team! I added a regex example and added me to the credits section. I also created a devel/abort_noattach_regexp feature branch for this.

P-EB:
I'd suggest to add a "default" regex in the Muttrc default configuration file.

Maybe there'd be a way to handle translations of this regex for other languages, I don't know. :)

flatcap:
We should provide examples for all the languages we speak:

English: `<attach(|ed|ments?)>` matches whole words:
- attach
- attached
- attachment
- attachments

flatcap:
I've just searched the code and we should keep all the regexes in lower case.

init.h defines the variable as DT_RX
```
{ "attach_keyword",  DT_RX,  R_NONE, UL &AttachKeyword, UL "attach" },
```
At startup, we get this stack:
- `main()`
- `mutt_init()`
- `mutt_restore_default()`

For a DT_RX type, it checks if there are any upper case letters `mutt_which_case()`
If there are none, then the regex will ignore the case of the strings: `REG_ICASE`

P-EB:
French: `<joint(|e?)s?>` matches whole words:
- joint
- joints
- jointe
- jointes

weisslj:
Thanks for the suggestions and examples! I agree that a default regex would be good. Since we cannot provide it for "all" languages I suggest we stick to English (@flatcap's example but with correct `<` and `>` escaping):

```
\\<attach(|ed|ments?)\\>
```

Also good find with the regex case sensitivity, it is even mentioned in the manual: http://www.mutt.org/doc/manual/#regexp

The new patch is 22d3219.

weisslj:
I won't be on my computer the next three weeks.

flatcap:
OK, thanks for letting me know.
Have a nice holiday :-)

flatcap:
Merged in 337f6e3.  Thanks.

weisslj:
Thanks @flatcap!

Issue #143, opened 2016-09-12
Crash when sending mail with Fcc set to current mailbox


jamessan:
When I reply to a mail, I typically set the Fcc to the mailbox containing the mail I'm replying to, in order to keep the threading intact.  I'm currently seeing a crash when I return to the index from the mail I replied to.

Steps (assuming an INBOX mailbox exists)
- Open email from INBOX mailbox
- Draft reply
- Set Fcc to `=INBOX`
- Send email
- Press `q` to return to index view from pager view

This results in the below backtrace:

```
 (gdb) bt
#0  update_index (menu=0x0, ctx=0xb71420, check=3, oldcount=29, index_hint=0) at curs_main.c:365
#1  0x000000000045e006 in mutt_pager (banner=0x0, fname=0x7fffffffd0b0 "/home/jamessan/tmp/mutt-freya-1000-6563-8036180115345718171", flags=66, extra=0x7fffffffcc50) at pager.c:2031
#2  0x0000000000414028 in mutt_display_message (cur=0x7fbeb0) at commands.c:225
#3  0x0000000000424a4c in mutt_index_menu () at curs_main.c:1902
#4  0x0000000000449553 in main (argc=1, argv=0x7fffffffe758) at main.c:877
```

The call to `update_index` in `mutt_pager` is

```c
            update_index (index, Context, check, oldcount, index_hint);
```

`index` is NULL.  This means that this code from `mutt_pager` was never called

```
1831       if (IsHeader (extra) && PagerIndexLines)
1832       {
1833         if (index == NULL)
1834         {
1835           /* only allocate the space if/when we need the index.
1836              Initialise the menu as per the main index */
1837           index = mutt_new_menu(MENU_MAIN);
1838           index->make_entry = index_make_entry;
1839           index->color = index_color;
1840           index->max = Context->vcount;
1841           index->current = extra->hdr->virtual;
1842           index->indexwin = index_window;
1843           index->statuswin = index_status_window;
1844         } 
```

Indeed, PagerIndexLines is 0 so index is never allocated.

The update_index code path was introduced by @sassmann in 4ed2b6f0f8e69b61c9e872ad9322bdc6f326ce5f.

Simply guarding the block that calls `update_index` with a check for non-NULL `index` isn't sufficient, because then the index isn't properly updated.

sassmann:
Hi jamessan,

I tried to reproduce what you're reporting, but I didn't see any crash. Few things I'd like to ask:
[...]

PR #161, opened 2016-09-25
Fixes repaint problem with $pager_index_lines #159

### LINKED ISSUE #159: Repaint problem with $pager_index_lines
NeoMutt: 2016-09-16 (doesn't happen in 2016-09-10)
Occurs: always
Steps to reproduce:
- `set pager_index_lines = 10`
- open email
- `set pager_index_lines = 0`

Result:
status bar isn't removed

Expected result:
status bar disappears

christopher-john-czettel:
Commit 6b4c4ca changed the behaviour of mutt_pager to build the
pager index even if pager_index_lines was set to 0 to (fixing #143).

The code redrawing the pager status line was relying on the pager index
beeing NULL to determine if a redraw should take place. Now
index_window->rows is used as an additional safeguard.

Also added a check if pager index is set to the terminal status line update
code as it was unguarded and old default behaviour might be introduced
later.

Fixes #159

### DIFF
--- a/pager.c
+++ b/pager.c
@@ -1958,25 +1958,28 @@ mutt_pager (const char *banner, const char *fname, int flags, pager_t *extra)
 	mutt_draw_statusline (pager_status_window->cols, bn, sizeof (bn));
       }
       NORMAL_COLOR;
-      if (option(OPTTSENABLED) && TSSupported)
+
+      /*
+       * update terminal status line
+       */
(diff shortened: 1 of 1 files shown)


flatcap:
Merged in 17d3e3f.  Thanks.

christopher-john-czettel:
happy to help :-)

Issue #168, opened 2016-09-29
NeoMutt not building on Mac OS X 10.5 PPC


lbschenkel:
Hi,

I'm the maintainer of the NeoMutt port in MacPorts. Today I got an e-mail from their buildbot notifying that it couldn't be built for Mac OS X 10.5 PPC:
https://build.macports.org/builders/ports-10.5_ppc_legacy-watcher/builds/632

The build output is here:
https://build.macports.org/builders/ports-10.5_ppc_legacy-builder/builds/5905/steps/install-port/logs/stdio

I'm at a loss to explain why the symbols are undefined since they are defined in `lib.c` and `strndup.c` and I can see both files being compiled. I don't know if this is an issue with the port, or NeoMutt, or upstream Mutt. I would appreciate any help on diagnosing/fixing this — I don't think I'll able to do it on my own.

flatcap:
Oh, hello.  I'm the maintainer of NeoMutt :-)

> NeoMutt port in MacPorts

I hadn't heard of that.  In a sentence, what's the difference between Homebrew and MacPorts?  (for a non-Mac user)  We've had Homebrew support almost since the beginning of NeoMutt.

> strndup.c and `safe_malloc()`

First, to get your build working, simply replace `safe_malloc()` with `malloc()`
I made the change to make sure everyone was allocating memory the "Mutt" way.

Why is it failing?  I really don't know.  As you point out lib.h is included and lib.c is compiled and linked in.  `safe_malloc()` is used everywhere in Mutt.  I shall do a little digging and get back to you.

flatcap:
I've found the problem.  It's not with mutt, but with mutt_dotlock and pgpewrap (two helper programs).
Even though they don't use `strndup()` or `strnlen()`, autoconf tries to link them, anyway.

I'll restore `malloc()` and it'll be in the next release -- due in the next few days.

lbschenkel:
Thanks a lot. It's not entirely clear to me if the autoconf bug was introduced in NeoMutt or is it present in upstream Mutt as well? And it's strange that it only triggers in OS X 10.5 PPC but not in any other version or architecture.

Regarding MacPorts and Homebrew, they're just different package managers. MacPorts is the older of the two and both have their legion of fans, not unlike apt vs rpm. :-)

flatcap:
`strndup()` is only used in the notmuch code, so it, and the autoconf "feature", don't appear in Mutt.

[...]

PR #172, opened 2016-10-02
Fix build failure for mutt_dotlock when strndup is missing

lbschenkel:
If strndup is missing, strndup.c is included when building mutt_dotlock.
However, strndup.c depends on safe_malloc which is implemented in lib.c and
lib.c was not included when building mutt_dotlock. This changeset corrects
that.

### DIFF
--- a/Makefile.am
+++ b/Makefile.am
@@ -92,9 +92,9 @@ EXTRA_DIST += UPDATING.kz OPS.NOTMUCH
 
 
 
-mutt_dotlock_SOURCES = mutt_dotlock.c
-mutt_dotlock_LDADD = $(LIBOBJS)
-mutt_dotlock_DEPENDENCIES = $(LIBOBJS)
+mutt_dotlock_SOURCES = mutt_dotlock.c extlib.c lib.c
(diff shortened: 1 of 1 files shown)


lbschenkel:
This should fix issue #168.

lbschenkel:
Whoops, I made a mistake. Please hold, I'll fix the branch.

lbschenkel:
OK, I should have fixed it now but I'm not happy about it since now we're depending on more code and linking against `gettext` on a setuid root binary.

I propose that this PR should be the starting point of a conversation on how to fix this.

lbschenkel:
Oh, I just noticed after I rebased to master that you fixed the dependency to `safe_malloc`. Now I feel dumb. Never mind. :-)

flatcap:
Thanks for taking the time to propose a fix (and having a conversation with yourself :-)
I committed the "simple" fix just before release.  Sorry it wasn't in time to save you some work.

The change to `safe_malloc()` **seemed** like the right thing to do, but it meant that the tools would have too heavy a dependency on Mutt.

PR #202, opened 2016-10-17
notmuch: option to affect visibility of tags

kdave:
Allow to show/hide notmuch labels on the index line, default is to show.
Useful in case there are a lot of labels that cover the subject line. No

Signed-off-by: David Sterba dsterba@suse.cz

### DIFF
--- a/hdrline.c
+++ b/hdrline.c
@@ -812,11 +812,14 @@ static const char *index_format_str(char *buf, size_t buflen, size_t col, int co
       tags = driver_tags_get_transformed(&hdr->tags);
       if (!optional)
       {
-        colorlen = add_index_color(buf, buflen, flags, MT_COLOR_INDEX_TAGS);
-        mutt_format_s(buf + colorlen, buflen - colorlen, prec, NONULL(tags));
-        add_index_color(buf + colorlen, buflen - colorlen, flags, MT_COLOR_INDEX);
+        if (ShowTags)
+        {
+          colorlen = add_index_color(buf, buflen, flags, MT_COLOR_INDEX_TAGS);
+          mutt_format_s(buf + colorlen, buflen - colorlen, prec, NONULL(tags));
+          add_index_color(buf + colorlen, buflen - colorlen, flags, MT_COLOR_INDEX);
+        }
       }
-      else if (!tags)
+      else if (!tags || !ShowTags)
         optional = 0;
       FREE(&tags);
       break;
--- a/init.h
+++ b/init.h
@@ -2151,6 +2151,12 @@ struct Option MuttVars[] = {
   ** connect to news server.
   */
 #endif
+  { "show_tags", DT_BOOL, R_TREE | R_BOTH, UL &ShowTags, 1 },
+  /*
+   ** .pp
+   ** Show tag vaules in format via %g specifier (default). Allow to hide the
+   ** tags if there's large number tags per message. In bindings use 'toggle'.
+   */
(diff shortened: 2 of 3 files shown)


flatcap:
So this is equivalent to changing `$index_format` to toggle the presence of "%g"?
OK, thanks.  Let me have think about that.

kdave:
Basically yes. Switching two index_formats is possible, but requires some macro magic.

toogley:
@kdave Do you mind adding your option to the [notmuch documentation](https://github.com/neomutt/neomutt/blob/neomutt/doc/manual.xml.head#L11929)? Thanks. :)

flatcap:
@kdave 
Hmm... useful as this is, it seems **too** specific.
Are there any other expandos (not just from notmuch) that might benefit from being hidden?
Labels from the Keywords feature?

@karelzak 
Any thoughts?

karelzak:
I think all %g is already notmuch specific, as well as tag-transforms and tag-formats used for the %g output. So extend it by nm_show_tags does not seem like a problem. Right?

Anyway, **it would be nice to have a generic way** how to hide something from index_format on-the-fly, then you can create "macro index" and toggle on/off, something like:
```
toggle_index_format = opt[:yes|no]
```
for example, to revert the current setting:
```
   toggle_index_format = %g
```
or force visibility:
```
  toggle_index_format = %g:yes
```
It would be nice for large things like %F (From: address) or so.

flatcap:
That's the sort of thing I was thinking of.

I don't think we need the '[yes|no]', or even the '%'.  A simple list of letters would do.
(The formatting, etc, lives in `$index_format`)

Then, add a vim style set method `+=` or `-=` and we're sorted.
`set toggle_index_format += g`

karelzak:
Ah yes,
```
set toggle_index_format += g
```
is definitely better!

kdave:
I'm all for making it generic. The += syntax sound good to me.

toogley:
@kdave do i understand you correctly that you aren't interested in developing the generic way?

flatcap:
@toogley - Issue #243 is our best solution so far.
@nobrowser was interested in working on it, but he's also assigned himself to #377 (creating buffer.c).

toogley:
@flatcap why did you close this PR?

toogley:
I think this PR was automatically closed, because the neomutt branch was deleted - therefore you have to set your tracking branch to master.

toogley:
blocked by https://github.com/neomutt/neomutt/issues/243

hdrline.c
```diff
@@ -841,12 +841,15 @@ static const char *hdr_format_str(char *dest, size_t destlen, size_t col, int co
     case 'g':
       if (!optional)
       {
-        colorlen = add_index_color(dest, destlen, flags, MT_COLOR_INDEX_TAGS);
-        mutt_format_s(dest + colorlen, destlen - colorlen, prefix,
-                      nm_header_get_tags_transformed(hdr));
-        add_index_color(dest + colorlen, destlen - colorlen, flags, MT_COLOR_INDEX);
+        if (option(OPT_NOTMUCH_SHOW_TAGS))
```
sileht: tags are no more a notmuch thing, so I suggest we rename OPT_NOTMUCH_SHOW_TAGS to OPT_SHOW_TAGS

flatcap:
@sileht Done.  This PR is a placeholder for a bigger solution (that hasn't been written yet).

Issue #180, opened 2016-10-04
Forgotten Attachment Improvement


flatcap:
This feature checks the body of the mail using a regex, e.g. "attach(|ment)"
If the email contains a match, but there's no attachment, notify the user.

This part works fine.  However, due to a bug in the code, the email is still
checked even if the feature is disabled:
```
set abort_noattach = no
```

This simply means that the code in `send.c` needs refactoring.
Look in send.c for the call to `mutt_search_attach_keyword()` in `ci_send_message()`

Additionally, I'd like you to change the search so that it ignores quoted text.
Any lines beginning with the text in `$indent_string` should be ignored.

---
**Easy tasks would be perfect for someone new to NeoMutt.**
You won't need to write lots of code, and you won't need any previous knowledge
of the NeoMutt code.  Obviously, you will need to be able to write C code :-)

**Easy tasks also come with a mentor** -- a NeoMutt project member -- who will help
you learn what you need to complete the work.

If you're interested in having a go, leave a comment or speak to us in our
IRC channel: #neomutt on irc.freenode.net

Thank you :heart:

nuclearsandwich:
I have a plane flight this afternoon where I'll be taking an initial crack at this. Discussing the particulars with flatcap in IRC.

PR #206, opened 2016-10-18
Avoid searching for attachment keywords when it's unnecessary

nuclearsandwich:
Resolves the "easy" portion of #180 by checking the `abort_noattach`/`OPT_ATTACH` option and does not search the message text if the option is set to `MUTT_NO`.

The second commit similarly short circuits avoiding the search if there is an attachment. At least, that's my understanding of what that portion of the conditional is. That commit is separate and can be excluded from the final merge.

I'll be looking at skipping quoted message sections from the search on the next flight. #180 mentions using `$indent_string` but in IRC @flatcap mentioned `$quote_regexp`. I've only had a chance to quickly grep through the changelog to see which is which, but it looks like the quote regexp would be the one to check message text against.

Feedback gladly appreciated. Thanks!

### DIFF
--- a/send.c
+++ b/send.c
@@ -1896,8 +1896,8 @@ main_loop:
 #endif
 
   if (quadoption(OPT_ATTACH) != MUTT_NO &&
-         mutt_search_attach_keyword (msg->content->filename) &&
          !msg->content->next &&
+         mutt_search_attach_keyword (msg->content->filename) &&
          query_quadoption (OPT_ATTACH, _("No attachments, cancel sending?")) != MUTT_NO)
   {
     /* if the abort is automatic, print an error message */
--- a/send.c
+++ b/send.c
@@ -1895,7 +1895,8 @@ main_loop:
   }
 #endif
 
-  if (mutt_search_attach_keyword (msg->content->filename) &&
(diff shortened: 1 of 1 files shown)


nuclearsandwich:
Updated to search lines matching the quote regexp and skip checking them for the attachment regexp.

flatcap:
That looks great, thanks.  I've tested it and pushed the changes into the branch [devel/forgotten](https://github.com/neomutt/neomutt/tree/devel/forgotten)

There's one more thing to do: update the docs.
Please can you add a few words to `doc/manual.xml.head` to explain how the feature ignores reply text.  Also, add yourself to the credits.

The docs are in DocBook (xml) format.  Each feature is in a separate `<sect1>`.  Look for `id="forgotten-attachment"`

When you're done, you can test out your new privs and push your commit into into the [devel/forgotten] branch of the NeoMutt repo.

nuclearsandwich:
Updated the manual in https://github.com/neomutt/neomutt/commit/4b9e19357a9b81e61b52ed003d535d8009523391

Let me know if you'd like a new PR from the `devel/forgotten` branch or have any feedback about the added docs. I opted to add minimally but I could see wanting to add similar text to the configuration section as well.

PR #210, opened 2016-10-20
Add a "Folder" line in notmuch output

immae:
This permits to know where (which "physical" folder) the message comes from, since
the information is not available anywhere else in the virtual mailbox.

### DIFF
--- a/copy.c
+++ b/copy.c
@@ -429,11 +429,27 @@ mutt_copy_header (FILE *in, HEADER *h, FILE *out, int flags, const char *prefix)
   }
 
 #ifdef USE_NOTMUCH
-  if ((flags & CH_VIRTUAL) && nm_header_get_tags(h))
+  if (flags & CH_VIRTUAL)
   {
-    fputs ("Tags: ", out);
-    fputs (nm_header_get_tags(h), out);
-    fputc ('\n', out);
+    /* Add some fake headers based on notmuch data */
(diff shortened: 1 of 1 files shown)


flatcap:
Hmm, that seems pretty useful.  I've made a few small changes.
I put the folder first, since there'll always be folder.
I also put in a call to `mutt_pretty_mailbox()` to abbreviate the name based on `$HOME` or `$folder`,
e.g. `=archive` rather than `/home/user/Mail/archive`

```C
#ifdef USE_NOTMUCH
  if (flags & CH_VIRTUAL)
  {       
    /* Add some fake headers based on notmuch data */
    char *folder = nm_header_get_folder(h);
    if (folder)
    { 
      char buffer[LONG_STRING];
      strfcpy (buffer, folder, sizeof (buffer));
      mutt_pretty_mailbox (buffer, sizeof (buffer));

      fputs ("Folder: ", out);
      fputs (buffer, out);
      fputc ('\n', out);
    }   
    char *tags = nm_header_get_tags(h);
    if (tags)
    {
      fputs ("Tags: ", out);
      fputs (tags, out);
      fputc ('\n', out);
    } 
  }   
#endif
```

immae:
Thanks for the suggestion, I updated the PR.

karelzak:
Nice idea!

PR #252, opened 2016-11-19
forward: $forward_references includes `References:` header

sgeb:
By default a forwarded message does not reference the messages it contains. This commit introduces a new boolean option `$forward_references`. When it is set, a forwarded message includes the `In-Reply-To:` and `References:` headers, just like a reply would. Hence the forwarded message becomes part of the original thread instead of starting a new one.

Sidenote: this is similar to how Gmail and some other providers handle forwards.


sgeb:
Should this change also add a line to file `PATCHES`?

flatcap:
Hi @sgeb.
That looks like a reasonable thing to do.
(We'll see if anyone else has any thoughts).

> Should this change also add a line to file PATCHES?

Yes, please.

sgeb:
>> Should this change also add a line to file PATCHES?

> Yes, please.

OK, done.

PR #260, opened 2016-11-28
Enable reconstruct-thread in the pager

mineo:
and make it behave like notmuchs entire-thread function - that is, keep displaying the message that was displayed when reconstruct-thread was called, but redisplay the index.

Note that the code changes themselves are fairly small, but there's an additional commit in this pull request that reindents the code in this branch to get rid of the mix of tabs and spaces. If that's not ok with you, I can remove that commit again :-)

Note number 2: This would need an update to the features website as well.

### DIFF
--- a/curs_main.c
+++ b/curs_main.c
@@ -1269,80 +1269,80 @@ int mutt_index_menu (void)
 	CHECK_ATTACH;
 	if (Context->magic == MUTT_NNTP)
 	{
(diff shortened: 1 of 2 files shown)


flatcap:
Hi @mineo.  Welcome to NeoMutt!
Thanks for the feature patch, it sounds very useful.
I shall review it very soon.

PR #263, opened 2016-12-01
New configure option --enable-everything

darnir:
Currently this configure option will enable:
* GPGME
* PGP
* SMIME
* Sidebar
* Compressed Folder Support
* Notmuch

In the future, this option will also enable:
* Header Cache (tokyocabinet)
* POP3
* SMTP
* IMAP
* GnuTLS
* NNTP

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -154,6 +154,12 @@ AS_IF([test x$enable_everything = "xyes"], [
 	use_imap="yes"
 	use_nntp="yes"
 	use_smtp="yes"
+	hcache_tokyocabinet="yes"
+	hcache_kyotocabinet="yes"
+	hcache_bdb="yes"
+	hcache_gdbm="yes"
+	hcache_qdbm="yes"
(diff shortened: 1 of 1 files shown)


darnir:
P.S.: This is broken since it does not allow overriding the value set by --enable-everything

flatcap:
hi @darnir.
That's starting to look good.

On IRC you asked:
> --enable_everything will require a significant change to configure.ac.
> Are we okay with that?

Yes, I'm happy with that.
Your work means that our `configure` is already superior to Mutt's.
If Mutt makes any changes, it's more likely to be a new feature than major restructuring.

darnir:
Turns out the restructuring wasn't really required. I could have managed without. But it's done now and I like how it looks. So it stays. 

Also, I've fixed the issue with overriding. Small subtlety that was missed. I'll send an updated patch soon

darnir:
Closed in favour of #264

Issue #233, opened 2016-11-15
NeoMutt Internationalization (i18n)


flatcap:
To translate a program into another language, there are two steps:

- **Internationalization** (i18n): Programmers marking the strings that need to be translated
- **Localization** (l10n): [Translators turning one language into another](https://github.com/neomutt/neomutt/issues/229)

This task is to find any strings that haven't been marked for translation.
It may be possible to script this job.

All strings should be surrounded with `_()`, e.g.

**Before**:
```
mutt_message ("I'm sorry Dave, I'm afraid I can't do that");
```

**After**:
```
mutt_message (_("I'm sorry Dave, I'm afraid I can't do that"));
```

PR #266, opened 2016-12-11
Internationalize strings

Alok:
This is a *very* crude effort that's nothing more than 2 `vim` macros
combined with `ag`. It is split into a few separate commits to make
reverting easier. Feel free to ignore the whole thing since I don't
actually know any C, or how `gettext`'s `_` works.

### DIFF
--- a/account.c
+++ b/account.c
@@ -184,7 +184,7 @@ int mutt_account_getuser (ACCOUNT* account)
   else
   {
     /* L10N: Example: Username at myhost.com */
-    snprintf (prompt, sizeof (prompt), _("Username at %s: "), account->host);
+    snprintf (prompt, sizeof (prompt), (_("Username at %s: ")), account->host);
     strfcpy (account->user, NONULL (Username), sizeof (account->user));
     if (mutt_get_field_unbuffered (prompt, account->user, sizeof (account->user), 0))
       return -1;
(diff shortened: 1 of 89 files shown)


Alok:
Is meant to address #233 .

flatcap:
Hi @Alok,

Thanks for this...

> Feel free to ignore the whole thing since I don't actually know any C

Ah, this could be a very interesting pull-request.

> Commit 1 - remove some spaces

That's fine.

> Commit 2 - Parenthesise all `_()` uses

This is a bit trickier.  `_` is a valid C function name (it's `#define`d so `_()` becomes `gettext()`.
So...
This change isn't needed
```diff
-  menu->title = _("Aliases");
+  menu->title = (_("Aliases"));  
```

But this one **is** (due to a widespread mistake in mutt):
```diff
-    mutt_error _("You have no aliases!");
+    mutt_error (_("You have no aliases!"));  
```

> Commit 3 - widespread wrapping of strings

Hmm...
The purpose of gettext is to mark strings that need to be translated.  In any given program, there will be lots of strings that shouldn't be translated:
- Contain only whitespace: `\n`
- Names of environment variables: `TERM`
- Standard email headers: `To:`, `Subject:`
- Function names
- File names
- Comments in the code (your macro is too powerful :-)

It'll take a little while to sort out -- I'll wait 'til there's something undemanding on TV.
Thanks.

Alok:
This is what happens when you indiscriminately run a macro 1376 times in Vim =(.

flatcap:
This PR is superceded by #376

PR #269, opened 2016-12-16
Apply patch to add relative message number patterns

nobrowser:
This patch adds new syntax to the ~m message number pattern:

~m/3 matches a range of 3 messages down from current one (including current one), ie. it is equivalent to ~m {N}-{N+2} where N is the user visible number of current message

similarly ~m,4 matches a range of 4 messages up from current one (including current one)

This makes it somewhat easier to mass-tag a range of messages and repeatedly perform a command on them, a feature I missed from Gnus.

Alas, I haven't updated the docs.  Maybe I can be badgered into it :-)

### DIFF
--- a/mutt.h
+++ b/mutt.h
@@ -125,12 +125,6 @@ typedef struct
   int destroy;	/* destroy `data' when done? */
 } BUFFER;
 
-typedef struct
-{
-  int ch; /* raw key pressed */
-  int op; /* function op */
-} event_t;
-
 /* flags for _mutt_system() */
 #define MUTT_DETACH_PROCESS	1	/* detach subprocess from group */
 
--- a/mutt_curses.h
+++ b/mutt_curses.h
@@ -76,6 +76,12 @@ void mutt_curs_set (int);
 #define CI_is_return(c) ((c) == '\r' || (c) == '\n')
 #endif
 
+typedef struct
+{
+  int ch; /* raw key pressed */
+  int op; /* function op */
+} event_t;
+
 event_t mutt_getch (void);
 
 void mutt_endwin (const char *);
@@ -146,7 +152,7 @@ typedef struct color_line
   regex_t rx;
   int match; /* which substringmap 0 for old behaviour */
   char *pattern;
-  pattern_t *color_pattern; /* compiled pattern to speed up index color
+  struct pattern_t *color_pattern; /* compiled pattern to speed up index color
                                calculation */
   short fg;
   short bg;
--- a/mutt_menu.h
+++ b/mutt_menu.h
@@ -127,6 +127,4 @@ int mutt_menuLoop (MUTTMENU *);
 void index_make_entry (char *, size_t, struct menu_t *, int);
 int index_color (int);
 
-int mutt_limit_current_thread (HEADER *h);
-
 #endif /* _MUTT_MENU_H_ */
--- a/pattern.c
+++ b/pattern.c
@@ -1460,52 +1460,6 @@ top_of_thread (HEADER *h)
   return t;
 }
 
-/**
- * mutt_limit_current_thread - Limit the email view to the current thread
- * @h: Header of current email
- *
- * Returns:
- *  1: Success
- *  0: Failure
- */
-int
-mutt_limit_current_thread (HEADER *h)
-{
-  int i;
-  THREAD *me;
-
-  if (!h)
-    return 0;
-
-  me = top_of_thread (h);
-  if (!me)
-    return 0;
-
-  Context->vcount    = 0;
-  Context->vsize     = 0;
-  Context->collapsed = 0;
-
-  for (i = 0; i < Context->msgcount; i++)
-  {
-    Context->hdrs[i]->virtual    = -1;
-    Context->hdrs[i]->limited    = 0;
-    Context->hdrs[i]->collapsed  = 0;
-    Context->hdrs[i]->num_hidden = 0;
-
-    if (top_of_thread (Context->hdrs[i]) == me)
-    {
-      BODY *body = Context->hdrs[i]->content;
-
-      Context->hdrs[i]->virtual = Context->vcount;
-      Context->hdrs[i]->limited = 1;
-      Context->v2r[Context->vcount] = i;
-      Context->vcount++;
-      Context->vsize += (body->length + body->offset - body->hdr_offset);
-    }
-  }
-  return 1;
-}
-
 int mutt_pattern_func (int op, char *prompt)
 {
   pattern_t *pat;


flatcap:
Hi @nobrowser, thanks for the feature.

> [Mutt devel mailing list]
> There have been no further comments, so I submitted a PR to neomutt

We'll overlook the fact that NeoMutt is your second choice :-)

All pull-requests are automatically built by TravisCI (continuous integration).
Please can you fix the build [(log)](https://travis-ci.org/neomutt/neomutt/builds/184617742) and I'll take a look.

nobrowser:
I can't see the log.  The page says "There was an error trying to fetch the log."  It says that in a color that is nearly the same as the background color.  Is this some plot to accelerate my slide into blindness? :-(

flatcap:
Hmm, I can see it.  The contrast isn't great, though -- here's a [raw version](https://s3.amazonaws.com/archive.travis-ci.org/jobs/184617743/log.txt)

```
BUILD: 1/2
    --enable-exact-address --enable-flock --enable-gpgme --enable-hcache
    --enable-imap --enable-locales-fix --enable-mailtool --enable-nfs-fix
    --enable-pop --enable-smtp --with-domain --with-gnutls --with-gss
    --with-homespool --with-mixmaster --with-regex --with-sasl --with-ssl
    --with-tokyocabinet --without-bdb --without-gdbm --without-qdbm

configure.ac:17: installing './compile'
configure.ac:21: installing './config.guess'
configure.ac:21: installing './config.sub'
configure.ac:12: installing './install-sh'
configure.ac:12: installing './missing'
Makefile.am: installing './depcomp'
configure: WARNING: unrecognized options: --enable-hcache
config.status: creating po/POTFILES
config.status: creating po/Makefile
configure: WARNING: unrecognized options: --enable-hcache
Making all in m4
Making all in intl
Making all in contrib
Making all in imap
In file included from ../mutt_menu.h:28:0,
                 from ../mutt.h:94,
                 from auth.c:27:
../mutt_curses.h:79:1: error: unknown type name â€˜event_tâ€™
 event_t mutt_getch (void);
 ^
In file included from ../mutt_menu.h:28:0,
                 from ../mutt.h:94,
                 from auth.c:27:
../mutt_curses.h:149:3: error: unknown type name â€˜pattern_tâ€™
   pattern_t *color_pattern; /* compiled pattern to speed up index color
   ^
In file included from ../mutt.h:94:0,
                 from auth.c:27:
../mutt_menu.h:130:32: error: unknown type name â€˜HEADERâ€™
 int mutt_limit_current_thread (HEADER *h);
                                ^
make[2]: *** [auth.o] Error 1
make[2]: *** Waiting for unfinished jobs....
In file included from ../mutt_menu.h:28:0,
                 from ../mutt.h:94,
                 from auth_login.c:25:
../mutt_curses.h:79:1: error: unknown type name â€˜event_tâ€™
 event_t mutt_getch (void);
 ^
In file included from ../mutt_menu.h:28:0,
                 from ../mutt.h:94,
                 from auth_login.c:25:
../mutt_curses.h:149:3: error: unknown type name â€˜pattern_tâ€™
   pattern_t *color_pattern; /* compiled pattern to speed up index color
   ^
In file included from ../mutt.h:94:0,
                 from auth_login.c:25:
../mutt_menu.h:130:32: error: unknown type name â€˜HEADERâ€™
 int mutt_limit_current_thread (HEADER *h);
                                ^
make[2]: *** [auth_login.o] Error 1
make[1]: *** [all-recursive] Error 1
make: *** [all] Error 2
```

flatcap:
Thanks for the fix.
It looks good.  I've made a couple of changes:
- I've restored mutt_limit_current_thread -- it **should** be used, but the call was accidentally trampled by a dodgy merge.
- A bit of light tidying to better conform to Mutt's loose coding standard

I _thought_ that I was able to add to a pull-request, but I can't get it to work.
(There's an option when you create a PR)

So instead, the changes are in the [[devel/rel-msgno-pattern]](devel/rel-msgno-pattern) branch.

How attached are you to the syntax?  I'd favour something a bit more sed/vim-like.

nobrowser:
@flatcap: suggest a syntax :-)

I'm an Emacs person, so I have no idea what vim-like means.
Some reasons why I did it just the way I did:

1. As I mentioned in the thread on mutt-users, I intentionally avoided allocating another top-level pattern type (like ~w).

2. The utmost possible brevity is important, not just in terms of characters but actual keystrokes, because it has to be quicker & easier than just typing the full range e.g. ~m 101-121.  So , and / because they are unshifted ...

3. ~m-10 won't work, because it is already valid and has a semantics (it is the same as 1-10).

flatcap:
OK, here goes...

We want:
- to specify absolute ranges
- to specify relative ranges
- to have open or closed ranges
- placeholders for the first/last email

Negation of ranges and other logic can be handled by Mutt.

## Proposal for `~m`

Add the `,` (comma) to represent relative ranges
Add shortcuts:
- `1` or `^` for the first email
- `.` (dot) for the current email
- `$` for the last email.
- Extend the `-` (dash) a little

## Absolute Ranges

Here are all the possible absolute ranges:

| Range   | Which emails  | Notes                          |
|:--------|:--------------|:-------------------------------|
| `4-10`  | 4 to 10       |                                |
| `4-.`   | 4 to current  |                                |
| `.`     | Current only  | Equivalent to `.-.`            |
| `.-10`  | Curent to 10  |                                |
| `1-$`   | First to last | Equivalent to `~A`             |
| `-10`   | First to 10   | Open range                     |
| `4-`    | 4 to last     | Open range                     |
| `-`     | First to last | Open range, equivalent to `~A` |
| `<10`   | First to 9    | Non-inclusive range            |
| `>4`    | 5 to last     | Non-inclusive range            |

## Relative Ranges

All the relative ranges contain `,` (comma):

| Range   | Which emails           | Notes                           |
|:--------|:-----------------------|:--------------------------------|
| `-3,.`  | Previous 3 and current |                                 |
| `.,.`   | Current only           | Equivalent to `.`               |
| `.,2`   | Current and next 2     |                                 |
| `.,$`   | Current to last        |                                 |
| `.,`    | Current to last        | Open range, equivalent to `.,$` |
| `,.`    | First to current       | Open range, equivalent to `1-.` |

## Parsing

Once any quotes have been removed, we should look for a comma which would
indicate a relative range.  They need not be in order, so "-4,1" and "1,-4"
would be equivalent.

The numbers can be converted to absolute indexes using the current message
position.  Ranges that extend before the first email, or beyond the last, should
be quietly truncated.

If there's no comma, then it must be an absolute range.  These can be parsed as
Mutt normally does it, with the addition of some extra symbols `^`, `.` and `$`.

## Notes

Vim has a `relativenumber` option.  It would change the index numbers to be
relative to the current message.  This is visually very helpful for relative
ranges.

```
-2  Email
-1  Email
 0  Current email
 1  Email
 2  Email
```

nobrowser:
What does "-4,1" mean though?  The table doesn't give it a meaning.

flatcap:
> What does "-4,1" mean though?

It has a comma, so it's relative.
It would mean the previous four emails, the current one and the next one (-4, -3, -2, -1, 0, +1).

I'm confident the notation:
- doesn't interfere with Mutt's interpretation
- is fairly compact
- isn't ambiguous

...but keep questioning me

nobrowser:
I think I like it :+1: 

Do you feel like writing some code on this one, or should I do it all myself?  If you volunteer, can you do the classical case without a comma, factoring it such that there is a subroutine for me to call from the relative case (after I convert the numbers to absolute).

No pressure :-P

flatcap:
Glad you like it :-)

If it's not a problem, please can **you** write all the code.
That way, I can concentrate on some other NeoMutt problems.

Any questions, just ask.
Cheers.

nobrowser:
Will do.

PR #272, opened 2016-12-24
fix new mail notifications on mbox on noatime filesystems

kilobyte:
There are many reasons to mount all your filesystems with noatime.  The only non-fringe downside of doing so is breaking new mail notifications on mbox files -- you're wrongly told that you have new mail.

But these notifications can be fixed by having every mail client (there's not that many of them) manually touch atime (via futimens(fd, {UTIME_NOW, UTIME_OMIT})) upon reading the mbox.  There's no cost for doing so -- if automatic atime updates are on, the inode will be dirty by the read you just did.


flatcap:
Thanks @kilobyte, the patch looks good.
I'll give it a test (and merge it) as soon as I can -- bit busy for the next few days :-)

PR #278, opened 2017-01-02
Searching with a window over notmuch vfolders

guyzmo:
When searching over a very large result, the result will be very slow for mutt
to load. So, inspired by sup's feature to iteratively load the results, I figured
adding a couple of commands to move an user set window backward/forward in time could be useful. And a new command extending the default vfolder-from-query function is being exposed.

The current window information is user settable, so the user can make a macro that would do useful stuff about a search (like moving the window's position around, or further tweaking current search).

* addition of four configuration variables, the first two are aimed at configuring
    the behaviour of the feature, the last two are exposing internals to make it
    hacking friendly:
  - `nm_query_window_duration`: duration for each window
  - `nm_query_window_timebase`: time base for each window
  - `nm_query_window_current_search`: the search being displayed as vfolder
  - `nm_query_window_current_position`: the current window's position
* addition of three bindable commands:
  - `windowed-vfolder-from-query`: execute a search with a window applied
  - `windowed-vfolder-backward`: move the window backwards in time
  - `windowed-vfolder-forward`: move the window forwards in time

I had a chat with @flatcap and @darnir about it on IRC, which led me to send [a mail to the neomutt-users mailing list][0] to have some ideas, feedback or advices… And I had so many that I implemented it following my own idea using what I could understand from the codebase… In the best way I could following what *I* would expect from such a feature.

[0]:http://mailman.neomutt.org/pipermail/neomutt-users-neomutt.org/2016-October/000017.html

### DIFF
--- a/mutt_notmuch.c
+++ b/mutt_notmuch.c
@@ -1284,8 +1284,8 @@ char *nm_uri_from_windowed_query(CONTEXT *ctx, char *buf, size_t bufsz, char *ti
 {
 	dprint(2, (debugfile, "nm_uri_from_windowed_query (%s, %s, %d)\n", buf, timebase, duration));
 
-	int beg = duration*(NotmuchQueryWindowCurrentPosition+1);
-	int end = duration*NotmuchQueryWindowCurrentPosition;
+	int beg = duration * (NotmuchQueryWindowCurrentPosition + 1);
+	int end = duration *  NotmuchQueryWindowCurrentPosition;
 
 	if (NULL == NotmuchQueryWindowCurrentSearch)
 	{
--- a/mutt_notmuch.c
+++ b/mutt_notmuch.c
@@ -1287,7 +1287,8 @@ char *nm_uri_from_windowed_query(CONTEXT *ctx, char *buf, size_t bufsz, char *ti
 	int beg = duration*(NotmuchQueryWindowCurrentPosition+1);
 	int end = duration*NotmuchQueryWindowCurrentPosition;
 
-	if (NULL == NotmuchQueryWindowCurrentSearch) {
+	if (NULL == NotmuchQueryWindowCurrentSearch)
+	{
 		nm_query_window_reset();
 		FREE(&NotmuchQueryWindowCurrentSearch);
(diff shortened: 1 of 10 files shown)


guyzmo:
Here's the configuration changes I added to test the feature:

```
macro index _ "<vfolder-from-query>" # old fashioned search
set nm_query_window_duration=2
set nm_query_window_timebase="week"
bind index \\ "windowed-vfolder-from-query"
bind index >  "windowed-vfolder-forward"
bind index <  "windowed-vfolder-backward"
```

also, both variables default to `2` and `week` in the codebase, but I guess it'd be a good idea to set `nm_query_window_duration` to `0`, which would disable the feature.

flatcap:
I've had a quick look, and given it a test, and it looks good.
I'll have a deeper look tomorrow (and have a play with GitHub new review functions).

guyzmo:
just fixed a mem leak I introduced.

flatcap:
Don't be alarmed.  This is a lot of very small changes (in separate commits for clarity) -- we can squash them later.

mutt_notmuch.c
```diff
@@ -1263,6 +1263,64 @@ char *nm_uri_from_query(CONTEXT *ctx, char *buf, size_t bufsz)
...
+char *nm_uri_from_windowed_query(CONTEXT *ctx, char *buf, size_t bufsz, char *timebase, int duration)
+{
+	dprint(2, (debugfile, "nm_uri_from_windowed_query (%s, %s, %d)\n", buf, timebase, duration));
+
+	int beg = duration*(NotmuchQueryWindowCurrentPosition+1);
+	int end = duration*NotmuchQueryWindowCurrentPosition;
+
+	if (NULL == NotmuchQueryWindowCurrentSearch) {
+		nm_query_window_reset();
+		FREE(&NotmuchQueryWindowCurrentSearch);
```
flatcap: If `NotmuchQueryWindowCurrentSearch` is NULL, then you don't need to FREE() it.

guyzmo: good point 😀

PR #279, opened 2017-01-02
add options to disable fake Folder: and Tags: headers for neomutt messages

kwantam:
Hi folks,

At present, notmuch messages are always displayed with fake Tags: and Folder: headers. This takes up a couple lines of screen real estate, which is inconvenient (at least for me).

This is a pretty simple patch that adds two boolean options to disable these. Default behavior is the same as now (show both).

Thanks for your hard work!

### DIFF
--- a/copy.c
+++ b/copy.c
@@ -200,9 +200,7 @@ mutt_copy_hdr (FILE *in, FILE *out, LOFF_T off_start, LOFF_T off_end, int flags,
 
       /* note: CH_FROM takes precedence over header weeding. */
       if (!((flags & CH_FROM) && (flags & CH_FORCE_FROM) && this_is_from) &&
-	  (flags & CH_WEED) &&
-	  mutt_matches_ignore (buf, Ignore) &&
-	  !mutt_matches_ignore (buf, UnIgnore))
+	  (flags & CH_WEED) && mutt_matches_ignore (buf))
 	continue;
       if ((flags & CH_WEED_DELIVERED) &&
 	  ascii_strncasecmp ("Delivered-To:", buf, 13) == 0)
--- a/copy.c
+++ b/copy.c
@@ -431,7 +431,7 @@ mutt_copy_header (FILE *in, HEADER *h, FILE *out, int flags, const char *prefix)
   {
     /* Add some fake headers based on notmuch data */
     char *folder = nm_header_get_folder(h);
-    if (folder)
+    if (folder && !(option (OPTWEED) && mutt_matches_ignore ("folder")))
     {
       char buffer[LONG_STRING];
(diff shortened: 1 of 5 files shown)


guyzmo:
maybe just one comment, I'm not sure that the option names are the most self-explicit. Maybe `nm_show_folder` and `nm_show_tags` could be `nm_insert_header_folder` and `nm_show_header_tags`

kwantam:
No objections here. I can update if you'd like.

guyzmo:
It's just my own opinion, I'm leaving the last word to @flatcap ☺

flatcap:
Hi @kwantam, welcome to NeoMutt and thanks for the patch.

However...

I think it would be better to change the NotMuch code to use Mutt's ignore/unignore mechanism, e.g.
```
unignore tags folder
```

As you point out, these are 'fake' headers, so we'd need to duplicate [some code from parse.c](https://github.com/neomutt/neomutt/blob/neomutt/parse.c#L1350-L1351).

This would mean the headers will disappear for the default user, but I'd prefer to give them one line of config, than add two new config options.

Alternatively, it might be possible to add a couple of default entries to the "unignore" list, so the headers default to "visible".

Would you be interested in some experimentation?

/cc @karelzak (author of the NotMuch code)

kwantam:
Sure, that would be fine. I'll try to find some time in the next couple days to play around with this. (Need to learn more about the ignore/unignore features to implement this.)

flatcap:
Thanks.
If you have any questions, just ask (here, or on our IRC channel #neomutt on freenode.net)

kwantam:
OK, new patch uses ignore/unignore to handle Folder and Tags headers.

First commit is a little refactoring: renamed `mutt_matches_ignore` to `mutt_matches_list` and added `mutt_matches_ignore` that checks for presence in Ignored and non-presence in UnIgnored. The commit also fixes the call sites using the original function.

Second commit adds call to `mutt_matches_ignore` in the right place in `copy.c` to make the fake headers work with Ignored / UnIgnored.

If you'd like, we can add a couple lines in init.c to turn on Folder and Tags by default. I originally had this, but removed it because I found that Debian's /etc/Muttrc overrides the ignore/unignore list by default, and I'm guessing that most other distributions will do the same.

karelzak:
Sounds like a good idea. No objections.

flatcap:
Lovely job.  Thanks.

> If you'd like, we can add a couple lines in init.c to turn on Folder and Tags by default. I originally had this, but removed it because I found that Debian's /etc/Muttrc overrides the ignore/unignore list by default.

Ah, I thought that might happen.  Never mind.
I'll add an explanation to the release notes for the next version.

PR #283, opened 2017-01-05
Relative ranges and goodies in oldie absolute ranges

nobrowser:
@flatcap 

Here is the first chunk of work on the relative range patterns.

What it does: it implements your proposal completely (IMO) , but _only_ for the relative syntax.  As of now, I've left the old absolute ranges exactly as they were - no new code, no new syntax.

I want you to criticize this part first so I get a sense of your taste in code, before I proceed to mess up the existing code path.  (And eventually write docs - cough.)

Also let me know if I need to clean up the git side.  (I re-based on neomutt branch once and somehow I ended up with two copies of the commits - I don't know why or how to fix it :-0 )

### DIFF
--- a/curs_main.c
+++ b/curs_main.c
@@ -880,6 +880,9 @@ int mutt_index_menu (void)
       unset_option (OPTREDRAWTREE);
     }
 
+    if (Context)
+      Context->menu = menu;
+
     if (Context && !attach_msg)
     {
       int check;
--- a/doc/manual.xml.head
+++ b/doc/manual.xml.head
@@ -5967,6 +5967,39 @@ string.  This example matches messages whose senders are known aliases.
 </screen>
 </example>
 
+<sect3 id="message-ranges">
+  <title>Message Ranges</title>
+
+  <para>
+    description of range types -- absolute, relative
+    mnemonics: ^ . $
+    closed vs open ranges
+  </para>
+
+  <table id="message-range-examples">
+    <title>Message Range Examples</title>
+    <tgroup cols="2">
+      <thead>
+        <row>
+          <entry>Range</entry>
+          <entry>Start</entry>
+          <entry>End</entry>
+          <entry>Description</entry>
+        </row>
+      </thead>
+      <tbody>
+        <row>
+          <entry>-3,2</entry>
+          <entry>Previous three emails</entry>
+          <entry>Next two emails</entry>
+          <entry>Relative to current email</entry>
+        </row>
+      </tbody>
+    </tgroup>
+  </table>
+
+</sect3>
+
 </sect2>
 
 <sect2 id="simple-searches">
--- a/mutt.h
+++ b/mutt.h
@@ -124,12 +124,6 @@ typedef struct
   int destroy;	/* destroy `data' when done? */
 } BUFFER;
 
-typedef struct
-{
-  int ch; /* raw key pressed */
-  int op; /* function op */
-} event_t;
-
 /* flags for _mutt_system() */
 #define MUTT_DETACH_PROCESS	1	/* detach subprocess from group */
 
@@ -1005,6 +999,8 @@ struct mx_ops
   int (*open_new_msg) (struct _message *, struct _context *, HEADER *);
 };
 
+#include "mutt_menu.h"
+
 typedef struct _context
 {
   char *path;
@@ -1032,6 +1028,8 @@ typedef struct _context
   int deleted;			/* how many deleted messages */
   int flagged;			/* how many flagged messages */
   int msgnotreadyet;		/* which msg "new" in pager, -1 if none */
+
+  MUTTMENU *menu;               /* needed for pattern compilation */
 
   short magic;			/* mailbox type */
 
--- a/mutt_curses.h
+++ b/mutt_curses.h
@@ -76,6 +76,12 @@ void mutt_curs_set (int);
 #define CI_is_return(c) ((c) == '\r' || (c) == '\n')
 #endif
 
+typedef struct
+{
+  int ch; /* raw key pressed */
+  int op; /* function op */
+} event_t;
+
 event_t mutt_getch (void);
 
 void mutt_endwin (const char *);
@@ -146,8 +152,8 @@ typedef struct color_line
   regex_t rx;
   int match; /* which substringmap 0 for old behaviour */
   char *pattern;
-  pattern_t *color_pattern; /* compiled pattern to speed up index color
-                               calculation */
+  struct pattern_t *color_pattern; /* compiled pattern to speed up index color
+                                      calculation */
   short fg;
   short bg;
   int pair;
--- a/pattern.c
+++ b/pattern.c
@@ -36,6 +36,7 @@
 #include "mutt_crypt.h"
 #include "mutt_curses.h"
 #include "group.h"
+#include "mutt_menu.h"
 
 #ifdef USE_IMAP
 #include "mx.h"
@@ -49,6 +50,7 @@
 static int eat_regexp (pattern_t *pat, BUFFER *, BUFFER *);
 static int eat_date (pattern_t *pat, BUFFER *, BUFFER *);
 static int eat_range (pattern_t *pat, BUFFER *, BUFFER *);
+static int eat_message_range (pattern_t *pat, BUFFER *s, BUFFER *err);
 static int patmatch (const pattern_t *pat, const char *buf);
 
 static const struct pattern_flags
@@ -79,7 +81,7 @@ Flags[] =
   { 'k', MUTT_PGP_KEY,		0,		NULL },
   { 'l', MUTT_LIST,		0,		NULL },
   { 'L', MUTT_ADDRESS,		0,		eat_regexp },
-  { 'm', MUTT_MESSAGE,		0,		eat_range },
+  { 'm', MUTT_MESSAGE,		0,		eat_message_range },
   { 'n', MUTT_SCORE,		0,		eat_range },
   { 'N', MUTT_NEW,			0,		NULL },
   { 'O', MUTT_OLD,			0,		NULL },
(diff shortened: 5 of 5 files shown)


flatcap:
> Also let me know if I need to clean up the git side. (I re-based on neomutt branch once and somehow I ended up with two copies of the commits - I don't know why or how to fix it :-0 )

(abridged log)
```
* 295f78a Implement vimified relative ranges as proposed by flatcap -- Ian Zimmerman
*   d03881f Merge branch 'devel/rel-msgno-pattern' of ssh://github.com/neomutt/neomutt into devel/rel-msgno-pattern -- Ian Zimmerman
|\
| * 4450505 light tidying -- Richard Russon
| * 2d43049 restore mutt_limit_current_thread -- Richard Russon
| * 776eb46 Fix build. -- Ian Zimmerman
| * 4ab20e9 Apply patch to add relative message number patterns -- Ian Zimmerman
* | 09518c1 First attempt at new relative range syntax -- Ian Zimmerman
* | 838deb4 light tidying -- Richard Russon
* | c34000d restore mutt_limit_current_thread -- Richard Russon
* | f1cb86b Fix build. -- Ian Zimmerman
* | fca12ec Apply patch to add relative message number patterns -- Ian Zimmerman
* | de6e7e1 notmuch: make 'Folder' and 'Tags' respect (un)ignore -- Riad Wahby
* | [lots and lots of commits...]
|/  
* a0239ad build: generate devel version suffix -- Richard Russon
```

Hmm...  It looks like you got it right to start with.
- You'd rebased your commits (fca12ec..838deb4) onto the latest version of neomutt (de6e7e1)
- Made a new commit (09518c1)
- Then you unexpectedly merge an old branch (4450505)
- Made another new commit (295f78a)

OK, that's fixable -- we need to delete the merge commit.  Here's how:

```
git clone https://github.com/neomutt/neomutt fix
cd fix

git checkout -b devel/rel-msgno-pattern origin/devel/rel-msgno-pattern

git rebase --interactive --preserve-merges HEAD~2
# In the editor, delete the first line (the merge commit)

# Check the results
git log --oneline neomutt..

    170394e Implement vimified relative ranges as proposed by flatcap
    09518c1 First attempt at new relative range syntax
    838deb4 light tidying
    c34000d restore mutt_limit_current_thread
    f1cb86b Fix build.
    fca12ec Apply patch to add relative message number patterns

# Check the code matches:
git diff devel/rel-msgno-pattern origin/devel/rel-msgno-pattern

    [no output]

# Finally, rebase onto the latest neomutt commits (from yesterday):
git rebase neomutt

# Forcibly save the changes to GitHub
git push --force origin devel/rel-msgno-pattern
```

nobrowser:
I squashed it successfully into a single commit, and I tried to clean up wrt. build, but mutt_limit_current_thread continues to plague me :( .  The problem is that mutt.h includes mutt_menu.h before the HEADER type is declared (in mutt.h), but that function has HEADER in its signature.

Let me know if you have ideas.  I suspect the proper solution is splitting off another header file, sigh.

nobrowser:
Here's one dirty way to "solve" it.

flatcap:
> mutt_limit_current_thread continues to plague me

The header files are a desperate mess :-(
I moved the definition to `protos.h` (which is where the other `pattern.c` functions live).

I've just pushed 5 small changes -- the first three are light tidying and self-explanatory.
- 0f00b2c tidy indent and trailing whitespace
- 3851f66 drop void cast
- ef96eae braces on new line

Next, is a personal preference, the log message says more:
- 5f0b671 bracket logic

Finally, a small change to the ranges.  To my mind, I expected the range `-1,1` to be the previous, current and next emails.
- 4bdeca5 extend ranges

flatcap:
20 few lines of code: Nice refactor!  :-)

The fake return type in `report_regerror()` is a bit odd, but it does make the caller look tidier.

I'd consider swapping the order of these if clauses, so that the comment "we're looking at a relative pattern" is followed by `eat_range_relative()`.
```c
/* We're looking at a new-style (relative) pattern iff a comma occurs in it */
if (strchr(s->dptr, ',') == NULL)
  tmp = eat_range_fixed(pat, s);
else
{ 
  tmp = eat_range_relative(pat, s, err);
  if (!tmp)
    return -1;
} 
```

So what's next?  Adding mnemonics to the fixed range?

nobrowser:
I am currently doing yet more cleanups to prepare for taking over the absolute code path.  Almost ready for that, probably will get there tomorrow.
The chunk you quote above won't survive that makeover, for sure.

nobrowser:
The fixed return value is a trick I learned and adopted long time ago.  You'd better get used to it, LOL

nobrowser:
It's almost all there now.  The only form I don't handle yet is a single bare number.

flatcap:
I've just given the absolute ranges (with mnemonics) a test.  Looking good.

I've spotted one problem with the range-order-swapping.
If a backwards range includes `$` the result is wrong.
Range `4,$` works, but `$-4` doesn't (it highlights `1-4`)

This stems from Mutt using `-1` for the MAX range.
Can it be changed to INT_MAX?
Where else is it used and compared, in Mutt?

I _know_ I asked for the ranges to be interchangeable, but perhaps it was wrong decision.
If you think we'd be better off without them, you can tell me :-)

One final request:  Now that `~m` accepts a single number, please can you extend that to include the mnemonics, so ranges: `^` and `.` and `$` work.

Cheers.

nobrowser:
Thinking about it, will be back

nobrowser:
The glitch you found is just a bug in my commit 68020b4.  It can be fixed and I'll do it.
Adding the magic chars may be more complex :-P:

nobrowser:
This is funny: the `MUTT_MAXRANGE` convention is in fact a problem, but with your 2nd request, not with the 1st.

I can easily add the magic characters to the syntax and `~m $` will be correctly "compiled" into the range bounds `{MUTT_MAXRANGE, MUTT_MAXRANGE}` i.e. `{-1, -1}`.  But mutt doesn't understand these bounds and tags everything.  This is already a problem with `~m $-$` in fact.

We have 2 options: either a bit of detective work to find the code that interprets the ranges, and make it understand `MUTT_MAXRANGE` as a lower bound.  Or I could find a way to translate `$` into the highest message number at pattern compilation time.  Of course in this latter case it better not be possible to save a pattern and apply it later - with different number of messages.  Do you know of any existing feature doing that?

flatcap:
Notes and questions in no discernible order:

How is the range stored / passed about?

I'm guessing that whoever uses the range must do _some_ range checking, so `{ -1, X }` will always start at `0` when applied to an array.  I don't know where the ranges are applied, but I reckon `{ X, INT_MAX }` (from limits.h), will always be truncated to however many messages are there.  As long as nobody changes the range to the _actual_ number of messages, we'll be OK.

nobrowser:
I think this nails it.  Light testing welcome :-)

flatcap:
> Light testing welcome

**Light** testing?  Pah!

I've performed 236 tests, covering every possible combination (I think).

These tests apply to a mailbox with 20 emails in it and the 10th email selected.

I've tested:
- Absolute offsets of 0, 1 (or ^), 5, 10 (or .), 15, 20 (or $), 99 (beyond the end).
- Relative offsets of -99 (before the beginning), -9 (or ^), -5, 0 (or .), 5, 10 (or $), 99 (beyond the end).
- Single character/number ranges

**Single Character**: `0` `1` `^` `5` `10` `.` `15` `20` `$` `99`

**Absolute Ranges**: `-^` `-1` `-5` `-10` `-.` `-15` `-20` `-99` `-$` `-` `0-^` `0-1` `0-5` `0-10` `0-.` `0-15` `0-20` `0-99` `0-$` `0-` `^-^` `^-1` `^-5` `^-10` `^-.` `^-15` `^-20` `^-99` `^-$` `^-` `1-^` `1-1` `1-5` `1-10` `1-.` `1-15` `1-20` `1-99` `1-$` `1-` `5-5` `5-7` `5-10` `5-.` `5-15` `5-20` `5-99` `5-$` `5-` `10-10` `10-.` `10-15` `10-20` `10-99` `10-$` `10-` `.-.` `.-10` `.-15` `.-20` `.-99` `.-$` `.-` `15-15` `15-17` `15-20` `15-99` `15-$` `15-` `20-20` `20-99` `20-$` `20-` `$-$` `$-` `99-99` `99-`

**Absolute Ranges (Backward)**: `7-^` `7-1` `7-5` `10-^` `10-1` `10-5` `10-7` `.-^` `.-1` `.-5` `.-7` `17-^` `17-1` `17-5` `17-7` `17-10` `17-.` `17-15` `20-^` `20-1` `20-5` `20-7` `20-10` `20-.` `20-15` `20-17`

**Relative Ranges**: `^,-99` `-9,-99` `-3,-99` `-3,^` `-3,-9` `-3,-5` `0,-99` `0,^` `0,-9` `0,-5` `.,-99` `.,^` `.,-9` `.,-5` `5,-99` `5,^` `5,-9` `5,-5` `5,0` `5,.` `5,3` `10,-99` `10,^` `10,-9` `10,-5` `10,0` `10,.` `10,5` `$,-99` `$,^` `$,-9` `$,-5` `$,0` `$,.` `$,5` `99,-99` `99,^` `99,-9` `99,-5` `99,0` `99,.` `99,5` `99,10` `99,$`

**Relative Ranges (Backward)**: `,-99` `,^` `,-9` `,-5` `,0` `,.` `,5` `,10` `,$` `,99` `,` `-99,-99` `-99,^` `-99,-9` `-99,-5` `-99,0` `-99,.` `-99,5` `-99,10` `-99,$` `-99,99` `-99,` `^,^` `^,-9` `^,-5` `^,0` `^,.` `^,5` `^,10` `^,$` `^,99` `^,` `-9,-9` `-9,-5` `-9,0` `-9,.` `-9,5` `-9,10` `-9,$` `-9,99` `-9,` `-5,-5` `-5,-3` `-5,0` `-5,.` `-5,5` `-5,10` `-5,$` `-5,99` `-5,` `0,0` `0,.` `0,5` `0,10` `0,$` `0,99` `0,` `.,0` `.,.` `.,5` `.,10` `.,$` `.,99` `.,` `5,5` `5,7` `5,10` `5,$` `5,99` `5,` `10,10` `10,$` `10,99` `10,` `$,$` `$,99` `$,` `99,99` `99,`

flatcap:
**Results**: 234/236 tests passed, 2 tests didn't do what I expected.

The two "failures" don't really bother me.
They are absolute ranges: `99-$` and `$-99`
I expected them to select the last entry, but they didn't.

If you're happy with the code, I'll look over it one last time before merging.

Also, are you happy for me to squash your 35 commits into one?

nobrowser:
Wait, I know of a bug (which may or may not have caused your test failures).  BRB.

I'll squash them myself when I think it's ready.

nobrowser:
Can you please repeat your battery of tests?

I took the other branch of the fork mentioned above, and I look up the high message number when I inspect the textual pattern.  This made it possible to get rid of the MUTT_MAXRANGE mess, for message number patterns.  It does mean that the parsing of this kind of range is now separate and distinct from other ranges (like size and score ranges).  I don't think many users will notice, but we'll need to at least hint at this in the doc.

flatcap:
Since the previous revision:
- Absolute ranges: `99-$` and `$-99` **now work correctly**.
- Single digit `0` used to select the current email, now it doesn't select anything.
  `0,0` selects the current email, as before.

There are two cases left that aren't perfect (but that I can live with): `,-99` and `99,` which select the first and last emails respectively.  I think they shouldn't select anything.

Anyway, I'm happy with the behaviour.  If you're not, here are my test config and script to play with :-)

```
color index green default ~T
macro index <F4> "\Ct.<enter>10<enter>"
macro index <F5> "T~m "
set sidebar_visible=no
set arrow_cursor
set index_format="%4C %Z %{%b %d} %-15.15L %s"
bind index <enter>  noop
bind index <return> noop
```

This is my test script: [mutt.sh.txt](https://github.com/neomutt/neomutt/files/716790/mutt.sh.txt)
Two terminals:
- Left running the script
- Right running mutt (20 emails)

I "drive" mutt using xdotool to generate mouse movements and keystrokes.
Note: xdotool works under X, but not Wayland.

nobrowser:
Thanks :) I'll squash the resisting cases, yet.

nobrowser:
Here is a trick question ;-)

What do you expect 99,$ to select?

flatcap:
> What do you expect 99,$ to select?

- rephrased it becomes `beyond-the-end,last`
- range is backwards, so reorder to `last,beyond-the-end`
- map to emails, `20,20`

So, only the last email

nobrowser:
So why should `99,` be different?  My assumption was that the empty slot was always equivalent to specifying the respective end (`$` on the right, `^` on the left).  If you disagree, you'll have to propose an alternative rule.

flatcap:
[light dawns on @flatcap]
Ah, I see what you mean.
You're right: `99,` will highlight the last and `,-99` will highlight the first.
Sorry for the confusion.

nobrowser:
Ok! It's commit squashing time for me.  I'll do it when I'm too tired for more creative tasks today.

I still need to update the docs, but maybe that can be another commit.

nobrowser:
Squashed!  Let me know if you'd like to have the doc fixes before merging :-)

flatcap:
Lovely job.

> Let me know if you'd like to have the doc fixes before merging

I can live without them for now, but I'll need them before the next release (probably next Friday).
I'll merge the feature and email the devel mailing list.

PR #284, opened 2017-01-06
Do a buffy check after shell escape

nobrowser:
Here's a trivial patch that made my user experience much better.

When the command called from a shell escape is something like fetchmail, rsync, or (in my case - I kid you not) unison, it helps to see which mailboxes were affected, immediately.

(I think the space kink in the diff is not my fault - looks like it was indented just 1 space before.)

### DIFF
--- a/commands.c
+++ b/commands.c
@@ -620,7 +620,8 @@ void mutt_shell_escape (void)
       mutt_endwin (NULL);
       fflush (stdout);
       if (mutt_system (buf) != 0 || option (OPTWAITKEY))
-	mutt_any_key_to_continue (NULL);
+        mutt_any_key_to_continue (NULL);
+      mutt_buffy_check(1);
     }
   }
 }


guyzmo:
hum… I'm not sure this is a good idea to implement as a patch. 👎
You should better use a macro that does your shell command followed by a `sync-mailbox`.

The issue here is that in the current situation you *can* make a macro to have what you want. But with your patch, it becomes impossible to prevent the buff-check after the sync.

nobrowser:
guyzmo, have you actually made a macro to do this?  I tried, and I couldn't make it work.  I think the switch from curses mode to normal line-buffered and back somehow discards all the stuffed command input in mutt.

If you show me a working macro I'll withdraw this proposal.

nobrowser:
By the way, I also don't see how doing the buffy check in this situation might be undesirable.  It _will _ happen anyway within the time given by mail_check; this just avoids having to wait some seconds or minutes for it.

But anyway, if I see this can really be done with a macro I'm fine with scrapping the PR.

nobrowser:
And does `sync-mailbox` check _all_ mailboxes?  Or just the one I'm looking at?

guyzmo:
ok, I just double checked the code, actually `buffy_check` only refreshes the count on the mailboxes, not an actual sync.

Sorry for the confusion, I vote then 👍 as it's mostly harmless, and I don't think it'll have much impact in terms of performance (maybe except for huge mbox, but anyway… whoever have a huge mbox is crazy ☺).

PR #285, opened 2017-01-06
Searching with a window over notmuch vfolders (part 2)

guyzmo:
following up on #278

Instead of adding a new command for searches that will result in
windowed searches, this commit patches a lower function of the search
mechanism, the `get_query_string()` function.

There, any search is buffered in the `nm_query_window_current_search`
global variable. And a newly introduced non-exposed function
`window_query_from_query` is taking a `data->db_query` and adds a
`date:` field implementing the window.

That function won't add a `date:` field if there's no duration set or if
there's already a date field in the query (to avoid risks of an empty
intersection).

To disable the feature, just set `nm_query_window_duration` to `0`.

Finally, only two public functions are exposed:

- `vfolder-window-forward`
- `vfolder-window-backward`

to slide the window over the search.

The improvements over #278 implementation is that now it's working with
any search done with notmuch, whether it's a `virtual-mailbox` or using
the `vfolder-from-query` command.

Have been removed:

- `windowed-vfolder-from-query`

Have been renamed:

- `windowed-vfolder-forward` → `vfolder-window-forward`
- `windowed-vfolder-backward` → `vfolder-window-backward`


guyzmo:
♥

iamkafai:
The vfolder-window is nice.

However, how is the existing '?limit=2000' query suppose to work now?  My existing limit in the uri stops working in my homebrew install at ~Jan/Feb timeframe.

Also, how to selectively en/disable window search in some virtual-mailboxes?

PR #286, opened 2017-01-08
fix coverity warnings

flatcap:


crypt-gpgme.c
```diff
@@ -2097,13 +2097,13 @@ static int pgp_gpgme_extract_keys (gpgme_data_t keydata, FILE** fp, int dryrun)
...
                  gpgme_pubkey_algo_name (subkey->pubkey_algo), subkey->length,
                  shortid, date, uid->uid);
       else
-        fprintf (*fp, "%s %5.5s %d/%8s %s\n", more ? "sub" : "pub",
+        fprintf (*fp, "sub %5.5s %d/%8s %s\n",
                  gpgme_pubkey_algo_name (subkey->pubkey_algo), subkey->length,
-                 shortid, date);      
+                 shortid, date);
       subkey = subkey->next;
       more = 1;
```
flatcap: For these two `fprintf`s, we know whether `more` is true or false, so we can drop the first parameter.
The "pub" and "sub" would only need to be params if they were being translated.

crypt-gpgme.c
```diff
@@ -3578,15 +3578,9 @@ static void print_key_info (gpgme_key_t key, FILE *fp)
...
-            s = "?";
+          s = gpgme_pubkey_algo_name (subkey->pubkey_algo);
 
-	  if (subkey)
-	    aval = subkey->length;
-	  else
-	    aval = 0;
+          aval = subkey->length;
 
           /* L10N: DOTFILL */
```
flatcap: The `for` loop ends if `subkey` is NULL, so these checks are redundant.

hcache-lmdb.c
```diff
@@ -99,6 +99,7 @@ hcache_lmdb_open(const char *path)
   if (rc != MDB_SUCCESS)
   {
     fprintf(stderr, "hcache_open_lmdb: mdb_env_create: %s", mdb_strerror(rc));
+    FREE(&ctx);
     return NULL;
   }
```
flatcap: Fix resource leak on error path.

hcache.c
```diff
@@ -719,19 +719,14 @@ static char* get_foldername(const char *folder)
...
-  if (stat (path, &st) == 0)
-  {
-    p = safe_malloc (PATH_MAX+1);
-    if (!realpath (path, p))
-      mutt_str_replace (&p, path);
-  } else
-    p = safe_strdup (path);
+  p = safe_malloc (PATH_MAX+1);
+  if (!realpath (path, p))
+    mutt_str_replace (&p, path);
```
flatcap: TOCTOU - Time of check vs Time of use -- if you check a filename exists, you can't be sure it'll be there later.
Don't `stat()` the file, just run `realpath()` on it.  If it doesn't exist, `realpath()` will tell us.

hcache.c
```diff
@@ -740,12 +735,11 @@ header_cache_t *
...
-  header_cache_t *h = safe_calloc(1, sizeof (header_cache_t));
-  struct stat sb;
-
   if (!ops)
     return NULL;
 
+  header_cache_t *h = safe_calloc(1, sizeof (header_cache_t));
+
   /* Calculate the current hcache version from dynamic configuration */
   if (hcachever == 0x0) {
```
flatcap: Reorder to avoid memory leak on error path.

hcache.c
```diff
@@ -799,7 +793,7 @@ mutt_hcache_open(const char *path, const char *folder, hcache_namer_t namer)
   else
   {
     /* remove a possibly incompatible version */
-    if (!stat (path, &sb) && !unlink (path))
+    if (unlink (path) == 0)
     {
       h->ctx = ops->open(path);
```
flatcap: TOCTOU - Time of check vs Time of use.
Don't `stat()` the file, just `unlink()` it.

imap/command.c
```diff
@@ -782,7 +782,7 @@ static void cmd_parse_lsub (IMAP_DATA* idata, char* s)
   url.path[strlen(url.path) - 1] = '\0';
   if (!mutt_strcmp (url.user, ImapUser))
     url.user = NULL;
-  url_ciss_tostring (&url, buf + 11, sizeof (buf) - 10, 0);
+  url_ciss_tostring (&url, buf + 11, sizeof (buf) - 11, 0);
   safe_strcat (buf, sizeof (buf), "\"");
   mutt_buffer_init (&token);
```
flatcap: Potential buffer overrun.

mutt_notmuch.c
```diff
@@ -152,12 +152,10 @@ static int url_parse_query(char *url, char **filename, struct uri_tag **tags)
...
 	if (!e)
-		return 0;
+		return 0;	/* only filename */
 
 	if (*filename && url_pct_decode(*filename) < 0)
 		goto err;
-	if (!e)
-		return 0;	/* only filename */
 
 	++e;	/* skip '?' */
```
flatcap: Remove redundant check.  `if (!e)` is done twice!

mutt_notmuch.c
```diff
@@ -269,9 +267,7 @@ static struct nm_ctxdata *new_ctxdata(char *uri)
 
 	if (url_parse_query(uri, &data->db_filename, &data->query_items)) {
 		mutt_error(_("failed to parse notmuch uri: %s"), uri);
-		data->db_filename = NULL;
-		data->query_items = NULL;
-		data->query_type = 0;
+		FREE(&data);
 		return NULL;
 	}
```
flatcap: Resource leak on error path.

mutt_notmuch.c
```diff
@@ -924,6 +920,8 @@ static void nm_progress_reset(CONTEXT *ctx)
 		return;
 
 	data = get_ctxdata(ctx);
+	if (!data)
+		return;
 
 	memset(&data->progress, 0, sizeof(data->progress));
```
flatcap: `get_ctxdata()` could return NULL

mutt_notmuch.c
```diff
@@ -936,7 +934,7 @@ static void nm_progress_update(CONTEXT *ctx, notmuch_query_t *q)
 {
 	struct nm_ctxdata *data = get_ctxdata(ctx);
 
-	if (ctx->quiet || data->noprogress)
+	if (ctx->quiet || !data || data->noprogress)
 		return;
```
flatcap: `get_ctxdata()` could return NULL

mutt_notmuch.c
```diff
@@ -970,9 +968,13 @@ static void append_message(CONTEXT *ctx,
...
+	struct nm_ctxdata *data = get_ctxdata(ctx);
+	if (!data)
+		return;
+
 	/* deduplicate */
 	if (dedup && get_mutt_header(ctx, msg)) {
-		get_ctxdata(ctx)->ignmsgcount++;
+		data->ignmsgcount++;
 		nm_progress_update(ctx, q);
 	        dprint(2, (debugfile, "nm: ignore id=%s, already in the context\n",
```
flatcap: `get_ctxdata()` could return NULL

pattern.c
```diff
@@ -1288,7 +1288,6 @@ mutt_pattern_exec (struct pattern_t *pat, pattern_exec_flag flags, CONTEXT *ctx,
         }
         return pat->not ^ result;
       }
-      return (pat->not ^ (h->env->x_label && patmatch (pat, h->env->x_label) == 0));
 #ifdef USE_NOTMUCH
     case MUTT_NOTMUCH_LABEL:
```
flatcap: Unreachable statement.

mutt_notmuch.c
```diff
@@ -1469,10 +1471,13 @@ static int rename_maildir_filename(const char *old, char *newpath, size_t newsz,
...
-	if (p)
+	if (p) {
 		*p = '\0';
+		p++;
+	} else {
+		p = folder;
+	}
 
-	p++;
 	strfcpy(filename, p, sizeof(filename));
```
flatcap: Potential problem if `folder` didn't contain a '/'

pop_lib.c
```diff
@@ -436,7 +436,8 @@ int pop_query_d (POP_DATA *pop_data, char *buf, size_t buflen, char *msg)
   mutt_socket_write_d (pop_data->conn, buf, -1, dbg);
 
   c = strpbrk (buf, " \r\n");
-  *c = '\0';
+  if (c)
+    *c = '\0';
   snprintf (pop_data->err_msg, sizeof (pop_data->err_msg), "%s: ", buf);
```
flatcap: `strpbrk()` could return NULL.

sidebar.c
```diff
@@ -532,7 +532,7 @@ static int prepare_sidebar (int page_size)
  */
 static int draw_divider (int num_rows, int num_cols)
 {
-  if ((num_rows < 1) || (num_rows < 1))
+  if ((num_rows < 1) || (num_cols < 1))
     return 0;
```
flatcap: Typo

sort.c
```diff
@@ -220,7 +220,9 @@ static int compare_spam (const void *a, const void *b)
   {
     result = strcmp(aptr, bptr);
     if (result == 0)
+    {
       AUXSORT(result, a, b);
+    }
   }
```
flatcap: Someone changed the `AUXSORT()` macro.  It now contains multiple statements meaning this `if` doesn't do what it used to.

system.c
```diff
@@ -129,8 +129,11 @@ int _mutt_system (const char *cmd, int flags)
...
 
-  sigaction (SIGCONT, &oldcont, NULL);
-  sigaction (SIGTSTP, &oldtstp, NULL);
+  if (!(flags & MUTT_DETACH_PROCESS))
+  {
+    sigaction (SIGCONT, &oldcont, NULL);
+    sigaction (SIGTSTP, &oldtstp, NULL);
+  }
 
   /* reset SIGINT, SIGQUIT and SIGCHLD */
```
flatcap: Only reset the signals if we set them.

hcache.c
guiniol: Aren't you missing a `struct stat sb;`? From the code it seems it's not used in the function, but since you deleted, I'd rather make sure. Other than that, LGTM!

flatcap: Ah, that's from the following change where I drop the call to `stat()`

flatcap:
Thanks for the review @guiniol.

PR #287, opened 2017-01-09
Add recursive mkdir function

flatcap:
@gahr 
I've altered the hcache to use the new `mutt_mkdir()` function.
I've tested:
- HAVE_ICONV & hcache directory
- HAVE_ICONV & hcache file
- !HAVE_ICONV & hcache directory
- !HAVE_ICONV & hcache file

They all seem to work correctly, but I'd appreciate a second opinion.
Cheers.


flatcap:
@gahr 
Since you last looked, I've completely refactored `mutt_hcache_per_folder()`.
The original was ~100 lines.  The new version, with extensive comments, is ~80 lines.

https://github.com/neomutt/neomutt/blob/devel/mkdir/hcache.c#L582

It's now much easier to see what directories are created and what filenames are used.

Again, if you have time, I'd appreciate your opinion on it.

Cheers.

PR #288, opened 2017-01-09
fixing vim syntax file for index-color commands

mrajner:
color higlighting of muttrc file in vim was not correct
for entries like:

color index_subject color... color...  "!~D ~sSPAM"

This was problem for all colors specified like `color index_...`
from patch-index-color-neomutt


flatcap:
Hi @mrajner, thanks for the patch.
I'll merge it once I understand it (that shouldn't take too long :-)

That file is actually just a copy of the upstream (vim) version.
We haven't reached out to the vim/neovim maintainers, yet.
When we do, I'll offer them the fix.

mrajner:
That's right, this file modified here is directly from vim. And it prevent higlighting color lines which starts with `color index...`. Probably it would be better idea to just touch `/doc/vimrc.index-color`.
I made such attempt here: https://github.com/mrajner/neomutt/commit/a388ebe02be7b98495267218b47439ff3f89fe5c
this has the same effect but do not touch `mutt-1.7.0-syntax.vim` file.

Probably it would be better to close this pull request and consider #289

Issue #162, opened 2016-09-25
Summary Pages


flatcap:
I'd like to create a set of summary pages for NeoMutt.

Vim-style
	:set     → show all modified vars
	:set all → all all vars
	:unset === set novar

:set -> pager full of settings
	list name, value, default value

:set VARIABLE=MAGIC_TOKEN => set VARIABLE to its default value
	save that default value in the history

:messages [clear]
	all messages, or limit to 100?

:bind / :macro / :???
	list bindings / macros

:color
	list all colours

:scriptnames
	list all config files that were loaded
	red-flag any with errors
	add cli option?  to save memory otherwise

for :set, etc how does help work when it overlays the pager?
	create a sample command
	then use the rest of the ideas for mentoring

:messages
	record progress bar info too, e.g.
	100% 3:45
	57% 1:24 cancelled

:messages should take ${read,write,net}_inc into account

:version in pager
	like vim's version
	or mutt -v command line

like noscript in firefox -- "neomutt newly updated message in inbox"
	or :neoversion => "fake email giving changelog"
	or :changelog

error message backlog
	cf vim :messages
	timestamped

mutt message backlog:
	quake like window
	promote to pager
	open in editor

:help
	show the formatted manual.txt
	=> generic <show-this-file> function

christopher-john-czettel:
will create the :test - prototype, as discussed in IRC

christopher-john-czettel:
Implemented basic scaffold for this feature-set. Even implemented :q :q! :qa and :quit commands with it. See 	968b5a6 for details. Will have to figure out how to effectively and elegantly handle the whole "opening a pager for this stuff" will work best. Will continue work sometime this week.

christopher-john-czettel:
Oh yeah: feedback and code-reviews while WIP always welcome! ;-)

flatcap:
I've had a quick glance at the code; it looks good so far.

> Even implemented :q :q! :qa and :quit commands

I had a feeling they'd show up :-)

I have a few more ideas/notes...

| Command  | Description |
|:---------|:----------------|
| about    | A brief description of the NeoMutt project |
| tips     | Enable "tips" mode, showing brief guides |
| credits | A list of the people who created NeoMutt |

**Ideas**:
- Opening a summary page, would close an existing page / help
- Status bar shows a list of other pages, when on a summary page

christopher-john-czettel:
just letting everyone know, that I will be working on this again. feedback and help always welcome.

[...]

PR #290, opened 2017-01-10
Support of relative paths sourcing and cyclic source detection

### LINKED ISSUE #282: Allow the 'source' command to use relative paths
Add the source command that will source external configuration files.

## Syntax

* to source a file located **relatively** to current configuration file:

```
source path/to/file
```

* to source a file in an **absolute** location:

```
source /path/to/file
```

as referenced by issues #281, #234 and #76.

guyzmo:
Added HEAP/FILO structure and global singleton to contain the the list
of config files that have been sourced, and detect cyclic sourcing.
Implemented with a simple heap API:

- `path_heap_push()`
- `path_heap_pop()`
- `path_heap_front()`
- `path_heap_find()`

Added a new method `to_absolute_path()` to convert a path into an absolute
path, relatively to another config file's path. Using POSIX methods to
do so, adding libgen's header in `mutt.h`.

fixes #282

### DIFF
--- a/init.c
+++ b/init.c
@@ -2607,23 +2607,117 @@ static int parse_set (BUFFER *tmp, BUFFER *s, unsigned long data, BUFFER *err)
   return (r);
 }
 
+/* Heap structure
+ * FILO designed to contain the list of config files that have been sourced
+ * and avoid cyclic sourcing */
+static struct config_path_heap {
+    char *path;
+    struct config_path_heap *prev;
+} *ConfigPathHeap;
+
+static void path_heap_push (char *path) {
+  struct config_path_heap *elt = safe_malloc(sizeof(struct config_path_heap));
+  elt->path = strdup(path);
+  elt->prev = ConfigPathHeap;
+  ConfigPathHeap = elt;
+}
+
+static int path_heap_pop () {
+  struct config_path_heap *elt = ConfigPathHeap;
+  if (!ConfigPathHeap)
+      return 0;
+  ConfigPathHeap = elt->prev;
+  free(elt->path);
+    free(elt);
+  return 1;
+}
+
+static const char *path_heap_front (void) {
+  if (!ConfigPathHeap || !ConfigPathHeap->path)
+      return "";
+  return ConfigPathHeap->path;
+}
+
+static int path_heap_find(char *path) {
+  for (struct config_path_heap* former=ConfigPathHeap; 
+       former != NULL;
+       former = former->prev)
+    if (strcmp(path, former->path) == 0)
+      return TRUE;
+  return FALSE;
+}
+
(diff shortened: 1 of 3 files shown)


guyzmo:
Currently the whole heap code is in `init.c` just next to the place of use. 
Either put the whole heap stuff inside its own compilation unit and/or plan the path to #162.
The good side of the heap abstraction I'm suggesting is that it can easily be switched.

guyzmo:
if you're ok with my changes, I'll move the heap stuff in `muttlib.c` along with `mutt_*_list`. ☺

guyzmo:
suggestion for changes:

- replace the `struct config_path_heap` with `LIST*`,
- keep a list of loaded script files in a global context,
- keep the heap static to the compilation unit,
- move all the data structure stuff in muttlib compilation unit

flatcap:
I can't get this to work at all.

Wherever I run mutt from, I get:
- Error: Cyclic sourcing of configuration file '/home/mutt/.muttrc'.

Adding `-F` doesn't help either.

guyzmo:
yup, my last commit did not help, a few preconditions that were designed in my heap algorithms went missing when I merged with muttlib's LIST implementation.

guyzmo:
@flatcap should be fixed, sorry for the issue!

flatcap:
I've just given it a good thrashing and everything works beautifully.

I tried config:
- relative to the current directory
- relative to other files
- mix and matching absolute and relative dirs
- a cycle of length 3
- a cycle of length 100
- an insane 4 dir deep, 775 config file loop of random sources

I just need to look at, and understand, the source.

guyzmo:
the implementation is rather straightforward:

* in `muttlib.c`
 * I've added a heap interface to the LIST structure (to keep DRY).
 * that interface adds a push front method (`mutt_push_heap()`) and a pop front method (`mutt_pop_heap()`)
* in `init.c`
 * I've added `to_absolute_path` that converts a given path string to absolute, relatively to a reference path string. That reference path shall always be a filename (the last part is of that path is removed).
 * the `source_rc()` function is taking a file path:
   * converts it to absolute relative to the file on the top of the heap, 
   * looks up if that file is already in the heap
   * if not, sources it.

So at the end of the execution, the Heap should be an empty pointer. TBH, the *one improvement* would be to make the pointer `MuttrcHeap` a static variable local to `source_rc()`.

flatcap:
@gahr found a problem when using pipes in a config file, e.g.
```
source "gpg -q -d ~/.mutt/password.gpg |"
```
gives the error: "Error in muttrc, line 6"

gahr:
The error is
`Error: issue converting path to absolute (No such file or directory)`

PR #293, opened 2017-01-12
Support of multiple config files as CLI arguments

### LINKED ISSUE #234: Allow Mutt to read multiple config files at startup
Currently, Mutt will only read one config file from the command line.

**Example**:

```
mutt -F apple.rc -F banana.rc -F cherry.rc
```

Only `cherry.rc` is read and acted upon.

guyzmo:
- changed the Muttrc global to a LIST*
- set it up to parse in order of parsing
- updated AliasFile to be a LIST* as well

fixes #234

### DIFF
--- a/alias.c
+++ b/alias.c
@@ -344,49 +344,51 @@ retry_name:
   else
     Aliases = new;
 
-  strfcpy (buf, NONULL (AliasFile), sizeof (buf));
-  if (mutt_get_field (_("Save to file: "), buf, sizeof (buf), MUTT_FILE) != 0)
-    return;
-  mutt_expand_path (buf, sizeof (buf));
-  if ((rc = fopen (buf, "a+")))
-  {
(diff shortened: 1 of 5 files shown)


guyzmo:
I've made as one shot and did not thoroughly check my patch, so please don't hesitate to check ☺
My test has been to create three bogus muttrc files and start muttrc, and I had:

```
 % for i in $(seq 1 3); do echo bogus$i > muttrc_test$i; done
 % ./mutt -F muttrc_test1 -F muttrc_test2 -F muttrc_test3
Error in muttrc_test1, line 1: bogus1: unknown command
source: errors in muttrc_test1
Error in muttrc_test2, line 1: bogus2: unknown command
source: errors in muttrc_test2
Error in muttrc_test3, line 1: bogus3: unknown command
source: errors in muttrc_test3
Press any key to continue...
```

PR #300, opened 2017-01-16
Added -S option to -D dump option to hide senstitive info

### LINKED ISSUE #236: Hide sensitive config
When fixing a bug, or helping a beginner, it's useful to see their config.
However, some items shouldn't be made public.

This user performs a safe -C config dump:

```
mutt -C | grep imap | sort
```

```
imap_authenticators=""
imap_check_subscribed is unset
imap_delim_chars="/."
imap_headers=""
imap_idle is set
imap_keepalive=120
imap_list_subscribed is unset
imap_login=""
imap_passive is set
imap_pass="**XXX**"
imap_peek is set
imap_pipeline_depth=15
imap_servernoise is set
imap_user="**XXX**"
```

By tagging certain config items (in `init.h`) as "sensitive", they can be obscured.

guyzmo:
- added MuttSensitiveVar list at the end of init.h
- added variable in main() to check for -S
- updated mutt_dump_variables

fixes #236

### DIFF
--- a/init.c
+++ b/init.c
@@ -3317,12 +3317,12 @@ int mutt_query_variables (LIST *queries)
 }
 
 /* dump out the value of all the variables we have */
-int mutt_dump_variables (void)
+int mutt_dump_variables (int hide_sensitive)
 {
-  int i;
-  
+  int i, j;
+
   char command[STRING];
-  
+
   BUFFER err, token;
 
   mutt_buffer_init (&err);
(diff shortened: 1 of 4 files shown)


gahr:
What I don't like of this approach is that the info about whether an option is sensitive is kept in a separate place as where the option is defined (MuttSensitiveVars instead of using flags).

guyzmo:
I prefer @gahr's approach of using a flag, so I'm closing this PR in favor of #302.

What I prefer in my approach though, is use of `-DS` switch.

PR #302, opened 2017-01-17
Add support for sensitive options and implement -D -S switch

gahr:
Issue: #236

### DIFF
--- a/init.h
+++ b/init.h
@@ -68,6 +68,7 @@
 #define R_SIDEBAR       (1<<7)  /* redraw the sidebar */
 #define R_BOTH		(R_INDEX | R_PAGER)
 #define R_RESORT_BOTH	(R_RESORT | R_RESORT_SUB)
+#define R_SENSITIVE     (1<<8)  /* hide in mutt -C */
 
 struct option_t
 {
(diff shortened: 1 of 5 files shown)


gahr:
TBD 
* rename `R_SENSITIVE` so it's not in the `R_*` namespace? If so, how to ensure bitwise flaggability?
* what exactly to print? `'***'`, `***`, `********`, `<***>`; should this be a config option? Overkill imho..

guyzmo:
another question:

- shall we go for `-C` or `-D -S` like [I implemented](https://github.com/neomutt/neomutt/pull/300/files#diff-2045016cb90d1e65d71c2407a2570927R306)?

init.h
```diff
@@ -68,6 +68,7 @@
 #define R_SIDEBAR       (1<<7)  /* redraw the sidebar */
 #define R_BOTH		(R_INDEX | R_PAGER)
 #define R_RESORT_BOTH	(R_RESORT | R_RESORT_SUB)
+#define R_SENSITIVE     (1<<8)  /* hide in mutt -C */
```
guyzmo: cf earlier comments, choose a prefix to properly namespace this, like `Q_` for qualifier…

gahr:
Yeah I'll take care of the docs as soon as we decide what option to implement ;)

gahr:
After much thought (...) I agree on the superiority of "-D -S" as opposed to "-C" :)

guyzmo:
one last thing, you need to update the [xml documentation](https://github.com/neomutt/neomutt/blob/neomutt/doc/manual.xml.head#L14147)

Issue #224, opened 2016-11-05
Create a suitable coding style option set via clang-format


toogley:
## introduction

After discussing the coding style with the main upstream developer (which is btw. Kevin J. McCarthy)[[1]](http://mailman.neomutt.org/pipermail/neomutt-devel-neomutt.org/2016-November/000094.html) [[2]](http://mailman.neomutt.org/pipermail/neomutt-devel-neomutt.org/2016-November/000093.html), he said, he would welcome an coding style enforcement via [clang-format](http://clang.llvm.org/docs/ClangFormat.html) (options: [link](http://clang.llvm.org/docs/ClangFormatStyleOptions.html)). 

## Why is `clang-format` better than `indent` ?

* Gnu indent had the last stable release in 2008/2009, which makes it obsolete. So i don't think most bugs are very likely to be fixed.
* To cite the [manual](https://www.gnu.org/software/indent/manual/indent.html#SEC15):

> indent does not understand C. In some cases this leads to the inability to join lines. The result is that running a file through indent is irreversible, even if the used input file was the result of running indent with a given profile (‘.indent.pro’). 

Clang-format **does** however understand C, because its part of the clang compiler.

## new coding style must be **really** an improvement.

For instance, 

```
$ clang-format -i -style=llvm  crypt-gpgme.c
```

produces, among others, the following diff:

```diff
-#define digitp(p)   (*(p) >= '0' && *(p) <= '9')
-#define hexdigitp(a) (digitp (a)                     \
-                      || (*(a) >= 'A' && *(a) <= 'F')  \
-                      || (*(a) >= 'a' && *(a) <= 'f'))
-#define xtoi_1(p)   (*(p) <= '9'? (*(p)- '0'): \
-                     *(p) <= 'F'? (*(p)-'A'+10):(*(p)-'a'+10))
-#define xtoi_2(p)   ((xtoi_1(p) * 16) + xtoi_1((p)+1))
+#define digitp(p) (*(p) >= '0' && *(p) <= '9')
+#define hexdigitp(a)                                                           \
+  (digitp(a) || (*(a) >= 'A' && *(a) <= 'F') || (*(a) >= 'a' && *(a) <= 'f'))
+#define xtoi_1(p)                                                              \
+  (*(p) <= '9' ? (*(p) - '0') : *(p) <= 'F' ? (*(p) - 'A' + 10)                \
+                                            : (*(p) - 'a' + 10))
[...]

PR #313, opened 2017-01-21
editorconfig

toogley:
This idea was mentioned by @guyzmo [in this issue](https://github.com/neomutt/neomutt/issues/224#issuecomment-269922902).

### DIFF
--- a/.editorconfig
+++ b/.editorconfig
@@ -1,0 +1,14 @@
+# top-most EditorConfig file
+root = true
+
+# Unix-style newlines with a newline ending every file
+[*]
+end_of_line = lf
+insert_final_newline = true
+charset = utf8
+trim_trailing_whitespace = true
+
+
+[{*.c,*.h}]
+indent_style = space
+indent_size = 2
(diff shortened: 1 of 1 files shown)


toogley:
just to be sure: the `indent_size = 2 ` setting doesn't interfere with tabs, right?

toogley:
@guyzmo  Do you know, what exactly do the [editorconfig docs](http://editorconfig.org/) mean by "columns used to represent a tab character." ?
Is one column one space or multiple spaces?

> **indent_size:** a whole number defining the number of columns used for each indentation level and the width of soft tabs (when supported). When set to "tab", the value of tab_width (if specified) will be used.
**tab_width:** a whole number defining the number of columns used to represent a tab character. This defaults to the value of indent_size and doesn't usually need to be specified.

guyzmo:
`tab_width` is what the representation of the tab would be in spaces one `\t` would be as large as `    ` if set to 4 or `  ` if set to 2.

`indent_size` the number of spaces in an indentation step.

Just make them both equal to `2`.

PR #317, opened 2017-01-22
Added waffle links to README

guyzmo:


toogley:
LGTM. 

Maybe should also add a graph like below?

[Throughput Graph](https://waffle.io/neomutt/neomutt/metrics/throughput)

guyzmo:
I'm not against. But it's not necessary to overload the readme with information… even more as the readme is stuffed under 6 pages of file listing!

toogley:
@guyzmo yep. But given that the history of mutt development was rather unfriendly at certain times, this graph may provide additional evidence that we're active. Therefore i think its not unnecessary.

guyzmo:
Maybe then on the neomutt.org website?

toogley:
@guyzmo its of course an option, but i don't know which one is better.

flatcap:
I've added the badges to the README as they're quite modest.
And, I've added the badges and graph to the [devel webpage](https://www.neomutt.org/dev.html).

PR #318, opened 2017-01-22
added gtags files to .gitignore

toogley:
The Source code tagging system [GNU Global](https://www.gnu.org/software/global/) (very similar to ctags and such) creates during its indexing step various binary files to save the results. Those files are git-ignored with this PR.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -65,3 +65,7 @@ txt2c
 
 neomutt-*.tar.gz
 
+# gtags things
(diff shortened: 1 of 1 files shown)


flatcap:
I think `.gitignore` should only contain the things that are likely to be built by neomutt, so this would be better off in `~/.gitignore`.

(unless you want to add a GNU Global config file and create a make target)

Issue #326, opened 2017-01-23
notmuch: Ability to toggle tag 


bbenne10:
Hoping I'm reaching out via proper avenues to make a feature request. If not, please let me know where I should be posting and I'll happily move this there.

Essentially what I am looking for is a more advanced form of the `modify-labels` function that exists. I would like a way to toggle the presence of a tag (or list of tags) on a message (or set of messages). I currently have a muttrc.keybinds that looks something like this ad infinitum:

```
macro index tt "<modify-labels>+todo\n" "Add tag todo"
macro index tT "<modify-labels>-todo\n" "Remove tag todo"
```
It'd be VERY nice if I could simply combine these two lines into one. Something along the lines of:

```
macro index "<toggle-labels>todo\n" "Toggle tag todo"
```

flatcap:
This is the right place :-)

How about another syntax that doesn't require a new command `!label`
```
macro index tt "<modify-labels>!todo\n" "Toggle tag todo"
```

It shouldn't be too hard to implement.

bbenne10:
In an attempt to help out here, I just looked up the implementation for `modify-labels` and my eyes immediately glazed over.  I should have expected some complexity from such a powerful tool, but that was eye opening. I think I'm going to have to leave this one to someone a bit more familiar with the project 😳

toogley:
> I should have expected some complexity from such a powerful tool

I think its just plain old and a lot of the complexity could be reduced (we are currently working hard to do so. But its hard.. we don't have many developers and a lot of the work is painful.) :D

flatcap:
Mutt, as a whole, is a bit of a mess, but the notmuch code is well-written.
To implement my syntax `!label` you would only need to alter [one function](https://github.com/neomutt/neomutt/blob/neomutt/mutt_notmuch.c#L1324-L1367) in mutt_notmuch.

Take a look.  If you're interested we can talk you through anything else you need to know.

guyzmo:
being a mutt+notmuch user, I can have a look at that, and see if I can provide a patch. I'll look into that a bit later (I'm a bit busy today to be able to do some code ☹).

guyzmo:
…or, @bbenne10, I can help you fixing it!

BTW, don't hesitate to join the `#neomutt` channel on irc.freenode.org IRC network.

bbenne10:
[...]

PR #328, opened 2017-01-24
Add ability to toggle notmuch tags on a message

bbenne10:
This addresses #326.


flatcap:
Many thanks.  Merged.

Issue #342, opened 2017-01-30
Crash closing a non-opened mailbox


gahr:
## How to reproduce
1. Make sure that trying to lock the `Postponed` folder is going to fail (e.g., set `dotlock_program` to some non-existent path).
2. Try to save a draft
3. Mutt crashes with `Couldn't lock /root/postponed?Segmentation fault (core dumped)`

## Reason
Failure to open an mbox causes it to be immediately fast-closed (see [here](https://github.com/neomutt/neomutt/blob/neomutt/mx.c#L637-L639)).

For mailbox, failing to lock the an mbox causes `mbox_open_mailbox_append` to [close it](https://github.com/neomutt/neomutt/blob/neomutt/mbox.c#L473). Later on, [closing it](https://github.com/neomutt/neomutt/blob/neomutt/mbox.c#L486) crashes because `ctx->fp` is `NULL`.

It is unclear to me where the error in the logic is, but either i) failing to open a mailbox should not close it, or ii) closing a mailbox should be resilient to the fact that the mailbox being closed is not open.

PR #343, opened 2017-01-30
Do not crash when closing a non-opened mailbox

gahr:
Issue: #342

### DIFF
--- a/mbox.c
+++ b/mbox.c
@@ -481,6 +481,11 @@ static int mbox_open_mailbox_append (CONTEXT *ctx, int flags)
 
 static int mbox_close_mailbox (CONTEXT *ctx)
 {
+  if (!ctx->fp)
+  {
+    return 0;
+  }
+
   if (ctx->append)
   {
     mx_unlock_file (ctx->path, fileno (ctx->fp), 1);


d-k-c:
Your fix is okay for now, but as you said in your issue the problem is in the logic: why is `close` called when `open` fails afterall? But this logic issue is way larger than the issue you intend to fix, so let's leave it for later.

riesebie:
Hi gents,

* Damien R. <notifications@github.com> [2017-01-30 08:11 -0800]:

> d-k-c approved this pull request.

* guiniol <notifications@github.com> [2017-01-30 09:48 -0800]:

> guiniol approved this pull request.
> 
> Looks good to me.

Why don't we manage those approvals by git? Just sign off the
patches. It would be good to manage patches in neomutt-devel as
well, or only?. We are working on a MUA and __not__ via unthreaded
and quite some non reproducible "issues" provided by a GitHub server
located in Trump-Land....

That are my thoughts for now ;-)

Elimar, who is a bit conservative on some circumstances, hehehehe.
-- 
  Numeric stability is probably not all that
  important when you're guessing;-)

flatcap:
> Why don't we manage those approvals by git? Just sign off the patches. 

It's a nice thought, to have the reviewers remembered.
However, it would mean two extra commits for every pull-request.
(unless you force the update, causing problems for others).

gahr:
As long as the issue is mentioned in the commit message, we'll be able to get back to the issue that resulted in the commit.

PR #345, opened 2017-01-30
Update documentation about modify-labels-then-hide

bbenne10:
This will inform new users regarding the need to call <sync-mailbox> after calling this function


bbenne10:
👍 Thanks for the quick turnaround!

guiniol:
Always a pleasure to have better documentation. So thank you!

PR #351, opened 2017-01-31
Handle BAD as IMAP_AUTH_UNAVAIL

### LINKED ISSUE #348: IMAP AUTH PLAIN fails if SASL is not available
Currently, the bundled SASL PLAIN implementation for IMAP doesn't differentiate between a NO (wrong credentials) and a BAD (invalid command, i.e., IMAP server knows nothing about SASL). Both responses are treated as authentication failures and the remaining authenticators are not tried.

This causes a regression when the user hasn't specified any `imap_authenticators` and the server supports only LOGIN.

gahr:
Closes #348

### DIFF
--- a/imap/auth_plain.c
+++ b/imap/auth_plain.c
@@ -31,6 +31,8 @@
 /* imap_auth_plain: SASL PLAIN support */
 imap_auth_res_t imap_auth_plain(IMAP_DATA *idata, const char *method)
 {
+  int rc;
+  imap_auth_res_t res = IMAP_AUTH_SUCCESS;
   char buf[STRING];
 
   if (mutt_account_getuser(&idata->conn->account))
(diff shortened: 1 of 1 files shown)


gahr:

imap/auth_plain.c
```diff
@@ -44,14 +46,23 @@ imap_auth_res_t imap_auth_plain(IMAP_DATA *idata, const char *method)
                       idata->conn->account.user, idata->conn->account.user,
                       idata->conn->account.pass);
 
-  if (!imap_exec(idata, buf, IMAP_CMD_FAIL_OK | IMAP_CMD_PASS))
+  imap_cmd_start(idata, buf);
+  do
+    rc = imap_cmd_step(idata);
+  while (rc == IMAP_CMD_CONTINUE);
+
```
flatcap: In `imap_exec()` this do-while is surrounded by:
```c
mutt_allow_interrupt(1);   
mutt_allow_interrupt(0); 
```

imap/auth_plain.c
```diff
@@ -44,14 +46,23 @@ imap_auth_res_t imap_auth_plain(IMAP_DATA *idata, const char *method)
...
                       idata->conn->account.pass);
 
-  if (!imap_exec(idata, buf, IMAP_CMD_FAIL_OK | IMAP_CMD_PASS))
+  imap_cmd_start(idata, buf);
+  do
+    rc = imap_cmd_step(idata);
+  while (rc == IMAP_CMD_CONTINUE);
+  
+  if (rc == -1)
   {
```
flatcap: `if (rc == IMAP_CMD_BAD)`

imap/auth_plain.c
```diff
@@ -44,14 +46,23 @@ imap_auth_res_t imap_auth_plain(IMAP_DATA *idata, const char *method)
...
-    mutt_clear_error(); /* clear "Logging in...".  fixes #3524 */
-    return IMAP_AUTH_SUCCESS;
+    res = IMAP_AUTH_UNAVAIL;
+  }
+  else if (rc == -2)
+  {
```
flatcap: `else if (rc == IMAP_CMD_NO)`

imap/auth_plain.c
gahr: No need here, I don't expect the while to run more than once.

guyzmo:
I'm not going to request changes, but I actually have a minor issue with this change: this patch is replicating what `imap_exec()` has been designed to do: [`imap/commands.c`](https://github.com/neomutt/neomutt/blob/neomutt/imap/command.c#L232,L265).

Instead, to follow DRY principle, you could be fixing `imap_exec()` by handling a new error case `IMAP_CMD_BAD`, and return appropriately, if need be using a flag from `flags` to handle that special case.

PR #352, opened 2017-01-31
If possible, only redraw after gpgme has invoked pinentry

### LINKED ISSUE #333: Double screen refresh during GPGME verification
When you open a PGP-signed email and you mutt verifies the signature using gpgme, a forced screen refresh happens.

The effect is going from index (where you `<enter>`'d the email) to a blank screen, to again index, and finally to pager, where the message is displayed.

This is due to [crypt-gpgme.c:811](https://github.com/neomutt/neomutt/blob/neomutt/crypt-gpgme.c#L811)

I do not understand why that line is there and would like to remove it (fixes the issue for me), but I'd like people to test commenting that line out and see whether anything bad happens, thus exposing the real purpose of the refresh.

gahr:
This requires gpgme >= 1.8.1, which exposes a new "redraw" status flag.
On older versions, fallback to always redraw if pinentry was potentially called.
Thanks to Werner Koch for the hints!

Closes #333

### DIFF
--- a/crypt-gpgme.c
+++ b/crypt-gpgme.c
@@ -127,18 +127,18 @@ static char *current_sender = NULL;
  * General helper functions.
  */
 
-/* accomodate for a redraw if needed */
+/* accommodate for a redraw if needed */
 static void
 redraw_if_needed (gpgme_ctx_t ctx)
 {
-#if GPGME_VERSION_NUMBER < 0x010800
+#if (GPGME_VERSION_NUMBER < 0x010800)
   /* gpgme_get_ctx_flag is not available in gpgme < 1.8.0. In this case, stay
    * on the safe side and always redraw. */
   (void)ctx;
   mutt_need_hard_redraw ();
 #else
   const char *s = gpgme_get_ctx_flag (ctx, "redraw");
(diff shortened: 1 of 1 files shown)


gahr:

crypt-gpgme.c
```diff
@@ -127,6 +127,24 @@ static char *current_sender = NULL;
...
+static void
+redraw_if_needed (gpgme_ctx_t ctx)
+{
+#if GPGME_VERSION_NUMBER < 0x010801
+  /* The "redraw" flag is not available in gpgme < 1.8.1. In this case, stay on
+   * the safe side and always redraw. */
+  (void)ctx;
+  mutt_need_hard_redraw ();
+#else
+  const char *s = gpgme_get_ctx_flag (ctx, "redraw");
```
d-k-c: Documentation of this function states that it returns NULL if "name" (second argument) is unknown, so it seems the #if block is not necessary

gahr: Oh, it's my comment which is wrong.. the reason for the ifdef block is that `gpgme_get_ctx_flag` didn't exist before 1.8.0. Let me fix that.

PR #353, opened 2017-01-31
Remove two use-after free in global hooks

### LINKED ISSUE #347: multiple timeout-hooks cause crash
This minimal config file causes mutt to crash.
**crash.rc**
```
timeout-hook 'exec beep'
timeout-hook 'exec previous-undeleted'
```

Steps to reproduce:
- start mutt
- source crash.rc
- wait for timeout

Reproducable?
- Always

guiniol:
Fixes #347
In both mutt_timeout_hook and mutt_startup_shutdown_hook, we iterate on
the list of hooks and try to execute the right ones, always using the
same buffer. The buffer must be free'd after all the hooks have been
checked and not after the first hook which fails. This also removes a
memory leak, because the buffer was never freed if no hook failed.

### DIFF
--- a/hook.c
+++ b/hook.c
@@ -587,7 +587,6 @@ void mutt_timeout_hook (void)
 
     if (mutt_parse_rc_line (hook->command, &token, &err) == -1)
     {
-      FREE (&token.data);
       mutt_error ("%s", err.data);
       mutt_sleep (1);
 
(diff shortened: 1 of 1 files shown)


guiniol:
@neomutt/reviewers ready for review!

PR #354, opened 2017-02-01
Simplify imap_keepalive + some cleanup

gahr:
1. introduce private `int imap_check (IMAP_DATA* idata, int force); `
2. turn `int imap_check_mailbox (CONTEXT* ctx, int* index_hint, int force);` into a wrapper around 1.
3. refactor and simplify `imap_keepalive`
4. remove some useless (and dangerous) casts
5. remove some unused parameters

### DIFF
--- a/imap/imap.h
+++ b/imap/imap.h
@@ -33,9 +33,9 @@ typedef struct
 
 /* imap.c */
 int imap_access (const char*, int);
-int imap_check_mailbox (CONTEXT *ctx, int *index_hint, int force);
-int imap_delete_mailbox (CONTEXT* idata, IMAP_MBOX mx);
-int imap_sync_mailbox (CONTEXT *ctx, int expunge, int *index_hint);
+int imap_check_mailbox (CONTEXT* ctx, int force);
+int imap_delete_mailbox (CONTEXT* ctx, IMAP_MBOX mx);
+int imap_sync_mailbox (CONTEXT *ctx, int expunge);
 int imap_close_mailbox (CONTEXT *ctx);
 int imap_buffy_check (int force, int check_stats);
 int imap_status (char *path, int queue);
--- a/imap/imap_private.h
+++ b/imap/imap_private.h
@@ -226,11 +226,9 @@ typedef struct
 } IMAP_DATA;
 /* I wish that were called IMAP_CONTEXT :( */
 
-/* -- macros -- */
-#define CTX_DATA ((IMAP_DATA *) ctx->data)
-
 /* -- private IMAP functions -- */
 /* imap.c */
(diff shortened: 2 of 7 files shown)


gahr:

imap/imap_private.h
```diff
@@ -226,11 +226,9 @@ typedef struct
 } IMAP_DATA;
 /* I wish that were called IMAP_CONTEXT :( */
 
-/* -- macros -- */
-#define CTX_DATA ((IMAP_DATA *) ctx->data)
```
d-k-c: Good riddance 👍

imap/imap.h
```diff
@@ -33,9 +33,9 @@ typedef struct
 
 /* imap.c */
 int imap_access (const char*, int);
-int imap_check_mailbox (CONTEXT *ctx, int *index_hint, int force);
-int imap_delete_mailbox (CONTEXT* idata, IMAP_MBOX mx);
-int imap_sync_mailbox (CONTEXT *ctx, int expunge, int *index_hint);
+int imap_check_mailbox (CONTEXT* ctx, int force);
+int imap_delete_mailbox (CONTEXT* ctx, IMAP_MBOX mx);
+int imap_sync_mailbox (CONTEXT *ctx, int expunge);
```
d-k-c: Here we have `CONTEXT* ctx` and `CONTEXT *ctx`, I know it's legacy but since you're modifying these prototypes, we can use this occasion to make them consistent.

imap/util.c
```diff
@@ -844,43 +844,20 @@ static void alrm_handler (int sig)
 void imap_keepalive (void)
 {
   CONNECTION *conn;
-  CONTEXT *ctx = NULL;
   IMAP_DATA *idata;
+  time_t now = time(NULL);
```
d-k-c: Missing space before parenthesis (but do we really want to enforce that?)

imap/util.c
```diff
@@ -844,43 +844,20 @@ static void alrm_handler (int sig)
...
-	   */
-	  need_free = 1;
-	}
-	/* if the imap connection closes during this call, ctx may be invalid
-	 * after this point, and thus should not be read.
-	 */
-	imap_check_mailbox (ctx, NULL, 1);
-	if (need_free)
-	  FREE (&ctx);
+        imap_check(idata, 1);
```
d-k-c: Ditto

imap/util.c
flatcap: I prefer the space before a parenthesis, but it seems that more people don't.

d-k-c: I prefer without space, but current style uses one

PR #358, opened 2017-02-03
Unicode 0x202F is a non-break space too

### LINKED ISSUE #357: Problem displaying narrow no-break spaces under freebsd 11
I’ve upgraded from freebsd 10.2 to freebsd 11 and since then I see 'M-}' instead of ' ' in the mails.
vim and weechat are displaying this character correctly. I’ve tested it with the versions 20161104 and 20170128.
All other non ASCII characters are displayed correctly.

Thanks,
\-- 
alarig

gahr:
Closes #357

### DIFF
--- a/pager.c
+++ b/pager.c
@@ -1287,11 +1287,11 @@ static int format_line (struct line_t **lineInfo, int n, unsigned char *buf,
       last_special = special;
     }
 
-    if (IsWPrint (wc) || (Charset_is_utf8 && wc == 0x00A0))
+    if (IsWPrint (wc) || (Charset_is_utf8 && (wc == 0x00A0 || wc == 0x202F)))
     {
       if (wc == ' ')
 	space = ch;
-      else if (Charset_is_utf8 && wc == 0x00A0)
+      else if (Charset_is_utf8 && (wc == 0x00A0 || wc == 0x202F))
       {
 	/* Convert non-breaking space to normal space. The local variable
 	 * `space' is not set here so that the caller of this function won't


gahr:

Although this commit fixes the wrong behaviour reported in the issue, I still consider it a workaround. My pager-foo-level is too low to be able to assess whether I'm touching the right place.
The reason for me doubting is that upstream mutt doesn't suffer from the same problem ***but*** doesn't handle Unicode 0x202F specially. Thus, something is done elsewhere which results in the character being printed correctly.

Anyway, since i) this isn't a particularly often used char, and ii) the problem is actually fixed by the commit, I'd suggest we put it into the next release. Feel free to change the milestone if you don't agree.

I plan to look a bit more into this, with less urgency.

Thanks!

gahr:
As always, the minute I finished writing this an idea popped up.

Upstream mutt also suffers from the same problem. The reason why I wasn't noticing is that I'd built it with `--enable-locales-fix`. Neomutt also displays the character correctly with that option turned on.

At this point, I think this fix is not necessary - or at least, not The Right Thing (tm).

gahr:
After some more thinking, I tend to consider the fix legitimate.

The locale fixes hides the problem because all chars >= 0xa0 are considered printable. This doesn't mean we shouldn't treat both non-breakable spaces - the regular one 0xa0 and the narrow one 0x202f - the same way.

Thoughts?

guyzmo:
ok, all that leads me to one question: why is there a problem with utf-8 spaces in the first place? Given the terminal can print it, and the font can display it, why mutt would need to transform those into ASCII spaces, instead of getting them printed all nice and well?

gahr:
> ok, all that leads me to one question: why is there a problem with utf-8 spaces in the first place? [...] why mutt would need to transform those into ASCII spaces

I don't know. Archeology shows that this originated as a [patch](https://bugs.freebsd.org/bugzilla/show_bug.cgi?id=166659) in FreeBSD a few years ago, which was then [imported](https://dev.mutt.org/hg/mutt/rev/41089c2982e3) upstream.

As I said, I will limit myself to make mutt treat the two non-breaking spaces the same way, within the scope of this fix.

PR #360, opened 2017-02-05
factorize base64 usage

d-k-c:
This pull request intends to factorize the usage of various base64 encoding functions.

sendlib.c and rfc2047.c had custom implementations which have been removed in favor of mutt_to_base64. This allows to make B64Chars static to base64.c.

I mostly tested the changes in sendlib.c by sending files and checking I got the same results, so it's only lightly tested and I might have missed some corner cases.

### DIFF
--- a/rfc2047.c
+++ b/rfc2047.c
@@ -197,35 +197,21 @@ static size_t b_encoder (char *s, ICONV_CONST char *d, size_t dlen,
   memcpy (s, "=?", 2), s += 2;
   memcpy (s, tocode, strlen (tocode)), s += strlen (tocode);
   memcpy (s, "?B?", 3), s += 3;
-  for (;;)
+
+  while (dlen)
   {
-    if (!dlen)
-      break;
-    else if (dlen == 1)
-    {
-      *s++ = B64Chars[(*d >> 2) & 0x3f];
-      *s++ = B64Chars[(*d & 0x03) << 4];
-      *s++ = '=';
-      *s++ = '=';
-      break;
-    }
-    else if (dlen == 2)
-    {
-      *s++ = B64Chars[(*d >> 2) & 0x3f];
-      *s++ = B64Chars[((*d & 0x03) << 4) | ((d[1] >> 4) & 0x0f)];
-      *s++ = B64Chars[(d[1] & 0x0f) << 2];
-      *s++ = '=';
-      break;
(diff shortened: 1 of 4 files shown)


d-k-c:

This is a low priority pull request, just making the code a bit cleaner

d-k-c:
I'll prepare some tests to compare the before and after results.

d-k-c:
I tested with the following gist:
https://gist.github.com/d-k-c/fe484ebf27e994d05488f8f4b94e54ee

It reads from /dev/urandom from 1 to 4096 bytes, and encode them using the old and new implementations, and compare the two. If an error is found, error is reported and program stops, otherwise it continues silently. When it's done, it prints a message saying the run is okay.

If you want to burn your CPU, euh test more inputs, you can pass a number as argument to the test program and it will run the whole test the requested number of times.

TL;DR: no errors reported

flatcap:
Wonderful!  Here are my results (truncated):
```
run  1024: OK
Test successful!
```

rfc2047.c
```diff
@@ -197,35 +197,21 @@ static size_t b_encoder (char *s, ICONV_CONST char *d, size_t dlen,
...
+    char encoded[11];
+    size_t ret, i;
+    size_t in_len = MIN(3, dlen);
+
+    ret = mutt_to_base64 (encoded, d, in_len, sizeof(encoded));
+    for (i = 0; i < ret; i++)
+      *s++ = encoded[i];
+
+    dlen -= in_len;
+    d += in_len;
```
gahr: Why is this done in chunks by 3?
Isn't calling `mutt_to_base64(s, d, d, dlen, STRING);` enough, as we anyhow suppose that `s` is big enough?

d-k-c: Well, the answer is... maybe it's big enough. First, this function should be modified to take the output buffer size as parameter, and then we need to be very sure that this output buffer is always bigger than 10B because `mutt_to_base64` has this very weird limitation.
I've done it this way just to be very conservative, I agree it would be better to remove the while loop altogether but I'm not sure it would work in every situation.

gahr: I see.. that 10 should be changed in a 4, but I have been too a chicken to do that :) Anyway, if this works and we don't forget to eventually clean it up, it's +1 for me

PR #366, opened 2017-02-05
Try (once) to reconnect after an IMAP failure

### LINKED ISSUE #346: IMAP: mailbox closed
I am enthusiastic to see an actively maintained fork of mutt.
Unfortunately the ages old issue with random closing of IMAP mailboxes is still present here.

http://mutt-users.mutt.narkive.com/gy14Cvl9/mailbox-closed-mutt-behaviour

Can neomutt be configured to connect every n minutes and during mailbox syncing (`$`) rather than maintaining active connection at all times that gets randomly killed?

gahr:
Closes #346

### DIFF
--- a/imap/command.c
+++ b/imap/command.c
@@ -440,7 +440,6 @@ static int cmd_status (const char *s)
 /* cmd_handle_fatal: when IMAP_DATA is in fatal state, do what we can */
 static void cmd_handle_fatal (IMAP_DATA* idata)
 {
-  static int try_reconnect = 1;
   idata->status = IMAP_FATAL;
 
   if ((idata->state >= IMAP_SELECTED) &&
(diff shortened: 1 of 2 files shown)


gahr:

imap/command.c
```diff
@@ -440,6 +440,7 @@ static int cmd_status (const char *s)
 /* cmd_handle_fatal: when IMAP_DATA is in fatal state, do what we can */
 static void cmd_handle_fatal (IMAP_DATA* idata)
 {
+  static int try_reconnect = 1;
```
d-k-c: Could we keep the function reentrant by moving this to IMAP_DATA? Otherwise 👍

gahr:
> Kill it again and the mailbox is left closed.

@flatcap that is surprising... the "once" in the comment was meant to be "once per failure", that is, if mutt can't re-establish a connection on failure, it doesn't try again. That's the reason for the `try_connect` static var.

If it doesn't reconnect upon a second (unrelated, long after reconnection) failure, then there's still something I have to look at.

imap/command.c
gahr: The function *is* reentrant, just it doesn't behave the same once reentred ;) I don't see a benefit to moving this local piece of data to IMAP_DATA. What am I missing?

PR #368, opened 2017-02-06
Advance by the correct number of position even for unknown characters

### LINKED ISSUE #245: Unicode and sidebar patch with overflowing lines bleeds into sidebar itself
I've noticed that unicode characters and sidebar patch sometimes leads to overflowing characters into the sidebar patch.  See screenshot below.  Beyond reporting this, I've done nothing to investigate why.
unicode-oops

gahr:
Closes #245

### DIFF
--- a/pager.c
+++ b/pager.c
@@ -1339,7 +1339,7 @@ static int format_line (struct line_t **lineInfo, int n, unsigned char *buf,
     {
       if (col + 1 > wrap_cols)
 	break;
-      ++col;
+      col += k;
       if (pa)
 	addch (replacement_char ());
     }


gahr:

pager.c
```diff
@@ -1339,7 +1339,7 @@ static int format_line (struct line_t **lineInfo, int n, unsigned char *buf,
     {
       if (col + 1 > wrap_cols)
 	break;
-      ++col;
+      col += k;
```
toogley: `k` is used in a wide scope (100 lines) but its never really explained what it does.

gahr: I'm still unsure whether this should be `col += wcwidth(replacement_char ());` Let me sleep over it some more :)

gahr: @toogley a refactoring of pager.c could be in order - but let's do that elsewhere. It's a particular bug I'm trying to get caught here :)

toogley: @gahr agreed.

PR #370, opened 2017-02-06
Use `mutt_buffer_init` instead of `memset`

### LINKED ISSUE #369: Use `mutt_buffer_init` in hook.c
A couple of buffer are being initialised using `memset` instead of the `mutt_buffer_init` dedicated function. This is mostly cosmetic but should be fixed.

guiniol:
The mutt_buffer_init function does a memset to 0 of its argument, so use
it instead of doing a manual memset.

I noticed (thanks to @guyzmo) that a number of buffer used in hooks where
initialised using memset instead of this dedicated function. This
remedies to that.

Fixes #369

### DIFF
--- a/alias.c
+++ b/alias.c
@@ -344,53 +344,49 @@ retry_name:
   else
     Aliases = new;
 
-  for (LIST *AliasFile = AliasFiles; AliasFile != NULL;
-      AliasFile = AliasFile->next)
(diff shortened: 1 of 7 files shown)


guiniol:

This is a low priority pull request, just making the code a bit cleaner

PR #373, opened 2017-02-07
Do not cast return values from malloc (et similia)

### LINKED ISSUE #372: Do not cast the return from malloc et similia

gahr:
Closes #372

### DIFF
--- a/mutt.h
+++ b/mutt.h
@@ -656,10 +656,26 @@ typedef struct spam_list_t
   struct spam_list_t *next;
 } SPAM_LIST;
 
-#define mutt_new_list() safe_calloc (1, sizeof (LIST))
-#define mutt_new_heap() safe_calloc(1, sizeof(HEAP))
-#define mutt_new_rx_list() safe_calloc (1, sizeof (RX_LIST))
-#define mutt_new_spam_list() safe_calloc (1, sizeof (SPAM_LIST))
+inline LIST *mutt_new_list()
+{
+  return safe_calloc (1, sizeof (LIST));
+}
+
+inline HEAP *mutt_new_heap()
+{
+  return safe_calloc (1, sizeof (HEAP));
+}
+
+inline RX_LIST *mutt_new_rx_list()
+{
+  return safe_calloc (1, sizeof (RX_LIST));
+}
+
+inline SPAM_LIST *mutt_new_spam_list()
+{
+  return safe_calloc (1, sizeof (SPAM_LIST));
+}
+
 void mutt_free_list (LIST **);
 void mutt_free_rx_list (RX_LIST **);
 void mutt_free_spam_list (SPAM_LIST **);
@@ -728,12 +744,22 @@ typedef struct envelope
   unsigned int refs_changed : 1; /* References changed to break thread */
 } ENVELOPE;
 
+inline ENVELOPE *mutt_new_envelope()
+{
+    return safe_calloc (1, sizeof (ENVELOPE));
+}
+
 typedef struct parameter
 {
   char *attribute;
   char *value;
   struct parameter *next;
 } PARAMETER;
+
+inline PARAMETER *mutt_new_parameter()
+{
+    return safe_calloc (1, sizeof (PARAMETER));
+}
 
 /* Information that helps in determing the Content-* of an attachment */
 typedef struct content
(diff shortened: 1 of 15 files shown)


gahr:
@neomutt/reviewers please review, thanks!

flatcap:
I agree with your view and your changes, but I want to suggest you make one exception.
Change the macros in `mutt.h`, from
```
#define mutt_new_list()      safe_calloc(1, sizeof(LIST))
#define mutt_new_heap()      safe_calloc(1, sizeof(HEAP))
#define mutt_new_rx_list()   safe_calloc(1, sizeof(RX_LIST))
#define mutt_new_spam_list() safe_calloc(1, sizeof(SPAM_LIST))
```
to
```
#define mutt_new_list()      (LIST*)      safe_calloc(1, sizeof(LIST))
#define mutt_new_heap()      (HEAP*)      safe_calloc(1, sizeof(HEAP))
#define mutt_new_rx_list()   (RX_LIST*)   safe_calloc(1, sizeof(RX_LIST))
#define mutt_new_spam_list() (SPAM_LIST*) safe_calloc(1, sizeof(SPAM_LIST))
```

That way, anyone using the "functions" can't accidentally assign it to the wrong variable type.
Alternatively, replace them with inline functions.

darnir:
As far as possible, we should replace all macros with inline functions. If you really want to be use it is inlined, use `__attribute__((always_inline))`

guyzmo:
I agree with @darnir inline is better than macro's suggestion.

gahr:
Great idea.. I've tracked down a few mode (`mutt_new_envelope`, `mutt_new_parameter`, `mutt_new_header`, and `mutt_new_enter_state`.

I don't have a strong opinion on where to put the externals.. I `muttlib.c` looks like a reasonable candidate though..

muttlib.c
```diff
@@ -53,6 +53,18 @@
 #include <sys/types.h>
 #include <utime.h>
 
+/*
+ * External definitions for inline functions in mutt.h
```
d-k-c: Why not including mutt.h?

mutt.h
```diff
@@ -656,10 +656,26 @@ typedef struct spam_list_t
   struct spam_list_t *next;
 } SPAM_LIST;
 
-#define mutt_new_list() safe_calloc (1, sizeof (LIST))
-#define mutt_new_heap() safe_calloc(1, sizeof(HEAP))
-#define mutt_new_rx_list() safe_calloc (1, sizeof (RX_LIST))
-#define mutt_new_spam_list() safe_calloc (1, sizeof (SPAM_LIST))
+inline LIST *mutt_new_list()
```
d-k-c: I think the idiomatic way is to mark the functions `static inline` to avoid unused functions warnings, but maybe `inline` is enough for that?

darnir: I don't think it makes much sense to have a `static inline` declaration inside a header file. Or did I miss something here?

regex.c
```diff
@@ -362,8 +362,8 @@ char *alloca ();
   (size1 && string1 <= (ptr) && (ptr) <= string1 + size1)
 
 /* (Re)Allocate N items of type T using malloc, or fail.  */
-#define TALLOC(n, t) ((t *) malloc ((n) * sizeof (t)))
-#define RETALLOC(addr, n, t) ((addr) = (t *) realloc (addr, (n) * sizeof (t)))
+#define TALLOC(n, t) (malloc ((n) * sizeof (t)))
+#define RETALLOC(addr, n, t) ((addr) = realloc (addr, (n) * sizeof (t)))
```
darnir: Why is this  not using `safe_realloc (addr, (n) * sizeof(t))`?

regex.c
```diff
@@ -5484,12 +5482,12 @@ re_comp (s)
 
   if (!re_comp_buf.buffer)
     {
-      re_comp_buf.buffer = (unsigned char *) malloc (200);	/* __MEM_CHECKED__ */
+      re_comp_buf.buffer = malloc (200);	/* __MEM_CHECKED__ */
```
darnir: Shouldn't these calls also be using the `safe_*` functions? Or is that something for a different patch?

guyzmo: well, I'm not sure what's the rationale about this, but the `malloc() __MEM_CHECKED__` calls have been made on purpose.

PR #375, opened 2017-02-07
debug: remove dprint in favor of mutt_debug

### LINKED ISSUE #362: Remove first argument from mutt_debug
`mutt_debug` is only called by the `dprint(N,X)` macro. The macro is defined as

```c
#  define dprint(N,X) do { if(debuglevel>=N && debugfile) mutt_debug X; } while (0)
```

and a common use is

```c
 dprint (1, (debugfile, "mutt_lookup_mime_type: Internal error, count = %d.\n", count));
````

It's pretty clear that:
1) `mutt_debug` is only called with `debugfile` as its first argument
2) if `dprint` was called with anything else than `debugfile` as part of the function call construction appearing in its second argument `X`, an inconsistent check would be done within the `dprint` body (check for `debugfile`, call `mutt_debug` with a `FILE*`pointing to something else)

My proposal is to get rid of the first argument of `mutt_debug` and hardcode the check and the use of `debugfile` within its body.

Thoughts?

### PR DESCRIPTION
**Pull request by d-k-c**
The dprint macro is error-prone. Any FILE* can be passed to dprint, but
the macro checks that `debugfile` is not NULL (ie. it's open), not the
file pointer it received as parameter. This had also the consequence
that `debugfile` is passed to every call to `dprint`, making lines
unnecessarily long.

This commit replaces calls to `dprint` by direct calls to `mutt_debug`.
It's now hardcoded in this function that it will write in `debugfile`.
Also, remove the MUTT_LIB_WHERE magic.

The substistution has been done with the following command:
   perl -0777 -pi -e 's/dprint[ ]*\((.+?),[ \t\r\n]*\(debugfile,[ ]*(.*?)\)\);/mutt_debug (\1, \2);/igs' *.c imap/*.c
And the code was then slightly ajusted to fix identation and line
length.

Closes #362

### DIFF
--- a/lib.h
+++ b/lib.h
@@ -129,31 +129,16 @@ static inline int is_email_wsp(char c)
 extern void (*mutt_error) (const char *, ...);
 # endif
 
-# ifdef _LIB_C
-#  define MUTT_LIB_WHERE 
-#  define MUTT_LIB_INITVAL(x) = x
-# else
-#  define MUTT_LIB_WHERE extern
-#  define MUTT_LIB_INITVAL(x)
-# endif
-
 void mutt_exit (int);
 
 
-# ifdef DEBUG
-
-MUTT_LIB_WHERE FILE *debugfile MUTT_LIB_INITVAL(0);
-MUTT_LIB_WHERE int debuglevel MUTT_LIB_INITVAL(0);
-
-void mutt_debug (FILE *, const char *, ...);
-
-#  define dprint(N,X) do { if(debuglevel>=N && debugfile) mutt_debug X; } while (0)
-
-# else
-
-#  define dprint(N,X) do { } while (0)
-
-# endif
+#ifdef DEBUG
+extern FILE *debugfile;
+extern int debuglevel;
+void mutt_debug (int level, const char *, ...);
+#else
+#define mutt_debug(...) do { } while (0)
+#endif
 
 
 /* Exit values used in send_msg() */
--- a/mbox.c
+++ b/mbox.c
@@ -147,7 +147,7 @@ int mmdf_parse_mailbox (CONTEXT *ctx)
       if (fgets (buf, sizeof (buf) - 1, ctx->fp) == NULL)
       {
 	/* TODO: memory leak??? */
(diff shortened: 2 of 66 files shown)

### CODE REVIEW

File: mbox.c [hunk 1]
```diff
@@ -338,11 +338,13 @@ int mbox_parse_mailbox (CONTEXT *ctx)
 	      fgets (buf, sizeof (buf), ctx->fp) == NULL ||
 	      mutt_strncmp ("From ", buf, 5) != 0)
 	  {
-	    dprint (1, (debugfile, "mbox_parse_mailbox: bad content-length in message %d (cl=" OFF_T_FMT ")\n", curhdr->index, curhdr->content->length));
-	    dprint (1, (debugfile, "\tLINE: %s", buf));
+	    mutt_debug (1, "mbox_parse_mailbox: bad content-length in message"
+	                "%d (cl=" OFF_T_FMT ")\n",
+	                curhdr->index, curhdr->content->length);
```
flatcap: When you folded the message, you lost the space between `message` and `%d`.

lib.h
```diff
@@ -129,31 +129,16 @@ static inline int is_email_wsp(char c)
...
-#  define dprint(N,X) do { } while (0)
-
-# endif
+#ifdef DEBUG
+extern FILE *debugfile;
+extern int debuglevel;
+void mutt_debug (int level, const char *, ...);
+#else
+#define mutt_debug(...) do { } while (0)
+#endif
```
gahr: Wouldn't it suffice to leave the macro empty?
A semicolon is a valid statement..

flatcap: gcc complains if you use it after an 'if'
```c
#define mutt_debug(...)
int main()
{
	if (1)
		mutt_debug("hello");
}
```

jim.c:6:8: warning: suggest braces around empty body in an ‘if’ statement [-Wempty-body]

gahr: I guess this would be a reason for always using braces? ;) - joking. I'm fine with the empty do-while then ;)

d-k-c: I think a pair of braces could do the trick as well, but that's how the dprint macro was stubbed so I kept the same thing

d-k-c:
@neomutt/reviewers Please review (sorry for the large changeset 👼)

PR #376, opened 2017-02-07
Fix dubious use macro for _() / gettext()

flatcap:
This PR supercedes #266 (Internationalize strings).

The gettext macro `_()` was pooly defined which led to some bad practices.
Note the missing `()`s in this:
```
mutt_message _("string")
```

@alok used some scripting skills to fix the problems, but it left a LOT of false positives.
This PR has filtered out the stuff that wasn't needed.

The first commit (LARGE) fixes `function _("string");` to `function(_("string"));`
The second commit (small) adds some missing translation markers.

Thanks @alok it's saved me a lot of time.

### DIFF
--- a/addrbook.c
+++ b/addrbook.c
@@ -142,7 +142,7 @@ void mutt_alias_menu (char *buf, size_t buflen, ALIAS *aliases)
   
   if (!aliases)
   {
-    mutt_error _("You have no aliases!");
+    mutt_error (_("You have no aliases!"));
     return;
   }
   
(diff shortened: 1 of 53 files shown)


toogley:
@flatcap Btw. is there a reason to keep this macro? i personally consider it a bit ugly..

flatcap:
> is there a reason to keep this macro?

That's the way it's always done in programs, not just mutt.
Being a macro means that if you compile **without** gettext support then you just get normal strings.
(there are no unnecessary function calls)

toogley:
@flatcap Ah, i didn't know that, thanks.

PR #380, opened 2017-02-08
Fixes name for the LIFO structure.


guyzmo:
Mea culpa: I used *heap* when I really meant *stack*, for the
structure used to detect cycles when sourcing muttrc files.


guyzmo:
ok, looks like @d-k-c agrees as well, I'll change that

guyzmo:
@gahr, @d-k-c hope you'll like it better ☺

BTW, @neomutt/reviewers plesae review !@#$ 😀

PR #384, opened 2017-02-09
Release KyotoCabinet data with kcfree()

### LINKED ISSUE #383: KyotoCabinet data should be free'd with kcfree()
... which internally calls c++ operator `delete[]`. The data type is a `char *` so no c++ destructors are involved and it *should* be functionally equivalent to release the memory with `free()`.. but let's stay on the safe side and do as they say.

From [upstream 09bb4a62ceb1](https://dev.mutt.org/hg/mutt/rev/09bb4a62ceb1).

gahr:
Obtained from https://dev.mutt.org/hg/mutt/rev/09bb4a62ceb1

Closes #383


gahr:

PR #387, opened 2017-02-10
Fix latest Coverity issues

### LINKED ISSUE #386: Fix coverity warnings
... reported by the latest coverity scan.

gahr:
Closes #386

### DIFF
--- a/imap/util.c
+++ b/imap/util.c
@@ -909,15 +909,15 @@ int imap_wait_keepalive (pid_t pid)
 
 void imap_allow_reopen (CONTEXT *ctx)
 {
-  IMAP_DATA* idata = ctx->data;
-  if (ctx && ctx->magic == MUTT_IMAP && idata->ctx == ctx)
+  IMAP_DATA* idata;
+  if (ctx && ctx->magic == MUTT_IMAP && (idata = ctx->data)->ctx == ctx)
     idata->reopen |= IMAP_REOPEN_ALLOW;
 }
 
 void imap_disallow_reopen (CONTEXT *ctx)
 {
-  IMAP_DATA* idata = ctx->data;
-  if (ctx && ctx->magic == MUTT_IMAP && idata->ctx == ctx)
+  IMAP_DATA *idata;
+  if (ctx && ctx->magic == MUTT_IMAP && (idata = ctx->data)->ctx == ctx)
     idata->reopen &= ~IMAP_REOPEN_ALLOW;
 }
 
--- a/imap/util.c
+++ b/imap/util.c
@@ -909,8 +909,12 @@ int imap_wait_keepalive (pid_t pid)
 
 void imap_allow_reopen (CONTEXT *ctx)
 {
-  IMAP_DATA* idata;
-  if (ctx && ctx->magic == MUTT_IMAP && (idata = ctx->data)->ctx == ctx)
+  IMAP_DATA *idata;
+  if (!ctx || !ctx->data || ctx->magic != MUTT_IMAP)
+      return;
+
+  idata = ctx->data;
+  if (idata->ctx == ctx)
     idata->reopen |= IMAP_REOPEN_ALLOW;
 }
 
--- a/imap/util.c
+++ b/imap/util.c
@@ -917,7 +917,11 @@ void imap_allow_reopen (CONTEXT *ctx)
 void imap_disallow_reopen (CONTEXT *ctx)
 {
   IMAP_DATA *idata;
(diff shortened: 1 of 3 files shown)


gahr:

gahr:
I'm done with this small batch. @neomutt/reviewers please review :)

parse.c
```diff
@@ -1626,9 +1626,10 @@ static int count_body_parts_check(LIST **checklist, BODY *b, int dflt)
     a = (ATTACH_MATCH *)type->data;
     mutt_debug (5, "cbpc: %s %d/%s ?? %s/%s [%d]... ",
                 dflt ? "[OK]   " : "[EXCL] ",
-                b->type, b->subtype, a->major, a->minor, a->major_int);
+                b->type, b->subtype ? b->subtype : "*",
+                a->major, a->minor, a->major_int);
     if ((a->major_int == TYPEANY || a->major_int == b->type) &&
-	!regexec(&a->minor_rx, b->subtype, 0, NULL, 0))
+        (!b->subtype || !regexec(&a->minor_rx, b->subtype, 0, NULL, 0)))
     {
```
flatcap: 👍

imap/util.c
```diff
@@ -909,15 +909,15 @@ int imap_wait_keepalive (pid_t pid)
...
 }
 
 void imap_disallow_reopen (CONTEXT *ctx)
 {
-  IMAP_DATA* idata = ctx->data;
-  if (ctx && ctx->magic == MUTT_IMAP && idata->ctx == ctx)
+  IMAP_DATA *idata;
+  if (ctx && ctx->magic == MUTT_IMAP && (idata = ctx->data)->ctx == ctx)
     idata->reopen &= ~IMAP_REOPEN_ALLOW;
 }
```
flatcap: I'd prefer if you could move the assignment out of the `if` statement.

init.c
```diff
@@ -506,11 +506,14 @@ static int add_to_spam_list (SPAM_LIST **list, const char *pat, const char *temp
...
+    rx = NULL;
     if (last)
       last->next = t;
     else
       *list = t;
   }
+  else
+    mutt_free_regexp(&rx);
 
   /* Now t is the SPAM_LIST* that we want to modify. It is prepared. */
```
flatcap: You'll need to expand the view above and below, to follow this comment.
- We find an entry `t` in the spam list (`for` loop above)
- We dereference `t->rx` (`if` below) without checking it

This isn't something you've introduced (and it may not even be a problem).

imap/util.c
gahr: I can't. Outside of the if I don't know if ctx is valid :(

d-k-c: You can move it inside the body of the `if`. Alternatively:
```
if (ctx && ctx->magic == MUTT_IMAP && ctx->data->ctx ==ctx)
{
  IMAP_DATA *idata = ctx->data;
  idata->reopen &= ~IMAP_REOPEN_ALLOW;
}
```

or 

```
if (ctx && ctx->magic == MUTT_IMAP)
{
  IMAP_DATA *idata = ctx->data;
  if (idata->ctx == ctx)
    idata->reopen &= ~IMAP_REOPEN_ALLOW;
}
```

guyzmo: wouldn't the following work?

``` c
if (ctx && ctx->magic && (ctx->data->tx == ctx))
{
  idata = ctx->data;
``

File: init.c (hunk 3 above)
Comment (guyzmo) [inline init.c]: I believe that:

```
 if (!(rx = mutt_compile_regexp (pat, REG_ICASE)))
```

always gives an assigned `rx` or exits. Then the `free(rx)` is fine.

Comment (gahr) [inline init.c]: I don't think we could ever end up with a SpamList entry without an `rx`, as `mutt_add_to_spam_list` is the only function adding to it.

File: imap/util.c [hunk 4]
```diff
@@ -909,15 +909,19 @@ int imap_wait_keepalive (pid_t pid)
 
 void imap_allow_reopen (CONTEXT *ctx)
 {
-  IMAP_DATA* idata = ctx->data;
-  if (ctx && ctx->magic == MUTT_IMAP && idata->ctx == ctx)
+  IMAP_DATA* idata;
+  if (ctx && ctx->magic == MUTT_IMAP && (idata = ctx->data)->ctx == ctx)
```
guyzmo: same comment as before — and below, that kind of assignment is better done outside of the `if`. ☺

imap/util.c
gahr: Except `ctx->data` is `void*`

PR #390, opened 2017-02-10
Add an option to show multipart/alternative

immae:
This pull request adds an option to ignore the multipart/alternative flag in emails. The reason for that is that there exist a lot of mailers that misuse that flag and doesn't have the same contents in the "alternative" part. 

Thus, clients that make the "wrong" choice in the alternative will see unhelpful or incomplete data and may miss important informations if he doesn't think to look at the attachment part.

NB: Is the documentation automatically generated from the comments?

### DIFF
--- a/handler.c
+++ b/handler.c
@@ -65,6 +65,23 @@ const int Index_64[128] = {
     41,42,43,44, 45,46,47,48, 49,50,51,-1, -1,-1,-1,-1
 };
 
+static void print_part_line (STATE *s, BODY *b, int n)
+{
+  char length[5];
+  mutt_pretty_size (length, sizeof (length), b->length);
+  state_mark_attach (s);
+  char *charset = mutt_get_parameter ("charset", b->parameter);
+  if (n != 0)
+      state_printf (s, _("[-- Alternative Type #%d: "), n);
+  else
+      state_printf (s, _("[-- Type: "));
+  state_printf (s, _("%s/%s%s%s, Encoding: %s, Size: %s --]\n"),
+          TYPE (b), b->subtype,
+          charset ? "; charset=" : "",
+          charset ? charset : "",
+          ENCODING (b->encoding), length);
+}
+
 static void state_prefix_put (const char *d, size_t dlen, STATE *s)
 {
   if (s->prefix)
@@ -1004,7 +1021,6 @@ static int alternative_handler (BODY *a, STATE *s)
   int type = 0;
   int mustfree = 0;
   int rc = 0;
-  char length[5];
   int count = 0;
 
   if (a->encoding == ENCBASE64 || a->encoding == ENCQUOTEDPRINTABLE ||
@@ -1135,10 +1151,7 @@ static int alternative_handler (BODY *a, STATE *s)
 
     if (mutt_strcmp ("info", ShowMultipartAlternative) == 0)
     {
-      mutt_pretty_size (length, sizeof (length), choice->length);
-      state_mark_attach (s);
-      state_printf (s, _("[-- Type: %s/%s, Encoding: %s, Size: %s --]\n"),
-                    TYPE (choice), choice->subtype, ENCODING (choice->encoding), length);
+      print_part_line (s, choice, 0);
     }
     mutt_body_handler (choice, s);
 
@@ -1153,13 +1166,10 @@ static int alternative_handler (BODY *a, STATE *s)
         if (choice != b)
         {
           count += 1;
-          mutt_pretty_size (length, sizeof (length), b->length);
           if (count == 1)
             state_putc ('\n', s);
 
-          state_mark_attach (s);
-          state_printf (s, _("[-- Alternative Type #%d %s/%s, Encoding: %s, Size: %s --]\n"),
-                        count, TYPE (b), b->subtype, ENCODING (b->encoding), length);
+          print_part_line (s, b, count);
         }
         b = b->next;
       }
@@ -1261,7 +1271,6 @@ int mutt_can_decode (BODY *a)
 static int multipart_handler (BODY *a, STATE *s)
 {
   BODY *b, *p;
-  char length[5];
   struct stat st;
   int count;
   int rc = 0;
@@ -1292,12 +1301,7 @@ static int multipart_handler (BODY *a, STATE *s)
 		    p->filename ? p->filename : p->form_name, s);
       }
       state_puts (" --]\n", s);
-
-      mutt_pretty_size (length, sizeof (length), p->length);
-      
-      state_mark_attach (s);
-      state_printf (s, _("[-- Type: %s/%s, Encoding: %s, Size: %s --]\n"),
-		    TYPE (p), p->subtype, ENCODING (p->encoding), length);
+      print_part_line (s, p, 0);
       if (!option (OPTWEED))
       {
 	fseeko (s->fpin, p->hdr_offset, 0);
--- a/handler.c
+++ b/handler.c
@@ -1004,6 +1004,8 @@ static int alternative_handler (BODY *a, STATE *s)
   int type = 0;
   int mustfree = 0;
   int rc = 0;
+  char length[5];
+  int count = 0;
 
   if (a->encoding == ENCBASE64 || a->encoding == ENCQUOTEDPRINTABLE ||
       a->encoding == ENCUUENCODED)
@@ -1130,7 +1132,38 @@ static int alternative_handler (BODY *a, STATE *s)
       fseeko (s->fpin, choice->hdr_offset, 0);
       mutt_copy_bytes(s->fpin, s->fpout, choice->offset-choice->hdr_offset);
     }
+
+    if (mutt_strcmp ("info", ShowMultipartAlternative) == 0)
+    {
+      mutt_pretty_size (length, sizeof (length), choice->length);
+      state_mark_attach (s);
+      state_printf (s, _("[-- Type: %s/%s, Encoding: %s, Size: %s --]\n"),
+                    TYPE (choice), choice->subtype, ENCODING (choice->encoding), length);
+    }
     mutt_body_handler (choice, s);
+
+    if (mutt_strcmp ("info", ShowMultipartAlternative) == 0)
+    {
+      if (a && a->parts)
+        b = a->parts;
+      else
+        b = a;
+      while (b)
+      {
+        if (choice != b)
+        {
+          count += 1;
+          mutt_pretty_size (length, sizeof (length), b->length);
+          if (count == 1)
+            state_putc ('\n', s);
+
+          state_mark_attach (s);
+          state_printf (s, _("[-- Alternative Type #%d %s/%s, Encoding: %s, Size: %s --]\n"),
+                        count, TYPE (b), b->subtype, ENCODING (b->encoding), length);
+        }
+        b = b->next;
+      }
+    }
   }
   else if (s->flags & MUTT_DISPLAY)
   {
@@ -1794,7 +1827,7 @@ int mutt_body_handler (BODY *b, STATE *s)
   {
     char *p;
 
-    if (ascii_strcasecmp ("alternative", b->subtype) == 0)
+    if ((mutt_strcmp ("inline", ShowMultipartAlternative) != 0) && (ascii_strcasecmp ("alternative", b->subtype) == 0))
       handler = alternative_handler;
     else if (WithCrypto && ascii_strcasecmp ("signed", b->subtype) == 0)
     {
--- a/globals.h
+++ b/globals.h
@@ -135,6 +135,7 @@ WHERE short SearchContext;
 WHERE char *SendCharset;
 WHERE char *Sendmail;
 WHERE char *Shell;
+WHERE char *ShowMultipartAlternative;
 #ifdef USE_SIDEBAR
 WHERE char *SidebarDelimChars;
 WHERE char *SidebarDividerChar;
--- a/init.c
+++ b/init.c
@@ -1978,6 +1978,13 @@ static int check_charset (struct option_t *opt, const char *val)
   return rc;
 }
 
+static bool valid_show_multipart_alternative(const char *val)
+{
+  return ((mutt_strcmp(val, "inline") == 0) ||
+          (mutt_strcmp(val, "info") == 0) ||
+          (val == NULL) || (*val == 0));
+}
+
 static int parse_setenv(BUFFER *tmp, BUFFER *s, unsigned long data, BUFFER *err)
 {
   int query, unset, len;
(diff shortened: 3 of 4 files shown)


immae:
Clearly it's not the ideal solution (if there is such solution) but in my opinion it's better than the current state, and hopefully better ideas will come from that

toogley:
hi and welcome to neomutt :)

Can you please add your option and its description to [our manual](https://github.com/neomutt/neomutt/blob/master/doc/manual.xml.head), to the section "configuration variables" ?

guyzmo:
@immae then the contents are shown as traditional multipart?

gahr:
Is mutt doing the "wrong choice"? IIRC (haven't checked the rfc) there is no choice to be made - the first part is the default one to show.

This is not to say that I'm against the change, but I don't immediately see what it brings. Am I wrong to assume that with this chance one would potentially have pretty much the same content displayed multiple times?

guyzmo:
could you update your PR with the missing stuff, as described in the upcoming [PR template](https://github.com/neomutt/neomutt/blob/75f312155a269ba78a25f3d6359c1454e833c054/.github/PULL_REQUEST_TEMPLATE.md)?

immae:
@guyzmo that's the idea

@gahr : you can configure mutt with an "alternative_order" setting, that will show the first one that matches. In theory, yes it would show duplicate information, since that's the idea of multipart/alternative. In practice, mail senders don't do that correctly and you may miss information depending on which alternative you chose. And I tried many settings, but never found the magic one that would work with every emails. That is very frustrating, but that's how it is nowadays

guyzmo:
@gahr the RFC requests to to use the first available choice that your client supports. But the issue with that is that sometimes the `text/plain` part is just empty, or a line saying `click on the link to see the message` or `your client is not supported, you cannot read this mail`.

I have seen mostly happening with viagra spams (where the `text/plain` is pure non-sense), but it starts to happen more and more with bulk mailing applications, that don't like that text based client don't render the webpages with all the JS and images that call back their services.

TBH, I'm not sure this change is a *MUST* have, as when you get such a mail and see obvious missing content, you just have to hit `v` and you'll see the list of contents.

immae:
@guyzmo : yes, you "just have" to do it. However it's not always obvious that you're missing something (well, to convince you of that, you should look at each and every mail and see what you're missing).

The most obvious one is when one of the alternative is a text/calendar, that you may be interested in including in your calendar, but you don't see it is one as the text is not always mentionning that it contains a calendar

d-k-c:
Maybe a more general approach would be to indicate somewhere in the view that there are other alternative parts available (a bit like `[-- Attachment #1 --]`, or the PGP informations)

guyzmo:
I agree with @d-k-c that showing up the list of other attachments like the `[-- … --]` line for traditional multipart messages is a nice way to go. Another alternative would be to generate some sort of fake header that just gives the list of attachments, so you can be informed they exist and press `v` to use them.

guyzmo:
you just need to add the `manual.xml.head` part, no need for doxygen as you're not adding a function.

immae:
@guyzmo : Yes, I'm trying to do it, but I'm struggling, sorry to take so long :/

immae:
Actually the `manual.xml.head` finishes at `<title>Configuration Variables</title>`, so I should add it somewhere *after* the last line, and I cannot find any suitable file to add the content (I tried to grep for other configuration variables, in vain...)

immae:
(My bet was that it was created with the ".pp" enclosing comments, but since it's not I have no idea where to look at..)

immae:
I just checked and the variable is actually referenced in the documentation as I bet. So I don't understand what you're expecting @guyzmo

toogley:
@guyzmo Hm.. I looked at the docs/ folder again and as far as i understand the list of configuration options seems to be generated by [makedoc.c](https://github.com/neomutt/neomutt/blob/master/doc/makedoc.c#L20-L26) - or do i confuse sth?

@immae don't worry :)

immae:
@d-k-c: here you are for your suggestion https://github.com/neomutt/neomutt/pull/391 (but I definitly prefer to have the setting flag)

gahr:
Thanks for clarifying the intent, @immae and @guyzmo. Most likely this option can't do any harm, can it?

immae:
The option only turns a multipart/alternative into a standard multipart, so I guess it cannot do any harm

guyzmo:
hum… @toogley, and what about changing [here](https://github.com/neomutt/neomutt/blob/master/doc/manual.xml.head#L12430) is it done automagically?

guyzmo:
@immae I'd suggest merging #391 and this one into a single option `show_multipart_alternative` with three values:

- `(none)` → when unset, have the default expected behaviour (show the first alternative) ;
- `inline` → what you want, display all the parts inlined,
- `info` → show multipart info as header (cf #391).

what do you think?

immae:
@guyzmo : I thought of that, but I cannot see any good reason to want `(none)` ? you already see all attachments, why not those one? To me, #391 is a bugfix

immae:
Other than that, no opinion, I can of course merge them

guyzmo:
@immae the main rationale than that it's the default behaviour and I guess some people might prefer this. I can't speak for everybody, and this could either be discussed, or we could just support the three options ☺

toogley:
@guyzmo [neomutt.org](https://www.neomutt.org/guide/reference) contains explanations of variables such as `abort_nosubject`. However grepping for this string in `manual.xml.head` returns nothing. Therefore i conclude `makedoc.c` generates that.

guyzmo:
This never bothered me up until now, and I might happily switch to the `info` mode (or whatever better name you can come up with). But I won't like the `inline` (same comment about name applies) mode.

immae:
Well, a default behavior that is a bug shouldn't have an option to remain a bug :p

But if you think it's a feature to see only partial attachment information then I'll merge them (I talking about the (none) vs info here, of course I won't force the inline version to everyone :p )

guyzmo:
@toogley and what about [here](https://github.com/neomutt/neomutt/blob/master/doc/manual.xml.head#L12121)?

immae:
@guyzmo : you're referencing "Example NeoMutt config file for the nntp feature", which is out of our scope?

guyzmo:
dammit, `manual.xml.head` is way too big, and I can't find crap in it. @flatcap, is makedoc doing proper documenting of variables?

@immae AFAICT, everything about `multipart/alternative` is a 20 year old troll that's very touchy, and I wouldn't want us to disrupt the way some people have configured their mailer because we changed something about it. So yes, what you think is a bug is likely to be seen as a feature by some users, so it's better to keep the default as is, and introduce new ways of operating on this.

toogley:
@guyzmo i think that's specifically for nntp, not the general list of variables. Rather, i think [init.h](https://github.com/neomutt/neomutt/blob/master/init.h#L110-L116) is the correct place, because there's also the description visible like on the website.^^

guyzmo:
@immae so please, merge your changes from #391 into this PR, and I'm testing it, and then approving it.

immae:
So I updated the PR with @guyzmo suggestion:

* (none) → when unset, have the default expected behaviour (show the first alternative) ;
* inline → what you want, display all the parts inlined,
*  info → show multipart info as header (cf #391).

and closed #391 which is now obsolete

handler.c
```diff
@@ -1130,7 +1132,35 @@ static int alternative_handler (BODY *a, STATE *s)
       fseeko (s->fpin, choice->hdr_offset, 0);
       mutt_copy_bytes(s->fpin, s->fpout, choice->offset-choice->hdr_offset);
     }
+
+    if (mutt_strcmp ("info", ShowMultipartAlternative) == 0) {
```
guyzmo: curly brace on new line

handler.c
```diff
@@ -1130,7 +1132,35 @@ static int alternative_handler (BODY *a, STATE *s)
...
+      mutt_pretty_size (length, sizeof (length), choice->length);
+      state_mark_attach (s);
+      state_printf (s, _("[-- Type: %s/%s, Encoding: %s, Size: %s --]\n"),
+                    TYPE (choice), choice->subtype, ENCODING (choice->encoding), length);
+    }
     mutt_body_handler (choice, s);
+
+    if (mutt_strcmp ("info", ShowMultipartAlternative) == 0) {
```
guyzmo: curly brace on new line

handler.c
```diff
@@ -1794,7 +1824,7 @@ int mutt_body_handler (BODY *b, STATE *s)
   {
     char *p;
 
-    if (ascii_strcasecmp ("alternative", b->subtype) == 0)
+    if (mutt_strcmp ("inline", ShowMultipartAlternative) != 0 && ascii_strcasecmp ("alternative", b->subtype) == 0)
```
guyzmo: parenthesis around each part of the `&&`

guyzmo:
@neomutt/reviewers need one more!

immae:
@guyzmo : changes made!

guiniol:
@immae Thanks for this. I receive calendar invitations where the text show no indication of an invitation and the invitation does not repeat the text so setting this option to `info` will help a lot. I'll test and review this today.

guyzmo:
I think we can improve on the existing info string, to show off the list of attachments in a nicer way, but that's material for another PR.

gahr:
Before considering merging this, I would like to see the config string properly parsed in init.c so that invalid values are properly detected and reported.

guyzmo:
@gahr AFAICT, that means introducing a new `DT_` type flag, which space is limited. I think this scheme will need to be refactored at some point. I'd rather see a validator function pointer passed as parameter to the init.h setting definition, so that we can differentiate the storage type (`int`, `char*` etc.) from the acceptable value.

immae:
@gahr : unless you show me something that already implements "enums" as you want in the code, I won't have time to do it properly in the forseable future, this seems completely out of the scope of that PR

gahr:
It can be done eithet with a new dedicated DT_ type or by variable name. Search init.c for DT_HCACHE for the first approach or "charset" for the second.
It's not at all a big deal.

guyzmo:
for the first approach, we cannot do that for each and every type. we got only a 16 values namespace for that.

for the second approach, then you need to add a type checking function and an `if` case in the `parse_set` function. That's a better way, but quite cumbersome. Also, note that `charset` is the only subtype being handled that way in the `DT_STR` type check.

As long as we don't have a generic way to handle typing of the set variables, it's gonna be a mess.

So, I'm ok with current unchecked situation, or with the second option of checking the variable value as subtype check in the `DT_STR` check.

gahr:
Oh, of course I'm not suggesting to rethink over the mechanism of handling config variables - that would definitely be out of scope. I'm just asking that the functionalities that are already there and being used are employed to do a bit of sanity checking and useful error reporting.

guyzmo:
@immae to make @gahr happy, you could add the checking [in a new if there](https://github.com/neomutt/neomutt/blob/master/init.c#L2331).

gahr:
If you want to make me super-happy, could you please also provide an option value, say "default", for the default behavior? Thanks a lot!

guyzmo:
I'm agnostic to that, whatever the default is, `unset show_multipart_alternative` will provide the default. @immae again, if you're happy to do the changes, please do, otherwise we can update that at merge time.

flatcap:
This is very nice.  Simple and effective.
I've just tested it on my junk dir -- very enlightening.

> is makedoc doing proper documenting of variables?

Yes, it converts `init.h` into both DocBook xml and nroff man page.

immae:
@guyzmo: If there are changes to do, I prefer the PR to have them. I'm just out of computer here, so it'll wait until tomorrow, but I'll add the checks :)

We've been decades without it, I'm sure it can wait a few more hours ;)

gahr:
@immae I just tried out your patch, and it's great, really! Thank you for that! One last thing that I think might be good to have displayed is the charset.

I have taken the freedom to implement that - along with a small refactoring of the printing code which was used in three different places - in the [devel/pull-390](https://github.com/neomutt/neomutt/commits/devel/pull-390) branch.

The [first commit](https://github.com/neomutt/neomutt/commit/f5a3061c1bed753a0bcd08eb944424e8b494c604) is your changes, the [second one](https://github.com/neomutt/neomutt/commit/99b89858c09de8d8ef945d3d20cb2e089ffa8d6c) is my suggestion.

guiniol:
I really like this!
I have one suggestion: what do you guys think of showing the alternative headers before the content? ie. moving the content of the second `if (mutt_strcmp ("info", ShowMultipartAlternative) == 0)` to the first one?
With a little dose of `color attach_headers <foreground> <background> "Alternative Type #[0-9]+:"`, I think it would be really nice, harder to miss while not being too unobtrusive. That's only for the `info` option, obviously.

immae:
I think I screwed up somewhere (I made a git commit --amend with the changes asked by @gahr for the check), and it removed the change added by @flatcap . Sorry, I'm more used to gitlab or "flat" git, and didn't expect that behavior from github...

However it's done, should I also include devel/pull-390 ?

immae:
No opinion as to @guiniol suggestion for the position of the information, but I don't like your screenshot example which is misleading ("current type" \n "other type 1" \n "other type 2" \n "current type content" is not an expected order)

guyzmo:
indeed that's fucked up, I was pretty sure that github was behaving like gitlab on this regard (keeping refs to old commits after `git push -f`). They're both so alike, yet so different.

As information within a given alternative I believe the `info` should be shown as:

```
[-- 3 alternatives attachments: text/html, text/plain, text/calendar, showing text/plain --]
``` 

so when the multipart/alternative is within the first attachment of a multipart:

```
[-- Attachment #1 --]
[-- Type: multipart/alternative, blablabla --]

[-- 3 alternatives attachments: text/html, text/plain, text/calendar, showing text/plain --]

...
```

guyzmo:
That being said, we can start with what you've committed, and improve the `info` mode with another commit. Or maybe offer different more or less verbose ways to display this.

immae:
I think we should consider improvements in other PR's yes, it will be easier from there to see what could be improved once we start using it

gahr:
I've updated the devel/pull-390 branch, in case you still want to apply [53f5f09](https://github.com/neomutt/neomutt/commit/53f5f09d06f1703fd3556509b8f9e0f98c375b5f).

immae:
@gahr : applied

gahr:
@immae, thanks!

gahr:
> If you want to make me super-happy, could you please also provide an option value, say "default", for the default behavior? Thanks a lot!

So, how about this? Are we going to have the three valid values be `inline`, `info`, and ``?

flatcap:
@gahr
I fixed the option parsing to allow an empty string, in commit: c80cb4a

gahr:
@flatcap yes I saw that. I am asking myself if we really want an empty string to be a good value. Would we not prefer to have three sensible values, `inline`, `info`, and `default`?

gahr:
Oh, forget about it. I though `unset` would restore the default, but it actually sets the value to NULL no matter what the default is. So, yeah, we'll have to take care of the empty string anyway..

immae:
Is there anything missing to merge that? As far as I see every issue has been solved?

flatcap:
No.  Merged.  Thanks.

PR #391, opened 2017-02-10
Add alternate types information in viewer

immae:
This MR adds informations about alternates when viewing an email:

- First, it show which content type we're reading in the alternates
- Also, it shows the other available alternates

Please comment on the wording/style

This came from discussions in https://github.com/neomutt/neomutt/pull/390 but is independent

cc @d-k-c

### DIFF
--- a/handler.c
+++ b/handler.c
@@ -1004,6 +1004,8 @@ static int alternative_handler (BODY *a, STATE *s)
   int type = 0;
   int mustfree = 0;
   int rc = 0;
+  char length[5];
+  int count = 0;
 
   if (a->encoding == ENCBASE64 || a->encoding == ENCQUOTEDPRINTABLE ||
       a->encoding == ENCUUENCODED)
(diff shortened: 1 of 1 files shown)


gahr:
Very nice! I'll try this out over the weekend and possibly come up with comments 👍🏻

toogley:
suggestion for improvement:

* if only one content type is given, display that and don't show additional information (or at least very minimal)
* If more than one content type is given, display a shortened variant:

```
Attachments: multipart/alternative, text/plain
```

Thoughts?

immae:
@toogley : if you have a multipart/alternative you have more than one type, or it's useless:p

However I'm closing it, please see #390 who will contain the modifications here to

guyzmo:
@toogley I agree with you, but if it's not too much work to achieve.

I changed from bug to enhancement, because even if you think it's wrong, it was a deliberate choice from the people who originally implemented `multipart/alternative`

Issue #393, opened 2017-02-12
SegFault in complete_all_nm_tags


WPettersson:
I'm a Gentoo user, and I recently upgraded from 20160530 to 20170206, and since then the notmuch integration has been playing up for me. I've tried with both notmuch 0.23.3 and 0.23.5. The default keybind (\`) for `modify-labels` has not been applying, which is annoying, but also possibly related so I thought I'd mention it.

However, the real problem is that I'm getting segfaults in `complete_all_nm_tags` when tab-completing tags, see the backtrace below. Clearly the issue is that `candidate()` is not expecting `src` to be `0x0`. However, I cannot work out why `src` is being set to `0x0`. I've checked through gdb, and `nm_tags[0] = 0x0`, and `nm_tags[tag_count_1] = 0x0`, but in between these two (so `nm_tags[1]` through `nm_tags[138]`) are all set to the names of my labels. It seems like, for some reason, `nm_get_all_tags` is also adding an initial `0x0` tag. I don't know why.

    Program received signal SIGSEGV, Segmentation fault.
    __strstr_sse2_unaligned () at ../sysdeps/x86_64/multiarch/strstr-sse2-unaligned.S:40
    40      ../sysdeps/x86_64/multiarch/strstr-sse2-unaligned.S: No such file or directory.
    (gdb) bt
    #0  __strstr_sse2_unaligned () at ../sysdeps/x86_64/multiarch/strstr-sse2-unaligned.S:40
    #1  0x000000000042d1ae in candidate (src=0x0, len=256, try=0x6c61c0 <User_typed> "se", dest=0x6c6080 <Completed> "") at init.c:2872
    #2  0x000000000042d3af in complete_all_nm_tags (pt=pt@entry=0x7fffffffc401 "se") at init.c:3132
    #3  0x00000000004318c8 in mutt_nm_tag_complete (buffer=buffer@entry=0x7fffffffc400 "+se", len=len@entry=1024, pos=<optimized out>, numtabs=1) at init.c:3232
    #4  0x000000000042b26d in _mutt_enter_string (buf=buf@entry=0x7fffffffc400 "+se", buflen=buflen@entry=1024, col=19, flags=flags@entry=1024, multiple=multiple@entry=0, files=files@entry=0x0, 
        numfiles=0x0, state=0x843f60) at enter.c:691
[...]

PR #395, opened 2017-02-13
Notmuch sending bad info

WPettersson:
Fixes crashes if the notmuch database has an empty tag. See #393

### DIFF
--- a/init.c
+++ b/init.c
@@ -2956,6 +2956,9 @@ static void matches_ensure_morespace(int current)
 */
 static void candidate (char *dest, char *try, const char *src, int len)
 {
+  if (!dest || !try || !src)
+    return;
+
   int l;
 
   if (strstr (src, try) == src)
--- a/mutt_notmuch.c
+++ b/mutt_notmuch.c
@@ -2109,6 +2109,7 @@ int nm_get_all_tags(CONTEXT *ctx, char **tag_list, int *tag_count)
   struct nm_ctxdata *data = get_ctxdata(ctx);
   notmuch_database_t *db = NULL;
   notmuch_tags_t *tags = NULL;
+  const char *tag;
   int rc = -1;
 
   if (!data)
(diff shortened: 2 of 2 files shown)


guyzmo:
Thank you for this PR! 🙌

could you `git commit --amend; git push -f` and change your commit to something more explicit:

```
fixes notmuch giving NULL pointers for empty strings causing segfaults

this adds a check to make sure we feed only non-empty strings to the canditate list

fixes #393
```

that way it references the issue Id, and gives a proper summary and explaination of what the commit does ☺

WPettersson:
Done :) Sorry, forgot that I only had a brief commit message, I wasn't originally expecting to share that commit anywhere.

guyzmo:
@neomutt/reviewers please review & merge 🙌

gahr:
I'm not sure this fix is correct - perhaps we should avoid getting in this situation to begin with. What about we just skip empty tags in nm_get_all_tags?

```diff
diff --git a/mutt_notmuch.c b/mutt_notmuch.c
index c686e8ca..48a9871a 100644
--- a/mutt_notmuch.c
+++ b/mutt_notmuch.c
@@ -2111,6 +2111,7 @@ int nm_get_all_tags(CONTEXT *ctx, char **tag_list, int *tag_count)
   struct nm_ctxdata *data = get_ctxdata(ctx);
   notmuch_database_t *db = NULL;
   notmuch_tags_t *tags = NULL;
+  const char *tag;
   int rc = -1;

   if (!data)
@@ -2125,11 +2126,14 @@ int nm_get_all_tags(CONTEXT *ctx, char **tag_list, int *tag_count)

   while (notmuch_tags_valid(tags))
   {
-    if (tag_list != NULL)
+    tag = notmuch_tags_get(tags);
+    /* Skip empty string */
+    if (*tag)
     {
-      tag_list[*tag_count] = safe_strdup(notmuch_tags_get(tags));
+      if (tag_list != NULL)
+        tag_list[*tag_count] = safe_strdup(tag);
+      (*tag_count)++;
     }
-    (*tag_count)++;
     notmuch_tags_move_to_next(tags);
   }

```

flatcap:
There are a few problems here:
1. Notmuch gives us dodgy data (we can't help that)
2. We put an empty string in our list
3. `candidate()` doesn't check its args
4. `nm_get_all_tags()` is called twice (once to count, then once to build the list)

@gahr's diff is a better solution.

(3) Really needs fixing (in this PR is fine)
If someone would like to look at (4), we can open an issue for it.

gahr:
I'm not sure 1. is an issue, it might well be that an empty tag is a valid one, in notmuch's semantics. I do not know notmuch much (...) so I can't really comment on that (so why am I..?)

PR #396, opened 2017-02-13
Do not cast or check returns from safe_calloc

### LINKED ISSUE #388: Fix safe_calloc calls
Similar to #372.

1. do not cast returns
2. do not check results
3. put `number` and `size` arguments in the correct order

gahr:
Fixes #388

### DIFF
--- a/browser.c
+++ b/browser.c
@@ -604,7 +604,7 @@ static void init_state (struct browser_state *state, MUTTMENU *menu)
 {
   state->entrylen = 0;
   state->entrymax = 256;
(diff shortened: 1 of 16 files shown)


gahr:

guyzmo:
is there any benefit to use `calloc(1, ...)` instead of using `malloc(...)`?

gahr:
> is there any benefit to use calloc(1, ...) instead of using malloc(...)?

@guyzmo, calloc initializes your memory to all bits 0.

PR #397, opened 2017-02-13
Update the german translation

mineo:


po/de.po
```diff
@@ -113,7 +113,7 @@ msgstr "Adresse eingetragen."
 
 #: alias.c:394
 msgid "Error seeking in alias file"
-msgstr "Fehler beim Suchen in alias Datei"
+msgstr "Fehler beim Suchen in Addressbuchdatei"
```
guyzmo: technically an alias file is not an addressbook. I believe calling that way might cause confusion.

po/de.po
```diff
@@ -14,7 +14,7 @@ msgstr ""
 "Project-Id-Version: neomutt-20161104\n"
 "Report-Msgid-Bugs-To: neomutt-devel@neomutt.org\n"
 "POT-Creation-Date: 2017-01-13 14:32+0000\n"
-"PO-Revision-Date: 2016-11-30 06:38+0100\n"
+"PO-Revision-Date: 2017-02-13 20:19+0100\n"
 "Last-Translator: André Berger <andre.berger@web.de>\n"
```
guyzmo: You might want to add your name here

po/de.po
mineo: Strings from https://github.com/mineo/neomutt/blob/3d05bf38a3f9d99168ea1eac8736c2da9bc22c1e/po/de.po#L62 onwards translate alias-related things as Addressbuch, including https://github.com/mineo/neomutt/blob/3d05bf38a3f9d99168ea1eac8736c2da9bc22c1e/po/de.po#L107, which is the same as the string here except for a single word.

guyzmo: I would advocate to differentiate `alias` stuff from addressbook which are inherently two different things. An alias is just a short mnemonic you use for a long address string.

An addressbook on the other hand is an index you lookup someone on, based on a given filed. There applications to manage addressbooks, and mutt provides alias mechanism.

My opinion is that calling aliases addressbook can lead to confusion of what it really is.

But I'm not using a localised neomutt, so whatever feels better for native germans using l10n mutt translation. I won't stand in the way 😉

mineo: Aliases in mutt are not just short mnemonics for longer strings. You can use the alias database just like a real address book - when (for example) composing mail, hit `<TAB>` when asked to provide a the recipient address and you get a list of all entries in the alias file that you can now search by whatever you want.

I'm also not sure what a better translation would be.

Issue #225, opened 2016-11-06
keywords patch and IMAP FLAGS?


lrosenman:
any chance of NeoMutt using the Keywords to set IMAP FLAGS?

I'm seeing weird interaction, and would love to have it work.

PR #399, opened 2017-02-14
 Add support to IMAP custom keywords

sileht:
Add IMAP keywords support                                                                                                                                                                                                                                                        
                                                                                                                                                                                                                                                                                 
This change introduces IMAP keyword support.                                                                                                                                                                                                                                     
                                                                                                                                                                                                                                                                                 
* index_format gets %j and %J to display them                                                                                                                                                                                                                                    
* pattern gets j to search/match imap keywords                                                                                                                                                                                                                                   
* colors.c gets index_keywords to color them                                                                                                                                                                                                                                     
* OPS gets imap-edit-keywords to edit them   
                                                                                                                                                
Closes neomutt/neomutt#225   

Initial/obsolete PR text:
```
This change introduces a new configuration options to save x_label
in IMAP flags: keywords_imap.

When set to true, x_labels are used stored in IMAP keyword flag.

This is work in progress, TODO:
* support more than one label per message.
* check PERMANENTFLAGS for \* presence.
* maybe their should be encoded, but I can't find in which charset in the RFC
```


sileht:
Hello,

I'm proposing this WIP stuff, to know if your are interested in this or not ?

If yes, I finish it and make it clean.

flatcap:
Hi @sileht, welcome to NeoMutt.  Thanks for the PR, it looks very interesting.
We're currently discussing it in #neomutt (on freenode), if you're free.

guyzmo:
Hi @sileht and thank you for your PR! As @flatcap says, we've been discussing it on the channel.

I think the main thing we all agree on so far is that the `X-Label` feature and the imap keywords feature are two distinct things, so imap keywords shouldn't be implemented as `X-Label`.

Maybe reusing the same fake header `Tags` that notmuch uses could be a solution (along with the `%Y` expando), but there would still need to be some thought given to how the tag information should be stored by mutt, without mutating remote IMAP files.

xdgc:
There is pending code upstream that will collide with this.

* On 14 Feb 2017, Mehdi ABAAKOUK wrote: 
> This change introduces a new configuration options to save x_label
> in IMAP flags: keywords_imap.
> 
> When set to true, x_labels are used stored in IMAP keyword flag.
> 
> This is work in progress, TODO:
> * support more than one label per message.
> * check PERMANENTFLAGS for \* presence.
> * maybe their should be encoded, but I can't find in which charset in the RFC
> 
> Closes neomutt/neomutt#225
> You can view, comment on, or merge this pull request online at:
> 
>   https://github.com/neomutt/neomutt/pull/399
> 
> -- Commit Summary --
> 
>   * Allow to store labels in imap flags
> 
> -- File Changes --
> 
>     M doc/manual.xml.head (6)
>     M doc/muttrc.keywords (3)
>     M doc/vimrc.keywords (1)
>     M imap/imap.c (27)
>     M imap/message.c (37)
>     M init.h (9)
>     M mutt.h (3)
> 
> -- Patch Links --
> 
> https://github.com/neomutt/neomutt/pull/399.patch
> https://github.com/neomutt/neomutt/pull/399.diff
> 
> -- 
> You are receiving this because you are subscribed to this thread.
> https://github.com/neomutt/neomutt/pull/399

-- 

flatcap:
@xdgc 
I've already merged your new label code from upstream.
Is there more code to come?

guyzmo:
> There is pending code upstream that will collide with this.

thank you for the note!

I found: https://dev.mutt.org/trac/ticket/3210 

Which might be useful to reference from our ticket (and vice-versa)

I guess you're talking about:

• http://marc.info/?l=mutt-dev&m=148565914432075&w=4

but anyway, I think we all agree that implementing IMAP keywords using
`X-Label` is not a good idea. It's better to use another mechanism, like
the fake header `Tags` implemented by mutt_notmuch, as the properties of
the `X-label`s and the IMAP keywords are not bijectives.

xdgc:
Yes. There's a full unification of labels, keywords, and imap keywords.
It's been discussed on mutt-dev but not merged yet.  I expect to commit
it very soon.

* On 14 Feb 2017, Richard Russon wrote: 
> @xdgc 
> I've already merged your new label code from upstream.
> Is there more code to come?
> 
> -- 
> You are receiving this because you were mentioned.
> https://github.com/neomutt/neomutt/pull/399#issuecomment-279835404

-- 

xdgc:
I don't want to say that this patch will be obviated by the new upstream
code, but it will require changes. There is feature overlap, but perhaps
still room for improvement.

* On 14 Feb 2017, Richard Russon wrote: 
> @xdgc 
> I've already merged your new label code from upstream.
> Is there more code to come?
> 
> -- 
> You are receiving this because you were mentioned.
> https://github.com/neomutt/neomutt/pull/399#issuecomment-279835404

-- 

flatcap:
OK, thanks.
What was the subject of the thread?  I can't seem to find it on mutt-dev.

sileht:
I have updated the code to not use labels, but have distinct code dedicated for IMAP keywords, like we have for notmuch tags and labels.

I'm interested about the upstream thread about imap keywords, but I can't find it neither.

functions.h
```diff
@@ -102,6 +102,9 @@ const struct binding_t OpMain[] = { /* map: index */
   { "delete-subthread",		OP_DELETE_SUBTHREAD,		"\033d" },
   { "edit",			OP_EDIT_MESSAGE,		"e" },
   { "edit-label",		OP_EDIT_LABEL,			"Y" },
+#ifdef USE_IMAP
+  { "edit-keywords",		OP_EDIT_KEYWORDS,			NULL },
```
guyzmo: could you use a more explicit name, like `imap-edit-keywords`?

On a side note, this makes me realise that `modify-labels` (for notmuch) is not very homogeneous nor namespaced.

functions.h
```diff
@@ -238,6 +241,9 @@ const struct binding_t OpPager[] = { /* map: pager */
   { "clear-flag",       OP_MAIN_CLEAR_FLAG,		"W" },
   { "edit",		OP_EDIT_MESSAGE,		"e" },
   { "edit-label",	OP_EDIT_LABEL,			"Y" },
+#ifdef USE_IMAP
+  { "edit-keywords",		OP_EDIT_KEYWORDS,			NULL },
```
guyzmo: same as for main.

imap/message.c
```diff
@@ -69,6 +69,79 @@ static void imap_update_context (IMAP_DATA *idata, int oldmsgcount)
...
+  new = buf;
+  SKIPWS(new);
+  if (*new == '\0')
+    new = NULL;
+
+  changed = 0;
+  if (h != NULL) {
+    changed += keywords_message(Context, h, new);
+  } else {
+#define HDR_OF(index) Context->hdrs[Context->v2r[(index)]]
```
guyzmo: please keep `#define`s either indented (if clang-format permits this, ping @flatcap / @toogley for code styling note) or out of the block codes.

imap/message.c
```diff
@@ -69,6 +69,79 @@ static void imap_update_context (IMAP_DATA *idata, int oldmsgcount)
   }
 }
 
+char *imap_header_get_keywords(HEADER *h)
```
guyzmo: shouldn't the return value be `const`? As I guess the returned value of this function is not meant to be mutated.

guyzmo: also a doxygen header for the `imap_*` functions would be 👍 ☺

imap/message.c
gahr: Or move the first occurrence of this macro from [headers.c:302](https://github.com/neomutt/neomutt/blob/master/headers.c#L302) to a header file

gahr:
@sileht I checked out your branch and did a quick test using my GMail account. I haven't debugged it thoroughly but I do see some problems. 

I have added `%-5.5j` to my `index_format`, so I can see the IMAP Flags.

* flags are correctly displayed in the index, but fail to update after modifying them with `edit-label`, for example by changing a `NonJunk` flag to `Junk`
* `edit-label` sometimes fails to report the current label, as shown by `%j` in the index_format, so the prompt turns up blank. After setting a label, the subsequent `edit-label` commands correctly report the changed value

I think some more testing is due.

flatcap:
@sileht, @gahr 
While I love the enthusiasm, if @xdgc (the author of the upstream label code) has already solved this problem can we perhaps wait for his code to land?

gahr:
@flatcap of course!

sileht:
I have fixed all yours remarks, fix a couple of missing FREE, and add the keywords input checker.

sileht:
@flatcap I have something that works well now for my use case. I will use it as-is until the upstream stuff is released.

@gahr my last push should fix that bug.

gahr:
@sileht it looks like UI-wise it's ok now. But flags are not persisted across session. Disconnecting / reconnecting loses the modifications, at least with GMail. Can you reproduce?

sileht:
@gahr I'm doing my test with a Zimbra server and an dovecot server and it works with both. Does GMAIL provides '\*' for FLAGS CAPABILITIES ?

sileht:
@gahr Also the mailbox need to be synced after the keywords have been edited, but I guess this is done in your case.

gahr:
>  Does GMAIL provides '*' for FLAGS CAPABILITIES ?

Yes - if it was not the case, I think you should warn that `imap-edit-keywords` won't work

imap/message.c
toogley: @guyzmo as far as i know, clang-format doesn't allow indentation of #define statements.

sileht:
I have added the warning for missing \* flag, just is case :)

guyzmo:
>  While I love the enthusiasm, if @xdgc (the author of the upstream label code) has already solved this problem can we perhaps wait for his code to land?

maybe @xdgc's code and @sileht's could be merged to make the best out of both, and then pushed upstream? I think using a fake header to represent imap keywords is a better idea, that makes it possible to still use `X-Label` in parallel of the IMAP feature, the same way it's done with notmuch.

xdgc:
* On 15 Feb 2017, guyzmo wrote: 
> >  While I love the enthusiasm, if @xdgc (the author of the upstream label code) has already solved this problem can we perhaps wait for his code to land?
> 
> maybe @xdgc's code and @sileht's could be merged to make the best out of both, and then pushed upstream? I think using a fake header to represent imap keywords is a better idea, that makes it possible to still use `X-Label` in parallel of the IMAP feature, the same way it's done with notmuch.

I don't know, these may be on completely separate tracks.  I'm
specifically trying to integrate all varieties of keywords, with
mailbox-agnostic support (i.e. no matter whether local or IMAP).  There
should be no difference between an X-Label and an IMAP keyword.

I get the sense that @sileht's approach is directly opposite to that.

There may, however, be better support for various IMAP keyword
implementations in his code.  I don't know really.

I've been using my code for years actually, and it's good.  It broke
badly with some recent updates, so I'm rewriting parts of it for current
mutt.  I will post it to mutt-dev after 1.8.0.

-- 

xdgc:
* On 15 Feb 2017, Tobias Angele wrote: 
> toogley commented on this pull request.
> 
> 
> 
> > +   * SPACE          = ( 32. )
> +   * CTLS           = ( 0.-31., 127.)
> +   *
> +   * And must be separated by one space.
> +   */
> +
> +  new = buf;
> +  SKIPWS(new);
> +  if (*new == '\0')
> +    new = NULL;
> +
> +  changed = 0;
> +  if (h != NULL) {
> +    changed += keywords_message(Context, h, new);
> +  } else {
> +#define HDR_OF(index) Context->hdrs[Context->v2r[(index)]]
> 
> @guyzmo as far as i know, clang-format doesn't allow indentation of #define statements.

I don't follow.  Do you mean like this:

#ifndef foo
  #define foo
#endif

or like this:

#ifndef foo
#  define foo
#endif

The former is supported by some compilers but not all, and is not
standard.  The latter should be supported everywhere, AFAIK.  But I
don't see the indentation in the quoted code so not sure what you're
referring to.

-- 

gahr:
> The former is supported by some compilers but not all, and is not
standard. 

@xdgc It is standard, see §6.10 in C99 standard:

A preprocessing directive consists of a sequence of preprocessing tokens that satisfies the
following constraints: The first token in the sequence is a # preprocessing token that (at
the start of translation phase 4) is either the first character in the source file (**optionally
after white space containing no new-line characters**) or that follows white space
containing at least one new-line character.

sileht:
@gahr I have fixed the GMail bug, but in fact, I was just lucky with dovecot/zimbra.

xdgc:
* On 15 Feb 2017, Pietro Cerutti wrote: 
> > The former is supported by some compilers but not all, and is not
> standard. 
> 
> @xdgc It is standard, see §6.10 in C99 standard:

Good to know - but I've dealt with a lot of pre-C99 compilers, and this
doesn't always work.  The wonderful things about standards....

-- 

toogley:
Hey,

> On February 16, 2017 at 3:42 AM David Champion <notifications@github.com> wrote:
> I don't follow.  Do you mean like this:
> 
> #ifndef foo
>   #define foo
> #endif
> 
> or like this:
> 
> #ifndef foo
> #  define foo
> #endif

Sorry for being unclear, i meant both. If i remain correctly, both versions would be reformatted to

#ifndef foo
#define foo
#endif

guyzmo:
for the matter of the macros, my *opinion*, is that if a macro is defined inside an algorithm, I advocate that it shall be declared at the same level of indentation as the code surrounding it, e.g.:

```
function foo() {
    if (bleh) {
        #define FOO 42
        printf("%d\n", FOO);
        #undef FOO
    }
}
```

if clang-format does not support it, and in all other cases the macro directives should be outside of the C algorithm, defined at the top level.
I can't actually think of a good reason to use macro directives inline with code. For the use case above, you can define and undef FOO around the function.

About the side topic of indentation after the `#`, I prefer always having the hash on column 0, and indent the macros (for loooong macros it's easier to read). But if clang does not support that I'm ok with how it works.

long story short: having moved the macro in the header is 👍

guyzmo:
On Wed, Feb 15, 2017 at 06:39:53PM -0800, David Champion wrote:
> * On 15 Feb 2017, guyzmo wrote: 
> > >  While I love the enthusiasm, if @xdgc (the author of the upstream label code) has already solved this problem can we perhaps wait for his code to land?
> > maybe @xdgc's code and @sileht's could be merged to make the best out of both, and then pushed upstream? I think using a fake header to represent imap keywords is a better idea, that makes it possible to still use `X-Label` in parallel of the IMAP feature, the same way it's done with notmuch.
> 
> I don't know, these may be on completely separate tracks.  I'm
> specifically trying to integrate all varieties of keywords, with
> mailbox-agnostic support (i.e. no matter whether local or IMAP).  There
> should be no difference between an X-Label and an IMAP keyword.

I don't get why you think there should be no differences between X-Label
and IMAP keyword? Those two features are two independant backends for
storing tags, and they can run in parallel.

The X-Label is to store tags within the mail's headers, whereas the IMAP
keywords is accessible only through an IMAP connection.

So with your approach, you'd be introducing a form of redundancy between
the two backends. Also, AFAICT, the X-Label is being designed to accept
free form text, whereas the IMAP keywords are having a strict grammar,
and backend implementations can vary in how they behave. The IMAP
keywords is also in use by other MUAs (like Mail.app) to colourise the
mails.

Now add the fact that notmuch is also a third backend for a similar
label/keyword feature, I think there's room to factorise the behaviour,
but there's no one size fits all approach here.

> I've been using my code for years actually, and it's good.  It broke
> badly with some recent updates, so I'm rewriting parts of it for current
> mutt.  I will post it to mutt-dev after 1.8.0.

if it broke badly, it might be a good time to actually gather forces and
make a proper design that fits all the label tagging situations!

xdgc:
* On 16 Feb 2017, guyzmo wrote: 
> I don't get why you think there should be no differences between X-Label
> and IMAP keyword? Those two features are two independant backends for
> storing tags, and they can run in parallel.

Why does the backend matter? This is just a storage format/access
mechanism. It's still the same email message. If I apply keywords to an
IMAP message and then save it to a Maildir, should I lose the keywords?
I don't lose any other properties of the message when I do that.  Why
are keywords different?

Should I have separate search mechanisms for each keyword type in my
mailbox?

There are two other keywording standards in documented use even for flat
on-disk (non-IMAP) mailboxes.  Do we implement distinct feature sets for
those?


> The X-Label is to store tags within the mail's headers, whereas the IMAP
> keywords is accessible only through an IMAP connection.

As you say, that's a backend.  It's still functionally a keyword.


> So with your approach, you'd be introducing a form of redundancy between
> the two backends. Also, AFAICT, the X-Label is being designed to accept
> free form text, whereas the IMAP keywords are having a strict grammar,

Not sure what you mean by redundancy.  My view is that I'm reducing
redundant functionality by combining all these into a single feature
set.  It should have been done long ago.

I implemented X-Label in mutt originally.  Nothing else had
labels/tags/keywords, so there was no standard to follow, but the idea
was very much to allow arbitrary, abstract tagging.  Now there are
several standards, and it makes sense for mutt to play along with the
rest of the world.  But nothing prevents continuing to have freeform
text - we are just going to let you have more than one of them now.


> free form text, whereas the IMAP keywords are having a strict grammar,
> and backend implementations can vary in how they behave. The IMAP
> keywords is also in use by other MUAs (like Mail.app) to colourise the
> mails.

You might know more about how keywords are supported in the IMAP
protocol than I do, but I haven't seen any indication that this can't
work.


> Now add the fact that notmuch is also a third backend for a similar
> label/keyword feature, I think there's room to factorise the behaviour,
> but there's no one size fits all approach here.

Again, I don't see why not.


> > I've been using my code for years actually, and it's good.  It broke
> > badly with some recent updates, so I'm rewriting parts of it for current
> > mutt.  I will post it to mutt-dev after 1.8.0.
> 
> if it broke badly, it might be a good time to actually gather forces and
> make a proper design that fits all the label tagging situations!

My point of view is that we did that a long time ago.  If there are
actual problems, of course I want to address them.  But I do not want a
mail application with a separate subsystem for each approach to message
keyword tagging.  That's not useful.

-- 

sileht:
Hello,

Main differences between X-label and IMAP keywords^Wflags, are:
* IMAP keywords^Wflags allow only a subset of ascii character. 
* X-label are part of the message and can be forwarded via SMTP, 
  'IMAP flags are not'

On Thu, Feb 16, 2017 at 02:22:13PM -0800, David Champion wrote:
>* On 16 Feb 2017, guyzmo wrote:
>> I don't get why you think there should be no differences between X-Label
>> and IMAP keyword? Those two features are two independant backends for
>> storing tags, and they can run in parallel.
>
>Why does the backend matter? This is just a storage format/access
>mechanism. It's still the same email message. If I apply keywords to an
>IMAP message and then save it to a Maildir, should I lose the keywords?
>I don't lose any other properties of the message when I do that.  Why
>are keywords different?
>
>Should I have separate search mechanisms for each keyword type in my
>mailbox?
>
>There are two other keywording standards in documented use even for flat
>on-disk (non-IMAP) mailboxes.  Do we implement distinct feature sets for
>those?

From the UX point of view, I agree it would be nice to have everything
at the same place, with same option/keybinding/format/... . But I don't
see where we can put this additional IMAP flags when a message is copied
to a maildir/mbox. We could add a new 'mutt-imap-keywords' header but
this one will be not supported by any other application.

For example, in thunderbird, IMAP flags are not copied to maildir/mbox
message, but in the thundebird index file of the maildir/mbox, so you
will lost this flags when you switch the MUA.

Other example, mbsync just does not copy imap flags to maildir. And
just because it not standardized to maildir format.

>> The X-Label is to store tags within the mail's headers, whereas the IMAP
>> keywords is accessible only through an IMAP connection.
>
>As you say, that's a backend.  It's still functionally a keyword.

If an user want to put keywords on it's IMAP message, how will we choice
where we store them ? Perhaps he want X-Label or "IMAP flags" or Both ?

What about a message that have X-Label from maildir with some utf-8 chars,
copied to IMAP server where user want use 'IMAP flags' for keywords ?

Because of the free-form of X-label and the very-limited-form of 'IMAP
flags', I think it will be really really tricky to merge them.

>> So with your approach, you'd be introducing a form of redundancy between
>> the two backends. Also, AFAICT, the X-Label is being designed to accept
>> free form text, whereas the IMAP keywords are having a strict grammar,
>
>Not sure what you mean by redundancy.  My view is that I'm reducing
>redundant functionality by combining all these into a single feature
>set.  It should have been done long ago.
>
>I implemented X-Label in mutt originally.  Nothing else had
>labels/tags/keywords, so there was no standard to follow, but the idea
>was very much to allow arbitrary, abstract tagging.  Now there are
>several standards, and it makes sense for mutt to play along with the
>rest of the world.  But nothing prevents continuing to have freeform
>text - we are just going to let you have more than one of them now.
>
>
>> free form text, whereas the IMAP keywords are having a strict grammar,
>> and backend implementations can vary in how they behave. The IMAP
>> keywords is also in use by other MUAs (like Mail.app) to colourise the
>> mails.
>
>You might know more about how keywords are supported in the IMAP
>protocol than I do, but I haven't seen any indication that this can't
>work.
>
>
>> Now add the fact that notmuch is also a third backend for a similar
>> label/keyword feature, I think there's room to factorise the behaviour,
>> but there's no one size fits all approach here.
>
>Again, I don't see why not.
>
>
>> > I've been using my code for years actually, and it's good.  It broke
>> > badly with some recent updates, so I'm rewriting parts of it for current
>> > mutt.  I will post it to mutt-dev after 1.8.0.
>>
>> if it broke badly, it might be a good time to actually gather forces and
>> make a proper design that fits all the label tagging situations!
>
>My point of view is that we did that a long time ago.  If there are
>actual problems, of course I want to address them.  But I do not want a
>mail application with a separate subsystem for each approach to message
>keyword tagging.  That's not useful.

If you can drop somewhere the patch/discussion about the "upstream
unification of labels", even it's unfinished. It will help a lot to see
if that easy to fit 'IMAP flags' in or not.


Maybe another solution here, is to not call this feature "IMAP keywords"
but 'IMAP custom flags'. And no more confusion about what a keywords is.
ie: "IMAP flags" are not keywords.

Some additional notes about upstream and neomutt difference about
keywords. notmuch backend already have it own code for tagging and this
is not shared with X-label neither.

I don't think, people want to have its notmuch tags copied to a maildir
or 'IMAP flags', too.

My point of view is "notmuch tags"/"IMAP flags" are user only visible
classifier that depends on the used backend, when "X-label" is part of
the message whatever the backend you use. Trying to implements a thing
standardized in a backend to a not standardized one will obviously
lead to a "Mutt specific behavior". And I think it's worse than have 3
different tagging systems.

Cheers,
--
Mehdi Abaakouk
mail: sileht@sileht.net
irc: sileht

gahr:
It is my interpretation of the standard and current custom that IMAP Flags are meant to carry **client**-specific info such as colouring (Mail.app) or implementation-specific labels (`Junk`,  `NotJunk` in GMail).

This is in contrast to the `X-Label` header, which is meant to carry **user**-defined tags.

There are in my opinion two consequences that come from this difference:
1. IMAP Flags should be machine-readable, whereas `X-Labels` should be user-readable. This is also clear from the fact that IMAP Flags have a well-defined syntax, while X-Labels can take wathever value a header can, including being encoded in something different than us-ascii.
2. IMAP Flags should not (or even cannot) be propagated outside of the specific session environment, meant as combination of a single server/client instance that agree on their semantics, whereas `X-Labels` are propagated as part of the rfc822 message.

For these reasons, I think it is wrong to allow the user to store arbitrary information in IMAP Flags.

As a contrasting example, I would be in favour of using this mechanism to enhance the user experience, e.g., to extend flagging (`!`) with multiple priorities.

sileht:
We have this kind of human readable vs machine readable idea in notmuch backend. We can define
tag-transform and tag-formats to transform notmuch tag format into more useful stuff for the UI.

sileht:
Here, the current capability of different tags/labels/keywords/flags:

With my current patch we have two concepts: tags and labels

* labels are editing of the 'X-label' header, I haven't touch the upstream code at all
* tags are editing of tags using the driver capabilities

| concept | drivers compat  | transferable between driver | transformable in UI
| --- | --- | --- | --- | 
| labels              | all | yes | no | 
| tags | notmuch and imap | no | yes | %g %Gx %J

Also the authorized set of characters on each differ. (I will create a list for each). For tag it also depends on the driver capability.

Current keybinding:
* labels: "edit-label"
* tags: "modify-labels",  "modify-labels-then-hide", "quasi-delete" 

I would suggest to rename tags keybinding to: edit-tags, edit-tags-then-hide, is that ok ?
Should I keep old name for compat ?

Current expando:
* labels: %y %Y
* tags: %g %J %Gx

%y is similar to %g, and %Y to %J. But if %y and %g is used at the same times we display both.
Merging the both concept will not allow this.

So, the big pending question is, should we merge these two concepts or not ?
* Does using both concepts at the same times make sense ? 
  * Since tags are not supported by all backends, I would say yes.
* Should we merge keybinding and expando of the two concepts and create a configuration option to select which concept we want ? 
  *This will break config of notmuch and/or X-labels user.
* If we merge them, which name to we keep ? labels ? tags ?

sileht:
Last push fixes the compilation issue with notmuch is not enabled and add the tidying code patch from @flatcap

gahr:
I know I might be a bit late to the party, but I don't think this was ready for merge. A few after-merge bugs have been spotted already, and my reviewing of the merge commit revealed more things that warrant at least some further discussion. If not on the design, at least in the implementation.

I don't see this pull req having been approved by anyone, and its status in still in-progress.

Would it be too much from me to ask to revert the merge and resume discussion?

gahr:
> So, the big pending question is, should we merge these two concepts or not ?

On the design side, I don't think this question was properly answered.

PR #401, opened 2017-02-14
Fixes missing failure return of notmuch msg open


### LINKED ISSUE #308: Segfault when opening a discarded postponed mail
### Neomutt version/configuration

* `NeoMutt 20161126 (1.7.1)` 
 * compiled with `Configure options: '--prefix=/usr' '--sysconfdir=/etc' '--enable-debug' '--enable-pgp' '--enable-gpgme' '--enable-notmuch' '--enable-pop' '--enable-imap' '--enable-smtp' '--enable-hcache' '--enable-sidebar' '--enable-compressed' '--with-gss=/usr' '--with-ssl=/usr' '--with-sasl' '--with-curses=/usr' '--with-regex' '--with-idn' '--with-gdbm' 'CFLAGS=-march=x86-64 -mtune=generic -O2 -pipe -fstack-protector-strong' 'LDFLAGS=-Wl,-O1,--sort-common,--as-needed,-z,relro' 'CPPFLAGS=-D_FORTIFY_SOURCE=2'`
* yes I'm still using the former version on the machine I'm reading my mail 😇 so yes, it has to be reproduced with latest
* I'm only using notmuch vfolder searches to list my mails

### How to reproduce?

* select a notmuch vfolder with `:sidebar-open`
* respond to a mail
* edit the message and save it
* exit and postpone the message
* sidebar-open the current mailbox, so it shows the postponed mail in the list
* write a new mail, and say `y`es to reopening a postponed message
* exit and discard the message
* `notmuch new`
* in the browser list, without refreshing, open that postponed mail in the list
 * 🚧🔥 **CRASH** 🔥🚧

guyzmo:
When calling the `mx_ops->open_msg()` method of the notmuch
implementation, if the message failed to open, it is not reported to the
user interface. This patch returns a non-zero value that will prevent a
crash.

fixes #308

### DIFF
--- a/mutt_notmuch.c
+++ b/mutt_notmuch.c
@@ -2430,7 +2430,7 @@ static int nm_open_message(CONTEXT *ctx, MESSAGE *msg, int msgno)
     msg->fp = maildir_open_find_message(folder, cur->path, NULL);
 
   mutt_debug (1, "%s\n", __func__);
-  return 0;
+  return !msg->fp;
 }
 
 static int nm_close_message(CONTEXT *ctx, MESSAGE *msg)


guyzmo:
@neomutt/reviewers please, review! 🙌

PR #402, opened 2017-02-15
Kill homebrewed (v)snprintf implementations, as they are already available in the C99 standard

gahr:


toogley:
@gahr what exactly means "as they are C99" ? i thought we would write C99 in mutt?

gahr:
@toogley yes, that's the point. Those functions are in standard C99, so if we assume a C99 compiler, there's no reason to keep an implementation in case they are not there.

d-k-c:
Amen to that. Do you know if that implies Neomutt will break on some platforms because of incomplete C99 support by some compilers?

gahr:
@d-k-c I have no idea, sorry.

toogley:
@neomutt/downstream-packagers do you guys know sth about [the question](https://github.com/neomutt/neomutt/pull/402#issuecomment-280063606) of @d-k-c ?

0-wiz-0:
That commit is no problem for NetBSD. NetBSD's man page says that both functions appeared in 4.4BSD, so it shouldn't be a problem for any of the BSDs.

lbschenkel:
I am not an expert, but I believe that all OS X versions supported by MacPorts implement these calls.

mutt.h
```diff
@@ -26,9 +26,6 @@
 #ifdef HAVE_UNISTD_H
 # include <unistd.h> /* needed for SEEK_SET */
 #endif
-#ifdef HAVE_UNIX_H
-# include <unix.h>   /* needed for snprintf on QNX. */
```
darnir: I'm not entirely sure, but if this allows us to keep compatibility with some esoteric platform, then it should be okay. 3 lines of code here won't hurt us much.

Does anyone know if Neomutt still compiles on QNX and if this ifdef is required on it?

neverpanic:
Even PPC Macs had those functions already, so this is fine from a Mac perspective.

mutt.h
guyzmo: I searched through the docs of QNX, and found that ① they're C99 compliant and ② there's a [`stdio.h` featuring snprintf](http://www.qnx.com/developers/docs/6.5.0SP1.update/index.html#./com.qnx.doc.dinkum_en_c99/stdio.html).

So I vote for removing those useless three lines.

xdgc:
* On 15 Feb 2017, Damien R. wrote: 
> Amen to that. Do you know if that implies Neomutt will break on some platforms because of incomplete C99 support by some compilers?

It certainly will.  Some compilers were made before '99....

-- 

Vaelatern:
It would break support for old compilers, but remotely modern gcc and friends have c99 available.

d-k-c:
> It certainly will. Some compilers were made before '99....

Mutt requires C99 since August 2015, so the question is "are there compilers that claim to be C99 compliant but which actually aren't".

guyzmo:
> the question is "are there compilers that claim to be C99 compliant but which actually aren't".

and the answer should be *screw them*.

If you're stuck with a system that does not have a C99 compliant compiler on which you cannot build one (like GCC) then you're stuck with using an old version of mutt.

toogley:
@d-k-c 
> Mutt requires C99 since August 2015, so the question is "are there compilers that claim to be C99 compliant but which actually aren't".

What's the source for that? I mean, can you cite your source?^^

flatcap:
@toogley 
`AC_PROG_CC_C99` was added to `configure.ac` by Kevin McCarthy on 2015-08-17 (a9676b02)
This means a C99-compliant compiler is required.

gahr:
Plus, [this check](https://github.com/neomutt/neomutt/blob/master/configure.ac#L33-L35)

xdgc:
* On 16 Feb 2017, guyzmo wrote: 
> > the question is "are there compilers that claim to be C99 compliant but which actually aren't".
> 
> and the answer should be *screw them*.
> 
> If you're stuck with a system that does not have a C99 compliant compiler on which you cannot build one (like GCC) then you're stuck with using an old version of mutt.

I talked about that in another thread.  Neomutt can do what you like of
course, but mutt should not abandon users without making it clear in
advance that there's a cutoff coming.


> @d-k-c 
> > Mutt requires C99 since August 2015, so the question is "are there compilers that claim to be C99 compliant but which actually aren't".

I didn't remember this.  If that decision has been made already, then I
have no objection on the basis of a support obligation.

But once again, I'm not sure what's the big win in removing passive
compatibility code that hasn't needed maintenance in over ten years.
The only thing achieved is less user support.  Nothing future-looking is
enabled by the change.

-- 

guyzmo:
On Thu, Feb 16, 2017 at 02:29:48PM -0800, David Champion wrote:
> I talked about that in another thread.  Neomutt can do what you like of
> course, but mutt should not abandon users without making it clear in
> advance that there's a cutoff coming.

Of course, my former remark is based on the fact below. ↓↓↓

> > @d-k-c 
> > > Mutt requires C99 since August 2015 […]
> I didn't remember this.  If that decision has been made already, then I
> have no objection on the basis of a support obligation.

👍

> But once again, I'm not sure what's the big win in removing passive
> compatibility code that hasn't needed maintenance in over ten years.

It's about removing bitrot, the part of code that, in theory, will never
get included at compile time by any user. And somehow, the only way to
prove that theory, it's to remove if from the codebase and wait for
someone to yell about it.

> The only thing achieved is less user support.  Nothing future-looking is
> enabled by the change.

It's like taking the garbage out for collecting. Nothing future-looking
is enabled by taking the garbage out, but you're still taking them out! ☺

-- 
Guyzmo

Issue #69, opened 2016-07-05
Option for robust macros with shell-escapes


scottkosty:
This is a feature request for an option that, if set, would cause a macro to end after any <shell-escape> that exits with error (perhaps other commands as well, such as <pipe-message>).

motivation: I have several macros with shell-escapes, and although I can report the error to the mutt user and e.g. set the muttrc wait_key option so that I know the user sees that the script exited with error, in some cases it makes for some frustration that mutt continues executing the macro after there was an error. For example, suppose I have a macro such as:

`macro compose [u "<shell-escape>foo<enter><send-message>"`

If 'foo' exits with error, I don't want for mutt to send the message. There are some hackish ways around this particular case (e.g. set a sendmail command that checks for a temporary file that 'foo' wrote containing the exit code), but I hope this case illustrates my point well enough.

Framed in another way, I would like for a way that <shell-escape>abc<def> is really interpreted as a shell would interpret `<shell-escape>abc && def`. Currently it is interpreted as `<shell-escape>abc ; def`.

I don't know much about how this would be implemented, but I'll give some uneducated ideas:
I'm guessing that a goal of macros is to be as true as possible to if the user had just typed in all the keys themself. In this case, it would seem to contradict my desired feature (because if a user keeps typing after an error, that is still of course input that should be interpreted). If this principle cannot be violated (with an option that defaults to the current behavior), then what about the following?

A muttrc variable "wait_on_error" of type "number" that designates the number of seconds that mutt would gobble any keystrokes after an error? For example, if set to "1", any keystroke after a <shell-escape> exits non-zero, would be gobbled up (= "not interpreted") for the duration of one second. This way, the behavior of macros would be the same as if the user types the macro (1000 wpm) directly themselves.

Note to @flatcap, I don't expect you to spend the time to implement this. I'm just curious if you're open to the idea. I'm guessing there might be reasons why such an option would be a bad idea. I hope not though. If you are open to the idea, then perhaps someone else interested in robust macros would want to work on implementing it.

flatcap:
A good idea, well argued :-)

If you're only going for the simple case `X && Y {&& Z ...}` then it should be quite easy to implement.

scottkosty:
@flatcap thanks for the feedback! I'm glad you are open to the idea. Hopefully someone else takes an interest in implementing it.

flatcap:
Sorry, this enhancement didn't attract someone to write some code.
[...]

PR #403, opened 2017-02-15
Notmuch: Fixed modify-labels command on folders

andir:
When calling `<modify-labels>` on non-virtual folders with the toggle
operator ("!") the notmuch code would break early and leave the rest of
the arguments within the input buffer. Mutt would then interpret the "!"
as OP_SHELL_ESCAPE and pass the label to /bin/sh.

To fix this behaviour two changes have been made:

- When not within a virtual folder the error will now be displayed using
  `mutt_error` which causes mutt to stop executing the current macro.
  Without this change other instructions in the macro would be called.
  e.g. causing the ui to change, reload of inboxes etc.

- Remaining arguments in the for the \<modify-labels\> command will be
  discarded.

### DIFF
--- a/curs_main.c
+++ b/curs_main.c
@@ -1801,11 +1801,13 @@ int mutt_index_menu (void)
       case OP_MAIN_MODIFY_LABELS:
       case OP_MAIN_MODIFY_LABELS_THEN_HIDE:
       {
-	if (Context->magic != MUTT_NOTMUCH) {
-	  mutt_message (_("No virtual folder, aborting."));
-	  break;
-	}
-	CHECK_MSGCOUNT;
+        if (Context->magic != MUTT_NOTMUCH) {
+          // remove remaining arguments
+          mutt_get_field ("", buf, sizeof(buf), MUTT_NM_TAG);
+          mutt_error (_("No virtual folder, aborting."));
+          break;
+        }
+        CHECK_MSGCOUNT;
         CHECK_VISIBLE;
 	*buf = '\0';
 	if (mutt_get_field ("Add/remove labels: ", buf, sizeof (buf), MUTT_NM_TAG) || !*buf)


guiniol:
@andir I didn't realise that you were using `mutt_error`. Is the `mutt_get_field` still necessary in that case?

andir:
Yes it is. Otherwise mutt does still execute `<sync-mailbox>` or whatever is left in the macro.

curs_main.c
```diff
@@ -1802,7 +1802,9 @@ int mutt_index_menu (void)
       case OP_MAIN_MODIFY_LABELS_THEN_HIDE:
       {
 	if (Context->magic != MUTT_NOTMUCH) {
-	  mutt_message (_("No virtual folder, aborting."));
+	  // remove remaining arguments
+	  mutt_get_field ("", buf, sizeof(buf), MUTT_NM_TAG);
```
guyzmo: I don't think using `mutt_get_field()` is necessary. If I'm not wrong, it's creating a useless input for the user, and in this case without any prompt.

The way the buffer is discarded by all functions is to do:

```
buf[0] = '\0';
```

guyzmo: Another thing: could you re-indent this block removing tabs? 😀

guyzmo:
Otherwise, good catch on this issue 👍

curs_main.c
andir: `buf[0] = '\0';` is probably there to truncate the input buffer size to zero.

I tried it instead of the `mutt_get_field` call and it brought back the original issue.

guyzmo:
err… actually I don't understand your issue.

Can you describe the exact set of actions you do to trigger the issue? More exactly how to you pass arguments to `<modify-labels>`?

guyzmo:
ok, so this is an [X-Y problem](http://xyproblem.info/):

* OP has defined a macro: `macro index,pager F "<modify-labels>!foobar<return>"`
* when not in notmuch context the macro fails and does not ask for labels
* the `!foobar` label is now executed like a mutt command,
* `!` launches a shell command `foobar`.

Is this an issue? Actually not. The `modify-labels` command really is built to be used as a binding, and when the binding is available we don't want to have typed useless tags.

Is this a liability? Yes, of course. You don't want to run arbitrary commands in that kind of context, even though the risk is pretty limited.

The solution?

Use a hook, that you unset everywhere you're not using it:

```
folder-hook * macro index,pager F noop
folder-hook notmuch:// macro index,pager F "<modify-labels>!foobar<return>"
```

but the better solution is being discussed in the issue #69: halt execution of any macro within which a command has failed.

PR #406, opened 2017-02-16
Remove snprintf.c from EXTRA_DIST

gahr:
Pull #402


gahr:
Merging this build fix.

Issue #407, opened 2017-02-16
Interrupted system call error with non-ssl imap makes neomutt unusable


EdwardBetts:
I read my mail via IMAP using the dovecot IMAP server on my laptop. The connection is not using ssl.

Mutt keeps crashing with this message: "Error talking to localhost (Interrupted system call)". It didn't used to do this. 

The problem goes away if I configure dovecot to use ssl.

I used git bisect to find when this problem started, it is commit 1f19f808dbcfe1cb02e4806a9f583ae9c657b7e5.

Steps to reproduce:
1. run an IMAP server with ssl disabled.
2. mutt -f imap://localhost/
3. while true; do echo WINCH; killall -WINCH mutt; sleep 1; done

Sending WINCH to mutt is a harmless windows resize signal. It will trigger the bug quicker.

gahr:
@EdwardBetts , I can test with a local dovecot over plain port 143. I have been running the loop for over a minute but I can't get it to crash. How long does it usually take?

Also, can you post your `mutt -v` and `mutt -DS`? Thanks!

EdwardBetts:
Normally within seconds.

Try with these settings:

set mail_check=1
set timeout=2
unset ssl_starttls

[mutt-v.txt](https://github.com/neomutt/neomutt/files/780343/mutt-v.txt)
[mutt-D.txt](https://github.com/neomutt/neomutt/files/780344/mutt-D.txt)

gahr:
`unset ssl_starttls` did it. Thanks, I can now reproduce.

EdwardBetts:
How come the "Interrupted system call" error only happens with SSL disabled?

EdwardBetts:
I often get "Mailbox closed" after the "Interrupted system call" error. The message list disappears and the status bar says "(no mailbox)".

gahr:
> How come the "Interrupted system call" error only happens with SSL disabled?

Because of the difference between how the [plain](https://github.com/neomutt/neomutt/blob/master/mutt_socket.c#L428) and the [ssl](https://github.com/neomutt/neomutt/blob/master/mutt_ssl.c#L319) sockets handle EINTR.

EdwardBetts:
The raw_socket_write function starts with mutt_allow_interrupt (1) and ends with mutt_allow_interrupt (0). The ssl_socket_read doesn't call mutt_allow_interrupt. Does that mean interrupts are permitted within raw_socket_write but disallowed within ssl_socket_read?

[...]

PR #408, opened 2017-02-16
Avoid "Interrupted system call" IMAP errors

EdwardBetts:
When trying to use mutt with non-SSL IMAP it intermittently throws 'Error talking to localhost (Interrupted system call)' and closes the mailbox. This fixes neomutt/neomutt#407.

### DIFF
--- a/mutt_socket.c
+++ b/mutt_socket.c
@@ -402,12 +402,7 @@ int raw_socket_read (CONNECTION* conn, char* buf, size_t len)
     mutt_error (_("Error talking to %s (%s)"), conn->account.host,
 		strerror (errno));
     mutt_sleep (2);
-  } else if (errno == EINTR) {
-    rc = -1;
-    mutt_error (_("Error talking to %s (%s)"), conn->account.host,
-               strerror (errno));
-    mutt_sleep (2);
-   }
+  }
   mutt_allow_interrupt (0);
 
   return rc;
@@ -422,11 +417,6 @@ int raw_socket_write (CONNECTION* conn, const char* buf, size_t count)
   {
     mutt_error (_("Error talking to %s (%s)"), conn->account.host,
 		strerror (errno));
-    mutt_sleep (2);
-  } else if (errno == EINTR) {
-    rc = -1;
-    mutt_error (_("Error talking to %s (%s)"), conn->account.host,
-               strerror (errno));
     mutt_sleep (2);
   }
   mutt_allow_interrupt (0);


guyzmo:
We'd need the input on this by @aradici especially about commit 1f19f808dbcfe1cb02e4806a9f583ae9c657b7e5 which sounds to me like a huge misusage of the `mutt_allow_interrupt` call as using `mutt_allow_interrupt(0)` does *NOT* disable the sighandler, it's only resetting it to the normal behaviour.

I believe that the `mutt_allow_interrupt()` calls used as some sort of "scope" is wrong and should go away. Also the name of that function is misleading and it might be a good idea to change it.

mutt_socket.c
```diff
@@ -423,11 +418,6 @@ int raw_socket_write (CONNECTION* conn, const char* buf, size_t count)
     mutt_error (_("Error talking to %s (%s)"), conn->account.host,
 		strerror (errno));
     mutt_sleep (2);
-  } else if (errno == EINTR) {
```
guyzmo: to catch if there was an error writing to the socket, one can simply compare `rc` with `count`, if they differ then the connection has ended abnormally, maybe as a result of a sigint, which can be checked with the presence of `SigInt` being set, something like:

```
if (rc != count)
{
  if (SigInt)
  {
    mutt_error (_("Connection to %s has been aborted"));
    SigInt = 0;
  }
  else
  {
    mutt_error (_("Unknown connection error while sending data to %s."), conn->account.host);
  }
  mutt_sleep (2);
}
```

of course that would need to be tested, especially with a `C-c` abort and forcing another issue. Not being an user of the IMAP features, I'm not sure how to create a testcase to check that code.

guyzmo: (for those who see this as mail, I moved this to the write function, and replaced the other with the read issue)

to catch if there was an error writing to the socket, one can simply compare `rc` with `count`, if they differ then the connection has ended abnormally, maybe as a result of a sigint, which can be checked with the presence of `SigInt` being set, something like:

```
if (rc != count)
{
  if (SigInt)
  {
    mutt_error (_("Connection to %s has been aborted"));
    SigInt = 0;
  }
  else
  {
    mutt_error (_("Unknown connection error while sending data to %s."), conn->account.host);
  }
  mutt_sleep (2);
}
```

of course that would need to be tested, especially with a `C-c` abort and forcing another issue. Not being an user of the IMAP features, I'm not sure how to create a testcase to check that code.

EdwardBetts: Here is the beginnings of a test case written in Python. It creates an IMAP server then runs mutt and connects to the the server.  

```python
#!/usr/bin/python3

import threading
import socketserver
import pexpect
import os
from tempfile import NamedTemporaryFile

verbose = True

class MyServer(socketserver.TCPServer):
    allow_reuse_address = True

class SimpleIMAPHandler(socketserver.StreamRequestHandler):
    continuation = None

    def setup(self):
        super().setup()
        self.server.logged = None

    def _send(self, message):
        if verbose:
            print("SENT: %r" % message.strip())
        self.wfile.write(message)

    def _send_line(self, message):
        self._send(message + b'\r\n')

    def _send_textline(self, message):
        self._send_line(message.encode('ASCII'))

    def _send_tagged(self, tag, code, message):
        self._send_textline(' '.join((tag, code, message)))

    def handle(self):
        # Send a welcome message.
        self._send_textline('* OK IMAP4rev1')
        while 1:
            line = b''
            while True:
                part = self.rfile.read(1)
                if part == b'':
                    return
                line += part
                if line.endswith(b'\r\n'):
                    break

            if verbose:
                print('GOT: %r' % line.strip())
            if self.continuation:
                try:
                    self.continuation.send(line)
                except StopIteration:
                    self.continuation = None
                continue
            splitline = line.decode('ASCII').split()
            tag = splitline[0]
            cmd = splitline[1]
            args = splitline[2:]

            if hasattr(self, 'cmd_' + cmd):
                continuation = getattr(self, 'cmd_' + cmd)(tag, args)
                if continuation:
                    self.continuation = continuation
                    next(continuation)
            else:
                self._send_tagged(tag, 'BAD', cmd + ' unknown')

    def cmd_CAPABILITY(self, tag, args):
        self._send_textline('* CAPABILITY IMAP4rev1')
        self._send_tagged(tag, 'OK', 'CAPABILITY completed')

    def cmd_LOGOUT(self, tag, args):
        self.server.logged = None
        self._send_textline('* BYE IMAP4ref1 Server logging out')
        self._send_tagged(tag, 'OK', 'LOGOUT completed')

    def cmd_LOGIN(self, tag, args):
        self.server.logged = args[0]
        self._send_tagged(tag, 'OK', 'LOGIN completed')

    def cmd_LIST(self, tag, args):
        self._send_textline(r'* LIST (\Noselect) "/" ""')
        self._send_tagged(tag, 'OK', 'LIST completed')

    def cmd_SELECT(self, tag, args):
        self._send_textline('* 0 EXISTS')
        self._send_textline('* 0 RECENT')
        self._send_tagged(tag, 'OK', '[READ-WRITE] SELECT completed')

    def cmd_NOOP(self, tag, args):
        self._send_tagged(tag, 'OK', 'NOOP completed')

    def cmd_CLOSE(self, tag, args):
        self._send_tagged(tag, 'OK', 'CLOSE completed')

def write_muttrc(port):
    f = NamedTemporaryFile(delete=False)
    muttrc = f.name
    f.write(b'''
set folder=imap://x:x@localhost:%d/
set mail_check=1
set timeout=2
unset help
unset ssl_starttls
''' % port)
    f.close()
    return muttrc

def read_loop(mutt):
    while True:
        try:
            buf = mutt.read_nonblocking(mutt.maxread, timeout=1)
        except pexpect.TIMEOUT:
            break
        print('from mutt:', repr(buf))

def run_mutt(port, muttrc):
    mbox = 'imap://x:x@localhost:{}'.format(port)
    mutt = pexpect.spawn('mutt', ['-F', muttrc, '-f', mbox], encoding='utf-8')

    read_loop(mutt)

    print('send mutt: imap-fetch-mail')
    mutt.send(':exec imap-fetch-mail\n')
    read_loop(mutt)

    print('send mutt: quit')
    mutt.send('q')
    try:
        read_loop(mutt)
    except pexpect.exceptions.EOF:
        pass

def main():
    server = MyServer(('localhost', 0), SimpleIMAPHandler)
    port = server.server_address[1]

    thread = threading.Thread(target=server.serve_forever)
    thread.daemon = True
    thread.start()

    muttrc = write_muttrc(port)
    run_mutt(port, muttrc)
    os.remove(muttrc)


if __name__ == '__main__':
    main()```

PR #411, opened 2017-02-18
handle sigint within socket operations

EdwardBetts:
I extended my previous attempt by adding @guyzmo's suggestion for handling SigInt. I've tested this with a mock IMAP server written in Python. It works and is able to abort a slow IMAP fetch.


gahr:
What happens if we get a sigint after read or write returns -1, say, because writing the first error or sleeping the first time is interrupted? Do we get two messages and a sleep of 4 seconds?

EdwardBetts:
Thanks. Good suggestions. I'll tidy up this change and resubmit.

EdwardBetts:
I've updated my change to set SigInt to 0 when a read/write error has already been reported, so the error isn't reported twice.

Issue #414, opened 2017-02-19
Functional and unit tests for CI


guyzmo:
following #222 and all the others discussion about testing. It took me some time, but I finally found out a decent way to implement functional and unit testing for (neo)mutt.

## functional tests

Those are tests being done using the mutt runtime, calling commands using mutt's eventloop. To achieve that, the Lua API is a good way to go #298.

So I've done exactly that: https://gist.github.com/35be5c581aaabf6f1c453cbc2a6e6147

A few tests that are ran by calling:

```
build/mutt -F tests/functional/test_runner.muttrc --batch
```

Those tests will *not* be included in the Lua PR #298.

## unit tests

those ones are a bit trickier. The idea is to to access each header of mutt and test the API exposed through the compiled objects. I've achieved this by using [Luajit]+[FFI], and compiling mutt as a big shared library. To complete the testing toolchain (and load all the headers and symbols), I've stolen a bunch of lua files from neovim/neovim 😀 like [helpers.lua].

And then I wrote a test case for the lists (babysteps!): https://gist.github.com/b12b948616341d8dffd825ae9a5efcea

[FFI]:https://en.wikipedia.org/wiki/Foreign_function_interface
[Luajit]:http://luajit.org
[helpers.lua]:https://github.com/neovim/neovim/blob/master/test/unit/helpers.lua

gahr:
While I think using Lua for functional tests could be a fair solution and something worth investigating, I object to using any foreign language + FFI for **unit** testing. There are a few reasons for that.

- unit tests are often the best documentation available on a particular function or data structure. In contrast to comments-based documentation (read: doxygen)- which we all know tends to rot away - unit tests are forced to be kept in sync with the functionality the test when this evolves. Having them in a different language prevents them from offering concrete examples on how to use or not use a particular API;
- having unit tests in C would allow us to at least have a discussion about where the unit tests code should go, either in separate files or near the code they test. Using a foreign language would preclude even the possibility to have such a discussion;
- unit tests in C would allow for coverage analysis. I understand not everybody is enthusiastic about coverage analysis and the level of confidence it gives about the goodness of code. However, what I don't think is arguable is that they provide a nice way to figure out what is fairly tested and what needs deeper inspection. As far as I know, FFI can't give you that;
- unit tests in C would allow for a much deeper introspection of data structures and would forcibly be much more integrated with C types system (think of pointers);
- unit tests in C would free us from having to maintain code to generate FFI code (...) and data structures;

[...]

PR #298, opened 2017-01-15
Lua Scripting feature

guyzmo:
Offers a basic API that's working and have some tests.

cf #153 for comments

PR #415, opened 2017-02-19
Functional and unit testing framework

guyzmo:
Adds the necessary for both unit testing and functional testing.
Unit testing will require building the code towards a shared library
Functional testing works using the Lua feature cf #298

For the CI script to work, you need to give:

* `${SOURCES}` the path to the checked out sources
* `${BUILD}` the path to the build directory

`build.sh` will build, and patch the makefiles at the `${BUILD}` target directory.
And `run_tests.sh` will run both tests.

cf #414

### DIFF
--- a/.ci/common.sh
+++ b/.ci/common.sh
@@ -1,0 +1,257 @@
+#!/bin/sh
+
+set -o nounset
+
+# Makes sure SOURCE_DIR path is absolute and relative to current file
+if [ ! -d ${SOURCE_DIR:-} ]; then
+  SOURCE_DIR=$(readlink -f -- "$(dirname $0/..)")
+fi
+
+# show a nice message with a coloured unicode char to say it worked
+function OK() {
+  echo -e "  [\033[32m✓\033[0m]" $*
+}
+
+# show a nice message with a coloured unicode char to say it failed
+function KO() {
+  echo -e "  [\033[31m✗\033[0m]" $*
+}
+
+# helper function to exit upon error with a nice message,
+# or show a nice message upon success
+function check_ret() {
+  if [ $? -ne 0 ]; then
+    KO $2
+    exit 1
+  else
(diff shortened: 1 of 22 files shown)


guyzmo:
so, the next step will be to work out automatic installation of the dependencies:

* LuaJit 2.0 for the unit tests, and the busted framework for LuaJit (or Lua 5.1)
* Lua ≥5.2 for the functional tests and compilation of mutt (should be there), and the busted framework for Lua ≥5.2.

toogley:
As far as i understand, unit tests should not rely on dependencies (by that i mean typical things like databases, external servers, internet connection, filesystem, etc. pp.) (I have read this in a book called ["The art of unit testing"](https://www.manning.com/books/the-art-of-unit-testing-second-edition?a_bid=f6ea80f5&a_aid=iserializable)). 

How would you solve this issue with your minimal unit testing framework? That means, how would you mock external dependencies, such as imap server, notmuch database, maildir, etc. ?

(I'm not sure if the answer to this question is different to @gahr 's [approach](https://github.com/neomutt/neomutt/issues/428), therefore i ask, just to be sure.)

guyzmo:
> As far as i understand, unit tests should not rely on dependencies (by that i mean typical things like databases, external servers, internet connection, filesystem, etc. pp.) (I have read this in a book called "The art of unit testing").
> How would you solve this issue with your minimal unit testing framework? That means, how would you mock external dependencies, such as imap server, notmuch database, maildir, etc. ?

cf my other post: the whole point of using a dynamic language is to make it easier to write mockups for all the dependency. What's important with unit testing is not to rely on no dependencies, but not use the dependencies and the runtime environment the program runs in. You should consider that like petridish experiments, where you use tooling that's specially crafted for the task of testing.

flatcap:
This branch has been archived to: https://github.com/neomutt/neomutt-old/tree/testing/testing

PR #417, opened 2017-02-20
notmuch: Allow to use untransformed tag for color

sileht:
Hi,

This change allow to use untransformed tag for index-tag color.

Cheers,

### DIFF
--- a/menu.c
+++ b/menu.c
@@ -52,6 +52,9 @@ get_color (int index, unsigned char *s)
     case MT_COLOR_INDEX_TAG:
       for (color = ColorIndexTagList; color; color = color->next)
       {
+        if (strncmp((const char *)(s+1), color->pattern,
+                    strlen(color->pattern)) == 0)
+          return color->pair;
         const char * transform = hash_find(TagTransforms, color->pattern);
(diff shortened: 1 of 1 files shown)


flatcap:
Hmm... there are too many `(`s

flatcap:
@neomutt/reviewers Please Review

menu.c
```diff
@@ -52,6 +52,9 @@ get_color (int index, unsigned char *s)
     case MT_COLOR_INDEX_TAG:
       for (color = ColorIndexTagList; color; color = color->next)
       {
+        if (strncmp((const char *)(s+1), color->pattern,
+                    strlen(color->pattern)) == 0)
+          return color->pair;
         const char * transform = hash_find(TagTransforms, color->pattern);
```
gahr: (perhaps my previous comment went missing)

Are you sure that `strncmp` is correct? Should "red" match "redish"?

flatcap: That's what I thought, but the pattern is being compared to the line which is "TAG" followed by a binary colour code.

gahr: Alright, thanks for the explanation!

PR #419, opened 2017-02-21
Remove ugly macros and casts from crypt-gpgme.c

gahr:


gahr:

crypt-gpgme.c
```diff
@@ -3162,12 +3147,12 @@ print_dn_parts (FILE *fp, struct dn_array_s *dn)
 
 
 /* Parse an RDN; this is a helper to parse_dn(). */
-static const unsigned char *
-parse_dn_part (struct dn_array_s *array, const unsigned char *string)
+static const char *
+parse_dn_part (struct dn_array_s *array, const char *string)
 {
-  const unsigned char *s, *s1;
+  const char *s, *s1;
```
toogley: why did you remove the `unsigned` here?

crypt-gpgme.c
```diff
@@ -3162,12 +3147,12 @@ print_dn_parts (FILE *fp, struct dn_array_s *dn)
...
   size_t n;
-  unsigned char *p;
+  char *p;
```
toogley: same: why did you remove the unsigned here?

crypt-gpgme.c
gahr: because they are not needed :)

toogley: @gahr Hm, for me its not obvious that they are not needed, can you explain it to me?^^

(btw. This wasn't meant as a review, i was just curious of your change -> I'm trying to understand mutt better :D)

gahr: @toogley for me it is not obvious that they were needed in the first place.. anyway, I *think* they were put there because some plain arithmetic was done with the characters, especially in the `xtoi_(1|2)` macros, so treating them as unsigned perhaps allows the original author to reason in a simpler domain (that of unsigned integers).

In general, if you find yourself having to cast a `char *` into an `unsigned char *`, you better stop and ask yourself *why* you're doing that. There are rare occasions where it is legitimate, but in most cases I'd say it's a symptom of poor understanding of how integer arithmetic works in C.

toogley: @gahr thanks :)

PR #420, opened 2017-02-21
Use getaddrinfo_a if possible

### LINKED ISSUE #176: blocks on start if the network isn't responding
The place I currently work at has a flakey wifi - from time to time, it will just stop working and things will just time out, even DNS requests.

Opening mutt when I have no internet connection is immediate and works (I use mbsync/isync, so it uses local maildirs). When the internet connection is available and works normally, mutt also opens up immediately.

But when there is an internet connection and it's timing out, opening mutt blocks for a long time (often more than 10s) showing nothing. It eventually gives up on whatever it's trying to do and finally shows me my inbox.

I have no idea what it is it's trying to do there, I would guess it's a DNS query since quite a lot of programs out there do those synchronously instead of asynchronously. I had that exact problem with rtorrent, which is why I had to end up dropping it - DNS timeouts made the whole program freeze.

If more information is needed, could anyone tell me how to debug what it's blocking on next time it blocks like this?

sileht:
When getaddrinfo_a is avialable we can use it to set a timeout for the
dns resolution. This avoid to have mutt that freeze on startup when the
network connection isn't working and the hosts/hostname misconfigured on
the system.

Closes #176

### DIFF
--- a/getdomain.c
+++ b/getdomain.c
@@ -33,40 +33,54 @@ int getdnsdomainname (char *d, size_t len)
 {
   int ret = -1;
 
-#ifdef HAVE_GETADDRINFO
-  char *node;
-  long node_len;
+#if defined HAVE_GETADDRINFO || defined HAVE_GETADDRINFO_A
+  char node[STRING];
+  if (gethostname(node, sizeof(node)))
+    return ret;
+
   struct addrinfo hints;
-  struct addrinfo *h;
-  char *p;
+  struct addrinfo *h = NULL;
 
   *d = '\0';
   memset(&hints, 0, sizeof (struct addrinfo));
   hints.ai_flags = AI_CANONNAME;
   hints.ai_family = AF_UNSPEC;
 
-  /* A DNS name can actually be only 253 octets, string is 256 */
-  if ((node_len = sysconf(_SC_HOST_NAME_MAX)) == -1)
-    node_len = STRING;
-  node = safe_malloc(node_len + 1);
-  if (gethostname(node, node_len))
-    ret = -1;
-  else if (getaddrinfo(node, NULL, &hints, &h))
-    ret = -1;
-  else
+#ifdef HAVE_GETADDRINFO_A
+
+  /* Allow 0.1 seconds to get the FQDN (fully-qualified domain name).
+   * If it takes longer, the system is mis-configured and the network is not
+   * working properly, so...
+   */
+  struct timespec timeout = {0, 100000000};
+  struct gaicb *reqs[1];
+  reqs[0] = safe_calloc(1, sizeof(*reqs[0]));
+  reqs[0]->ar_name = node;
+  reqs[0]->ar_request = &hints;
+  if ((getaddrinfo_a(GAI_NOWAIT, reqs, 1, NULL) == 0) &&
+      (gai_suspend((const struct gaicb * const *) reqs, 1, &timeout) == 0) &&
+      (gai_error(reqs[0]) == 0))
   {
-    if (!h->ai_canonname || !(p = strchr(h->ai_canonname, '.')))
-      ret = -1;
-    else
-    {
-      strfcpy(d, ++p, len);
-      ret = 0;
-      mutt_debug (1, "getdnsdomainname(): %s\n", d);
-    }
+    h = reqs[0]->ar_result;
+  }
+  FREE(&reqs[0]);
+
+#else /* !HAVE_GETADDRINFO_A */
+
+  getaddrinfo(node, NULL, &hints, &h)
+
+#endif
+
+  char *p;
+  if (h != NULL && h->ai_canonname && (p = strchr(h->ai_canonname, '.')))
+  {
+    strfcpy(d, ++p, len);
+    ret = 0;
+    mutt_debug (1, "getdnsdomainname(): %s\n", d);
     freeaddrinfo(h);
   }
-  FREE (&node);
-#endif
+
+#endif /* HAVE_GETADDRINFO || defined HAVE_GETADDRINFO_A */
 
   return ret;
 }
--- a/configure.ac
+++ b/configure.ac
@@ -658,11 +658,12 @@ need_socket="no"
 
 dnl -- socket dependencies --
 
-dnl getaddrinfo is used by getdomain.c, and requires libnsl and
-dnl libsocket on some platforms
+dnl getaddrinfo/getaddrinfo_a is used by getdomain.c, and requires libnsl and
+dnl libsocket on some platforms, and libanl
 AC_CHECK_FUNC(gethostent, , AC_CHECK_LIB(nsl, gethostent))
 AC_CHECK_FUNC(setsockopt, , AC_CHECK_LIB(socket, setsockopt))
 AC_CHECK_FUNCS(getaddrinfo)
+AC_SEARCH_LIBS([getaddrinfo_a], [anl], [AC_DEFINE(HAVE_GETADDRINFO_A, 1, [Define to 1 if you have the `getaddrinfo_a' function.])])
 
 AS_IF([test x$use_pop = "xyes"], [
 	AC_DEFINE(USE_POP, 1, [Define if you want support for the POP3 protocol.])


flatcap:

sileht:
I have fixed the ifdef with none are present, I don't really know how I can make it more readable without duplicate the code.

gahr:
A couple of suggestions:
- make node an array, say, STRING bytes (no need for sysconf and free at the end)
- turn if (ok) { many lines } into if (nok) { return -1; }
- free reqs[0] immeditely after assigning h

gahr:
how about his:

```c
int getdnsdomainname (char *d, size_t len)                                      
{                                                                               
  int ret = -1;                                                                 
                                                                                
#if defined HAVE_GETADDRINFO || defined HAVE_GETADDRINFO_A                      
                                                                                
  char node[STRING];                                                            
  if (gethostname(node, sizeof(node)))                                          
  {                                                                             
    return ret;                                                                 
  }                                                                             
                                                                                
  struct addrinfo hints;                                                        
  struct addrinfo *h = NULL;                                                    
                                                                                
  *d = '\0';                                                                    
  memset(&hints, 0, sizeof (struct addrinfo));                                  
  hints.ai_flags = AI_CANONNAME;                                                
  hints.ai_family = AF_UNSPEC;                                                  
                                                                                
#ifdef HAVE_GETADDRINFO_A                                                       
                                                                                
  /* 0.1 seconds, do we really need more just to get the fqdn                   
   * If it takes longer, the system is miss configured and the                  
   * network is not working properly, so...                                     
   **/                                                                          
  struct timespec timeout = { 0, 100000000};                                    
  struct gaicb *reqs[1];                                                        
  reqs[0] = safe_calloc(1, sizeof(*reqs[0]));                                   
  reqs[0]->ar_name = node;                                                      
  reqs[0]->ar_request = &hints;                                                 
  if (getaddrinfo_a(GAI_NOWAIT, reqs, 1, NULL) == 0 && \                        
      gai_suspend((const struct gaicb * const *) reqs, 1, &timeout) == 0 && \   
      gai_error(reqs[0]) == 0)                                                  
  {                                                                             
    h = reqs[0]->ar_result;                                                     
  }                                                                             
  free(reqs[0]);                                                                
                                                                                
#else /* !HAVE_GETADDRINFO_A */                                                 
                                                                                
  getaddrinfo(node, NULL, &hints, &h);                                          
                                                                                
#endif /* HAVE_GETADDRINFO_A */                                                 
                                                                                
  char *p;                                                                      
  if (h != NULL && h->ai_canonname && (p = strchr(h->ai_canonname, '.')))       
  {                                                                             
    strfcpy(d, ++p, len);                                                       
    ret = 0;                                                                    
    mutt_debug (1, "getdnsdomainname(): %s\n", d);                              
    freeaddrinfo(h);                                                            
  }                                                                             
                                                                                
#endif /* HAVE_GETADDRINFO || HAVE_GETADDRINFO_A */                             
                                                                                
  return ret;                                                                   
}        
```

sileht:
Thanks for the suggestion, removing FREE(&node), simplify a lot the
flow :)

On Wed, Feb 22, 2017 at 12:28:31AM -0800, Pietro Cerutti wrote:
>how about his:
>
>```c
>int getdnsdomainname (char *d, size_t len)
>{
>  int ret = -1;
>
>#if defined HAVE_GETADDRINFO || defined HAVE_GETADDRINFO_A
>
>  char node[STRING];
>  if (gethostname(node, sizeof(node)))
>  {
>    return ret;
>  }
>
>  struct addrinfo hints;
>  struct addrinfo *h = NULL;
>
>  *d = '\0';
>  memset(&hints, 0, sizeof (struct addrinfo));
>  hints.ai_flags = AI_CANONNAME;
>  hints.ai_family = AF_UNSPEC;
>
>#ifdef HAVE_GETADDRINFO_A
>
>  /* 0.1 seconds, do we really need more just to get the fqdn
>   * If it takes longer, the system is miss configured and the
>   * network is not working properly, so...
>   **/
>  struct timespec timeout = { 0, 100000000};
>  struct gaicb *reqs[1];
>  reqs[0] = safe_calloc(1, sizeof(*reqs[0]));
>  reqs[0]->ar_name = node;
>  reqs[0]->ar_request = &hints;
>  if (getaddrinfo_a(GAI_NOWAIT, reqs, 1, NULL) == 0 && \
>      gai_suspend((const struct gaicb * const *) reqs, 1, &timeout) == 0 && \
>      gai_error(reqs[0]) == 0)
>  {
>    h = reqs[0]->ar_result;
>  }
>  free(reqs[0]);
>
>#else /* !HAVE_GETADDRINFO_A */
>
>  getaddrinfo(node, NULL, &hints, &h);
>
>#endif /* HAVE_GETADDRINFO_A */
>
>  char *p;
>  if (h != NULL && h->ai_canonname && (p = strchr(h->ai_canonname, '.')))
>  {
>    strfcpy(d, ++p, len);
>    ret = 0;
>    mutt_debug (1, "getdnsdomainname(): %s\n", d);
>    freeaddrinfo(h);
>  }
>
>#endif /* HAVE_GETADDRINFO || HAVE_GETADDRINFO_A */
>
>  return ret;
>}
>```
>
>-- 
>You are receiving this because you were assigned.
>https://github.com/neomutt/neomutt/pull/420#issuecomment-281601965

-- 
Mehdi Abaakouk
mail: sileht@sileht.net
irc: sileht

flatcap:
I've done a little tidying:

Make the comments in `config.h` match
```diff
-AC_SEARCH_LIBS([getaddrinfo_a], [anl], [AC_DEFINE(HAVE_GETADDRINFO_A, 1, [Define this symbol if you have getaddrinfo_a])])
+AC_SEARCH_LIBS([getaddrinfo_a], [anl], [AC_DEFINE(HAVE_GETADDRINFO_A, 1, [Define to 1 if you have the `getaddrinfo_a' function.])])
```

- Remove unnecessary `\`s
- Add `()`s to logic
```diff
-  if (getaddrinfo_a(GAI_NOWAIT, reqs, 1, NULL) == 0 && \
-      gai_suspend((const struct gaicb * const *) reqs, 1, &timeout) == 0 && \
-      gai_error(reqs[0]) == 0)
+  if ((getaddrinfo_a(GAI_NOWAIT, reqs, 1, NULL) == 0) &&
+      (gai_suspend((const struct gaicb * const *) reqs, 1, &timeout) == 0) &&
+      (gai_error(reqs[0]) == 0))
```

Use Mutt's "free and NULL" function
```diff
-  free(reqs[0]);
+  FREE(&reqs[0]);
```

I'll fold the changes into your commit.

sthen:
The version of this which was committed misses the ; from the getaddrinfo() call in the "#else /* !HAVE_GETADDRINFO_A */" case, causing compilation to fail.

getdomain.c
```diff
@@ -33,41 +33,55 @@ int getdnsdomainname (char *d, size_t len)
...
-    }
-    freeaddrinfo(h);
+    h = reqs[0]->ar_result;
   }
-  FREE (&node);
+  FREE(&reqs[0]);
+
+#else /* !HAVE_GETADDRINFO_A */
+
+  getaddrinfo(node, NULL, &hints, &h)
```
guyzmo: cf last comment

flatcap:
@sthen Thanks for spotting that.  Fixed.
That's what comes of late-night releases.  Sorry.

sileht:
The debian approach to do less dns lookup: https://anonscm.debian.org/cgit/pkg-mutt/mutt.git/tree/debian/patches/debian-specific/Md.etc_mailname_gethostbyname.patch

flatcap:
Ah, yes.  That's also in [[distro/debian](https://github.com/neomutt/neomutt/tree/distro/debian)] -- 1df0ae0

Is `/etc/mailname` used anywhere but Debian? -- it doesn't seem to be part of any standard
Is it worth adding anyway? -- it's quick to check

sileht:
On Wed, Mar 01, 2017 at 04:18:45AM -0800, Richard Russon wrote:
>Ah, yes.  That's also in [[distro/debian](https://github.com/neomutt/neomutt/tree/distro/debian)] -- 1df0ae0
>
>Is `/etc/mailname` used anywhere but Debian? -- it doesn't seem to be part of any standard
>Is it worth adding anyway? -- it's quick to check

It's a Debian thing, Red Hat uses HOSTNAME=foo.bar in
/etc/sysconfig/network. And I don't really know for *BSD, but I think
you have to do the DNS lookup.

gahr:
FreeBSD has `hostname = foo` in `/etc/rc.conf`. Anyway, it should be libc's problem to get it from wherever the system puts it.. we should rely on much more than `gethostname`..

flatcap:
Fair enough.  We'll leave it as it is, for now.

PR #431, opened 2017-02-25
Unit test backend using BDD-For-C (only)

guyzmo:
- implemented the simple list test using that backend
- updated test runners

TODO:

many parts of this PR is done in shell scripts and should be done in autoconf crap. M4 just creeps me out, so if someone can help me out on that. Basically what would be needed is:

- add a `mutt.so` target in the Makefile that compiles everything with `-fPIC` and generate a `-shared`
- add a `test` target and a `Makefile.am` in `tests/units` that compiles `*_spec.c` file into an executable.


flatcap:
This branch has been archived to: https://github.com/neomutt/neomutt-old/tree/testing/testing-bdd-for-c

PR #437, opened 2017-02-26
small refactoring: replacing macros with functions, readability changes.

toogley:
* replace ESC_CHAR macro with function

* replace CHECK_PAGER macro with function
return 0 was added to prevent warning `control reaches end of non-void function
[-Wreturn-type]`

* added braces to if logic to improve readability

### DIFF
--- a/commands.c
+++ b/commands.c
@@ -1047,7 +1047,9 @@ int mutt_check_traditional_pgp(struct Header *h, int *redraw)
     for (i = 0; i < Context->vcount; i++)
       if (Context->hdrs[Context->v2r[i]]->tagged &&
           !(Context->hdrs[Context->v2r[i]]->security & PGP_TRADITIONAL_CHECKED))
+      {
         rv = _mutt_check_traditional_pgp(Context->hdrs[Context->v2r[i]], redraw) || rv;
+      }
   }
   return rv;
 }
--- a/init.c
+++ b/init.c
@@ -69,12 +69,18 @@
 #include "mutt_notmuch.h"
 #endif
 
-#define CHECK_PAGER                                                                  \
-  if ((CurrentMenu == MENU_PAGER) && (idx >= 0) && (MuttVars[idx].flags & R_RESORT)) \
-  {                                                                                  \
-    snprintf(err->data, err->dsize, _("Not available in this menu."));               \
-    return -1;                                                                       \
+static int check_pager(int idx, struct Buffer *err)
+{
+  if ((CurrentMenu == MENU_PAGER) && (idx >= 0) && (MuttVars[idx].flags & R_RESORT))
+  {
+    snprintf(err->data, err->dsize, _("Not available in this menu."));
+    return (-1);
   }
+  else
+  {
+    return 0;
+  }
+}
 
 struct MyVar
 {
@@ -2451,7 +2457,7 @@ static int parse_set(struct Buffer *tmp, struct Buffer *s, unsigned long data,
       }
       else
       {
-        CHECK_PAGER;
+        check_pager(idx, err);
         if (myvar)
           myvar_del(myvar);
         else
@@ -2489,7 +2495,7 @@ static int parse_set(struct Buffer *tmp, struct Buffer *s, unsigned long data,
         return 0;
       }
 
-      CHECK_PAGER;
+      check_pager(idx, err);
       if (unset)
         unset_option(MuttVars[idx].data);
       else if (inv)
@@ -2503,7 +2509,7 @@ static int parse_set(struct Buffer *tmp, struct Buffer *s, unsigned long data,
     {
       if (unset)
       {
-        CHECK_PAGER;
+        check_pager(idx, err);
         if (myvar)
           myvar_del(myvar);
         else if (DTYPE(MuttVars[idx].type) == DT_ADDR)
@@ -2561,7 +2567,7 @@ static int parse_set(struct Buffer *tmp, struct Buffer *s, unsigned long data,
       }
       else
       {
-        CHECK_PAGER;
+        check_pager(idx, err);
         s->dptr++;
 
         if (myvar)
@@ -2661,7 +2667,7 @@ static int parse_set(struct Buffer *tmp, struct Buffer *s, unsigned long data,
         break;
       }
 
-      CHECK_PAGER;
+      check_pager(idx, err);
       s->dptr++;
 
       /* copy the value of the string */
@@ -2715,7 +2721,7 @@ static int parse_set(struct Buffer *tmp, struct Buffer *s, unsigned long data,
         break;
       }
 
-      CHECK_PAGER;
+      check_pager(idx, err);
       s->dptr++;
 
       /* copy the value of the string */
@@ -2745,7 +2751,7 @@ static int parse_set(struct Buffer *tmp, struct Buffer *s, unsigned long data,
         break;
       }
 
-      CHECK_PAGER;
+      check_pager(idx, err);
       s->dptr++;
 
       mutt_extract_token(tmp, s, 0);
@@ -2815,7 +2821,7 @@ static int parse_set(struct Buffer *tmp, struct Buffer *s, unsigned long data,
         break;
       }
 
-      CHECK_PAGER;
+      check_pager(idx, err);
       if (*s->dptr == '=')
       {
         s->dptr++;
@@ -2888,7 +2894,7 @@ static int parse_set(struct Buffer *tmp, struct Buffer *s, unsigned long data,
                  (*((short *) MuttVars[idx].data) & SORT_LAST) ? "last-" : "", p);
         return 0;
       }
-      CHECK_PAGER;
+      check_pager(idx, err);
       s->dptr++;
       mutt_extract_token(tmp, s, 0);
 
@@ -2908,7 +2914,7 @@ static int parse_set(struct Buffer *tmp, struct Buffer *s, unsigned long data,
         break;
       }
 
-      CHECK_PAGER;
+      check_pager(idx, err);
       s->dptr++;
 
       /* copy the value of the string */
--- a/init.c
+++ b/init.c
@@ -2066,6 +2066,13 @@ static void restore_default(struct Option *p)
     mutt_set_current_menu_redraw_full();
 }
 
+static void ESC_char(char c, char *p, char *dst, size_t len)
+{
+  *p++ = '\\';
+  if (p - dst < len)
+    *p++ = c;
+}
+
 static size_t escape_string(char *dst, size_t len, const char *src)
 {
   char *p = dst;
@@ -2073,25 +2080,18 @@ static size_t escape_string(char *dst, size_t len, const char *src)
   if (!len)
     return 0;
   len--; /* save room for \0 */
-#define ESC_CHAR(C)                                                            \
-  do                                                                           \
-  {                                                                            \
-    *p++ = '\\';                                                               \
-    if (p - dst < len)                                                         \
-      *p++ = C;                                                                \
-  } while (0)
   while (p - dst < len && src && *src)
   {
     switch (*src)
     {
       case '\n':
-        ESC_CHAR('n');
+        ESC_char('n', p, dst, len);
         break;
       case '\r':
-        ESC_CHAR('r');
+        ESC_char('r', p, dst, len);
         break;
       case '\t':
-        ESC_CHAR('t');
+        ESC_char('t', p, dst, len);
         break;
       default:
         if ((*src == '\\' || *src == '"') && p - dst < len - 1)
@@ -2100,7 +2100,6 @@ static size_t escape_string(char *dst, size_t len, const char *src)
     }
     src++;
   }
-#undef ESC_CHAR
   *p = '\0';
   return p - dst;
 }


commands.c
```diff
@@ -1068,10 +1068,12 @@ int mutt_check_traditional_pgp (HEADER *h, int *redraw)
   else
   {
     for (i = 0; i < Context->vcount; i++)
-      if (Context->hdrs[Context->v2r[i]]->tagged && 
-	  !(Context->hdrs[Context->v2r[i]]->security & PGP_TRADITIONAL_CHECKED))
-	rv = _mutt_check_traditional_pgp (Context->hdrs[Context->v2r[i]], redraw)
-	  || rv;
+      if ((Context->hdrs[Context->v2r[i]]->tagged) && 
+           (Context->hdrs[Context->v2r[i]]->security & PGP_TRADITIONAL_CHECKED))
+      {
```
gahr: Where have the negation gone?

init.c
```diff
@@ -55,14 +55,18 @@
...
+  if ((CurrentMenu == MENU_PAGER) && (idx >= 0) && (MuttVars[idx].flags & R_RESORT))
+  {
+    snprintf(err->data, err->dsize, _("Not available in this menu."));
+    return (-1);
+  }
+  else
+  {
+    return 0;
+  }
+}
```
gahr: Invoking check_pager as a function no longer makes the caller return. At least you'll have to turn all the calls below into

If (check_pager(...) == -1)
   return -1;

I'm not sure there's any advantage in doing so, though...

init.c
```diff
@@ -1905,26 +1909,35 @@ static void mutt_restore_default (struct option_t *p)
...
+{
+  do
+  {
+     *p++ = '\\';
+     if (p - dst < len)
+       *p++ = C;
+  } while(0);
+}
+
 static size_t escape_string (char *dst, size_t len, const char* src)
```
gahr: The whole point of the do { ... } while(0); construct is to turn the body of the macro into a block. By replacing the macro with a function this point fails to exist.

commands.c
toogley: Ah, sorry, that's an fault.

init.c
toogley: what exactly do you mean by "caller" ? do you mean with that the surrounding function that calls somewhere in its body check_pager() ?

toogley:
@gahr I thought that macros in general were worse than functions, because static analyzer's can't find as many errors as with functions.

Furthermoore:

 I do understand now the reason the `ESC_CHAR` macro exists.  
However, I don't understand the difference between these macros [in this PR](https://github.com/neomutt/neomutt/pull/419/files) and the `CHECK_PAGER` macro. Why do you consider `CHECK_PAGER` as more readable as one of the helper macros in #419 ?

gahr:
Those macros were replaced with a single call to a standard function:
hexdigitp -> isxdigit
hextoi_2 -> sscanf

These macros here are just a couple of statements that are needed in several places. I don't see much gain in building functions around them, and I'm not sure the resulting code will be as readable, but fine if you're absolutely positive about it - if done correctly of course.

init.c
gahr: You moved the return -1 one frame down, so it's a different function that's returning now. Before it was whatever function the macro was put in, now it's your new function itself.

toogley:
@gahr Ah, i see. I had overlooked that. Hm, at least i personally like the  `check_pager()` function slightly better - but i'm not sure how others think about that.

gahr:
Turning CHECK_PAGER into a function might be okay, provided it's given a name that actually makes sense and reflects what it does... can you think of anything?

toogley:
@gahr As far as i understand, the  `CHECK_PAGER` macro checks whether the given option (`MuttVars[idx].flags`) is available in the current menu (which is the pager). 

* But i don't understand why `MuttVars[idx].flags` is & connected with `R_RESORT`.  I mean, i understand that this is just a shortcut for testing both variables if they're set. But i don't understand why its important here to check if the mailbox needs to be resorted - because that contradicts IMHO the error message `"Not available in this menu."`.

* I also don't understand, why those Options/macros below are defined that way. I mean, why is bitshifting used here? Why is this even a macro and not just a simple variable?

```
#define R_INDEX		(1<<0)
#define R_PAGER		(1<<1)
#define R_RESORT	(1<<2)	/* resort the mailbox */
#define R_RESORT_SUB	(1<<3)	/* resort subthreads */
#define R_RESORT_INIT	(1<<4)  /* resort from scratch */
#define R_TREE		(1<<5)  /* redraw the thread tree */
#define R_REFLOW        (1<<6)  /* reflow window layout */
#define R_SIDEBAR       (1<<7)  /* redraw the sidebar */
```

init.c
guyzmo: I have to say, looking at the code, I prefer the function call solution.

We could inline those functions, but I believe there won't be any significant gain by doing so, so I'm ok with current prototype qualitfiers.

Otherwise I'd change the semantics a bit of that function, to have it use a boolean:

```
static bool check_pager(int idx, BUFFER *err)
{
    if (A)
    {
        return false;
    }
    else
    {
        return true;
    }
```

And then do in the usage places below:

```
if (!check_pager(idx, err))
    return -1;
```

guyzmo: Otherwise I'd change the semantics a bit of that function, to have it use a boolean:

```
static bool check_pager(int idx, BUFFER *err)
{
    if (A)
    {
        return false;
    }
    else
    {
        return true;
    }
```

And then do in the usage places below:

```
if (!check_pager(idx, err))
    return -1;
```

init.c
guyzmo: I follow: `do { } while (0);` shall go away for reasons @gahr states.

commands.c
```diff
@@ -1068,10 +1068,12 @@ int mutt_check_traditional_pgp (HEADER *h, int *redraw)
...
+        rv = _mutt_check_traditional_pgp (Context->hdrs[Context->v2r[i]],redraw)
+                  || rv;
+      }
   }
```
guyzmo: why did you do the change above? it's unrelated to the other parts of that commit. I think it should be either another commit or another PR for cosmetics only.

init.c
```diff
@@ -1905,26 +1909,35 @@ static void mutt_restore_default (struct option_t *p)
 #endif
 }
 
+static void ESC_char(char C, char* p, char *dst, size_t len)
```
guyzmo: change the name of the function to all lower case, maybe `esc_char(…)` or `escape_char(…)` or `escaped_char(…)`

init.c
```diff
@@ -2200,7 +2212,7 @@ static int parse_set (BUFFER *tmp, BUFFER *s, unsigned long data, BUFFER *err)
       }
       else
       {
-	CHECK_PAGER;
+        check_pager(idx,err);
```
guyzmo: cf my comment above:

```
if (!check_pager(idx, err))
    return -1;
```

also, space after comma.

init.c
```diff
@@ -2237,7 +2249,7 @@ static int parse_set (BUFFER *tmp, BUFFER *s, unsigned long data, BUFFER *err)
 	return 0;
       }
 
-      CHECK_PAGER;
+      check_pager(idx,err);
```
guyzmo: idem

init.c
```diff
@@ -2252,7 +2264,7 @@ static int parse_set (BUFFER *tmp, BUFFER *s, unsigned long data, BUFFER *err)
     {
       if (unset)
       {
-	CHECK_PAGER;
+        check_pager(idx,err);
```
guyzmo: idem

init.c
```diff
@@ -2309,7 +2321,7 @@ static int parse_set (BUFFER *tmp, BUFFER *s, unsigned long data, BUFFER *err)
       }
       else
       {
-	CHECK_PAGER;
+        check_pager(idx,err);
```
guyzmo: idem

init.c
```diff
@@ -2394,7 +2406,7 @@ static int parse_set (BUFFER *tmp, BUFFER *s, unsigned long data, BUFFER *err)
 	break;
       }
       
-      CHECK_PAGER;
+      check_pager(idx,err);
```
guyzmo: idem

init.c
```diff
@@ -2487,7 +2499,7 @@ static int parse_set (BUFFER *tmp, BUFFER *s, unsigned long data, BUFFER *err)
 	break;
       }
 
-      CHECK_PAGER;
+      check_pager(idx,err);
```
guyzmo: idem

init.c
```diff
@@ -2517,7 +2529,7 @@ static int parse_set (BUFFER *tmp, BUFFER *s, unsigned long data, BUFFER *err)
...
```
guyzmo: idem

init.c
```diff
@@ -2571,7 +2583,7 @@ static int parse_set (BUFFER *tmp, BUFFER *s, unsigned long data, BUFFER *err)
...
```
guyzmo: idem

init.c
```diff
@@ -2645,7 +2657,7 @@ static int parse_set (BUFFER *tmp, BUFFER *s, unsigned long data, BUFFER *err)
 		  p);
 	return 0;
       }
-      CHECK_PAGER;
+      check_pager(idx,err);
```
guyzmo: idem

init.c
```diff
@@ -2665,7 +2677,7 @@ static int parse_set (BUFFER *tmp, BUFFER *s, unsigned long data, BUFFER *err)
 	break;
       }
 
-      CHECK_PAGER;
+      check_pager(idx,err);
```
guyzmo: bis repetita.

commands.c
toogley: on [our website](https://www.neomutt.org/dev/coding-style#braces) is a note about that lines being difficult to read.

guyzmo: well you can leave it for the clang-format commit.

toogley: Clang format doesn't include extra characters, like braces. It only puts whitespace into files^^

gahr:
Before proceeding with turning CHECK_PAGER into a function, I suggest to check wheter it is actually needed to begin with. Hint: I am running without and I don't see any problems.

guyzmo:
@gahr did you try doing different `set` commands while in pager mode?

gahr:
@guyzmo of course, changing sort and thread settings seems to work fine.
Again, I am not seeing any issues but I'm not 100% sure the macro is not needed. I just would like us to assess that, since we're into this part of the codebase.

guyzmo:
ok, I"ll also have a look

toogley:
1. i'll wait some time with fixing this PR, until @zmo and @gahr have notified any bugs because of the removal of those `CHECK_PAGER` macro calls.

2. I'd be still very happy if someone could explain me  [these questions](https://github.com/neomutt/neomutt/pull/437#issuecomment-282598559) :)

gahr:
@toogley, aren't you the one supposed to figure it out? There's your name on this pr after all ;) I'm just offering hints here..

toogley:
@gahr Well, maybe :D

  i've looked again at the bitshifting parts, and i think i understand it now. i just forgot how bitshifting actually looks like.

flatcap:
The second question (bit shifting) has two answers: compact, clear.
- writing all the constants as `(1<<16)`, `(1<<17)` is much tidier than `65536`, `131072` -- less risk of a typo
- `(1<<17)` shows you immediately that EXACTLY one bit is set, which tells you that the constants might be `|`'d together

toogley:
@flatcap thanks. i guess you don't know the answer the the first question?

flatcap:
> i guess you don't know the answer the the first question?

I've only glanced at the code, so not yet :-)

guyzmo:
> writing all the constants as (1<<16), (1<<17) is much tidier than 65536, 131072 -- less risk of a typo

though in the XXIst century, you can use `0b10000000` instead of `1<<7` which is also quite explicit ☺

>  i'll wait some time with fixing this PR, until @zmo and @gahr have notified any bugs because of the removal of those CHECK_PAGER macro calls.

it's a good exercise that's within your reach as well.

Try to produce the steps necessary to produce a call to CHECK_PAGER and have it `return -1`. Using a breakpoint could be a good way, using `printf()` printouts another. And then assess whether THERE is a path in the code to actually trigger the failing condition or not.

guyzmo:
> But i don't understand why MuttVars[idx].flags is & connected with R_RESORT. I mean, i understand that this is just a shortcut for testing both variables if they're set. But i don't understand why its important here to check if the mailbox needs to be resorted - because that contradicts IMHO the error message "Not available in this menu.".

it's the A-B-C of programming and boolean logic. flags is an `int`, `R_RESORT` another int with a single bit set. If you lookup on wikipedia AND boolean table, you'll see that:

```
0&1 → 0
1&1 → 1
```

what it means is that you can check the presence of a flag in a byte using:

```
flags = 0b11011010
flag_1 = 0b100
flag_2 = 0b10
(flag_1&flags) == 0
(flag_2&flags) == flag_2
```

and you can set it using the `OR` operation:

```
flags= 0b00000000;
flags= flags|flag_1
flags == 0b100
```

flatcap:
The `CHECK_PAGER` macro was introduced in commit dd913a9

> Ensure that set doesn't change variables from within the help screen. I think this is a gross approach, and probably overkill, but it's the fault of the original code design, not Tamo's patch. Until a better way falls from the sky, this Closes: [#2011](https://dev.mutt.org/trac/ticket/2011).

>Some variables can be changed from the help screen. (SEGV!)

toogley:
@flatcap thanks.

I cannot reproduce the behavior described in [the mutt ticket #2011](https://dev.mutt.org/trac/ticket/2011):

* updated current master
* I removed CHECK_PAGER completly from the sources (both the calls and its definiton)
* `./prepare --enable-everything --without-tokyocabinet --without-qdbm && make`
* neither `./mutt -RnF /dev/null -e 'push "?:set sort=reverse-threads\n"'` nor `./mutt -RnF /dev/null` and doing `?:set sort=reverse-threads\n` manually with pressing `q` afterwards, produces an segfault (i guess that's meant by "SEGV")

Therefore i'm also for the removal for this macro.

flatcap:
The pager and index are very inter-dependent.  They pass around large lists of pointers (to emails).
Anything that changes the order of one of these lists unexpectedly might cause severe problems.

guyzmo:
i'm suggesting to hold off this PR with the `CHECK_PAGER` thing, because as @gahr suggests in #298 I should refactor `parse_set` to use the new `mutt_option_set` and `mutt_option_get` API I've introduced, which will thoroughly break `parse_set`.

flatcap:
I've split this into two parts.
- ESC_char() is an improvement. merged.
- CHECK_PAGER() doesn't function correctly because the original macro would cause the calling **function** to return.  not merged.

As has been discussed, fixing CHECK_PAGER() would mean making a mess of the code, or some significant restructuring, so I'm going to close this PR for now.

PR #439, opened 2017-02-27
editorconfig: support for new files, fix whitespace

toogley:
* Our code contains a lot of whitespace issues, so every commit is
  polluted by whitespace changes with the current editorconfig. However, i'm not sure whether disabling `trim_trailing_whitespace` is the best way to go until the codebase has been formatted with clang-format (which will also fix trailing whitespace)

* Add support for Makefile.am, xml, vim and yml files.

@flatcap what do you think?

### DIFF
--- a/.editorconfig
+++ b/.editorconfig
@@ -4,18 +4,18 @@ root = true
 end_of_line = lf
 insert_final_newline = true
 charset = utf8
-trim_trailing_whitespace = true
+trim_trailing_whitespace = false
 
-[*.{c,h}]
+# those two settings may be overwritten below.
 indent_size = 2
 indent_style = space
 
+
+
 [ChangeLog.neomutt]
-indent_style = space
 indent_size = 1
 
 
-[configure]
-indent_size = 2
-indent_style = space
-
+[{[mM]akefile*,*.am}]
+indent_size = 4
+indent_style = tab


.editorconfig
```diff
@@ -4,9 +4,9 @@ root = true
 end_of_line = lf
 insert_final_newline = true
 charset = utf8
-trim_trailing_whitespace = true
+trim_trailing_whitespace = false
```
gahr: why?

toogley: @ghar [see my first bullet point](https://github.com/neomutt/neomutt/pull/439#issue-210464653). That's just a temporary fix, until the clang-format has run on our codebase - which is, according to @flatcap, to ready at the moment.

gahr: Sorry, I failed to read the bullets...

toogley: @gahr np :)

toogley:
@gahr are you also okay with this generic variant? note, that it specifies for every file in the neomutt repo (except for makeand *.am files) to use spaces based indendation.

gahr:
If it matches clang-format I'm fine with it 👍🏻

toogley:
i compared the options of editorconfig and clang-format, therefore it **shouldn't** interfere with each other. But i haven't found a sensible test strategy, and therefore i haven't tested very well.

Issue #440, opened 2017-02-27
Bug under fmemopen + torify


gahr:
See #361

I have not been able to reproduce the crash, but I do see an unexpected behaviour.

When running under torify with fmemopen enabled, closing a FILE opened with `open_memstream` does not automatically update the size pointer. This results in assuming the body length is zero.

gahr:
This only happens with relatively small messages and seems to be due to internal buffers not being automatically flushed if not filled. This also confirms that the fix is likely correct.

PR #441, opened 2017-02-27
Get the correct buffer size under fmemopen/torify

gahr:
#440

### DIFF
--- a/handler.c
+++ b/handler.c
@@ -1701,17 +1701,16 @@ static int run_decode_and_handler (BODY *b, STATE *s, handler_t handler, int pla
 
     mutt_decode_attachment (b, s);
 
-#ifdef USE_FMEMOPEN
-    /* When running under torify, safe_fclose(&s->fpout) does not seem to
-     * update tempsize. On the other hand, fflush does.  See
-     * https://github.com/neomutt/neomutt/issues/440 */
-    fflush (s->fpout);
-#endif
-
     if (decode)
     {
       b->length = ftello (s->fpout);
       b->offset = 0;
+#ifdef USE_FMEMOPEN
+      /* When running under torify, safe_fclose(&s->fpout) does not seem to
+       * update tempsize. On the other hand, fflush does.  See
+       * https://github.com/neomutt/neomutt/issues/440 */
+      fflush (s->fpout);
+#endif
       safe_fclose (&s->fpout);
 
       /* restore final destination and substitute the tempfile for input */
@@ -1757,8 +1756,7 @@ static int run_decode_and_handler (BODY *b, STATE *s, handler_t handler, int pla
       /* restore the original source stream */
       safe_fclose (&s->fpin);
 #ifdef USE_FMEMOPEN
-      if (tempsize)
-        FREE(&temp);
+      FREE(&temp);
 #endif
       s->fpin = fp;
     }
--- a/handler.c
+++ b/handler.c
@@ -1701,6 +1701,13 @@ static int run_decode_and_handler (BODY *b, STATE *s, handler_t handler, int pla
 
     mutt_decode_attachment (b, s);
 
+#ifdef USE_FMEMOPEN
+    /* When running under torify, safe_fclose(&s->fpout) does not seem to
+     * update tempsize. On the other hand, fflush does.  See
+     * https://github.com/neomutt/neomutt/issues/440 */
+    fflush (s->fpout);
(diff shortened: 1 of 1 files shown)


gahr:

gahr:
Fixed commit message length...

guyzmo:
I kinda follow @d-k-c on that one. Is it a fix to the general case, or a workaround for a specific use case?

If it's the second, as long you can guarantee there's no side effects in the general situation, I see no harm in adding such a patch. But maybe torify should be notified they break POSIX APIs.

flatcap:
I've stared at the code and the man page for a while now...

> I don't understand why it's not next to the safe_fclose line

There are two `fclose()`s in the function.  Having the `fflush()` there means that `tempsize` will get updated whichever path the code takes.

> Is the bug only reproducible with torify?

That's the only case we know of.

> I don't know why it would mess with this API in the first place

Torify clearly intercepts network traffic.
I wonder if it's messing with temporary files too, for security.

> a specific use case

probably

> as long you can guarantee there's no side effects

We're flushing a small temporary file shortly before closing it.
That shouldn't cause any problems.

d-k-c:
> There are two fclose()s in the function. Having the fflush() there means that tempsize will get updated whichever path the code takes.
We only care about `tempsize` when decode = 1.

And if I read the manuel for `open_memstream` correctly, a null byte is maintained at the end of the buffer, but it's not included in the size. So testing if the buffer should be freed with `if (tempsize)` seems like a potential memleak to me (which is what Valgrind confirms on my laptop)

TL;DR: only the first use of `tempsize` is necessary, the second one is an error; that means the fflush can be moved next to `safe_fclose (&s->fpout);`, which garanties that s->fpout is sanely defined.

gahr:
> @d-k-c  We only care about tempsize when decode = 1.

Yes, and all uses of `tempsize` are within `if (decode) { }` blocks.

> @dkc And if I read the manuel for open_memstream correctly, a null byte is maintained at the end of the buffer, but it's not included in the size. So testing if the buffer should be freed with if (tempsize) seems like a potential memleak to me (which is what Valgrind confirms on my laptop)

That is correct, there is indeed a memleak, but not for the reason you're mentioning. The buffer won't contain a null byte unless it's written to, which implies that tempsize > 0. Anyway, the buffer must be freed no matter what.

Also, I moved the `fflush` inside the `if (decode) {}` block.

gahr:
Here is some more evidence:

```c
/* memstream.c */
#include <assert.h>
#include <errno.h>
#include <stdbool.h>
#include <stdio.h>
#include <stdlib.h>
#include <string.h>

int main(int argc, char **argv)
{
    FILE *fp;
    char *buf;
    size_t len;
    bool flush;

    flush = (argc == 2 && *argv[1] == '1');

    fp = open_memstream(&buf, &len);
    assert(fp);
    fputs("Hello", fp);
    if (flush)
        fflush(fp);
    fclose(fp);
    printf("Flush : %d\n", flush);
    printf("Length: %zu\n", len);
    free(buf);
    return 0;
}
```

```sh
$ ./memstream 1
Flush : 1
Length: 5
$ ./memstream
Flush : 0
Length: 5
$ torsocks ./memstream 1
Flush : 1
Length: 5
$ torsocks ./memstream                                                                                                      
Flush : 0
Length: 0
Segmentation fault (core dumped)
```

flatcap:
One other thing (which clearly isn't causing a problem, but I'd like to see) is setting the variables used as out-params.
```c
 #ifdef USE_FMEMOPEN
   char *temp = NULL;
   size_t tempsize = 0;
#endif
```

d-k-c:
If you think this fix is worth including, then I'm quite happy with it as is.

Has this bug been reported to the Tor Project/torsocks/torify? A link to their bug tracker would be more useful than a link to ours.

gahr:
> Has this bug been reported to the Tor Project/torsocks/torify? A link to their bug tracker would be more useful than a link to ours.

Not yet, I'm trying to debug it - torsocks works by `LD_PRELOAD`ing libc stuff, so it's a bit messy.

gahr:
> One other thing (which clearly isn't causing a problem, but I'd like to see) is setting the variables used as out-params.

Done in eb2438a.

gahr:
OP reports (via private email) that NeoMutt still crashes with this patch. I'll have to investigate further. Please do not merge just yet.

Issue #377, opened 2017-02-07
Creating buffer.c


flatcap:
This task is more investigation, than coding (to start with).
There are 9 functions that deal with `BUFFER`s:

```c
int mutt_extract_token (BUFFER *, BUFFER *, int);
BUFFER *mutt_buffer_new (void);
BUFFER * mutt_buffer_init (BUFFER *);
BUFFER * mutt_buffer_from (char *);
void mutt_buffer_free(BUFFER **);
int mutt_buffer_printf (BUFFER*, const char*, ...);
void mutt_buffer_add (BUFFER*, const char*, size_t);
void mutt_buffer_addstr (BUFFER*, const char*);
void mutt_buffer_addch (BUFFER*, char);
```

I wonder if they can be moved into `buffer.c` and their headers into `buffer.h`?
What do the functions depend upon?
What's the minimum we can change to achieve this?

nobrowser:
Taking this as the first step towards general header break up.

nobrowser:
I think you missed at least some.  The code I worked on in pattern.c takes BUFFERs for error tracking purposes.

guyzmo:
@nobrowser there are a huge number of functions that take buffers it's like one of the main types, but I guess @flatcap wants to gather all basic buffer manipulation in one place, eventually breaking general headers like `protos.h`, with the compilation unit being not clear (is it defined in `main.c`? `muttlib.c`? `init.c`? some place else?).

gahr:
I've been thinking for a while about adding small string optimization to buffers. However, that would require making them opaque. Just dropping it here so I don't forget, I'm fine if it's going to me taken care of separately (or even not at all if there's no consensus).

nobrowser:
gahr, can you explain what optimization you mean? thanks.

gahr:
@nobrowser, google for small string optimization

flatcap:
@gahr 
Hmm... a quick google shows lots of interesting ideas.
I think that would first need a good bit of research into how we use strings.
I think that should be discussed on the mailing list.

flatcap:
@nobrowser 
The functions I mentioned are self-contained.  There may be others.
I suppose I'd like to see two files that don't contain: `#include <mutt.h>` :-)

nobrowser:
@flatcap, ok, I get it now.

Is there a branch?  What should it be named?

flatcap:
As a member, you can create a branch.
[...]

PR #443, opened 2017-02-28
Devel/buffer refactor

nobrowser:
* What does this PR do?

Buffer struct refactor as planned

If it merges, it should be fine.  Only shuffling code around has been done.

Well, not really - I just shuffled it around.  Will be another pass if needed.

#377 

Edited: Tidied (@flatcap)

### DIFF
--- a/buffer.c
+++ b/buffer.c
@@ -1,0 +1,340 @@
+/*
+ *     This program is free software; you can redistribute it and/or modify
+ *     it under the terms of the GNU General Public License as published by
+ *     the Free Software Foundation; either version 2 of the License, or
+ *     (at your option) any later version.
+ * 
+ *     This program is distributed in the hope that it will be useful,
+ *     but WITHOUT ANY WARRANTY; without even the implied warranty of
+ *     MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ *     GNU General Public License for more details.
+ * 
+ *     You should have received a copy of the GNU General Public License
+ *     along with this program; if not, write to the Free Software
+ *     Foundation, Inc., 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301, USA.
+ */
+
+#include <string.h>
+#include <stdarg.h>
+#include <stdio.h>
+#include <stdlib.h>
+
+#include "lib.h"
+#include "myvar.h"
+#include "filter.h"
+#include "buffer.h"
+
+/* creates and initializes a BUFFER */
+BUFFER *mutt_buffer_new(void) {
+  BUFFER *b;
+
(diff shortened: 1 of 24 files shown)


flatcap:

gahr:
Should we try to keep .h/.c consistency? Specifically, should anything declared in myvar.h be defined in myvar.c?

gahr:
Also, I am not sure `mutt_extract_token` really belongs to `buffer.(c|h)`.. I don't have a strong opinion.. 

other than that, good to go for me!

flatcap:
> Should we try to keep .h/.c consistency?

Yes.

> Specifically, should anything declared in myvar.h be defined in myvar.c?

Ideally, yes, but myvar.h is a bit of a mixed bag.
That probably needs more rearranging.

> I am not sure mutt_extract_token really belongs to buffer.(c|h)

Hmm... It's slightly more BUFFER-y than not.

For now, I'm happy with this experiment.
I'll add the above to the list of what should considered to be broken out next.

nobrowser:
I also considered creating a myvar.c, but note that the issue said *minimal* set of changes to to isolate `BUFFER` ... so I resisted that urge.  Besides, it may have required *further* splitting if, for example, something in init.c depended on the new myvar.c .

buffer.c
```diff
@@ -0,0 +1,340 @@
...
+{
+  mutt_buffer_add (buf, s, mutt_strlen (s));
+}
+
+void mutt_buffer_addch (BUFFER* buf, char c)
+{
+  mutt_buffer_add (buf, &c, 1);
+}
+
+int mutt_extract_token (BUFFER *dest, BUFFER *tok, int flags)
```
nobrowser: This is line 145.  What to reindent here?

flatcap:
@P-EB 
This change is **just** a reorganisation of functions, so I'm not sure it could be split meaningfully.
Yes, the code's a mess, but that's not @nobrowser's fault :-)

When "The Big Tidy<sup>™️</sup>" occurs, most of the problems will disappear.

I'm happy with the code as it is, for now.

nobrowser:
I should like to point out that the root of all evil is protos.h.

It's that file that needs to be split more than any other.
The reason I created myvar.h and filter.h was to avoid an `#include "protos.h"` in the new buffer.c file.

flatcap:
Yes, I was referring to the formatting problems in `buffer.[ch]`.

I'm open to suggestions about what you'd like to split out next?

P-EB:
@flatcap the reindent issue seems to be introduced by the commit. I might be wrong but my diff seems to say so.

flatcap:
@P-EB In GitHub's diff view the indent **looks** awful because the tabs aren't handled well, but adding `?ts=8` to the URL makes it look ok.

Issue #341, opened 2017-01-29
Current position changes when displaying message


guiniol:
Under some conditions (not sure which, and I received new email during my investigations which stopped the issue triggering), when displaying a message, neomutt shows the right message but think its position is on another message. This shows when going back to the index (the selected message is not the message that was being read) or when moving to the next message (the next message is not the message following the message that was being read. When this happens, the erroneous position is always the same, even when closing and reopening the message, and the message displayed when trying to show the next message is the message following the one when going back to the index.

This is with neomutt 20170128 (1.7.2), and the bug disappeared before I could try with older versions of neomutt...

guyzmo:
next time it happens, could you try hitting <kbd>Control</kbd>-<kbd>L</kbd>? Sometimes displaying can look fishy with some 3-byte unicode characters.

guiniol:
@guyzmo I can, but I'm pretty sure it was the position being wrong. It took me back to almost the beginning of my mailbox, always in the same place, and I had to scroll back down. Except for the position, everything looked right.

guyzmo:
were you able to reproduce the error? And nail a way to make it reprocduceable?

guiniol:
Not by my own volition, but it does happen once in a while. And rivarun on irc was (maybe) having the same bug.
I can attach a gdb but I'm not sure where to put a breakpoint, and I haven't had time to figure it out. I need to spend more time in the pager/index code/event loop. Simply doing `continue`s does not work as gdb complains about not knowing the scope of the function at some point (not sure of the exact message) and refuses to `continue`.

guyzmo:
ok I'll keep those tags on as long as you're not providing a way for someone else to actually reproducing the issue.

guiniol:
No worries. I'll try to debug it on my own too.

gahr:
@guinol how do you mean that this happens in index-only too? I thought it was when getting back from the pager into the index. Thanks for clarifying!

guiniol:
@gahr I'm pretty sure I've had it jump around in the index only too. I'll have to check next time it appears.

guiniol:
btw, it's guiniol, with and `i`. You just mentionned someone else ☺

gahr:
Sorry for that, blame it on the phone ;)

guiniol:
@gahr no worries, just saying ^^

dset0x:
[...]

Issue #11, opened 2016-04-07
Folders periodically switching places in sidebar


alexpearce:
Firstly: this repo is a wonderful effort, thank you for your hard work! My apologies that my first interaction is reporting a problem.

I've modified the [Homebrew formula for mutt](https://github.com/Homebrew/homebrew-core/blob/master/Formula/mutt.rb) to use neomutt, adding the `--enable-sidebar` flag to the call to `./prepare`. The build succeeds, but when I start mutt the first and last folders in the sidebar switch periodically:

![Folder switching](https://cloud.githubusercontent.com/assets/540853/14344973/d4f2ac48-fcaa-11e5-9fa8-5a7e98319b06.gif)

The periodicity seems to take the value of `timeout`, which is set to 3 in my `muttrc`.

I thought the sidebar might be trying to sort the folders by itself, rather than the order defined in the `mailboxes` setting, but setting `sidebar_sort_method = "unsorted"` doesn't change the behaviour (as expected, really, as [the README](https://github.com/neomutt/neomutt/blob/fba0ecd82522c2eacb1aaa70044cae5a77c93acf/README.sidebar) says that “Unsorted isn't” is a known bug).

I usually define my mailboxes by using `tree` and `sed` to scan my Maildir for folders, manually inserting `+gmail/INBOX` before this scan so that it appears at the top of my folder list. For the purposes of debugging this problem, I just define them by hand:

```
mailboxes \
  "+gmail/INBOX" \
  "+gmail/archive" \
  "+gmail/Family" \
  ...
```

If I have only the following mailboxes defined, no folder switching occurs:

```
mailboxes \
  "+gmail/INBOX" \
  "+gmail/archive" \
  "+gmail/FIFO" \
  "+gmail/Family" \
  "+gmail/GitHub" \
  "+gmail/House"
```

If I add one extra folder, the switching occurs:

```
mailboxes \
  "+gmail/INBOX" \
  "+gmail/archive" \
  "+gmail/FIFO" \
  "+gmail/Family" \
  "+gmail/GitHub" \
  "+gmail/House" \
  "+gmail/L"
```

But now it's `INBOX` and `Family` that switch, not the first and last folders! 😄 

My only thought is that "L" follows "I" alphabetically, so maybe that somehow triggers the switching. It I move `INBOX` to come just before `L`, I still see switching, but now it's `archive` and `GitHub` that switch (so still the same indices). So, it's probably not that.

What could be causing this behaviour?

[...]

PR #447, opened 2017-03-01
Fix uncollapse_new in pager

### LINKED ISSUE #305: In pager mode threads are not expanded when new email is received
This report is against neomutt 20170113.

The following configuration is set in muttrc:

set sort=threads
set sort_aux=last-date-received
folder-hook . "exec collapse-all"
unset collapse_unread
set uncollapse_new

This should according to the options description cause new emails to uncollapse previously collapsed thread, and indeed it works fine in the index, but not in the pager mode, and new emails received on collapsed threads cause the "n" status flag to appear next to them:

```
34160     Jan 18 osstest service (   0) [Xen-devel] [ovmf test] 104239: all pass - PUSHED
34161 r   Jan 17 Stefano Stabell (#003) [Xen-devel] [RFC] Device memory mappings for Dom0 on ARM64 ACPI systems
34164     Jan 18 Dario Faggioli  (#004) [Xen-devel] [PATCH] xen: credit2: clear bit instead of skip step in runq_tickle()
34168     Dec 07 Juergen Gross   (#019) [Xen-devel] Xenstore domains and XS_RESTRICT
34187 n   Aug 19 mcgrof@kernel.o (#130) [Xen-devel] [PATCH v4 00/16] linux: generalize sections, ranges and linker tables
34317 N   Jan 18 Platform Team r (   0) [Xen-devel] [seabios baseline-only test] 68387: trouble: blocked/broken
```

sileht:
In pager, the index refresher uses the current count as oldcount.

This change fixes that.

Closes #305

### DIFF
--- a/pager.c
+++ b/pager.c
@@ -2059,8 +2059,6 @@ mutt_pager (const char *banner, const char *fname, int flags, pager_t *extra)
 
 	  if (index && Context)
 	  {
-	    oldcount = Context->msgcount;
-	    index->max = Context->vcount;
 
 	    /* After the mailbox has been updated,
 	     * index->current might be invalid */
@@ -2071,6 +2069,8 @@ mutt_pager (const char *banner, const char *fname, int flags, pager_t *extra)
 	    Context->quiet = 1;
 	    update_index (index, Context, check, oldcount, index_hint);
 	    Context->quiet = q;
+
+	    index->max = Context->vcount;
 
 	    /* If these header pointers don't match, then our email may have
 	     * been deleted.  Make the pointer safe, then leave the pager */


gahr:
Do you think this could fix #341 too?

guiniol:
I would think not because I think I've had it appear even in the index only.

guiniol:
So. I can't remove the approval, but it's not working here. May be related to #341. Don't merge yet.

flatcap:
@guiniol 
> I can't remove the approval

In the "Changes approved" section, "show all reviewers"
Inside you can "Dismiss review"

guiniol:
@flatcap thanks

guiniol:
@gahr @sileht can any one of you check that you can move while in the pager with this patch? I would like to know the bug I'm seeing is due to #341 or not

dset0x:
I have also been experiencing #341, but with this this I see the "reverse" behavior of that one. (Not always reproducible)

With this patch, given a mailbox of 12 messages {0..11},

* Pager open on any message. Pressing `next-entry` jumps to #11.
* Pager open on any message. Pressing `previous-entry` jumps to #10.
* Pager open on any message. Pressing `exit` jumps to #11.
* Pager open on any message. Pressing `display-toggle-weed`: jumps to #11 and toggles weeding.

guiniol:
I'm actually seeing the exact same behaviour as @dset0x 
Could be notmuch related.

sileht:
I have pushed a new version. I have found that update_index is called earlier with a wrong oldcount too.

gahr:
Sorry folks I haven't been able to look at this in details yet. I'll get back with comments.

guiniol:
@sileht This second take seems to fix the earlier bug.

dset0x:
It looks like it's solved it for me as well.

dset0x:
Disregard that, just reproduced it.

sileht:
I'm bet is, this change is correct but show up another display bug. update_index is now called with the correct arguments, I have checked the oldcount and msgcount contents when the display bug occurs and it's always good. 

I don't think we should merge that, until we fix the display bug too.

I my case, the display bug is, sometimes I have 2, 3 or 4 almost empty lines  inserted before the index.

all inserted lines start with some random characters.

sileht:
I think I have found the culprit, index->max was updated before update_index() with Context->vcount.
While update_index() update Context->vcount in some circumstance. Making max and vcount out of sync, then the redraw messup.

This bug can't occurs if msgcount == oldcount, that why we didn't see it before.

I will continue to test this change, just in case I have added it to my mutt in production :)

dset0x:
I'm afraid I'll be the bearer of bad news and let you know I just reproduced it again with 013f9d1bf06cc3b.

Do you think you could add some debug/print statements that are relevant to you, so that one can run with those? (I'm sorry for not doing it myself, but I haven't looked at the relevant code at all)

sileht:
@dset0x, what did you reproduce ?

Does the pager was open ? or maybe this #341 that have the same issue on index.

dset0x:
@sileht I've never had this happen in the index and I don't know if @guiniol ever confirmed it.

What happens is that the selection changes _while_in the pager, but only _shows that it has changed on-screen_ when exiting the pager, or moving to another message.

PR #448, opened 2017-03-02
Add debug_level/debug_file options

### LINKED ISSUE #445: Make the debug level configurable within neomutt
Starting neomutt with the option `-d 2` enables debug logging (if enabled by the build).
Often, the user hasn't enabled this logging when then encounter a problem.
Often, restarting will leave neomutt in a slightly different state to the one that caused the problem.

I propose changing the internal `debuglevel` variable into a user-configurable `$debug_level`

Use case: User sees some unexpected behaviour
- `set debug_level=5`
- repeat actions to repeat error

When `debug_level` is set > 0, a new log file will be opened.
When `debug_level` is set to 0, the log file will be closed.

sileht:
* What does this PR do?

This change allows to change debug_level and debug_file on the fly.

We can't really debug the early startup of mutt with these new options.

* Why was this PR needed?

To solve #445 

   - [X] Documentation created/updated
   - [X] All builds are passing
   - [X] Added [doxygen code documentation](https://www.neomutt.org/dev/doxygen)
   - [X] Code follows the [style guide](https://www.neomutt.org/dev/coding-style)

Closes #445

### DIFF
--- a/init.c
+++ b/init.c
@@ -1913,10 +1913,17 @@ static void restore_default(struct option_t *p)
       break;
     case DT_PATH:
       FREE((char **) p->data);
-      if (p->init)
+      char *init = NULL;
+#ifdef DEBUG
+      if (mutt_strcmp(p->option, "debug_file") == 0 && debugfile_cmdline)
+        init = debugfile_cmdline;
+      else
+#endif
+        init = (char *) p->init;
+      if (init)
       {
         char path[_POSIX_PATH_MAX];
-        strfcpy(path, (char *) p->init, sizeof(path));
+        strfcpy(path, init, sizeof(path));
         mutt_expand_path(path, sizeof(path));
         *((char **) p->data) = safe_strdup(path);
       }
@@ -1938,7 +1945,12 @@ static void restore_default(struct option_t *p)
     case DT_NUM:
     case DT_SORT:
     case DT_MAGIC:
-      *((short *) p->data) = p->init;
+#ifdef DEBUG
+      if (mutt_strcmp(p->option, "debug_level") == 0 && debuglevel_cmdline)
+        *((short *) p->data) = debuglevel_cmdline;
+      else
+#endif
+        *((short *) p->data) = p->init;
       break;
     case DT_RX:
     {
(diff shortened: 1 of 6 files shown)


sileht:
Does the init.h options documentation are enough for this advanced stuff (they are appended at the end of manual), or should I create a new section in the manual ?

gahr:
I think we should be consistent with respect to log rotation. In your current approach:

- if `debug_file` is not set, the default `~/.muttdebug.%d` is used and rotated
- if `debug_file` is set, no rotation happen

Wouldn't it be acceptable to have a `.%d` appended automatically and rotation happening every time the debug is re-initialized?

If we proceed like this, you could have `DebugFile` default to "~/.muttdebug" and avoid the need for a second `debugfilepath` variable. This one also leaks memory, although it's not too severe a problem.

sileht:
I have hesitated to do append the '%d' too, so if other agree, I will do it.

init.c
```diff
@@ -2336,6 +2403,8 @@ static int parse_set (BUFFER *tmp, BUFFER *s, unsigned long data, BUFFER *err)
 	  strfcpy (scratch, tmp->data, sizeof (scratch));
 	  mutt_expand_path (scratch, sizeof (scratch));
 	  *((char **) MuttVars[idx].data) = safe_strdup (scratch);
+      if (mutt_strcmp (MuttVars[idx].option, "debug_file") == 0)
+	    restart_debug ();
```
guyzmo: build is broken, certainly because when `DEBUG` is not defined, this line cannot compile.

https://travis-ci.org/neomutt/neomutt/builds/206916932#L483

init.c
```diff
@@ -2540,6 +2609,13 @@ static int parse_set (BUFFER *tmp, BUFFER *s, unsigned long data, BUFFER *err)
 	  *ptr = 0;
 	mutt_init_history ();
       }
+      else if (mutt_strcmp (MuttVars[idx].option, "debug_level") == 0)
+      {
+	if (*ptr < 0)
+	  *ptr = 0;
+	restart_debug ();
```
guyzmo: same.

sileht:
timestamps are already added by mutt_debug()

sileht:
@guyzmo all your points have been fixed.

init.c
```diff
@@ -87,6 +87,64 @@ static char **nm_tags;
...
+static void restart_debug(void)
+{
+   if (debuglevel > 0)
+   {
+     mutt_debug (1, "NeoMutt/%s (%s) stop debugging\n",
+                 PACKAGE_VERSION, MUTT_VERSION);
+     safe_fclose(&debugfile);
+   }
+
+   memcpy(&debuglevel, &DebugLevel, sizeof(int));
```
gahr: debuglevel = DebugLevel ?

sileht: obviously. done.

gahr:
I'm not sure it's wise to restart debug upon a change in debug_level. Why is that needed?

sileht:
@gahr I have added newl logic to not restart debug when just debug_level is changed.

sileht:
@flatcap that was just a logging issue, when the level change nothing was printed, I fixed that.

Also in  current implementation, muttrc is loaded after the debug is started, so "set devel_level = 3" will override any mutt -d option.

flatcap:
OK, that's better.

> muttrc is loaded after the debug is started

Hmm... Is that what we want?  I'd expect the command line to have a higher priority than the config file.
(I suppose ideally, we'd have a debugfile command line option too)

ATM, if I run `mutt -d 2` then mutt creates `~/muttdebug0` before it reads the `set debug_file` line in my config.  I can't prevent it creating that file.

sileht:
On Thu, Mar 30, 2017 at 03:14:07PM -0700, Richard Russon wrote:
>OK, that's better.
>
>> muttrc is loaded after the debug is started
>
>Hmm... Is that what we want?  I'd expect the command line to have a higher priority than the config file.
>(I suppose ideally, we'd have a debugfile command line option too)

I would prefer to have command line that override muttrc, too.

>ATM, if I run `mutt -d 2` then mutt creates `~/muttdebug0` before it reads the `set debug_file` line in my config.  I can't prevent it creating that file.

I have kept that to allow to debug the init as earlier as possible,
especially to have the muttrc parsing. But adding a new command line
option to override debug_file on startup will solve that.

So to resume my next version:
* setting only debug* in muttrc, will start the debug when the line is read
  from the config file.
* Setting debug* on command line will start the debug at init beginning.
* Setting both, command line always wins.

-- 
Mehdi Abaakouk
mail: sileht@sileht.net
irc: sileht

flatcap:
> I have kept that to allow to debug the init as earlier as possible, especially to have the muttrc parsing

Ah, I see.

---

\<thinking aloud\>

Hmm... does the file parsing actually have any **debugging**?
Aren't the errors just printed to stderr?
```
NeoMutt/20170306 (1.8.0) debugging at level 2
getdnsdomainname(): flatcap.org
Reading configuration file '/home/mutt/.muttrc'.
Reading configuration file '/home/mutt/.mutt/lists.rc'.
In mutt_reflow_windows
Reading configuration file '/home/mutt/.mutt/alternates.rc'.
Reading configuration file '/home/mutt/.mutt/crypto.rc'.
Reading configuration file '/home/mutt/.mutt/sidebar-dev.rc'.
In mutt_reflow_windows
In mutt_reflow_windows
In mutt_reflow_windows
Reading configuration file '/home/mutt/.mutt/imap.rc'.
Reading configuration file '/home/mutt/.mutt/mailbox.rc'.
Reading configuration file '/home/mutt/.mutt/colours-dev.rc'.
Reading configuration file '/home/mutt/.mutt/status-color.rc'.
Reading configuration file '/home/mutt/.muttrc.local'.
NeoMutt/20170306 (1.8.0) stop debugging 
```
Could we do something like the linux kernel at startup?  It stores its messages in a ring buffer, that gets dumped as soon as there's a tty.
Hmm... too complicated and if mutt crashes nothing will get logged.

\</thinking\>

---

> So to resume my next version:

Great!  Thanks.

flatcap:
Merged in 40f81d1.  Thanks.

PR #449, opened 2017-03-02
attachment/pager: Use mailcap for test/* except plain

### LINKED ISSUE #430: Pressing return doesn't view text/html attachment with mailcap
My distro switched me from mutt to neomutt and I immediately noticed that I could no longer view html attachments using my mailcap rule.   (If I select a message, hit "v" and then hit "enter" the HTML is simply shown in text.)  I finally dug into this issue and I noticed that the help text for <Enter> says "view-attach" and describes this as "view attachment using mail cap entry if necessary".  I observe another command, "m" which is bound to "view-mailcap" which says "force viewing of attachment using mailcap".   

So I don't understand if this is a bug or if things are working as intended, but when I press return, I just see the text of the HTML, which is mostly useless.   Is use of mailcap deemed "not necessary" in this case because the attachment has a "text" type?  

Pressing "m" invokes my mailcap rule and displays the HTML as desired.

sileht:
This change makes attachment view using mailcap for everything
not text/plain.

And in pager, don't assume text/* are human readable except text/plain.
This can still be overridable with auto_view.

Closes #430

### DIFF
--- a/muttlib.c
+++ b/muttlib.c
@@ -734,10 +734,9 @@ int mutt_needs_mailcap (BODY *m)
   switch (m->type)
   {
     case TYPETEXT:
-      /* we can display any text, overridable by auto_view */
-      return 0;
+      if (!ascii_strcasecmp ("plain", m->subtype))
+        return 0;
       break;
-
     case TYPEAPPLICATION:
       if((WithCrypto & APPLICATION_PGP) && mutt_is_application_pgp(m))
 	return 0;
--- a/handler.c
+++ b/handler.c
@@ -1823,7 +1823,7 @@ int mutt_body_handler (BODY *b, STATE *s)
     else if (ascii_strcasecmp ("enriched", b->subtype) == 0)
       handler = text_enriched_handler;
     else /* text body type without a handler */
-      plaintext = 1;
+      plaintext = 0;
   }
   else if (b->type == TYPEMESSAGE)
   {


muttlib.c
```diff
@@ -76,6 +76,18 @@ static const char *xdg_defaults[] =
   [kXDGConfigDirs] = "/etc/xdg",
 };
 
+static const char * const mailcap_text_as_application[] =
+{
+  "html",
+  "xml",
+  "calendar",
+};
```
guyzmo: I think this should be defined as a `LIST*` exposed as a global `optiont_t` in `init.h`, with a `set` or a command to update that list.

guyzmo:
cf the linked issue. `auto_view` enables displaying of non `text/*` mime contents, whereas all `text/*` is automatically viewed. So the patch makes it so some `text/...` that should belong to `application/...` are not automatically viewed.

Though I've been thinking, and maybe a much simpler approach is to only always open `text/plain` per default, and add all `text/...`isms explicitely.

sileht:
Agreed, that's what my last push does

guyzmo:
@neomutt/reviewers please review (more than 2, it's trivial but it's a visible change).

gahr:
@sileth have you tested it? It doesn't seem to work here: `text/x-vcard` is still being displayed inline.
I have no `auto_view` and no `mailcap` entries for that MIME type.

sileht:
@gahr, if you have no mailcap entry, mutt always fallback to builtin viewer for attachments, with or without my patch.

sileht:
And should print 'No matching mailcap entry found.  Viewing as text.'

gahr:
I thought your patch was supposed to only view `text/plain` within the pager, and treat all others as non-viewable.

With - or without - your patch I get:

```
[-- Attachment #2: image001.jpg --]                                                                                                                              
[-- Type: image/jpeg, Encoding: base64, Size: 15K --]                                                                                                            
                                                                                                                                                                 
[-- image/jpeg is unsupported (use 'v' to view this part) --]                                                                                                    
                                                                                                                                                                 
                                                                                                                                                                 
[-- Attachment #2: Prof  Whatever.vcf --]                                                                                                                  
[-- Type: text/x-vcard; charset=us-ascii, Encoding: base64, Size: 7.0K --]                                                                                       
                                                                                                                                                                 
BEGIN:VCARD                                                                                                                                                      
VERSION:2.1                                                                                                                                                      
X-MS-SIGNATURE:YES
...
```

I would have expected:

```
[-- Attachment #2: image001.jpg --]                                                                                                                              
[-- Type: image/jpeg, Encoding: base64, Size: 15K --]                                                                                                            
                                                                                                                                                                 
[-- image/jpeg is unsupported (use 'v' to view this part) --]                                                                                                    
                                                                                                                                                                 
                                                                                                                                                                 
[-- Attachment #2: Prof  Whatever.vcf --]                                                                                                                  
[-- text/x-vcard is unsupported (use 'v' to view this part) --]                                                                                                    
```

gahr:
> @sileht  And should print 'No matching mailcap entry found. Viewing as text.'

Yes, this happens in the attachment browser, with or without your patch.

sileht:
This PR is only for the attachments brower, I will try to propose another PR, to fix the pager.

gahr:
Oh, I see. Sorry for the noise, then.

sileht:
I have added a second commit that does the same thing on the pager, to have a consistent behavior.

gahr:
Wonderful, I was about to propose that diff!

guyzmo:
Wondermarvelful! ♥

PR #455, opened 2017-03-05
getdnsdomainname: cancel getaddrinfo_a if needed

### LINKED ISSUE #453: Memory corruption after using getaddrinfo_a()
This is the result of a **lot** of testing over two days. (I have run neomutt over 100,000 times.)

I found that the problem:

- corrupts several global strings (maybe not at the same time)
  - Tempdir (`$tmpdir`)
  - SmtpUrl (`$smtp_url`)
- affects strings that we have malloc'd
  - writes 0xFFFFFFFE (4 bytes) over the beginning
  - only seems to affect the start of strings
- is intermittent
- is caused by the hostname lookup
  - dependent on `getaddrinfo_a()`
  - only the async version
- is strongly influenced by timing
- is probably the cause of #438 (Fedora builds broken).
- seems to be IMAP-dependent

I started investigating #438 in a minimal VM and managed to repeat the problem occasionally (a few percent of the runs).  I started running running `mutt -e 'exec exit'` in blocks of 1000 to make the problem reoccur.

The OP reported that running `mutt -d2` "fixed" the problem.  I found that the problem still occurred, but less frequently.  Then the problem stopped happening.  For the rest of the night, I couldn't make it happen.

Next day, it started happening again.  Same setup.  Whatever was causing the problem was in the latest code and strongly timing-dependent.  I narrowed it down to `getaddrinfo_a()`.

The problem can be "fixed", by:

- getting `getdnsdomainname()` to return immediately (without doing anything)
- disabling the use of `getaddrinfo_a()`
- changing `GAI_NOWAIT` to `GAI_WAIT`
- adding `127.0.0.1 myhost.example.com` to `/etc/hosts`
  (so that the hostname lookup is instantaneous)

I tried altering the `timeout` variable (making it much smaller or larger), but I didn't manage to get any reliable results.

By adding a lot of `printf()` debugging, I managed to find one place where `Tempdir` was being corrupted.  In `tls_negotiate()` the call to `gnutls_handshake()` caused the corruption.

I *think* that the problem may be caused by someone else doing a `getaddrinfo()` lookup, but I can't imagine how that would cause random string corruption.

/cc @sileht

sileht:
* What does this PR do?

if getaddrinfo_a() is not finish on time we must call
gai_cancel() and wait getaddrinfo_a() to finish.

I can't reproduce the initial issue #453, but now, the missing gai_cancel() looks obvious.

* Why was this PR needed?

Closes #453

### DIFF
--- a/getdomain.c
+++ b/getdomain.c
@@ -52,16 +52,31 @@ int getdnsdomainname (char *d, size_t len)
    * If it takes longer, the system is mis-configured and the network is not
    * working properly, so...
    */
+  bool status;
   struct timespec timeout = {0, 100000000};
   struct gaicb *reqs[1];
   reqs[0] = safe_calloc(1, sizeof(*reqs[0]));
   reqs[0]->ar_name = node;
   reqs[0]->ar_request = &hints;
-  if ((getaddrinfo_a(GAI_NOWAIT, reqs, 1, NULL) == 0) &&
-      (gai_suspend((const struct gaicb * const *) reqs, 1, &timeout) == 0) &&
-      (gai_error(reqs[0]) == 0))
+  if (getaddrinfo_a(GAI_NOWAIT, reqs, 1, NULL) == 0)
   {
-    h = reqs[0]->ar_result;
+    gai_suspend((const struct gaicb * const *) reqs, 1, &timeout);
+    status = gai_error(reqs[0]);
+    if (status == 0)
+      h = reqs[0]->ar_result;
+    else if (status == EAI_INPROGRESS)
+    {
+      mutt_debug(1, "getdnsdomainname timeout\n");
+      /* request is not finish, cancel it to free it safely */
+      if (gai_cancel(reqs[0]) == EAI_NOTCANCELED)
+      {
+        while (gai_suspend ((const struct gaicb * const *) reqs, 1, NULL) != 0)
+          continue;
+      }
+    }
+    else
+      mutt_debug(1, "getdnsdomainname fail: (%d) %s\n",
+                 status, gai_strerror(status));
   }
   FREE(&reqs[0]);
 


flatcap:
Thanks, @sileht.  I'll leave this running overnight.

getdomain.c
```diff
@@ -52,16 +52,31 @@ int getdnsdomainname (char *d, size_t len)
    * If it takes longer, the system is mis-configured and the network is not
    * working properly, so...
    */
+  bool status;
```
gahr: This should be ìnt`.

getdomain.c
```diff
@@ -52,16 +52,31 @@ int getdnsdomainname (char *d, size_t len)
...
+    status = gai_error(reqs[0]);
+    if (status == 0)
+      h = reqs[0]->ar_result;
+    else if (status == EAI_INPROGRESS)
+    {
+      mutt_debug(1, "getdnsdomainname timeout\n");
+      /* request is not finish, cancel it to free it safely */
+      if (gai_cancel(reqs[0]) == EAI_NOTCANCELED)
+      {
+        while (gai_suspend ((const struct gaicb * const *) reqs, 1, NULL) != 0)
```
gahr: `while (... == EAI_NOTCANCELLED)` perhaps?

flatcap: I /think/ the code is right.  We **forcibly** cancel the request, then we wait for confirmation.  `gai_suspend()` could conceivably be interrupted by a signal, so that loop's needed.

P-EB: Btw in the commit, there is a typo, CANCELLED takes two 'L'.

gahr: Yes, that makes sense. Sorry for the noise.

flatcap: @P-EB hehe, blame [Noah Webster](https://en.wikipedia.org/wiki/English-language_spelling_reform) for that.
Perhaps we should write some `en_GB.h` compatibility headers.

flatcap:
I racked up 300,000 runs last night without any problems.

flatcap:
Merged.

getdomain.c
P-EB: Oh, I wasn't aware of that difference.

PR #457, opened 2017-03-06
Abort when message is empty

### LINKED ISSUE #191: Quick way to skip sending empty mail
Say I don't want to send a mail, then I can exit the editor, press 'q' in the compose screen, confirm that I don't want to postpone the mail and go back to index.

Desired behaviour: let composer check if the text of the mail is empty and then skip asking, ie. go directly to index. The test could be simple, empty file (or just whitespace), or parse the file and look if body is empty.

A non working attempt (send.c::ci_send_message):

```

     fcc_error = 0; /* reset value since we may have failed before */
     mutt_pretty_mailbox (fcc, sizeof (fcc));
+    if (msg->lines == 0) {
+      /* abort */
+      mutt_message _("Mail empty, sending aborted.");
+      goto cleanup;
+    }
     i = mutt_compose_menu (msg, fcc, sizeof (fcc), cur);
     if (i == -1)
     {
```

sileht:
When the message to send is empty, this doesn't really make sense to
show the composer. We should return directly to the index.

Closes #191

### DIFF
--- a/send.c
+++ b/send.c
@@ -1662,7 +1662,7 @@ ci_send_message (int flags,		/* send mode */
       if (stat (msg->content->filename, &st) == 0)
       {
 	/* if the file was not modified, bail out now */
-	if (mtime == st.st_mtime && !msg->content->next &&
+	if ((mtime == st.st_mtime || st.st_size == 0) && !msg->content->next &&
 	    query_quadoption (OPT_ABORT, _("Abort unmodified message?")) == MUTT_YES)
 	{
 	  mutt_message (_("Aborted unmodified message."));


flatcap:

gahr:
I was under the impression that `abort_unmodified` made #191 a non-issue. What erroneous behaviour of what use case is this PR fixing?

sileht:
@gahr: When `abort_unmodified = yes`, instead of only checking the `mtime` of the edited file, this will also check that the file is not empty.

gahr:
@sileht I understand what the change does. I don't understand which use case does it fix.

sileht:
I don't have a strong opinion this change. I have understood the use case described in the original bug report, as:
 * we need a way to return to the index directly when we have modified the mtime of the edited file.

Checking file size is the easier way to resolve this use case.

gahr:
So, are we saying that

* edit a message
* remove the content (size -> 0)
* save the file (mtime -> modified)
* close the editor
* abort composing (**this change**)

is the wanted behaviour? I am not sure I agree.

flatcap:
Saving a zero-length file does seem to be sending a clear message &mdash; I'm done.

sileht:
I just realize I use `edit_headers=yes` and for me deleting headers mean I will don't send this mail.

But if `edit_headers=no`, this change makes impossible to send a empty mail with only a subject...

toogley:
> But if edit_headers=no, this change makes impossible to send a empty mail with only a subject...

which is important IMHO, because some mailinglists do the registration confirmation by replying with only a subject.

sileht:
So I'm more in favour of not fixing #191 at all. (And abandon this)

flatcap:
OK, @sileht.
Please can you add a quick summary to #191 (cut'n'paste from above), then close both.

kdave:
Actually, this fixes the problem for me. The initial wording in #191 "or parse the file and look if body is empty" was unfortunatelly misleading as I have `edit_headers=yes` and thus body == the whole mail for me. This got clarified in followup comments at least.

PR #463, opened 2017-03-11
fix: Garbage in chdir prompt due to unescaped string

flatcap:
I upgraded one of my stable systems to stretch and noticed that when
changing the directory with 'c', there was some garbage at the end of
the string in the prompt. I've not noticed that before on my main
amd64 system, but this stable system is i386, so different alignment
and word size that can affect this kind of thing.

I tracked it down to an unescaped string handling.

Debian [BTS#857433](https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=857433)


flatcap:

PR #466, opened 2017-03-13
Tidy the code - Part 1

flatcap:
Here is the first batch of tidying-up commits.
They are all quite simple and unlikely to cause offence :-)

- 2243a26 drop control characters from the source
- ac1380d drop vim modelines
- 8b61a2a delete trailing whitespace
- 2693203 mark all local functions as static
- 99e4473 delete unused functions
- 296098b replace FOREVER with while (true)
- e55c4f2 drop #if HAVE_CONFIG_H
- 52c2e03 use #ifdef for potentially missing symbols
- 0d5475d remove #if 0 code blocks
- 6379690 drop commented out source


flatcap:
@neomutt/reviewers 
Anyone have time to give their views on Part 1?
I'd like to get it committed because Part 2 is built on top (and already finished).

Issue #462, opened 2017-03-10
Fallback from iso-8859-1 to cp1252 if needed


gahr:
Some MUAs - I have an example sent by Microsoft Outlook 12.0 - pretend to be sending iso8859-1 quoted-printable stuff when they in fact are using cp1252.

Let's provide a fallback mechanism so we can properly convert and avoid weird `\205`, `\222`, and similar sequences in the decoded message.

gahr:
Sample email:

```
From example@example.com Fri Mar 10 10:44:56 2017
From: <example@example.com>
To: <example@example.com>
Subject: Fake cp1252
Date: Fri, 10 Mar 2017 11:44:57 +0100
Message-ID: <001501d2998b$5c8946d0$159bd470$@ch>
MIME-Version: 1.0
Content-Type: multipart/alternative;
    boundary="----=_NextPart_000_0016_01D29993.BE4DAED0"
X-Mailer: Microsoft Office Outlook 12.0

------=_NextPart_000_0016_01D29993.BE4DAED0
Content-Type: text/plain;
    charset="iso-8859-1"
Content-Transfer-Encoding: quoted-printable

Hey,

some ellipsis=85.(!)

An an apostrophe =92

------=_NextPart_000_0016_01D29993.BE4DAED0--
```

gahr:
I'd rather not treat this as a bug, since the originating MUA is clearly lying.

gahr:
After some more reading and investigation, I am pretty convinced that a simpler solution would be to just always use cp1252 when the advertised charset is iso-8859-1. I'll commit something along these lines in the next few days.

PR #467, opened 2017-03-13
Use cp1252 instead of iso-8859-1

gahr:
Some MUAs, e.g., Microsoft Outlook 12.0, pretend to be sending text encoded in iso-8859-1 when they're actually using cp1252. Being cp1252 a superset of iso-8859-1 - characters in the range 0x80-0x9f are printable characters and symbols in the former and control characters in the latter - this change suggests using cp1252 instead of iso-8859-1 when doing iconv conversion.

Fixes: #462

### DIFF
--- a/charset.c
+++ b/charset.c
@@ -74,14 +74,15 @@ PreferredMIMENames[] =
   { "csISO2022JP",           "iso-2022-jp"   },
   { "csISO2022JP2",          "iso-2022-jp-2" },
 
-  { "ISO_8859-1:1987",       "iso-8859-1"    },
-  { "iso-ir-100",            "iso-8859-1"    },
-  { "iso_8859-1",            "iso-8859-1"    },
-  { "latin1",                "iso-8859-1"    },
-  { "l1",                    "iso-8859-1"    },
-  { "IBM819",                "iso-8859-1"    },
-  { "CP819",                 "iso-8859-1"    },
-  { "csISOLatin1",           "iso-8859-1"    },
+  { "ISO_8859-1:1987",       "cp1252"        },
+  { "iso-ir-100",            "cp1252"        },
+  { "iso_8859-1",            "cp1252"        },
+  { "latin1",                "cp1252"        },
+  { "l1",                    "cp1252"        },
+  { "IBM819",                "cp1252"        },
+  { "CP819",                 "cp1252"        },
+  { "csISOLatin1",           "cp1252"        },
+  { "iso-8859-1",            "cp1252"        },
 
   { "ISO_8859-2:1987",       "iso-8859-2"    },
   { "iso-ir-101",            "iso-8859-2"    },
@@ -215,6 +216,7 @@ PreferredMIMENames[] =
 };
 // clang-format on
 
+static const char * const DefaultCharset = "cp1252";
 #ifdef HAVE_LANGINFO_CODESET
 # include <langinfo.h>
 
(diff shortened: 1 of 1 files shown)


gahr:

gahr:
I just realized that `charset-hook ^iso-8859-1$ cp1252` does the same. I still think this should be the default, as there's nothing to be lost, or is there?

Thoughts are welcome!

guyzmo:
I don't think this is a wise change, especially given there's a working solution to get around that issue.

This is because we cannot decide what codepage is the most popular amongst mail clients that send mail with codepage with an ISO charset. Also, when specifying an ISO charset when I send a mail, I'm expecting receiver to show the mail with that charset.

If we need to bind the rules for every stupid thing MS/Outlook devs have made, we would need to rewrite mutt just for receiving mail from an Outlook.

Finally, we're in the Unicode era, and such considerations should be less a worry, as ill-encoded e-mails is slowly becoming something of the past. Just alike the fact that Outlook 12 is 10 years old and shouldn't be a concern for a patch.

gahr:
@guyzmo, I agree with you, in theory. Just two notes that you might have missed:
- I am modifying the handling of only iso-8859-1, not all ISOs
- cp1252 is a superset of iso-8859-1
- there are people out there still using those old mailers
- should we expect all mutt users to understand charsets and regexps? (The bare minimum to setup a charset-hook)

It's about a trade off between being purists and being user friendly, imho.

gahr:
As a side note, other mailers I have at hand (roundcube, osx mail, ios mail) handle that conversion transparently for me.

flatcap:
There may be fewer bad emailers out there, but think how much old mail you've got stored!
(unless you want a filter to convert all your emails to utf-8)

gahr:
Good point!

guyzmo:
Well, over 5 years of mails (10GB Maildir), I never received any MS Outlook 12 mails.

If that happens and when that happens, I'd rather change the default behaviour for a single user, or using a default hook to change the charset binding the way you've suggested.

For the stats, in my personal mail corpus I got 803 MSO14, 415 MSO15, 31 MSO16, 267 MSOE ; 2143 mails that comes from a variant of Outlook.

Please, check your mailboxes and tell me what proportion of mails have you got coming from MSO12, and how many users were they? I think we're dealing with an exception to the rule, and that does not justify changing the default, but rather dealing with it using an exceptional feature.

Note that I'm not raising the argument that codepages using private pages of the ASCII standard to display characters is a bad thing. Nope. Not doing it 😇

guyzmo:
> > @guyzmo N.B.: I believe I made my point, I won't veto this PR, but
> > I'm not in favor of such a change.
> I don't believe I read a valid point, but I might have missed it.

the main point is the one *you* made: there's a way, as hackish as it is,
to fix the issue locally without a patch.

You might want to make that fix global, or just as a `message-hook` that
will apply the fix for mails coming from MSO12 senders or with X-Mailer.

> imply "if my mailer specified charset A when it's actually sending
> text in charset B, I'm expecting the receiver to see garbage"?

Basically yes. If you set an encoding of utf-16 and write ASCII
characters in it, it's *your* responsibility and fault as the sender to
misdirect the reader.

In the current case, the tool displaying the text is supposed to display
nothing instead of garbage for all the private pages. So lack of support
should only make invisible characters non vital for reading.

> Which one(s) of these are your strong points?
> Also, what is the *downside* of such a change?

there is no major reason to make into a patch something that can be
handled properly with a `message-hook`.

N.B.: I'm only precising because you've asked. I see no point in
applying such a patch. But I see no point in fighting over this issue
further, so I'm leaving this to the community.

gahr:
Thank you for your elaboration.

guyzmo:
you're welcome ☺

gahr:
For the record, Roundcube does [the same](https://github.com/roundcube/roundcubemail/blob/master/program/lib/Roundcube/rcube_charset.php#L121-L126).

gahr:
For the record, upstream has turned down this proposal:
https://dev.mutt.org/trac/ticket/3925

gahr:
Closing this, as there is no consensus. I still fail to see a proper use case for using undefined character codes (0x80-0x9f) with iso-8859-1 proper (binary encoding should be used for binary data anyway), but I can live with a charset hook.

flatcap:
This issue has been added to a list of [closed discussions and blue-sky ideas](https://www.neomutt.org/dev/discussions).

PR #470, opened 2017-03-15
Warn on bindkey aliasing

### LINKED ISSUE #451: Bindkey aliasing without warnings
* Why do you not like the current state?

When you have several bindkeys with the same prefix, the one with the shortest prefix is always executed, making all others unreachable. It's even worse because the order of bindkey definition matters.

Take the minimal muttrc alternatives:

    bind index g group-reply
    bind index gg first-entry

When you click `g`, nothing happens because the bindkey that is defined is the second one. `gg` gets executed.

    bind index gg first-entry
    bind index g group-reply

When you click `g`, a group reply is initiated. The first entry binding was lost.

* What would you like to see to change?

I don't think mutt need a Vim-like approach using timeouts, that is too complicated. But at the very least a warning message should be produced when you bind a key combination that aliases other.

somini:
* What does this PR do?

Warn the user when binding a key combination that will alias others.

* Why was this PR needed?

   - [ ] All builds are passing

Fixes #451


somini:
This is still in the printf phase, so don't merge this yet. This is just an easier way to track progress of the #451 issue.

guyzmo:
sure that's what the `in-progress` status tag is for! 👍

gahr:
What's the reason for moving `km_keyname` ?

somini:
@gahr Compile error otherwise, it's defined after I use it. I though that was what headers were for, but apparently not.

gahr:
@somini perhaps in a previous version. Currently, your changes do not call `km_keyname` and reverting e74afe2a8 works just fine.

somini:
@gahr Thanks for spotting that, it was indeed one of many experiments to pretty print key bindings. Force pushed a new version without that commit.

As for warning the user without `printf`, I was checking the `mutt_parse_rc_line`, but I can't understand how this calls the `km_bind` function I changed.
https://github.com/somini/neomutt/blob/5c03e09b3c8976a7fb94fa44d0c4dbefa94291cf/init.c#L2885
Where is that `Commands` variable defined?

somini:
I've force pushed because I rebased this into ``master``.

I've browsed around and found a way to warn the user using similar mechanisms as the existing errors on non-existing functions. This produces an error message with the rc file path and line number, which is a big help. It's not really an error, since it must not invalidate loading the file.

The code quality is dubious at best, but I think this can be overcome with some major refactoring. I'm counting on reviewers' help for guidance.

guyzmo:
force push is 👌 for PRs.

I'll try to have a look at your code ASAP, and test your changes this week. 

ping other @neomutt/reviewers 🙌

somini:
I'll be unavailable from this Thursday until the next Monday, but I'll read your comments when I come back.

somini:
Here's some more commits.

The noop detection is important, otherwise you will get spurious warnings since so many keys have bindings by default.

I'm not sure this is the correct way to treat PO files, but it's a separate commit to make sure I can revert it easily.

somini:
There's also this:
https://github.com/somini/neomutt/blob/bd9691b7b79dcbeaba9077ddea11c887ac39b96f/keymap.c#L749-L866

Shouldn't this be moved to the default config file? Is it necessary to keep this stuff in C code?

flatcap:
Rebased to master.
I've dropped the .po commit, for now.  I'll regenerate them all just before release.

somini:
Rebased on the new master, it required many changes.

It compiles and still works as expected, there were no non-cosmetic changes to this part of the code.

flatcap:
I made a few slight changes to your work.  I found that some valid mappings were displaying error messages (due to `try_bind()` being called twice).

This is the test muttrc that I used:
```
bind index,pager a noop
bind index,pager p noop
bind index,pager g noop

# alias -> generic
macro index a1  ':bind alias 5 delete-entry<enter>'
macro index a2  ':bind alias 6 bottom-page<enter>'
macro index a3  ':bind alias 7 wibble<enter>'
macro index a4  ':bind alias 55 delete-entry<enter>'
macro index a5  ':bind alias 66 bottom-page<enter>'

# pager (only)
macro pager p1 ':bind pager j tag-message<enter>'
macro pager p2 ':bind pager k bottom-page<enter>'
macro pager p3 ':bind pager k wibble<enter>'
macro pager p4 ':bind pager jj tag-message<enter>'

# generic (only)
macro index g1 ':bind generic 5 bottom-page<enter>'
macro index g2 ':bind generic 6 wibble<enter>'
macro index g3 ':bind generic 55 bottom-page<enter>'
```

I testes a binding in a real menu, then one in the generic part, then a non-existent one.
Finally, I tested overwriting mappings.

It all seems fairly solid to me.

Thanks.

somini:
Thanks!

Closes #451

Issue #460, opened 2017-03-08
Use external gettext


gahr:
* **feature requests**

    * Why do you not like the current state?

Mutt bundles an old implementation of the gettext library in the `intl` subdirectory. It is a pain to maintain and adds little value. I think it's reasonable to assume that anybody using Mutt with i18n has a gettext runtime installed.  

    * What would you like to see to change?

Get rid of the `intl` subdirectory and rely on the system's gettext.

flatcap:
This commit "fixes" the build for me (see the log message :-)
If it's no good, delete it.

guyzmo:
I do like the idea

gahr:
Merged

PR #472, opened 2017-03-15
Switch to using an external gettext runtime

gahr:
This PR allows us to get rid of the `intl` directory that used to keep an old version of the `gettext` implementation. The former approach is changed into using whatever `gettext` is installed in the system.

Generated `m4` files have been removed and added to `.gitignore`.

No code changes have been introduced. Please just check your build workflow :)

Issue: #460


gahr:

gahr:
You can start reviewing, but I have to sort out a problem with `--disable-iconv`. Perhaps it makes sense not to have it altogether. See also what's written in `INSTALL`.

gahr:
7004f96 removes support for building neomutt without iconv support. Please comment.

gahr:
Rebased

flatcap:
@gahr I've made one slight change -- I've reinstated `m4/funcs.m4` (temporarily).

In `configure.ac` there's a command `XIPH_FUNC_VA_COPY` (which I'd never seen before) which checks for the presence of `va_copy` which is part of C99.  Without that custom M4, I get warnings during configure.

I _think_ we're probably safe to delete both, but I'll wait 'til you're back.

Issue #473, opened 2017-03-15
hcache error: Cannot allocate memory


viric:
* Bug reports
    * Expected behavior
hcache should work
    * Actual behavior
hcache fails. It runs a mmap() call with 2GB. Strace lines:
open("/home/viric/.hcache", O_RDWR|O_CREAT, 0644) = 4
mmap2(NULL, 2147483648, PROT_READ, MAP_SHARED, 4, 0) = -1 ENOMEM (Cannot allocate memory)
The .hcache file I am using is 8KB big. Erasing this file creates it anew for 8KB and fails the same way.
    * Steps to reproduce
Use the hcache in a system with 1GB of RAM.
    * Used program versions
NeoMutt 20170128 (1.7.2)
    * Operating System and its version
NixOS 17.03 on Raspberry Pi 2

viric:
The terminal has this text: `hcache_open_lmdb: mdb_env_open: Cannot allocate memoy`

guyzmo:
is it actually crashing?

viric:
It is not crashing.

flatcap:
Hmm... It may be because LMDB is trying to `mmap()` more memory than exists.
If you're building your own neomutt, try changing: `hcache-lmdb.c` line 31:
from
```c
const size_t LMDB_DB_SIZE = 2147483648;
```
to
```c
const size_t LMDB_DB_SIZE = 536870912;
```

gahr:
Perhaps I should make it configurable and use HeaderCachePageSize as gdbm does. Although the name would be a bit misleading in the case of lmdb, where the value is actually the max db size.

gahr:
Or maybe a better approach would be to handle the failure and reduce the map size by >> 1 until it succeeds.

gahr:
a0777aa should fix that. Unfortunately I cannot test it, as on my system mmap succeeds no matter what.

gahr:
@viric please feel free to reopen if you get around to test my patch.

PR #475, opened 2017-03-16
Try to decrease LMDB's db size if the default is too much

gahr:
Issue: #473

* What does this PR do?

This fixes an issue with the LMDB hcache backend not being able to allocate memory on some systems (Linux) if the [default db size](https://github.com/neomutt/neomutt/blob/master/hcache-lmdb.c#L31) is larger than available memory. The approach is to reduce the default until allocation succeeds.

### DIFF
--- a/hcache/lmdb.c
+++ b/hcache/lmdb.c
@@ -20,13 +20,14 @@
 
 #include "config.h"
 #include <stddef.h>
+#include <errno.h>
 #include <lmdb.h>
 #include "backend.h"
 #include "lib.h"
 
 /* The maximum size of the database file (2GiB).
  * The file is mmap(2)'d into memory. */
-const size_t LMDB_DB_SIZE = 2147483648;
+static size_t LMDB_DB_SIZE = 2147483648;
 
 enum mdb_txn_mode
 {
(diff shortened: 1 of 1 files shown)


gahr:
@viric Could you please test this branch and see whether it fixes the problem for you?

viric:
Give me some time. Thank you!

On Thu, Mar 16, 2017 at 03:14:44PM -0700, Richard Russon wrote:
> flatcap commented on this pull request.
> 
> While the changes look good, I'd rather wait for the OP to check that it actually fixes the problem :-)
> 
> 
> 
> -- 
> You are receiving this because you were mentioned.
> https://github.com/neomutt/neomutt/pull/475#pullrequestreview-27473396

-- 
(Escriu-me xifrat si saps PGP / Write ciphered if you know PGP)
PGP key D4831A8A - https://emailselfdefense.fsf.org/

viric:
`Reading /var/spool/mail/viric... 0/895 (0%)Segmentation fault`

That's it. :)

Stracing, it does not retry the mmap after the failed 2GB attempt.

hcache-lmdb.c
```diff
@@ -108,9 +109,18 @@ hcache_lmdb_open(const char *path)
     return NULL;
   }
 
-  mdb_env_set_mapsize(ctx->env, LMDB_DB_SIZE);
+  rc = ENOMEM;
+  while (LMDB_DB_SIZE)
+  {
+      mdb_env_set_mapsize(ctx->env, LMDB_DB_SIZE);
+      rc = mdb_env_open(ctx->env, path, MDB_NOSUBDIR, 0644);
```
viric: Are you sure this returns 'errno'?

flatcap: The LMDB docs for `mdb_env_open` suggest it will return E error numbers on failure and lists a couple (but not ENOMEM).  I might have some time, tonight, to do a little debugging (note neither @gahr nor I can reproduce the problem).

Let us know if you discover any other clues.

gahr: > @viric Are you sure this returns 'errno'?

[Yes](https://github.com/LMDB/lmdb/blob/mdb.master/libraries/liblmdb/mdb.c#L4358-L4363).

gahr:
> @viric Reading /var/spool/mail/viric... 0/895 (0%)Segmentation fault
That's it. :)
Stracing, it does not retry the mmap after the failed 2GB attempt.

Would you be able to provide some more info? Like, what does `mdb_env_open` return on failure?

toogley:
blocked by: additional info from @viric  needed.

flatcap:
Hi @viric 
We'd really like to know if this fixes your problem, so we can close this issue.

PR #476, opened 2017-03-17
Tidy the code - Part 2

flatcap:
Another batch of tidying.
All of the commits are quite simple ideas and shouldn't cause concern.
There's more detail in the commit logs.

- e4c9180 build: force OPS to be rebuild after a reconfigure
- c6b3eb2 build: fix minor code warnings / style
- 7b616ef tidy: be specific about void functions
- 8f1bad8 build: expand a few more alloc macros
- bda3628 tidy: add argument names to function prototypes
- c32c180 build: drop local copy of regex code
- 3d9acdb tidy: indent the docbook manual
- 7d2904a tidy: rearrange code to avoid forward declarations
- c929808 tidy: limit the scope of some functions
- 5e7abae build: give the compress functions a unique name
- ce83910 tidy: use snake_case for function names
- 390d16b build: add missing newlines to mutt_debug

I recommend looking at the commits individually, because there's a lot of overlap.


gahr:
I have two points I would like to discuss:

1. minor - I don't see any advantage in 51610dc. I personally prefer `int foo()` over `int foo(void)`. Sorry if this has been discussed before

2. medium - won't 9c3a55d make merges a nightmare? What's wrong with fwd decls?

flatcap:
> I personally prefer int foo() over int foo(void)

Ah, but there's a difference in C.  `int foo()` means the parameters are unspecified, so these would both be permitted by the compiler:
- `x = foo();`
- `x = foo(1, 2, 3);`
Adding "void", means specifically : no parameters allowed.

> won't [avoid forward declarations](https://github.com/neomutt/neomutt/commit/9c3a55d2f6dcd1747af4c5c9dfdcb697a28a3aa9) make merges a nightmare?

I'm not sure it'll be a problem.  (Or at least, the changes yet to come will be more problematic :-)
Because of the context of the diff, most changes should be correctly located.

> What's wrong with fwd decls?

Nothing particularly.  They're just extra work to maintain.
Some are required where two functions refer to each other -- there are a few instances of this in mutt.

In my own code, I tend to put all the static functions at the top of file (before their use).
I tried to sneak this idea into these tidying changes, but there are still a lot of large files that need to be split up first.

gahr:
>> @gahr I personally prefer int foo() over int foo(void)

> @flatcap Ah, but there's a difference in C.

My bad - I was reasoning c++ .

>> @gahr won't avoid forward declarations make merges a nightmare?

> @flatcap I'm not sure it'll be a problem. (Or at least, the changes yet to come will be more problematic :-)
Because of the context of the diff, most changes should be correctly located.

Alright, then!

compress.h
```diff
@@ -17,7 +17,7 @@
  */
 
 #ifndef _COMPRESS_H_
-#define _COMPRESS_H_
+#define _COMPRESS_H
```
d-k-c: Doesn't belong in this commit and is not in sync with the #ifndef just above

charset.c
```diff
@@ -311,7 +311,7 @@ int mutt_chscmp (const char *s, const char *chs)
 			     a > b ? chs : buffer, MIN(a,b));
 }
 
-char *mutt_get_default_charset ()
+char *mutt_get_default_charset (void)
```
d-k-c: Should be in another commit

flatcap:
@d-k-c 
charset.c : moved change to correct commit
compress.h : removed typo (I don't know what _that_ came from)

flatcap:
@neomutt/reviewers Any other comments?

PR #477, opened 2017-03-17
Get rid of a few more left overs from a long time ago

gahr:
* What does this PR do?

Get rid of a few m4 macros that are useless in a C99 world.


gahr:

Issue #277, opened 2016-12-29
Subject separate from tree?


iKarith:
At the moment, `$index_format` does not include a string for the tree characters--they're simply assumed to be part of the subject.  That's not always desirable if, say, you want to do something with %y or %Y ala Gmail.

Changing this, I realize, is somewhat complex because it alters established behavior.  Currently, `%s` includes the tree, if the tree would make sense to display.  You could create a new expando for the tree, but either this introduces the need to update configs if you wish the tree to be displayed, or you need to add a variable for whether or not the subject includes tree characters which only makes sense if you understand that it's there so that people need not update their configs for a new default.

The feature is desired all the same.  :)

flatcap:
We need examples...

Here's a normal conversation (from our devel mailing list):
```
Sven      Enter PGP passphrase
Elimar    └─>
Sven        └─>
Elimar        └─>
Guillaume       ├─>
Sven            └─>Re: Enter PGP passphrase -> GPGME
Elimar            └─>
```

Inserting labels, makes a mess of the tree:
```
Sven      LABEL Enter PGP passphrase
Elimar    LABEL └─>
Sven         └─>  
Elimar         └─>  
Guillaume        ├─>
Sven             └─>Re: Enter PGP passphrase -> GPGME
Elimar    LABEL         └─>
```

Giving them a fixed width helps:
```
Sven      LABEL    Enter PGP passphrase
Elimar    LABEL    └─>
Sven                 └─>
Elimar                 └─>
Guillaume                ├─>
Sven                     └─>Re: Enter PGP passphrase -> GPGME
Elimar    LABEL            └─>
```

Ideal situation, separate "tree", "label" and "subject" expandos:
```
Sven      LABEL Enter PGP passphrase
Elimar    └─>LABEL
Sven        └─>
Elimar        └─>
Guillaume       ├─>
Sven            └─>Re: Enter PGP passphrase -> GPGME
Elimar            └─>LABEL 
```

iKarith:
I'd say you have captured what I'm after.  I'm working on a relatively narrow screen by today's standards, barely 90 columns wide, so ideally I need to minimize the widths of everything to ensure that I can read date, sender, labels, and as much subject as I can fit.  Anything else in the index is mostly unnecessary--I mark things like unread/deleted/both with colors, for example, because it saves precious characters.

guiniol:
I have the same issue, and I think the visual solution I would go for is the second one:
```
[...]

PR #482, opened 2017-03-21
add %z format string to index_format

sassmann:
The %Z format string consists of three columns. Add %z which only
consists of 2 columns, sacrificing the middle column. The cost here is
to omit sign and crypto flags, but not everybody cares for that.
Delete flags have been merged into the first column.

Signed-off-by: Stefan Assmann <sassmann@kpanic.de>

* What does this PR do?
Add %z which only consists of 2 columns to index_format.

Check hdr_format_str()

* Why was this PR needed?
Allows to have 2 column message status flags.

* Screenshots (if relevant)

     Patch docs in init.h, should be enough?

   - [ ] All builds are passing
     How can I test this? Did a local build.

     Not required

      I sure hope so. :)

### DIFF
--- a/globals.h
+++ b/globals.h
@@ -233,6 +233,11 @@ WHERE short NewsPollTimeout;
 WHERE short NntpContext;
 #endif
 
+#ifdef DEBUG
+WHERE short DebugLevel;
+WHERE char *DebugFile;
+#endif
+
 WHERE short ConnectTimeout;
 WHERE short HistSize;
 WHERE short MenuContext;
--- a/hdrline.c
+++ b/hdrline.c
@@ -445,7 +445,11 @@ static char *apply_subject_mods(ENVELOPE *env)
  * %X = number of MIME attachments
  * %y = `x-label:' field (if present)
  * %Y = `x-label:' field (if present, tree unfolded, and != parent's x-label)
- * %Z = status flags    */
+ * %zs = message status flags
+ * %zc = message crypto flags
+ * %zt = message tag flags
+ * %Z = combined message flags
+ */
 static const char *hdr_format_str(char *dest, size_t destlen, size_t col, int cols,
                                   char op, const char *src, const char *prefix,
                                   const char *ifstring, const char *elsestring,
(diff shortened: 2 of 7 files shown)


somini:
I don't mean to badmouth your PR, but I think instead of adding yet another combination of flags to the index format, we should break down the existing ones so that everyone can be happy with its special `index_format` configuration.

toogley:
@somini what do you mean by "break down" ?

somini:
I mean splitting up flags that produce several characters at once into several flags that produce a character each. So that you can pick and choose different combinations and locations. The formats already support conditional formatting, so very complex formats can be created without touching C code. 
This Z flag is the most glaring example. For  an example of complex formatting, see the date thinghy in the site.

sassmann:
Hi Somini,
appreciate your feedback. It took me a moment to realize exactly what you mean by that. So if I get that right you're saying, split up those 3 columns from %Z in 3 separate flags. Sounds like a worthwhile consideration, few things that need to be considered.
- we're running out of single character flags and I'm not sure if we can have multi character flags like z1,z2,z3 for example without non-trivial changes.
- it will probably cause code duplication, which I tried to avoid, as we should not remove %Z for backwards compatibility

flatcap:
I like the idea, and if it works, there are other options that could be split the same way.

> we're running out of single character flags and I'm not sure if we can have multi character flags like z1,z2,z3 for example without non-trivial changes.

I think using descriptive letters for the suffixes might be easier to use / remember.

Adding a single letter/digit suffix to `%Z` _may_ not be too hard.
If you have the time to investigate, I'd really appreciate it.

> it will probably cause code duplication, which I tried to avoid

Hmm... possibly.

> as we should not remove `%Z` for backwards compatibility

indeed.

sassmann:
I've change the patch to have 3 single column format strings %zs %zc %zt.
Let me know what you think.

somini:
Sounds just like what I had in mind. 😊
I agree with keeping old flags too.

flatcap:
I like the changes, but I'm just hesitating from approving them...
I'm wondering about changing `%z` to something non-lettery, maybe `%!`

ATM, you're using `%z` as an extension of `%Z` -- all very logical.
However, I'd like to reuse this extension idea to fix #277 too.

What should we have as the extension character?

somini:
Most are already taken:

https://www.neomutt.org/guide/reference.html#index-format

I think `%!` or `%+` are our best bet.

flatcap:
Ooh, I like `%+` :-)

@neomutt/reviewers any thoughts?

gahr:
Would it be bad to enhance `%Z` to support `%Zs`, `%Zc`, and `%Zt` extensions? Of couse this would break settings where currently the `%Z` flag is surrounded by one of those three characters, literally, but would such a setting even make sense?

flatcap:
I did consider `%Z{letter}` and wonder how we'd distinguish from adjacent text.

It might be unlikely that `%Z` is followed by text, but if we extend the pattern to other expandos, that might not be the case.

Another possibility is shell style: `%{Zs}` -- whilst **very** flexible, it's going to be a lot more work to parse.

gahr:
We will get to the end of the alphabet quite soon, so `%{Mv}` for majors and variants might be worth pursuing while there's interest in adding something (i.e., now) ;)

flatcap:
> while there's interest in adding something (i.e., now)

Good point.

@sassmann Would you be interested in investigating/coding `%{Mx}` ?
If so, please put the code in a separate branch.
(as a member, you can create [devel/*] branches in the neomutt repo, if you wish).

somini:
That will break `%{fmt}`, for dates. We can always go for `%+Mx+`.

flatcap:
> That will break %{fmt}, for dates

Ah, yes.  So much for that idea.
Hmm... all the paired characters are in use.

sassmann:
Did I understand this right, you'd like to have %+Zs+, %+Zc+, %+Zt+ ?

flatcap:
> you'd like to have %+Zs+, %+Zc+, %+Zt+ ?

Hmm... that would work, but it doesn't look very nice.
Aesthetics are everything :-)

Let's keep thinking for now.

sassmann:
so, what about we reserve + for 2 letter format strings. That way we gain a whole lot of new format options and it doesn't look too crazy. %+Zs doesn't seem so bad, imho.

flatcap:
@sassmann 
> reserve + for 2 letter format strings
> `%+Zs`

That's not a bad idea...

@somini 
> With delimiters ... `%+unread+`

Yes, that's what I imagined in the future.

---

Thoughts...

If we started with `%+Zs` can we still extend it to `%+token+` in the future?
Not easily.  An existing `%+ap` would make it hard to parse `%+apple+`

Should we use `%+Zs` for now, then use `%!token!` in the future?

Do the formatting strings still work?  `%-3.3Zs`
Could any of the potential schemes cause problems with formatting?

Current `$index_format` has ~40 expandos.  Splitting up some of these will lead to some more.
When scripting arrives this number will explode.  Which format will give us enough scope (and remain legible)?

Are we stretching the printf string too far?
Ages ago, I mocked up a [json formatting string](https://github.com/neomutt/flatcap-brain-dump/blob/master/rich-format.json) -- should we consider this for the future?

somini:
> Should we use `%+Zs` for now, then use `%!token!` in the future?

I like this option. `man 3 printf` shows a conflict with the plus sign: `%+d` should produce a number including the plus even if the number is positive, but for `%!` I can't find conflicts.

Something like that JSON format, while unambiguous, is hassle to write and doesn't give the immediate feedback. That idea can be used if we write a parser that outputs the JSON and uses it to write the string. Very useful for regression testing too.
The performance on the index, for example, where we have hundreds of lines must be evaluated.

----

TL;DR: I vote for `%!token!`

sassmann:
Looks like there are no further comments.  I'll go ahead with the %+Zx options then and leave %|token| for somebody else to implement.

sassmann:
I've pushed a devel branch named devel/message_status_flags as flatcap suggested.

flatcap:
Thanks @sassmann

toogley:
blocked by https://github.com/neomutt/neomutt/issues/531

sassmann:
Toogley,
I'm confused, this feature is independent of #531. #531 is a project of its own and shouldn't block the merge of this request. Care to explain?

toogley:
@sassmann aah, i see. sorry, my fault.

flatcap:
Merged in d3c2d8f.  Thanks.

PR #488, opened 2017-03-26
Tidy the code - Part 3

flatcap:
Another batch of tidying.
There's more detail in the commit logs.

1. 5da2282 fix: typo in browser_compare_count{_new}
2. dde4b09 build: remove generated files from repo
3. fd4d386 tidy: comments
4. 328ccb4 build: look for translations in all files
5. aaa909e fix: arguments to printf-style functions
6. 1b11da2 tidy: license text
7. 7c57850 tidy: unify include-guards
8. f935a50 build: tidy makefiles
9. 2e0437e tidy: initialise pointers
10. 37087a6 tidy: make strcmp-like functions clearer
11. 6430f1b tidy: unify sizeof usage
12. a8dc8a6 build: remove forward declarations
13. 7462578 tidy: remove ()s from return
14. ~~fd18984 tidy: add ()s to return~~

I recommend looking at the commits individually, because there's a lot of overlap.


flatcap:
@guiniol 
Thanks.
I'm not too worried about fd18984 (adding ()s to return).  I was looking for simple ways to tidy up.
The real problem is code like this:
```c
return a1->flags & MUTT_ACCT_USER && a1->user[0] ? 0 : 1;
```

@d-k-c 
> This is not really idiomatic to add this trailing "1"

OK.  I don't mind dropping the "1".  Now that all the guards are unified, it's trivial to do.
Anyone else have a preference?

> #ifdef __GNUC__

Yeah, that's the easy bit.  I just couldn't decide where to put it.
There are quite a few `__attribute__`s I'd like to use, so they might warrant their own "compiler.h"

guiniol:
> The real problem is code like this:
> 
> ```C
> return a1->flags & MUTT_ACCT_USER && a1->user[0] ? 0 : 1;
> ```

Humm. That should probably be:
```C
return ((a1->flags & MUTT_ACCT_USER) && a1->user[0]) ? 0 : 1;
```
But that's more of a general problem than a `return` specific, no?

gahr:
I'm also not in favour of having return values surrounded by `()`.

flatcap:
> But that's more of a general problem than a return specific, no?

Yes.  I'm working on something, but there are a lot of false positives.

> I'm also not in favour of having return expressions surrounded by ().

OK, I'll drop them.
I'll have a think about splitting up all return statements containing conditionals.

gahr:
> I'll have a think about splitting up all return statements containing conditionals.

Aren't you perhaps going a bit too far? Are you considering changing

```c
return a > b ? c : d;
```
into
```c
int ret = a > b ? c : d;
return ret;
```
or
```c
if (a > b)
   return c;
else
   return d;
```
?

None provides any obvious advantage over the original single-line statement, imho. Of course a ten-lines boolean expression in a `return` statement is ugly, but be careful in making a rule out of it and generalizing it to _all return statements containing conditionals_.

flatcap:
> Aren't you perhaps going a bit too far?

Don't worry, I'm just thinking aloud...

There are quite a lot of ugly return statements.
I'm working parsing some of the unclear logic that would turn your first example into:
```c
return (a > b) ? c : d;
```

but there are still quite a lot which would be better off in the form:
```c
if (a > b)
   return c;
else
   return d;
```
But they would be manual tasks, so they're not happening any time soon.

flatcap:
I've removed the final commit (Add ()s to return).
Any other comments?

PR #490, opened 2017-03-27
Fix inbox-first functionality when using mutt_pretty_mailbox

### LINKED ISSUE #489: Inbox-first sorting broken after 2299e673c
* Expected behavior

The INBOX folder of an IMAP account should be sorted before all other folders when
`sort_browser = alpha`. This was achieved in #190 and #356, mostly.

* Actual behavior

After 2299e673c, the `proto://user@host` part of the `folder` variable is not displayed anymore, and the `=` sign is displayed instead. This breaks the inbox-first functionality, which relied on the fact that '/' was the path separator.

gahr:
Fixes #489


gahr:

PR #491, opened 2017-03-27
use w3m as default for generating UTF8 manual.txt

sassmann:
The input for manual.txt is manual.html which is UTF8 encoded. Now
forcing manual.txt to be ascii leads to all sorts of "?" in the ascii
file, which is not nice. If we have UTF8 input we should also use UTF8
for the output.
Moved w3m to be the default to generate UTF8 manual.txt. If you prefer
manual.txt to be ascii the fallbacks lynx or elinks will still produce
plain ascii.

Signed-off-by: Stefan Assmann <sassmann@kpanic.de>

   - [ ] All builds are passing

### DIFF
--- a/doc/Makefile.am
+++ b/doc/Makefile.am
@@ -127,8 +127,8 @@ uninstall-local:
 
 check:
 manual.txt: manual.html
-	-LC_ALL=C lynx -dump -nolist -with_backspaces -display_charset=us-ascii manual.html > $@ || \
-	LC_ALL=C w3m -dump manual.html > $@ || \
+	-LC_ALL=C w3m -dump -O UTF8 manual.html > $@ || \
+	LC_ALL=C lynx -dump -nolist -with_backspaces -display_charset=us-ascii manual.html > $@ || \
 	LC_ALL=C elinks -dump -no-numbering -no-references manual.html | sed -e 's,\\001, ,g' > $@
 
 Muttrc: stamp-doc-rc


flatcap:
TBH, I hadn't really **looked** at manual.txt before.
For testing purposes I made sure that I had all three of `lynx`, `w3m` and `elinks` installed, but I didn't compare the output.

Now I have, my results don't match yours.
I ran the three conversions and for comparison, also ran them in my locale en_GB.UTF-8.

```sh
LC_ALL=C lynx -dump -nolist -with_backspaces -display_charset=us-ascii manual.html    > mc-lynx.txt
LC_ALL=C w3m -dump manual.html                                                        > mc-w3m.txt
LC_ALL=C elinks -dump -no-numbering -no-references manual.html | sed -e 's,\\001, ,g' > mc-elinks.txt

lynx -dump -nolist -with_backspaces -display_charset=us-ascii manual.html             > m-lynx.txt
w3m -dump manual.html                                                                 > m-w3m.txt
elinks -dump -no-numbering -no-references manual.html | sed -e 's,\\001, ,g'          > m-elinks.txt
```

| Description    | Example   | C lynx   | lynx   | C w3m  | w3m    | C elinks | elinks   |
|----------------|-----------|----------|--------|--------|--------|----------|----------|
| Extended ASCII | Björn     | Bjoern   | Bjoern | Bj?rn  | Björn  | Bjoern   | Björn    |
| Quoted text    | \<quote\> | "Fefe"   | "Fefe" | ?Fefe? | “Fefe” | "Fefe"   | “Fefe”   |
| Unicode        | Hyphen    | --       | --     | ??     | —      | --       | —        |
| Unicode        | Copyright | (c)      | (c)    | ?      | ©      | (c)      | ©        |
| Unicode        | Dagger    | /-       | /-     | ?      | †      | /-       | †        |

On my machine w3m performed worse at converting to plain text.
If you allowed manual.txt to be utf-8, then w3m and elinks did a nice job.

If you want to **force** a particular rendering, without altering what's installed on your machine, you can cheat.  I found that putting a "fake" lynx in my path, meant that w3m was used:

 `ln -s /usr/bin/false lynx`

sassmann:
Nice comparison! Did you notice the -O UTF8 I added to the w3m call? That results in manual.txt being encoded in UTF8 instead of plain ascii. Therefore all UTF8 characters are rendered appropriately, same as when you used locale en_GB.UTF-8 in your example.

The input (manual.html) is UTF8 encoded already and nowadays most distros should be able to handle UTF8. For anybody who wants to stick with plain ascii there's still lynx as well as elinks.

flatcap:
> Did you notice the -O UTF8

Ah, sorry, I didn't.

Yes, I agree, it's reasonable to default to utf-8.

toogley:
@flatcap @sassmann Are there any more things needed to be fixed before this PR can be merged?

sassmann:
Good to go from my side.

PR #493, opened 2017-03-31
Tidy the code - Part 4

flatcap:
## THIS IS A WORK-IN-PROGRESS

- I'm still doing lots of checking.
- It may change without warning.

1. c878692 **shadow variables** — Eliminate variables that shadow other variables or function arguments
2. 20dfe85 **0 → NULL** — Eliminate one occasion where a pointer is compared to 0
3. 15fb09e **(!x) → (x == 0)** — Replace all tests on integers
4. 0f73a2a **(x) → (x != 0)** — Replace all tests on integers
5. 2ea4f41 **(!x) → (x == NULL)** — Replace all tests on pointers
6. d93b74e **(x) → (x != NULL)** — Replace all tests on pointers
7. 1b8e5ed **(x + y > z) → ((x + y) > z)** — Start bracketing complex logic
8. 663cb96 **logic1** — More bracketing of complex logic
9. 5c517e0 **logic2** — More bracketing of complex logic

@neomutt/reviewers Do you agree with the changes being made?

This is my final batch of tidying.
If anyone has any global changes they'd like to perform, speak now.
After this it will be time for clang-format.


gahr:
* It is pretty idiomatic to use `x` for `x != NULL` and `!x` for `x == NULL`, think for example a typical list traversal loop `for (tmp = head; tmp; tmp = tmp->next) {...}`. Does adding explicit NULL really add anything?

* I would not label `(x + y > z)` as *complex logic*. C has operators priorities, I am not sure surrounding any single binary expression with parenthesis adds any clarity. Personally, I find `x + 6 * y / 3` much clearer than `x + ((6 * y) / 3)`. Any accustomed C programmer can read that without having to stop and wonder whether + or * comes first. Same for binary addition and greater-than comparison in your original example.

Bottom line of both points is: do you really think that being so much more verbose helps anybody?

gahr:
I started reviewing the diff and I noticed that you turned many `if (foo)` into `if (foo != 0)`. When `foo` is meant as a boolean flag (be it bool, char, int, doesn't matter) the original version is much clearer in intent.

toogley:
I personally like things like

```c
  if ((fc->ibl == sizeof(fc->bufi)) ||
      (fc->ibl && (fc->ib + fc->ibl) < (fc->bufi + sizeof(fc->bufi))))
```

Of course, i can also understand what it does without the parenthesis. But with the parenthesis, i can read it faster.

gahr:
@toogley yes - the original code sucks a bit, but it's not because of parentheses. Would the following achieve exactly the same?

```c
if (fc->ibl <= sizeof(fc->bufi))
```

toogley:
@gahr yes.

flatcap:
Oh, for nested comments...

> It is pretty idiomatic to use `x` for `x != NULL` and `!x` for `x == NULL`

I agree, but at one of the early discussions about coding style, one of the participants expressed a dislike for `!` being used on pointers.

I don't mind dropping that commit.
What _was_ useful, was to identify where `(!x)` was being applied to an integer.

> `bad` is clearly used as a boolean flag here.
> The original version is much clearer than this one.
> etc

Yes.  In those examples the variable would be better off as a bool and the condition left as `(!x)`
The two commits do "successfully" distinguish between pointers and integers.

> IRC: perhaps boolification should come first

I made several attempts to automatically identify bools in the code, but none was successful.
Perhaps I can rework the integer commit into a bool commit.

> I would not label `(x + y > z)` as _complex_ logic

hehe, "complex" was probably too strong a word

This example with three single-letter operands is trivial, but that's not representative of most of the code.  Here's a simple example:

```c
while (curcol < win->col_offset + win->cols)
while (curcol < (win->col_offset + win->cols))
```

```c
if (tempbuf && templen == state->lastchar - i &&
if (tempbuf && templen == (state->lastchar - i) &&
```

In each, the `(` tells you that you're expecting more than one token.
There are quite a lot of `x < sizeof(TYPE) - 1` where the `-1` could easily be missed.

> C has operators precedence

Of course.  However, as a left-to-right-reading human, some of mutt's code is downright misleading.
Here are two examples from thread.c
```c
x = safe_malloc (width * depth + 2);
x = safe_malloc (2 + depth * width);
```

The first reads quite nicely.
The second means you have to re-read:
"two plus depth times, no, depth times width, [scan back to start] plus 2"

Parenthesising this leads to:
```c
x = safe_malloc (2 + (depth * width));
```

Note: my regex replacements didn't deal with this one.

> do you really think that being so much more verbose helps anybody?

**I** find them easier to parse with the parentheses.
Some of the mental effort has been done for me.

---

I'll investigate whether I can magically turn some integers into bools.

My personal preferences are:
- **keep** the **integer** replacements
- **drop** the **pointer** replacements.
- **keep** the **logic** replacements.

guiniol:
> It is pretty idiomatic to use `x` for `x != NULL` and `!x` for `x == NULL`

I'm with @gahr on that one.


> I don't mind dropping that commit.
> What was useful, was to identify where `(!x)` was being applied to an integer.
> 
>  >   bad is clearly used as a boolean flag here.
>  >  The original version is much clearer than this one.
>  > etc
> 
> Yes. In those examples the variable would be better off as a bool and the condition left as `(!x)`
> The two commits do "successfully" distinguish between pointers and integers.
> 
>  >   IRC: perhaps boolification should come first
> 
> I made several attempts to automatically identify bools in the code, but none was successful.
> Perhaps I can rework the integer commit into a bool commit.

That sounds like the best option to me. What was not successful in your attempts?

flatcap:
> > I made several attempts to automatically identify bools in the code, but none was successful.

> What was not successful in your attempts?

I separated the pointers from the integers by replacing all the tests with NULL and then parsing all the compiler warnings.

I hoped that I could do something similar with bools: replace all integer variable declarations with bool.
Unfortunately, the C compiler doesn't issue warnings for `bool b = 42;`
Nor does it complain for: `enum my_bool { true, false } mb = 42;`

flatcap:
It was a lovely dream...  I should create a new "too ambitious" issue tag :-)

toogley:
wouldn't it make sense to keep https://github.com/neomutt/neomutt/commit/c878692f19582eec26e5402c94bc7792e4399083 ?

flatcap:
Good spot.
When I realised this PR was too much, I put that commit into another batch.
It is now: 328e2e7

PR #494, opened 2017-04-05
Assorted small fixes

flatcap:
These are mostly simple fixes and tidy-ups.

1. 8cbfdee build: fix warnings
2. 25e474c add full neomutt version to log startup
3. 7811113 drop duplicate versions of the sidebar contrib
4. 6a24b74 drop comment from sidebar docs
5. 2882209 build: remove unused macros
6. 3037f73 build: use SEEK_SET, SEEK_CUR, SEEK_END
7. d170e60 bug: add missing break
- ~a292ff2 fix: avoid type overflow~
8. f3f80f4 build: remove constant code
9. 97fa5e5 bug: don't pass large object by value
10. dbeeb46 fix: use correct buffer size
11. 9ecfc17 fix: bug in uncolor for notmuch tag
12. e0f340a build: workaround slang warnings
13. ec2dfbc shadow variables
14. e00d707 0 -> NULL

The commit log explains a little more about each change.


pgppacket.c
```diff
@@ -119,7 +119,8 @@ unsigned char *pgp_read_packet (FILE * fp, size_t * len)
 	  perror ("fread");
 	  goto bail;
 	}
-	material += b + 192;
+	material += b;
+	material += 192;
```
gahr: I don't understand why this is needed

flatcap: `b` is un unsigned char which is read from outside of mutt.
If `b` is large, then `b + 192` will overflow before it's added to `material`.

An alternative solution would be to cast `b` to the type of `material` first, then add.

gahr: Uhm. I don't think so. The type of the expression `b + 192` is `int`.

flatcap: @gahr's right (of course :-) -- I'll drop this commit

flatcap:
Rebased; unnecessary commit dropped.
Any other comments?

PR #497, opened 2017-04-09
refactoring // cleanup of legacy files

toogley:
* What does this PR do?

    * remove files whose content is already included in other parts of the project or are otherwise legacy
        * [neomutt.github.io#33](https://github.com/neomutt/neomutt.github.io/pull/33) is part of this PR and includes the content of some legacy files.
    * don't generate vim-neomutt syntax file and instead store the complete file in git.
    * remove Makefile target to create pdf version of the manual, as it doesn't seem to be used at the moment.


flatcap:
In principle, this looks fine.

We just need to skim through the deleted docs and make sure there's nothing worth cut'n'pasting into the main guide.

When we're done, the Makefiles will need fixing: "make dist" and "make distcheck" reference the deleted docs.

toogley:
@flatcap i think i have fixed the Makefiles.

toogley:
@flatcap oooh, sorry. I missed those.

toogley:
Ah, i forgot to ask: What should we do with `UPDATING.kz ` ? I'm not sure if a merge into `UPDATING` is a good idea? or should we just delete it?

toogley:

toogley:
(forgot to clarify why `mutt_xtitle` has been removed)

toogley:
* forgot to remove also pdf creation routine from configure.ac
* squashed some commits together

toogley:
> further TODO for this PR:
>     * clean up removed files from Makefile.am

i think that was done in [this commit](https://github.com/neomutt/neomutt/pull/497/commits/4aac5a895afba74d2c218810b91d4149beb1c8d3) already.

PR #515, opened 2017-04-09
Fix Coverity Defects

flatcap:
Here's another batch of preventative fixes.
These address >80 Coverity defects.

They are grouped by type (and better viewed as separate commits):

1. 785959e **strfcpy**
    Replace strncpy with strfcpy to avoid unterminated strings.

2. **add variable** -- add tests to conditionals
    - a331218 **function args**
      Check parameters passed to a function

    - fa7443c **failed function**
      Some functions when they fail return -1.
      This return value is often used as an array index which would cause a buffer underrun.

    - f86fa08 **Context**
      The Context can become NULL without warning

    - 54ac444 **alloc/strdup**
      If alloc is passed 0, or strdup is passed NULL, they can return NULL

    - b6d67a8 **route through code**
      A possible route through the code leads to an invalid variable.

3. c1a457f **remove variable test**
    Remove a few simple tests where the condition is guaranteed to be true.
    There are a few cases where we already know the outcome -- usually due to a previous test.

4. 7287f31 **test functions**
    Test the return value of some functions that could fail.

5. 1af16f5 **tidy switches**
    - Factor out trivial fall throughs
    - Document any other fall throughs

6. 0298ca0 **unused variables**
    Remove unused variables.

7. d390b7e **refactor only**
    Refactor code for clarity -- no functional change.

8. a6dc284 **check for buffer underruns**
    Prevent buffer underruns caused by functions that can return -1.
    Some functions that _could_ return -1 are used as array indices.
    Factor out the function and test the result first.

9. e65462d **fix leaks**
    Fix resource leaks.

10. c429f39 **minor fixes**
    Fix a few minor bugs.

With these changes, the number of defects will drop from 174 to 88.

### DIFF
--- a/addrbook.c
+++ b/addrbook.c
@@ -214,6 +214,8 @@ new_aliases:
         break;
       case OP_GENERIC_SELECT_ENTRY:
         t = menu->current;
+	done = 1;
+	break;
       case OP_EXIT:
 	done = 1;
 	break;
--- a/addrbook.c
+++ b/addrbook.c
@@ -164,6 +164,8 @@ new_aliases:
 
   safe_realloc (&AliasTable, menu->max * sizeof (ALIAS *));
   menu->data = AliasTable;
+  if (!AliasTable)
+    return;
 
   for (i = omax, aliasp = aliases; aliasp; aliasp = aliasp->next, i++)
   {
--- a/alias.c
+++ b/alias.c
@@ -285,7 +285,7 @@ retry_name:
 
   mutt_addrlist_to_local (adr);
 
-  if (adr)
+  if (adr && adr->mailbox)
     strfcpy (buf, adr->mailbox, sizeof (buf));
   else
     buf[0] = 0;
--- a/attach.c
+++ b/attach.c
@@ -729,7 +729,8 @@ int mutt_save_attachment (FILE *fp, BODY *m, char *path, int flags, HEADER *hdr)
       hn->msgno = hdr->msgno; /* required for MH/maildir */
       hn->read = true;
 
-      fseeko (fp, m->offset, SEEK_SET);
+      if (fseeko (fp, m->offset, SEEK_SET) < 0)
+        return -1;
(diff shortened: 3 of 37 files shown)


gahr:
2. fd0c5a7 add variable test

Perhaps here we should differentiate between three cases:
1. things that cannot possibly be `NULL` - e.g., return from `safe_realloc` - should not be checked, as checking them only introduces uncertainty about whether they can or cannot actually be NULL
2. things that must not be `NULL` by contract - e.g., parameter to some function - should be `assert`ed.
3. things that can legitimately be NULL should be checked. I guess there are not many occasions where a NULL is a sensible value for parameters. Please do not add complexity to handle programmers errors. For example, `mutt_buffer_read` returns a `void` and guarantees the input parameter `buf` parameter. I think this is a clear case for point 2.

flatcap:
The errors I've "fixed" have been raised by Coverity.  In each case it has found a path through the code that (however unlikely) could cause a NULL-dereference.

I'll see what I can do to separate out the cases.

gahr:
@flatcap I'm not sure coverity can figure out that e.g., a return from safe_malloc cannot be NULL.

darnir:
> @flatcap I'm not sure coverity can figure out that e.g., a return from safe_malloc cannot be NULL.

You can either add a coverity modelling file or use in-code comments to let coverity know of such scenarios. Though, I've used `safe_malloc` like functions in the past and coverity almost always realises what the invariants there are.

flatcap:
There's one slight wrinkle, though...  both these cases return NULL (and don't `exit(1);`)
- `safe_malloc(0);`
- `safe_remalloc(ptr, 0);`

Coverity spotted that the variables passed in _could_ be zero.

I've just categorised the 40 changes in fd0c5a7, I'll split them up shortly.

flatcap:
@darnir Thanks, I'll look into the modelling files.

flatcap:
I've split up the "add variable test" commit and updated the original comment.
Hopefully they're much clearer now.

PR #521, opened 2017-04-13
browser: small refactoring of %f of folder_format

toogley:
* What does this PR do?

As i found the code to process the `%f' char in folder_format_str a bit unreadable, i refactored it a bit. This version improves the readablity, in my opinion slightly.


gahr:
how about this?

```c
folder->ff->local ? (S_ISLNK (folder->ff->mode)   ? "@" :
                     S_ISDIR (folder->ff->mode)   ? "/" :
                     (folder->ff->mode & S_IXUSR) ? "*" : ""
                    ) : "";
```

flatcap:
@gahr, if we're heading for automatic clang-formatting, then any simple rearrangement will be lost.

clang-format renders that line (rather verbosely) as:
```c
snprintf(fn, sizeof(fn), "%s%s", s,
         folder->ff->local ?
             (S_ISLNK(folder->ff->mode) ?
                  "@" :
                  (S_ISDIR(folder->ff->mode) ?
                       "/" :
                       ((folder->ff->mode & S_IXUSR) != 0 ? "*" : ""))) :
             "");
```
@toogley is that any better?

gahr:
Ah true! Yeah, the clang version is better than both the original code and the proposed change, imo.

toogley:
@flatcap the clang-format version is of course better than the original code, but i still like mine a bit more. But maybe this is also a because i know the if style far longer than the `() ? true : false` version.

gahr:
If we want to go with your version, please could you at least turn `else { if (...) {...} }` into `else if (...) { ...}` ?

toogley:
rebased onto master.

gahr:
I still prefer the clang-format'd version, and I thought you were going to address [my comment](https://github.com/neomutt/neomutt/pull/521#issuecomment-293866417). Weren't you?

gahr:
Alternatively, if you really can't stand nested `expr ? trueval : falseval;`:

```diff
diff --git a/browser.c b/browser.c
index a9f485ed..8bca2b88 100644
--- a/browser.c
+++ b/browser.c
@@ -303,14 +303,14 @@ static const char *folder_format_str(char *dest, size_t destlen, size_t col, int
 #endif
         s = NONULL(folder->ff->name);

-      snprintf(fn, sizeof(fn), "%s%s", s,
-               folder->ff->local ?
-                   (S_ISLNK(folder->ff->mode) ?
-                        "@" :
-                        (S_ISDIR(folder->ff->mode) ?
-                             "/" :
-                             ((folder->ff->mode & S_IXUSR) != 0 ? "*" : ""))) :
-                   "");
+      char c = '';
+      if (S_ISLINK(folder->ff->mode))
+        c = '@';
+      else if (S_ISDIR(folder->ff->mode))
+        c = '/';
+      else if ((folder->ff->mode & S_IXUSR) != 0)
+        c = '*';
+      snprintf(fn, sizeof(fn), "%s%c", s, c);

       mutt_format_s(dest, destlen, fmt, fn);
       break;
```

flatcap:
I dislike the nested ternary operators throughout mutt, but this example isn't one of the bad ones.
Now that clang-format has reformatted it, it's fairly legible.

In this case, I think we should leave it as it is.
(this from the coder who just devastated the code :-)

toogley:
Hm, okay. Well then I'll close this.

PR #527, opened 2017-04-15
Add guix-neomutt.scm.

ng-0:
This is used to build neomutt with guix in the git checkout.

* Added Comments:

Currently this needs to be in the top directory, a fix to support the "contrib" folder as a location is being worked on.

### DIFF
--- a/guix-neomutt.scm
+++ b/guix-neomutt.scm
@@ -1,0 +1,49 @@
+;;; Copyright (C) 2017 ng0 <ng0@no-reply.pragmatique.xyz>
+;;;
+;;; This program is free software: you can redistribute it and/or modify it under
+;;; the terms of the GNU General Public License as published by the Free Software
+;;; Foundation, either version 2 of the License, or (at your option) any later
+;;; version.
+;;;
+;;; This program is distributed in the hope that it will be useful, but WITHOUT
+;;; ANY WARRANTY; without even the implied warranty of MERCHANTABILITY or FITNESS
+;;; FOR A PARTICULAR PURPOSE.  See the GNU General Public License for more
+;;; details.
+;;;
+;;; You should have received a copy of the GNU General Public License along with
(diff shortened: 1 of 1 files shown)


d-k-c:
Sorry I am not familiar with Guix, but why does it need to be in the top directory, and not in the contrib folder as comment on line 18 may suggest?

EmilePouget:
I'm having difficulties figuring out the correct way to walk the path up from current-workdir (it's not just "..") and then take that directory as a source.
If you move it in the current form into contrib and then do "guix build -f contrib/guix-neomutt.scm", the "contrib" is taken as the sourcedir and attempts to build the content of contrib only.

If it's bad practice to put such a file into the top directory for you, I will take this question to either guile-users or guix-help list and then send an updated patch.

d-k-c:
As there is already a neomutt package in Guix, isn't there a way to add a neomutt-git package ? I know some distros like ArchLinux do that. But if something prevents you from doing that and this file needs to be in neomutt's tree, it would definitely be better if it could be put in contrib. So yes, it would be appreciated if you could take this question to guile-users and guix-help ML.

EmilePouget:
No, Guix master does (at this moment in time) run the policy to only use git when it can not be avoided, one such example is that the software developers will not produce release tarballs/snapshots.

The file I'm trying to introduce is for developers of neomutt with the intention to just build whatever you have worked on, but run it reproducible with Guix.
In GNUnet, for the software "gnunet", we did the same with Nix and Guix files (or at least we are trying, relations are more complex there and my maintenance is mostly outside of the gnunet.git tree to improve the procedure.. again, complex, more than just one software and not applicable to neomutt).

As you can see [here](https://gnunet.org/git/gnunet.git/tree/) they live in the top dir, as they can be regarded as extentions to the Makefiles, enabling Nix and Guix on all supported systems including but not limited to their own System Distributions.

ng-0:
obsolete by #723

Issue #389, opened 2017-02-10
remove configuration flags which don't depend on external dependencies


toogley:
@d-k-c  said in [issue #310](https://github.com/neomutt/neomutt/issues/310#issuecomment-275164875):

> Remove some configuration flags and their corresponding ifdefs. I think that options that don't depend on external dependencies should be built-in: smtp, imap, pop, debug, sidebar, compressed, etc.

toogley:
This issue was meant as a tracking issue, because IMHO its one of the easiest things to realize (compared to e.g. removing global variables or so).

Does anyone disagree with this change?

gahr:
Sounds reasonable, as long as default behaviour doesn't change. Example, if --enable-sidebar currently defaults to off and we remove the autoconf switch, I'd like to have sidebar_visible default to no.

somini:
The debug configuration flag should remain in theory, but I'm not sure it's worth it to keep non--debug builds.

toogley:
@somini good point. What exactly does the debug flag do? does it more than enabling the -D option?

gahr:
... most of which are in regex.c.. (I can't... resist... the urge... to killl...)

toogley:
@flatcap i don't understand why you've deleted all mutt_debug() operations. In my understanding, mutt_debug() did just write to `.muttdebug0` - which isn't harming anything IMHO. so i don't understand why you removed the additional debugging information completely instead of just removing the #ifdef DEBUG.

flatcap:
@toogley That was **just a bit of fun**.  I WILL NOT COMMIT THAT BRANCH.

I was just curious to see what DEBUG actually did -- I'm not sure I've ever asked that question of a large project before.

I figured that the simplest way to see the `#ifdef`s and `mutt_debug()`s was to delete them and look at the diff.

toogley:
@flatcap aaah, okay. 👍

xdgc:
* On 14 Feb 2017, Tobias Angele wrote: 
> @somini good point. What exactly does the debug flag do? does it more than enabling the -D option?

The theory is that a non-debug version is more performant, because
dprint()s are never compiled.  I'm sure this is true on older systems.
I'm not sure it's true on anything modern.

I'm not sure why it's desirable to remove either, though.

-- 
David Champion • dgc@bikeshed.us

gahr:
> I'm not sure why it's desirable to remove either, though.

I agree with this.

flatcap:
[...]

PR #528, opened 2017-04-15
make sidebar persistent

toogley:
* **What does this PR do?** it makes the sidebar persistent in every compilation, as it doesn't rely on external libraries. 

* **Are there points in the code the reviewer needs to double check?** `configure.ac`, because i didn't knew where i should place the line to include `sidebar.o`.

* **Why was this PR needed?** This reduces the amount of builds we have to run for travis and coverity, which is desired because those services limit the amount of builds per week.

* **What are the relevant issue numbers?** #389

### DIFF
--- a/Makefile.am
+++ b/Makefile.am
@@ -8,7 +8,7 @@ EXTRA_PROGRAMS = mutt_dotlock pgpring pgpewrap mutt_md5
 # Test the .tar file by building everything
 AM_DISTCHECK_CONFIGURE_FLAGS = --enable-compressed --enable-debug \
 	--enable-gpgme --enable-imap --enable-nntp --enable-notmuch \
-	--enable-pop --enable-sidebar --enable-smtp --with-bdb --with-gdbm \
+	--enable-pop --enable-smtp --with-bdb --with-gdbm \
 	--with-gnutls --with-gss --with-kyotocabinet --with-lmdb --with-qdbm \
 	--with-sasl --with-tokyocabinet
 
@@ -46,7 +46,7 @@ mutt_SOURCES = addrbook.c alias.c ascii.c attach.c base64.c browser.c buffer.c \
 	mutt_idna.c mutt_sasl_plain.c mx.c pager.c parse.c pattern.c \
 	postpone.c query.c recvattach.c recvcmd.c rfc1524.c rfc2047.c \
 	rfc2231.c rfc3676.c rfc822.c safe_asprintf.c score.c send.c sendlib.c \
-	signal.c sort.c status.c system.c thread.c url.c version.c
+	sidebar.c signal.c sort.c status.c system.c thread.c url.c version.c
 
 nodist_mutt_SOURCES = $(BUILT_SOURCES)
 
@@ -68,7 +68,7 @@ EXTRA_mutt_SOURCES = account.c bcache.c bcache.h browser.h compress.c \
 	mutt_lua.c mutt_sasl.c mutt_socket.c mutt_ssl.c mutt_ssl_gnutls.c \
 	mutt_tunnel.c newsrc.c nntp.c pgp.c pgpinvoke.c pgpkey.c pgplib.c \
 	pgpmicalg.c pgppacket.c pop.c pop_auth.c pop_lib.c remailer.c \
-	remailer.h resize.c sha1.c sidebar.c smime.c smtp.c url.h utf8.c \
+	remailer.h resize.c sha1.c smime.c smtp.c url.h utf8.c \
 	wcwidth.c
 
 EXTRA_DIST = account.h ascii.h attach.h browser.h buffer.h buffy.h ChangeLog \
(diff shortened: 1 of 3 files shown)


toogley:
Hm. interesting, i had checked my commits locally (and they build fine).

d-k-c:
Nice :) overall looks good to me, but I'll review and test it more thorougly tomorrow

toogley:
(i think i need to fix the manual, there are suggestions in the ifdef part to load the sidebar settings, if USE_SIDEBAR is defined.)

flatcap:
Some thoughts:

- Do we want to remove the `USE_SIDEBAR` markers yet?  (defined without a config option)
  ATM, the symbol shows us all the places that the sidebar touches.  It would be nice to reduce the sidebar's impact.  This would be easier if the symbols were still there.

- Do we want to remove "patch-sidebar" from `PATCHES` (and `USE_SIDEBAR` from version.c)
  Even though it's compiled-in, perhaps we should tell the user that the 'patch' is still there.

toogley:
hm, additionally: At the moment, this PR breaks configs which have lines in it, such as:


    ifndef USE_SIDEBAR finish

because `USE_SIDEBAR` is undefined. I think it would be a better idea to remove the `USE_something` symbols completely in one commit (if we have finished the work on #389) to limit problems/error messages user experience.

d-k-c:
We can definitely keep the USE_SIDEBAR. That would help us identify parts of the code that need to be reworked to have a better integration with the rest of the code, and that would also make this PR smaller and more focused on the changes in configure/Makefile.

On the code itself: there are some stuff that should be cleaned up in configure.ac: the help string, the summary of build options, etc. And I think the way to compile files that are always built-in is to add them to the mutt_SOURCES list in Makefile.am

toogley:
@d-k-c I addressed your wishes in my commit. Is there anything I've forgotten?

Furthermore: Should we merge these parts of #389 all together at once? (i mean, to be all at the same time released?)

d-k-c:
@toogley Are you sure sidebar.c is actually compiled in this new version? ^^

toogley:
@d-k-c Aaah. Sorry. :D I was getting erros and just cared about fixing them. But i forgot to actually control the result. So this new version is my "best attempt" - it just fails due to errors like [these](https://travis-ci.org/neomutt/neomutt/builds/223187970#L520-L522):

```
sidebar.c: In function ‘cb_format_str’:
sidebar.c:235:43: error: ‘SidebarWidth’ undeclared (first use in this function)
     mutt_FormatString(dest, destlen, col, SidebarWidth, ifstring, cb_format_str,
                                                                 ^
```

I also know that this happens because the header file wasn't included. Or rather, gcc can't find `sidebar.h` or so. But i don't know how i can solve that here. Also [Sources](https://www.gnu.org/software/automake/manual/html_node/Sources.html#Sources) and [headers](https://www.gnu.org/software/automake/manual/html_node/Headers.html#Headers) from the automake manual are not really helping here - although they seem to cover this topic.

Makefile.am
```diff
@@ -47,7 +47,7 @@ mutt_SOURCES = addrbook.c alias.c ascii.c attach.c base64.c browser.c buffer.c \
 	mutt_idna.c mutt_sasl_plain.c mx.c pager.c parse.c pattern.c \
 	postpone.c query.c recvattach.c recvcmd.c rfc1524.c rfc2047.c \
 	rfc2231.c rfc3676.c rfc822.c safe_asprintf.c score.c send.c sendlib.c \
-	signal.c sort.c status.c system.c thread.c url.c version.c
+	signal.c sort.c status.c system.c thread.c url.c version.c sidebar.c
```
d-k-c: Please keep that list alphabetically sorted

configure.ac
```diff
@@ -281,13 +275,6 @@ AS_IF([test x$use_smime != "xno"], [
 	SMIMEAUX_TARGET="smime_keys"
 ])
 
-dnl --enable-sidebar
-AS_IF([test x$use_sidebar = "xyes"], [
-	AC_DEFINE(USE_SIDEBAR, 1, [Define if you want support for the sidebar.])
```
d-k-c: By removing this block, you also removed the corresponding #define USE_SIDEBAR 1 in config.h, that's why you get the errors you mentionned, because some symbols (like SidebarWidth) are defined only if USE_SIDEBAR is defined. I'll push a fixup commit, feel free to squash it in your own commit or even move that line if it makes more sense somewhere else.

toogley: @d-k-c Thanks for the explanation.

Makefile.am
toogley: fixed.

toogley:
Rebased onto master. i'm wondering, would it improve the power of static analysis if we'd define USE_SIDEBAR 1 in the code and not the Makefile?

flatcap:
> would it improve the power of static analysis if we'd define USE_SIDEBAR 1 in the code and not the Makefile? (Because then, static analysis can figure out that sidebar is always compiled in)

I don't think moving the define will make a difference here, because I enable it in the build and sidebar's #defines don't have any #else clauses.

However, because the GnuTLS code is enabled, the OpenSSL code isn't tested at all.

toogley:
> However, because the GnuTLS code is enabled, the OpenSSL code isn't tested at all.

How does the sidebar  relate to TLS/SSL ?

flatcap:
I've just move the OPS back into OPS.SIDEBAR for now.
I'm considering changing the format of the OPS file.

d-k-c:
Could you give more info on changing fthe ormat of the OPS file? I think OPS files should be dropped. I understand why there were here before, because it was easier to keep the enum and the strings in sync, but now that Mutt uses C99 we can use the enum values as array indexes of `HelpStrings` and get rid of keymap*.h generation.

Issue #533, opened 2017-04-17
from_chars strange effect


nobrowser:
I'm the original author of the from_chars patch, so I think I know how it should work.

In my `.muttrc` I have:

`from_chars=">+-\r"`

In mutt.org mutt plus my patch, this adds a `>` prefix to the `%F` column of mails I sent. (I'm one of those people who like to have sent mails in the inbox.)  It leaves untouched index entries of mails I received.

But in neomutt, it prefixes the `%F` column of received mails with a question mark.  This feels as if some code, perhaps related to i18n or wide chars, misinterprets the `\r`.

I'll eventually assign myself after a few days if nobody takes it.

nobrowser:
The problem is that there is confusion about whether the configuration variable `Fromchars` is a plain string, or a multi-byte character table.

I remember when @flatcap initially merged my patch, he said that he modified it such that multi-byte character could be used for the prefixes.  And at present in `HEAD`, the variable is declared in `init.h` as a `DT_MBCHARTBL`.  But at the point where it is used -- namely in the function `make_from_prefix` in `hdrline.c` -- it is used very much like a plain string.

I tried to change the code in `make_from_prefix`, but I don't know how to convert a multi-byte character with `snprintf` to get it into a normal string buffer.  Anyone able to lend a hand/brain?

nobrowser:
How is [this](https://github.com/neomutt/neomutt/blob/master/init.c#L287) not completely broken?  Where is `d` getting set?  (It _should_ be the first argument to `mbrtowc`, methinks, instead of `NULL`.)

d-k-c:
Well, it seems this piece of code is only interested in determining how many bytes are in a multi-byte char, but the value itself is copied the old-fashion way (L301-303). As for `d`, it's allocated line 284, and set line 302. Does that answer your question or I'm completely off?

nobrowser:
@d-k-c you do answer my question.

The confusion was mine: I assumed that function was decoding into wide chars (UCS-4) while it apparently just copies the encoded representation (UTF-8 or whatever).

flatcap:
I've just had a chance to look at the code.
We started with @nobrowser's patch, which I tweaked to allow multi-byte chars.  Everyone was happy :-)
Then upstream adopted it and extended it to flag_chars, from_chars, status_chars, to_chars.

There are two problems:
- `\r` isn't magic any more -- a small tweak to `parse_mbchar_table()` will fix that
[...]

PR #534, opened 2017-04-18
Fixed issue #533: from_chars redux

nobrowser:
* What does this PR do?

There are only two (very small) real changes.

1. Previously `get_nth_wchar` checked for `\n` as the special placeholder value, but the documented and expected behavior for the `Fromchars` configuration item is that the special value is `\r`. I checked that the `Fromchars` code is the only user of  `get_nth_char` which uses the special value at all, so this should be OK.

2. Previously `make_from_prefix` used direct indexing to get the prefix value, I replaced that with a call to `get_nth_wchar` as explained above.

* Why was this PR needed?

Without these changes the `Fromchars` feature was broken.

   - [ ] Documentation created/updated - in fact this PR brings the code into correspondence with the doc :-)

   - [ ] All builds are passing - yes

   - [ ] Code follows the style guide - yes

#533

### DIFF
--- a/hdrline.c
+++ b/hdrline.c
@@ -156,6 +156,27 @@ enum FieldType
 };
 
 /**
+ * get_nth_wchar - Extract one char from a multi-byte table
+ * @table:  Multi-byte table
+ * @index:  Select this character
+ * @return: String pointer to the character
+ *
+ * Extract one multi-byte character from a string table.
+ * If the index is invalid, then a space character will be returned.
+ * If the character selected is '\n' (Ctrl-M), then "" will be returned.
+ */
+static char *get_nth_wchar(mbchar_table *table, int index)
+{
+  if (!table || !table->chars || (index < 0) || (index >= table->len))
+    return " ";
+
+  if (table->chars[index][0] == '\r')
+    return "";
+
+  return table->chars[index];
+}
+
+/**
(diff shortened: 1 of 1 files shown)


flatcap:
clang-format & rebase

nobrowser:
@flatcap, is that a hint for me?  Do you see things that are obviously misformatted?

flatcap:
Ah, that's what I've done.  There was one `char* var` and one line indented too much.

nobrowser:
Thanks for doing that.  I'm in the middle of installing clang so I can do it next time.

hdrline.c
```diff
@@ -156,6 +156,27 @@ enum FieldType
...
+ * Extract one multi-byte character from a string table.
+ * If the index is invalid, then a space character will be returned.
+ * If the character selected is '\n' (Ctrl-M), then "" will be returned.
+ */
+static char *get_nth_wchar(mbchar_table *table, int index)
+{
+  if (!table || !table->chars || (index < 0) || (index >= table->len))
+    return " ";
+
+  if (table->chars[index][0] == '\r')
```
gahr: Should the `\r` read `\n` perhaps?

gahr: Ah no sorry, I just read your comment regarding this. Disregard.

flatcap:
Fixed by 98d4025

PR #536, opened 2017-04-19
Remove -Iimap from CFLAGS and include imap/imap.h explicitely

gahr:


gahr:

PR #537, opened 2017-04-19
change 1/0 int parameter to boolean

toogley:
* **What does this PR do?** 
    * the `overwrite` parameter only holds `1`/ `0`
    * and has a boolean meaning in the context

Therefore i think changing the type makes sense.


toogley:
@d-k-c okay. i can fix that in this PR.

PR #539, opened 2017-04-20
Add configure options for address and undefined behaviour sanitizers

darnir:
* What does this PR do?

Adds 2 new configure flags to enable compile time instrumentation code. This helps check at runtime for some bugs

This should be used by the developers locally on their system. Since it will slow down the binary and any issues found will cause the binary to crash.

None

* Why was this PR needed?

Checking for undefined behaviour and address violations helps make the Mutt codebase more robust and secure

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -42,6 +42,9 @@ fi
 
 CFLAGS="-Wall -pedantic $CFLAGS"
 
+# Non-verbose make
+m4_ifdef([AM_SILENT_RULES], [AM_SILENT_RULES([yes])])
+
 AC_PROG_CPP
 AC_PROG_MAKE_SET
 AC_PROG_INSTALL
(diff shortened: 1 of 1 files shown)


toogley:
Something is really odd here. If i do ``, your branch fails, but current master not.

**current master:** (`./prepare --enable-everything --without-tokyocabinet --without-qdbm | grep kclangc.h`)

```
checking kclangc.h usability... yes
checking kclangc.h presence... yes
checking for kclangc.h... yes
```

**your add-sanitizers branch:** (`./prepare --enable-fsanitize-asan --enable-everything --without-tokyocabinet --without-qdbm | grep kclangc.h`)2

```
checking kclangc.h presence... yes
checking kclangc.h usability... no
configure: WARNING: kclangc.h: present but cannot be compiled
configure: WARNING: kclangc.h:     check for missing prerequisite headers?
configure: WARNING: kclangc.h: see the Autoconf documentation
configure: WARNING: kclangc.h:     section "Present But Cannot Be Compiled"
configure: WARNING: kclangc.h: proceeding with the compiler's result
checking for kclangc.h... no
```
I'm not sure if that's expected?

Anyways, this reveals at least that our autoconf config is legacy and should be updatet. ([Source](https://www.gnu.org/software/autoconf/manual/autoconf-2.67/html_node/Present-But-Cannot-Be-Compiled.html))

configure.ac
```diff
@@ -888,6 +892,22 @@ if test x$GCC = xyes && test $mutt_cv_warnings = yes; then
...
+  [gl_cc_sanitize_asan=yes], [gl_cc_sanitize_asan=no])
+
+if test "$gl_cc_sanitize_ubsan" = yes; then
+  CFLAGS="$CFLAGS -fsanitize=undefined -fno-sanitize-recover=undefined"
+fi
+
+if test "$gl_cc_sanitize_asan" = yes; then
+  CFLAGS="$CFLAGS -fsanitize=address -fsanitize-memory-track-origins"
+fi
+
```
flatcap: As these options are 'off-by-default' and don't have anything in the 'else' clause, can we abbreviate them -- move the CFLAGS into the AC_ARG_ENABLE?
```
AC_ARG_ENABLE([fsanitize-ubsan],
    [AS_HELP_STRING([--enable-fsanitize-ubsan], [Turn on Undefined Behavior Sanitizer (for developers)])],
    [CFLAGS="$CFLAGS -fsanitize=undefined -fno-sanitize-recover=undefined"])
```

gahr:
At this point, why the `autoconf` flag at all? Isn't the ability to specify `CFLAGS` at configure time enough?

darnir:
No. The reason they are separate is for future extensions. Not all of the sanitizers can be run together. So, we want to be able to identify fsanitize options that don't play well with each other and prevent the user from compiling the binary at all. The way these options are currently split, it is possible to write mutual exclusivity tests for various options.

gahr:
Not sure this belongs to NeoMutt's `configure.ac`. Using the flag assumes knowledge of the flags - or why would you enable them anyway? - so I still don't see much value. At least, I don't see much more value than providing flags for the different `-W` levels and incompatibilities.

toogley:
regarding my comment above:

The actual reason is that my gcc doesn't understand `-fsanitize-memory-track-origins`

from my `config.log` (after rebasing locally onto master)
```
configure:10075: checking kclangc.h usability
configure:10075: gcc -c -Wall -pedantic -Wno-long-long -g -O2 -fno-delete-null-pointer-checks -fsanitize=undefined -fno-sanitize-recover=undefined -fsanitize=address -fsanitize-memory-track-origins  conftest.c >&5
gcc: error: unrecognized command line option '-fsanitize-memory-track-origins'; did you mean '-fsanitize-recover=address'?
configure:10075: $? = 1
```

flatcap:
Thanks for the PR @darnir.
Rather than add all possible flags to the build, we've added `EXTRA_CFLAGS` and `EXTRA_LDFLAGS`.  These allow you to simply add any custom options to the build.

PR #547, opened 2017-04-24
configure: search by default for all cache backends

d-k-c:
All header cache backends are now scanned, even if not selected by `--with-<backend>`, and built if found. This behaviour can be disabled by adding `--without-<backend>` to configure options. If a backend is "force-selected" with `--with-<backend>`, a failure will be triggered if this backend cannot be found.

I'm posting this PR to open the discussion, and will be happy to rework it if this behaviour is not wanted. Usually, packagers are against automagic dependencies [1], but as backend scanning can be disabled by `--without-<backend>`, we are not in this situation.

Maybe another option as @gahr suggested, like `--with-all-hcache`, would be preferable. Or alternatively, do the auto-scanning but instead of automatically adding found backends to the build, report them at the end of the configure as being available like this:

```
Header Cache(s):   lmdb bdb [gdbm] ([backend] selected for build)
```

or

```
Header Cache(s):   gdbm
Other available Cache(s): lmdb bdb
```

Do we want that? What do you prefer?

[1] https://wiki.gentoo.org/wiki/Project:Quality_Assurance/Automagic_dependencies

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -622,38 +622,44 @@ AC_ARG_WITH(gdbm,
 	AS_HELP_STRING(
 		[--with-gdbm@<:@=DIR@:>@],
 		[Use gdbm for the header cache]),
-		[hcache_gdbm=$withval])
+		[hcache_gdbm=$withval],
+		[hcache_gdbm=check])
 AC_ARG_WITH(tokyocabinet,
 	AS_HELP_STRING(
 		[--with-tokyocabinet@<:@=DIR@:>@],
 		[Use tokyocabinet for the header cache]),
-		[hcache_tokyocabinet=$withval])
+		[hcache_tokyocabinet=$withval],
+		[hcache_tokyocabinet=check])
 AC_ARG_WITH(kyotocabinet,
 	AS_HELP_STRING(
 		[--with-kyotocabinet@<:@=DIR@:>@],
 		[Use kyotocabinet for the header cache]),
-		[hcache_kyotocabinet=$withval])
+		[hcache_kyotocabinet=$withval],
+		[hcache_kyotocabinet=check])
 AC_ARG_WITH(qdbm,
 	AS_HELP_STRING(
 		[--with-qdbm@<:@=DIR@:>@],
 		[Use qdbm for the header cache]),
-		[hcache_qdbm=$withval])
+		[hcache_qdbm=$withval],
+		[hcache_qdbm=check])
 AC_ARG_WITH(bdb,
 	AS_HELP_STRING(
 		[--with-bdb@<:@=DIR@:>@],
 		[Use BerkeleyDB for the header cache]),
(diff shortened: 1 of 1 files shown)


somini:
I built this branch excluding some backends and everything works as expected.

I noticed the output of `mutt -v` includes the selected backends:
```
NeoMutt 20170421-1-a85389 (1.8.2)
Mutt comes with ABSOLUTELY NO WARRANTY; for details type `mutt -vv'.
Mutt is free software, and you are welcome to redistribute it
under certain conditions; type `mutt -vv' for details.

System: Linux 4.10.11-1-ARCH (x86_64)
libidn: 1.33 (compiled with 1.33)
hcache backends: tokyocabinet, kyotocabinet, gdbm, bdb
```
This can help solve/reopen neomutt/aur-build#1

d-k-c:
Thinking about it, this isn't really appropriate from a packaging perspective. Package managers take care of dependencies, and warn you if you try to remove a package that is a dependence of another one. So if it autoscans and enables stuff without that relationship being handled by the package manager, we will end up in bad situations, and will (rightly) get blamed for it.

Arch Linux's `optdepends` is appropriate for dependencies that can come & go without impacting too much (like w3m, lynx, optional "runtime" dependencies so to speak), but as long as Neomutt can't dynamically load header cache backends (spoiler: it's not coming soon), Neomutt package maintainers will have to choose which backends they want to depend on. I think the only distro that can really take advantage of multiple backends currently is Fedora, but use only one by default.

In the end, I think the best solution would be another option like `--enable-hcache-detect` to make it explicit that it's somewhat unstable, or at least dependent on a flag set by the user; or just hint at the end of the configure that additional options could be passed to configure:
```
$ ./configure --with-gdbm
[...]
Neomutt could additionally be built on your system with:
    --with-tokyocabinet --with-qdbm
```

configure.ac
```diff
@@ -1107,16 +1113,16 @@ if test -n "$hcache_bdb" && test "$hcache_bdb" != "no"; then
...
 	fi
 fi
 
 dnl -- LMDB --
-if test -n "$hcache_lmdb" && test "$hcache_lmdb" != "no"; then
+if test "x$hcache_lmdb" != "xno"; then
 	OLDCPPFLAGS="$CPPFLAGS"
 	OLDLDFLAGS="$LDFLAGS"
-	if test "$hcache_lmdb" != "yes"; then
+	if test "x$hcache_lmdb" != "xyes" -a "x$hache_lmdb" != "xcheck"; then
```
gahr: Typo: hache

flatcap:
Thanks for the PR @d-k-c.
It's an interesting idea for developers, but I think it's just adding complexity that we don't need.
The code branch will be archived in the [neomutt-old](https://github.com/neomutt/neomutt-old) repo.

PR #549, opened 2017-04-25
configure: uniformize notmuch and remove dead code

d-k-c:
* What does this PR do?

Cleans up configure.ac to avoid messing with `MUTTLIBS` and `LIBS` in confusing ways, build notmuch the same way as other mutt's components, remove dead code related to configure symbols that are never defined.

Making sure that the `HAVE_` symbols are really not defined anywhere and that the changes to configure.ac don't break anything.

* Why was this PR needed?

Improve code readability by removing dead code and uniformize configure.ac and Makefile.am.


flatcap:
Rebased and fixed 'make distcheck'

PR #550, opened 2017-04-25
Show alternating padding in help menus

somini:
* What does this PR do?

Show alternating padding on help menus, to make it easier to line up the columns on larger screens.

* Why was this PR needed?

As suggested by tuxic in the mailing list.

http://mailman.neomutt.org/pipermail/neomutt-users-neomutt.org/2017-April/000116.html

### DIFF
--- a/help.c
+++ b/help.c
@@ -193,21 +193,28 @@ static int get_wrapped_width(const char *t, size_t wid)
   return n;
 }
 
-static int pad(FILE *f, int col, int i)
+static int pad(FILE *f, int col, int i, bool is_alternate)
 {
-  char fmt[32] = "";
-
   if (col < i)
   {
-    snprintf(fmt, sizeof(fmt), "%%-%ds", i - col);
-    fprintf(f, fmt, "");
+    if (col <= i + 2) {
+      fputc(' ', f);
+      for (int j = 0; j < i - col - 2; j++) {
+        fputs(is_alternate ? "―" : "•", f);
+      }
+      fputc(' ', f);
+    }
+    else {
+      /* Only a single character of difference */
+      fputc(' ', f);
+    }
     return i;
   }
   fputc(' ', f);
   return (col + 1);
 }
 
-static void format_line(FILE *f, int ismacro, const char *t1, const char *t2, const char *t3)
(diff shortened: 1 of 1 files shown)


somini:
This is just v0 of the code, it needs some extra work.
I just put it here to make sure I don't forget about it.

somini:
Rebased to master.

I'm still working to clean this up.

flatcap:
The code looks fine, but the output is crazy-looking.

I have a few thoughts:
- highlighting alternate lines is probably enough (dots or lines)
- two spaces after the key might be wise
  make it clear where the key ends and the dots begin
- colour-highlighting the alternate lines is a good alternative to dots/lines
  not sure how much work it would be (there's already some colour highlighting in the help)
- just colour the dots/lines (say, dark grey) to reduce the visual impact

somini:
I spent some time thinking about this three-column format, and after struggling to align and colour things properly, I went hunting for libraries and found this one, battle-tested:

https://github.com/karelzak/util-linux/tree/master/libsmartcols

This works perfectly, I can now output the help section in a clean way, using ANSI sequences. My struggle now is with the mutt builtin pager. I just can't get it to accept ANSI sequences and dump that raw into the terminal so that you get colours per-column.
By default the pager strips ANSI sequences, which seems nice for random emails, but for our help screen I can't get it to accept them. If I enable colours, the ANSI sequences are stripped too, and the same regexes from the colour scheme are used. Even though the docs explicitly mention this:

https://www.neomutt.org/guide/gettingstarted.html#tab-ansi-esc

If I change the pager to `less -R`, I get colours allright, but not in the mutt pager.

somini:
Anyway, I'll make the changes to address your comments. Since colors is out of the question, I'll alternate between space and dots, and add two spaces after the key.

gahr:
I have given the branch a go and I have to say that I don't like the dotted filling.

flatcap:
Thanks for the PR @somini.  Both are quite striking!
Unfortunately, the dotty one is a bit visually cluttered.
If only we could colouring working I think it would be brilliant.
As for libsmartcols, it looks good, but I'd rather not add it as a dependency.
This issue will now be closed and listed on the [closed discussions page](https://www.neomutt.org/dev/discussions)
The code branches will be archived in the [neomutt-old](https://github.com/neomutt/neomutt-old) repo.

PR #557, opened 2017-04-28
remove --enable-mailtool configure option

toogley:
* **What does this PR do?** Completely removes the `--enable-mailtool` configure flag.

* **Why was this PR needed?** It looks legacy. However I have not found much evidence about its legacy - still i doubt the contrary.

* **What are the relevant issue numbers?** #389 


gahr:
I have zero ideas whether anybody's using Solaris' `mailtool` nowadays, and if so, whether `mailtool` still employs these extensions. Anybody on Solaris around?

gahr:
What I would do instead is remove the configure option but leave the code there. Worst case, something that used to not work for a user all of a sudden will start working.

flatcap:
Following @gahr's suggestion, I've changed the PR to hide the option and leave the code enabled.
I've added this to my "review in 6 months" list.

PR #559, opened 2017-04-29
build: remove check_sec.sh

flatcap:
This script looked for mistakes when using mutt functions such as
safe_free().  It was crude.  Coverity can do a better job without us
having to litter the code with `/* __FREE_CHECKED__ */` comments.

This is an extension of @toogley's change in #497 - refactoring / cleanup of legacy files


toogley:
why did you commit *.po files? I don't think, this commit should change anything in *.po since this has nothing to do with the code.

flatcap:
Argh!  Pure carelessness.  Fixed and rebased.

PR #560, opened 2017-04-29
build: change gpgme requirement back to 1.1.0

flatcap:
RHEL6 is still using GPGME 1.1.8, which doesn't have the function
`gpgme_op_export_keys`.  Revert the GPGME requirement to 1.1.0 and put
back the #ifdef's around the missing function.

Note: The improvements to `configure.ac` are still there.

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -245,8 +245,9 @@ dnl == Declare all the checks and code for options / features below this line ==
 dnl --enable-gpgme
 AS_IF([test x$use_gpgme = "xyes"], [
 	ifdef([AM_PATH_GPGME], [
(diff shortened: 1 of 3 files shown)


gahr:
Should we do so, and keep supporting 9 years old software (gpgme 1.1.8, dated 08-DEC-2008), or should be up to whatever distro to backport? In this case I think it's fair to just put the check back. In general, I'd rather not waste time chasing crashes that might be caused by 3rd party bugs fixed ages ago. I'm pretty sure RHEL6 doesn't ship *vanilla* gpgme-1.1.

PR #564, opened 2017-05-01
Strip down configure.ac using POSIX:2001 as a guide

### LINKED ISSUE #538: Require POSIX compliance?
I was working on cleaning up some stuff in configure.ac, but I went too far and I ended up reading C99 standard and POSIX manual. Time to take a step back!

It seems that Mutt is in a strange situation. We check for a lot of stuff in our configure, have some workarounds if some of them are not defined, but it seems that in the end we require a system that is POSIX compliant. For instance, a lot of POSIX headers are included unconditionally like sys/time.h which is included in init.c without checking if HAVE_SYS_TIME_H is defined or not.

My point is that it looks to me like Mutt already requires a POSIX compliant system in order to build. If this is the case and we agree on that, then we could make the configure more focused on checking real dependencies, like external libs for header caches or crypto features; remove checks for C99/POSIX features (test it just once), and remove HAVE_FOOBAR in the code that we know always end up to be set to 1.

I started some work here, if you want to take a look but I think last commit breaks the compilation:
https://github.com/d-k-c/neomutt/tree/devel/configure-diet

I'd like more inputs. Do some of you use systems that are not POSIX compliant but still manage to build Mutt?

flatcap:
Closes #538 Require POSIX compliance?

This PR improves and extends @d-k-c's work.
We have declared [POSIX:2001](https://github.com/neomutt/management/blob/master/standard-functions/README.md) as the minimum requirement to build NeoMutt.

We no longer have to check for all the header files, functions and types declared in POSIX:2001.

1. bc4158e remove check on unix.h
2. 4266235 remove check on stdarg.h
3. 53295e0 remove check on inline and const
4. fc5ed28 remove check on wide chars
5. 2e54f27 remove check on fgetpos
6. b4887df remove check on memmove
7. 8acad85 remove check on strerror
8. 7fccc0c remove check on srand48
9. 29e244e remove check on long long int
10. 69aae11 signal: remove test on SYS_SIGLIST_DECLARED
11. b18f750 simplify sig_atomic_t check
12. ~~0000000 Test for POSIX:2001~~
13. 02cc7f1 remove check for <dirent.h> d_ino
14. ee97e00 drop LANGINFO checks
15. 48498f2 remove stuff defined in POSIX:2001
16. 2a1caf6 use basic getopt from unistd.h
17. 8917fdc drop HAVE_UNISTD_H
18. 539a57d unused prototypes - unsupported systems
19. e9a08f8 sys/types.h
20. 5e42b3e sys/select.h
21. 4d1c19d inttypes.h includes stdint.h
22. c424b9b Drop obsolete iconv check
23. ec4dd65 Drop another obsolete iconv check
24. afc1b7d tidy md5


md5.c
```diff
@@ -26,45 +26,22 @@
 #include "config.h"
 #include <stddef.h>
 #include <string.h>
+#include <endian.h>
```
gahr: `endian.h` is [glibc specific](http://man7.org/linux/man-pages/man3/endian.3.html). My system doesn't ship it, although it has a `sys/endian.h`. This doesn't look neither POSIX.2001 nor C99 to me ;) Currently this branch doesn't build on FreeBSD.

flatcap:
- Rebased
- POSIX:2001 test dropped
- @gahr's md5 fix merged

PR #565, opened 2017-05-02
Devel/configure builtin

### LINKED ISSUE #389: remove configuration flags which don't depend on external dependencies

d-k-c:
This closes #389 by dropping `--enable-compressed`, `--enable-imap`, `--enable-nntp`, `--enable-pop`, and `--enable-smtp`.

### DIFF
--- a/browser.c
+++ b/browser.c
@@ -1398,11 +1398,7 @@ void _mutt_select_file(char *f, size_t flen, int flags, char ***files, int *numf
           }
         }
 
-#ifdef USE_NNTP
-        if (buffy || option(OPTNEWS))
-#else
-        if (buffy)
-#endif
+        if (buffy || option(OPTNEWS)) /* USE_NNTP */
         {
           strfcpy(f, state.entry[menu->current].name, flen);
           mutt_expand_path(f, flen);
--- a/main.c
+++ b/main.c
@@ -249,13 +249,9 @@ int main(int argc, char **argv, char **environ)
         argv[nargc++] = argv[optind];
     }
 
-#ifdef USE_NNTP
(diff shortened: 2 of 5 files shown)


gahr:
If `imap` is not an option anymore, perhaps it makes little sense to keep building a dedicated `.a` for it. Wanna do it now? Wanna do it in a further PR? Either is good to me, I just thought I'd drop the thing here for reference.

flatcap:
I've just added a commit to drop the #else clauses from the preprocessor statements.

gahr:
I've committed in a separate branch an attempt to get rid of `libimap.a`. Feel free to integrate it or leave it for afterwards, if it builds fine ;)

flatcap:
rebased onto master

flatcap:
I'm in favour of this PR, but I have a few thoughts...

- When mutt was created it could only read a couple of local mailbox types.
  We're choosing to bake-in networking (IMAP, POP, SMTP), compression and NNTP.
  Is everyone happy with this?

- After a chat with @gahr, I'm not so sure about dropping `imap.a` and `imap/Makefile`.
  They wrap up a bit of the logic and keep it out of the main `Makefile`
  When we come to moving the hcache into a separate dir, it makes sense to move the build logic with it.

- We're leaving in the `#ifdef`s as markers.  As we develop and refactor the code we'll need to be careful that we keep the markers in the right places -- they're only useful if they're correct.

Issue #567, opened 2017-05-02
Segfault when trying to unset a variable using lua


darnir:
I built neomutt using the current git master. I'm trying to unset a variable using lua, my lua scrip looks like this:
```lua
-- Switch Mutt accounts
function switch_account(name)
	if name == "UdS" then
		_switch_account {
			user="username",
			pass="UdS/password",
			url="urlhere"
		}
	elseif name == "mlists" then
		_switch_account {
			user = "usernme",
			pass = "GMail/darnirmlist",
			url  = "imaps://imap.gmail.com",
			record = "",
			postponed = "[Gmail]/Drafts",
			trash = "[Gmail]/Trash"
		}
	end
end

function _switch_account(opt)
	_switch_to_imap_account(opt.user,
							get_pass(opt.pass),
							opt.url,
							opt.inbox or opt.url.."/INBOX",
							opt.record or opt.url.."/Sent",
							opt.postponed or opt.url.."/Drafts",
							opt.trash or opt.url.."/Trash")
end

function _switch_to_imap_account(user, pass, url, inbox, record, drafts, trash)
	mutt.set("imap_user", user)
	mutt.set("imap_pass", pass)
	mutt.set("folder", url)
	mutt.set("spoolfile", inbox)
	mutt.set("record", record)
	mutt.set("postponed", drafts)
	mutt.set("trash", trash)
end
```

Using such a function when I invoke switch_account("mlists"), mutt segfults and crashes.

A preliminary look into mutt with `gdb` shows me this:

The `_lua_mutt_set()` method is invoked to set a mutt config variable. This method in turn calls the `lua_tostring()` method to get the value from the lua interpreter. Since the value passed to the variable was "", the interpreter correctly returns an empty string back to mutt.

However, we pass this empty string pointer to `safe_strdup`. Here the real issue arises. `safe_strdup` checks if the pointer is `NULL` _or_ if the string passed to it is empty and in both cases it returns `NULL`. However, this time we indeed wanted an empty string. The `NULL` returned by `safe_strdup` is passed to `strfcpy` which attempts to dereference it and causes a segfault.

Now the question is where exactly do we fix this? Should `safe_strdup` return an empty string (may break expectations in other parts of the code), or should `strfcpy` handle the `NULL` pointer cases gracefully? I think the former is a better solution.

guyzmo:
simpler way to replicate the issue:

```config
lua mutt.set('record', '')
```

crash backtrace:

```
[...]

PR #568, opened 2017-05-02
* lib.c (safe_strdup): Don't return a NULL pointer when trying to cop…

darnir:
Potential fix for #567

### DIFF
--- a/lib.c
+++ b/lib.c
@@ -225,7 +225,7 @@ char *safe_strdup(const char *s)
   char *p = NULL;
   size_t l;
 
-  if (!s || !*s)
+  if (!s)
     return 0;
   l = strlen(s) + 1;
   p = safe_malloc(l);


gahr:
Please don't do this - it will likely break a zillion of other things. Mutt runs under the assumption that `safe_strdup` doesn't allocate a one-byte `"\0"` string but return NULL instead. Client code is supposed to use `NONULL` with the return value.

darnir:
I agree. I mentioned it in my original issue that this will likely break things. I quickly cooked this up and it seemed to work (didn't crash my mutt while I viewed and handled my mail)

gahr:
Of course it won't crash your mutt. You're killing an optimization and changing an assumption, though. What about you use `NONULL` like it's done all over the place instead?

guyzmo:
I agree with @gahr. This is a false good idea. Whether returning a NULL pointer for empty strings being a good idea is debatable, that choice has been made 19 years ago. So I'm closing that issue, and fix my own mistake!

PR #573, opened 2017-05-04
The trouble with dotlock...

flatcap:
1. 90263ab build: make dotlock optional
2. edc988b build: completely drop dotlock

Commit 1 makes dotlock work better with the other configure options.
If we decide that we don't want dotlock any more, commit 2 removes it completely.              

## What used to happen                                                                         

- Something you enabled forced dotlock to be built
- When you installed it locally, `make` tried to make dotlock **setgid**
- `configure` checked for local directories, which makes no sense for packaging / cross-compiling

## New behaviour                                                                               

- `--with-mailpath` defaults to `/var/mail` unless overridden
- dotlock won't be built/install unless you `./configure --enable-dotlock`                     
- if installed dotlock will be setgid to "mail" unless overridden, e.g.                        
  `make DOTLOCK_GROUP='' install`
  
## Discussion

Mutt has three locking methods, `fcntl()`, `flock()` and `dotlock`

- `fcntl()` is widely available (being in POSIX:2001) and it even works across NFS
- `flock()` is a good alternative
- `dotlock` is a very clumsy solution

@neomutt/reviewers Should we get rid of it altogether?


gahr:
> Should we get rid of it altogether?

oh yes, pretty please!

configure.ac
```diff
@@ -471,76 +472,18 @@ if test x$with_homespool != x; then
...
-	AC_DEFINE(DL_STANDALONE,1,[ Define if you want to use an external dotlocking program. ])
+AC_ARG_ENABLE(dotlock, AS_HELP_STRING([--enable-dotlock],[Define to build mutt_dotlock.]), [build_dotlock=$enableval])
+if test "x$build_dotlock" = "xyes"; then
+	AC_DEFINE(USE_DOTLOCK,1, [ Define to use dotlock to lock mailboxes. ])
 	DOTLOCK_TARGET="mutt_dotlock\$(EXEEXT)"
-else
-	MUTT_LIB_OBJECTS="$MUTT_LIB_OBJECTS dotlock.o"
 fi
-
+AM_CONDITIONAL(BUILD_DOTLOCK, test "x$build_dotlock" = "xyes")
```
d-k-c: Is this needed?

flatcap: Ah, no it isn't.  Good catch.
That was left over from one my experiments.
Fixed

xdgc:
* On 04 May 2017, Richard Russon wrote: 
> 
> Mutt has three locking methods, `fcntl()`, `flock()` and `dotlock`
> 
> - `fcntl()` is widely available (being in POSIX:2001) and it even works across NFS
> - `flock()` is a good alternative
> - `dotlock` is a very clumsy solution
> 
> @neomutt/reviewers Should we get rid of it altogether?

dotlock is the universal fallback.  No one thinks it's a good solution,
but fnctl and flock are not always available.  They're questionable over
smbfs, for example.

-- 

toogley:
As we have declared [POSIX:2001 to be the minimum build requirement](https://github.com/neomutt/management/blob/master/standard-functions/README.md): Is there a reason why we can't just use `fnctl()` everywhere? And get rid of `flock()` too (as well as dotlock) ?

configure.ac
```diff
@@ -471,78 +472,12 @@ if test x$with_homespool != x; then
 		MAILPATH should be set to the filename of the spool mailbox
 		relative the the home directory.
 		use: configure --with-homespool=FILE])
-	AC_DEFINE(USE_DOTLOCK,1,[ Define to use dotlocking for mailboxes. ])
-	mutt_cv_setgid=no
 else
 	AC_ARG_WITH(mailpath, AS_HELP_STRING([--with-mailpath=DIR],[Directory where spool mailboxes are located]),
-		[mutt_cv_mailpath=$withval],
```
toogley: Hm. i understand the dotlock changes in this file, but i don't understand why the initialization of `mutt_cv_mailpat` belongs in this commit? i mean, what does this have to do with dotlock?

gahr: Messages from code that is being removed.

gahr: Ehm.. this reply was for your other comment about po files :-/

toogley: @gahr aah, i see. But still, maybe it would be tidier if it is in a separate commit? (the po changes)

flatcap:
Updates:
- Commited change to make dotlock configurable
- Split out the .po file changes
- Rebased onto master

flatcap:
> but fnctl and flock are not always available. They're questionable over smbfs, for example.

@xdgc 
Are they still problematic over smbfs?  (I'm no expert).
Googling, I couldn't find any reported problems in the last five years (a very few within ten years).

toogley:
@flatcap why was this closed? is it dismissed or integrated?

flatcap:
Integrated.  Merged by #617

PR #574, opened 2017-05-06
build: add separate hcache dir

flatcap:
**Notes**:
- Move all hcache source to `/hcache`
- Other source only refers to "hcache/hcache.h"
- Shortened filenames `s/hcache_//`
- Changed `configure.ac` references from object files to symbol name
- hcache's `Makefile.am` has rules to make sure that `keymap_defs.h` and `mutt_md5` have been built
  This could probably be improved

**Tested**:
- Build with **no** hcache
- 6 builds with **one** hcache
- Build with **all** hcaches
- `make dist`, `make distcheck`
- `make install`, `make uninstall`

@neomutt/reviewers 
I _think_ I've got it right, but welcome your testing and comments.


d-k-c:
I think it would be better to move the hcache "public" header to a `include/mutt` folder, and add `-Iinclude` to Makefile.am so that other files could include them using `#include <mutt/hcache.h>`. Yes it would be a bit awkward to have only one file in that folder at first, but we should start somewhere, right?

flatcap:
I've added a readme and tweak the lib handling slightly.

@d-k-c 
Yes, perhaps we can move it when we have a second candidate to put in the new dir.

PR #579, opened 2017-05-13
Several build improvements

flatcap:
These changes _shouldn't_ cause any problems.

The main changes are to "silent-ify" the rest of the make rules.
I've tweaked the building of the docs to remove the need for the timestamp files.

1. d06779d remove unused definitions
2. f1179ba update distcheck build flags
3. 76b63ec move configure-generated files
4. 3c3e57d drop docs timestamps
5. fce65a1 add silent rules, but default to off
6. 2e24ad4 tidy po/Makevars
7. 43bc341 light tidy 
8. c148f58 strerror

### DIFF
--- a/doc/makedoc.c
+++ b/doc/makedoc.c
@@ -35,10 +35,6 @@
 #include <unistd.h>
 #include "makedoc_defs.h"
 
-#ifndef HAVE_STRERROR
-#define strerror(x) ((x) > 0 && (x) < sys_nerr) ? sys_errlist[(x)] : 0
-#endif /* !HAVE_STRERROR */
-
 extern int optind;
 
 #define BUFFSIZE 2048


gahr:
I am personally against silent make rules. What added value does it have?

flatcap:
> What added value does it have?

Value?  Very little :-)
As developers, we spent a LOT of time building and anything that makes the build output easier to read seems like a win to me.  I imagine we've all got a build script _that_.

For the silent rules to be effective, all the build steps need to be made "silenceable".
Mostly, this means prefixing recipes with `$(AM_V_GEN)`

If you're strongly dislike "silent rules", I'll leave in the changes but set configure's default to verbose.

gahr:
Well, I don't care too much about the default, as you say, I do have my build script wrapper that could set the right option. My concern is that the added churn (look at the diff just to implement that) has a high cost/benefit ratio. More steps to get useful info in case of build failures. For developers it could be trivial, but any additional step requires time, and users reporting issues might just not want to go through the trouble.
My point is, if you don't care about the out just pipe it to > /dev/null or look the other way when it builds.

flatcap:
**Update**:  I've changed the default for `--enable-silent-rules` to **`false`**
They're there and working for anyone who wants them, but off by default.
This means that newbies with build problems will have something to cut'n'paste.

PR #580, opened 2017-05-13
build: split up mutt.h, protos.h

flatcap:
After moving the header cache code into its own directory, @d-k-c suggested moving `hcache/hcache.h` into a new dir `include/mutt`
It sounded like a good idea, but it included "mutt.h"  This would mean the include dir would need pretty much all the other headers, too.

This set me working...

Though this is a large diff, it's only a few small changes.
**Note**: Because it's the result of ~100 small, iterative, commits, it's now too much work to separate then.  Sorry.

**Main changes**:
- Split up `mutt.h`
- Split up `protos.h`
- Run IWYU (include-what-you-use) on all source -- https://include-what-you-use.org/

**New files**:
- **address.h** -- ADDRESS
- **alias.h** -- ALIAS
- **body.h** -- BODY
- **content.h** -- CONTENT
- **context.h** -- CONTEXT
- **enter_state.h** -- ENTER_STATE
- **envelope.h** -- ENVELOPE
- **format_flags.h** -- enum MUTT_FORMAT* used by mutt_FormatString
- **header.h** -- HEADER
- **list.h** -- LIST
- **mbyte_table.h** -- mbchar_table
- **options.h** -- enum of OPT*
- **parameter.h** -- PARAMETER
- **pattern.h** -- pattern_t
- **state.h** -- STATE
- **thread.h** -- THREAD
- **where.h** -- evil macro

After the split, all the `#include`s were adjusted using IWYU.
Most of the .c files now have many more includes (showing the complexity of the source).
Many of the headers can get away with forward declarations of structs/typedefs.

@neomutt/reviewers 
This is quite a crude carve up of the source.  Better suggestions are welcomed.


gahr:
While I am in favour of IWYU, I fear it might still be a bit too early for us to adopt it, especially if it means introducing large blocks of pseudo forward declarations like this:

```c
typedef struct account ACCOUNT;
typedef struct body_cache body_cache_t;
typedef struct buffer BUFFER;
typedef struct hash HASH;
typedef struct header HEADER;
typedef struct imap_header_data IMAP_HEADER_DATA;
typedef struct imap_mbox IMAP_MBOX;
typedef struct list_t LIST;
typedef struct progress progress_t;
typedef struct _connection CONNECTION;
typedef struct _context CONTEXT;
typedef struct _message MESSAGE;
`
Note that these do not provide any decoupling between implementation and use: you need both the original structure name and the typedef'd name.

Is the code base really improving because of this commit?

**Comment by flatcap** [comment]
> introducing large blocks of pseudo forward declarations

Any types that aren't dereferenced, don't need a `#include`
and that's the standard C way of doing it.

In these days of fast disks, fast compilers and caching, it probably doesn't make a difference to the build time.
If you prefer, I can replace all the forward-decls with `#include`s.

> Is the code base really improving because of this commit?

IMHO, yes.  But not by much (yet).

The more we can split things up, the easier it is to see the inter-dependencies.
Now, by looking at the head of the file, you know **exactly** what it depends on.

**Comment by gahr** [comment]
> > introducing large blocks of pseudo forward declarations

> Any types that aren't dereferenced, don't need a #include and that's the standard C way of doing it.

Of course, I'm not arguing on that.

In a perfect world you'd have `foo.h` defining `struct foo { ... };` and in `bar.h` you'd only need to do this:
```c
struct foo;
void bar(struct foo*);
```

In our world, on the other hand, inside `foo.h` you have nonsense like `typedef struct FOO {...} foo;`, and you end up with this in `bar.h`:
```c
typedef struct FOO foo;
void bar(foo*);
```

This, in my opinion, loses the benefit of decoupling `bar` from `foo`, because you need to know that a particular `struct FOO` was typedef'd to `foo`.

Perhaps it would be better to remove those typedefs in the first place, and *then* aim at decoupling by using forward declaration on plain structs. This is why I like the idea but I think it's a bit early in the refactoring process.

flatcap:
> introducing large blocks of pseudo forward declarations

You're right, of course.

> Perhaps it would be better to remove those typedefs in the first place

As if by magic c938600 :-)

Now all the ugly forward decls have gone, leaving a small handful of `struct Address;` etc

gahr:
The beauty itself! Thumbs up, @flatcap!

PR #592, opened 2017-05-24
Add reorganised collapse_flagged patch

petelewis:
* **What does this PR do?**
Integrates the collapse_flagged patch, which was part of several other patched versions of mutt and provides a useful feature. 

While it seems simple enough, and works for me, I would appreciate someone looking over the whole thing, as it is not my code, and I have not contributed to (neo)mutt before.

* **Why was this PR needed?**
Specifically, it adds a new option, "collapse_flagged", which when set, makes Mutt keep threads open if they contain a flagged message. It works similarly to collapse_unread.

Can provide if necessary, but hopefully this is straightforward enough to understand.

I wasn't quite sure how this should be documented, since there is nothing in manual.xml.head about collapse_unread anyway. Please let me know if we should add something there.

Also please note that this isn't my code, it just came from a previous mutt collection I maintained [here](https://aur.archlinux.org/packages/muddy-mutt/). I have just done a small bit of reorganisation and tidying, based on recent changes, and rebased it on current neomutt master.  I am not sure who the original author is, although the patch I found was made by Fabian Groffen <grobian@gentoo.org>. The patch's original home (http://debian.lpr.ch/Mutt/) has gone.


flatcap:
Welcome to NeoMutt @petelewis.
Thanks for the patch.  I'll have a close look at it later.

init.h
```diff
@@ -440,6 +440,12 @@ struct Option MuttVars[] = {
   ** When \fIunset\fP, Mutt will not collapse a thread if it contains any
   ** unread messages.
   */
+  { "collapse_flagged",	DT_BOOL, R_NONE, OPTCOLLAPSEFLAGGED, 1 },
+  /*
+  ** .pp
+  ** When \fIunset\fP, Mutt will not collapse a thread if it contains any
+  ** flagged messages.
+  */
   { "compose_format",   DT_STR,  R_MENU, UL &ComposeFormat, UL "-- NeoMutt: Compose  [Approx. msg size: %l   Atts: %a]%>-" },
```
flatcap: You don't need to add anything to the manual.xml.head because the comments here in `init.h` are processed into docs.

thread.c
```diff
@@ -1156,7 +1156,7 @@ int _mutt_traverse_thread(struct Context *ctx, struct Header *cur, int flag)
   struct MuttThread *thread = NULL, *top = NULL;
   struct Header *roothdr = NULL;
   int final, reverse = (Sort & SORT_REVERSE), minmsgno;
-  int num_hidden = 0, new = 0, old = 0;
+  int num_hidden = 0, new = 0, old = 0, flagged = 0;
   int min_unread_msgno = INT_MAX, min_unread = cur->virtual;
```
flatcap: While you're editing this file, please could you change `new`, `old` and `flagged` into `bool`, then replace `0`/`1` with `false`/`true`.

curs_main.c
```diff
@@ -2681,7 +2686,7 @@ int mutt_index_menu(void)
         }
         else
         {
-          mutt_error(_("Thread contains unread messages."));
+          mutt_error(_("Thread contains unread or flagged messages."));
           break;
```
flatcap: Ah, this isn't true, we're missing a test.

If I set `collapse_flagged = no`, then when I enter a folder, flagged messages aren't folded.
However, I can still manually collapse the thread even if it contains a flagged message.

With `collapse_unread = no`, the thread is open when I enter the folder and I **cannot** manually fold it.

flatcap:
Hi @petelewis,

Have you had a chance to look at the changes I suggested?

PR #594, opened 2017-05-26
Fail early if ncursesw cannot be found

gahr:
Issue: #585

* **What does this PR do?**

Adds a check to configure.ac so that if `waddnwstr` cannot be found, the configure phase fails and the user is not left with link errors.


gahr:

gahr:
Great, thanks for your tests, @flatcap !

PR #600, opened 2017-06-01
Do not try to create Maildir if it is an NNTP URI

### LINKED ISSUE #599: "news://news.gmane.org/ does not exist. Create it? ([yes]/no)" at every startup with news:// spoolfile
I have a muttrc for catching up with unsubscribed mailing lists via gmane NNTP. This works fine, but it askss me to create news://news.gmane.org/ at startup which is just annoying.

muttrc: 
- https://github.com/julian-klode/dotfiles/blob/master/.mutt/account.news
- https://github.com/julian-klode/dotfiles/blob/master/.mutt/muttrc.common (sourced from above)

gahr:
Fixes #599

* **What does this PR do?**

At startup, if Maildir (the `folder` setting) does not exists, mutt tries to create it. A pre-existing check skips this logic if Maildir is an IMAP URI. This PR extends this skip logic to NNTP URIs.

### DIFF
--- a/main.c
+++ b/main.c
@@ -504,19 +504,23 @@ int main(int argc, char **argv, char **environ)
 
     strfcpy(fpath, Maildir, sizeof(fpath));
     mutt_expand_path(fpath, sizeof(fpath));
+    bool skip = false;
 #ifdef USE_IMAP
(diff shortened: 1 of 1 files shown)


gahr:

Issue #601, opened 2017-06-01
Consider requerying when sync-mailbox notmuch vfolders


julian-klode:
When using notmuch vfolders, and you remove a tag from a message that is part of the current query, it does not disappear when you sync-mailbox unless you quasi-delete it. This has some disadvantages:

- If you quasi-deleted a message, you can't undelete it later on in case you accidentally removed the label and want to re-add it.
- You cannot write generic macros for tagging. For example, if I have tags "inbox" and "later", and the action "+inbox -later" and "-inbox +later", I need two variants for each: In inbox, the -inbox hides the message; in later, the -later one should hide the message. You can use folder-hooks to some extent, but it's not that pretty.

So it might be a good idea to just re-issue the query when syncing the mailbox, and preserve the position (if possible, probably keep ids of current message +/- 10 and then jump to the closest match after requerying).

julian-klode:
It might be too expensive to reopen in sync-mailbox, or someone might rely on sync-mailbox to not add new emails (which could happen if new ones were tagged in the background), so maybe it might be a good idea to have a new function for that.

julian-klode:
Alternatively, and what I initially thought about, but then discarded is automatically setting the quasi-delete state based on whether the message would be in the query results. I initially thought that's not possible but here's a new idea.

If you modify-labels-smart (let's call it this for now), after modifying, run a new query for `id:MSGID and (ORIGINAL-QUERY)`. Then do the following:

- If the message is *not* in the returned results, quasi-delete it
- Otherwise, undo the quasi-delete

That practically solves the issue in the least intrusive way, and maintains the current performance.

flatcap:
Can this be closed now?

julian-klode:
I think we might want to keep it open. The commit is a workaround that solves most issues, but perhaps requerying might be more practical long term (if it's fast enough)

PR #602, opened 2017-06-01
make modify-labels-then-hide hide or unhide

julian-klode:
If the message is no longer in the current query after modifying
the label it will be hidden as before, but if it is in the current
query (again), it will be unhidden.

Related to #601

### DIFF
--- a/curs_main.c
+++ b/curs_main.c
@@ -1864,9 +1864,12 @@ int mutt_index_menu(void)
               if (!Context->quiet)
                 mutt_progress_update(&progress, ++px, -1);
               nm_modify_message_tags(Context, Context->hdrs[Context->v2r[j]], buf);
+
+              bool still_queried =
+                  nm_message_is_still_queried(Context, Context->hdrs[Context->v2r[j]]);
               if (op == OP_MAIN_MODIFY_LABELS_THEN_HIDE)
               {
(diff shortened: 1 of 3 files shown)


julian-klode:
Not sure whether reusing the then-delete makes sense, and it's missing documentation for now, but this gives us a chance to look at it :)

julian-klode:
Performance evaluation (ThinkPad X230 with Core i5-3320M, slowish SSD):

I tagged 827 emails with and without the change. The runtime was 1.5 to 1.6 seconds in both cases. I measured the overhead inside the nm_message_is_still_queried() function by calling clock_gettime() at beginning and end and building some static total values:

[2017-06-01 22:03:06] nm: overhead: 0.116273 ms, avg: 0.182960 ms, worst: 0.423519, best: 0.077543, total: 151.307598 ms in 827.000000 entries

This seems to be OK.

julian-klode:
This is still a bit strange if your query type is thread or you used entire-thread, but it's definitely better than than current then-hide semantics (although the name is odd now...)

toogley:
is this ready for review or still in progress?

julian-klode:
@toogley I guess we wanted to discuss if this is the right move to do that here or if we want a new function or something, but otherwise, this is ready.

toogley:
@julian-klode ah, okay. Then @neomutt/reviewers please review.

PR #606, opened 2017-06-03
Decode CRLF line endings to LF when copying headers

julian-klode:
This fixes display issues with files that consist of CRLF
line endings rather than just LF line endings, for example,
emails exported directly from the gmail API.

This only kicks in when CH_DECODE is set which might be a bit
strange, because that does RFC2047 header decoding, but it seems
a safe place to place this line end decoding without potentially
affecting anything else.


julian-klode:
@neomutt/reviewers

Issue #583, opened 2017-05-16
how to disable "warning in muttrc"


Alok:
I bind `gi`, `gg`, and a few other chords to be like gmail. Now when I launch `mutt`, I get a warning about aliasing. How do I disable that warning?

toogley:
As far as i know, this isn't possible at the moment.

flatcap:
The trick, for now is to unbind the mapping that's overwritten.
It would have been helpful if you'd told us what the error message said.
I guess, that this might help:
```
bind index,pager g noop
```

Vim, and probably other editors, have a timeout, which allows them to map both 'x' and 'xy'.
Commit: 744fe99 was just a quick fix to tell the user when a mapping was being overwritten.

somini:
Yeah, I was the one who created this feature (#451).

I think this is the correct behaviour, specially if you have a big muttrc, otherwise you can accidentally alias commands.

With @Alok bindings, the group reply (g) binding is lost.

ahmedelgabri:
It will be nice if we can turn off the warnings I have the same issue as @Alok with the exactly the same keybindings.

guyzmo:
well, you should rather unbind the `g` command as suggested above. The issue is not the warning, but the fact that you're asking something mutt cannot resolve !

ahmedelgabri:
I actually had a couple more than just those `g` ones.

So to get this straight, every time I bind keys now they won't override the default bindings by default & I have to unbind first?

somini:
They override the defaults, but now you get a warning. 
The old version overrode the defaults silently, so you could lose keybindings. Now the defaults are overridden so that you know that some keybindings are invalid. If you don't mind, unbind them first.

oblitum:
I did override it on purpose, this completely screw up my workflow....

oblitum:
If it warned but started up mutt anyway, but warns and stops execution to ask for continuing.

PR #608, opened 2017-06-04
unbind mappings before overwriting in vim-keys

pigmonkey:
* **What does this PR do?**

Sets mappings to `noop` before overwriting them to prevent warnings.

No.

* **Why was this PR needed?**

Changes to key bindings in release 2017-06-02.

#583


flatcap:
Thanks, @pigmonkey.

Issue #616, opened 2017-06-06
integrate external patches


toogley:
## Here's a list of patches I've found, which are not already integrated in neomutt.

- ~[better Gmail integration ](https://github.com/daniel-w/mutt)~
- [x] [Multiple source patch](http://cedricduval.free.fr/mutt/patches/download/patch-1.5.4.cd.source_multiple.2) --> #626 
- [x] [Editing threads patch](http://cedricduval.free.fr/mutt/patches/download/patch-1.5.5.1.cd.edit_threads.9.5) --> cef50ebe86951eefd03bb3de3dccb3d43a0d0fa7 (thanks for the note @somini )
- [ ] [My_list patch](http://cedricduval.free.fr/mutt/patches/download/patch-1.5.1.cd.my_list.0.4) 
- [ ] [Signatures menu patch](http://cedricduval.free.fr/mutt/patches/download/patch-1.5.5.1.cd.signatures_menu.2.1)
- [x] [Pattern for duplicated messages patch](http://cedricduval.free.fr/mutt/patches/download/patch-1.3.25.cd.pattern_duplicates.1) --> 8f64b402e38af6b68ecc8a1a87cfa82b0aa765bd (thanks for the note @somini )
- [X] [Pattern for broken threads patch](http://cedricduval.free.fr/mutt/patches/download/patch-1.5.1.cd.pattern_broken.1) --> #623 
- [ ] [ignore thread](https://gist.github.com/SteveClement/6c6bcf9c7c2f09174e790033e2feef7f)

-----------------

Some Lists:

* https://github.com/hihellobolke/mutt-extra/tree/master/mutt-patched
* https://dev.mutt.org/trac/wiki/PatchList

toogley:
@daniel-w are you interested in joining us and/or integrating your gmail patch into neomutt?

somini:
Duplicated Messages: 8f64b402e38af6b68ecc8a1a87cfa82b0aa765bd

somini:
Editing threads: cef50ebe86951eefd03bb3de3dccb3d43a0d0fa7

toogley:
I'll take the my-list patch.

flatcap:
@toogley Where did the "better Gmail integration" come from?
I've just checked daniel-w's repo and it doesn't contain **any** changes from upstream mutt.

toogley:
@flatcap Thanks for the hint. I honestly don't know where i got that from, nor can i find any references in blogs/patchlists and so on.

PR #623, opened 2017-06-07
Include pattern for broken threads

somini:
* **What does this PR do?**

From Frederic Duval:
http://cedricduval.free.fr/mutt/patches/download/patch-1.5.1.cd.pattern_broken.1

#616


somini:

This is a trivial patch.

PR #624, opened 2017-06-08
Fix latest coverity issues

gahr:
ncrypt/crypt.c
CID 170494:  Resource leaks  (RESOURCE_LEAK)
Variable "from" going out of scope leaks the storage it points to.

nntp.c
CID 170495:  Uninitialized variables  (UNINIT)
Declaring variable "buf" without initializer.

parse.c
CID 76957:  Error handling issues  (NEGATIVE_RETURNS)
"loc" is passed to a parameter that cannot be negative.

### DIFF
--- a/ncrypt/crypt.c
+++ b/ncrypt/crypt.c
@@ -182,7 +182,7 @@ int mutt_protect(struct Header *msg, char *keylist)
   {
     /* Set sender (necessary for e.g. PKA).  */
     const char *mailbox = NULL;
-    const struct Address *from = msg->env->from;
+    struct Address *from = msg->env->from;
 
     if (!from)
       from = mutt_default_from();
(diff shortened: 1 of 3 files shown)


ncrypt/crypt.c
```diff
@@ -195,6 +195,9 @@ int mutt_protect(struct Header *msg, char *keylist)
       crypt_smime_set_sender(mailbox);
     else if ((WithCrypto & APPLICATION_PGP) && (msg->security & APPLICATION_PGP))
       crypt_pgp_set_sender(mailbox);
+
+    if (!msg->env->from)
+      FREE(&from);
   }
```
flatcap: I think this needs to `FREE(&from->personal)` too.
`mutt_default_from()` either generates a `rfc822_new_address()` or calls `rfc822_cpy_adr_real(From)`
`rfc822_cpy_adr_real()` generates a `rfc822_new_address()` and sets `->personal`

Issue #552, opened 2017-04-26
configure related tasks


d-k-c:
Management task to keep track of development on configure/autools

- [ ] kill prepare.sh & document build process
- [ ] remove MUTTLIBS, rely only on LIBS (or find out why they should stay separate)
- [x] remove build flags that don't have external dependencies (#389). Please keep related `#ifdef HAVE_` in the code for now, so don't forget to keep the definitions for these symbols. See #528 for details.
  - [x] --enable-sidebar (#528)
  - [x] --enable-compressed
  - [x] --enable-pop
  - [x] --enable-imap
  - [x] --enable-nntp
  - [x] --enable-smtp
- [ ] add a developer flag `--enable-everything-available` that selects if available the following features: gpgme, gpg, smime, notmuch, lua, header cache backends, sasl, ssl, idn, gss. See #547 for the concept, limited to hcache for now.
- [ ] make `mutt -v` expand `--enable-everything`. So instead of `Configure options: --enable-everything`, it would display `Configure options: --enable--notmuch`. Same thing for `--enable-everything-available` once implemented.

toogley:
>  kill prepare.sh 

why? I find its useful.

flatcap:
I imagine all the developers have their own script / aliases to build neomutt their way.
`prepare.sh` doesn't really add anything.

Now that we're stripping out lots of configure options, there's less value that `prepare.sh` could add.
Anywhere a doc says "run prepare.sh", it might as well say run "autoreconf -i; ./configure [OPTIONS]"

toogley:
Ah, okay. well that's true. => i'm fine with the removal.

flatcap:
@d-k-c  Have you made any progress with MUTTLIBS vs LIBS?
I'd like to see that (user-visible bit) tidied up.

d-k-c:
Not yet, I was quite busy moving to a new place. I'll give it a shot as soon as possible

flatcap:
I think this can be closed now.
`prepare.sh` will go away when the autotools dies.
`LIBS` vs `MUTTLIBS` has been sorted
#547 --enable-everything-available has been closed (no thanks)
PR #946 expanded `--everything` in the autosetup build.

PR #625, opened 2017-06-08
make +DEBUG persistent

toogley:
* **What does this PR do?**

    * removes `--enable-debug` compile option
    * remove `#define` DEBUG and corresponding `#ifdef`. This is fine IMHO, because debugging code usually doesn't need to be refactored. Or more precisely: "making a tighter integration into the sidebar easier" was a reason to not touch `#ifdef` of e.g. IMAP. This doesn't apply to `DEBUG`.
    * non-debug being the default is **not** changed.

    * [imap/auth_gss.c](https://github.com/neomutt/neomutt/compare/master...devel/persistent-debug#diff-092fac0ebe696f5a8bd7856d7fba9621R82)
    * [rfc3676.c](https://github.com/neomutt/neomutt/compare/master...devel/persistent-debug#diff-f0eb956c12d2e70484cf45af91dc0f5fR354)
    * [imap/auth_gss.c](https://github.com/neomutt/neomutt/compare/master...devel/persistent-debug#diff-092fac0ebe696f5a8bd7856d7fba9621R220)
    the above are variables where its not obvious that their purpose is for debugging. I agree that the `#ifdef` around these places makes the purpose clearer as with comments. However, i the removal of `#ifdef DEBUG` is still more "lean" in the other places. Therefore, I'm unsure what to do here, if commenting `// debug ` is really good.

#552 #389 

* **further TODO:**

    * update translations (i guess? i'm not sure)
    * update documentation

### DIFF
--- a/auto.def
+++ b/auto.def
@@ -27,7 +27,6 @@ options {
   with-ncurses:path         => "Location of ncurses"
   with-slang:path           => "Location of S-Lang"
 # Features w/o 3rd party dependencies
-  logging=0                 => "Enable debug logging"
   doc=1                     => "Disable building the documentation"
   full-doc=0                => "Build the full documentation set"
   docdir:path               => "Documentation root"
@@ -96,7 +95,7 @@ if {1} {
   # Keep sorted, please.
   foreach opt {
     bdb doc everything fcntl flock fmemopen full-doc gdbm gnutls gpgme gss
-    homespool idn kyotocabinet lmdb locales-fix logging lua mixmaster nls
+    homespool idn kyotocabinet lmdb locales-fix lua mixmaster nls
     notmuch pgp qdbm sasl smime ssl tokyocabinet
   } {
     define want-$opt [opt-bool $opt]
@@ -304,10 +303,6 @@ if {[get-define want-everything]} {
 } else {
   set conf_options "$::argv"
 }
-
-###############################################################################
-# Debug
-if {[get-define want-logging]} {define DEBUG}
 
 ###############################################################################
 # flock(1)
@@ -826,7 +821,6 @@ set auto_rep {
   BINDIR
   BUILD_DOC
   CRYPT_*
-  DEBUG
   DOMAIN
   ENABLE_*
   HAVE_*
--- a/configure.ac
+++ b/configure.ac
@@ -591,12 +591,6 @@ AM_CONDITIONAL(USE_SASL, test x$need_sasl = xyes)
 
 dnl -- end socket --
 
-AC_ARG_ENABLE(debug, AS_HELP_STRING([--enable-debug],[Enable debugging support]),
-	[ if test x$enableval = xyes; then
-		AC_DEFINE(DEBUG,1,[ Define to enable debugging info. ])
-	fi
-])
-
 AC_ARG_ENABLE(flock, AS_HELP_STRING([--enable-flock],[Use flock() to lock files]),
 	[if test $enableval = yes; then
 		AC_DEFINE(USE_FLOCK,1, [ Define to use flock() to lock mailboxes. ])
--- a/conn/ssl.c
+++ b/conn/ssl.c
@@ -130,9 +130,7 @@ static int ssl_load_certificates(SSL_CTX *ctx)
   X509 *cert = NULL;
   X509_STORE *store = NULL;
   int rc = 1;
-#ifdef DEBUG
   char buf[STRING];
-#endif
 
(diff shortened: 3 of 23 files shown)


toogley:
rebased onto master.

gahr:
I fail to understand how these two statements go together:
1. removes --enable-debug compile option
2. non-debug being the default is not changed.

toogley:
@gahr i understood mutt debugging in that way, that it **only** gets activated when called with the -d flag.

gahr:
Oh I see what you mean now. I agree with removing `--enable-debug`, but haven't looked at the code in detail yet. Partial 👍 for now ;)

toogley:
Sorry about the formatting changes. I forgot to disable Clang autoformatting.

But those changes will be anyways commited in the next clang-format -i *.c run. Should i still remove them?

guiniol:
If there's going to be a clang-format soon, then remove them, it's better if all those formatting changes are together, otherwise just keep them. At least, that's what I would do. You contribute much more than me, so you decide 👍

toogley:
@guiniol okay, then i'll remove them. (& rebase onto master)

toogley:
fixed formatting + rebased onto master.

toogley:
In IRC, @flatcap, thomas_adam and I discussed #625, here's a summary for future discussion:

* The basic problem with the solution proposed by #625 is that some users might say

    > oooh, i didn't thought level >2 would expose my passwords, why didn't you tell me?

    * however, thomas_adam pointed out, that this thought might be a bit
      over-the-top, because a corefile's backtrace also can reveal passwords.
        * However, they are usually disabled by default, as far as i know.

* the solution (so far)
    * debuglevel 1 and 2 are built in (without recompiling it explicitly in)
      in these levels, we'd make sure that no sensible data is exposed.
    * if the user wants to have debuglevel 3 and 4, they have to recompile their binary
      in these levels, passwords and sensible data is exposed in logfiles and so on.
        * maybe it would make sense to force the user to use a different directory?
        * because the user might forget their logfiles with sensible data in it?


## future changes

* make sure that two instances of mutt can't contaminate the logs of the other mutt program.

* ~/.mutt/log/FILE (where FILE can contain strftime() strings)
  e.g. log-%Y-%m-%d.txt

* let mutt receive SIGUSR1 when its unresponsive and debug_level can't be
  manually turned on. There's been countless times where a problem is occuring
  and logging isn't on, yet trying to reproduce the problem again is tricky.

d-k-c:
We should not write such sensitive information in log files, regardless of the log level. Maybe these logs were necessary when protocols were being implemented, but they are mostly irrelevant now and really expose information in a sneaky way. I mean that you might think logs are safe to be shared because you grep'd for your password and it was not found, but the pass is actually there, just encoded.

My advise would be to substitute them by generic strings such as "LOGIN" and "PASSWORD".

toogley:
What should we even consider sensitive information?

I mean, should we consider meta data from emails as such? I think so. But that would in turn make the logfiles much less useful, i guess.

d-k-c:
Agreed, but at least it's clear that email metadata are logged by a quick glance at the debug file, you can see email subjects or whatnot, whereas your credentials might be logged without you noticing. I think the bare minimum is to *NOT log credentials* because they put accounts at risk, instead of "just" leaking some data.

Overall, we should be more privacy-concerned about what is logged, and the `-S` flag should also affect log files. (Arguably, I'd prefer safe by default, with a flag to switch to unsafe).

flatcap:
There were ~50 `#ifdef DEBUG`s in the code.
Most of them surrounded logging.

- 1f15704a drop #ifdef DEBUG
  Compile all the code, unconditionally

- 36d1cdca remove debug/logging from makefile/configure
  We can put back `--debug` if/when we need it

- bb21a8f7 raise logging levels for ex-debug code
  All the `mutt_debug`s that were within a `#ifdef DEBUG` block have been raised to level 5.
  The developer clearly didn't intend the user to see these messages.

@neomutt/reviewers 
There's one `DEBUG` left, which now can't be enabled from `./configure`.
If debug isn't enabled, `crypt_valid_passphrase()` calls `disable_coredumps()`.
It's done to protect PGP credentials, but it's not used elsewhere.  Should we:
- Leave the developer to enable `DEBUG` and enable coredumps?
- Always enable it?  No coredumps, ever.
- Delete it?  Coredumps will be determined by the system/ulimit.

flatcap:
Updated:
- autosetup fixed (`--logging` removed)
- dropped "messing-with-logging-levels" commit (I was over-thinking things)

PR #626, opened 2017-06-08
allow source command to load multiple files at once

toogley:
* **What does this PR do?**
     allows the `source` command to take multiple arguments.

Issue #616

### DIFF
--- a/doc/manual.xml.head
+++ b/doc/manual.xml.head
@@ -5485,8 +5485,7 @@ set spam_separator=", "
             readability</title>
             <screen>
 set my_cfgdir = $HOME/mutt/config
-source $my_cfgdir/hooks
-source $my_cfgdir/macros
+source $my_cfgdir/hooks $my_cfgdir/macros
 <emphasis role="comment"># more source commands...</emphasis>
 </screen>
           </example>
@@ -5593,8 +5592,8 @@ mailboxes $my_mx +mailbox3</screen>
       <para>Usage:</para>
       <cmdsynopsis>
         <command>source</command>
-        <arg choice="plain">
-          <replaceable class="parameter">filename</replaceable>
+        <arg choice="plain" rep="repeat">
+          <replaceable class="parameter">file</replaceable>
         </arg>
       </cmdsynopsis>
       <para>This command allows the inclusion of initialization commands from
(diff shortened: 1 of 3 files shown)


toogley:

guyzmo:
I don't think this is really a good idea to include this patch, because having multiple source files in a single line is not much added value.

And it's encouraging the use of the following pattern: `source `echo ~/.mutt/*``.

It might be a better idea to have a special parsing of the file path to support the globbing, and include multiple files without shell expansion. Also Lua scripting could leverage that.

gahr:
While I don't mind the feature per se - I could see a use for that - I would still fail early and stop sourcing at the first error.

flatcap:
I agree with @gahr, it's reasonable and it should fail on the first error.
After all, mutt thinks this config line is OK:
```
set help sidebar_visible sidebar_width = 10 pager_index_lines = 5
```

toogley:
@gahr i addressed your complaints, it works perfectly.

flatcap:
Please limit the clang-format to just your changes (where possible).
I'll sort out the remainder before a release.

toogley:
fixed.

gahr:
... at which point `rc` is not needed and you simply return 0 after the loop ;)

toogley:
@fixed ah, thanks. fixed.

gahr:
Rebased onto master

gahr:
I have tested this branch it works fine.

Perhaps a minor modification would be to turn the`do { ... } while (MoreArgs(token));` into a `while (MoreArgs(token)) { ... }`.

The only difference is that a line with `source` alone in a config file would simply be skipped instead of being reported like this:
```
Error in /home/gahr/.mutt/muttrc, line 169: source: file  could not be sourced. 
```

What do people think?

flatcap:
hehe, I like the "cleanliness" of it, but...
I think I'd stick with `source<enter>` as an error.

gahr:
fair enough :)

flatcap:
I made one slight tweak to the docs to show optional multiple parameters:
```diff
       <para>Usage:</para>
       <cmdsynopsis>
         <command>source</command>
-        <arg choice="plain">
-          <replaceable class="parameter">file1</replaceable>
-          <replaceable class="parameter">file2</replaceable>
+        <arg choice="plain" rep="repeat">
+          <replaceable class="parameter">file</replaceable>
         </arg>
       </cmdsynopsis>
       <para>This command allows the inclusion of initialization commands from
```

Issue #613, opened 2017-06-05
Berkeley DB configure check fails on some systems


qsuscs:
`configure.ac` checks for BDB in a way that fails at least on Gentoo:

- The directory is named like `/usr/include/db5.3`, which is missed by [configure.ac#L803](https://github.com/neomutt/neomutt/blob/044cc65932922c017ecf8913f9b02b2d627d0636/configure.ac#L803)
- The version test in [configure.ac#L810-L813](https://github.com/neomutt/neomutt/blob/044cc65932922c017ecf8913f9b02b2d627d0636/configure.ac#L810-L813) fails because of Gentoo’s multilib setup:
```c
/* This file is auto-generated by multilib-build.eclass
 * as a multilib-friendly wrapper. For the original content,
 * please see the files that are #included below.
 */

#if defined(__x86_64__) /* amd64 */
#       if defined(__ILP32__) /* x32 ABI */
#               error "abi_x86_x32 not supported by the package."
#       else /* 64-bit ABI */
#               include <x86_64-pc-linux-gnu/db5.3/db.h>
#       endif
#elif defined(__i386__) /* plain x86 */
#       error "abi_x86_32 not supported by the package."
#elif defined(__mips__)
// etc.
```

gahr:
@qsuscs Could you please check and see whether the `devel/issue-613` branch fixes the issue for you?

qsuscs:
Yes, although the check is still a bit weird:
```
checking for BerkeleyDB in /usr/local/include//... no
checking for BerkeleyDB in /usr/local/include/db-5.3... no
checking for BerkeleyDB in /usr/local/include/db53... no
checking for BerkeleyDB in /usr/local/include/db5.3... no
checking for BerkeleyDB in /usr/local/include/db-5... no
checking for BerkeleyDB in /usr/local/include/db5... no
checking for BerkeleyDB in /usr/local/include/db-6.2... no
checking for BerkeleyDB in /usr/local/include/db62... no
checking for BerkeleyDB in /usr/local/include/db6.2... no
checking for BerkeleyDB in /usr/local/include/db-6... no
checking for BerkeleyDB in /usr/local/include/db6... no
checking for BerkeleyDB in /usr/local/include/db-4.8... no
checking for BerkeleyDB in /usr/local/include/db48... no
checking for BerkeleyDB in /usr/local/include/db4.8... no
checking for BerkeleyDB in /usr/local/include/db-4... no
[...]

PR #632, opened 2017-06-12
More sophisticated check for BDB version + support for DB6 (non default)

gahr:
This commit enhances configure.ac to actually parse (as in CPP) and
extract the values of DB_VERSION_MAJOR and DB_VERSION_MINOR from the
Berkeley DB include files. The previous approach to just use grep failed
on some system where the includes are just wrappers around the real bdb
headers.

Also, add support for the current db6 version (6.2), although db5 is
still the first one looked for.

Issue #613


gahr:
Waiting for feedback from @qsuscs to see whether this fixes his build on Gentoo.

PR #633, opened 2017-06-14
[WIP] use ~/.mutt/log/$pid-$date by default as log file

toogley:
* **What does this PR do?**

    * add new config variable `debug_dir`, which defaults to `~/.mutt/log`
    * create directory tree of `debug_dir` if it doesn't exist
    * use the pid of the current process and the current date as default for the log filename.

    * part of #625 

*  **TODO**

    * fix `restart_debug`
        * When debug_file is changed, start_debug() is not executed.
    * fix documentation.

### DIFF
--- a/globals.h
+++ b/globals.h
@@ -235,6 +235,7 @@ WHERE short NntpContext;
 #endif
 
 WHERE short DebugLevel;
+WHERE char *DebugDir;
 WHERE char *DebugFile;
 
 WHERE short MenuContext;
--- a/init.c
+++ b/init.c
@@ -32,6 +32,7 @@
 #include <stdlib.h>
 #include <string.h>
 #include <sys/stat.h>
+#include <sys/types.h>
 #include <sys/utsname.h>
 #include <unistd.h>
 #include <wchar.h>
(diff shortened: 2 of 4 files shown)


toogley:
@gahr for what exactly?

gahr:
I just don't understand what problem you're trying to solve and what use case you're addressing.
Do you expect people to want to keep log files around for a long time (more than 3 invocation, that is)?

toogley:
* i don't delete my ~/.muttdebug files. That means, when i debug things, i have to open every ~/.muttdebug[0..3]  to find out which is the correct one. i know i can change this, but it think it makes sense to change the default - because date sytax doesn't .hurt and is much more expressive. (you can find the most current file much easier).
* creating a new config variable `debug_dir` was just easier for me to implement. But I can of course remove this option and solely rely on `debug_file'
* prepending the log files with the current PID of the program means, multiple mutt instances cannot contaminate logfiles from other mutt instances.

gahr:
Thanks for the clarification. Your use case does indeed make sense.

toogley:
$pid-$date is not enough for the case where user run nmutt multiple times during the day. Is $pid-$date-$time (including seconds) better ?

Additionally: users probably will confuse what each number means (Was the $pid first or last number?) , they will have to look it up very often. ==> Would it make sense to change the filename to P${pid}D${date}T${time} ?

flatcap:
This PR needs work, but has been dead for quite a while.
For now, I'm closing the PR and I have archived the branch to the [neomutt-old](https://github.com/neomutt/neomutt-old) repo.

Issue #556, opened 2017-04-28
libreoffice odt file not open with proper application


sassmann:
As a test create a file with libreoffice writer, save it, add it as attachment to a new mail.
The file is recognized as application/octet-stream.
`  A     2 /dev/shm/test.odt              [applica/octet-stre, base64, 7,8K]`
in .muttrc I have set
`mime_lookup application/octet-stream application/X-Lotus-Manuscript`
so that application/octet-stream should be looked up via /etc/mime.types instead of mailcap.
Also note what xdg-mime says about the file
```
xdg-mime query filetype /dev/shm/test.odt
application/vnd.oasis.opendocument.text
```
From /etc/mime.types
application/vnd.oasis.opendocument.text                         odt

Back in mutt, when trying to view the attachment while editing new mail it is shown in plaintext giving the error "No matching mailcap entry found.  Viewing as text."

Expected result: Open the document with libreoffice
Version: NeoMutt 20170421 (1.8.2)

christopher-john-czettel:
# REPLICATION

I tried to replicate your problem with my debian stretch default neomutt (NeoMutt 20170113 (1.7.2)) and config - no luck.

After that I built NeoMutt 20170421 (1.8.2) directly from the git source. (`./prepare; make`) and was able to reproduce your bug (even without the `mime_lookup` option running `./mutt -nF /dev/null -d 9`). I noticed that the mime-type was already set incorrectly to `application/octet-stream` whilst the attaching process.

# ANALYSIS

A short attached strace unveiled the problem:

```
...
rt_sigaction(SIGINT, {sa_handler=0x56436798a110, sa_mask=[], sa_flags=SA_RESTORER|SA_RESTART, sa_restorer=0x7fd11f9f2030}, NULL, 8) = 0 
open("/home/chris/.mime.types", O_RDONLY) = **-1 ENOENT (No such file or directory)**
open("/usr/local/etc/mime.types", O_RDONLY) = **-1 ENOENT (No such file or directory)**
open("/usr/local/share/mutt/mime.types", O_RDONLY) = **-1 ENOENT (No such file or directory)**
stat("/home/chris/mutt.odt", {st_mode=S_IFREG|0640, st_size=7998, ...}) = 0 
open("/home/chris/mutt.odt", O_RDONLY)  = 4    
fstat(4, {st_mode=S_IFREG|0640, st_size=7998, ...}) = 0 
lseek(4, 0, SEEK_SET) 
...
```
[...]

Issue #635, opened 2017-06-15
log when no mime.types file is found


toogley:
See https://github.com/neomutt/neomutt/issues/461#issuecomment-285422963

flatcap:
Fixed by #637

PR #637, opened 2017-06-16
look in /etc/ for mime.types + message user if no mime.types is found.

christopher-john-czettel:
see issue  #556.

### DIFF
--- a/sendlib.c
+++ b/sendlib.c
@@ -956,6 +956,7 @@ int mutt_lookup_mime_type(struct Body *att, const char *path)
   char subtype[STRING], xtype[STRING];
   int szf, sze, cur_sze;
   int type;
+  bool found_mimetypes = false;
 
   *subtype = '\0';
   *xtype = '\0';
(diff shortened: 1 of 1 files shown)


christopher-john-czettel:
probably also closes: log when no mime.types file is found #635

sendlib.c
```diff
@@ -956,6 +956,7 @@ int mutt_lookup_mime_type(struct Body *att, const char *path)
   char subtype[STRING], xtype[STRING];
   int szf, sze, cur_sze;
   int type;
+  int found = 0;
```
toogley: i think, this can be changed into a boolean value.

christopher-john-czettel:
@gahr the code below the switch case keeps on searching until all files have been parsed and only remembers the last instance of (mime-)type and xtype.

The order was:
```
/etc/mime.types
~/.mime.types
SYSCONFDIR/mime.types
PKGDATADIR/mime.types
```
Therefore the mime.types in SYSCONF/PKGDATADIR would override any changes in the users ~/.mime.types

I changed to order to the following:
```
/etc/mime.types
SYSCONFDIR/mime.types
PKGDATADIR/mime.types
~/.mime.types
```
It should now work as expected.

christopher-john-czettel:
Although I am not quite sure if this change (although making sense) could break someones config, as now the users .mime.types is higher rated.

flatcap:
I've clang-formatted and squashed (no code changes).

PR #641, opened 2017-06-20
notmuch: drop VirtIncoming

flatcap:
NotMuch kept its mailboxes in a separate list, `VirtIncoming` rather than `Incoming`.
This lead to some code duplication and complex logic in the sidebar.

1. be36e9f tidy some buffy code
2. c58c838 split mutt_parse_mailboxes into mutt_parse_unmailboxes
3. 5dca538 drop VirtIncoming

From a user's point-of-view nothing has changed, you still use `<sidebar-toggle-virtual>` to switch between normal and virtual mailboxes.  (For now)

I've tested:
- adding normal mailboxes
- removing a single normal mailbox by shortcut `=mailbox`
- removing a single normal mailbox by path `/home/user/mail/mailbox`
- removing all normal mailboxes `*`
- adding virtual mailboxes
- removing a single virtual mailbox by description `neomutt`
- removing a single virtual mailbox by short path `notmuch://?query=neomutt`
- removing a single virtual mailbox by canonical path `notmuch:///home/user/mail/linode?query=neomutt`
- removing all virtual mailboxes `*`

... but I could do with a NotMuch expert giving it a try.

---

This is step one.  What comes next is open to discussion...

I'd like to see a fake mailbox used as a container for the virtual mailboxes.
This mailbox would live in a unified sidebar.

### DIFF
--- a/browser.c
+++ b/browser.c
@@ -738,10 +738,10 @@ static int examine_directory(struct Menu *menu, struct BrowserState *state,
 #ifdef USE_NOTMUCH
 static int examine_vfolders(struct Menu *menu, struct BrowserState *state)
 {
-  struct Buffy *tmp = VirtIncoming;
-
-  if (!VirtIncoming)
+  struct Buffy *tmp = Incoming;
+  if (!tmp)
     return -1;
+
   mutt_buffy_check(false);
 
   init_state(state, menu);
--- a/buffy.c
+++ b/buffy.c
@@ -519,8 +519,24 @@ int mutt_parse_mailboxes(struct Buffer *path, struct Buffer *s,
 
   while (MoreArgs(s))
   {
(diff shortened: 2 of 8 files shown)


guiniol:
I tried it and all my usual tricks seem to work fine.

somini:
I think I detected a bug fixed by this branch.

When you add a virtual mailbox to a sidebar that shows regular mailboxes, the virtual mailbox will appear mixed with the regular mailboxes. If you `<sidebar-toggle-virtual>` twice, the virtual mailboxes correctly disappear. `<sidebar-toggle-virtual>` once and they appear (this is expected behaviour).

flatcap:
@somini Thanks, good to know :-)

flatcap:

PR #649, opened 2017-06-25
Add Cc, In-Reply-To, and References to default mailto_allow

### LINKED ISSUE #115: Default mailto_allow should include Cc, In-Reply-To, and References
Mutt's default value for mailto_allow includes only Subject and Body.  I'd like to suggest changing the default to also include Cc, In-Reply-To, and References.

In-Reply-To and References allows mailing list archives to include mailto links that won't break threading.  Debian mailing lists and the Debian bug-tracking system include such links.  Cc allows those links to mail both the list and the author of the mail.

I don't see any obvious way that allowing those additional headers could cause harm, and I think allowing them by default would improve usability.

julian-klode:
These headers are quite useful for replying to mailing lists
via mailto links, as they don't break threading.

Debian has been shipping these in the /etc/Muttrc file since
September 2016 without much issues.

Fixes #115


julian-klode:
Hmm, does not actually seem to work yet. I thought it did, because that's what's been shipped in Debian, but apparently it's not enough.

julian-klode:
I take that back, it works, my test url was just broken.

julian-klode:
We can also do this by patching the default Muttrc, not sure what's nicer.

PR #660, opened 2017-06-30
Handle native Solaris GSSAPI.

0-wiz-0:
Signed-off-by: Thomas Klausner <wiz@NetBSD.org>

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -355,7 +355,7 @@ main ()
 
 	AC_CHECK_FUNC(initscr,,[
 	cf_ncurses=""
-	for lib in ncursesw ncurses
+	for lib in ncursesw ncurses curses
 	do
 		AC_CHECK_LIB($lib, waddnwstr, [cf_ncurses="$lib"; break])
 	done
--- a/configure.ac
+++ b/configure.ac
@@ -489,7 +489,7 @@ if test "$gss_prefix" != "no"; then
 		AC_CACHE_SAVE
 		AC_MSG_RESULT([GSSAPI libraries not found])
 	fi
-	if test "$GSSAPI_IMPL" = "Heimdal"; then
+	if test "$GSSAPI_IMPL" = "Heimdal" || test "$GSSAPI_IMPL" = "Solaris"; then
 		AC_DEFINE(HAVE_HEIMDAL,1,[ Define if your GSSAPI implementation is Heimdal ])
 	fi
 	CPPFLAGS="$CPPFLAGS $GSSAPI_CFLAGS"
--- a/curs_main.c
+++ b/curs_main.c
@@ -50,6 +50,9 @@
 #include "protos.h"
 #include "sort.h"
 #include "thread.h"
+#ifndef USE_SLANG_CURSES
+#include <term.h>
+#endif
 #ifdef USE_SIDEBAR
 #include "sidebar.h"
 #endif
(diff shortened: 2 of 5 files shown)


flatcap:
Beyond this patch, how is NeoMutt on Solaris?

A while back we created a list of platforms on which NeoMutt was supported.
- Linux, *BSD, MacOS -- these we could easily support and test
- AIX, Ultrix, etc -- clearly obsolete

Solaris fell into the middle.  We couldn't find anyone to ask (we didn't look very hard :-)
Is there a NeoMutt-using Solaris community out there?

0-wiz-0:
9941939 is also needed for compilation on Solaris, while 528c7d9 removes a warning there (helps with -Werror).
afc3f4c fixes a configure error on NetBSD, and d238ff1 fixes a compilation error on NetBSD.
decdcc5 is just a cleanup (fixing an autoconf/automake warning).

neomutt is in pkgsrc, and Illumos builds all of their packages from pkgsrc. The package has been fixed for Illumos recently, so there are no download numbers for neomutt yet.

configure.ac
```diff
@@ -46,7 +46,6 @@ fi
 AC_PROG_CPP
 AC_PROG_MAKE_SET
 AC_PROG_INSTALL
-AC_PROG_MKDIR_P
```
flatcap: We use `$(MKDIR_P)` in a couple of Makefiles.
The build works without them (for me).
Are you sure it's safe for all users?

curs_main.c
```diff
@@ -50,6 +50,7 @@
 #include "protos.h"
 #include "sort.h"
 #include "thread.h"
+#include <term.h>
```
flatcap: I guess this is for `tigetstr`, etc.
`<term.h>` is part of the curses install but doesn't exist in a slang build.

Putting this around it might do, for now:
```
#if defined(HAVE_NCURSESW_NCURSES_H) || defined(HAVE_NCURSES_H) || defined(HAVE_NCURSES_NCURSES_H)
#endif
```

flatcap:
> neomutt is in pkgsrc, and Illumos builds all of their packages from pkgsrc.

Who manages the neomutt package in pkgsrc/Illumos?

> there are no download numbers for neomutt yet.

If we can find a few people to test NeoMutt regularly, we can declare Solaris as "supported".

configure.ac
0-wiz-0: You're right. Forget about this part, I misinterpreted a warning, the obsolete macro use is in a gettext m4 file. Can I rebase and remove this patch and force-push, or is there a better way to get rid of this?

curs_main.c
0-wiz-0: I don't understand. `tigetstr()` is, if I read the code, used unconditionally, so the header file needs to be included unconditionally as well.
Addiitonally, on NetBSD, neither of the defines is defined–we're using curses there, not ncurses.

0-wiz-0:
I maintain the pkgsrc package, but I'm not using Illumos. jperkins committed the Solaris fixes, but he's currently still using mutt because there were no binary packages for neomutt so far.

configure.ac
flatcap: Yes, we're happy for people to force-push to PRs to update them.

curs_main.c
flatcap: It exists in both curses and slang, but in different headers.
In curses, `tigetstr()` is in `<term.h>`, but in slang it's (a null op) in `slcurses.h`

0-wiz-0: I've added the header except for the slang build. Ok that way?

flatcap:
OK, if there's anything we can do to streamline getting neomutt built, let me know.

PR #661, opened 2017-07-03
Fixes when using colors

fbellet:
This PR fixes some colors misbehavior:
 * [**merged**] the background color of the "tree" element, used to display mail threads is meaning less, because it mixes with the background color of the indexes subject in unexpected ways (and this latter background color is variable). It is better to create new color pairs dynamically, using the fixed "tree" foreground color, and whatever background color the "tree" element is supposed to use for a given message.

  * [**workaround found**] the color having "bright" in its color name should not always have the bold attribute, because some colors schemes like solarized don't have the notion of color "brightness", which just represents of position in the color palette.

 * [**merged**] the flags modifications in a collapsed sub-tree should modify the root message of the tree **last**, because some color patterns updates that immediately follows, may take into account the status of the other hidden messages in this collapsed sub-tree, for example `~v~(~D)`

### DIFF
--- a/color.c
+++ b/color.c
@@ -426,11 +426,7 @@ static int parse_color_name(const char *s, int *col, int *attr, int is_fg, struc
 
   if (is_bright)
   {
-    if (is_fg)
-    {
-      *attr |= A_BOLD;
-    }
-    else if (COLORS < 16)
+    if (COLORS < 16)
     {
       /* A_BLINK turns the background color brite on some terms */
       *attr |= A_BLINK;


flatcap:
Welcome, @fbellet!  Thanks for the patches.

flatcap:
I've merged your tree-colour fix.  Thanks.

The other two, I'm not so sure about.

> the color having "bright" in its color name should not always have the bold attribute

I agree, it _shouldn't_ be bold, but generations of mutt users may expect this behaviour.
If this were vim, we could suggest they add `cterm=bold`, but there's no way in mutt.

I suggest changing the theme to use to `colorNNN` notation.
It works, even if you only have a 16 colour palette.
Here's some sed to help:

```sed
s/\<brightblack\>/color8/
s/\<brightred\>/color9/
s/\<brightgreen\>/color10/
s/\<brightyellow\>/color11/
s/\<brightblue\>/color12/
s/\<brightmagenta\>/color13/
s/\<brightcyan\>/color14/
s/\<brightwhite\>/color15/
```

Fixing the colour handling in mutt is one of the most-asked-for enhancements.
Unfortunately, it would probably require massive changes.

> the flags modifications in a collapsed sub-tree should modify the root message of the tree last

This sounds good, but I can't find an example that shows the difference.
I've tried lots of different flags within lots of folders and compared your branch to [master].

Can you give me an example?

What flags should be set on the top message?
What flags should be set on a child message?
What `color`s should be applied?

fbellet:
On 07/05/17 at 01:31pm, Richard Russon wrote:
> > the flags modifications in a collapsed sub-tree should modify the root message of the tree last
> 
> This sounds good, but I can't find an example that shows the difference.
> I've tried lots of different flags within lots of folders and compared your branch to [master].
> 
> Can you give me an example?
> 
> What flags should be set on the top message?
> What flags should be set on a child message?
> What `color`s should be applied?

with this muttrc file:

set sort=threads
color normal        brightyellow    default
color index         magenta         white           "~v~(~D)"
color index         blue            white           "~v~(~N!~D)"

let's consider a mailbox with new messages, with threads collapsed, the
top message should appear "blue", then delete the thread with ^D
(keeping it collapsed), the top message remains "blue", instead of
changing to "magenta". 

-- 
fabrice

flatcap:
Thanks, that works for me.

fbellet:
On 07/05/17 at 01:31pm, Richard Russon wrote:
> I suggest changing the theme to use to `colorNNN` notation.
> It works, even if you only have a 16 colour palette.

Yes, the workaround works for me, that's a good idea!

-- 
fabrice

PR #662, opened 2017-07-04
fix gpgme S/MIME non-detached signature handling

stbuehler:
- gpgsm ends the session after the decrypt error, so restart gpgsm when retrying with "maybe_signed".
- evolution sends encrypted + signed S/MIME mails with non-detached signatures

### DIFF
--- a/ncrypt/crypt_gpgme.c
+++ b/ncrypt/crypt_gpgme.c
@@ -1705,6 +1705,9 @@ restart:
       {
         maybe_signed = true;
         gpgme_data_release(plaintext);
+        /* gpgsm ends the session after an error; restart it */
+        gpgme_release(ctx);
+        ctx = create_gpgme_context(is_smime);
         goto restart;
       }
     }


stbuehler:
Theoretically `mutt_is_application_smime` does detect that the contained message is not encrypted, but signed only (by looking at the `smime-type` parameter); but this information is not passed to `decrypt_part`, which is always used to "decode" the non-multipart `application/[x-]pkcs7-mime` messages.

Adding the following lines at the beginning of `decrypt_part` fix this:
```c
  /* don't try to decrypt S/MIME messages which are only signed */
  if (is_smime && (SMIMEENCRYPT != (SMIMEENCRYPT & mutt_is_application_smime (a))))
    maybe_signed = 1;
```

stbuehler:
I looked at strace logs again, and found that gpgsm does NOT terminate the connection; mutt (i.e. libgpgme) does.

This is because libgpgme treats gpgsm "protocol" errors as IO errors; see [`status_handler()` in `engine-gpgsm.c:758`](https://github.com/gpg/gpgme/blob/aea2c168fc9c12148181dbcc33d7085aad8e6d90/src/engine-gpgsm.c#L758): it returns the error as function result. "operational errors" should be returned through the [`io_cb_data`](https://github.com/gpg/gpgme/blob/aea2c168fc9c12148181dbcc33d7085aad8e6d90/src/wait.h#L82) struct (passed by reference) instead. The assuan-engine in [`engine-assuan.c:588`](https://github.com/gpg/gpgme/blob/aea2c168fc9c12148181dbcc33d7085aad8e6d90/src/engine-assuan.c#L588) demonstrates the correct usage. IO errors lead to session termination through [`wait-private.c:119`](https://github.com/gpg/gpgme/blob/aea2c168fc9c12148181dbcc33d7085aad8e6d90/src/wait-private.c#L119) and [`gpgsm_cancel()` in `engine-gpgsm.c:L196`](https://github.com/gpg/gpgme/blob/aea2c168fc9c12148181dbcc33d7085aad8e6d90/src/engine-gpgsm.c#L196).

All this means that the issue is actually a libgpgme bug, but it is probably better to be on the safe side and restart the gpgme context after failures (i.e. what this pull request does).

Regarding my extra three lines: it is probably better to handle them as a separate pull request (if wanted at all); if the return value of `mutt_is_application_smime` is "safe", the patch should be safe - but otherwise it might break when `mutt_is_application_smime` says the content is not encrypted, but it actually is.  And those lines don't fix a bug, they just optimize the handling when the content is not encrypted.

stbuehler:
Related bug reports:
- gpgme: https://dev.gnupg.org/T2919
- mutt: https://dev.mutt.org/trac/ticket/3808

PR #663, opened 2017-07-05
Small build improvements

flatcap:
These changes are just decluttering.

1. 9f428e1 build: move smime_keys into contrib
2. 1102152 build: drop instdoc script
3. fa4b54e build: remove unnecessary steps
4. 840814f build: tidy makefile

Edit: Dropped the "magic" install dir commit


d-k-c:
Overall looks good, but I don't like the change introduced in d8e3e4e. I think Mutt doesn't need more magic/non-obvious behaviour, and it's not obvious to me why configuring with `--enable-debug` should alter the default install path.

Also, maybe change `AM_DISTCHECK_CONFIGURE_FLAGS` to one option per line. It would take a bit more room but easier to keep it sorted without introducing unnecessary diff.

flatcap:
> I think Mutt doesn't need more magic/non-obvious behaviour

Yeah, fair enough.  It was a little shortcut, partly to reclaim `--enable-debug` for developers (but mostly for myself :-)  I'll drop it.

> change AM_DISTCHECK_CONFIGURE_FLAGS to one option per line

done

PR #666, opened 2017-07-08
compile-time output: use two lists

ThomasAdam:
**PoC:  I'm really not anticipating this to be merged, but if it means SvenG is appeased, maybe we'd all appreciate it.**

When asking NeoMutt for a list of compile-time features, the current
implementation assumes a mixture of options which can be enabled through
./configure, and some which default to on, without compile-time
exclusion.

Instead, separate this output in NeoMutt.


flatcap:
I've changed the order of the output.  The strings that follow the options are user-configurable.
Also, I added a check in `feature_enabled()` otherwise the change would have broken the `ifdef` command.

I've committed this, but the `mutt -v` output really needs a complete rework.

Issue #667, opened 2017-07-10
configure without slang does the opposite


nobrowser:
I was trying to setup a "minimal" build, so I did

`./configure --without-slang ${WITHOUT_MANY_OTHER_THINGS} `

And I got:

```
configure: Summary of build options:

  Version:           20170707
  Host OS:           linux-gnu
  Install prefix:    /usr/local
  Compiler:          gcc
  CFlags:            -Wall -pedantic -Wno-long-long -g -O2 -fno-delete-null-pointer-checks 
  LDFlags:            -Lno/lib
  Libs:              -lanl 
  Mutt libs:          -lslang -lm 

  GPGME:             no
  PGP:               no
  SMIME:             no
  Notmuch:           no
  Header Cache(s):   no
  Lua:               no
```
Then I removed the option, and I got:

```
configure: Summary of build options:

  Version:           20170707
  Host OS:           linux-gnu
  Install prefix:    /usr/local
  Compiler:          gcc
  CFlags:            -Wall -pedantic -Wno-long-long -g -O2 -fno-delete-null-pointer-checks 
  LDFlags:           
  Libs:              -lanl 
  Mutt libs:          -lncursesw 

  GPGME:             no
  PGP:               no
  SMIME:             no
  Notmuch:           no
  Header Cache(s):   no
  Lua:               no
```
This is normal autotools madness - a good example of the motivation people have for switching build systems :-P (Not that I'm suggesting it myself - no sir, not me!)

flatcap:
Good find!

It comes from some clumsy conditionals in configure.
One of the two is required, but they don't coordinate.

> This is normal autotools madness

Yes, it is :-)

> ... switching build systems ...

It's a common suggestion, though nobody has come up with a viable alternative yet.
In the mean time, we're trying to make the build: "as simple as possible, but no simpler" -- Einstein.

PR #676, opened 2017-07-17
Fix the configure check for S-Lang

gahr:
Issue: #667


gahr:

gahr:
Thanks for the test, @flatcap

Issue #664, opened 2017-07-06
"color index_subject..." is bleeding to the next line


jmtd:
I have `color index_subject brightwhite  default '~O'`, and if the subject is truncated, then the colour bleeds into the next line down (in my case, over the next email's number):

![neomutt_bug_color_index_subject](https://user-images.githubusercontent.com/63694/27913604-981cbcae-6258-11e7-97bb-88a9fd458291.png)

Note in the example that email 597 has the subject brighwhite, the subject is truncated at the end of the line, and the number 598 is alsounintentionally brightwhite.

By contrast, email 604 matches, the subject is brightwhite, the subject is not truncated, and so email 605 is unmolested. (Email 605 is a mail *from* me, but that is not relevant, elsewhere in this mailbox I can see the bug manifest where this is not the case).

Specific neomutt version is Debian package 1.8.3+neomutt20170609-2

fbellet:
It seems that the function that formats the index string simply truncates the output buffer to the number of columns of the terminal, ignoring the escape sequence that may be located at the end of the input buffer to restore the terminal attributes. Does this patch work for you ?

4nd3r:
similar to #64

flatcap:
Fixed by #685

PR #685, opened 2017-07-24
Fix truncated formating

fbellet:
This patch fixes the leak of index line attributes on the next line due to the truncation of the output buffer based on terminal number of columns, that forgets to include escape sequence from the end of the input buffer (#664). Characters from an escape sequence have w=0, so they'll continue to be added to the output buffer.

### DIFF
--- a/curs_lib.c
+++ b/curs_lib.c
@@ -1199,7 +1199,7 @@ void mutt_simple_format(char *dest, size_t destlen, int min_width, int max_width
     if (w >= 0)
     {
       if (w > max_width || (k2 = wcrtomb(scratch, wc, &mbstate2)) > destlen)
-        break;
+        continue;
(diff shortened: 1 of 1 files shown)


gahr:
Isn't this going to overflow the buffer pointed-to by `p`?

fbellet:
On 07/24/17 at 09:27am, Pietro Cerutti wrote:
> Isn't this going to overflow the buffer pointed-to by `p`?

from my understanding of the code, dest has a size of destlen (1024
bytes initially in my case, decreasing as formating progresses), and p
is a forward moving index in this buffer. So the overflow test for p (in
dest) is (k2 > destlen). 

-- 
fabrice

flatcap:
Very nice.  Thanks.  Sorry it took a while to check.

Issue #603, opened 2017-06-01
notmuch queries with tags containing & need to be double encoded


julian-klode:
I have a tag called Privat/1&1. This tag does not really work. When using `vfolder-from-query`, I have to search `tag:Privat/1%261`, and when directly specifying an URI it is `notmuch://?query=tag:Privat/1%25261`. 

So it seems that & is not url-encoded properly and possibly decoding is wrong too.

Tested versions: NeoMutt 20170306 (1.8.0) on Debian, and commit dfc89660326fbd175e090f505ab5267df6134048

julian-klode:
Oh nevermind, I actually forgot to run the test in git before hitting send. Seems to work fine in the git commit. Let me check some more.

julian-klode:
Ah no, it fails there too, I accidentally searched for it without the `tag:` prefix and got some results back and mistakenly thought it was fixed :(

flatcap:
NotMuch contributors @bbenne10 @paretje @wpettersson
Would you be interested in investigating this?

bbenne10:
Looking at the implementation (specifically [here](https://github.com/neomutt/neomutt/blob/5a951b768eb3f546327da938267c7d2b6b680de7/mutt_notmuch.c#L289)), I can see how this would be problematic in the completely unescaped case (but that's somewhat to be expected), but I'm unsure of the deeper ramifications. I'm at work today, but might be able to check this more deeply this evening. I imagine that part of it is this rather wonky uri parsing, but that should only affect the unencoded string.

WPettersson:
I took a shot at this, couldn't find out how to assign myself though. I decided to use `url_pct_encode` to encode the ampersand. This means that I tweaked `url_pct_encode` to also encode ampersands (as well as `/:%`). Given that `url_pct_encode` is only called in two places, once when encoding a username and once when encoding a password, I figured that this made sense.

Note that when specifying a URI, you still have to use `notmuch://?query=tag:Privat/1%261` to encode the ampersand. I don't see any way around that, the ampersand is used as a delimiter in URIs.

I also couldn't see any format testing suite, so I don't have anything there beyond me stating "I tested it a bit myself" which is rather useless.

Pull request #692

flatcap:
Fixed by #692

PR #692, opened 2017-07-30
Encode URIs created with nm_uri_from_query

WPettersson:
* **What does this PR do?**

Encodes URIs created in nm_uri_from_query using url_pct_encode. This means url_pct_encode is no longer static. This violates the coding style, as this function is now global but doesn't start with 'mutt_'. However, leaving 'mutt_' off seems to be consistent with the other url functions. Can change this if desired.

`url_pct_encode` now also encodes ampersands (`&`s). As `url_pct_encode` is only used in two other places, where it encodes either usernames or passwords, this seems valid.

Addresses issue #603


WPettersson:
My bad, missed a coding style issue. I can create a new squashed PR if needed, or I think Github can do that for you. Not sure what your preferred approach is.

flatcap:
@WPettersson Brilliant!  Thanks for that.
I'll take a look at it shortly.

> I can create a new squashed PR if needed

I can tidy that all up at merge time.

PR #694, opened 2017-07-31
Break up dependencies / Create library

flatcap:
**NOTE**: One important questions need answering before we can commit this.

- What shall we call it?

It's currently the `lib` library, because I was lacking in imagination :-)

It's tempting to call it `libmutt`, but that would be misleading as there's no mutt-specific code in it.
I was thinking: `libsimple`, `libplain`, but I'm open to persuasion.

The files were called `lib_*` to avoid clashes whilst renaming, but the `lib_` prefix can probably be removed on all except `lib_string.h`.

All the functions could probably do with better names, but we'll leave that for another day.

## Changes

1. d5d7551 split out state functions
   move `state_*()` from `muttlib.c` to `state.c`
1. 7871697 split out parameter functions
   move `parameter_*()` from `muttlib.c` to `parameter.c`
1. 9c4bdef create library
   create an empty library `lib/liblib.a`
1. 3d6bbd7 change code to use lib/lib.h
1. d85c3bb move the ascii functions to the library
   `ascii.c` to `lib_ascii.c`
1. ed93571 move the base64 functions to the library
   `base64.c` to `lib_base64.c`
1. c3973b5 move the date functions to the library
   `date.c` to `lib_date.c`
1. c4ee7f1 move the exit functions to the library
   create an overridable `mutt_exit()` function
1. be39c4a move the md5 functions to the library
   `md5.c` to `lib_md5.c`
1. 63e532d move the sha1 functions to the library
   `sha1.c` to `lib_sha1.c`
1. 3a873d3 move the memory functions to the library
   move the memory management functions from `lib.c` to `lib_memory.c`
1. 45896c5 move the message functions to the library
   move the message/error functions into the library
1. 9757e63 move the debug functions to the library
   create a stub `mutt_debug()` in the library
1. 4603725 move the string functions to the library
   move the string management functions from `lib.c` to `lib_string.c`
1. 78538c6 move the file functions to the library
   move the file management functions from `lib.c` to `lib_file.c`
1. 2bb98e7 move the buffer functions to the library
   `buffer.c` to `lib_buffer.c`
1. 83552a3 move the hash functions to the library
   `hash.c` to `lib_hash.c`
1. a77d96d drop lib.c
   `lib.c` is almost empty.  move the remainder into `muttlib.c`
1. 7de7f7c doxygen: update config
   Show off all the new documentation

## Dependencies

The library has zero external dependencies beyond the standard C libraries.
It provides stub versions of `mutt_message()`, `mutt_error()`, etc.
When linking your program, as long as `liblib.a` is last, then the stubs will only get used if you don't provide your own.

## Testing

The independent nature of the library, means you can now test all the function with ease.
Here's a working sample program `test_lib.c`

```c
#include <stdio.h>
#include <stdarg.h>
#include <unistd.h>
#include "lib/lib.h"

int main()
{
  void *ptr = NULL;

  ptr = safe_calloc(1, 128);
  printf("%p\n", ptr);

  FREE(&ptr);
  return 0;
}
```

Compile with:
`gcc -Wall -o test_lib test_lib.c -Llib -llib`

---


flatcap:
I've also documented all the functions, adding doxygen pages to the source.
This gives us a new section in the code docs: https://www.neomutt.org/code/lib.html

flatcap:
I've put some sample code in a new repo: [library-test](https://github.com/neomutt/library-test).
`test_lib` covers each of the library source files
`test_hcache` creates a header cache entry and retrieves it (the code has a few dependencies)

flatcap:
I've done a quick bit of experimenting and it seems that I can strip the `lib_` prefix from all the files except `lib_string.h`.  I've renamed that to `string2.h`.

On reflection, I'll change the library to `libmutt`.
It's no big deal if we need to rename it in the future.

Issue #683, opened 2017-07-20
`change-vfolder` doesn't work anymore


guiniol:
The first bad commit seems to be 1116748c6e52bdf8c4c0b457f4a7a26346dc3b30 "drop VirtIncoming"

# Setup
1. Define a virtual mailbox e.g. 
   `virtual-mailboxes "inbox" "notmuch://?query=tag:inbox and NOT tag:archive"`
2. Add a macro to open that virtual mailbox (not required but makes it easier) e.g.
   `macro index X '<change-vfolder>inbox<enter>'`

# Expected behaviour
The virtual inbox is opened

# Actual behaviour
Error message `inbox is not a mailbox.`

dsdeiz:
I am experiencing this as well. One thing I noticed is that it looks like `change-vfolder` seems to like URI now instead of virtual mailboxes. An example is `macro index X '<change-vfolder>notmuch://?query=tag:inbox'`. It also doesn't seem to like spaces in the URIs e.g. `macro index X '<change-vfolder>notmuch://?query=tag:inbox and NOT tag:archive'`.

Right now my `folder-hook`s don't work.

esovetkin:
I have created a pull request  #695  that fixes this issue

flatcap:
Thanks for the fix @esovetkin!  Merged.

PR #695, opened 2017-08-02
fix `change-vfolder` (issue #683)

esovetkin:
nm_description_to_path: fix a mailbox type                                                                                                                                               
                                                                                                                                                                                              
In nm_description_to_path a mailbox type is supposed to be MUTT_NOTMUCH,                                                                                                                  
since the function queries the path of a *notmuch* mailbox using                                                                                                                          
a description string.                                                                                                                                                                     
                                                                                                                                                                                              This commit fixes issue #683.

### DIFF
--- a/mutt_notmuch.c
+++ b/mutt_notmuch.c
@@ -2164,7 +2164,7 @@ int nm_description_to_path(const char *desc, char *buf, size_t bufsz)
 
   for (struct Buffy *b = Incoming; b; b = b->next)
   {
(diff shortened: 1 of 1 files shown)


d-k-c:
Thanks for this proposed patch. It would be better if the commit message could more expressive and state what the commit actually does, it helps us navigate through the history of the project more easily.
This [blog post](https://chris.beams.io/posts/git-commit/) has good tips.

flatcap:
Merged.

Issue #699, opened 2017-08-04
Kill ascii.h functions


gahr:
* **feature requests**

* Why do you not like the current state?

The `ascii_strcasecmp` and `ascii_strncasecmp` functions declared in `lib/ascii.h` are problematic for a couple of reason:

  1. they do not state what they do, should any characters not be confined within the ASCII character set
  2. they use the standard C function `tolower` (to case-convert the strings to lowercase before comparison), which does take the current `LC_CTYPE` locale into consideration

As such, I'd say that the `ascii` prefix is incorrect at best, and useless anyway.

* What would you like to see to change?

I propose to remove the two aforementioned functions and replace any calls to the correspective Std 1003.1, 2001 POSIX.1's `strcasecmp` and `strncasecmp`.

The third function of `ascii.h`, `ascii_strlower` is used only within `ncrypt/crypt_gpgme.c`. I suggest to convert it to a macro or static inline function and move it there.

gahr:
Edit: I suggest to replace `ascii_strcasecmp` and `ascii_strncasecmp` by their `mutt_strcasecmp` and `mutt_strncasecmp`, which in turn call the POSIX one after wrapping their argument through `NONULL`.

flatcap:
Yes, using the NONULL wrappers sounds good.  Go for it.

PR #700, opened 2017-08-04
Remove ascii_* functions

gahr:
Issue #699


flatcap:
I suggest leaving `mutt_strlower()` in the library.
It might only be used in one place, but it's a useful function.
If it's in the library, it's more visible should anyone need it.

The library does mean we need to do an regular audit to make sure everything's still needed.
I'll make a note.

gahr:
done :)

PR #709, opened 2017-08-07
Accept abbreviated commands

### LINKED ISSUE #701: Please accept :se as a synonym for :set
This is a minor item, but a very regularly encountered pain point for me.

Vim accepts commands starting with a colon, just like mutt does. However, vim allows abbreviating commands using any unambiguous prefix. In particular, you can write `:se` instead of `:set`. I regularly find myself typing this in mutt and expecting it to work (e.g. `:se mime_forward=yes`), only to get an error, and have to go back and change it to `:set`.

I don't know if mutt should *in general* move to accepting unambiguous prefixes, but please consider accepting `:se` in particular as a synonym for `:set`.

flatcap:
Many commands could be abbreviated and still be unique, e.g.
- `:co` instead of `:color`
- `:so` instead of `:source`

This PR allows all the shortest abbreviations for commands, with one exception: `:se` is allowed for `:set`

Closes #701

### DIFF
--- a/init.c
+++ b/init.c
@@ -3317,16 +3317,15 @@ static int parse_source(struct Buffer *tmp, struct Buffer *token,
 
 /**
  * find_command - Find a command matching a (partial) string
- * @param cmd String to compare
+ * @param cmd     String to compare
+ * @param cmd_len Length of string
  * @retval num Index into Commands
  * @retval  -1 Error, no match
  */
-static int find_command(const char *cmd)
+static int find_command(const char *cmd, int cmd_len)
 {
   if (!cmd)
     return -1;
-
-  int cmd_len = mutt_strlen(cmd);
 
   for (int i = 0; Commands[i].name; i++)
   {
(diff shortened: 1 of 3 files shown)


gahr:
A slightly dangerous thing is that the addition of any command must now trigger a review of unrelated lines in init.h to catch potential clashes of minimum prefix. Is that an issue? Just a thought..

flatcap:
That's a good point.
A simple solution would be to add comments to that effect.
Alternatively, we could generate the lengths at startup.
(the command list is now sorted, so it would be fairly simple to do)

gahr:
+1 for the runtime generation of prefixes. However, I'm not fond of exceptions (for set).
How about we just alias se to set? Is the whole set (no pun intended) really needed?

flatcap:
I've looked through my logs and I only tend to `:set` and `:source`.  I'd benefit from `:so`.

Do we need the others?
Not really, but I don't see any harm in having them.

As for the runtime generation, I'll have a think how to do it.
(I used vim and awk for the initial list)

gahr:
I really don't know, I feel a bit uncomfortable with this one. I think what's worrying me is that prefixes can change between compilations - addition / removal of commands or potentially even different configure options - which will make them useless or even incorrect at muttrc level.
Having slept on it some more, I think I'd prefer to hardcode just the set of prefixes people require as aliases.
Just my .2CHF ;)

flatcap:
This PR works, but it's been decided to solve the problem in another way.
See the comments for #701

toogley:
@flatcap when it has been decided to solve it differently, is there a reason why this PR is still open then?

flatcap:
> is there a reason why this PR is still open

It contains a _little_ useful info and the PR's a good template of where the code needs changing.
You can kill it next week.

gahr:
An alternative approach was preferred.

PR #713, opened 2017-08-09
include term.h from wherever we find ncurses.h

flatcap:
Solaris needs `term.h` for `tigetstr()`, but `term.h` doesn't always live in `/usr/include`

The only tests that need to be done are:
- Does Ncurses compile?
- Does Slang compile?

It works in Fedora and Debian, but I'd appreciate feedback for any systems you can test.

/cc @0-wiz-0

### DIFF
--- a/curs_main.c
+++ b/curs_main.c
@@ -58,7 +58,7 @@
 #include <ncursesw/term.h>
 #elif defined(HAVE_NCURSES_NCURSES_H)
 #include <ncurses/term.h>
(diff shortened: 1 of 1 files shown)


gahr:
Builds & starts up fine both with ncurses and slang on FreeBSD (y)

d-k-c:
Compile with ncurses and slang on Arch Linux

ng-0:
With the patch applied, it still builds on GNU Guix with ncurses and starts up.

PR #714, opened 2017-08-09
Small cleanups to configure.ac

gahr:
* Remove old checks that are not relevant anymore
* Group some `AC_ARG_WITH` in the same section
* Document a bit `--with-slang` / `--with-ncurses` interaction
* Group the generation of `OPS` in one place


flatcap:
I've made one small tweak to the OPS generation.

The OPS are now `cat`d in the same order in `configure.ac` and `Makefile.am`
This means that `keymap_defs.h` and `keymap_alldefs.h` contain their symbols in the same order.
This isn't _required_ -- it's just a nicety.

(you can squash this into the first commit)

Issue #551, opened 2017-04-26
Maildir creation fails (errno=21)


jwynn6:
When trying to save a message to a non-existent maildir  the interface prompts to create a new directory and then when yes is selected, it errors with an errno=21, indicating the directory already exists.  Which it does not. 

NeoMutt 20170421 (1.8.2)
System: Linux 4.10.6-1-ARCH (x86_64)
libidn: 1.33 (compiled with 1.33)
hcache backends: gdbm

flatcap:
Hmm... it works for me.
```
# muttrc
set mbox_type = maildir
```
Docs for [`$mbox_type`](https://www.neomutt.org/guide/reference#mbox-type)

I tried:
- <kbd>s</kbd> `=new-dir`
- <kbd>s</kbd> `=new-dir/`
- <kbd>s</kbd> `/path/to/new-dir`
- <kbd>s</kbd> `/path/to/new-dir/`

If `mbox_type` is set to its default `mbox`, then I can repeat your problem by specifying a non-existent dir with a trailing slash:

- <kbd>s</kbd> `=new-dir/`
> /home/user/mail/new-dir/: Is a directory (errno = 21)

christopher-john-czettel:
@jwynn6 I was able to reproduce your problem (with `$mbox_type` set to maildir as well as mbox):

 <kbd>s</kbd> `=new-dir/subdir`

erroneously reports `/home/user/mail/new-dir/subdir: Is a directory (errno = 21)`
even though it not exists. 

It seems the code that should create the actual dir in `mutt_save_confirm()` was missing. I have hopefully fixed this with commit 5cc4bfc. 

Please be so kind, check if everything now works as expected and let me know.

@flatcap: You actually found another bug - gratz! ;-)

 <kbd>s</kbd> `=new-dir/` 

The `"/home/user/mail/new-dir/: Is a directory (errno = 21)` error occurs when $mbox_type is set to `MUTT_MBOX` as this save call needs a filename and not a directory name (opposed to `MUTT_MAILDIR`).

The error message should be more user friendly like "is a directory, need a filename for mbox", so I also committed e1db65ec.

PR #720, opened 2017-08-16
trim trailing slash from completed dirs

### LINKED ISSUE #719: Neomutt thinks `mbox_type` is `mbox` when it should be `maildir`
This occurs with neomutt from the gentoo_neomutt overlay for gentoo.

I store my mail locally, in maildir format, and until recently I
had never set the `mbox_type` variable in my config. Prior to commit
eebf72008a790dc3214d89beeea4daf815c2b5bc , this wasn't a problem,
but since that commit, `:set ?mbox_type` shows `mbox`, and
`save-message` has become unusable.

In particular, if I'm in maildir foo and want to move a message to
bar, I `s`, then `=bar/`, then `Return`, on which I get the error
message

    '=bar/' is a directory, need a filename for mbox.

The fix is to add `set mbox_type = maildir` to my config somewhere,
but that used to not be necessary (and since I don't want mbox, I
wouldn't have guessed to set a variable with "mbox" in the name).

flatcap:
When saving a email, selecting a maildir directory should override the setting of $mbox_type.

- default mailbox (`$mbox_type`) is set to `mbox`
- save an email to a **Maildir** directory with a trailing '/'

Before #551, NeoMutt would do the "right thing" and safe it in Maildir format.
Since, it complains: '=bar/' is a directory, need a filename for mbox.

By trimming the trailing '/' from the completed name, means:
- maildirs are selected correctly
- non-existent paths respect `$mbox_type`

Fixes #719

### DIFF
--- a/addrbook.c
+++ b/addrbook.c
@@ -65,7 +65,7 @@ static const char *alias_format_str(char *dest, size_t destlen, size_t col, int 
       mutt_format_s(dest, destlen, fmt, alias->name);
       break;
     case 'r':
-      adr[0] = 0;
+      adr[0] = '\0';
       rfc822_write_address(adr, sizeof(adr), alias->addr, 1);
       snprintf(tmp, sizeof(tmp), "%%%ss", fmt);
       snprintf(dest, destlen, tmp, adr);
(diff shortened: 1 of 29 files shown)


flatcap:
Thanks.  I've removed the duplicate check and changed the assignment to a `\0`

PR #723, opened 2017-08-20
contrib: Add guix-neomutt.scm.

ng-0:
* **What does this PR do?**

guix-neomutt.scm can be used to build the git checkout of neomutt at HEAD without hash checking in the build environment of GNU Guix. This is similar to what other software packages already do, like Mediagoblin, GNUnet, etc. There's no formalized way how to do it yet, but it helps with development on those platforms.

     for that)

   - All builds are passing

### DIFF
--- a/contrib/guix-neomutt.scm
+++ b/contrib/guix-neomutt.scm
@@ -1,0 +1,49 @@
+;;; Copyright (C) 2017 ng0 <ng0@infotropique.org>
+;;;
+;;; This program is free software: you can redistribute it and/or modify it under
+;;; the terms of the GNU General Public License as published by the Free Software
+;;; Foundation, either version 2 of the License, or (at your option) any later
+;;; version.
+;;;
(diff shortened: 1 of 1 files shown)


ng-0:
For previous discussion read https://github.com/neomutt/neomutt/pull/527

I had to delete the old repository.

flatcap:
Hi @ng-0.  Do we need #527 any more?  If not, please can you close it.

I see that you've signed your commit, but GitHub hasn't recognised it (the commit above says "Unverified").  In your GitHub settings you can add all your public keys, then I should see "Verified".

**Note**: With NeoMutt's workflow, code tends to get rebased a lot before merging (which would break the signing).  This commit is already on top of the latest master, so there's no problem.

ng-0:
Yeah, I don't know what the problem is with Github. You can check the commit locally where it is alright.
A problem of SaasS I'd guess.

I will close #527 then. I'm still looking for the best solution for this, but this work okay'ish enough to be usable (I did the same for us in GNUnet).

PR #730, opened 2017-08-23
Unsubscribe after deleting an imap folder

seishinryohosha:
Only added two line to `browser.c` in `_mutt_select_file` case `OP_DELETE_MAILBOX`.

* **What does this PR do?**
Calls `imap_subscribe(char *, 0)` after successful call to `imap_delete_mailbox(struct Context, struct ImapMbox)` if user called `delete-mailbox` in browser.
Probably not.
Didn't found one. Should I create a issue first next time?


flatcap:
Thanks for the PR, @seishinryohosha.

> Should I create a issue first next time?

If you were planning a lot of work then it would be a good idea to create an issue first.

The code change looks sensible, but... please can you fix the typo in the comment :-)

seishinryohosha:
Oh, really missed that.
Right now I also noticed, that renaming a folder does not update it's subscription.
Also `delete-mailbox` does not delete recursively.

Should I open one Issue with "Subscription Bugs" or for each bug an individual one?

Unfortunately I'm in the exam phase right now, so I kinda lack the time to properly do it myself.
(I assume the fixes aren't that complicated)

I'm new to contributing on public projects, so please excuse my asking for every detail.

flatcap:
> renaming a folder does not update its subscription.
> delete-mailbox does not delete recursively.

Ah, yes.

> Should I open one Issue with "Subscription Bugs" or for each bug an individual one?

As they're closely related, open one bug.

> Unfortunately I'm in the exam phase right now, so I kinda lack the time to properly do it myself.

That's OK.  Real life comes first.

> I'm new to contributing on public projects, so please excuse my asking for every detail.

Not a problem.  Our newbie tutorial answers lots of common questions,
https://www.neomutt.org/dev/newbie-tutorial
and we're always happy to help people help us :-)

gahr:
I do have reservations towards this change but I'm currently really short on time. Please do not hurry it, I'll try to get back to it with detailed reasoning later today or tomorrow. Thanks!

gahr:
The change might look innocent and sensible at a first glance, but there
are a few subtleties I'd like to point out. If I understand correctly,
the reason for this change would be to avoid "dangling subscriptions".
What problems / unexpected behaviour are they causing?

As I see it, there is no such thing as a "dangling subscription" in IMAP
semantics. In IMAP, subscriptions are simply mailbox names, not anything
like pointers to actual instantiation of mailboxes having a particular
name. What this means is that you can subscript to Foo, remove Foo,
create a new Foo, and your subscription is still there, so when you LSUB
you'll see Foo.

This is especially important with rotating or temporary mailboxes.

Say I rotate my Spam messages weekly and keep them for one month. I
subscribe to Spam and I additionally have Spam1, Spam2, and Spam3 to
hold spam messages that I received 1, 2, or 3 weeks ago. Every Sunday, I
remove Spam3, rename Spam2 to Spam3, Spam1 to Spam2, Spam to Spam1, and
create a new Spam folder from scratch. This way, I always have a
relatively small set of emails to go through to check for false
positives, and have the possibility to check back one month's worth of
data if anything should have gone wrong.

Or say I have a single Spam folder, which I check daily. I move all
false-positives to INBOX, delete Spam, and create it anew. This might be
easier and/or faster than deleting all messages.

In both cases, with the change you propose, I'd need to regularly
re-Subscribe.

What are the issues you're experiencing with the current behaviour?

-- 
Pietro Cerutti
gahr@gahr.ch

gahr:
> Right now I also noticed, that renaming a folder does not update it's subscription.

The fact that I rename Spam to Spam2 in my first example does not mean that I want to transfer Subscription to Spam2, so I'd vote no for this one too :)

seishinryohosha:
Ok, your arguments are reasonable.

I just came into the situation, where I deleted a folder, but still had a subscription to it.
In other mailclients it appears for example grayed out (Thunderbird) others just ignore it (K9-Mail).
Of course neither of them is right or wrong.
So I had some options, to unsubscribe it anyway:

1) Recreate the folder in mutt and unsubscribe and delete it again.

2) Use another mailclient and use it's subscription handling feature.

3) Log in to my mailserver and do it there.

4) Use openssl s_client and do it manually.

5) Maybe some other methods I forget or have overseen.

As I used 1), it felt like dangling around.
I noticed, the folder still appears in the "subscription view" (If someone hits <TAB> in the folder-view)
But you can't unsubscribe, as it errors "Bad mailbox name" for `=test` as the name.
(Other IMAP-features also don't work in this view)

So I've come to some other suggestions:

* Do nothing, but may mention this in the docs
* Allow to use `:unsubscribe  <Foldername>`
* Use an option like `set delete_and_unsubscribe = yes` or `set auto_subscribe = yes`
* Have a subscription view

gahr:
> I noticed, the folder still appears in the "subscription view" (If someone hits in the folder-view)
But you can't unsubscribe, as it errors "Bad mailbox name" for =test as the name.

This is the case for all folders in folder-view. The IMAP code doesn't recognize your `=` prefix and think it's not an imap mailbox.

> Allow to use :unsubscribe <Foldername>

Yes, I'd like that

seishinryohosha:
> This is the case for all folders in folder-view. The IMAP code doesn't recognize your = prefix and think it's > not an imap mailbox.
Yes, I just mentioned it in order to show, that where is no direct way to unsubscribe

> Allow to use :unsubscribe
>> Yes, I'd like that

Great, I'm also perfectly well with this option.

I'd implement this myself, if I get some free time and nobody is quicker to do it instead ;)
As neomutt moves from globals to a ctx struct, I'll need to look a bit deeper in the code to make it right.

If somebody has different ideas, how to solve this, I gladly appreciate other opinions.

Thank you all for your quick replies!

seishinryohosha:
So, I got some free time.

My proposal would be to add `unsubscribe-from` in `init.h` with a typical `parse_unsubscribe_from` function. So the user can use:
```
:unsubscribe-from =Folder1 =Folder2
```
to unsubscribe from Folder1 and Folder2.

Of course, if you have a better name, I'd gladly accept it.
To keep it simple, I'd not add pattern matching.

Where would be the best place for `parse_unsubscribe_from`? `init.c` like others or somewhere else?

seishinryohosha:
**UPDATE**
Added `unsubscribe-from` command and implemented it with `parse-unsubscribe-from` in `init.c`
Uses `imap/imap.h` to unsubscribe.
Tokenize the inputbuffer, expand each token (path) and unsubscribe from it.

* **What does this PR do?**
Calls `imap_subscribe(char *, 0)` for each tokenized expanded path of given arguments.
Should I copy and free the expanded path?
none

gahr:
@seishinryohosha I have finally had a chance to test your branch. `unsubscribe-from` is working fine, but the first thing I noticed is the lack of a counterpart `subscribe-to`. Would you agree that it would be nice to have both?

seishinryohosha:
Thinking about adding subscribe-to or not I've chose to not include it for two reasons:

1) Minimize potential conflicts
2) unsubscribe-from arose from a missing direct way of doing it. (Without creating it first). As `subscribe-to` is already directly doable (change-folder -> ? -> Select entry -> subscribe) I've not included it.

Of course it's no big deal to add it, if it is wished for.

seishinryohosha:
Added subscribe-to command as requested.

gahr:
No error is given when trying to subscribe to a non-existent mailbox:

```
:subscribe-to =NonExistent
Subscribing to NonExistent... 
```

The output message remains like that.

From the command line:

```
a2 subscribe NonExistent                                                                                                                                          
a2 NO Mailbox doesn't exist: NonExistent (0.001 + 0.000 secs).                                                                                                    
a3 unsubscribe NonExistent
a3 OK Unsubscribe completed (0.001 + 0.000 secs). 
```

gahr:
Thanks, it works fine. Just one thing I'm not sure about: both `subscribe-to` and `unsubscribe-from` currently accept a space-separated list of mailboxes. This works, but the confirmation message is perhaps a bit confusing, as only the last mailbox is seen (the previous ones get printed but the message is overwritten by the subsequent ones).

```
:subscribe-to =Test =Test2 =Test3
Subscribed to Test3
```

I'm not sure how / whether you would like to improve this. Perhaps `subscribe-to` and `unsubscribe-from` should just accept one token?

seishinryohosha:
Added better output:

```
subscribe-to =exists =notexists
Successful: =exists Could not subscribe to =notexists

subscribe-to =notexists
Successful: None Could not subscribe to =notexists

subscribe-to =exists =existsAlso
Subscribed to =exists =existsAlso
```

gahr:
```
:subscribe-to<CR>
Subscribed to (null)
```

gahr:
@seishinryohosha there's been some discussion on FreeNode/#neomutt regarding the message strings, and we (me and @flatcap) came to the conclusion that perhaps it would be better to have these two new commands only accept a single mailbox name. It solves the original use-case of being able to unsubscribe from a non-existent mailbox, and it reduces code complexity considerably.
Would you be willing to implement this change?

seishinryohosha:
Of course.
I'll revert it a bit later today.

seishinryohosha:
Hope I didn't miss trivial cases this time ;)

gahr:
Great - subscribing to a non-existent mailbox still leaves the `Subscribing to NonExistent...` message in the status bar and gives no error, though

seishinryohosha:
Oh interesting, before changing to the err struct Buffer it worked, missed something where.
I'll fix it

seishinryohosha:
Just to clarify:
the `struct Buffer *err` is already initialized by the caller?
As `strfcpy` worked perfectly well on it.
I used `mutt_buffer_addstr` only in the nonexists folder case. Apparently I had to reset it first.

gahr:
Watch out, your call to `imap_subscribe` in `parse_unsubscribe_from` has a second parameter of `1`. Should be `0`.

seishinryohosha:
Never copy & paste...
I completely look like fool :(

I saw, that init.c uses stdbool.h, but imap.{h,c} doesn't.
Is it planed to change all int bool flags into `bool`?

gahr:
> Is it planed to change all int bool flags into bool?

Yes, that was the plan, most|some of the work has already been done. Not sure you want to do that within this PR, which in my opinion is ready for merging.. :)

flatcap:
Nice work @seishinryohosha!

I have a few simple tidying up requests before we merge.
Please could you:

* Expand the tabs in the source
* Replace the two pairs of `mutt_buffer_addstr()` with `mutt_buffer_printf()`
* Add gettext calls `_()` around the two "Could not..." messages
* Remove the gettext calls around the "Corrupted buffer" messages and change them to `mutt_debug()`
  These messages are very unlikely to be reached and if they are, it's a programming problem, not something the user has done wrong.

Thanks.

seishinryohosha:
Added suggested cleanup

* Expanded tabs
* changed to mutt_debug
* merged mutt_buffer_addstr -> mutt_buffer_printf
* vimdiff'ed with clang-format to get sure
* changed 0 -> false and 1 -> true in imap_subscribe()

Thank you for being so patient with me

gahr:
> changed 0 -> false and 1 -> true in imap_subscribe()

Sorry, I wouldn't do this unless you also i) change the signature, and ii) change all other callers

seishinryohosha:
Ok, I misinterpreted the previous post.

I could do this, but for the sake of clarity, I would do this in a separate PR and revert this for now.

gahr:
Cool, thanks!

PR #731, opened 2017-08-24
convert OPS files to a C header

d-k-c:
* **What does this PR do?**

This PR converts the OPS* files to a C header named opcodes.h. There are several benefits :

- Header is always in sync with regards to `config.h`. In some situations it could fall out of sync and trigger weird errors. Now, one can mess with config.h and OPS will no longer be a problem.
- Reduce complexity by removing a step in the build process.
- Remove generation of a useless header that was used only by gettext.

I expect some troubles with translations but I'm not really familiar with the smartness of gettext, so let me know if there is anything more I need to do with this PR regarding that.


opcodes.h
```diff
@@ -0,0 +1,286 @@
...
+  OPS_MIX(_fmt) \
+  OPS_NOTMUCH(_fmt) \
+  OPS_PGP(_fmt) \
+  OPS_SMIME(_fmt) \
+  OPS_CRYPT(_fmt) \
+
+enum mutt_ops {
+#define DEFINE_OPS(opcode, help_string) opcode,
+  OPS(DEFINE_OPS)
+  OP_MAX,
```
gahr: When doing thing kind of expnsion, I usually `#undef` the temporary macro just after use. I find it makes it clearer what it's his scope. Here, I'd do that just after the expansion of `OPS`, or if it looks nicer just before the `enum` closing brace.

help.c
```diff
@@ -32,15 +30,21 @@
...
+#include "opcodes.h"
 #include "options.h"
 #include "pager.h"
 #include "protos.h"
 
+static const char *HelpStrings[] = {
+#define DEFINE_HELP_MESSAGE(opcode, help_string) [opcode] = help_string,
+  OPS(DEFINE_HELP_MESSAGE)
+  [OP_MAX] = NULL,
+};
```
gahr: As in the other comment, I'd undef the temporary macro just after use.
Also, do you ever expect `opcodes` to come unordered or with gaps? If so, how? If not, why the explicit array index?

d-k-c: Okay for undefs. For the explicit array index, it's just an artifact of the way I thought about it I guess, no real reasons to have an explicit index here.

gahr:
Got this while generating the manual:

```sh
( mutt_ver=`sed -n -e '/MUTT_VERSION/s/.*"\(.*\)".*/\1/p' ../config.h` &&  sed -e "s/@VERSION\@/20170714 ($mutt_ver)/" /usr/home/gahr/github/neomutt/doc/manual.xml.head &&  /usr/local/libexec/ccache/cc -E -I. -I.. -I/usr/local/include -I/usr/home/gahr/github/neomutt -DSYSCONFDIR=\"/usr/local/etc\" -DBINDIR=\"/usr/local/bin\" -Ino/include -I/usr/local/include -I/usr/local/include -D_MAKEDOC -C /usr/home/gahr/github/neomutt/init.h | ./makedoc -s &&  /usr/local/libexec/ccache/cc -E -I. -I.. -I/usr/local/include -I/usr/home/gahr/github/neomutt -DSYSCONFDIR=\"/usr/local/etc\" -DBINDIR=\"/usr/local/bin\" -Ino/include -I/usr/local/include -I/usr/local/include -D_MAKEDOC -C /usr/home/gahr/github/neomutt/functions.h |  perl /usr/home/gahr/github/neomutt/doc/gen-map-doc /usr/home/gahr/github/neomutt/doc/manual.xml.tail /usr/home/gahr/github/neomutt/opcodes.h  ) > manual.xml
map mixmaster undefined at /usr/home/gahr/github/neomutt/doc/gen-map-doc line 62, <XML> line 22.
```

I don't know how much of an issue that is.

Issue #587, opened 2017-05-17
Better support for multipart/alternative?


mrzool:
I'm new to mutt and would like to use it for all my emails. I have spent the last couple of days trying to set up a sensible environment, but I'm hitting my head against (several) walls.

The struggle started when I realized that sending emails in plain text is not viable anymore nowadays, because hard line breaks render the text hardly readable on small screens, and I don't want my emails to be ugly and hardly readable. There's a [couple](http://wcm1.web.rice.edu/mutt-tips.html) [of](http://brianbuccola.com/line-breaks-in-mutt-and-vim/) [suggestions](http://www.mdlerch.com/emailing-mutt-and-vim-advanced-config.html) as to how to work around this with the unlucky standard `format=flowed` (`set text_flowed` in mutt) and some ad-hoc setting in Vim (`setlocal fo=aw`). I soon realized that this doesn't work in several major email clients. Text looking fine in Gmail web interface looked terrible in Outlook. `format=flowed` is thus not a viable solution to this issue.

It appears that the only way to get nicely wrapped, responsive text in emails is (horrors) marking it up with HTML. So I wrote a macro to convert the body of the message in the 'compose' window to HTML and change the `Content-Type` to `text/html` just before sending — I use Pandoc, which has the nice side-effect of allowing me to use Markdown in my emails. So far so good, things are working and I'm happy. But what about email clients that are unable to display HTML? I don't want to give up plain text wholesale, I want my emails to be as accessible as possible.

The standard solution to this is to use `Content-Type: multipart/alternative` and embed two versions of the same text in the body, one in plain text and one in HTML, delimiting the regions through a string named `boundary`:

```
Content-Type: multipart/alternative; boundary="myboundary"
```

So I wrote a (rudimentary) shell script that takes the email body, converts it as needed, sets the various boundaries between regions, and outputs a new body readily formatted for `multipart/alternative`. I integrated the script into my mutt macro and did some testing. 

Well, nothing works as expected.

I write an email containing the following:

```
First paragraph.

Second paragraph.

Third paragraph.
```

Then I close and drop into the 'Compose' window. I filter the body through my script with `F` and I get this back, as expected:

```
--myboundary
Content-Type: text/plain; charset=utf-8

First paragraph. 
 
Second paragraph. 
 
Third paragraph. 

--myboundary
Content-Type: text/html; charset=utf-8
Content-Transfer-Encoding: quoted-printable

<html code omitted>

--myboundary--

```

[...]

PR #734, opened 2017-08-26
Support for multipart/alternative and multipart/mixed

somini:
* **What does this PR do?**

- Allows reordering email attachments.

- Allow joining several attachments as a single `multipart/alternative` attachment.

This is still WIP, mutt segfaults right after sending the email.

This is the work of @oblitium, based on this unattributed patch:
https://github.com/altercation/arch-packages/blob/master/mutt-dgc-hg/dgc.groupalts

Everything?

     for that)

   - All builds are passing

References #587

### DIFF
--- a/handler.c
+++ b/handler.c
@@ -1189,7 +1189,7 @@ static int multilingual_handler(struct Body *a, struct State *s)
   struct Body *b = NULL;
   bool mustfree = false;
   int rc = 0;
-  struct Body *first_multilingual_part = NULL;
+  struct Body *first_part = NULL;
   struct Body *zxx_part = NULL;
   char *lang;
 
@@ -1226,8 +1226,8 @@ static int multilingual_handler(struct Body *a, struct State *s)
     {
       if (mutt_can_decode(b))
       {
-        if (b->language && !first_multilingual_part)
-          first_multilingual_part = b;
+        if (!first_part)
+          first_part = b;
 
         if (b->language && (mutt_str_strcmp("zxx", b->language) == 0))
           zxx_part = b;
@@ -1264,7 +1264,7 @@ static int multilingual_handler(struct Body *a, struct State *s)
     if (zxx_part)
       mutt_body_handler(zxx_part, s);
     else
-      mutt_body_handler(first_multilingual_part, s);
+      mutt_body_handler(first_part, s);
   }
 
   if (mustfree)
--- a/handler.c
+++ b/handler.c
@@ -1178,6 +1178,102 @@ static int alternative_handler(struct Body *a, struct State *s)
 }
 
 /**
+ * multilingual_handler - Parse a multi-lingual email
+ * @param a Body of the email
+ * @param s State for the file containing the email
+ * @retval 0 Always
+ */
+static int multilingual_handler(struct Body *a, struct State *s)
+{
+  struct Body *choice = NULL;
+  struct Body *b = NULL;
+  bool mustfree = false;
+  int rc = 0;
+  struct Body *first_multilingual_part = NULL;
+  struct Body *zxx_part = NULL;
+  char *lang;
+
+  mutt_debug(2, "RFC8255 >> entering in handler multilingual handler\n");
+  if ((a->encoding == ENCBASE64) || (a->encoding == ENCQUOTEDPRINTABLE) ||
+      (a->encoding == ENCUUENCODED))
+  {
+    struct stat st;
+    mustfree = true;
+    fstat(fileno(s->fpin), &st);
+    b = mutt_body_new();
+    b->length = (long) st.st_size;
+    b->parts = mutt_parse_multipart(
+        s->fpin, mutt_param_get(&a->parameter, "boundary"), (long) st.st_size,
+        (mutt_str_strcasecmp("digest", a->subtype) == 0));
+  }
+  else
+    b = a;
+
+  a = b;
+
+  if (a->parts)
+    b = a->parts;
+  else
+    b = a;
+
+  mutt_debug(2, "RFC8255 >> preferred_languages set in config to '%s'\n", PreferredLanguages);
+  char *preferred_languages = mutt_str_strdup(PreferredLanguages);
+  lang = strtok(preferred_languages, ",");
+
+  while (lang)
+  {
+    while (b)
+    {
+      if (mutt_can_decode(b))
+      {
+        if (b->language && !first_multilingual_part)
+          first_multilingual_part = b;
+
+        if (b->language && (mutt_str_strcmp("zxx", b->language) == 0))
+          zxx_part = b;
+
+        mutt_debug(2, "RFC8255 >> comparing configuration preferred_language='%s' to mail part content-language='%s'\n",
+                   lang, b->language);
+        if (lang && b->language && (mutt_str_strcmp(lang, b->language) == 0))
+        {
+          mutt_debug(2, "RFC8255 >> preferred_language='%s' matches content-language='%s' >> part selected to be displayed\n",
+                     lang, b->language);
+          choice = b;
+          break;
+        }
+      }
+
+      b = b->next;
+    }
+
+    if (choice)
+      break;
+
+    lang = strtok(NULL, ",");
+
+    if (a->parts)
+      b = a->parts;
+    else
+      b = a;
+  }
+
+  if (choice)
+    mutt_body_handler(choice, s);
+  else
+  {
+    if (zxx_part)
+      mutt_body_handler(zxx_part, s);
+    else
+      mutt_body_handler(first_multilingual_part, s);
+  }
+
+  if (mustfree)
+    mutt_body_free(&a);
+
+  return rc;
+}
+
+/**
  * message_handler - handles message/rfc822 body parts
  */
 static int message_handler(struct Body *a, struct State *s)
@@ -1902,6 +1998,11 @@ int mutt_body_handler(struct Body *b, struct State *s)
         (mutt_str_strcasecmp("alternative", b->subtype) == 0))
     {
       handler = alternative_handler;
+    }
+    else if ((mutt_str_strcmp("inline", ShowMultipartAlternative) != 0) &&
+             (mutt_str_strcasecmp("multilingual", b->subtype) == 0))
+    {
+      handler = multilingual_handler;
     }
     else if ((WithCrypto != 0) && (mutt_str_strcasecmp("signed", b->subtype) == 0))
     {
--- a/body.h
+++ b/body.h
@@ -36,6 +36,7 @@ struct Body
 {
   char *xtype;                  /**< content-type if x-unknown */
   char *subtype;                /**< content-type subtype */
+  char *language;               /**< content-language (RFC8255) */
   struct ParameterList parameter;  /**< parameters of the content-type */
   char *description;            /**< content-description */
   char *form_name;              /**< Content-Disposition form-data name param */
--- a/compose.c
+++ b/compose.c
@@ -367,6 +367,8 @@ static int check_attachments(struct AttachCtx *actx)
 
   for (int i = 0; i < actx->idxlen; i++)
   {
+    if (actx->idx[i]->content->type == TYPEMULTIPART)
+      continue;
     mutt_str_strfcpy(pretty, actx->idx[i]->content->filename, sizeof(pretty));
     if (stat(actx->idx[i]->content->filename, &st) != 0)
     {
@@ -637,6 +639,39 @@ static void compose_menu_redraw(struct Menu *menu)
     menu_redraw_motion(menu);
   else if (menu->redraw == REDRAW_CURRENT)
     menu_redraw_current(menu);
+}
+
+/**
+ * compose_attach_swap - Swap two adjacent entries in the attachment list
+ * @param msg   Body of email
+ * @param idx   Array of Attachments
+ * @param first Index of first attachment to swap
+ */
+static void compose_attach_swap(struct Body *msg, struct AttachPtr **idx, short first)
+{
+  /* Reorder Body pointers.
+   * Must traverse msg from top since Body has no previous ptr.
+   */
+  for (struct Body *part = msg; part; part = part->next)
+  {
+    if (part->next == idx[first]->content)
+    {
+      idx[first]->content->next = idx[first + 1]->content->next;
+      idx[first + 1]->content->next = idx[first]->content;
+      part->next = idx[first + 1]->content;
+      break;
+    }
+  }
+
+  /* Reorder index */
+  struct AttachPtr *saved = idx[first];
+  idx[first] = idx[first + 1];
+  idx[first + 1] = saved;
+
+  /* Swap ptr->num */
+  int i = idx[first]->num;
+  idx[first]->num = idx[first + 1]->num;
+  idx[first + 1]->num = i;
 }
 
 /**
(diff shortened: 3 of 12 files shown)


somini:
For reference, here is an email generated by this:
```
From somini@localhost Sat Aug 26 20:14:17 2017
Date: Sat, 26 Aug 2017 20:14:17 +0100
From: somini@localhost
To: me@localhost
Subject: sub
Message-ID: <20170826191417.a2qrhk3lrs6hiia2@localhost>
MIME-Version: 1.0
Content-Type: multipart/alternative; boundary="v4yduxschm2zjeua"
Content-Description: Alternatives for "TESTING/ex.markdown"
Content-Disposition: inline
Content-Transfer-Encoding: x-unknown
User-Agent: NeoMutt/20170714-132-f1c5c3-dirty (1.8.3)
Status: RO
Content-Length: 537
Lines: 18


--v4yduxschm2zjeua
Content-Disposition: attachment; filename="ex.markdown"

This in an example markdown file
--------------------------------

Use *italic* and **bold** faces, to convert to HTML.

--v4yduxschm2zjeua
Content-Type: text/html; charset=us-ascii
Content-Disposition: attachment; filename="ex.html"

<h2 id="this-in-an-example-markdown-file">This in an example markdown file</h2>
<p>Use <em>italic</em> and <strong>bold</strong> faces, to convert to HTML.</p>

--v4yduxschm2zjeua--
```

and an real-world email, truncated to the important parts:
```
From noreply@steampowered.com Sat Aug 26 19:24:31 2017
From: "Steam" <noreply@steampowered.com>
Subject: An item on your Steam wishlist is on sale!
Content-Type: multipart/alternative;
 boundary="np59a1bcaea590b"

This is a multi-part message in MIME format.
--np59a1bcaea590b
Content-Type: text/plain; charset=UTF-8; format=flowed

Hello somini


--np59a1bcaea590b
Content-Type: text/html; charset=UTF-8

<!DOCTYPE html PUBLIC "-//W3C//DTD HTML 4.01 Transitional//EN">
<html>
<!-- HTML HERE -->

--np59a1bcaea590b--
```

oblitum:
Other issues that happened while testing it out:

- Using the original text mail attachment and just adding a single additional html one, tagging both and making them multipart/alternative doesn't create a multipart message with the two, just with the HTML one, leaving the original text mail as is. It seems it's necessary to have at least one attachment that won't be part of the multipart step for it to work, this extra attachment should then be deleted before sending to avoid creating a multipart/mixed message.

flatcap:
[This message is from @walkerlala -- transferred from #1094]

* **What does this PR do?**

This work is based on commit 5b4f7a1342 from @somini and was discussed at <https://github.com/neomutt/neomutt/issues/587>. As of now, one can tag entries in the compose menu and use <group-alternatives> to group those entries into one. If all the entries are tagged (including the
fundamental part), then the email become a multipart/alternative one, otherwise it becomes a multipart/mixed one.

There is some weird things to be noted, though: when postponing a multipart/alternative mail and then recall it, those entries get split up. This does not occurred for multipart/mixed mails.

Also, please note:
https://github.com/neomutt/neomutt/blob/93278967e0c205f62b865f71d39f486d2aa2b143/compose.c#L1160

walkerlala:
@oblitum This situation you described has already been fixed by commit 93278967e . Please see the rebase by @flatcap in branch devel/multipartalternative.

oblitum:
hi @walkerlala, I'm not particularly into this anymore since I've been relying on [my own solution for HTML emails](http://nosubstance.me/post/mutt-secret-sauce/). I can try it out, but solely when it reaches release state and on my distro's package manager.

oblitum:
If you think it's fine and what I commented at the time has been fixed, then great :)

walkerlala:
@oblitum I read your post and found lots of interesting things! But still I encourage you to use neomutt's multipart/alternative features so that we can have more feedback.

walkerlala:
@flatcap Here is a multipart/alternative email generated with neomutt:

```
From ablacktshirt@gmail.com Wed Mar 14 11:38:17 2018
Date: Wed, 14 Mar 2018 11:38:17 +0800
From: Yubin Ruan <ablacktshirt@gmail.com>
To: Yubin Ruan <ablacktshirt@gmail.com>
Subject: Test of multipart/alternative email
Message-ID: <20180314033817.fzqgpkca4v7gb5fp@HP>
MIME-Version: 1.0
Content-Type: multipart/alternative; boundary="7bq6zb2a5fxeq4hu"
Content-Description: Alternatives for "/tmp/neomutt-HP-1000-9648-11188063731682912527"
Content-Disposition: inline
Content-Transfer-Encoding: x-unknown
X-Organization: http://fastdrivers.org
User-Agent: NeoMutt/20180223-102-079cbc-dirty


--7bq6zb2a5fxeq4hu
Content-Disposition: inline

   NOTE! This copyright does *not* cover user programs that use kernel
 services by normal system calls - this is merely considered normal use
 of the kernel, and does *not* fall under the heading of "derived work".
 Also note that the GPL below is copyrighted by the Free Software
 Foundation, but the instance of code that it refers to (the Linux
 kernel) is copyrighted by me and others who actually wrote it.

 Also note that the only valid version of the GPL as far as the kernel
 is concerned is _this_ particular version of the license (ie v2, not
 v2.2 or v3.x or whatever), unless explicitly otherwise stated.

--7bq6zb2a5fxeq4hu
Content-Type: text/html; charset=us-ascii
Content-Disposition: attachment; filename="neomutt-multipart.html"

<!DOCTYPE HTML PUBLIC "-//W3C//DTD HTML 4.01//EN" "http://www.w3.org/TR/html4/strict.dtd">
<html>
<head>
<meta http-equiv="content-type" content="text/html; charset=UTF-8">
<title>/tmp/neomutt-HP-1000-9648-17190040443524511568.html</title>
<meta name="Generator" content="Vim/8.0">
<meta name="plugin-version" content="vim7.4_v2">
<meta name="syntax" content="mail">
<meta name="settings" content="use_css,pre_wrap,no_foldcolumn,expand_tabs,prevent_copy=">
<meta name="colorscheme" content="default">
<style type="text/css">
<!--
pre { white-space: pre-wrap; font-family: monospace; color: #000000; background-color: #ffffff; }
body { font-family: monospace; color: #000000; background-color: #ffffff; }
* { font-size: 1em; }
-->
</style>

<script type='text/javascript'>
<!--

-->
</script>
</head>
<body>
<pre id='vimCodeElement'>
   NOTE! This copyright does *not* cover user programs that use kernel
 services by normal system calls - this is merely considered normal use
 of the kernel, and does *not* fall under the heading of &quot;derived work&quot;.
 Also note that the GPL below is copyrighted by the Free Software
 Foundation, but the instance of code that it refers to (the Linux
 kernel) is copyrighted by me and others who actually wrote it.

 Also note that the only valid version of the GPL as far as the kernel
 is concerned is _this_ particular version of the license (ie v2, not
 v2.2 or v3.x or whatever), unless explicitly otherwise stated.
</pre>
</body>
</html>
<!-- vim: set foldmethod=manual : -->

--7bq6zb2a5fxeq4hu--
```

Here is what I do:

  - write the email body (in plain text) as usual.
  - use vim's TOHtml feature to generate a piece of html from that plain-text email body and save it as "neomutt-multipart.html"
  - quit the editor and return to the compose menu
  - attach the "neomutt-multipart.html" that was generated before.
  - tag both attachments and press & to group them into a multipart/alternative one
  - send it to myself

As you can see, the `Content-Type` of this email and the boundary is fine. However, there are some defects:

  - the `Content-Transfer-Encoding` header of this email is `X-unknown`. It will be better to provide a default known one.
  - the html part of this email still have a `Content-Disposition` of *attachment* rather than *inline* or anything else. Should it be inline or what? Most of the multipart/alternative email I receive does not have the `Content-Disposition` header so I cannot check. Does anyone know what is the default?
  - this email cannot be view normally in the gmail web interface. Both the two parts are now treated as attachments. I guess there are something wrong with the header (maybe it is because of the `Content-Disposition` header above)
  - when attaching a html attachment, if the filename of that attachment does not end with `.html`, then neomutt will treat it as a `Content-Type: text/plain` attachment rather than a `text/html` attachment.

oblitum:
@walkerlala On my blog post I point to the script I use to create multipart emails, and at the time I've read the references (wikipedia, RFCs, etc) to do it right. You may use it as reference because they display fine as multipart on Gmail.

oblitum:
If you run:

```bash
curl --silent https://raw.githubusercontent.com/mikel/mail/master/spec/fixtures/emails/plain_emails/raw_email_simple.eml \
    | MIMEmbellish
```
You will at least see how a simple multipart text/html should be.

oblitum:
Here's a sample output for that command:

```
Content-Type: multipart/alternative; boundary="===============1769514270870265629=="
MIME-Version: 1.0
Return-Path: <mikel@nowhere.com>
Received: from mikel091a by oaamta05sl.mx.bigpond.com with SMTP id <20071021093820.JFMT24025.oaamta05sl.mx.bigpond.com@mikel091a> for <mikel@nowhere.com.else>; Sun, 21 Oct 2007 19:38:20 +1000
Date: Sun, 21 Oct 2007 19:38:13 +1000
From: Mikel Lindsaar <mikel@nowhere.com>
To: Mikel <mikel@somewhere.com>
Message-Id: <009601c813c6$19df3510$0437d30a@mikel091a>
Subject: Testing outlook

--===============1769514270870265629==

Hello Mikel


--===============1769514270870265629==
Content-Type: text/html; charset="utf-8"
MIME-Version: 1.0

<!DOCTYPE html>
<html>
<head>
  <meta charset="utf-8">
  <meta name="ProgId" content="Word.Document">
  <meta name="viewport" content="width=device-width, initial-scale=1">
  <meta http-equiv="X-UA-Compatible" content="IE=edge" />
  <meta http-equiv="Content-Type" content="text/html; charset=utf-8" />
  <title>Testing outlook</title>
</head>
<body>
<p>Hello Mikel</p>
</body>
</html>

--===============1769514270870265629==--
```

walkerlala:
On Fri, Mar 16, 2018 at 11:07:37PM +0000, Francisco Lopes wrote:
> If you run:
> 
> ```bash
> curl --silent https://raw.githubusercontent.com/mikel/mail/master/spec/fixtures/emails/plain_emails/raw_email_simple.eml \
>     | MIMEmbellish
> ```
> You will at least see how a simple multipart text/html should be.

Thanks! But we are already able to generate multipart emails which are suitable for Gmail. Have you tried?

somini:
I tested this and it seems to be working as intended.

Based on https://github.com/neomutt/neomutt/pull/734#issuecomment-372897236 to try and send an email that could be seen in Gmail and here's my flow:

- Generate a plain text and HTML files. I used:
  - Write plaintext email in Markdown
  - Use pandoc to generate the corresponding HTML into a separate file
- Add both to the email
- Change both `Content-disposition` to `inline`
- Delete the description for both parts
- Tag both and create a `multipart/alternative` part
- Remove the description on this

What's currently not possible to do from mutt.
- [ ] Add a way to edit the full "Content-Disposition" header for parts, to remove the extra "filename" part that make the HTML downloadable. This is probably Gmail breaking the standards, because the part is inline. There's a `Content-Description` header for explaining what the part is.
- [ ] Set `Content-Transfer-Encoding` to `binary? From the [RFC](https://tools.ietf.org/html/rfc1521#section-5) it seems it might be unnecessary.

Here's an example I generated and manually hacked with very long lines.
https://gist.github.com/somini/9ae7a3cec8297fd308a7d252565cb737

oblitum:
@walkerlala ok, I was just  trying to help on "- this email cannot be view normally in the gmail web interface". Also, sorry but I didn't try it.

walkerlala:
@somini 

> Change both Content-disposition to inline

This is done automatically when you use <group-alternatives>

> Add a way to edit the full "Content-Disposition" header for parts, to remove the extra "filename" part that make the HTML downloadable. This is probably Gmail breaking the standards, because the part is inline. There's a Content-Description header for explaining what the part is.

Yes we should remove the extra filename for parts. I will take this

>  Set Content-Transfer-Encoding to `binary? From the RFC it seems it might be unnecessary.

Then shall we just leave this as "X-unknown"

BTW, if all these are done, can this branch be merged into master? @flatcap

somini:
>  Yes we should remove the extra filename for parts. I will take this

When turning parts inline, this should be done automatically, there is no need to restrict this to multipart emails.

I don't think it makes sense to have inline parts with filenames, but the [RFC](https://tools.ietf.org/html/rfc2183#section-2.8) is a free-for-all.

flatcap:
Now that the release is out of the way, I can have a good look at this.

@walkerlala, @somini, @oblitum 
Meanwhile... please can you turn your wisdom to #1104 (multi-lingual).
Give it a test, review the code, help to get it NeoMutt-ready.

Thanks

flatcap:
It works!  Yay!

I noticed one oddity.  I assembled a packet, but my email wasn't addressed.
After the error message ("No recpients"), the packet was two separate attachements again.

I've tidied up the code a little (the commit message explains what I've done).
I still need to run through it in the debugger to verify _how_ it's doing it.

walkerlala:
@flatcap 

> I noticed one oddity. I assembled a packet, but my email wasn't addressed. After the error message ("No recipients"), the packet was two separate attachments again.

This will happen when the message is **multipart/alternative**, but not **multipart/mixed**. It is the same situation when one recalls a postponed message, as I described above. Probably a bug. I am looking into it.

BTW, for the debugging purpose, are you attaching a debugger (gdb) to it in another console? Or are you using some other tools?

walkerlala:
@flatcap If you are looking into that oddity (multipart/alternative message broken apart), note the comment at `postpone.c:616` (try to postpne some multipart/alternative message and then recall it). I have scanned through the problem, and it seems that it will not be a easy fix... because relevant code are scattered around.

walkerlala:
And, BTW, after compiling from source in the `devel/multipartalternative` branch, it seems that there is something broken when trying to recall postponed message (using `R` does not work and there is no `<recall-message>` function there). Will that be a configuration problem?

flatcap:
> for the debugging purpose, are you attaching a debugger (gdb) to it in another console?

Yes.  I recently updated our [debugging page](https://www.neomutt.org/dev/debug).
I use two windows: cgdb (curses front-end to gdb) in one and neomutt in the other.

> If you are looking into that oddity

I wasn't...

> ...it seems that it will not be a easy fix...
> because relevant code are scattered around.

...but if you're stuck, I'll happily investigate.
Otherwise, I have lots to do :-)

> there is no <recall-message> function there.
> Will that be a configuration problem?

`<recall-message>` seems to work (well, it restores an email)
If `$postponed` isn't set to a folder, then `<recall-message>` won't do anything (not even an error).

walkerlala:
@flatcap The reason why  `multipart/alternative` emails get broken aparts at error and when recalled from postponed is that the original author explicitly do that. It is achieved by the function `mutt_remove_multipart` and the reason why it is done this way is explained in *postpone.c:616-624* (at branch multipart/alternative), that is, to prevent a multipart message to be the primary part.

You can use grep to see that this function exists in multiple places.

And for this reason above, I will take it as a feature rather than a bug.

And, if this is taken as a feature, can this branch be merged? Or do we still have to wait for https://github.com/neomutt/neomutt/pull/1104 ?

By the way, there is really something going wrong when recalling postponed messages. Specifically, if there are multiple postponed messages in the postponed mailbox and I try to use <recall-message> to recall just **one** message, then the whole  index go pale (all new messages turned into read). I haven't found out the reason yet. Just FYI.

walkerlala:
We can now write multipart/multilingual mails in neomutt! I have transferred the commit from https://github.com/neomutt/neomutt/pull/1104 so that we can view it too.

@flatcap There is still something left:

  1. we need to add more doc to that.

  2. there is some code duplicate in compose.c . I did that because there might be some need in the future.

  3. We need more tests.

flatcap:
Wow!  You **have** been busy!

I shall have a good look at it tomorrow.

As for release...

- Fix any bugs we find
- Document:
  - Multi-part vs multi-lingual
  - How to create each one
  - How to view your preferred **part**
  - How to view your preferred **language**
  - What works (and what does "unexpected" things)
  - That this feature is a **work-in-progress** and may change
- Get the devel mailing list to test it
  - Send out some multi-part/multi-lingual emails
  - Get the developers to reply with the same

Before we release this feature, we need to:
- List all the bugs / "features"
- Decide what the "correct" behaviour should be for each

gahr:
Should we perhaps allow users to define their own `BodyLanguages` ? A nice default in neomuttrc and the ability to change it site-wide / per-user would be nice.

walkerlala:
@gahr It's probably not a good idea to allow users to defined their own `BodyLanguage`. I think it is better to comply with some standards, otherwise the MUA receiving the mails might not be able to decode them.

handler.c
```diff
@@ -1176,6 +1176,102 @@ static int alternative_handler(struct Body *a, struct State *s)
...
+  while (lang)
+  {
+    while (b)
+    {
+      if (mutt_can_decode(b))
+      {
+        if (!first_part)
+          first_part = b;
+
+        if (b->language && (mutt_str_strcmp("zxx", b->language) == 0))
```
gahr: I think this was introduced for testing and should be removed.

gahr:
@walkerlala fair enough. My point is more than given the complexity of the language tag (see RFC4646) and that we're just offering exact string comparison as a matching algorithm (see RFC4647), a user might want to add to that list to better guide the choice of language to display.

But anyway I'm fine with this first simple approach.

handler.c
walkerlala: No, there is indeed a "zxx" part, which is a "fallback" if the MUA cannot interpret that email at all. Please see RFC8255.

walkerlala:
@gahr In my first implementation I did try to comply with ISO639-1, but without zones. But then I realized that country codes in ISO639-1 come with no zones, and I cannot find a standard with zones (just not sure which is right and which is not). So I gave up, temporarily...

Can you please provide a exact list?

gahr:
> @flatcap Do we need a language list at all?

Of course, you're right, I couldn't agree more! Let's not bother trying to validate the contents :)

flatcap:
Rebased and tidied.

flatcap:
@walkerlala 
Do you agree with my suggestion that we don't need the language list?
If so, please can you remove it.
Thanks.

flatcap:
> Sorry for the delay.

No problem

> language list, is used to checked whether users enter a valid languages when composing emails

Ah, OK.
Hmm... do we care which language code they use? (as long as it's not 'zxx')

walkerlala:
@flatcap I think it is better to be more constraint to prevent some unexpected mistake such as typo.

flatcap:
OK.  Fair enough.

walkerlala:
@flatcap Are there any test going on for this issue?

walkerlala:
@gahr Fair enough. @flatcap what do you think?

flatcap:
OK.  Please drop the language list.

flatcap:
> Are there any test going on for this issue?

Not really.
I tried it a while back.
Generally, it worked for me, but it's a lot of work to construct an email.

I'd like to merge it, but I'd prefer to see some more testing first.
You tried to encourage the devel mailing list, but I think they'd only respond if you set them on fire :-)

Fortunately, most users won't see the changes, so they won't be affected if there's a problem.

---

Before then, please can you

### Document

1. What this feature is capable of: multipart, multilingual
2. Create some use cases, listing the steps, e.g.
   - User sending an HTML email (with a text alternative)
   - User sending a multi-lingual email
3. What can be scripted?
   - Within NeoMutt (macro)
   - Outside of NeoMutt (shell script)
4. List of all the known limitations of the feature
   (from a user's point of view)

4 is important.
I don't want bug reports for things that don't work, aren't supported and aren't going to be fixed.
Once the feature is released, I don't want to have to take it away.

### Plan

I would like a plan:
(notes would do, it's not for the public)

- What are all problems (from a code point of view)?
  I don't have the time to investigate these, myself.
- If this feature were finished and perfect, what would it look like?
  Just your blue-sky ideas

If I know what this feature **can** and **cannot** do, and
I have an idea of what its future _could be_, then
I'll be happy to merge it.

I'm not committing **you** to make this future happen :-)

Thanks.

walkerlala:
language list driopped.

I would take some time to document these. But first I have to figure out how these document work...

flatcap:
> I have to figure out how these document work

Any questions, just ask.
I'm also in our IRC channel a lot (#neomutt irc.freenode.net) -- mostly during European daylight hours.

walkerlala:
Oh...I didn't expect removing a function will cause a build error. It seems like that there are multiple definition of `multilingual_handler` in `handler.c`. There must be double commit. I can't figure out what is the right way to do that. @flatcap Can you take some time to investigate on this?

flatcap:
Wow that's a good one.
Your latest commit introduces 1517 empty lines into compose :-)
Give me a second...

walkerlala:
Oh my... I guess there might be editor mistake when I was editing that file. Sorry for that.

walkerlala:
The commit history now look more pleasant. What is the issue previously? @flatcap

flatcap:
I think you merged the old branch (dating back several months).

walkerlala:
@flatcap @gahr @somini @igit 
I has finished the doc for `multipart/{alternative,multilingual}`, please take some time to review.

As of the problems and plan,

These features are almost complete, and you can happily read/write `multipart/alternative` and `multipart/multilingual` emails (please see the doc). With some macros and scripts, you can even transform a plain text email into a `multipart/alternative` bundle in one single key stroke.

There are still some problems, though:

  1. some code has to be re-factored / re-used. e.g., code for multipart/alternative and multipart/multilingual can be factored out to reduce code duplicate. I intentionally left that out previously for easy modification.

  2. Error and warning messages for user can be better. There is some "can't stat (null)" message on error.

igit:
@walkerlala @flatcap
Sorry for not been available earlier to review the code and the doc. This is now done on my side : all seems OK for me.

I've just compiled and tested source code from master branch. The multipart/multilingual edition works well. So now, we can consider that NeoMutt is fully compatible with RFC8255 on both sides : writing and reading. #cool !
On the writing side, the writer must know exactly what he does and he has to create, by hand, all the parts (preface, parts with languages, part with zxx). Maybe in a future version the interface can help (I don't know exactly how) the writer building this ?

Anyway, thank you all for working of this and making the RFC8255 IETF101 Hackathon Proof Of Concept a reality !

PR #740, opened 2017-08-31
mutt_notmuch: fix setting custom query_type in notmuch query

### LINKED ISSUE #484: Unable to override nm_query_type in virtual mailbox
I set `nm_query_type` to `threads`, and defined some virtual mailboxes like this:

`notmuch://?query=tag:flokli and (tag:inbox or tag:new)`

The results are displayed in threads. When changing `nm_query_type` to `messages`, results are displayed in messages, so this works as expected.

The documentation mentions its also possible to override the query type per virtual folder, so something like ``notmuch://?type=messages&query=tag:flokli and (tag:inbox or tag:new)`` should always display results as messages, regardless of the `nm_query_type`.

Currently, the custom type however seems to be ignored, and the value for `nm_query_type` is always used, instead of the override, which is a bug.

NeoMutt 20170306 (1.8.0) and notmuch 0.23.5 on Arch Linux amd64 here.

flokli:
nm_normalize_uri re-builds notmuch queries, but `nm_uri_from_query` doesn't set `query_type` back again, so fix that.
Also, use the `NmQueryType` `enum` where possible instead of `int`.

Fixes #484.

### DIFF
--- a/mutt_notmuch.c
+++ b/mutt_notmuch.c
@@ -142,10 +142,10 @@ struct NmCtxData
 {
   notmuch_database_t *db;
 
-  char *db_filename; /**< Filename of the NotMuch database */
-  char *db_query;    /**< Previous query */
-  int db_limit;      /**< Maximum number of results to return */
-  int query_type;    /**< Messages or Threads */
+  char *db_filename;           /**< Filename of the NotMuch database */
+  char *db_query;              /**< Previous query */
+  int db_limit;                /**< Maximum number of results to return */
+  enum NmQueryType query_type; /**< Messages or Threads */
 
   struct UriTag *query_items;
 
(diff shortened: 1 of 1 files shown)


andir:
The change looks good to me. I've not test-run it but the code looks a lot simpler to me then the original version. Also that bug is kinda annoying (but wasn't a game breaker for me yet). 

Looking forward to seeing this merged :-)

mutt_notmuch.c
```diff
@@ -450,22 +450,25 @@ static struct NmCtxData *get_ctxdata(struct Context *ctx)
...
+static enum NmQueryType string_to_query_type(const char *str)
 {
-  if (!str)
-    str = NmQueryType; /* user's default */
-  if (!str)
-    return NM_QUERY_TYPE_MESGS; /* hardcoded default */
-
-  if (strcmp(str, "threads") == 0)
+  if (mutt_strcmp(str, "threads") == 0)
     return NM_QUERY_TYPE_THREADS;
```
flatcap: @flokli 
I've changed the two `strcmp()`s here to `mutt_strcmp()` which checks for `str` being `NULL`.
(Otherwise mutt would crash if the user unset '$nm_query_type')

PR #745, opened 2017-09-07
Autosetup-based configuration

gahr:
* **What does this PR do?**

Introduce the new [Autosetup](https://msteveb.github.io/autosetup/)-based build system. Please refer to [AUTOSETUP.md](https://github.com/neomutt/neomutt/blob/devel/autosetup/AUTOSETUP.md) for details.

**TL;DR**

* An alternative zero-dependency configuration mechanism is available as `./configure.autosetup`
* It supersedes `autoreconf --install && ./configure`
* It's intended to be feature-complete wrt the current autotools
* Options might be a bit different, e.g. `--enable-notmuch=/opt` → `--notmuch --with-notmuch=/opt`
* Run with `--help` for a list of supported options

**Thanks to**

* @flatcap for the support and pre-PR testing on Linux
* @msteveb for reviewing the whole work and providing useful feedback

Your feedback is especially wanted: @neomutt/downstream-packagers @neomutt/notmuch-users @neomutt/reviewers


lbschenkel:
I have tested it on MacOS 10.12, I'm getting the following output:
```
./configure.autosetup
Host System...x86_64-apple-darwin16.7.0
Build System...x86_64-apple-darwin16.7.0
C compiler...ccache cc -g -O2
C++ compiler...ccache c++ -g -O2
Build C compiler...cc
Error: invalid command name "lmap"
```

gahr:
Thanks @lbschenkel, this reminds me that I still have to test it on Mac.

`[lmap]` appeared in Tcl 8.6 and I fear OSX 10.12 ships Tcl 8.5 by default. You can double check that by looking at the output of `echo "puts [info patchlevel]" | tclsh`.

Since 8.5 is so widespread, it might make sense to patch Autosetup to do without `[lmap]`.

gahr:
Doh! [959cc](https://github.com/msteveb/autosetup/commit/959cc1bd0e3fe91c816fdd0bc6e67cefad95ed9c). Thanks @msteveb!

msteveb:
Yes, I noticed the lmap problem too but I forgot to push the fix to autosetup. You can find the fix in the latest commit to msteveb/autosetup (sorry, I don't know how to add a link here)

lbschenkel:
Next try:
```
Error: invalid command name "try"
```

lbschenkel:
I also noticed that the new autosetup build fails to detect LibreSSL:
```
Checking for openssl/bio.h...ok
Checking for openssl/err.h...ok
Checking for openssl/ssl.h...ok
Checking libs for X509_STORE_CTX_new...-lcrypto
Checking libs for SSL_new...-lssl
Checking for RAND_status...ok
Checking for RAND_egd...not found
Error: Unable to find OpenSSL
```
This was not the case previously. The old autotools just didn't set `HAVE_RAND_EGD`.

gahr:
Thank you! I have fixed both issues.

lbschenkel:
I can configure it now, tried `--prefix=/opt/local` (since I'm testing for MacPorts):
```
Summary of build options:

  Version:           20170907
  Host OS:           darwin16.7.0
  Install prefix:    /opt/local
  Compiler:          ccache cc
  CFlags:            -g -O2 -std=c99 -D_ALL_SOURCE=1 -D_GNU_SOURCE=1 -D__EXTENSIONS__ -I/opt/local/include
  LDFlags:           -L/opt/local/lib
  Libs:              -lz -lssl -lcrypto -lncursesw -lintl -liconv
  Header cache libs:

  GPGME:             no
  PGP:               yes
  SMIME:             yes
  Notmuch:           no
  Header Cache(s):
  Lua:               no
```

but later `make` fails:
```
ar cr libimap.a imap/auth.o imap/auth_anon.o imap/auth_cram.o imap/auth_login.o imap/auth_plain.o imap/browse.o imap/command.o imap/imap.o imap/message.o imap/utf7.o imap/util.o
ranlib libimap.a
ar cr libhcache.a
ar: no archive members specified
usage:  ar -d [-TLsv] archive file ...
        ar -m [-TLsv] archive file ...
        ar -m [-abiTLsv] position archive file ...
        ar -p [-TLsv] archive [file ...]
        ar -q [-cTLsv] archive file ...
        ar -r [-cuTLsv] archive file ...
        ar -r [-abciuTLsv] position archive file ...
        ar -t [-TLsv] archive [file ...]
        ar -x [-ouTLsv] archive [file ...]
make: *** [libhcache.a] Error 1
```

gahr:
Thanks for the report. I'll look at it later when I have my macbook at hand.

lbschenkel:
From what I can see, the new build is not prepared to handle the case when header cache is not enabled: it tries to create a library without any object files. The whole command should have been skipped instead.

gahr:
True - let's see if the latest commit fixed that.

lbschenkel:
I managed to get a successful build now. I didn't test all combinations, but it looks like the common cases are working at least.

gahr:
Thanks a lot @lbschenkel !

flatcap:
This isn't a comprehensive list of options.

| OLD autoconf              | NEW autosetup           | Notes                                        |
| :------------------------ | :---------------------- | :------------------------------------------- |
| `--disable-fcntl`         | `--disable-fcntl`       | fcntl(2) file locking                        |
| `--enable-flock`          | `--flock`               | flock(2) file locking                        |
| `--enable-locales-fix`    | `--disable-locales-fix` | Workaround for broken locales                |
| `--disable-nls`           | `--disable-nls`         | Localisation (translations)                  |
| `--disable-pgp`           | `--disable-pgp`         | Pretty Good Privacy                          |
| `--disable-smime`         | `--disable-smime`       | Secure/Multipurpose Internet Mail Extensions |
|                           |                         |                                              |
| `--without-idn`           | `--idn`                 | Internationalised domain names               |
| `--enable-debug`          | `--logging`             | Logging                                      |
| `--with-slang`            | `--slang`               | Slang screen drawing                         |
| `--with-curses`           | **MISSING**             | Curses screen drawing                        |
| `--with-domain=DOMAIN`    | `--with-domain=DOMAIN`  | Default email domain                         |
| `--with-mailpath=DIR`     | `--with-mailpath=DIR`   | Location of spooled mail                     |
| `--with-homespool`        | `--homespool`           | Spooled mail is in user's home dir           |
|                           |                         |                                              |
| `--enable-gpgme`          | `--gpgme`               | GPG Made Easy                                |
| `--with-gnutls`           | `--gnutls`              | Gnu TLS (SSL)                                |
| `--with-gss`              | `--gss`                 | Generic Security Services                    |
| `--with-sasl`             | `--sasl`                | Simple Authentication and Security Layer     |
| `--with-ssl`              | `--disable-ssl`         | OpenSSL                                      |
|                           |                         |                                              |
| `--enable-fmemopen`       | `--fmemopen`            | Optional Feature (Dangerous)                 |
| `--enable-lua`            | `--lua`                 | Optional Feature                             |
| `--enable-notmuch`        | `--notmuch`             | Optional Feature                             |
| `--with-mixmaster`        | `--mixmaster`           | Optional Feature                             |
|                           |                         |                                              |
| `--with-bdb`              | `--bdb`                 | Header cache backend                         |
| `--with-gdbm`             | `--gdbm`                | Header cache backend                         |
| `--with-kyotocabinet`     | `--kyotocabinet`        | Header cache backend                         |
| `--with-lmdb`             | `--lmdb`                | Header cache backend                         |
| `--with-qdbm`             | `--qdbm`                | Header cache backend                         |
| `--with-tokyocabinet`     | `--tokyocabinet`        | Header cache backend                         |
|                           |                         |                                              |
| `--disable-doc`           | `--disable-doc`         | Don't build the docs                         |
| `--disable-full-doc`      | `--full-doc`            | Document disabled features                   |
| `--enable-everything`     | `--everything`          | Build all features                           |
| `--quiet`                 | `--quiet`               | Only show the summary                        |
|                           |                         |                                              |
| `--disable-iconv`         | GONE                    | Compulsory                                   |
| `--enable-dotlock`        | GONE                    | Obsolete                                     |
| `--enable-exact-address`  | GONE                    | Obsolete                                     |
| `--enable-mailtool`       | GONE                    | Compulsory                                   |
| `--enable-nfs-fix`        | GONE                    | Obsolete                                     |
| `--with-included-gettext` | GONE                    | Obsolete                                     |
| `--without-wc-funcs`      | GONE                    | Obsolete                                     |
| `--disable-po`            | GONE                    | Obsolete, use `--disable-nls`                |

flatcap:
I notice there are a few changes between the versions.
By default:
- OpenSSL is turned on
- IDN is turned off
- Locales Fix is turned on

Also, there's no explicit `--with-curses` option.

gahr:
Thanks, I'll review and fix

Please note that autosetup recognizes `--feature` as a shorthand for `--enable-feature`.

The extended version would reduce the differences in the table you have produced ;)

gahr:
The behavior of IDN relies on the fact that `--with-idn`, `--without-idn`, and no `*-idn` argument at all are all different configuration options. This does not exist in autosetup, where a boolean is either, well, `true` or `false`.

I can switch the default to `true`, but that would *require* idn in the default configuration. I'm fine with that, if it's not an issue.

flatcap:
> autosetup recognizes --feature as a shorthand for --enable-feature.

Oh nice.  I quite like the shorter options.

> `--with-idn`, `--without-idn`, and no `*-idn` argument at all are all different configuration options

Ah, I didn't know that.

0-wiz-0:
I just tried the current head of the branch on NetBSD.
First problem: `autosetup` doesn't use `pkg-config` to find libraries, in particular:
````
Checking for stringprep.h...not found
Checking for idn/stringprep.h...not found
Checking for idna.h...not found
Checking for idn/idna.h...not found
Checking libs for stringprep_check_version...no
Error: Unable to find GNU libidn
Try: 'autosetup --help' for options
````

**Comment by gahr** [comment]
Thanks for your tests, @0-wiz-0 . I actually considered using `pkg-config`, but that'd mean introducing a build dependency. I'd like to avoid that.

You can tell autosetup where to find libidn with `--with-idn=/usr/pkg`

**Comment by 0-wiz-0** [comment]
Well, it could try using `pkg-config` and use it if it's there.

Anyway, I tried with `--with-idn=/usr/pkg` as you suggested, and that worked.

Then I noticed that gpgme and notmuch were not enabled, so I added `--with-gpgme=/usr/pkg --with-notmuch=/usr/pkg`. However, that was not sufficient, I had to additionally add `--gpgme --notmuch`. I think that the `--with-foo` options should imply `--foo`.
When I added those, the build succeeded.

One other problem: autosetup does not correctly detect my lua. It's in `/usr/include`, but neither `--lua` nor `--lua --with-lua=/usr` are sufficient. With `--lua` alone:

````
Checking for include/lua53/lua.h...no
Checking for include/lua5.3/lua.h...no
Checking for include/lua/lua.h...no
Checking for include/lua.h...no
Checking for include/lua52/lua.h...no
Checking for include/lua5.2/lua.h...no
Checking for include/lua/lua.h...no
Checking for include/lua.h...no
````
With `--lua --with-lua`:
````
Checking libs for luaL_openlibs...no
Checking for include/lua52/lua.h...no
Checking for include/lua5.2/lua.h...no
Checking for include/lua/lua.h...no
Checking for include/lua.h...yes
Checking libs for luaL_openlibs...(cached) no
Error: Unable to find Lua
````

The files exist:
````
# grep -r luaL_openlibs /usr/include/lua* 
/usr/include/lualib.h:LUALIB_API void (luaL_openlibs) (lua_State *L);
````

I don't see what fails and I don't see where this information is cached (`git status --ignored` output is empty)

**Comment by gahr** [comment]
I think you might be having lua5.1 installed, and that requires `-lm`. I'm seeing the same behaviour on NetBSD 7.1. Working on it...

**Comment by flatcap** [comment]
> the `--with-foo` options should imply `--foo`

I like that idea

**Comment by msteveb** [comment]
FYI, Jim Tcl ( e.g. msteveb/jimtcl@5dd45e7baf34854b588dd77553f000fd1) uses pkg-config if available or falls back to manual checks if not

**Comment by 0-wiz-0** [comment]
Thank you, `--lua --with-lua=/usr` works now on NetBSD. I wonder why `--lua` is not enough, since `/usr` is in the compilers default search paths.

**Comment by msteveb** [comment]
@0-wiz-0 You can use --debug and look in config.log to find out why something wasn't found.

**Comment by 0-wiz-0** [comment]
Thanks for the hint, that makes it clearer what happens. Not for `--lua` though. I guess the fix would be to define a default for `lua_prefix` (of `/usr`) if nothing else is defined.

**Comment by gahr** [comment]
I will need to review the lua part, I got it working on NetBSD too but it is kind of hackish right now..

**Comment by gahr** [comment]
So, the default prefix is `/usr/local` on anything by Linux. The Lua checks explicitely look for the lua include file in the locations you're seeing in the `Checking for...` messages, prepended by the value of `prefix`.

My solution on NetBSD was to define `--prefix=/usr/pkg`, which is where 3rd party packages seem to be installed. How does that sound to you?

**Comment by gahr** [comment]
> FYI, Jim Tcl ( e.g. msteveb/jimtcl@5dd45e7) uses pkg-config if available or falls back to manual checks if not

Thanks @msteveb , but I'm not sure the added complexity is worth it, since we have to be able to find 3rd party packages even without pkg-config... unless we decide that we want to actually require pkg-config, at which point yes, the code could be simplified.

**Comment by 0-wiz-0** [comment]
@gahr: I think the best solution would be (in absence of a `--with-lua=/path`) to search in the same paths that the compiler will look for the headers, and that does include `/usr`. Or use the compiler itself to look for the headers.
But this is just polish, as long as `--with-lua=/usr` works, I can get it to work.

**Comment by gahr** [comment]
@0-wiz-0 I will try to remind myself of why I did it this way for lua...

**Comment by gahr** [comment]
The thing is, if you:

* have Lua *both* in `/usr` and in `/usr/pkg`
* have, say, *some_package* in `/usr/pkg`
* configure with `--prefix=/usr/pkg` or `--with-some_package=/usr/pkg`

your compile lines will end up containing `-I/usr/pkg/include` and your link lines `-L/usr/pkg/lib`, which means you'll pick up the Lua headers and libs from `/usr/pkg` anyway.

The system doesn't really support picking 3rd party dependencies from several places, if one exists in at least two places.

On the NetBSD system I have access to, I actually have Lua 5.1 in /usr and 5.2 in /usr/pkg. If I default to do the checks against the compiler's default search path I'll end up using the 5.1 headers at configure stage and the 5.2 at build stage.

I don't have a ready solution for this.

**Comment by gahr** [comment]
@0-wiz-0, @d-k-c 
Having said that, I think you input has convinced me that defaulting to prefix to look for dependencies is not the way to go. I'll review and fix that logic, let's see where it brings us then.

**Comment by gahr** [comment]
I had a good sleep on it and re-checked what autoconf is doing. It turns out that it also looks for dependencies in prefix.

See [here](https://github.com/neomutt/neomutt/blob/master/configure.ac#L27-L33) where it sets `mutt_cv_prefix` and [here](https://github.com/neomutt/neomutt/blob/master/configure.ac#L322-L325) and [here](https://github.com/neomutt/neomutt/blob/master/configure.ac#L751-L757) where it uses this value to look for include files.

The problem it solves is when a package installs header files into a subdirectory of the default search path, e.g., `$prefix/include/lua-5.3/lua.h`. The code uses `#include <lua.h>`. How do I tell the compiler to use `-I$prefix/include/lua-5.3` without using `$prefix` in CFLAGS?

**Comment by msteveb** [comment]
Like this?

```
cc-with [list -cflags -I[get-define prefix]/include/lua-5.3] {
    cc-check-includes lua.h
}
```

**Comment by gahr** [comment]
Yes - that's basically what I'm doing. The point raised was that prefix should only be used to define the installation directories, not where to look for dependencies.

But I'm not convinced anymore that that's correct.

Further checks with the current autoconf reveal that a prefix different from the local installation base results in not being able to build.

I think I'lll leave things as they currently are, and fix minor specific issues such as cross compilation as we go.

**Comment by msteveb** [comment]
Yes, in general the problem of finding packages on an arbitrary system is almost intractable.
Many packages expect the user to provide CFLAGS, etc. to find dependencies. Thus the appeal of something like pkg-config, but then you still need to know where to look for .pc files.
Otherwise you can do something similar to fossil:

                set ssldirs "{} /usr/sfw /usr/local/ssl /usr/lib/ssl /usr/ssl \
                             /usr/pkg /usr/local /usr /usr/local/opt/openssl"

i.e. just look everywhere remotely plausible.

**Comment by flatcap** [comment]
I've pushed the renaming changes to master and rebased this PR.
21430b9 syncs most of the changes.  All that's left is to move is:
```diff
< usr/bin/pgpewrap
< usr/bin/pgpring
< usr/bin/smime_keys
---
> usr/lib64/neomutt/pgpewrap
> usr/lib64/neomutt/pgpring
> usr/lib64/neomutt/smime_keys
```

gahr:
Thank you @flatcap, I've fixed the remaining paths.

PR #761, opened 2017-09-15
tidy: use explicit NUL for strings

sileht:
Recently we have used explicit NUL everywhere. But the new
imap flags feature have reintroduce one.

This change fixes it.


gahr:
Your change doesn't do what you say it does. The code as it is is just fine :)

flatcap:
I agree with @gahr, there's nothing to be gained by this change.
However, if you were to make the change at the same time as fixing the dodgy `sizeof(buf)`... :-)

sileht:
The sizeof fix is here: https://github.com/neomutt/neomutt/pull/762
I close this one it's useless.

PR #763, opened 2017-09-18
Rename installed files to avoid clashes with Mutt package

flatcap:
I've made some naming changes to the installed files.
Hopefully this will mean that NeoMutt and Mutt packages could be installed in parallel, without conflict.

**Your suggestions are welcomed.**

### 811692de rename man pages

The duplicated man pages have been given a `_neomutt` suffix
These pages are still easily discovered with 'man -k'

* /usr/share/man1/pgpewrap_neomutt.1
* /usr/share/man1/pgpring_neomutt.1
* /usr/share/man1/smime_keys_neomutt.1
* /usr/share/man5/mbox_neomutt.5
* /usr/share/man5/mmdf_neomutt.5

### dcfba3d2 move helper programs to lib dir

Moving these programs out of the path means changing the config to find them.
This has been done, but any distro that installs to, say, `/usr/local` will need to change their config.
The alternative is a little ugly.  Leave them in `/usr/bin`, but add a `_neomutt` suffix.

* /usr/lib/neomutt/pgpewrap
* /usr/lib/neomutt/pgpring
* /usr/lib/neomutt/smime_keys

### cb7f2deb move mime.types to doc dir

Many distros already have a package that provides `/etc/mime.types`
I've moved our copy into the docs.  Distros may install this to `/etc` if they wish.

* /usr/share/doc/neomutt/mime.types

### 5f92df6c contrib: change mutt references to neomutt

This is a quick'n'dirty search and replace of 'Mutt' with 'NeoMutt' in the contrib dir.
smime_keys should now work.
The lua config hasn't been changed, because it still refers to a 'mutt' object in the main code.

/cc @neomutt/downstream-packagers @neomutt/arch-users @neomutt/debian-users @neomutt/gentoo-users @neomutt/homebrew-users


somini:
Since this changes the default binary name to `neomutt`, it might be important to send a message about this in the installation script.

hobarrera:
Have you considered `nmutt` instead of `neomutt`, simply because it's shorter to type (also, it can be written using only the right hand). `neovim`'s binary is actually called `nvim`, as a common precedent.

flatcap:
> Have you considered nmutt instead of neomutt

I'll put it to a vote.

> (also, it can be written using only the right hand)

Hmm... Yes, but not if you're a touch-typer ('t' is left index-finger).

flatcap:
@somini I'm planning to work with each of the downstream distros.
The Arch Team is largest, so they get to start.  [My rough notes are here](https://gist.github.com/flatcap/a27a31769008ba38f363bb5e6f6a4b42).
They document all the changes we've made and are going to make, and things they might need to think about.

flatcap:
The votes came in about 2:1 in favour of `neomutt`,
but as we're all going to `alias mutt=neomutt` it barely matters :-)

PR #764, opened 2017-09-18
Add support for IMAP custom keywords

flatcap:
This is a continuation of #399 by @sileht 

This change introduces IMAP keyword support.

* index_format gets %j and %J to display them
* pattern gets j to search/match imap keywords
* colors.c gets index_keywords to color them
* OPS gets imap-edit-keywords to edit them


gahr:
Some random points on the technical (mostly) aspect of things, which some notes on the concept, too:

1. Member `tags` of `struct Header`
I think it should be properly typeized as `struct HeaderTags * tags`. The `HEADER_TAGS` macro in `mutt_tags.c` wouldn't be needed, then. **`mutt_tags.h`**

2. For tags that do not have a transform, the original tag is strdup'd into the transform member. This duplicates the memory needed. How about using NULL to indicate that no transform is wanted?

3. `struct HeaderTag` vs. `struct HeaderTags`
I sense some confusion looking at these two structures.. there's some duplication too.

4. function that are meant to be private to `mutt_tags.c` should not be declared in `mutt_tag.h` and should be static.

5. why doesn't `hdr_tags_free` actually free the `tags` member? This is left to do to `mutt_free_header` in `muttlib.c`.

6. declaration of `nm_header_get_tags_transformed` is a leftover

7. inclusion of `mutt_tags.h` should be either always or never guarded by `USE_IMAP || USE_NOTMUCH`

More on the conceptual side of things.

1. I am very confused about the terminology: `edit-label` impacts the `X-Label` header, `modify-label` impacts "tags". So, what are "tags" and what are "labels" now?

2. What would you expect to happen on `:exec modify-label<CR>unread<CR>` on an IMAP-based message?

3. are you sure that tag-transforms and tag-formats should be applied to IMAP labels as well? I fear most of the use cases (at least the documented ones) would result in a "Invalid IMAP keyword". Does this make you re-think about the universality of the concept of "label" or "tag"?

sileht:
On Mon, Sep 18, 2017 at 02:21:13PM +0000, Pietro Cerutti wrote:
>More on the conceptual side of things.
>
>1. I am very confused about the terminology: `edit-label` impacts the `X-Label` header, `modify-label` impacts "tags". So, what are "tags" and what are "labels" now?

This confusion exists even without my patch. My intent is just to
implement the same tag thing that notmuch have.

I don't thinkg it's a good idea to fix this terminology in the same pull
request. That should be discussed into another issue even if we never
merge this PR.

>2. What would you expect to happen on `:exec modify-label<CR>unread<CR>` on an IMAP-based message?

I guess you're thinking that currently that will do two "set flag"
operations on the imap servers. I don't think it's a issue.

>3. are you sure that tag-transforms and tag-formats should be applied to IMAP labels as well?
> I fear most of the use cases (at least the documented ones) would result in a "Invalid IMAP keyword".

This is the exact reason why we need tag-transforms and tag-formats,
the imap flag have very limited set of character. It's up to the client to
show in the UI fancy names. Thunderbird uses $label1, $label2, $label3
for example, I don't want to see them as-is. I want to transform them
with the same label I have in thunderbird. I also associates colored emojy to
my flag.

> Does this make you re-think about the universality of the concept of "label" or "tag"?

This concept are implicitly already defined in mutt, label is X-label
and tag are notmuch_tag (and I hope soon imap flag too).

But I agree the terminology looks bad. Note that 'tag' in mutt also refer to the
'tag-pattern' key bindings not only "notmuch tag" :(.

So, I haven't changed the current terminology because I think it should be in
another PR and it will break backward config compat if we change
keybinding, configuration names, ...

(I will other technical issues)
-- 
Mehdi Abaakouk
mail: sileht@sileht.net
irc: sileht

sileht:
On Mon, Sep 18, 2017 at 02:21:13PM +0000, Pietro Cerutti wrote:
>Some random points on the technical (mostly) aspect of things, which some notes on the concept, too:

These technical issues have been inherited from the tags code of notmuch, I
have just moved them to mutt_tags.* . Not sure they should be part of this PR. But
they looks simple enough, so I have tried to fix them.

>1. Member `tags` of `struct Header`
>I think it should be properly typeized as `struct HeaderTags * tags`. The `HEADER_TAGS` macro in `mutt_tags.c` wouldn't be needed, then. **`mutt_tags.h`**

Fixed

>2. For tags that do not have a transform, the original tag is strdup'd into the transform member. This duplicates the memory needed. How about using NULL to indicate that no transform is wanted?

I have tried, and I don't think that worse it. This adds a bunch
of if/else everywhere to save a small amount of memory. All this
duplication is because how the UI works. Everything should be ready in
memory, hdrline just want where all the strings are located. Having
these dups make the thing more understandable and less error prone.

It's just my opinion, but I prefer simple code and not saving memory for
this kind of software.

>3. `struct HeaderTag` vs. `struct HeaderTags`
>I sense some confusion looking at these two structures.. there's some duplication too.

I have renamed HeaderTag to TagList to avoid confusion

>4. function that are meant to be private to `mutt_tags.c` should not be declared in `mutt_tag.h` and should be static.

Done.

>5. why doesn't `hdr_tags_free` actually free the `tags` member? This is left to do to `mutt_free_header` in `muttlib.c`.

Done.

>6. declaration of `nm_header_get_tags_transformed` is a leftover

Removed.

>7. inclusion of `mutt_tags.h` should be either always or never guarded by `USE_IMAP || USE_NOTMUCH`

I have deliberately done this, because I changed the MxOps structure and
most methods of this structure are optional. tags is just one more thing
that a backend can implement.

If you really thing I should add the "define" guard, I can.

-- 
Mehdi Abaakouk
mail: sileht@sileht.net
irc: sileht

gahr:
Re string duplication for transforms, since you got the encapsulation correct, wouldn't the "bunch of if/else everywhere" be limited to `hdr_tags_get_transformed[_for]`?

gahr:
Re guarded inclusion, I just said "either always or never". If you feel like having them unguarded, fine, but please let them all be unguarded then.

gahr:
I see a strange behaviour with IMAP. I have this config:

```
set index_format    = "%4C %?M?⤷%5M &(%4c) ? %-5.5Z [%-15.15D]  %-35.35F  (%g)  %s"
tag-transforms "unread" "not_read"
```

And I do the following:

1. open a mailbox, with the first message selected
2. `:exec modify-labels<CR>unread<CR>`
3. selection moves to the next message, but the `%g` group is not populated (empty parentheses)
4. go back to the previous message
5. apply the label again: `:exec modify-labels<CR>`, mutt says `Keywords: unread`, so the label is there

If I close/reopen the mailbox, the `%g` group is correctly populated and I do see `not_read`.

gahr:
The '%G' group of `index_fomat` is not working here. It does not print any tags and it eats the subsequent character.

Take a message with the "Foo" tag applied:

`:set index_format="[%g]"` --> `[Foo]`
`:set index_format="[%G]"` --> `[`

gahr:
Please, do chose a term for the thing and use it consistently. We cannot have `tags`, `labels`, and `keywords` all being used interchangeably and mean the same thing except when they don't (see `modify-labels` vs `edit-label`.

sileht:
On Tue, Sep 19, 2017 at 06:58:04AM +0000, Pietro Cerutti wrote:
>Re guarded inclusion, I just said "either always or never". If you feel like having them unguarded, fine, but please let them all be unguarded then.

That was already the case. But I have rechecked a found a new useless
ifdef that I just removed.

-- 
Mehdi Abaakouk
mail: sileht@sileht.net
irc: sileht

sileht:
On Tue, Sep 19, 2017 at 07:35:54AM +0000, Pietro Cerutti wrote:
>Please, do chose a term for the thing and use it consistently. We cannot have `tags`, `labels`, and `keywords` all being used interchangeably and mean the same thing except when they don't (see `modify-labels` vs `edit-label`.

This is unrelated to my feature, the wording is already bad in neomutt.
Just check: https://www.neomutt.org/feature/keywords the three words are
used in the documentation (and the code also reflect that).

Or https://www.neomutt.org/feature/notmuch it uses sometimes labels,
sometimes tags.

I'm OK to work of this renaming issue. But that have nothing to do with
this new feature.

My take is to name the backend thing "tags" (for imap/notmuch) and the
"X-label" thing "label".

I don't want to change this in this change because it will/may break
backcompatibility.

--
Mehdi Abaakouk
mail: sileht@sileht.net
irc: sileht

sileht:
On Tue, Sep 19, 2017 at 07:34:49AM +0000, Pietro Cerutti wrote:
>The '%G' group of `index_fomat` is not working here. It does not print any tags and it eats the subsequent character.
>
>Take a message with the "Foo" tag applied:
>
>`:set index_format="[%g]"` --> `[Foo]`
>`:set index_format="[%G]"` --> `[`

That occurs also without my change. And this is expected. %G expects
another char and here the other char is ']'. So it looks to tag-formats
for ']' and don't find it so it print nothing.

-- 
Mehdi Abaakouk
mail: sileht@sileht.net
irc: sileht

sileht:
On Tue, Sep 19, 2017 at 06:54:48AM +0000, Pietro Cerutti wrote:
>Re string duplication for transforms, since you got the encapsulation correct, wouldn't the "bunch of if/else everywhere" be limited to `hdr_tags_get_transformed[_for]`?

That part is clean. The function that looks a bit awful with all the if/else
is hdr_tags_add(). But I can do the optimisation if needed, that's not a
big deal.

-- 
Mehdi Abaakouk
mail: sileht@sileht.net
irc: sileht

sileht:
On Tue, Sep 19, 2017 at 07:17:02AM +0000, Pietro Cerutti wrote:
>And I do the following:
>
>1. open a mailbox, with the first message selected
>2. `:exec modify-labels<CR>unread<CR>`
>3. selection moves to the next message, but the `%g` group is not populated (empty parentheses)
>4. go back to the previous message
>5. apply the label again: `:exec modify-labels<CR>`, mutt says `Keywords: unread`, so the label is there

That weird, I can't reproduce the issue when I type unread<CR>, the row
is instantly updated.

gahr:
> That part is clean. The function that looks a bit awful with all the if/else is hdr_tags_add()

Wouldn't this be enough?

```diff
diff --git a/mutt_tags.c b/mutt_tags.c
index 65651a55..742a6c15 100644
--- a/mutt_tags.c
+++ b/mutt_tags.c
@@ -162,8 +162,6 @@ static void hdr_tags_add(struct Header *h, char *new_tag)
   char *new_tag_transformed = NULL;

   new_tag_transformed = hash_find(TagTransforms, new_tag);
-  if (!new_tag_transformed)
-    new_tag_transformed = new_tag;

   ttmp = safe_calloc(1, sizeof(*ttmp));
   ttmp->name = safe_strdup(new_tag);
@@ -189,7 +187,8 @@ static void hdr_tags_add(struct Header *h, char *new_tag)
   mutt_str_append_item(&h->tags->tags, new_tag, ' ');

   /* expand the transformed tag string */
-  mutt_str_append_item(&h->tags->tags_transformed, new_tag_transformed, ' ');
+  if (new_tag_transformed)
+    mutt_str_append_item(&h->tags->tags_transformed, new_tag_transformed, ' ');
 }

 /**
```

gahr:
> My take is to name the backend thing "tags" (for imap/notmuch) and the "X-label" thing "label".

The backend thing for imap is called Flags. Tags are already two separate and unrelated things in mutt: mutt's own method of selecting one or more messages to perform actions on them, and notmuch's filtering mechanism. Pretty please, don't add yet another new concept under the same name.

sileht:
@gahr do you have a idea/proposal for the naming ?

gahr:
Not really, sorry, I don't have a good solution for the unification of these two concepts. Just asking not to introduce more confusion of terms that there already is (see edit-labels vs. modify-label).

gahr:
Thanks for your prompt improvements so far! I think you have (almost) reahed a technically sound implementation.
If you don't mind, I'd like to have another round of review by the end of the week and come back to you with further comments, if any.
Again, thanks for bearing with my critical input ;)

gahr:
> That was already the case. But I have rechecked a found a new useless ifdef that I just removed.

Ok, I saw the change. What I meant is this: "mutt_tags.h" is conditionally included in muttlib.c, and unconditionally included everywhere else.

gahr:
One thing that's been giving me a strange feeling is the relation between the two structures `HeaderTags` and `TagList`.

Here's my understanding, please correct it if it's wrong.

The `HeaderTags` structure store per-message tags (`tags`) and their transformations (`tags_transformed`) in a space-separated string. Each element in these two strings (which must count the same number of elements) is also a node in the linked list `tag_list`.

The `TagList` structure stores each of the corresponding pairs from (`tags`, `tags_transformed`) as a separate node.

Also, the `HeaderTags` structure containes a space-separated `tags_with_hidden` string which contains all tags, hidden or not.

If my understanding is correct, I think we could simplify matters a lot by just using a linked-list of `TagNode` structures, each with a name, a transformed name, and a hidden flag. Using the macros from queue.h, that would be something like this:

```c
STAILQ_HEAD(TagHead, TagNode);
struct TagNode
{
    char *name;
    char *transformed;
    bool hidden;
    STAILQ_ENTRY(TagNode) entries;
};

Thoughts?

sileht:
> Ok, I saw the change. What I meant is this: "mutt_tags.h" is conditionally included in muttlib.c, and unconditionally included everywhere else.

Thanks, an oversight, fixed

sileht:
On Wed, Sep 20, 2017 at 08:11:53AM +0000, Pietro Cerutti wrote:
>One thing that's been giving me a strange feeling is the relation between the two structures `HeaderTags` and `TagList`.
> ...
>Thoughts?

Another thing already present in nm_tags_* code I have moved and reused. But I
agree that should be replaced by a simple linked-list.

I'm maybe enough familliar with mutt code now to do this kind of change. I will try.

sileht:
I have pushed some new changes:
* I have continued to remove the word header/label from the driver tags code to remove some confusion.
* I have removed the header.h dependency of the tagging code. mutt_tags.c is only about tags list manipulation now.
* I have implemented the @gahr linked-list proposal to remove the three memory allocations for each tags.

gahr:
Cool, thank you! One thing I noticed at a first glance:  `STAILQ_INIT` should be called at the end of `driver_tags_free`, see [mutt_list_free](https://github.com/neomutt/neomutt/blob/master/list.h#L84-L95). This also allows you to get rid of `driver_tags_init`, which just wrap a call to `STAILQ_INIT`.

sileht:
I have removed driver_tags_init and fixed driver_tags_free

gahr:
Thanks! I don't know if that's ported code too, but ` char *tags = safe_calloc(0, sizeof(char));` in `driver_tags_getter` looks bogus to me. Wouldn't `NULL` be enough as an initializer?

gahr:
Also, it looks like the `sep` argument to `mutt_str_append_item` should really be a `char` and `\0` be used instead of 0, when needed. Especially, the use of `32` instead of ` ` (space) in `msg_parse_flags` is weird.

sileht:
On Thu, Sep 21, 2017 at 02:53:30PM +0000, Pietro Cerutti wrote:
>Thanks! I don't know if that's ported code too, but ` char *tags = safe_calloc(0, sizeof(char));` in `driver_tags_getter` looks bogus to me. Wouldn't `NULL` be enough as an initializer?

Indeed mutt_str_append_item allocate it correctly for me, fixed.

-- 
Mehdi Abaakouk
mail: sileht@sileht.net
irc: sileht

sileht:
On Thu, Sep 21, 2017 at 07:56:20AM -0700, Pietro Cerutti wrote:
>Also, it looks like the `sep` argument to `mutt_str_append_item` should really be a `char` and `\0` be used instead of 0, when needed. Especially, the use of `32` instead of ` ` (space) in `msg_parse_flags` is weird.

I have noticed that too in notmuch code, good cleanup to do. Fixed.

gahr:
> I have noticed that too in notmuch code, good cleanup to do. Fixed.

I think you missed the definition :)

sileht:
> I think you missed the definition :)
> 
> Edit: and the declaration

Fixed, I have thought it was already char.

I have also resolved the merge conflict.

PR #767, opened 2017-09-18
Pull request for devel/keywords

sileht:
tags: rename HeaderTag to TagList
tags: make private method private
tags: remove needs of HEADER_TAGS macro
notmuch: remove unused methods declaration
imap: fix memory leak
tags: always pass correct buffer length
Tidying the code
Share label editing between imap/notmuch
Add IMAP keywords support
tags: Add index_format %J
Make the notmuch tags generic


sileht:
Not pushed at the correct place

Issue #774, opened 2017-09-23
Change int boolean flags into bool (aka _Bool) for better clarity


seishinryohosha:
* Why do you not like the current state?
  Using int for boolean flags is not very intuitive in a method prototype.
* What would you like to see to change?
  Change all int flags into bool flags.
  
I would start with changing all headers alphabetically and make for each letter 
a unique branch for easier reviewing.
After changing all top level headers I would go alphabetically in each folder.

For example the first branch would be `devel/switch_to_bool_a`:

So the branch would change:
* account.h
* address.h
* alias.h
* attach.h
* all calls to the prototypes in other files

Now I have some questions before I would be able to start:

1. Is the breakdown appropriate? 
    Or is it too granular or too shallow?
2. Are some files using int intentionally? 
    E.g. planning on using int in the future for other reasons, except boolean flags?
3. Other points I might be missing right now?

flatcap:
Yes, it's a good idea.  But a little backstory first...

Even before we started NeoMutt, Mutt had a C99 requirement in the build.
This should allow:
- `bool`
- local for variables: `for (int i = 0; ...`
- variables declared nearer their use

Mutt didn't make use of any of these.
Using a lot of unix- and vim-fu, I made a lot of changes, I:
- changed the type of variables in functions to `bool`
- shrank the scope of `for` variables

and fixed the naming of:
- Global variables
- `struct`s
- function names

Besides vim, I also used [coccinelle](https://github.com/neomutt/coccinelle) -- a code transformation tool.

However careful I was, each time I managed to **** something up.
After days of staring at changes, you become blind to mistakes.

**Tip:** Keep the changes small.  I'd suggest one commit per prototype changed.
It will be a pain to organise, but it will mean we can divide up the job of reviewing.


If you're a CS student, you may be able to use Coccinelle, or Clang's AST to automate the changes.
If so, please teach us how :-)

seishinryohosha:
I definitely look into the tools!
My proposed strategy would be as follow:
Follow the alphabetical PR versioning, but only for each header a single PR and doing a commit for each prototype a single commit.
This approach would even further divide it, and make the review easier.

First only change `int -> bool`. After that, start a new iteration with for example shrank the scope of `for` variables.

I looked into `account.h` and noticed that all functions, which return `int` only return `-1` or `0`.
Is it planned to change them also to `bool` or just leave them?
A quick `grep` on `mutt_account_fromurl` only used it in a boolean way.

[...]

PR #776, opened 2017-09-23
Devel/issue 774 attach.h

seishinryohosha:
@neomutt/reviewers 
First PR of hopefully a series ;)
First commits are very granular, later ones are a bit more packed, please
tell which is more appropriate.
If rebasing is needed, e.g. unify more commits, just notify me.

* **What does this PR do?**
Refactors `int -> bool` in `attach.h` and it's deps.
Strips leading space e.g. `foo ()` -> `foo()`, splits too long prototypes to match ~90 chars
Yes
    * `struct Menu`s `tagprefix` changed from `int` to `bool`.
      I checked all occurrences, and only found boolean uses of it.
    * `static int can_print()` in `recvattach.c` now returns `bool`
    * `static short check_can_decode()` in `recvcmd.c` now returns `bool`
#774

### DIFF
--- a/attach.h
+++ b/attach.h
@@ -68,28 +68,35 @@ struct AttachCtx
   short body_max;
 };
 
-void mutt_attach_init (struct AttachCtx *actx);
+void mutt_attach_init(struct AttachCtx *actx);
 void mutt_update_tree(struct AttachCtx *actx);
 int mutt_view_attachment(FILE *fp, struct Body *a, int flag, struct Header *hdr,
                          struct AttachCtx *actx);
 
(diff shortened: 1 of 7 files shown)


gahr:
> Strips leading space e.g. foo () -> foo(), splits too long prototypes to match ~90 chars

Not sure about these.. if they were clang-format -compliant already, then there shouldn't be any need for these modifications.. @flatcap?

flatcap:
Rebase, fix two unconverted function args.

> if they were clang-format -compliant already

I don't routinely format the header files, because clang would mess up the alignment of lists.
Those spaces crept in where I've merged upstream changes, so it's OK for them to die.

PR #781, opened 2017-09-24
fix: crash in mutt_substrdup

flatcap:
- Make `mutt_substrdup()` more robust
- Protect a few `printf`s against `NULL` args

### DIFF
--- a/browser.c
+++ b/browser.c
@@ -1590,6 +1590,8 @@ void _mutt_select_file(char *f, size_t flen, int flags, char ***files, int *numf
               mutt_message(_("Mailbox deleted."));
               init_menu(&state, menu, title, sizeof(title), buffy);
             }
+            else
+              mutt_error(_("Mailbox deletion failed."));
           }
           else
             mutt_message(_("Mailbox not deleted."));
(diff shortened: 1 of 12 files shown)


gahr:
> Make mutt_substrdup() more robust

Wonderful example of having an assert(begin) and clearly state the contract in the documentation. There is no use case where calling `mutt_substrdup` with the first parameter being NULL makes sense, so why handle the case at all?

If you think an assert is too much (i.e., you don't want to crash under any circumstances), then at least a `mutt_error` would be nice:

`Programming error: mutt_substrdup was passed a NULL as first argument, please report...`

flatcap:
> assert(begin)

I think that's a bit much, especially for string functions.

> at least a mutt_error would be nice:

Yes.  I'll add one.

This is the situation where my macro would be used (mentioned on neo-devel :-)
```c
return_val_if_fail(begin, -1);
```

> so why handle the case at all?

With the exception of `malloc()`, etc, I'd like each function to have a defined return for all possible inputs.

The string functions are an awful mess.  There should be no excuse for not checking pointers before use.
But any changes would need to be done carefully to make sure none of the callers are relying on side-effects.

PR #782, opened 2017-09-25
devel/issue 774 bcache.h transit to `bool`

seishinryohosha:
@neomutt/reviewers 
Next is `bcache.h`. This only affects `mutt_bcache_put()`s `tmp` flag.

Affected files
* `bcache.c` -> `mutt_bcache_put()`
* `bcache.h` -> `mutt_bcache_put();`
* `imap/message.c` -> `msg_cache_put()`
* `nntp.c` -> `nntp_open_message()`
* `pop.c` -> `pop_fetch_message()` -> if can use body cache

* **What does this PR do?**
Refactors bcache.h to make use of `bool`
Probably not
#774
* **Further mentions**
`int mutt_bcache_exists()` may return `bool`, as it answers the question: "Does bcache exists? Y/N"

### DIFF
--- a/bcache.c
+++ b/bcache.c
@@ -150,13 +150,19 @@ FILE *mutt_bcache_get(struct BodyCache *bcache, const char *id)
   return fp;
 }
 
-FILE *mutt_bcache_put(struct BodyCache *bcache, const char *id, bool tmp)
+FILE *mutt_bcache_put(struct BodyCache *bcache, const char *id)
 {
   char path[LONG_STRING];
   struct stat sb;
 
   if (!id || !*id || !bcache)
     return NULL;
+
+  if (snprintf(path, sizeof(path), "%s%s%s", bcache->path, id, ".tmp") >= sizeof(path))
+  {
+    mutt_error(_("Path too long: %s%s%s"), bcache->path, id, ".tmp");
+    return NULL;
+  }
 
   if (stat(bcache->path, &sb) == 0)
   {
@@ -175,7 +181,6 @@ FILE *mutt_bcache_put(struct BodyCache *bcache, const char *id, bool tmp)
     }
   }
 
-  snprintf(path, sizeof(path), "%s%s%s", bcache->path, id, tmp ? ".tmp" : "");
   mutt_debug(3, "bcache: put: '%s'\n", path);
 
   return safe_fopen(path, "w+");
--- a/bcache.c
+++ b/bcache.c
@@ -150,7 +150,7 @@ FILE *mutt_bcache_get(struct BodyCache *bcache, const char *id)
   return fp;
 }
 
-FILE *mutt_bcache_put(struct BodyCache *bcache, const char *id, int tmp)
+FILE *mutt_bcache_put(struct BodyCache *bcache, const char *id, bool tmp)
 {
   char path[LONG_STRING];
   struct stat sb;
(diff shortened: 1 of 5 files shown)


bcache.c
```diff
@@ -25,6 +25,7 @@
 #include <dirent.h>
 #include <errno.h>
 #include <limits.h>
+#include <stdbool.h>
```
gahr: This is already included in bcache.h for the prototype.

imap/message.c
```diff
@@ -26,6 +26,7 @@
 #include "config.h"
 #include <ctype.h>
 #include <limits.h>
+#include <stdbool.h>
```
gahr: Same as before.

nntp.c
```diff
@@ -25,6 +25,7 @@
 #include "config.h"
 #include <ctype.h>
 #include <limits.h>
+#include <stdbool.h>
 #include <string.h>
```
gahr: Same

pop.c
```diff
@@ -24,6 +24,7 @@
 #include "config.h"
 #include <errno.h>
 #include <limits.h>
+#include <stdbool.h>
 #include <stdio.h>
```
gahr: Ditto :)

seishinryohosha:
> It occurs to me that this tmp parameter is never used.. How about we kill it altogether?
Yes, it's always `true`. Also `mutt_bcache_commit` always uses ".tmp".

If we always use ".tmp", should we reduce the size of `BodyCache->path` to get rid of the Warning 
message below?
```C
bcache.c:178:37: warning: '%s' directive output may be truncated writing up to 4 bytes into a region of size between 1 and 256 [-Wformat-truncation=]
   snprintf(path, sizeof(path), "%s%s%s", bcache->path, id, tmp ? ".tmp" : "");
                                     ^~
bcache.c:178:3: note: 'snprintf' output 1 or more bytes (assuming 260) into a destination of size 256
   snprintf(path, sizeof(path), "%s%s%s", bcache->path, id, tmp ? ".tmp" : "");
   ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
```

gahr:
You wouldn't know how much to reduce, since the `id` parameter is now known in advance.. how about checking?

```diff
--- a/bcache.c
+++ b/bcache.c
@@ -158,6 +158,13 @@ FILE *mutt_bcache_put(struct BodyCache *bcache, const char *id, bool tmp)
   if (!id || !*id || !bcache)
     return NULL;

+  if (snprintf(path, sizeof(path), "%s%s%s", bcache->path, id, ".tmp")
+      >= sizeof(path))
+  {
+      mutt_error(_("Path too long: %s%s%s"), bcache->path, id, ".tmp");
+      return NULL;
+  }
+
   if (stat(bcache->path, &sb) == 0)
   {
     if (!S_ISDIR(sb.st_mode))
@@ -175,7 +182,6 @@ FILE *mutt_bcache_put(struct BodyCache *bcache, const char *id, bool tmp)
     }
   }

-  snprintf(path, sizeof(path), "%s%s%s", bcache->path, id, tmp ? ".tmp" : "");
   mutt_debug(3, "bcache: put: '%s'\n", path);

   return safe_fopen(path, "w+");
```

seishinryohosha:
Sure even better, thought it was intentionally silently dropped.

flatcap:
- Rebased
- Applied @gahr's patch
- Dropped unused 'tmp' param
- Buffer size increased (in a separate commit, to fix build warnings)

PR #795, opened 2017-09-27
fix: overflow in mutt_mktime()

### LINKED ISSUE #794: Integer overflow in mutt_mktime()
On 32-bit systems, signed integer overflow in `mutt_mktime()` is possible.
To reproduce, build NeoMutt with `-fsanitize=undefined` and run:

    $ printf 'From Wed Nov 0 0: 200' > intoverflow.mbox
    $ export UBSAN_OPTIONS=halt_on_error=1:abort_on_error=1:print_stacktrace=1
    $ neomutt -R -f intoverflow.mbox 2>&1 >/dev/null | cat
    date.c:276:5: runtime error: signed integer overflow: 68812040 * 60 cannot be represented in type 'long int'
        #0 0x8030208c in mutt_mktime .../neomutt/lib/date.c:276
        #1 0x801c6e41 in is_from .../neomutt/from.c:185
        #2 0x80206bb6 in mbox_parse_mailbox .../neomutt/mbox.c:290
        #3 0x80208e90 in mbox_open_mailbox .../neomutt/mbox.c:458
        #4 0x80229cb2 in mx_open_mailbox .../neomutt/mx.c:480
        #5 0x80202a88 in main .../neomutt/main.c:929
        #6 0xb7089285 in __libc_start_main (/lib/i386-linux-gnu/libc.so.6+0x18285)
        #7 0x8015dd70  (.../neomutt/neomutt+0x6ad70)

Tested with git master (0df7d370d81b174e87ca3a29ba3e92e81acd76f1).

Found using [American Fuzzy Lop](http://lcamtuf.coredump.cx/afl/).

flatcap:
mutt_mktime() can overflow when used on 32-bit arches.
This was due to a bogus comparision.

Now, extreme future dates will be interpreted as 2038-01-19.

Fixes #794

### DIFF
--- a/from.c
+++ b/from.c
@@ -134,6 +134,8 @@ int is_from(const char *s, char *path, size_t pathlen, time_t *tp)
     return 0;
   if (sscanf(s, "%d", &tm.tm_mday) != 1)
     return 0;
+  if ((tm.tm_mday < 1) || (tm.tm_mday > 31))
+    return 0;
 
   /* time */
   s = next_word(s);
@@ -146,6 +148,10 @@ int is_from(const char *s, char *path, size_t pathlen, time_t *tp)
   else if (sscanf(s, "%d:%d", &tm.tm_hour, &tm.tm_min) == 2)
     tm.tm_sec = 0;
   else
+    return 0;
+
+  if ((tm.tm_hour < 0) || (tm.tm_hour > 23) || (tm.tm_min < 0) ||
+      (tm.tm_min > 59) || (tm.tm_sec < 0) || (tm.tm_sec > 60))
     return 0;
 
   s = next_word(s);
@@ -173,6 +179,8 @@ int is_from(const char *s, char *path, size_t pathlen, time_t *tp)
 
   /* year */
   if (sscanf(s, "%d", &yr) != 1)
+    return 0;
+  if ((yr < 0) || (yr > 9999))
     return 0;
   tm.tm_year = yr > 1900 ? yr - 1900 : (yr < 70 ? yr + 100 : yr);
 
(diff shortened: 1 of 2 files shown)


jwilk:
You should also check for years far in the past.
For example `From Wed Jan 1 0:0 1901` causes:

    date.c:276:5: runtime error: signed integer overflow: -36288979 * 60 cannot be represented in type 'long int'

AFAICS you will also need range checks for `tm_mday`, `tm_hour`, `tm_min` and `tm_sec`.

flatcap:
Ah, that's an interesting point.
Elsewhere in neomutt's code, the lower limit for dates is 1901...

https://github.com/neomutt/neomutt/blob/0df7d370d81b174e87ca3a29ba3e92e81acd76f1/from.c#L177

but by the time the date's displayed in the index, anything older than 1970 is displayed as 1970-01-01.
For now, I've added a lower limit to prevent arithmetic overflow.

Anyone wishing to display mediaeval emails, will have to wait for us to perform a thorough audit of date handling :-)

jwilk:
With 24512c05d3d2fc71c88b71e11a120070b0fe01af, these From_ lines still trigger integer overflow:

    From Sat Sep 2000000000 0:0:0 0
    From Sat Sep 0 2000000000:0:0 0
    From Sat Sep 0 0:2000000000:0 0
    From Sat Sep 0 0:0:2000000000 0

flatcap:
You were right; I should have listened :-)  You're good at this.

I've added a heap more range-checking both where the date is first parsed and deeper where the date is converted.

Is there some automatic way to generate these test strings?
Or are they done by hand with your expertise in fuzzing?

jwilk:
All my test cases in this PR were made by hand.

PR #797, opened 2017-09-28
pager: index must be rebuild when check mailbox returns MUTT_REOPENED

### LINKED ISSUE #796: crash when <pipe-message> on deleted message
Steps to reproduce:

1. Open two separate instances of mutt.
2. Open an email in instance 2.
3. In instance 1, delete that email, and sync.
4. In instance 2, do :exec pipe-message
5. For the command to pipe to, put e.g. "echo" and press enter.

Result: SIGSEGV

I can reproduce on recent master (0df7d370 with "neomutt" binary) and with NeoMutt 20170912.

I did a bisect, which led to 4ed2b6f0. Before that commit, when I did the above recipe I would get the following:

> /home/scott/.mail/INBOX/cur/1506565365_2.13864.chro,U=39124,FMD5=7e33449f656h1e6l9m79b29t3f82c57e:2,S: No such file or directory (errno = 2)

Then if I press e.g. "q" to go back to Inbox, I see
Mailbox was externally modified.  Flags may be wrong.

To me, that behavior (of giving the error) makes sense. On current upstream mutt I don't think it gives an error anymore. I think it just doesn't do anything, but doesn't crash.

gdb gives this bt:

```
Thread 1 "neomutt" received signal SIGSEGV, Segmentation fault.
0x0000000000454a47 in snprintf (__fmt=0x4dad56 "%s/%s", __n=256, __s=0x7fffffffabb0 "\001") at /usr/include/x86_64-linux-gnu/bits/stdio2.h:64
64	  return __builtin___snprintf_chk (__s, __n, __USE_FORTIFY_LEVEL - 1,
(gdb) bt
#0  0x0000000000454a47 in snprintf (__fmt=0x4dad56 "%s/%s", __n=256, __s=0x7fffffffabb0 "\001") at /usr/include/x86_64-linux-gnu/bits/stdio2.h:64
#1  maildir_mh_open_message (msgno=<optimized out>, is_maildir=1, msg=<optimized out>, ctx=<optimized out>, ctx=<optimized out>) at mh.c:1447
#2  0x000000000045cacb in mx_open_message (ctx=ctx@entry=0x7ce550, msgno=112) at mx.c:1201
#3  0x000000000042173e in mutt_copy_message (fpout=0x820ca0, src=0x7ce550, hdr=hdr@entry=0x820b50, flags=0, chflags=16) at copy.c:766
#4  0x0000000000417be3 in pipe_msg (h=h@entry=0x820b50, fp=<optimized out>, decode=decode@entry=0, print=print@entry=0) at commands.c:384
#5  0x0000000000417ca5 in _mutt_pipe_message (h=h@entry=0x820b50, cmd=cmd@entry=0x7fffffffadc0 "copy_message_id", decode=0, print=print@entry=0, 
    split=<optimized out>, sep=<optimized out>) at commands.c:418
#6  0x0000000000418d86 in mutt_pipe_message (h=0x820b50) at commands.c:506
#7  0x000000000046b58e in mutt_pager (banner=banner@entry=0x0, 
    fname=fname@entry=0x7fffffffbe40 "/tmp/neomutt-steph-1000-10495-10518590499035943711", flags=flags@entry=66, extra=extra@entry=0x7fffffffbe10)
    at pager.c:2867
#8  0x0000000000418512 in mutt_display_message (cur=0x820b50) at commands.c:232
#9  0x0000000000427d8d in mutt_index_menu () at curs_main.c:2152
#10 0x0000000000409a16 in main (argc=1, argv=<optimized out>, env=<optimized out>) at main.c:936
(gdb) 
```

sileht:
When we call mx_check_mailbox() and something change
header in Context->hdrs may have been updated.

In case of MUTT_NEW_MAIL, the page rebuild the index, and fixup
rd.index->current/max and extra->hdr. If the extra->hdr change
like when an above the opened one, or the selected one, is deleted, the
pager is closed.

In case of MUTT_REOPENED we do nothing. Making rd.index->current/max
wrong and extra->hdr perhaps pointing to something that doesn't exist
anymore.

This change makes the MUTT_REOPENED behavior similar to MUTT_NEW_MAIL.

Note that we this the pager have good change to be closed with the
mailbox is modified externally, just like the current MUTT_NEW_MAIL
behavior.

Closes #796

### DIFF
--- a/pager.c
+++ b/pager.c
@@ -2231,7 +2231,10 @@ int mutt_pager(const char *banner, const char *fname, int flags, struct Pager *e
               break;
             }
           }
+        }
 
+        if ((check == MUTT_NEW_MAIL) || (check == MUTT_REOPENED))
+        {
           if (rd.index && Context)
           {
             /* After the mailbox has been updated,
@@ -2247,7 +2250,9 @@ int mutt_pager(const char *banner, const char *fname, int flags, struct Pager *e
             rd.index->max = Context->vcount;
 
             /* If these header pointers don't match, then our email may have
-             * been deleted.  Make the pointer safe, then leave the pager */
+             * been deleted.  Make the pointer safe, then leave the pager.
+             * This have a unpleasant behaviour to close the pager even the
+             * deleted message is not the opened one, but at least it's safe. */
             if (extra->hdr != Context->hdrs[Context->v2r[rd.index->current]])
             {
               extra->hdr = Context->hdrs[Context->v2r[rd.index->current]];


gahr:
This doesn't fix the crash for me, when trying to reproduce the issue.

sileht:
Did you get the same backtrace as reporter ?

Because since https://github.com/neomutt/neomutt/commit/4ed2b6f0f8e69b61c9e872ad9322bdc6f326ce5f
before we run any OP in pager, we check the mailbox status. So we already knows that the mailbox was modified at this point.

The code that decides if we should redraw the index or not was rebuilding it only on MUTT_NEW_MAIL. My change also does that on MUTT_REOPENED.

But maybe an other race is present in mh.c. My change fixes the issue only if mx_check_mailbox doesn't add/remove item in ctx->hdrs when it returns 0 or MUTT_FLAGS .

gahr:
I haven't verified this, but I conjecture that `mh_check_mailbox` doesn't notice that the message was deleted.

```
#0  maildir_mh_open_message (ctx=0x804a952a0, msg=0x804a433b0, msgno=0, is_maildir=1) at mh.c:1447                                                                   
#1  0x00000000004726a8 in maildir_open_message (ctx=0x804a952a0, msg=0x804a433b0, msgno=0) at mh.c:1466                                                              
#2  0x0000000000481bd9 in mx_open_message (ctx=0x804a952a0, msgno=0) at mx.c:1201                                                                                    
#3  0x000000000042ab23 in mutt_copy_message (fpout=0x803442f10, src=0x804a952a0, hdr=0x804b8f460, flags=0, chflags=16) at copy.c:766                                 
#4  0x00000000004202c5 in pipe_msg (h=0x804b8f460, fp=0x803442f10, decode=0, print=0) at commands.c:384                                                              
#5  0x000000000041dee1 in _mutt_pipe_message (h=0x804b8f460, cmd=0x7fffffff9c20 "echo", decode=0, print=0, split=0, sep=0x804a41c00 "\n") at commands.c:418          
#6  0x000000000041dd92 in mutt_pipe_message (h=0x804b8f460) at commands.c:506                                                                                        
#7  0x000000000049018f in mutt_pager (banner=0x0, fname=0x7fffffffb590 "/tmp/neomutt-ptrcrt-1001-87715-12363117053403874314", flags=66, extra=0x7fffffffb128) at page
r.c:2870                                                                                                                                                             
#8  0x000000000041d649 in mutt_display_message (cur=0x804b8f460) at commands.c:232                                                                                   
#9  0x0000000000434c82 in mutt_index_menu () at curs_main.c:2152                                                                                                     
#10 0x0000000000467582 in main (argc=1, argv=0x7fffffffe9e8, env=0x7fffffffea18) at main.c:936  
```

PR #798, opened 2017-09-28
bool: convert extern variable in mbyte.h

seishinryohosha:
@neomutt/reviewers 
next up is mbyte.h
Affected variable/prototypes
* Charset_is_utf8
* mutt_hcache_restore()
* mutt_set_charset()

Charset_is_utf8 affects `convert` in mutt_set_charset() and mutt_hcache_restore()
-> affects several static functions in hcache/hcache.c

* dump_address()
* dump_body()
* dump_buffer()
* dump_char()
* dump_char_size()
* dump_envelope()
* dump_parameter()
* dump_stailq()

* restore_address()
* restore_body()
* restore_buffer()
* restore_char()
* restore_envelope()
* restore_parameter()
* restore_stailq()

* **What does this PR do?**
Refactors mbyte.h to make use of `bool`
No
#774

### DIFF
--- a/hcache/hcache.c
+++ b/hcache/hcache.c
@@ -181,7 +181,7 @@ static inline bool is_ascii(const char *p, size_t len)
 }
 
 static unsigned char *dump_char_size(char *c, unsigned char *d, int *off,
-                                     ssize_t size, int convert)
+                                     ssize_t size, bool convert)
 {
   char *p = c;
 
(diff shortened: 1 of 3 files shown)


seishinryohosha:
> I would not reflow the function signatures.. they exceed the 80-chars limit just by a small amount, I think they're prettier in the current form :)

I see the char-limit as follows:
* Soft limit: 80
* Hard limit: 90

I reflowed them, as some exceeded 90 chars. (E.g. dump_address())
As they have similar signatures, I also reflowed the 80 < width < 90 ones.
Other than that, I agree with you.

flatcap:
> Soft limit: 80
> Hard limit: 90

Speficically in the [`.clang-format`](https://github.com/neomutt/neomutt/blob/master/.clang-format):
```
ColumnLimit: 80
# Allow some slightly over-long lines
PenaltyExcessCharacter: 1
```

I try to keep all the `.c` files tidy, so after making your changes it should be safe to:
```
find . -name '*.c' -exec clang-format -i {} +
```

flatcap:
Ah, I see what you mean about wrapping the prototypes.
Unfortunately, those changes will be lost when I next run clang-format.  Sorry.
(I try to keep the source tidy as we go, but I always run clang-format just before a release).

I like whitespace, so I'd love to make changes like:
```diff
-d = dump_char(e->supersedes, d, off, false);
-d = dump_char(e->date, d, off, false);
+d = dump_char(e->supersedes, d, off, false);
+d = dump_char(e->date,       d, off, false);
```

but no formatting tool is smart enough to be able to recreate all the patterns we like as humans.
With so much source, we quickly realised that we'd have to pick a tool and use its output without exceptions.  clang-format won and did a remarkably good job on Mutt's source.

PR #809, opened 2017-10-01
Fix memory leaks in pgpewrap

jelly:
Although the wrapper probably won't cause a real memory leak, cleaning up all allocated memory when exiting could save time spend on debugging memory leaks. Apart from fixing a leak this also leads a small refactoring of the duplicate argc check.


flatcap:
> check of argc to be `argc < 2`

done

PR #812, opened 2017-10-02
url: make notmuch query string parser generic

sileht:
notmuch: remove custom UriTag linked list
url: make notmuch query string parser generic

### DIFF
--- a/imap/util.c
+++ b/imap/util.c
@@ -325,6 +325,7 @@ int imap_parse_path(const char *path, struct ImapMbox *mx)
   {
     if (mutt_account_fromurl(&mx->account, &url) < 0 || !*mx->account.host)
     {
+      url_free(&url);
       FREE(&c);
       return -1;
     }
@@ -334,11 +335,13 @@ int imap_parse_path(const char *path, struct ImapMbox *mx)
     if (url.scheme == U_IMAPS)
       mx->account.flags |= MUTT_ACCT_SSL;
 
+    url_free(&url);
     FREE(&c);
   }
   /* old PINE-compatibility code */
   else
   {
+    url_free(&url);
     FREE(&c);
     if (sscanf(path, "{%127[^}]}", tmp) != 1)
       return -1;
--- a/mutt_notmuch.c
+++ b/mutt_notmuch.c
@@ -211,7 +211,7 @@ static void free_ctxdata(struct NmCtxData *data)
 #endif
   data->db = NULL;
 
-  url_qs_free(&data->db_url.query_strings);
+  url_free(&data->db_url);
   FREE(&data->db_url_holder);
   FREE(&data->db_query);
   FREE(&data);
@@ -238,7 +238,7 @@ static struct NmCtxData *new_ctxdata(const char *uri)
   data->db_limit = NmDbLimit;
   data->db_url_holder = safe_strdup(uri);
 
-  if (url_parse_with_qs(&data->db_url, data->db_url_holder) < 0)
+  if (url_parse(&data->db_url, data->db_url_holder) < 0)
   {
     mutt_error(_("failed to parse notmuch uri: %s"), uri);
     FREE(&data);
@@ -1722,7 +1722,7 @@ bool nm_normalize_uri(char *new_uri, const char *orig_uri, size_t new_uri_sz)
 
   rc = 0;
 gone:
-  url_qs_free(&tmp_ctxdata->db_url.query_strings);
+  url_free(&tmp_ctxdata->db_url);
   FREE(&tmp_ctxdata->db_url_holder);
   FREE(&tmp_ctxdata);
   if (rc < 0)
@@ -1913,7 +1913,7 @@ int nm_nonctx_get_count(char *path, int *all, int *new)
 
   memset(&url, 0, sizeof(struct Url));
 
-  if (url_parse_with_qs(&url, url_holder) < 0)
+  if (url_parse(&url, url_holder) < 0)
   {
     mutt_error(_("failed to parse notmuch uri: %s"), path);
     goto done;
@@ -1976,7 +1976,7 @@ done:
 #endif
     mutt_debug(1, "nm: count close DB\n");
   }
-  url_qs_free(&url.query_strings);
+  url_free(&url);
   FREE(&url_holder);
 
   mutt_debug(1, "nm: count done [rc=%d]\n", rc);
--- a/mutt_notmuch.c
+++ b/mutt_notmuch.c
@@ -1714,8 +1714,6 @@ bool nm_normalize_uri(char *new_uri, const char *orig_uri, size_t new_uri_sz)
   if (nm_uri_from_query(&tmp_ctx, buf, sizeof(buf)) == NULL)
     goto gone;
 
-  FREE(&tmp_ctxdata);
-
   strncpy(new_uri, buf, new_uri_sz);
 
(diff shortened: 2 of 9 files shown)


mutt_notmuch.c
```diff
@@ -370,9 +253,9 @@ static void free_ctxdata(struct NmCtxData *data)
 #endif
   data->db = NULL;
 
-  FREE(&data->db_filename);
+  STAILQ_INIT(&data->db_url.query_strings);
```
gahr: Are you perhaps forgetting to free the members of the linked list here?

gahr:
In general, I think you’re assuming that `STAILQ_INIT` also iterates the nodes and frees them, which it doesn’t do. I just had a quick look at this on my phone, so I might have overlooked some bits.

sileht:
url.c parser is a bit special, it doesn't allocate any data, it replaces all / ? & : @ by \0. Then all char* in Url and UrlQueryString are just pointers to the initial char* passed to url_parse().

So when you free the initial char* passed to url_parse(), all pointers in Url and UrlQueryString struct becomes obsolete, we just have to free the Url and UrlQueryString struct and that is. That why I don't free member of the linked list.

gahr:
The first part I do get.

I probably missed the spot in the code where you free the UrlQueryString nodes that make up the linked list.

sileht:
Oh I just got what I missed I will fix that.

sileht:
I have added url_qs_free() to clean the linked list nodes.

gahr:
Yep, that part :) I’ll reivew it as a whole as soon as I get a larger screen... looks promising already, thanks!

url.c
```diff
@@ -96,20 +96,72 @@ enum UrlScheme url_check_scheme(const char *s)
...
+    qs = safe_calloc(1, sizeof(struct UrlQueryString));
+    if ((k = strrchr(src, '&')))
+      *k = '\0';
+
+    if ((v = strchr(src, '=')))
+    {
+      *v = '\0';
+      qs->value = v + 1;
+      if (url_pct_decode(qs->value) < 0)
+        return -1;
```
gahr: You are leaking `qs` here....

url.c
```diff
@@ -96,20 +96,72 @@ enum UrlScheme url_check_scheme(const char *s)
...
+    }
+    qs->name = src;
+    if (url_pct_decode(qs->name) < 0)
+      return -1;
```
gahr: ... and here.

url.c
```diff
@@ -96,20 +96,72 @@ enum UrlScheme url_check_scheme(const char *s)
...
+
+int _parse_query_string(struct Url *u, char *src)
+{
+  struct UrlQueryString *qs = NULL;
+  char *k = NULL, *v = NULL;
+
+  while (src && *src)
+  {
+    qs = safe_calloc(1, sizeof(struct UrlQueryString));
+    if ((k = strrchr(src, '&')))
```
gahr: Why `strrchr` instead of `strchr` here?

url.c
```diff
@@ -96,20 +96,72 @@ enum UrlScheme url_check_scheme(const char *s)
     return (enum UrlScheme) i;
 }
 
+
+int _parse_query_string(struct Url *u, char *src)
```
gahr: Should this be static?

gahr:
I'm not especially fond of prepending static function with an underscore... 

I also don't especially like having both `url_parse` and `url_parse_with_qs`. I expect the query string is empty in most use cases (all but notmuch actually), so I'd just always parse it. URLs are short enough and performance shouldn't be an issue here. I would just modify `url_parse` to always try to parse a query string.

url.c
sileht: Worse typo ever....

url.c
sileht: Done

url.c
sileht: Done

url.c
sileht: Done, and I also freed the query_strings when in url_parse() return -1.

sileht:
I daily do python and often use underscore. But that doesn't make sense here, I removed them.

I made url_parse and url_parse_with_qs, just because I was lazy the day I wrote this to find where should I free the query_strings everywhere url_parse() is called. But I will cleanup that.

url.c
```diff
@@ -96,20 +96,76 @@ enum UrlScheme url_check_scheme(const char *s)
...
+ * @retval 0       if src is valid
+ * @retval -1      if src is invalid
+ *
+ * char* elements are pointers into src, which is modified by this call
+ * (duplicate it first if you need to).
+ *
+ * This method doesn't allocated any additional char* of the Url and
+ * UrlQueryString structs.
+ *
+ * To free Url, caller must free "src" and call url_free(&u);
```
gahr: Perhaps `call url_free(u)` since `u` is already a pointer. It's a minor thingy, but this works a bit different than most other free function in NeoMutt, where you pass in a pointer-to-pointer and the free function also sets the pointer to NULL at the caller.

url.c
```diff
@@ -154,32 +220,45 @@ static int parse_userhost(struct Url *u, char *src)
...
 
-  return parse_userhost(u, tmp);
+void url_free(struct Url *u)
+{
+  struct UrlQueryString *np = STAILQ_FIRST(&u->query_strings), *next = NULL;
+  while (np)
+  {
+      next = STAILQ_NEXT(np, entries);
+      /* NOTE(sileht): We don't free members, they will be freed when
+       * the src char* passed to _url_parse() is freed */
```
gahr: passed to `url_parse()`...

imap/util.c
```diff
@@ -335,6 +336,7 @@ int imap_parse_path(const char *path, struct ImapMbox *mx)
     if (url.scheme == U_IMAPS)
       mx->account.flags |= MUTT_ACCT_SSL;
 
+    url_free(&url);
     FREE(&c);
   }
   /* old PINE-compatibility code */
```
gahr: yuou are leaking url in the `else` clause following. Admittedly, this is a non-problem, since I don't expect a query string in an imap uri. Anyway, it would be better to call `url_free` anyway...

sileht:
Hum github haven't detected the new head: 59e1e08  ...

sileht:
I force pushed c60fd2bbf1fa6b7fcbd8b9e8a50a188daf04aea2 and it doesn't see it...

sileht:
It's an ongoing github issue

flatcap:
I've rebased this and made a few tweaks:
- fix double free in nm_normalize_uri()
- set path to NULL in url_parse()
  If we're setting all the other members, we should set the path too
- tidy some doxygen docs

mutt_notmuch.c
```diff
@@ -611,13 +502,13 @@ static const char *get_db_filename(struct NmCtxData *data)
...
 
-  db_filename = data->db_filename ? data->db_filename : NmDefaultUri;
+  db_filename = data->db_url.path ? data->db_url.path : NmDefaultUri;
   if (!db_filename)
     db_filename = Folder;
   if (!db_filename)
     return NULL;
   if (strncmp(db_filename, "notmuch://", 10) == 0)
-    db_filename += 10;
+    db_filename += 9;
```
flatcap: I'm not sure about this change, though I couldn't work out what settings would actually get me there.
A url with a path in it would have '///' in it (three slashes), so
`notmuch:///home/user/nm` would become `//home/user/nm`

sileht: That's a leftover, it should be 10. I remember having changing this to exercice a bug in url_parse().

PR #821, opened 2017-10-06
Install doc into `neomutt`

nicolasbock:
The documentation is installed into the `mutt` subdirectory by default.
This directory should be renamed to `neomutt`.


flatcap:
Thanks.  Merged in b2034674

PR #823, opened 2017-10-07
fix actions on tagged messages

### LINKED ISSUE #697: Tagged actions on collapsed threads don't work
**Note:** This bug exists in upstream Mutt, too.

**Working example:**
- View mail in the index
- Emails are threaded: `set sort=threads`
- All threads are open (not collapsed)
- Tag one thread: `<tag-thread>`
- Perform an action, e.g. `<tag-prefix><flag-message>`

**Result:**
- All the emails in the thread are now flagged

**Broken example:**
- View mail in the index
- Emails are threaded: `set sort=threads`
- All threads are closed (collapsed)
- Tag one thread: `<tag-thread>`
  This correctly tags all the emails in the thread
- Perform an action, e.g. `<tag-prefix><flag-message>`

**Result:**
- Only the root email is flagged

Related to #687 
/cc @ahmedelgabri

flatcap:
Many actions aren't performed on all tagged messages if the messages are within a collapsed thread.

Fixes #697

### Overview

The `Context` has two lists of emails:

- `struct Header **hdrs`
  A list of all emails
- `int v2r []`
  A lookup table of **visible** emails (what's seen in the index)

An email may be invisible if it's in a collapsed thread,
or if it doesn't match a `limit` pattern.

### Bug

Actions aren't being applied to tagged, but collapsed, threads.
e.g.

- \<tag-thread\>\<collapse-thread\>\<tag-prefix\>\<flag-message\>
- Expected action: all the emails in the thread are flagged
- Actual action: only the parent email is flagged

A closer inspection showed the problem to be widespread.
The incorrect code was looping through the `v2r` table of visible emails looking
for those that were tagged.

### Tests

1. Threads open, Tag a thread, Perform an action
   Result: Action is performed on every email in the thread
1. Threads closed, Tag a thread, Perform an action
   Result: Action is performed on every email in the thread
1. Threads open, Tag everything, Limit to a single thread, Perform an action
   Result: Action is performed on every email in the thread, but nothing else

### Fix

My first attempt at a fix, changed the code to loop through `hdrs` looking for
tagged emails.  This suffered from a different problem -- it applied functions
to tagged emails that were outside the limit pattern.

Now, the checks have been factored out into a new function `bool message_is_tagged()`
If the message is tagged **and** within the current limit (if applicable).

### Results

There are thirty changes to twenty functions.
I suspect this broken code has been cut'n'pasted for a long time.

Functions/actions that acted on messages outside of the limit pattern:

- ci_bounce_message
- imap_copy_messages * 2
- mutt_compose_menu (attach-message)
- mutt_index_menu (modify-tags)
- _mutt_bounce_message

Functions/actions that didn't act on collapsed threads:

- crypt_extract_keys_from_messages
- make_reference_headers
- mutt_check_traditional_pgp
- mutt_edit_message
- mutt_index_menu (flag-message)
- mutt_index_menu (quasi-delete)
- mutt_index_menu (resend-message)
- mutt_index_menu (tag-tag)
- mutt_index_menu (toggle-new)
- mutt_label_message
- mutt_link_threads
- mutt_save_message
- mutt_save_message
- _mutt_pipe_message * 3
- _mutt_set_flag

### DIFF
--- a/curs_lib.c
+++ b/curs_lib.c
@@ -43,8 +43,10 @@
 #include <wchar.h>
 #include "lib/lib.h"
 #include "mutt.h"
+#include "context.h"
 #include "enter_state.h"
 #include "globals.h"
+#include "header.h"
 #include "mbyte.h"
 #include "mutt_curses.h"
 #include "mutt_menu.h"
@@ -1460,4 +1462,33 @@ int mutt_strwidth(const char *s)
     w += wcwidth(wc);
   }
   return w;
+}
+
+/**
+ * message_is_visible - Is a message in the index within limit
+ * @param ctx   Open mailbox
+ * @param index Message ID (index into `ctx->hdrs[]`
+ * @retval bool True if the message is within limit
+ *
+ * If no limit is in effect, all the messages are visible.
+ */
+bool message_is_visible(struct Context *ctx, int index)
+{
+  if (!ctx || !ctx->hdrs || (index >= ctx->msgcount))
+    return false;
+
+  return !ctx->pattern || ctx->hdrs[index]->limited;
+}
+
+/**
+ * message_is_tagged - Is a message in the index tagged (and within limit)
+ * @param ctx   Open mailbox
+ * @param index Message ID (index into `ctx->hdrs[]`
+ * @retval bool True if the message is both tagged and within limit
+ *
+ * If a limit is in effect, the message must be visible within it.
+ */
+bool message_is_tagged(struct Context *ctx, int index)
+{
+  return message_is_visible(ctx, index) && ctx->hdrs[index]->tagged;
 }
--- a/commands.c
+++ b/commands.c
@@ -425,27 +425,32 @@ static int _mutt_pipe_message(struct Header *h, char *cmd, int decode,
     unset_option(OPT_KEEP_QUIET);
   }
   else
-  { /* handle tagged messages */
-
+  {
+    /* handle tagged messages */
     if (WithCrypto && decode)
     {
       for (int i = 0; i < Context->msgcount; i++)
-        if (message_is_tagged(Context, i))
-        {
+      {
+        if (!message_is_tagged(Context, i))
+          continue;
+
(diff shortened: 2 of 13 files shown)


gahr:
Nice!

For the sake of reviewing, would it be easy for you to split this inversion of logic in a second commit?

```
if (!message_is_tagged(Context, i))
  continue;
```

flatcap:
OK, a bit fiddly, but done.
I've also tried a little experiment.

- 3ce081eb fix actions on tagged messages
  Make the changes
- 9f2fa0e4 de-indent if-clauses
  This just 'de-indents' the clauses that will changed in the next commit
- e6824654 refactor: invert logic
  This is now much clearer now that the indentation isn't bloating the diff

gahr:
Thank you! I’ll have a close look asap!

curs_lib.c
```diff
@@ -1461,3 +1463,44 @@ int mutt_strwidth(const char *s)
...
+    return false;
+
+  if (!ctx->hdrs[index]->tagged)
+    return false;
+
+  if (!ctx->pattern)
+    return ctx->hdrs[index]->tagged;
+
+  return ctx->hdrs[index]->tagged && ctx->hdrs[index]->limited;
+}
```
gahr: So this whole function could be
```c
return message_is_visible(ctx, index) && ctx->hdrs[index]->tagged;
```

curs_lib.c
```diff
@@ -1461,3 +1463,44 @@ int mutt_strwidth(const char *s)
...
+ */
+bool message_is_visible(struct Context *ctx, int index)
+{
+  if (!ctx || !ctx->hdrs || (index >= ctx->msgcount))
+    return false;
+
+  if (!ctx->pattern)
+    return true;
+
+  return ctx->hdrs[index]->limited;
```
gahr: how about grouping the two returns?
```
return !ctx->pattern || ctx->hdrs[index]->limited;
```

Issue #826, opened 2017-10-08
Error: Unable to find GPGME


riesebie:
Expected behavior:
configure.autobuild  runs through

Actual behavior:
configure.autobuild stops with
```
Checking for gpgme.h...not found
Error: Unable to find GPGME
```
config .log says:
```
/usr/include/gpgme.h:104:2: error: GPGME was compiled with _FILE_OFFSET_BITS = 64, please see the section "Largefile support (LFS)" in the GPGME manual.                                          
  #error GPGME was compiled with _FILE_OFFSET_BITS = 64, please see the section "Largefile support (LFS)" in the GPGME manual.
```

Steps to reproduce:
Build on an i386. 
Build on amd64 runs just fine.

Used program versions:
gpgme 1.8.0 1.9.0
neomutt 20171006
neomutt 20170912 builds fine

Operating System and its version:
Debian sid

riesebie:
I can confirm that devel/issue826 let me build both i386 and ppc32 successfully.
Thanks

PR #829, opened 2017-10-09
If set, use -D_FILE_OFFSET_BITS=64 when checking for GPGMe

gahr:
Issue #826

* **What does this PR do?**

GPGMe gets configured a bit differently depending on whether LFS needs the `_FILE_OFFSET_BITS=64` define. This commit reinstates `cc-check-lfs` that might define  `_FILE_OFFSET_BITS`, and if defined, adds it to `CFLAGS` when checking for GPGMe

### DIFF
--- a/auto.def
+++ b/auto.def
@@ -225,6 +225,7 @@ if {1} {
   cc-with [list -cflags [get-define CFLAGS]]
 
   # Large file support
+  cc-check-lfs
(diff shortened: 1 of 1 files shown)


riesebie:
I can confirm that the build works on both i386 and ppc32 plattforms. 
Thanks

gahr:
Thank you!

PR #846, opened 2017-10-12
fix crash in mutt_pattern_exec()

flatcap:
This was reported by summit_ on IRC, who provided the following backtrace:
```
Program terminated with signal SIGSEGV, Segmentation fault.
#0  0x0000558864263662 in mutt_pattern_exec (pat=pat@entry=0x55886494f6f0, flags=flags@entry=MUTT_MATCH_FULL_ADDRESS, ctx=ctx@entry=0x5588648eacb0,
    h=h@entry=0x5588649d5ec0, cache=cache@entry=0x7ffff5523340) at pattern.c:1689
1689              set_pattern_cache_value(cache_entry, match_user(pat->alladdr, h->env->from, NULL));
 -------------
(gdb) p h->env
$2 = (struct Envelope *) 0x0
```

Steps to reproduce:
- '#' (break-thread)
- '\<enter\>' (view-message)

Crash occurred in `mutt_pattern_exec()` when `~P` pattern was evaluated.
A backtrace showed that `h->env` was NULL.

Thanks to summit_ for the bug report.

### DIFF
--- a/pattern.c
+++ b/pattern.c
@@ -1618,20 +1618,32 @@ int mutt_pattern_exec(struct Pattern *pat, enum PatternExecFlag flags,
       return -1;
 #endif
     case MUTT_SENDER:
+      if (!h->env)
+        return 0;
       return (pat->not ^ match_adrlist(pat, flags & MUTT_MATCH_FULL_ADDRESS, 1,
                                        h->env->sender));
     case MUTT_FROM:
+      if (!h->env)
+        return 0;
       return (pat->not ^
               match_adrlist(pat, flags & MUTT_MATCH_FULL_ADDRESS, 1, h->env->from));
     case MUTT_TO:
+      if (!h->env)
+        return 0;
       return (pat->not ^
               match_adrlist(pat, flags & MUTT_MATCH_FULL_ADDRESS, 1, h->env->to));
     case MUTT_CC:
+      if (!h->env)
+        return 0;
       return (pat->not ^
               match_adrlist(pat, flags & MUTT_MATCH_FULL_ADDRESS, 1, h->env->cc));
     case MUTT_SUBJECT:
+      if (!h->env)
+        return 0;
       return (pat->not ^ (h->env->subject && patmatch(pat, h->env->subject) == 0));
     case MUTT_ID:
+      if (!h->env)
+        return 0;
       return (pat->not ^ (h->env->message_id && patmatch(pat, h->env->message_id) == 0));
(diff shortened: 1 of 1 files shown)


flatcap:
Update from summut_:

> thanks, but that hasn't fixed it  :/

```
Program terminated with signal SIGSEGV, Segmentation fault.
#0  mutt_pattern_exec (pat=pat@entry=0x565506a20750, flags=flags@entry=MUTT_MATCH_FULL_ADDRESS, ctx=ctx@entry=0x565506ab2a80, h=h@entry=0x5655079acd60, cache=cache@entry=0x7ffc1b6515e0) at pattern.c:1715
1715              set_pattern_cache_value(cache_entry, match_user(pat->alladdr, h->env->from, NULL));
(gdb) p pat->alladdr
$1 = false
(gdb) p h->env->from
Cannot access memory at address 0x49
(gdb) p h->env
$2 = (struct Envelope *) 0x41
(gdb) p *h
$3 = {security = 0, mime = false, flagged = false, tagged = false, deleted = true, purge = false, quasi_deleted = false, changed = true, attach_del = false, old = false, read = false, expired = false, superseded = false,
  replied = false, subject_changed = false, threaded = false, display_subject = false, recip_valid = false, active = false, trash = false, xlabel_changed = false, zhours = 0, zminutes = 0, zoccident = false, searched = false,
  matched = false, attach_valid = false, collapsed = false, limited = false, num_hidden = 94923199794608, recipient = -22496, pair = 22101, date_sent = 94923184554256, received = 94923184554256, offset = 33, lines = 112346032,
  index = 22101, msgno = 127585616, virtual = 22101, score = 0, env = 0x41, content = 0x1b0, path = 0x0, tree = 0x5655079ace10 "\260", thread = 0x5655079acf30, attach_total = -12976, refno = 22101, tags = {stqh_first = 0x565506b24820,
    stqh_last = 0x565506b24820}, data = 0x21, free_cb = 0x565506b25040, maildir_flags = 0x5655079acdb0 "\260\001"}
```

> i think one of the callers must have corruptable state or smthing
> in fact - following the stracktrace all the way back shows that h->env is always 0x41 so something isn't setting it appropriately
> i can also reproduce this with no config other than to connect to an imap server and it still segfaults in the same way
> i have to go now though - thanks for your time

flatcap:
**Note:**  This PR doesn't fix Summit's problem, but it's still worthwhile adding the checks.
Issue #874 has been opened to track the original problem.

Issue #849, opened 2017-10-13
Virtual mailbox's notmuch URI ignores limit= parameter


Austin-Ray:
* **bug reports**

    * Expected behavior
    Virtual mailboxes using notmuch queries should return the number of emails specified by the limit= parameter e.g. 1000 emails.

    * Actual behavior
    notmuch queries return all emails matching a query.

    * Steps to reproduce
      1. Setup virtual mailbox using notmuch's URI with limit= parameter
          * I am using: 
             > virtual-mailboxes "unified inbox"   "notmuch://?query=tag:inbox&limit=1000"
      2. Open virtual mailbox in neomutt
      3. Compare number of emails in mailbox to specified limit

    * Used program versions
      NeoMutt 20170912 (1.9.0)

    * Operating System and its version
      System: Linux 4.13.5-1-ARCH (x86_64)

Running neomutt with debug level 2 show this message in the log:

> [2017-10-12 23:35:06] Reading configuration file '/home/aray/.mutt/mailboxes'.
> [2017-10-12 23:35:06] nm_normalize_uri (notmuch://?query=tag:inbox&limit=1000)
> [2017-10-12 23:35:06] nm_normalize_uri #<span></span>1 () -> db_query: (null)
> [2017-10-12 23:35:06] nm: get_query_string(0)
> [2017-10-12 23:35:06] nm: query 'tag:inbox'
> [2017-10-12 23:35:06] nm_normalize_uri #<span></span>2 () -> db_query: tag:inbox
> [2017-10-12 23:35:06] nm_uri_from_query (tag:inbox)
> [2017-10-12 23:35:06] nm: db filename '/home/aray/.mail'
> [2017-10-12 23:35:06] nm: uri from query 'notmuch:///home/aray/.mail?type=messages&query=tag%3Ainbox'
> [2017-10-12 23:35:06] nm_normalize_uri #<span></span>3 (notmuch://?query=tag:inbox&limit=1000) -> notmuch:///home/aray/.mail?type=messages&query=tag%3Ainbox

Normalization seems to drop the limit= parameter.

Thanks for taking a look!

Austin-Ray:
I have narrowed down the cause of the issue to how nm_normalize_uri() functions.

nm_normalize_uri() creates a temporary Context and NmCtxData structure to hold the URI's state while it is manipulated.

First nm_normalize_uri() calls get_query_string() to parse the URI. It properly parses the limit parameter and stores the value in the temporary NmCtxData structure.
[...]

PR #856, opened 2017-10-14
Fix Issue #849

Austin-Ray:
During notmuch URI creation, the limit= parameter is never added once
parsed. This issue is rectified by checking if db_limit is equal to 0
(the default value) and appending limit= if db_limit != 0.

NeoMutt initializes db_limit with the value of nm_db_limit. If the
limit= parameter is included, db_limit is assigned the value. However,
db_limit is never used after reassignment, resulting in issue #849.

* **What does this PR do?**
Resolves the issue present in #849 where notmuch URI ignores limit= parameter

Please examine my changes with extra scrutiny.

N/A

#849

### DIFF
--- a/mutt_notmuch.c
+++ b/mutt_notmuch.c
@@ -1741,9 +1741,16 @@ char *nm_uri_from_query(struct Context *ctx, char *buf, size_t bufsz)
   int added;
 
   if (data)
-    added = snprintf(uri, sizeof(uri),
-                     "notmuch://%s?type=%s&query=", get_db_filename(data),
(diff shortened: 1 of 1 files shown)


Austin-Ray:
In my eagerness to resolve this issue, I have introduced a bug. I am closing this pull request until such time that the bug is fixed.

flatcap:
Thanks for the brief PR, @Austin-Ray :-)
If you have any questions / need help, just ask.
We're also in #neomutt on irc.freenode.net

PR #860, opened 2017-10-14
Fix issue #849

### LINKED ISSUE #849: Virtual mailbox's notmuch URI ignores limit= parameter

Austin-Ray:
nm_uri_from_query() checks if there is a custom limit and includes it
in the URI. This keeps the limit intact after nm_normalize_uri()
terminates. Otherwise, the limit is stored within nm_normalize_uri()'s
temporary NmCtxData struct and lost upon termination.

Additionally, nm_normalize_uri() properly initializes tmp_ctxdata, which
sets the default db_limit value. This ensures db_limit is either
NmDbLimit or the custom limit. My earlier attempt to resolve #849 did
not initialize db_limit and suffered from random db_limit values.

The url_parse_query() check was removed from nm_normalize_uri() as
new_ctxdata() contains the same block of code. Error message numbering
has been updated the reflect the change.

* **What does this PR do?**
Fixes issue #849
I am unsure if I need to free the NmCtxData struct in nm_normalize_uri(). I had tried using free_ctxdata(), but neomutt would segfault on start. 
N/A
#849

### DIFF
--- a/mutt_notmuch.c
+++ b/mutt_notmuch.c
@@ -1741,9 +1741,20 @@ char *nm_uri_from_query(struct Context *ctx, char *buf, size_t bufsz)
   int added;
 
   if (data)
-    added = snprintf(uri, sizeof(uri),
-                     "notmuch://%s?type=%s&query=", get_db_filename(data),
-                     query_type_to_string(data->query_type));
+  {
+    if (get_limit(data) != NmDbLimit)
+    {
+      added = snprintf(uri, sizeof(uri),
+                       "notmuch://%s?type=%s&limit=%d&query=", get_db_filename(data),
+                       query_type_to_string(data->query_type), get_limit(data));
+    }
+    else
+    {
+      added = snprintf(uri, sizeof(uri),
+                       "notmuch://%s?type=%s&query=", get_db_filename(data),
+                       query_type_to_string(data->query_type));
+    }
+  }
   else if (NmDefaultUri)
     added = snprintf(uri, sizeof(uri), "%s?type=%s&query=", NmDefaultUri,
                      query_type_to_string(string_to_query_type(NmQueryType)));
@@ -1792,23 +1803,22 @@ bool nm_normalize_uri(char *new_uri, const char *orig_uri, size_t new_uri_sz)
   struct Context tmp_ctx;
   struct NmCtxData tmp_ctxdata;
 
+  struct NmCtxData *tmp_data_from_uri = new_ctxdata(strdup(orig_uri));
+
+  if (!tmp_data_from_uri) 
+    return false;
+
+  tmp_ctxdata = *tmp_data_from_uri;
+
   tmp_ctx.magic = MUTT_NOTMUCH;
   tmp_ctx.data = &tmp_ctxdata;
-  tmp_ctxdata.db_query = NULL;
-
-  if (!url_parse_query(orig_uri, &tmp_ctxdata.db_filename, &tmp_ctxdata.query_items))
-  {
-    mutt_error(_("failed to parse notmuch uri: %s"), orig_uri);
-    mutt_debug(2, "nm_normalize_uri () -> error #1\n");
-    return false;
-  }
 
   mutt_debug(2, "nm_normalize_uri #1 () -> db_query: %s\n", tmp_ctxdata.db_query);
 
   if (get_query_string(&tmp_ctxdata, false) == NULL)
   {
     mutt_error(_("failed to parse notmuch uri: %s"), orig_uri);
-    mutt_debug(2, "nm_normalize_uri () -> error #2\n");
+    mutt_debug(2, "nm_normalize_uri () -> error #1\n");
     return false;
   }
 
@@ -1819,7 +1829,7 @@ bool nm_normalize_uri(char *new_uri, const char *orig_uri, size_t new_uri_sz)
   if (nm_uri_from_query(&tmp_ctx, buf, sizeof(buf)) == NULL)
   {
     mutt_error(_("failed to parse notmuch uri: %s"), orig_uri);
-    mutt_debug(2, "nm_normalize_uri () -> error #3\n");
+    mutt_debug(2, "nm_normalize_uri () -> error #2\n");
     return true;
   }
 
--- a/mutt_notmuch.c
+++ b/mutt_notmuch.c
@@ -1801,6 +1801,7 @@ bool nm_normalize_uri(char *new_uri, const char *orig_uri, size_t new_uri_sz)
   char buf[LONG_STRING];
 
   struct Context tmp_ctx;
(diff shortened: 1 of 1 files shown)


mutt_notmuch.c
```diff
@@ -1792,23 +1799,22 @@ bool nm_normalize_uri(char *new_uri, const char *orig_uri, size_t new_uri_sz)
   struct Context tmp_ctx;
   struct NmCtxData tmp_ctxdata;
 
-  tmp_ctx.magic = MUTT_NOTMUCH;
-  tmp_ctx.data = &tmp_ctxdata;
-  tmp_ctxdata.db_query = NULL;
+  struct NmCtxData *tmp_data_from_uri = new_ctxdata(strdup(orig_uri));
```
flatcap: I can't see the need for this `strdup()`
Inside `new_ctxdata()` there are calls to `mutt_substrdup()` to keep the bits we want to keep.

Austin-Ray: Without the `strdup()` I receive a discarded qualifier warning during compilation. Would casting it to `char *` be a better alternative?

gahr: Not a good one :) How about changing `new_ctxdata` to take a `const char*`?

gahr:
How about not using an additional temporary variable for the result of `new_ctxdata(orig_uri);` and instead assign it directly to `tmp_ctx.data`?

gahr:
Here's the diff I propose against your branch:

```diff
diff --git a/mutt_notmuch.c b/mutt_notmuch.c
index daf21c0a..061fe596 100644
--- a/mutt_notmuch.c
+++ b/mutt_notmuch.c
@@ -334,7 +334,7 @@ static void free_ctxdata(struct NmCtxData *data)
  * A new nm_ctxdata struct is created, then the query is parsed and saved
  * within it.  This should be freed using free_ctxdata().
  */
-static struct NmCtxData *new_ctxdata(char *uri)
+static struct NmCtxData *new_ctxdata(const char *uri)
 {
   struct NmCtxData *data = NULL;

@@ -1797,38 +1797,38 @@ bool nm_normalize_uri(char *new_uri, const char *orig_uri, size_t new_uri_sz)
   char buf[LONG_STRING];

   struct Context tmp_ctx;
-  struct NmCtxData tmp_ctxdata;
+  struct NmCtxData *tmp_ctxdata = new_ctxdata(orig_uri);

-  struct NmCtxData *tmp_data_from_uri = new_ctxdata(strdup(orig_uri));
-
-  if (!tmp_data_from_uri) 
+  if (!tmp_ctxdata) 
     return false;

-  tmp_ctxdata = *tmp_data_from_uri;
-
   tmp_ctx.magic = MUTT_NOTMUCH;
-  tmp_ctx.data = &tmp_ctxdata;
+  tmp_ctx.data = tmp_ctxdata;

-  mutt_debug(2, "nm_normalize_uri #1 () -> db_query: %s\n", tmp_ctxdata.db_query);
+  mutt_debug(2, "nm_normalize_uri #1 () -> db_query: %s\n", tmp_ctxdata->db_query);

-  if (get_query_string(&tmp_ctxdata, false) == NULL)
+  if (get_query_string(tmp_ctxdata, false) == NULL)
   {
     mutt_error(_("failed to parse notmuch uri: %s"), orig_uri);
     mutt_debug(2, "nm_normalize_uri () -> error #1\n");
+    FREE(&tmp_ctxdata);
     return false;
   }

-  mutt_debug(2, "nm_normalize_uri #2 () -> db_query: %s\n", tmp_ctxdata.db_query);
+  mutt_debug(2, "nm_normalize_uri #2 () -> db_query: %s\n", tmp_ctxdata->db_query);

-  strfcpy(buf, tmp_ctxdata.db_query, sizeof(buf));
+  strfcpy(buf, tmp_ctxdata->db_query, sizeof(buf));

   if (nm_uri_from_query(&tmp_ctx, buf, sizeof(buf)) == NULL)
   {
     mutt_error(_("failed to parse notmuch uri: %s"), orig_uri);
     mutt_debug(2, "nm_normalize_uri () -> error #2\n");
+    FREE(&tmp_ctxdata);
     return true;
   }

+  FREE(&tmp_ctxdata);
+
   strncpy(new_uri, buf, new_uri_sz);

   mutt_debug(2, "nm_normalize_uri #3 (%s) -> %s\n", orig_uri, new_uri);
```

Austin-Ray:
I have updated my branch with the suggested changes.

Thank you for the constructive feedback! It is a very beneficial learning experience for me. My skill level with C is novice at best.

flatcap:
I've rebased your changes and added [a `memset()`](https://github.com/Austin-Ray/neomutt/blob/96997dc163bce218c65a1a3222f83d7b22e38dec/mutt_notmuch.c#L1804).

```c
struct Context tmp_ctx;
memset(&tmp_ctx, 0, sizeof(tmp_ctx));
```

We were calling `nm_uri_from_query()` with a mostly uninitialised `struct Context`.
If someone changed this function to use other members, it might crash.

Issue #877, opened 2017-10-19
does <save> for IMAP really need to display User/Pass?


lrosenman:
* **bug reports**

    * Expected behavior
        &lt;save&gt; shows save to imap://server.do.main/path/to/folder
    * Actual behavior
        &lt;save&gt; shows Append messages to <br>imap://&lt;user&gt;:&lt;pass&gt;@imap.lerctr.org/lists/noaa/changes? ([yes]/no):
    * Steps to reproduce
         save a message to a folder with folder=imap URL.
    * Used program versions
NeoMutt 20171013-25-e05db1
    * Operating System and its version
macOS High Sierra (10.13)

I'd like to see us suppress the username/password (or at least the password) on the Append message.

gahr:
Are you using the user and pass as part of your `folder` or `spoolfile` settings? Does this happen with user and pass in `imap_user` and `imap_pass` too?

lrosenman:
It's part of imap_folder, because I have multiple imap accounts, and swap between them:
```set folder=imap://ler@lerctr.org:$my_pass_lerctr@imap.lerctr.org
set smtp_url = "smtp://ler@lerctr.org:$my_pass_lerctr@smtp.lerctr.org:587"
set record="=SENT"
set imap_check_subscribed=yes
set imap_idle=yes
set from = "ler@lerctr.org"
set realname = "Larry Rosenman"
set spoolfile = "=INBOX
set postponed = "=Drafts"

mailboxes =FreeBSD/* =lists/noaa/* =lists/waze/*

subscribe freebsd-.*@freebsd.org  pgsql-.*@postgresql.org mailman-.*@python.org svn-src-all@freebsd.org
subscribe ports-committers@freebsd.org commits@spamassassin.apache.org developers@freebsd.org
subscribe no-reply@waze.com Melod.*@noaa.gov

account-hook imap://ler@lerctr.org@imap.lerctr.org \
      "set imap_headers='X-LERCTR-SPAM-FLAG X-LERCTR-SPAM-REPORT X-LERCTR-SPAM-SCORE '"
spam "X-LERCTR-Spam-Flag: YES" "SA"
spam "X-LERCTR-Spam-Score: ([0-9.-]*)" "%1"
```

is one of my identity switch files.

and the macros I use to invoke them:
```#Default#
set status_format="-%r-ler@lerctr.org: %f [Msgs:%?M?%M/?%m%?n? \
[...]

PR #886, opened 2017-10-21
call the folder-hook before saving to $record

flatcap:
cli neomutt asks for a password when sending because the `folder-hook` isn't executed for the `$record` mailbox.

This problem was discovered in #877.
Thanks to @lrosenman for his help debugging.

Setup:
- IMAP server
- 'sent' mailbox
- Password that's set using a folder-hook

Here's a minimal config.
```
set smtp_url="smtp://example.com"
set folder="imaps://USER@example.org"
set record="=SENT"
set imap_check_subscribed=yes
set from = "flatcap"
set realname = "Richard Russon"
set spoolfile = "=INBOX"
folder-hook "imaps://USER@example.com" set imap_pass="PASSWORD"
```

Test, by sending an email from the command line:

```
neomutt -n -F test.rc user@example.com
```

Actions:
- Follow the prompts to the compose email
- Hit <kbd>y</kbd> to send the mail

Result:
- A prompt for the password

Expected behaviour:
- `folder-hook` is run
- No prompt for a password
- Email saved to `$record` folder

### DIFF
--- a/hook.c
+++ b/hook.c
@@ -329,7 +329,7 @@ int mutt_parse_unhook(struct Buffer *buf, struct Buffer *s, unsigned long data,
   return 0;
 }
 
-void mutt_folder_hook(char *path)
+void mutt_folder_hook(const char *path)
 {
   struct Hook *tmp = NULL;
   struct Buffer err, token;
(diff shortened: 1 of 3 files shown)


gahr:
Would it make sense to instead call `folder_hook` on entry in `mx_open_mailbox`? I haven’t check the other call points, but I intend to.

flatcap:
This fixes the immediate problem, but it wasn't a deep examination.

I was cautious about touching `mx_open_mailbox()`.
There's one call to `mutt_folder_hook()`, but 15 calls to `mx_open_mailbox()`

I didn't want to break the existing `folder-hook`s and
I didn't want to have a situation where the hook was called twice.

gahr:
Makes sense, let's fix the immediate problem.

Issue #844, opened 2017-10-12
Smart wrap problem


tcpaiva:
* **bug reports**

    * Expected behavior
When reading e-mails, lines should wrap between words if "smart_wrap" is set.

    * Actual behavior
Lines wrap in the middle of the word. 

    * Steps to reproduce
According to flatcap, who helped me to dig into the problem at #neomutt irc channel, lines starting with space do not "smart wrap".

    * Used program versions
neomutt (1.8.3)

    * Operating System and its version
Ubuntu 16.04.3

gahr:
The (intermediate fix)[https://github.com/neomutt/neomutt/commit/f5e472b13a98fb84173e058fed462d3109508bb0] was brought to completion. It shouldn't be hard to check whether there are other spaces in between index 0 and `Wrap`.

Austin-Ray:
Since no one has stepped up for the issue, I would like to try my hand at it.

flatcap:
@Austin-Ray 
Great.  Be warned though!
You may find you just need to adjust one line, by one character, and it works.
Or, you may go insane trying to figure it out :-)

I found that lines starting with whitespace didn't **word**-wrap.

Good:
```
This long line gets
wrapped
```

Leading space, bad:
```
␣However this line isn't so pret
ty
```

Austin-Ray:
The merged PR resolves this issue so I am closing the issue.

PR #887, opened 2017-10-21
Fix smart wrap in pager without breaking header

Austin-Ray:
This commit addresses the issue of smart wrap not working if a line
started with whitespace.

This behavior was a result commit f5e472b where smart wrap was disabled
on lines beginning with a space or tab to prevent headers from having
weird wrapping.

This commit addresses this issue by only applying smart wrapping if the
pager is displaying a message body. This allows smart_wrap to work
inside of a message and preserves what commit f5e472b addressed.

I'm not the greatest with bitwise operations, but I based my change off of what I saw in pager.c and pager.h. I'm sure I didn't mess up the bitwise operations, but not entirely confident.

Issue #844

### DIFF
--- a/init.h
+++ b/init.h
@@ -4319,7 +4319,7 @@ struct Option MuttVars[] = {
   ** When \fIset\fP, NeoMutt will weed headers when displaying, forwarding,
   ** printing, or replying to messages.
   */
-  { "wrap",             DT_NUMBER,  R_PAGER, UL &Wrap, 0 },
+  { "wrap",             DT_NUMBER,  R_PAGER_FLOW, UL &Wrap, 0 },
   /*
   ** .pp
   ** When set to a positive value, NeoMutt will wrap text at $$wrap characters.
--- a/pager.c
+++ b/pager.c
@@ -1562,8 +1562,8 @@ static int display_line(FILE *f, LOFF_T *last_pos, struct Line **line_info,
   {
     if (cnt < b_read)
     {
-      if (ch != -1 && buf[0] != ' ' && buf[0] != '\t' && buf[cnt] != ' ' &&
-          buf[cnt] != '\t' && buf[cnt] != '\n' && buf[cnt] != '\r')
+      /* Wrap if the line is not part of the header and buf[cnt] is not whitespace */
+      if (ch != -1 && !ISHEADER((*line_info)[n].type) && !ISSPACE(buf[cnt]))
       {
         buf_ptr = buf + ch;
         /* skip trailing blanks */


Austin-Ray:
Mentioning @neomutt/reviewers per flatcap's welcome email.

flatcap:
This fixes the lines that weren't wrapping because of leading whitespace,
but it can cause problems with the header.

I've found a test case that shows the problem.
Edit an email and give it a new header line, e.g.
(three strings of say 60 chars)
```
X-Label: VeryLongStringWithNoSpaces VeryLongStringWithNoSpaces VeryLongStringWithNoSpaces
```

Resize your window to be less than 120 chars wide.
When the is email prepared for display (before display_line() is called), the header line will be wrapped to:
```
X-Label: VeryLongStringWithNoSpaces
        VeryLongStringWithNoSpaces
        VeryLongStringWithNoSpaces
```

Now shrink the window.  NeoMutt from `[master]` will display something like:
```
X-Label:
nccyrnccyrnccyrnccyrnccyrnccyrnccyr
        nccyrnccyrnccyrnccyrnccyrnccyrncc
yr
        nccyrnccyrnccyrnccyrnccyrnccyrncc
yr
```

Ugly, but still together as a block.
Your branch will look something like:
```
X-Label:
nccyrnccyrnccyrnccyrnccyrnccyrnccyr

nccyrnccyrnccyrnccyrnccyrnccyrnccyr

nccyrnccyrnccyrnccyrnccyrnccyrnccyr
```

The leading whitespace has been left and the words wrapped, causing blank lines.
I think this is what's referred to in the [commit message](https://github.com/neomutt/neomutt/commit/f5e472b13a98fb84173e058fed462d3109508bb0):

> on a line by itself detaching the header value from the header name

flatcap:
I think I may have a solution, though.
In `display_line()`, we want to leave the `buf[0]` checks when we're displaying a header line.
The caller can work out whether it's a header line or not.
Add a `bool header = true;` to this loop:

https://github.com/neomutt/neomutt/blob/c06e730d18afc8d0229b056ee09a283b08dd7589/pager.c#L1925-L1937

When we find a line `"\n"`, then `header = false;`
We can pass this info to `display_line()` in the flags.

I also had a quick look at all the callers of `display_line()`
The only two that are displaying actual emails are: `mutt_display_message()` and `mutt_view_attachment()`
It might be necessary to alter both.

pager

I hope this make sense.  It's quite late, now.

Austin-Ray:
@flatcap I had trouble implementing your solution due to lack of familiarity with the codebase (I got overwhelmed.) However, I discovered a simpler solution. 

I checked my most recent commit for detached headers using 10 different emails on terminals at 40, 80, and 120 characters wide. I was unable to find any. Could you verify that the headers aren't detached on your machine?

gahr:
Sorry, this doesn't solve the detached-header problem:

Header:
```
X-Spam-Symbols: ARC_NA,FROM_HAS_DN,FROM_EQ_ENVFROM,RCVD_COUNT_ONE,                                                                                                   
        HAS_REPLYTO,BAYES_HAM,MIME_GOOD,RCVD_TLS_ALL,PREVIOUSLY_DELIVERED,                                                                                           
        RCPT_COUNT_ONE,REPLYTO_EQ_FROM,TO_DN_NONE,TO_EQ_FROM 
```

`:set wrap=40`

* Original code

```
X-Spam-Symbols:                                                                                                                                                      
+ARC_NA,FROM_HAS_DN,FROM_EQ_ENVFROM,RCVD                                                                                                                             
+_COUNT_ONE,                                                                                                                                                         
        HAS_REPLYTO,BAYES_HAM,MIME_GOOD,                                                                                                                             
+RCVD_TLS_ALL,PREVIOUSLY_DELIVERED,                                                                                                                                  
        RCPT_COUNT_ONE,REPLYTO_EQ_FROM,T                                                                                                                             
+O_DN_NONE,TO_EQ_FROM 
```

* Your patch

```
X-Spam-Symbols:                                                                                                                                                      
+ARC_NA,FROM_HAS_DN,FROM_EQ_ENVFROM,RCVD                                                                                                                             
+_COUNT_ONE,                                                                                                                                                         
                                                                                                                                                                     
+HAS_REPLYTO,BAYES_HAM,MIME_GOOD,RCVD_TL                                                                                                                             
+S_ALL,PREVIOUSLY_DELIVERED,                                                                                                                                         
                                                                                                                                                                     
+RCPT_COUNT_ONE,REPLYTO_EQ_FROM,TO_DN_NO                                                                                                                             
+NE,TO_EQ_FROM  
```

Issue #888, opened 2017-10-22
Minor infelicity in logging mailbox flags parsing


lrosenman:
* **bug reports**

    * Expected behavior
       each flag on it's own line with time stamp.
    * Actual behavior
       bunched up mess
    * Steps to reproduce 
       mutt -d5
       select a mailbox with a bunch of PERMANENTFLAGS
    * Used program versions
       NeoMutt 20171013-32-53e6d4
    * Operating System and its version
       macOS High Sierra (10.13)

see https://pastebin.com/CGDc50cM for an example output

PR #889, opened 2017-10-22
Issue 888: Fix imap mailbox flag logging

lrosenman:
* **What does this PR do?**
   changes the imap mailbox flags logging to be more readable

   no

#888

### DIFF
--- a/imap/imap.c
+++ b/imap/imap.c
@@ -175,7 +175,8 @@ int imap_delete_mailbox(struct Context *ctx, struct ImapMbox *mx)
 
   if (!ctx || !ctx->data)
   {
-    idata = imap_conn_find(&mx->account, option(OPT_IMAP_PASSIVE) ? MUTT_IMAP_CONN_NONEW : 0);
+    idata = imap_conn_find(&mx->account,
+                           option(OPT_IMAP_PASSIVE) ? MUTT_IMAP_CONN_NONEW : 0);
     if (!idata)
     {
       FREE(&mx->mbox);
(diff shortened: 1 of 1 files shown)


gahr:
How about storing all the flags in a nocely formatted temporary buffer and log them at once? The disadvantage would be an additional memory allocation, unless we want to chech for the debug level.

flatcap:
One possibility is to use `struct Buffer` and `mutt_buffer_printf()` to generate the full string.

gahr:
Yep that’s what I had in mind. Also, it should be possible to use a buffer on the stack to avoid the additional memory allocation.

lrosenman:
I guess one question is how big to make the buffer since on a server that allows random flags (dovecot, et al), it could be a LONG list, see my pastebin.

lrosenman:
And, it looks like it might not be that bad (and Buffer seems like it does the right thing to auto-expand).  I'm working on a patch :) which I will add to this PR.

gahr:
I would personally prefer to keep a single log entry on a single line, no matter how long that might be. Would help with grepping etc.

gahr:
The code looks good, thanks!

However, I test-ran it and saw how this formatting is different from the one specified by the IMAP protocol as a return from the `FLAGS` command or as untagged responses, e.g., when opening a mailbox.

```
6< * FLAGS (\Answered \Flagged \Deleted \Seen \Draft Old $Forwarded $MDNSent $NotJunk NonJunk $Junk Junk Unread)
...
6< * OK [PERMANENTFLAGS (\Answered \Flagged \Deleted \Seen \Draft Old $Forwarded $MDNSent $NotJunk NonJunk $Junk Junk Unread \*)] Flags permitted.
...
Mailbox flags: [\Answered] [\Flagged] [\Deleted] [\Seen] [\Draft] [Old] [$Forwarded] [$MDNSent] [$NotJunk] [NonJunk] [$Junk] [Junk] [\*]
```
So, I have two questions:

1. Do we still want to print that log entry, given that the response was logged just above anyway?
2. If we do, do we want to wrap each flag inside square brackets? If we do, why?

lrosenman:
The original code listed each flag with square brackets and as a separate log message (but without a new line, see the pastebin in #888). 

I think we do want to list the flags that we noticed as confirmation.

PR #917, opened 2017-11-05
Respect --with-ssl path

### LINKED ISSUE #916: Fix usage of --with-openssl
The path specified with `--with-openssl=/path`is only used by the linker during configure time.

gahr:
Fixes #916


flatcap:
I've just tweaked the commit message `s/--with-openssl/--with-ssl/` to save confusion.

gahr:
Yep, thanks!

Issue #897, opened 2017-10-26
consider removing hard-coded color commands from pgp tags; and when a message-hook filter fails, mutt should display the unfiltered content


libBletchley:
I have this config line, which worked fine with older versions of mutt:
```
message-hook '~f .*@hushmail.com'  "set display_filter='dos2unix'"
```
The purpose of that line is to deal with the fact that we see a caret-M ("`^M`") at the end of each line if receiving an encrypted message that was sent from a hushmail server.

   * Expected behavior
The message should display correctly without being junked up with `^M`s.  And even if `dos2unix` (or any other filter tool) were to have a non-zero exit code, I would expect mutt to display the unfiltered content.  In fact this worked as expected in the previous version of mutt that was distributed with Debian Jessie.

    * Actual behavior
Now after entering the PGP password to decrypt the message, this blob appears on the screen, and the plaintext is no longer shown:
```
PGP message successfully decrypted.dos2unix: Binary symbol 0x1B found at line 8                                                             
                                                                               dos2unix: Skipping binary file stdin
                                                                                                                   Press any key to continue...
```
The knee-jerk thought is that the bug is in dos2unix, not mutt.  So to test that theory, I saved the pgp attachment to a file, and ran the same command mutt is configured to run: `gpg --decrypt --passphrase-fd 0 --pinentry-mode loopback --no-verbose --quiet  --batch  --output - message.html.pgp | dos2unix` and it worked fine (no error).  So it seems mutt is now introducing a "Binary symbol 0x1B" somehow.

   * Steps to reproduce
1. put your public key on the hushmail keyring at hushtools.com, and make sure the key contains your external email address
2. have a hushmail user send you an encrypted email to the address on the key you just submitted
3. add this to your configs `message-hook '~f .*@hushmail.com'  "set display_filter='dos2unix'"`
4. when the encrypted message arrives, run it through this sed command before opening it:
```
sed -e 's@^Content-Type: text/plain.*@Content-Type: application/pgp; format=text; x-action=encrypt@'
```
5. Try to open it in mutt.

Note about step 4: it may not be necessary to do step 4.  That sed command is what my procmail recipe runs on all inbound hushmail msgs to make mutt treat them as encrypted messages.  It's possible that over the years hushmail has fixed that (in which case the sed command would have no impact).. I have no idea ATM.

   * Used program versions
failing version: NeoMutt 20170113 (the current Debian Stretch version)
working version: whatever version of mutt is distributed with debian jessie

[...]

PR #924, opened 2017-11-07
display_filter: Don't output escape sequences

flatcap:
When NeoMutt opens a signed/encrypted message, it adds some lines to the
display, e.g.

    [-- PGP output follows ... --]

In order to be able to colour them, it adds a marker to the beginning of the line -- `<esc>]9;XXX<bel>` -- where "XXX" is a random 64-bit number.
This marker means that other strings in the body of the email will never match.

Unfortunately, this escape sequence will be visible to the `display_filter` command.

issue #897

---

Normally, NeoMutt processes the email, writing it to a temporary file.  That file is displayed in the pager.
When `$display_filter` is set, it invisibly inserts the filter into the processing.

    Process email -> filter -> temp file -> pager

https://github.com/neomutt/neomutt/blob/3a6f8fbc8de5d309c2ea556817889714d2e8c0f6/init.c#L4075-L4076

https://github.com/neomutt/neomutt/blob/3a6f8fbc8de5d309c2ea556817889714d2e8c0f6/pager.c#L828-L829

This PR adds a new flag to the `struct State` which gets passed down through many code layers.
The backtrace looks something like:
```
#1 mutt_index_menu()
#2 mutt_display_message()
#3 mutt_copy_message()
#4 _mutt_copy_message()
#5 mutt_body_handler()
#6 handler = mutt_signed_handler
#7 run_decode_and_handler()
#8 handler() -> mutt_signed_handler()
.. many pgp functions
#11 state_mark_attach()
#12 state_puts(AttachmentMarker, s);
```

### DIFF
--- a/commands.c
+++ b/commands.c
@@ -146,6 +146,7 @@ int mutt_display_message(struct Header *cur)
       unlink(tempfile);
       return 0;
     }
+    cmflags |= MUTT_CM_DISPLAY_FILTER;
   }
 
   if (!Pager || (mutt_strcmp(Pager, "builtin") == 0))
--- a/copy.c
+++ b/copy.c
@@ -655,6 +655,8 @@ int _mutt_copy_message(FILE *fpout, FILE *fpin, struct Header *hdr,
       s.prefix = prefix;
     if (flags & MUTT_CM_DISPLAY)
       s.flags |= MUTT_DISPLAY;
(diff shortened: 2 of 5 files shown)


flatcap:
The code has been archived to https://github.com/neomutt/neomutt-old/tree/display_filter

PR #929, opened 2017-11-07
Don't start function names with an underscore

flatcap:
- 9f540482f rename static functions
- 7e347662d rename public functions

One function has had its parameters reordered (to match another function).
Otherwise, there are no code changes.

##  Static Functions
Most of their names have been simplified by stripping leading components.

| Old Name                    | New Name              |
| :-------------------------- | :-------------------- |
| _alternates_clean           | alternates_clean      |
| _attachments_clean          | attachments_clean     |
| _crypt_compare_address      | compare_key_address   |
| _crypt_compare_date         | compare_key_date      |
| _crypt_compare_keyid        | compare_keyid         |
| _crypt_compare_trust        | compare_key_trust     |
| _handle_error               | handle_error          |
| _handle_panic               | handle_panic          |
| _lua_expose_command         | lua_expose_command    |
| _lua_init                   | lua_init              |
| _lua_mutt_call              | lua_mutt_call         |
| _lua_mutt_enter             | lua_mutt_enter        |
| _lua_mutt_error             | lua_mutt_error        |
| _lua_mutt_get               | lua_mutt_get          |
| _lua_mutt_message           | lua_mutt_message      |
| _lua_mutt_set               | lua_mutt_set          |
| _maildir_commit_message     | md_commit_message     |
| _maildir_open_find_message  | md_open_find_message  |
| _menu_status_line           | status_line           |
| _mh_commit_message          | mh_commit_msg         |
| _mutt_append_message        | append_message        |
| _mutt_bounce_message        | bounce_message        |
| _mutt_check_traditional_pgp | check_traditional_pgp |
| _mutt_fmt_smime_command     | fmt_smime_command     |
| _mutt_list_hook             | list_hook             |
| _mutt_parse_color           | parse_color           |
| _mutt_parse_uncolor         | parse_uncolor         |
| _mutt_pipe_message          | pipe_message          |
| _mutt_string_hook           | string_hook           |
| _pgp_compare_address        | compare_key_address   |
| _pgp_compare_date           | compare_key_date      |
| _pgp_compare_keyid          | compare_keyid         |
| _pgp_compare_trust          | compare_key_trust     |
| _pgp_free_key               | free_key              |
| _pgp_parent                 | key_parent            |
| _smime_getkeys              | getkeys               |

## Public Functions
Some have just had their leading `_` stripped.
The others have been renamed to distinguish them from similarly named functions.
e.g.  Suffix describing extra parameter.

| Old Name               | New Name                 |
| :--------------------- | :----------------------- |
| mutt_copy_message      | mutt_open_copy_message   |
| rfc2047_encode_string  | rfc2047_encode_string32  |
| _mutt_aside_thread     | mutt_aside_thread        |
| _mutt_copy_message     | mutt_copy_message        |
| _mutt_enter_fname      | mutt_enter_fname_full    |
| _mutt_enter_string     | mutt_enter_string_full   |
| _mutt_expand_path      | mutt_expand_path_regex   |
| _mutt_get_field        | mutt_get_field_full      |
| _mutt_make_string      | mutt_make_string_flags   |
| _mutt_mktemp           | mutt_mktemp_full         |
| _mutt_save_message     | mutt_save_message_ctx    |
| _mutt_select_file      | mutt_select_file         |
| _mutt_set_flag         | mutt_set_flag_update     |
| _mutt_traverse_thread  | mutt_traverse_thread     |
| _pgp_keyid             | pgp_this_keyid           |
| _rfc2047_encode_string | rfc2047_encode_string    |

I'm happy to change any of the names.  (naming is hard :-)

### DIFF
--- a/attach.c
+++ b/attach.c
@@ -795,7 +795,7 @@ int mutt_save_attachment(FILE *fp, struct Body *m, char *path, int flags, struct
       if (ctx.magic == MUTT_MBOX || ctx.magic == MUTT_MMDF)
         chflags = CH_FROM | CH_UPDATE_LEN;
       chflags |= (ctx.magic == MUTT_MAILDIR ? CH_NOSTATUS : CH_UPDATE);
-      if (mutt_copy_message(msg->fp, fp, hn, hn->content, 0, chflags) == 0 &&
+      if (mutt_copy_message_fp(msg->fp, fp, hn, 0, chflags) == 0 &&
           mx_commit_message(msg, &ctx) == 0)
         r = 0;
       else
--- a/attach.c
+++ b/attach.c
@@ -795,7 +795,7 @@ int mutt_save_attachment(FILE *fp, struct Body *m, char *path, int flags, struct
       if (ctx.magic == MUTT_MBOX || ctx.magic == MUTT_MMDF)
         chflags = CH_FROM | CH_UPDATE_LEN;
       chflags |= (ctx.magic == MUTT_MAILDIR ? CH_NOSTATUS : CH_UPDATE);
-      if (_mutt_copy_message(msg->fp, fp, hn, hn->content, 0, chflags) == 0 &&
+      if (mutt_copy_message(msg->fp, fp, hn, hn->content, 0, chflags) == 0 &&
           mx_commit_message(msg, &ctx) == 0)
         r = 0;
       else
--- a/browser.c
+++ b/browser.c
@@ -1066,7 +1066,7 @@ void mutt_browser_select_dir(char *f)
   mutt_get_parent_path(LastDir, OldLastDir, sizeof(LastDir));
 }
 
-void _mutt_select_file(char *f, size_t flen, int flags, char ***files, int *numfiles)
+void mutt_select_file(char *f, size_t flen, int flags, char ***files, int *numfiles)
 {
   char buf[_POSIX_PATH_MAX];
   char prefix[_POSIX_PATH_MAX] = "";
--- a/color.c
+++ b/color.c
@@ -484,14 +484,14 @@ static void do_uncolor(struct Buffer *buf, struct Buffer *s,
 }
 
 /**
- * _mutt_parse_uncolor - Parse an 'uncolor' command
+ * parse_uncolor - Parse an 'uncolor' command
  *
  * usage:
  * * uncolor index pattern [pattern...]
  * * unmono  index pattern [pattern...]
  */
-static int _mutt_parse_uncolor(struct Buffer *buf, struct Buffer *s, unsigned long data,
-                               struct Buffer *err, short parse_uncolor)
+static int parse_uncolor(struct Buffer *buf, struct Buffer *s, unsigned long data,
+                         struct Buffer *err, short parse_uncolor)
 {
   int object = 0, do_cache = 0;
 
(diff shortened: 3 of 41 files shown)


gahr:
* Perhaps leave `mutt_enter_string` unrenamed and rename `_mutt_enter_string` as `mutt_enter_string_full`, if we want to go with the `_full` suffix for extended calls (I’ve seen prior art of using `ex` and possibly `ex2` in other projects)

* maybe a per-module (i.e., .c and .h pair) prefix would still be nice for static functions. It would immediately show the scope and avoid unadvertently override a global name (did you know we have two `mutt_exit` functions ;) )

gahr:
Just throwing random ideas in...

How about
* mutt_open_copy_message -> mutt_copy_message_ctx
* mutt_copy_message -> mutt_copy_message_fp

gahr:
`rfc2047_encode_string32` is curious. I have followed the use of that `col` parameter down to after `try_block` is called somewhere in the depths of rfc2047 encoding, but I can’t seem to be able to figure out what its purpose is. Can anybody explain?

flatcap:
Rebased.

> Perhaps leave mutt_enter_string unrenamed ...

Done

> (did you know we have two mutt_exit functions ;) )

Yes, that's a artifact of the libraries.
The alloc functions need it and I wanted the libraries to have as few unresolved symbols as possible.

>     mutt_open_copy_message -> mutt_copy_message_ctx
>     mutt_copy_message -> mutt_copy_message_fp

I think not for this pair.
`mutt_open_copy_message()` calls `mx_open_message()` before delegating to `mutt_copy_message()`

> rfc2047_encode_string32 is curious
> Can anybody explain?

I traced though the code and I think I understand.
It's encoding various header fields and `col` is the length of the field name, e.g. "From".

The RFC2047 encoding produces blocks like: `=?ISO-8859-1?Q?a?=`
There's no limit to how many there are, but the header lines are limited to 76 chars.
The `col` parameter makes sure we don't overflow.  Usually it's given the exact length of the field name: `strlen("From")`.  In other places, it's just an arbitrary 32 chars.

[RFC2047, section 2](https://tools.ietf.org/html/rfc2047#section-2) says:
```
   An 'encoded-word' may not be more than 75 characters long, including
   'charset', 'encoding', 'encoded-text', and delimiters.  If it is
   desirable to encode more text than will fit in an 'encoded-word' of
   75 characters, multiple 'encoded-word's (separated by CRLF SPACE) may
   be used.

   While there is no limit to the length of a multiple-line header
   field, each line of a header field that contains one or more
   'encoded-word's is limited to 76 characters.
```

gahr:
>> mutt_open_copy_message -> mutt_copy_message_ctx
>> mutt_copy_message  -> mutt_copy_message_fp

> I think not for this pair. mutt_open_copy_message() calls mx_open_message() before delegating to mutt_copy_message()

The fact that it calls `mx_open_message()` is an implementation details that should the caller should not care or even be aware of. `mx_close_message()` is called before the function returns, so the net effect - as far as the caller is concerned - is *not* to open the message.

The real difference between these two functions is where they get the info about what to copy, either `FILE *fpin` or `struct Context *src`.

Also, `mutt_copy_message` would benefit from having the `body` argument remove, as it's always `hdr->content` anyways.

The result would be cleaner in my opinion:

```h
int mutt_copy_message_fp (FILE *fpout, FILE *fpin,          struct Header *hdr, int flags, int chflags);
int mutt_copy_message_ctx(FILE *fpout, struct Context *src, struct Header *hdr, int flags, int chflags);
```

I took the liberty to provide evidence in fedfa2b, which is based on devel/underscore. Feel free to merge it into your branch if you think my arguments make sense. I am not going to push any more than this, promise :)

flatcap:
Thanks, I've added your changes.

PR #941, opened 2017-11-09
A mixed bag of library changes

flatcap:
The first batch are simple.  These don't make any changes to the code.

- 290646b98 build: rename headers.c to header.c
  Match the `.c` to the `.h` for Header

- 01e064006 build: move *_free functions
  Move `free_X()` from `muttlib.c` to `X.c`

- c8a5e096d doxygen: add custom layout
  Tweak the docs layout

- 891a2728b lib: tidy headers
  Light tidying of headers

- 8cde8adb4 lib: move NLS magic to message.c
  The NLS gettext macros seem better suited in message.h

- 297560fbe lib: move function to file.c
  * `file_read_keyword()`
  * `mbox_check_empty()`
  * `to_absolute_path()`

- d9ad7cb7a lib: move functions to string.c
  * `find_word()`
  * `is_ascii()`
  * `next_word()`

- aabe469ab lib: move function to list.c
  * `strict_cmp_stailq()`

---

There's one slight code change in `mbyte.c`
The `replacement_char()` function depended on a widely used global variable: `Charset_is_utf8`.
It's been replaced with a global variable `ReplacementChar` that the library maintains.
When `Charset_is_utf8` is changed, `mutt_set_charset()` sets `ReplacementChar` appropriately.

- 50bb99279 lib: move functions to mbyte.c
  * `get_initials()`
  * `is_shell_char()`
  * `mutt_charlen()`
  * `my_mbstowcs()`
  * `my_wcstombs()`
  * `my_wcswidth()`
  * `my_wcwidth()`
  * `my_width()`
  * `width_ceiling()`

---

~This commit moves a load of functions into `lib/address.c`~
~It's not the ideal place for it.  Ideally, I'd like to see a [libemail](https://gist.github.com/flatcap/58a8ca42d30ef9d0cb54f189ce28b9fa).~

~For now, these functions have been cleaned up and documented.~
~Being in libmutt means they can be easily unit-tested.~

- ~89fa41b99 lib: move functions to `address.c`~
- ~62f570a21 lib: boolify two address functions~


body.h
```diff
@@ -101,5 +101,6 @@ struct Body
 struct Body *mutt_new_body(void);
 int mutt_copy_body(FILE *fp, struct Body **tgt, struct Body *src);
 void mutt_free_body(struct Body **p);
+int strict_cmp_bodies(const struct Body *b1, const struct Body *b2);
```
d-k-c: mutt_body_strict_cmp? mutt_strict_cmp_body? But please avoid plural, it makes it even harder to spot similar function

envelope.h
```diff
@@ -67,5 +67,6 @@ struct Envelope
 struct Envelope *mutt_new_envelope(void);
 void mutt_free_envelope(struct Envelope **p);
 void mutt_merge_envelopes(struct Envelope *base, struct Envelope **extra);
+int strict_cmp_envelopes(const struct Envelope *e1, const struct Envelope *e2);
```
d-k-c: mutt_strict_cmp_envelope? (same reasoning as before)

header.h
```diff
@@ -118,14 +118,8 @@ struct Header
...
-static inline struct Header *mutt_new_header(void)
-{
-  struct Header *h = safe_calloc(1, sizeof(struct Header));
-#ifdef MIXMASTER
-  STAILQ_INIT(&h->chain);
-#endif
-  STAILQ_INIT(&h->tags);
-  return h;
-}
+int mbox_strict_cmp_headers(const struct Header *h1, const struct Header *h2);
```
d-k-c: mutt_strict_cmp_header?

lib/file.h
```diff
@@ -59,5 +61,6 @@ int         safe_fsync_close(FILE **f);
 int         safe_open(const char *path, int flags);
 int         safe_rename(const char *src, const char *target);
 int         safe_symlink(const char *oldpath, const char *newpath);
+int         to_absolute_path(char *path, const char *reference);
```
d-k-c: If there's only one consumer of this function, the need to move it to file.c is discutable. But if you choose to do so, a name like file_abspath or mutt_abspath (the header seems schizophrenic on the prefix that should be used) would definitely be a plus.

lib/address.h
```diff
@@ -0,0 +1,77 @@
...
+
+/**
+ * struct Address - An email address
+ */
+struct Address
+{
+  char *personal; /**< real name of address */
+  char *mailbox;  /**< mailbox and host address */
+  int group;      /**< group mailbox? */
+  struct Address *next;
```
d-k-c: Definitely electable to a switch to TAILQ (this comment is more a note than an actual request for change)

lib/address.h
```diff
@@ -0,0 +1,77 @@
...
+void            rfc822_cat(char *buf, size_t buflen, const char *value, const char *specials);
+struct Address *rfc822_cpy_adrlist(struct Address *addr, bool prune);
+struct Address *rfc822_cpy_adr(struct Address *addr);
+void            rfc822_free_address(struct Address **p);
+struct Address *rfc822_new_address(void);
+struct Address *rfc822_parse_adrlist(struct Address *top, const char *s);
+void            rfc822_qualify(struct Address *addr, const char *host);
+int             rfc822_remove_from_adrlist(struct Address **a, const char *mailbox);
+bool            rfc822_valid_msgid(const char *msgid);
+bool            strict_addrcmp(const struct Address *a, const struct Address *b);
```
d-k-c: could this whole bunch of functions use a better prfix than rfc822, as it's mostly dealing with addresses? This is a bit out-of-scope of the "move functions to lib/address.c", but we should really use that opportunity for cleanup

lib/list.h
```diff
@@ -52,5 +52,6 @@ struct ListNode *mutt_list_insert_after(struct ListHead *h, struct ListNode *n,
 struct ListNode *mutt_list_insert_head(struct ListHead *h, char *s);
 struct ListNode *mutt_list_insert_tail(struct ListHead *h, char *s);
 bool             mutt_list_match(const char *s, struct ListHead *h);
+int              strict_cmp_stailq(const struct ListHead *ah, const struct ListHead *bh);
```
d-k-c: mutt_list_strict_cmp?

init.c
```diff
@@ -4275,7 +4220,7 @@ void mutt_init(int skip_sys_rc, struct ListHead *commands)
   /* Do we have a locale definition? */
   if (((p = getenv("LC_ALL")) != NULL && p[0]) || ((p = getenv("LANG")) != NULL && p[0]) ||
       ((p = getenv("LC_CTYPE")) != NULL && p[0]))
-    set_option(OPT_LOCALES);
+    OPT_LOCALES = true;
```
gahr: Why this change?

flatcap:
@d-k-c 

> I know this PR is mostly moving stuff,
> but I think it should also take care of uniformizing and renaming functions

I'd rather move, then rename.  It'll make it easier to keep track of functions that way.

> Also, I'm getting more and more skeptical to moving stuff to "lib".
> That is a very generic name so I'm not sure what is the actual scope of lib.

Fair point.
I'm going to drop the two 'address' commits from this PR (they're confusing),
and I'll sort out the naming of the libs and their contents before moving anything else.

body.h
flatcap: > But please avoid plural

I hadn't thought about that.  Noted.

lib/address.h
flatcap: > Definitely electable to a switch to TAILQ

Yep.

With independent libraries, it should be easy for someone to make and test these changes.
(someone who needn't be familiar with neomutt's code).

lib/address.h
flatcap: hehe, you prefer not to use 'rfc822_*' and below @gahr's in favour of it.
see comment below

lib/file.h
flatcap: > If there's only one consumer of this function

There are a handful of one-off functions like this.
I've chosen to move them, for now, so they're more discoverable.
The more comparisons we can make between, say file functions, the easier it should be to refactor and improve.

init.c
flatcap: `OPT_LOCALES` is a global variable, which is bundled with the general neomutt config bools.
(packed into the `Options` variable).

`OPT_LOCALES` really belongs with the multi-byte code, so I've split it out into a real bool global variable.

flatcap:
@gahr

> The whole address.h file is empty now?

Yeah, that's an oversight.  I'll drop the address changes for now.

> What is the rationale for the (partial) move from rfc822 to address?

The bits that have moved are 'pure' address functions.  They don't have any dependencies.
The remaining functions make reference to global variables, charsets and hooks.

> I think the former name makes more sense, 
> as the functions really deal with email addresses as specified in rfc822

I understand, but I see a few problems:
- "rfc822" only makes sense if you know what that RFC is for
- "RFC822 was obsoleted 15 years ago, there's a small set of RFCs that are now current
- The filenames don't match the struct they represent
  The actual struct has always had the human-friendly name `Address`, so I'd like the files to be `address.[ch]`

In a similar direction, I favour renaming `rfc1524.c` to `mailcap.c`

gahr:
Fair enough, thanks. As long as the RFCs are referenced somewhere at file level, so one knows where to look for authoritative knowledge :)

flatcap:
> As long as the RFCs are referenced somewhere at file level,
> so one knows where to look for authoritative knowledge :)

Absolutely.

PR #942, opened 2017-11-10
Tidy libconn

d-k-c:
81d5cbae2 conn: replace mutt_socket_write_d by mutt_socket_send_d
14a29c799 conn: rename mutt_socket_write to mutt_socket_send
412c5f2a6 conn: introduce mutt_socket_{read,write}
79faf247b conn: don't try to determine buffer size if length was negative
This first group of commits makes sure that functions of the mutt_socket_write_* family honours the buffer size and don't try to be smart about it. They also free up the "mutt_socket_write" symbol so that it can be reused as a wrapper over conn->conn_write. It's not used yet, but might become useful to rewrite some parts of SSL/SASL sockets

790424af4 conn: rename socket_new_conn to mutt_socket_new
Make this function name consistent with its pairs

7d792f652 conn: define mutt_ssl_socket_setup even when building without SSL
To factorise error paths and make the code a little bit easier to parse

58374d9a1 move mutt_socket_free to conn/socket.c
7f5cc01da conn: let mutt_socket_new add Connection to the list
a270395fb move Connections declaration to conn/socket.c
Let mutt_socket_{new,free} deal with the list of Connections. As a consequence, move that list definition to conn/socket.c, that's where it makes the most sense

a37571c57 conn: add parameter to mutt_socket_new to specify socket type
Let mutt_socket_new deal with socket initialisation

35f3af79c conn: move raw connection to its own file
The point is to make conn/socket.c the entry point to deal with sockets, and move the specific implementations elsewhere.

d8f69df57 conn/ssl.c: rename local var data to ssldata
66dc031d6 conn: factorise ssl_socket_open and mutt_ssl_starttls
Self-explanatory. These two commits could be squashed, but renaming different variables with the same name makes the two function easier to compare to one another

3c8452dda conn: rename tls_close to ssl_socket_close_and_restore
Cosmetic change to 1) remove mixed terminology (SSL vs. TLS) and 2) better describe what the function does


gahr:
Good job! I’m not convinced by the API naming of the various mutt_socket_write/send/_d functions. I think we can come up with something better and more consistent.

In particular, “send” reminds me of sendto(2) (udp) or sendfile(2). I would avoid that verb in this context.

conn/ssl.h
```diff
@@ -28,6 +28,11 @@ struct Connection;
 
 int mutt_ssl_starttls(struct Connection *conn);
 int mutt_ssl_socket_setup(struct Connection *conn);
+#else
+static inline int mutt_ssl_socket_setup(struct Connection *conn)
```
gahr: Uhm.. so the forward definition of `struct Connection` above is needed in all cases.

gahr:
Do I see it right that `mutt_socket_write` and `mutt_socket_read` are never used? How about using those names as a base to express the basis for that family?

I'd see something like this:

```c
int mutt_socket_write(struct Connection *, const char *); // strlen, default debug lvl  - simple API
int mutt_socket_write_n(struct Connection *, const char *, size_t); // explicit len, default debug lvl
int mutt_socket_write_d(struct Connection *, const char *, int); // strlen, explicit debug lvl
int mutt_socket_write_nd(struct Connection *, const char *, size_t, int); // explicit len and debug lvl
```

gahr:
Also, at this point would it make sense to get rid of mutt_socket.(h|c) altogether and move the debug level definitions and mutt_conn_find in lib/conn?

conn/ssl.h
d-k-c: Right. I guess I didn't get bitten by that because it was already done by another header. I'll fix that.

d-k-c:
You're absolutely right about `mutt_socket_write` and `mutt_socket_read`. The thing is that there are already a set of functions: `mutt_socket_open`, `mutt_socket_close`, `mutt_socket_poll`, etc. which are wrappers over the `Connection->conn_{open,close,poll}` callbacks. The idea was to let `mutt_socket_write` also be a wrapper over conn_write, so to me it would make more sense if they have the same prototype.

To give you some context, initially I wanted to address SSL/SASL sockets which are setup in a weird way (it doesn't use mutt_socket_new, and they hijack existing Connections to replace conn_{open,close,read,write,...} functions with their owns). The idea I had in my head was to add a new member to the Connection struct, a `struct Connection *{raw,transport}` that would be non-NULL for sockets that do encryption over a raw socket; and use mutt_socket_{read,write}(conn->raw); to do stuff with it. But I didn't get this far and I'm not even sure this solution works, so introducing these two functions was maybe a bit premature.

For `mutt_socket_send` I'll try to come up with a better name, I just wanted to avoid `mutt_socket_write_str` :)

About `mutt_conn_find`, I feel like this function is just a hack over the fact that `struct Context` doesn't keep track of its Connection, so this has to be done at another place. I would be very happy if I could see this function gone. But in the meantime, we can definitely move it to conn/.

julian-klode:
Hmm, just a quick comment: To me it makes no sense to have the functions have _socket_ in the name, but the type being Connection. I'd expect Connection to be Socket or mutt_socket_foo to be mutt_connection_foo.

julian-klode:
Oh, I see these are specific connection types. Hmm. I'm kind of used to have the functions include the type name first, so you know these are connection "methods" (object-oriented C, really). But then my initial complaint is wrong of course.

PR #946, opened 2017-11-11
autosetup: expand --everything in `neomutt -v`

flatcap:
If configure.autosetup is called with `--everything`, this commit expands that to `--gpgme --pgp ...` for display purposes.

`$ ./neomutt -v | grep Configure`

**Before:**
Configure options: --everything

**After:**
Configure options:  --gpgme --pgp --smime --notmuch --lua --tokyocabinet --kyotocabinet --bdb --gdbm --qdbm --lmdb

### DIFF
--- a/auto.def
+++ b/auto.def
@@ -299,7 +299,10 @@ if {[get-define want-everything]} {
   foreach opt {gpgme pgp smime notmuch lua tokyocabinet kyotocabinet bdb 
                gdbm qdbm lmdb} {
     define want-$opt
+    append conf_options "--$opt "
   }
+} else {
+  set conf_options "$::argv"
 }
 
 ###############################################################################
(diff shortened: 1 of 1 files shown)


auto.def
```diff
@@ -296,10 +296,14 @@ set prefix [get-define prefix]
 ###############################################################################
 # Everything
 if {[get-define want-everything]} {
+  set conf_options ""
```
gahr: Not needed

auto.def
```diff
@@ -296,10 +296,14 @@ set prefix [get-define prefix]
...
   foreach opt {gpgme pgp smime notmuch lua tokyocabinet kyotocabinet bdb 
                gdbm qdbm lmdb} {
     define want-$opt
+    set conf_options "$conf_options --$opt"
```
gahr: `append conf_option —$opt`

flatcap:
Thanks.  Fixed.

Issue #815, opened 2017-10-03
View source of a message in read-only mailbox


reisradomil:
Sometimes it is useful to see the complete raw message, e.g. to debug a mail problem or looking for spurious encoded stuff.
Currently there is no command to view the source of a message.
For read-write mailboxes there is the 'edit' command which can be abused for that.
However, this fails for read-only mailboxes with the message 'Mailbox is read-only.'

Interestingly enough, mutt is able to detect if the 'edit' command has modified the message.
It prints 'Message not modified!' in case the message was not modified.

Question: Did I miss something or is it currently not possible to view the message source from within neomutt?

Idea: Enhance the 'edit' command to print a notice if the message is from a read-only mailbox but still open the message (as it does at the moment).
If the use still modifies the message politely reject those when the editor closes.

reis

flatcap:
How about if neomutt sets the temporary file as read-only?
The editor will force the user to save a copy and neomutt won't have deal with saving a tmp file.

gahr:
I'm all for this!

reisradomil:
Since you flagged this as easy, I'm willing to give it a shot and implement that feature.

flatcap:
That's great.  Thanks.
First you'll need a copy of the source, a heap of building dependencies, a debugger...
(If you use IRC, I'll be around for a few more hours today : #neomutt on irc.freenode.net)

How good are your programming/debugging skills?
How much would you like to discover yourself?

reisradomil:
This is how I would implement that feature:
* Implement a function `view-raw-message` which opens a message as a read-only file in the editor. (regardless if the mailbox is read-only or not)
* Implement a function `edit-or-view-raw-message` which is a dispatcher for either `edit` or `view-raw-message` depending whether the mailbox is writeable or not.
* Bind the key `e` to `edit-or-view-raw-message`. This yields still the old behaviour (i.e. `edit`) in a writeable mailbox, so does not break backwards compatibility.

Reasoning behind introducing `view-raw-message`:
1. Naming a function which only views a message in a read-only mailbox `edit` is quite misleading.
Renaming `edit` to `edit-or-view-raw-message` will likely break some config files out there, so is not a solution.
2. Some people might want to ensure that a message is not accidentally edited in a writeable mailbox (e.g. if their editor does some kind of autosaving).
3. It allows a clear separation of the intension viewing and editing a message. Personally I never wanted to edit a message but viewing happens every now and then.

[...]

Issue #947, opened 2017-11-12
lib/mbyte.c does not exists


reisradomil:
The commit "Add missing source file to po/POTFILES.in" 7e2efa280d44a8da024e454140b2a8d90aaf7da5 introduced lib/mbyte.c in po/POTFILES.in. However, this file does not exists and thus lets make fail with

```
make[3]: *** No rule to make target '../lib/mbyte.c', needed by 'neomutt.pot-update'.  Stop.
```

Note that there is a file mbyte.c in the root directory.

flatcap:
That, my friend, is because flatcap is an idiot :-)
In one of my many branches, I noticed that POTFILES.in was missing a file.
I pushed the fix thinking that the corresponding code was already in master.
I've (force) fixed master.

PR #948, opened 2017-11-12
[RFC] Feature/edit or view raw message

reisradomil:
Implementation for the suggestions in #815.

# Notes for the implementation

* This introduces two new mutt-functions `view-raw-message` and `edit-or-view-raw-message`. The latter is a dispatcher for either `edit` or `view-raw-message` depending on whether the mailbox is writeable or not.

  `view-raw-message` is implemented via `mutt_view_message()` in editmsg.c. Since its functionality is 80% that of `edit`/`mutt_edit_message()`, we reused the `mutt_edit_message()` function (and its relatives). The distinction between the two cases (view or edit) is based upon an enum constant. I personally think this is cleaner than using a `bool` as it does not obfuscate its purpose or do you know what `edit_or_view_one_message(true, ctx, hdr)` does?

  Regarding my question in #815: In this implementation we use two functions `mutt_edit_message()` and `mutt_view_message()` and not a single function with a parameter `mutt_edit_or_view_messega(EditOrView, ...)`. Here I'm open for discussion in case there are other preferences.

* in any OP_..._RAW_MESSAGE (curs_main.c) we check for CHECK_ATTACH.
  Unclear if we really need that for viewing the message but better safe than sorry.

* The implementation is, I think, POSIX specific. (Uses chmod(...) call), so does not work on Windows. Setting the mode was a bit tricky as `sb.st_mode` also contains bits which describe what kind of file it is (directory, special device, ...). Therefore we project out the bits related to permissions and then modify these.

* The online help/description (inside the help menu (press `?`)) for `edit-or-view-raw-message` is a bit long. Here I'm open for suggestions.

* Default key binding for `e` is now `edit-or-view-raw-message`. This is backwards compatible as for writeable mailbox this is the same as `edit`.

* We renamed `OP_EDIT_MESSAGE` to `OP_EDIT_RAW_MESSAGE` to make it coherent with the new constants.

# Open Issues 

In `curs_main.c` we dispatch the different mutt-functions into actions/C-functions.
The code below has two(!) implementations for the dispatch of the three functions `edit`, `edit-or-view-raw-message` and `view-raw-message` and thus does *not* compile!

a) Dispatch all three mutt-functions in a single case-statement.
b) Dispatch each mutt-function in its own case-statement.

I'm kind of unsure which of the two variants is the best in this monster of switch-statement. I would appreciate your opinion here.
b) certainly introduces small code duplications (the tests at the start, the PGP stuff and the redraw at the end). a) is more complicated as we distinguish the different functions inside the case-statement again. I personally would have preferred a single function for the complete cases and then implement `edit-or-view-raw-message` as

```
case OP_EDIT_OR_VIEW_RAW_MESSAGE:
  if (!Context->readonly && mutt_bit_isset(Context->rights, MUTT_ACL_INSERT))
    edit_case_function();
  else
    view_case_function();
  break;
```

but this would require refactoring a lot.

# Further thoughts 

* It would be nice to rename `edit` to `edit-raw-message` to make it coherent with the other names.
However, this would break backwards compatibility. As a solution we could rename the function and introduce an alias `edit`.

* We could export the C-function `mutt_edit_or_view_message()`. This requires also exporting the enum EditOrView.


flatcap:
Wow, you **have** been busy!  And thorough.  I see you added docs and updated the vim syntax file.

First, please can you fix the build (it's set to fail on warnings).
Click the "Details" button in the "All checks have failed" box below.

reisradomil:
I rebased the pull request to the new master (where #947 is fixed).
I also added a commit to remove the proposal for the separate case-statements per mutt-function (to prove that the code compiles).

@reviewers: You might nevertheless have a look at the proposed code for three separate case-statements for handling the `edit`/`edit-or-view-raw-message`/`view-raw-message` functions in case this fits the style of neomutt coding better.

I might also add that there are the following scenarios to test

read-only mailbox
* do not modify the selected message
  - edit -> expect error "Mailbox is read-only"
  - edit-or-view -> expect opening the file read-only, no other messages
  - view -> expect opening the file read-only, no other messages
* modify the selected message
  - edit -> expect error "Mailbox is read-only"
  - edit-or-view -> expect opening file read-only, warning about modification and rejecting the modification
  - view -> expect opening file read-only, warning about modification and rejecting the modification

writeable mailbox
* do not modify the selected message
  - edit -> expect warning "message not modified"
  - edit-or-view -> expect warning "message not modified"
  - view -> expect opening the file read-only, no other messages
* modify the selected message
  - edit -> expect to show, edit and accept the modification.
  - edit-or-view -> expect to show, edit and accept the modification.
  - view -> expect opening file read-only, warning about modification and rejecting the modification

flatcap:
Thanks.  btw, we're @neomutt/reviewers, I'm not sure who you just pinged :-)

reisradomil:
> Thanks. btw, we're @neomutt/reviewers, I'm not sure who you just pinged :-)

I used it there to emphasise that the word 'you' following does not mean 'you flatcap' but 'you all reviewers interested'. :)

gahr:
Will definitely have a good look at it, thank you!

editmsg.c
```diff
@@ -118,7 +122,22 @@ static int edit_one_message(struct Context *ctx, struct Header *cur)
...
+    /* Note: glibc has an ALLPERMS constant */
+    mode_t allperms = S_ISUID | S_ISGID | S_ISVTX | S_IRWXU | S_IRWXG | S_IRWXO;
+    mode_t perms = sb.st_mode & allperms; /* get only bits related to permissions */
+    rc = chmod(tmp, perms & ~(S_IWUSR | S_IWGRP | S_IWOTH));
+    if (rc == -1)
+    {
+      mutt_error(_("Could not remove write permissions of %s: %s"), tmp, strerror(errno));
+      goto bail;
+    }
+  }
```
gahr: I'm not sure you really need this part. Most editors (well, vim at leat) will allow you to write the file anyway since the directory it's in belongs to the user and is writable. You are correctly checking `editOrView == VIEW && sb.st_mtime != mtime` below, so I'd say this just adds complexity.

gahr:
A couple of notes:

* nice work!
* I'd kill `mutt_view_message` and `mutt_edit_message` in favour of a single function with a flag
* I'd kill the enumeration and use a `bool readonly` or `bool readwrite` parameter. It looks readable enough in my eyes. You could pass directly the flag `edit` you are setting in `curs_main.c`.

editmsg.c
reisradomil: I'm unsure what you mean by "this part" (as the snippet only shows 4 lines):
1) bailing out.
2) the error message + bailing out.
3) marking the file as read-only.

In case 1) and 2). I can remove that since it is not essential to ensure the file is read-only (because---as you mentioned---we check for modification later / do not allow editing anyway).

In case 3): It is true that some editors allow you to write a read-only file. However, no editor AFAIK allows you to do it by accident. You have to explicitly request it (e.g. in vim with a bang!). This gives the user (hopefully) a strong indication that they are doing something which will not work as they expect.

Also most editors show some visual clue that the current file is read-only. This gives the user an additional clue that they are now in the 'view' mode and not in the 'edit' mode. This might be handy if you've forgotten whether your mailbox was read-only (remember `e` is now bound to `<edit-or-view-raw-message>`).

Though, I understand your point with the complexity.

reisradomil:
>    nice work!

Thanks.

>    I'd kill mutt_view_message and mutt_edit_message in favour of a single function with a flag
>    I'd kill the enumeration and use a bool readonly or bool readwrite parameter. It looks readable enough in my eyes. You could pass directly the flag edit you are setting in curs_main.c.

Using a single function instead of the two is no problem. In fact that was my original idea (see my question at #815). However, I'm *very* reluctant in using a bool as a flag. From inside (i.e. at the callee site) the code might be readable as we have the variable name, e.g.
```
int edit_or_view_one_message(bool readonly, struct Context *ctx, struct Header *cur)
{
...
    if (readonly) ...
...
}
```
But on caller site this looks completely different:
```
edit_or_view_one_message(true, ctx, hdr);  /* 1 */
edit_or_view_one_message(false, ctx, hdr); /* 2 */
```
Ask yourself the question which of the calls is the edit which is the view one? (If you said 1 is view then you'd be wrong as I implemented it like edit_or_view_one_message(bool writeable, ...)`). Hope you get my point. Here are some more resources about boolean flags [1,2,3].

[1] http://www.nickhodges.com/post/How-Not-To-Code-2-Dont-Use-Boolean-Method-Parameters.aspx
[2] https://martinfowler.com/bliki/FlagArgument.html
[3] https://softwareengineering.stackexchange.com/questions/147977/is-it-wrong-to-use-a-boolean-parameter-to-determine-behavior

gahr:
I meant the whole chmod thing. I understand your point about having an editor somehow show that the file is read-only, so I’m fine with removing the write bit anyway. If you do, however, then I think it would be better to get the actual mode (stat) and just remove what you don’t want (write bits).

Regarding using an enum instead of a bool, uhm.. I think I might somewhat agree on the principle case, but there are a few things to note in the specific case of our codebase, and on the fact that we’re using C:

* C’s enums are not strongly types, so nobody prevents you from writing `f(6, ...);` where f’s first parameter is some enum
* C’s enums (just like anything else, fwiw) are not namespace’d, which is why they usually embed a namespace in the tags. To reduce the likelyhood of clashes you’d have something like `EOV_VIEW` and `EOV_VIEW`... pretty ugly
* neomutt’s codebase does not currently employs enums for that, so your code would read somewhat foreign

I prefer the alternative approach where you don’t call your function with `true` or `false` directly, but assign the bool value to a variable to make it clear at call site. This is usually more natural to write, and your code is evidence of it: you have an `edit` variable already, you just need to use it. Compare these two examples:

```c
// your approch
if (edit)
  mutt_view_or_edit_message(VOE_EDIT, ...);
else
  mutt_view_or_edit_message(VOE_VIEW, ...);

// my approach
mutt_view_or_edit(edit, ...);
```

gahr:
Anyway, I don’t have a strong opinion on either approach. Perhaps what you have done - two entry points in the public API that dispatch to a single static function - is still the cleanest. In this case, being the call scope of the static function very clear and being the two exposed functions only calling the static one, having a bool or an enum doesn’t change much. Just nitpicking on potential improvements;)

reisradomil:
Thanks for the feedback.

> Regarding using an enum instead of a bool, uhm.. I think I might somewhat agree on the principle case, but there are a few things to note in the specific case of our codebase, and on the fact that we’re using C: [...some points...]
> [...] Perhaps what you have done - two entry points in the public API that dispatch to a single static function - is still the cleanest.

Ok. I keep the two functions `mutt_view_raw_message()` and `mutt_edit_raw_message()` but exchange the enum for a `bool`.

> I meant the whole chmod thing. [...] I’m fine with removing the write bit anyway. If you do, however, then I think it would be better to get the actual mode (stat) and just remove what you don’t want (write bits).

You mean like this direct approach
```c
rc = stat(tmp, &sb);
//...
rc = chmod(tmp, sb.st_mode & ~(S_IWUSR | S_IWGRP | S_IWOTH));
//...
```

This was my initial thought. However, `sb.st_mode` does not only contain bits for permissions but also bits for what type of file `tmp` is (link, special device, directory, ...). So `sb.st_mode & ~(S_IWUSR | S_IWGRP | S_IWOTH)` would not only be a permission mask but also a filetype mask.

Sadly the documentation of `chmod()` [1,2a,2b] does not describe what happens if the mode parameter has non-permission bits set. [1] even says

> The new file mode is specified in mode, which is a bit mask created by ORing together zero or more of the following: [list of permission constants only]

which can be read that only masks with permission bits are allowed. To me this reeks of undefined behaviour/grey area. Therefore, I added the clearing of all non-permission bits. If the direct approach above is deemed safe, I'm more than happy to remove this bit juggling part.


[1] http://man7.org/linux/man-pages/man2/chmod.2.html
[2a] https://www.gnu.org/software/libc/manual/html_node/Permission-Bits.html#Permission-Bits
[2b] https://www.gnu.org/software/libc/manual/html_node/Setting-Permissions.html#Setting-Permissions


To incorporate your remarks, I would amend and rebase my branch. Is this ok or should I keep this history intact and add a correcting commit?
 
Another question: Do you have any opinion about 1 case vs 3 cases for handling the three related mutt-functions [3]?

[3] https://github.com/neomutt/neomutt/pull/948/commits/8df48e347fe6ba244fcbbb4424d5956670481be7

And what is your opinion about trying to unify the function names by migrating `<edit>` to `<edit-raw-message>` by aliasing the two (removing `<edit>` would be a breaking change, so shouldn't be done in the near future)?

gahr:
I'm pretty sure you don't have to care about non-permission related flags in chmod. Calling with just `S_IRUSR| S_IRGRP|S_IROTH` should be fine.

> To incorporate your remarks, I would amend and rebase my branch. Is this ok or should I keep this history intact and add a correcting commit?

Feel free to either rebase or add another commit, whatever is easier for you. We might eventually squash the branch on merging.

> Another question: Do you have any opinion about 1 case vs 3 cases for handling the three related mutt-functions [3]?

1 single case is fine to me.

> And what is your opinion about trying to unify the function names by migrating <edit> to <edit-raw-message> by aliasing the two..?

Uhm, let's see how it would look like :)

Thank you!

flatcap:
I understand your point about bool vs enum params, but I'd prefer bools for now.

The NeoMutt source is on a long journey...
We've tidied up function names, struct/typedef names, global variables and we're working on bools
(Mutt doesn't use bools at all).  When that's done, we can consider enums.

Other than that, the code looks good.
I've tested it on a normal mailbox and a truly readonly mailbox (compressed).
Using a "readonly" mailbox, using `%`, works as expected.

Elsewhere in the code, NeoMutt simply does `chmod(400)` to remove permissions (ignoring existing perms).  Your code's smarter, but I'd prefer to see it in `lib/file.c` -- perhaps a `chmod_rm()`.

editmsg.c
```diff
@@ -118,7 +121,19 @@ static int edit_one_message(struct Context *ctx, struct Header *cur)
...
 
-  mtime = mutt_decrease_mtime(tmp, &sb);
+  /* remove write permissions */
+  if (!edit)
+  {
+    rc = chmod(tmp, sb.st_mode & ~(S_IWUSR | S_IWGRP | S_IWOTH));
+    if (rc == -1)
+    {
+      mutt_error(_("Could not remove write permissions of %s: %s"), tmp, strerror(errno));
+      goto bail;
```
gahr: Yep, sorry for not getting to this earlier. I'd probably not bail out, just log something and let it be :)

reisradomil: Will do it shortly.

reisradomil:
> Elsewhere in the code, NeoMutt simply does chmod(400) to remove permissions (ignoring existing perms). Your code's smarter, but I'd prefer to see it in lib/file.c -- perhaps a chmod_rm().

I can do that but I would prefer to do it in a separate pull-request as this also touches code from attach.c which is not related to this issue.

To be on the same page: `int mutt_chmod_rm(const char *p, mode_t mode)` should remove the permissions specified by mode. To me the name looks a bit odd (chmod_rm reads like "change mode remove"), how about `mutt_rmmod(...)`?

reisradomil:
> I'd probably not bail out, just log something and let it be :)

Changed the fail into a warning. The caveat is that if there is a second warning (e.g. the user modifies the file) then this first warning will not be seen.

gahr:
Oh, I wouldn't even bother informing the user, I'd just log it (`mutt_debug`)

flatcap:
> I can do that but I would prefer to do it in a separate pull-request as this also touches code from attach.c which is not related to this issue.

Yes, that's a good idea.

> To be on the same page: int mutt_chmod_rm(const char *p, mode_t mode) should remove the permissions specified by mode.

Yes.  I imagined:
- `chmod` setting the perms exactly
- `chmod_rm` removing specific perms
- `chmod_add` adding specific perms

Of course we only need to implement 'rm' for now.

> To me the name looks a bit odd (chmod_rm reads like "change mode remove"), how about mutt_rmmod(...)?

Yeah.  `mutt_rmmod()` will do for now.

gahr:
Thanks @reisradomil, I don't have any other comments related to the code. Do you need a hand with testing or have you tested all cases you could think of already?

reisradomil:
I've tested all the cases (ro/rw mailbox, modified/unmodified message, 3 functions) described in my second commend. At the moment I cannot think of any other test cases.

reisradomil:
Are you going to squash all these commits? If not, then I would rebase and incorporate the last 6 commits into the "main" commits of this pull-request to have a saner history.

flatcap:
> Are you going to squash all these commits?

We tend to squash commits where possible.
If you'd like to tidy the history (and the log message) yourself, then please go ahead.

flatcap:
One other thing.
Please can you update your git config with your name and an email that you're happy being public.

Issue #944, opened 2017-11-10
pgp_decrypt_command fails to decrypt saved outbound PGP/MIME msgs w/self encryption


libBletchley:
`gpg.conf` includes the `encrypt-to` parameter to ensure that all outbound messages are encrypted to myself (in addition to the recipient).  So I can go into my _sent_ mail and review an encrypted message that I've sent if I sent it in the old PGP/inline format.  Since I've recently discovered that protonmail can handle PGP/MIME for messages that have no attachment, I've switched to PGP/MIME for those cases.  Now when I try to open a message that I sent in the PGP/MIME format, mutt asks for my password and then I get:

"Could not decrypt PGP message"

Note that in the post-GPG-2.1 state of things, the mutt GPG password prompt is useless.  I generally enter junk for that and enter the correct passphrase when GPG graphically prompts for it.  But in the case at hand, gpg does not prompt for it.  Of course I've tried giving mutt my correct gpg passphrase but it makes no difference.

These are the relevant settings:
```
unset crypt_use_gpgme
set pgp_decrypt_command="/usr/bin/gpg --passphrase-fd 0 --pinentry-mode loopback --no-verbose --quiet --batch --output - --decrypt %f"
```

Workaround 1: do a `copy-message` to a file then manually run `gpg` on that file.  (In fact if you use the command above in `pgp_decrypt_command`, and replace `%f` with the filename and pipe your password in, it works, proving that this is a mutt problem)
Workaround 2: `set fcc_clear = yes` so the saved msgs are in plaintext.
Workaround 3: enable GPGME (which breaks other things)

libBletchley:
I got this working.  It wasn't broken code, just a change to the config file.  Apparently the passphrase had to be an optional parameter.

PR #949, opened 2017-11-12
Introduce GPG config file changes to accomodate gpg 2.1

libBletchley:
* **What does this PR do?**
 * renames `gpg.rc` to `gpg2.rc`
 * introduces `gpg2_1.rc` which brings in a parameter that does not exist in older GPG versions
 * places gpg commands at the end of the commandline, as required by the (often unenforced) syntax rules
 * remove use of `--always-trust` in the new version (on the advice of GPG folks)
 * ultimately fixes bug https://github.com/neomutt/neomutt/issues/944

I didn't do any testing on the removal of `--always-trust`.

I think so.  I do not believe this change requires any of the following.

     for that)

   - All builds are passing

https://github.com/neomutt/neomutt/issues/944

### DIFF
--- a/contrib/gpg.rc
+++ b/contrib/gpg.rc
@@ -2,7 +2,13 @@
 #
 # Command formats for gpg.
 #
-# This version uses gpg-2comp from
+# Version notes:
+#
+#   GPG 2.1 introduces the option "--pinentry-mode", which requires
+#   the "loopback" argument in instances where "--passphrase-fd" is
+#   used.
+#
+# The gpg-2comp found in some comments comes from
 #   http://70t.de/download/gpg-2comp.tar.gz
 #
 # %p    The empty string when no passphrase is needed,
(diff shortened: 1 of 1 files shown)


gahr:
Given that 2.1.0 is 3 years old, do you think it would make sense to just upgrade the file to be 2.1-compliant and drop support for < 2.1?

libBletchley:
Debian is always a bit behind.  Debian _Stretch_ was released ~5 months ago.  The predecessor (Jessie) uses the pre-2.1 config.  There have been moments that I was 2 Debian releases behind due to stuff breaking in next release.  I'm not sure how common that is - probably rare.

The end of life date for Jessie long-term-support is ~June 6th 2020 (https://wiki.debian.org/DebianReleases).  I'm not sure if that's a factor.. If someone is stuck on Jessie that long, they should in principle also be using the old version of mutt 1.5.23-3.  But if such a user would take a bleeding edge version of mutt it'll break no matter what, but having the old config file there would hopefully provide a quick fix.

It's your call.. I just noticed there are multiple versions of `pgp*.rc` files and followed the pattern.

libBletchley:
Another option would be to put the contents of `gpg2.rc` in `gpg2_1.rc`, as comments, and rename back to `gpg.rc`.

gahr:
Debian has a record for maintaining their own patched versions of most software. I don't think we should wait on them to drop support for EOL'd 3rd party software. Debian is free to maintain patches against gpg.rc as well.

libBletchley:
I've consolidated back to `gpg.rc`.  The history should be preserved the way that I did it (`git mv gpg2.rc gpg.rc`, `cp gpg2_1.rc gpg.rc`, `git rm gpg2_1.rc`).

I also noticed the `default.auto` file was a Tcl file that appears in fundamental mode in emacs due to the non-standard extension, so I added the tag to fix that.

gahr:
Thanks @libBletchley. A couple of notes: 

* `autosetup/default.auto` is imported from upstreams, we do not maintain that, and it has nothing to do with this issue. Please revert that change.
* I'd rather not say `Command formats for GnuPG 2.1`. If you want to specify a version, please use `>=`. I'm running 2.2.2 here.
* Could you please try to keep the diffs to a minimum? There are a few things that do not belong to this issue. Ideally, we should only see changes in `set *_command = ....` lines, right?

libBletchley:
> I'd rather not say Command formats for GnuPG 2.1. If you want to specify a version, please use >=. I'm running 2.2.2 here.

I've removed the version number, as I'd rather not predict what to expect of future versions.

> Could you please try to keep the diffs to a minimum? There are a few things that do not belong to this issue. Ideally, we should only see changes in set *_command = .... lines, right?

I'm not quite sure what you're referring to.  There was some noise b/c I had two config files and went back to one, but if you just diff prior to the first commit with the last commit that noise won't appear.  Or if you mean my removal of "PGP/MIME" in some of the comments, those were misleading because they were commenting commands that are also used for PGP/inline operations.  So I removed the PGP/MIME comment in the cases that were clearly both pgp/mime and inline.  I'm not sure if I got them all though.

contrib/gpg.rc
```diff
@@ -1,10 +1,18 @@
...
+#
+#   GPG 2.1 introduces the option "--pinentry-mode", which requires
+#   the "loopback" argument in instances where "--passphrase-fd" is
+#   used.
+#
+# The gpg-2comp found in some comments comes from
 #   http://70t.de/download/gpg-2comp.tar.gz
 #
+# $Id: gpg.rc,v 1.12 2001/12/11 09:33:57 roessler Exp $
+#
```
gahr: Where does this come from?

contrib/gpg.rc
```diff
@@ -29,58 +37,72 @@
...
+set pgp_clearsign_command="gpg %?p?--pinentry-mode loopback --passphrase-fd 0? --no-verbose --batch --quiet --output - --armor --textmode %?a?--local-user %a? --clearsign %f"
 
-# create a pgp/mime encrypted attachment
+# Create an encrypted attachment (note that some users include the --always-trust option here)
+#
 # set pgp_encrypt_only_command="/usr/lib/neomutt/pgpewrap gpg-2comp -v --batch --output - --encrypt --textmode --armor --always-trust -- -r %r -- %f"
-set pgp_encrypt_only_command="/usr/lib/neomutt/pgpewrap gpg --batch --quiet --no-verbose --output - --encrypt --textmode --armor --always-trust -- -r %r -- %f"
+#
+set pgp_encrypt_only_command="/usr/lib/mutt/pgpewrap gpg --batch --quiet --no-verbose --output - --textmode --armor -- --recipient %r -- --encrypt %f"
```
gahr: please keep `/usr/lib/neomutt/`

contrib/gpg.rc
```diff
@@ -29,58 +37,72 @@
...
 
-# create a pgp/mime encrypted and signed attachment
+# Create an encrypted and signed attachment (note that some users include the --always-trust option here)
+# 
 # set pgp_encrypt_sign_command="/usr/lib/neomutt/pgpewrap gpg-2comp %?p?--passphrase-fd 0? -v --batch --output - --encrypt --sign %?a?-u %a? --armor --always-trust -- -r %r -- %f"
-set pgp_encrypt_sign_command="/usr/lib/neomutt/pgpewrap gpg %?p?--passphrase-fd 0? --batch --quiet --no-verbose --textmode --output - --encrypt --sign %?a?-u %a? --armor --always-trust -- -r %r -- %f"
+#
+set pgp_encrypt_sign_command="/usr/lib/mutt/pgpewrap gpg %?p?--pinentry-mode loopback --passphrase-fd 0? --batch --quiet --no-verbose --textmode --output - %?a?--local-user %a? --armor -- --recipient %r -- --sign --encrypt %f"
```
gahr: please keep `/usr/lib/neomutt/`

contrib/gpg.rc
libBletchley: I forgot that I did that.  Those are anchors.  Some projects want comments to appear above the code they refer to, and some want comments to appear below the referenced code.  I've also worked on a project that had it both ways, depending on what kind of code it was.  The purpose of the anchors are to clearly attach the comment to the SLOC it refers to.

contrib/gpg.rc
libBletchley: Ah, I overlooked that.  Debian renames neomutt to mutt, so that was Debian stuff creeping in.  I'll fix it.

contrib/gpg.rc
```diff
@@ -29,58 +35,72 @@
...
 
-# fetch keys
+# Fetch keys
 # set pgp_getkeys_command="pkspxycwrap %r"
 
 # pattern for good signature - may need to be adapted to locale!
-
-# set pgp_good_sign="^gpgv?: Good signature from "
-
-# OK, here's a version which uses gnupg's message catalog:
```
gahr: This command was probably useful.

libBletchley: I was merging a couple versions, and didn't notice the `v?` difference from this one:
`# set pgp_good_sign="^gpg: Good signature from"`

Indeed the `v?` is good to have.

gahr: I meant, this comment: `-# OK, here's a version which uses gnupg's message catalog:`

libBletchley:
Is there anything left for me to do?  Do you want the comment anchors removed?

flatcap:
That looks good @libBletchley

Before I merge it, please can you:
- Squash the commits into one neatly described commit
- Fix your commit credentials.  The commits currently say "GIT user"
- Send @flatcap an email, so he can send you an project invitation

Thanks

PR #950, opened 2017-11-12
Progressive debugging messages

flatcap:
There are about 700 calls to `mutt_debug()`.
These commits make `mutt_debug()` smarter.
About 300 of them have the function's name in their message.

- 3bef0460 replace mutt_debug with a macro
  The macro passes `__FILE__` and `__LINE__` to `mutt_debug_real()`

- ba4f04c8 Strip function names from debug messages
  Remove function names and a few references to `__FILE__` and `__LINE__`

- ~b817d5aa1 Add progressive debugging info~

### DIFF
--- a/alias.c
+++ b/alias.c
@@ -75,7 +75,7 @@ static struct Address *expand_aliases_r(struct Address *a, struct ListHead *expn
         {
           if (mutt_str_strcmp(a->mailbox, np->data) == 0) /* alias already found */
           {
-            mutt_debug(1, "expand_aliases_r(): loop in alias found for '%s'\n", a->mailbox);
+            mutt_debug(1, "loop in alias found for '%s'\n", a->mailbox);
             i = true;
             break;
           }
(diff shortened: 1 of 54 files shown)


gahr:
I like the idea, but I'm more and more convinced that debug levels make little sense in mutt. My understanding is that:

* people do not usually run with debug turned on (I mean runtime `-d`)
* people turn on debugging when they need to, ehm, debug a problem
* at this point, all info we can give is valuable.

How about we take a more radical approach, get rid of debugging levels altogether, and always print the full `[DATE TIME] FILE:LINE:FUNCTION() MESSAGE` log?

Especially, I don't see a practical difference between printing and not printing `FILE:LINE:FUNCTION()`, or why one would not want to see them.

flatcap:
> people turn on debugging when they need to, ehm, debug a problem
> at this point, all info we can give is valuable.

Yes, true.

> always print the full `[DATE TIME] FILE:LINE:FUNCTION() MESSAGE` log?

We'll see what others think, but OK.

> get rid of debugging levels altogether

Hmm... Just because they're not being used well doesn't mean they're not useful.
Unfortunately, to make them useful would require quite a lot of editing.

d-k-c:
I don't really like the `FILE:LINE` thing,  I believe `FUNCTION()` is enough (mostly because the line info tend to get inaccurate fairly quickly).

As for log level, I'd rather seem them wrapped in functions/macros with explicit log levels in their names: `mutt_trace`, `mutt_debug`, `mutt_info`, `mutt_warn`, `mutt_error` (that's just examples), and let `-d` take a string with log level to use. So far, I still have to look at the source code to know if mutt_debug(0,...) is more verbose than mutt_debug(5,...) or not.

And I think we should keep log levels. Sometimes, we want to dump what goes over the wire (like everything that goes in and out of sockets), and some other times we just want to get of broad idea of the code path.

flatcap:
[Sorry, this gets a bit rambly]

> I don't really like the FILE:LINE thing, I believe FUNCTION() is enough (mostly because the line info tend to get inaccurate fairly quickly).

I take your point.  A quick grep shows there are 19 functions that have duplicated mutt_debug messages.
Changing them to "msg1", "msg2" would mean we wouldn't need FILE and LINE.

> As for log level, I'd rather seem them wrapped in functions/macros with explicit log levels in their names: mutt_trace, mutt_debug, mutt_info, mutt_warn, mutt_error (that's just examples)

That sounds good to start with, but `mutt_debug()` isn't the primary way to communicate with the users.
All of the debugging messages are just that: "debugging".  What would a debug "error" mean?
If it's an error, why haven't we communicated it to the user with `mutt_error()`?

`mutt_message()` and `mutt_error()` already get logged, so what would these different levels of debug-ness mean?

Replacing `mutt_debug(1, ` with `mutt_debug_info(` or `mutt_debug(INFO,` would certainly help us use them more consistently, but we have two problems:
- Defining meaningful 'levels'
- Fixing the 700 existing uses

Then there's the users.  We don't want the situation where we ask for `-d 2` logging, only to tell the user, sorry we need `-d 3` logging, ...  When we turn it on, we might as well have **all** the info available.

> Sometimes, we want to dump what goes over the wire

Yes, but now we have a security problem.
We can't have neomutt leaking account info to a file (that the user may not be aware of).

The alternative to levels would be some sort of topical logging:
- LOG_IMAP_COMMANDS
- LOG_FS_COMMANDS

but I don't think the user should be able to enable these levels.

Unless the CLI gets a 'sensitive` logging option `-dS`

gahr:
> Unless the CLI gets a 'sensitivelogging option-dS`

Let's go for it :)

flatcap:
Update.  I've pushed two commits.
-b306ed99 Use a macro
  * Messages logged to file will always contain the function name
  * Macro still has `__FILE__` and `__LINE__` parameters (not currently used)
- ec4c2208 Strip back the debug messages
  * Remove the function names from the debug messages
  * Make sure that each message is unique within a given function

Removing the logging levels (or adding a sensitive option) will require a audit of the logging.
This will have to wait until someone has some free time.

Issue #945, opened 2017-11-11
Renaming functions


flatcap:
So far we've taken over a hundred functions moved then into a library.
All of them are fully documented and independent of neomutt.
As such, they can be easily unit-tested.

Now it's time to sort out the naming.
These are my suggestions, all of which are open to discussion.

## 1. Rename the directory

`lib` is meaningless.  These are utility functions so, perhaps:

* `util/` and `libutil`
* `utils/` and `libutils`

## 2. Pick a library prefix

Name all of our functions to prevent clashes with any of the standard libraries.
Ideally the prefixes want to be short, but meaningful.

* `mutt_`
* `util_` (or `utils_`)
* `m_`
* `m`
* `u_`
* `u`
* `mu_`
* `mu`
* \<none\> (rely on the library prefixes)

I quite like `u_` or `mu_`.

## 3. Pick library prefixes

The functions in the library are grouped into files by type.
Suggested prefixes:

| File        | Prefix   |
| :---------- | :------- |
| `base64.c`  | `b64`    |
| `buffer.c`  | `buffer` |
| `date.c`    | `date`   |
| `file.c`    | `file`   |
| `hash.c`    | `hash`   |
| `list.c`    | `list`   |
| `mapping.c` | `map`    |
| `md5.c`     | `md5`    |
| `memory.c`  | `mem`    |
| `sha1.c`    | `sha1`   |
| `string.c`  | `str`    |

`buffer` hasn't been abbreviated because it refers to `struct Buffer`

## 4. Rename the functions

* Apply the library prefixes
* Tidy up some of the more ridiculous naming, e.g. `safe_`
* Ignore the static functions
* Ignore some 'magic' functions, e.g. `mutt_debug()`

| Original function          | Tidied base name   | New function              |
| :------------------------- | :----------------- | :------------------------ |
| mutt_from_base64           | decode             | mu_b64_decode             |
| mutt_to_base64             | encode             | mu_b64_encode             |
|                            |                    |                           |
| mutt_buffer_addch          | addch              | mu_buffer_addch           |
| mutt_buffer_addstr         | addstr             | mu_buffer_addstr          |
| mutt_buffer_free           | free               | mu_buffer_free            |
| mutt_buffer_from           | from               | mu_buffer_from            |
| mutt_buffer_init           | init               | mu_buffer_init            |
| mutt_buffer_new            | new                | mu_buffer_new             |
| mutt_buffer_printf         | printf             | mu_buffer_printf          |
[...]

PR #955, opened 2017-11-16
Sort out naming of the library functions

flatcap:
Issue #945 - Renaming functions

I'm not sure what I'm expecting people to do with this enormous PR.
(just give people one last chance to change their minds)
I've updated the original issue to represent the final naming decisions.

- b8bff4f32 rename lib to mutt
- 26ee8370c rename base64 functions
- a7db5c520 rename buffer functions
- 35a328e68 rename date functions
- cfcc2d200 rename file functions
- cd49d502e rename hash functions
- 79662908b rename list functions
- 199b0e09c rename mapping functions
- 560b58712 rename mbyte functions
- 9f844b4c2 rename md5 functions
- 03f5fece5 rename memory functions
- f50d5b40a rename sha1 functions
- 226540c79 rename string functions


toogley:
regarding b8bff4f32: Just curiosity, why do you guys think `lib` is less readable than `mutt` ?

flatcap:
'lib' is a bit vague. While the was one library it didn't matter.
As we break apart more code dependencies, there'll be more libraries, so it'll be important to name them clearly.
It's 'mutt' for now as it's the main mutt-y bits.  It may get renamed when we have a better understanding of our needs.

PR #957, opened 2017-11-17
neomutt-syntax.vim: Recognise <edit-raw-message> function

reisradomil:
* **What does this PR do?**

I'm deeply sorry but I missed to update the `neomutt-syntax.vim` file to recognise the newly introduced function `<edit-raw-message>` (which is  synonym for `<edit>`). The function was introduced in pull-request #948.


flatcap:
Merged in https://github.com/neomutt/neomutt.vim/commit/766dbf3d5db5114eb62f4a29c791edbff6565324

PR #958, opened 2017-11-18
Add chmod like functions to mutt/file.[hc]

reisradomil:
As discussed in pull-request #948 this implements three functions to set, add, and remove file permissions.

**Notes**

* I followed flatcap's naming convention for the functions as suggested in #948.

* `mutt_file_chmod()` is really just a `chmod()`. I added it anyway to have a complete API in `file.c`.

* I copied the approach of `mutt_file_decrease_mtime()` and gave `mutt_file_chmod_add/remove()` an optional parameter `struct stat sb` through which the caller can pass the `struct stat` for the file. Theoretically this saves a systemcall, although I doubt its usefulness.

* I converted the three calls of `chmod()` to the new functions. Please double check that the use of `mutt_file_chmod_add/remove()` is correct and shouldn't be `mutt_file_chmod()`.

* I tested the code by saving an attachment and comparing its permissions which a vanilla neomutt's saved attachment. They are the same (`rw-------`). Don't know if there are any other tests to perform (in particular how to test the error branch with `mutt_file_chmod_add()`).

### DIFF
--- a/editmsg.c
+++ b/editmsg.c
@@ -120,10 +120,10 @@ static int edit_or_view_one_message(bool edit, struct Context *ctx, struct Heade
     goto bail;
   }
 
-  /* remove write permissions */
   if (!edit)
   {
-    rc = chmod(tmp, sb.st_mode & ~(S_IWUSR | S_IWGRP | S_IWOTH));
+    /* remove write permissions */
+    rc = mutt_file_chmod_rm_stat(tmp, S_IWUSR | S_IWGRP | S_IWOTH, &sb);
     if (rc == -1)
     {
       mutt_debug(1, "Could not remove write permissions of %s: %s", tmp, strerror(errno));
--- a/mutt/file.c
+++ b/mutt/file.c
@@ -29,6 +29,11 @@
  * | :---------------------------- | :-----------------------------------------------------------
  * | mutt_file_basename()          | Find the last component for a pathname
  * | mutt_file_check_empty()       | Is the mailbox empty
+ * | mutt_file_chmod()             | Set permissions of a file
+ * | mutt_file_chmod_add()         | Add permissions to a file
+ * | mutt_file_chmod_add_stat()    | Add permissions to a file
+ * | mutt_file_chmod_rm()          | Remove permissions from a file
+ * | mutt_file_chmod_rm_stat()     | Remove permissions from a file
  * | mutt_file_concatn_path()      | Concatenate directory and filename
  * | mutt_file_concat_path()       | Join a directory name and a filename
  * | mutt_file_copy_bytes()        | Copy some content from one file to another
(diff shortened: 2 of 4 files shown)


editmsg.c
```diff
@@ -123,7 +123,7 @@ static int edit_or_view_one_message(bool edit, struct Context *ctx, struct Heade
   /* remove write permissions */
```
flatcap: How about moving the comment next to the chmod?

mutt/file.c
```diff
@@ -52,6 +52,9 @@
  * | mutt_file_set_mtime()         | Set the modification time of one file from another
  * | mutt_file_symlink()           | Create a symlink
  * | mutt_file_touch_atime()       | Set the access time to current time
+ * | mutt_file_chmod()             | Set permissions of a file
+ * | mutt_file_chmod_add()         | Add permissions to a file
+ * | mutt_file_chmod_rm()          | Remove permissions from a file
```
flatcap: This list is in alphabetical order

mutt/file.h
```diff
@@ -59,6 +59,9 @@ void        mutt_file_set_mtime(const char *from, const char *to);
 int         mutt_file_symlink(const char *oldpath, const char *newpath);
 int         mutt_file_to_absolute_path(char *path, const char *reference);
 void        mutt_file_touch_atime(int f);
+int         mutt_file_chmod(const char *path, mode_t mode);
+int         mutt_file_chmod_add(const char *path, mode_t mode, struct stat *st);
+int         mutt_file_chmod_rm(const char *path, mode_t mode, struct stat *st);
```
flatcap: This list is in alphabetical order

mutt/file.c
```diff
@@ -972,6 +975,77 @@ void mutt_file_touch_atime(int f)
...
+ * @param path Filename
+ * @param mode the permissions to remove
+ * @param st   struct stat for the file (optional)
+ * @retval same as chmod(2)
+ *
+ * Removes the given permissions from the file. Permissions not mentioned in
+ * mode will stay as they are. This function resembles the `chmod ugoa-rwxXst`
+ * command family. Example:
+ *
+ *    mutt_file_chmod_remove(path, S_IWUSR | S_IWGRP | S_IWOTH, NULL);
```
flatcap: Typo, the function's called `mutt_file_chmod_rm()`

reisradomil:
Thanks for having a look and spotting my mistakes. Should now all be fixed.

flatcap:
I've just pushed some minor fixes to the doxygen comment blocks:
- `@retval` highlights the first word after it, so I've added and 'int'
- I've indented the examples by 1 more char to make doxygen highlight it as a code block
  (that leaves 4 spaces after the space-star-space which is eaten by doxygen)

```diff
diff --git a/mutt/file.c b/mutt/file.c
index a05a9713..acca1b80 100644
--- a/mutt/file.c
+++ b/mutt/file.c
@@ -979,7 +979,7 @@ void mutt_file_touch_atime(int f)
  * mutt_file_chmod - change permissions of a file
  * @param path Filename
  * @param mode the permissions to set
- * @retval same as chmod(2)
+ * @retval int same as chmod(2)
  *
  * This is essentially chmod(path, mode), see chmod(2).
  */
@@ -993,13 +993,13 @@ int mutt_file_chmod(const char *path, mode_t mode)
  * @param path Filename
  * @param mode the permissions to add
  * @param st   struct stat for the file (optional)
- * @retval same as chmod(2)
+ * @retval int same as chmod(2)
  *
  * Adds the given permissions to the file. Permissions not mentioned in mode
  * will stay as they are. This function resembles the `chmod ugoa+rwxXst`
  * command family. Example:
  *
- *    mutt_file_chmod_add(path, S_IWUSR | S_IWGRP | S_IWOTH, NULL);
+ *     mutt_file_chmod_add(path, S_IWUSR | S_IWGRP | S_IWOTH, NULL);
  *
  * will add write permissions to path but does not alter read and other
  * permissions.
@@ -1022,13 +1022,13 @@ int mutt_file_chmod_add(const char *path, mode_t mode, struct stat *st)
  * @param path Filename
  * @param mode the permissions to remove
  * @param st   struct stat for the file (optional)
- * @retval same as chmod(2)
+ * @retval int same as chmod(2)
  *
  * Removes the given permissions from the file. Permissions not mentioned in
  * mode will stay as they are. This function resembles the `chmod ugoa-rwxXst`
  * command family. Example:
  *
- *    mutt_file_chmod_rm(path, S_IWUSR | S_IWGRP | S_IWOTH, NULL);
+ *     mutt_file_chmod_rm(path, S_IWUSR | S_IWGRP | S_IWOTH, NULL);
  *
  * will remove write permissions from path but does not alter read and other
  * permissions.
```

reisradomil:
Your fixes are fine with me.

gahr:
One thing might be to have two separate sets of functions: those that take a struct stat and those that don’t. Or macros to avoid having a useless NULL paramether there. But we don’t like macros like these, do we? ;)

reisradomil:
@gahr commented
> One thing might be to have two separate sets of functions: those that take a struct stat and those that don’t. Or macros to avoid having a useless NULL paramether there. But we don’t like macros like these, do we? ;)

Be careful what you wish for! And yes we don't like macros. :)

PR #962, opened 2017-11-21
Remove vim syntax file from the main repo

guiniol:
This PR removes the vim syntax file from the main repo since the syntax file is now in an extension at https://github.com/neomutt/neomutt.vim


flatcap:
Please can you remove the file from the two makefiles, too.
- `doc/Makefile.am`
- `doc/Makefile.autosetup`

PR #966, opened 2017-11-26
Devel/autosetup path fixes

julian-klode:
* **What does this PR do?**

This pull requests adds support for ${prefix}, properly uses libexecdir instead of libdir for helper tools, and updates autosetup to the latest master version in order to fix msteveb/autosetup#33.

### DIFF
--- a/Makefile.autosetup
+++ b/Makefile.autosetup
@@ -38,6 +38,7 @@ datadir=	@PKGDATADIR@
 docdir=		@PKGDOCDIR@
 mandir=		@mandir@
 libdir=		@libdir@
+libexecdir=		@libexecdir@
 sysconfdir=	@SYSCONFDIR@
 
 # targets for specific subdirectories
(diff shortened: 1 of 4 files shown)


julian-klode:
It seems we also should change the .rc contrib files to use libexecdir correctly instead of hardcoding /usr/lib. The best approach, that evaluates any variables like ${prefix} is to just do this from the Makefile.

Otherwise, autosetup has to be patched so I can specify stuff like --libexecdir=\${prefix}/lib and it writes /usr/lib for @libexecdir@. Not sure if that's something @msteveb would like to do?

flatcap:
The location, `/usr/lib/neomutt` was a compromise.
Some systems have `/usr/lib64`, but most still have a `/usr/lib`,
but not everyone has `/usr/libexec` (which was one of my first thoughts).

By picking one **fixed** location, it meant I didn't need to edit the config files at install time.

julian-klode:
Well, but it's not /usr/lib/neomutt everywhere, it's $libdir/neomutt. Which on Debian normally would be /usr/lib/<triplet>/neomutt. $libexecdir is no different, but actually meant for that kind of binaries. On systems without a separate libexec, $libexecdir and $libdir are usually the same.

The rc files also don't work if you install into /usr/local, as the binaries are in /usr/local/lib then.

msteveb:
If you want libexecdir to be $prefix/lib instead of $exec_prefix/libexec, then you can either do:

`--prefix='${prefix}/lib'`

Or you can set the default with the latest autosetup by adding:

```
options-defaults {
    prefix /usr
    libexecdir ${prefix}/lib
}
```

julian-klode:
@msteveb What I was trying to say was that if I wanted to do such substitutions of non-Makefile files, like documentation or example configuration files, these files would end up with variables in them. So, any such substitution would have to be done from a Makefile with sed or something. I don't care much about that, though :)

PR #967, opened 2017-11-27
Unused parameters

flatcap:
### Unused

These 13 functions have unused parameters.
I suggest we remove them (and add them back _if_ we ever need them).
They're in separate commits for ease of review; I'll squash them all before merging.

| Commit   | File                 | Function(argument)                    |
| :------- | :------------------- | :------------------------------------ |
| 99a1c25d | curs_main.c          | main_change_folder(flags)             |
| ea00c482 | imap/imap.c          | imap_buffy_check(force)               |
| 58c295e6 | init.c               | mutt_nm_tag_complete(pos)             |
| fcbf897c | mutt/file.c          | mutt_file_lock(path)                  |
| 1b4ee25b | mutt/file.c          | mutt_file_unlock(path)                |
| 2c0e95df | ncrypt/crypt_gpgme.c | verify_sender(protocol)               |
| 31dbecf9 | recvcmd.c            | attach_forward_bodies(flags)          |
| e504ea0c | recvcmd.c            | attach_forward_msgs(hdr)              |
| 6aab5181 | recvcmd.c            | attach_include_reply(flags)           |
| 460cfff0 | recvcmd.c            | mutt_attach_bounce(hdr)               |
| 32ce5a51 | recvcmd.c            | mutt_attach_resend(hdr)               |
| b5038b6b | send.c               | mutt_make_misc_reply_headers(hdr,ctx) |
| 13df4039 | sendlib.c            | encode_8bit(istext)                   |

### Macro

I'd like to add a macro to wrap all the other unused parameters.
They are common in various APIs: string formatting, hcache, etc.
With the macro in place, we can enable the compiler option: `-Wunused-parameter` to catch any new unused parameters.

```c
#define UNUSED(x) UNUSED_ ## x __attribute__((__unused__))
```

### DIFF
--- a/attach.h
+++ b/attach.h
@@ -87,8 +87,7 @@ void mutt_pipe_attachment_list(struct AttachCtx *actx, FILE *fp, bool tag,
 void mutt_print_attachment_list(struct AttachCtx *actx, FILE *fp, bool tag,
                                 struct Body *top);
 
-void mutt_attach_bounce(FILE *fp, struct Header *hdr, struct AttachCtx *actx,
-                        struct Body *cur);
+void mutt_attach_bounce(FILE *fp, struct AttachCtx *actx, struct Body *cur);
 void mutt_attach_resend(FILE *fp, struct Header *hdr, struct AttachCtx *actx,
                         struct Body *cur);
 void mutt_attach_forward(FILE *fp, struct Header *hdr, struct AttachCtx *actx,
--- a/attach.h
+++ b/attach.h
@@ -88,8 +88,7 @@ void mutt_print_attachment_list(struct AttachCtx *actx, FILE *fp, bool tag,
                                 struct Body *top);
 
 void mutt_attach_bounce(FILE *fp, struct AttachCtx *actx, struct Body *cur);
-void mutt_attach_resend(FILE *fp, struct Header *hdr, struct AttachCtx *actx,
-                        struct Body *cur);
+void mutt_attach_resend(FILE *fp, struct AttachCtx *actx, struct Body *cur);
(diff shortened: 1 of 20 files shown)


toogley:
> Macro
> 
> I'd like to add a macro to wrap all the **other** unused parameters.
> They are common in various APIs: string formatting, hcache, etc.
> With the macro in place, we can enable the compiler option: -Wunused-parameter to catch >any new unused parameters.
> 
> #define UNUSED(x) UNUSED_ ## x __attribute__((__unused__))

I have trouble to understand that.

My understanding so far: At the moment, those paramers are not used. GCC will higlight it if we run it with `-Wunused-parameter`.  Removing those parameters would probably mean that we have to re-add it in a few months/years, because we need that parameter then.  Therefore, we **don't** want to remove those unused parameters?

Do i understand that correctly?  In short, why don't we remove those other unused paramers, given that simpler code would make refactoring and understanding the code easier?

flatcap:
The usual place to find the unused parameters is in API functions.
The Mailbox API (MxOps, mx.h) defines a `msg_open` function which all the mailboxes must implement.

- `int msg_open(struct Context *ctx, struct Message *msg, int msgno);`

However the `mbox` backend doesn't need the `msgno` parameter.

Turning on the gcc option will give us a false-positive for this function.
Adding `attribute((unused))` to that parameter would stop the warnings from the compiler.

You're right about the others.
Any other functions with unused parameters should have them removed.
I haven't checked recently.  There are probably a few to find.

There's one final type of parameter-redundancy: always-X. These can be hard to find.
I've found some functions where a parameter is always called with the same value.
With a bit of refactoring the code can be simplified.

toogley:
@flatcap aah, thanks for explaining.

PR #968, opened 2017-11-28
Add UNUSED macro for parameters

flatcap:
OK, I didn't hear any (negative) feedback on my suggestion, so I've created a PR.
It's not _quite_ as clean as I'd liked.
There are 119 parameters marked as unused, but there are a further 16 which are only used conditionally.

What do you think?

- 5165b482 Mark unused parameters
  Add UNUSED macro `__attribute__((__unused__))`

- ad49abd7 Mark conditional parameters
  Add CONDIT macro for parameters in `#ifdef` clauses

- cb671110 Add gcc warning options
  Add two new warning options to the build:
  * `-Wunused-parameter`
  * `-Wshadow`

### DIFF
--- a/addrbook.c
+++ b/addrbook.c
@@ -71,10 +71,11 @@ static const struct Mapping AliasHelp[] = {
  * | \%r     | Address which alias expands to
  * | \%t     | Character which indicates if the alias is tagged for inclusion
  */
-static const char *alias_format_str(char *buf, size_t buflen, size_t col, int cols,
-                                    char op, const char *src, const char *prec,
-                                    const char *if_str, const char *else_str,
-                                    unsigned long data, enum FormatFlag flags)
+static const char *alias_format_str(char *buf, size_t buflen, size_t UNUSED(col),
+                                    int UNUSED(cols), char op, const char *src,
+                                    const char *prec, const char *UNUSED(if_str),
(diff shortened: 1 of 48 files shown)


flatcap:
After a brief chat with @julian-klode it turns out there are three possible options.

- **Prefix**
  `#define UNUSED __attribute__((__unused__))`
  `int function(UNUSED int arg)`
  This looks good.  The `UNUSED` marker looks like a type modifer

- **Postfix**
  `#define UNUSED __attribute__((__unused__))`
  `int function(int arg UNUSED)`
  This looks OK and the `UNUSED` marker could be nicely concealed in vim.

- **Macro**
  `#define UNUSED(x) UNUSED_ ## x __attribute__((__unused__))`
  `int function(int UNUSED(arg))`
  It's visually less appealing, but the macro also renames `arg`.
  This prevents accidental use.

flatcap:
Maybe in the future...
For now this branch has been archived to https://github.com/neomutt/neomutt-old/tree/unused-macro

toogley:
I like the idea. Is there a reason why you closed this PR?

flatcap:
These attributes (wrapped in macros) aren't to everyone's taste, so I dropped the idea.
Meanwhile, I've been marking a lot of them in the Doxygen headers
```c
/**
 * @param size (UNUSED)
 */
```

Issue #928, opened 2017-11-07
Bring consistency among sizes


gahr:
Currently, a mix of `size_t`, `unsigned long`, `unsigned`, and `int` is used in our code base to express sizes, which are usually conveying information about the length of a string or the capacity of a buffer.

Scope of this issue is to perform a single swipe over the code base and convert all sizes to `size_t`.

justinvasel:
I can give this a go.

PR #971, opened 2017-11-29
Bring consistency among sizes

justinvasel:
* **What does this PR do?**

There are instances in the code that refer to lengths or sizes of strings and
buffers and the like, but they are not typed in a consistent manner. The purpose
of this commit is to provide some consistency to these scenarios by changing
their type from `unsigned int`, `int`, `unsigned long`, etc. to `size_t`.

In general, these situations were identified by globally searching the code
for patterns along the lines of:
`(unsigned|unsigned int|int|long) (len|blen|olen|length|l|size)`
and changing the type to `size_t` on a case-by-case basis.

The following situations were explicitly ignored:

- autosetup/jimsh0.c, because it appears to be third-party code

- inputs to some function calls relating to ssl and sasl, because they don't take
`size_t` as inputs and recasting in the function call didn't seem worth the
increased messiness

- for loop indices, because in some cases unsigned indices can breed bugs

This commit takes care of most obvious cases. There may certainly be subtler
ones that were missed in this pass.

Sanity checks are always welcome, but no specific review is requested.

Resolves: #928


gahr:
Before I take a close look, I’d like to bring to your attention the fact that in some (many) cases signed integers are currently checked for > 0. This won’t work for size_t. So, have you also checked that your changes take that into account?

justinvasel:
For signed integer to `size_t` conversions, I double checked that those variables would not potentially hold a negative value. Is that what you mean?

For example: some functions return the size of an input or -1 if unsuccessful. I would avoid switching to `size_t` in such a case.

conn/ssl.c
```diff
@@ -740,7 +740,7 @@ static int tls_close(struct Connection *conn)
 static bool check_certificate_cache(X509 *peercert)
 {
   unsigned char peermd[EVP_MAX_MD_SIZE];
-  unsigned int peermdlen;
+  size_t peermdlen;
```
flatcap: `X509_digest()` is expecting `unsigned int *`
When might `size_t` and `unsigned int` differ in size?

conn/ssl.c
```diff
@@ -769,7 +769,7 @@ static bool check_certificate_cache(X509 *peercert)
 static int check_certificate_file(X509 *peercert)
 {
   unsigned char peermd[EVP_MAX_MD_SIZE];
-  unsigned int peermdlen;
+  size_t peermdlen;
```
flatcap: `X509_digest()` again

pop.c
```diff
@@ -83,7 +83,7 @@ static int pop_read_header(struct PopData *pop_data, struct Header *h)
 {
   FILE *f = NULL;
   int rc, index;
-  long length;
+  size_t length;
```
flatcap: If we're using `size_t` here, we probably want to use `%zd` in the `sscanf()` below.

compress.c
```diff
@@ -684,7 +684,7 @@ static int comp_check_mailbox(struct Context *ctx, int *index_hint)
   if (!ops)
     return -1;
 
-  int size = get_size(ctx->realpath);
+  size_t size = get_size(ctx->realpath);
```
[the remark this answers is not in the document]
justinvasel: This is a good point. The size returned by `get_size()` comes from the `stat` struct, which uses the signed `off_t` type. I've reverted this back to `int`.

conn/ssl.c
justinvasel: As far as I can tell, there seems to be no risk in going from an `unsigned int` to `size_t`. Going the other way would be riskier as `unsigned int` is more platform-dependent in its size.

pop.c
justinvasel: Good catch. I changed it to `%zu` since `size_t` is inherently unsigned.

Issue #813, opened 2017-10-02
Sidebar display multiple (but not all) components.


lrosenman:
I have mailboxes in a hierarchy in my IMAP server:
e.g. (not the full list):
```
lists
lists/neomutt
lists/neomutt/devel
lists/neomutt/users
lists/noaa
lists/noaa/changes
lists/waze
lists/waze/beta
lists/exim
lists/exim/users
lists/exim/announce
lists/exim/dev
lists/mailman
lists/mailman/announce
lists/mailman/users
lists/openssl
lists/openssl/announce
lists/openssl/users
lists/mutt
lists/mutt/users
lists/cone
lists/lopsa-us-tx-austin
lists/ietf
lists/ietf/sieve
lists/pgsql
lists/pgsql/novice
lists/pgsql/docs
lists/pgsql/announce
lists/pgsql/bugs
lists/pgsql/hackers
lists/pgsql/general
lists/pgsql/committers
lists/pgsql/sql
lists/pgsql/advocacy
lists/pgsql/jdbc
lists/pgsql/odbc
lists/pgsql/www
lists/pgsql/jobs
lists/SprintBuzz
lists/spi
lists/spi/private
lists/spi/general
lists/SpamAssassin
lists/SpamAssassin/users
lists/SpamAssassin/commits
lists/pfsense
lists/pfsense/users
lists/pfsense/security
lists/isoc
lists/isoc/pubsoft
lists/isoc/memberpubpol
lists/surbl
lists/surbl/discuss
lists/php
lists/php/announce
lists/freebsd
lists/freebsd/ports-developers
lists/freebsd/doc-svn
lists/freebsd/developers
lists/freebsd/ports-commiters
lists/freebsd/jobs
lists/freebsd/rc
lists/freebsd/mobile
lists/freebsd/x11
lists/freebsd/ports-bugs
lists/freebsd/amd64
lists/freebsd/advocacy
lists/freebsd/virtualization
lists/freebsd/ports-announce
lists/freebsd/wireless
lists/freebsd/arch
lists/freebsd/scsi
lists/freebsd/arm
lists/freebsd/net
lists/freebsd/fs
lists/freebsd/proliant
lists/freebsd/stable
lists/freebsd/cvs-all
lists/freebsd/announce
lists/freebsd/current
lists/freebsd/security
lists/freebsd/java
lists/freebsd/usb
lists/freebsd/cloud
lists/freebsd/ports
lists/freebsd/acpi
lists/freebsd/hackers
[...]

PR #973, opened 2017-11-30
Devel/sidebar strip component

justinvasel:
* **What does this PR do?**

In cases of a complex maildir directory structures, one may want to hide the
highest `N` directories from display in the sidebar to keep things looking
clean and legible.

This PR introduces a new config parameter, `sidebar_component_depth`, which
takes a positive integer as its value.

For example, suppose I have the following directory structure:
```
Inbox
dir1/dir2
dir1/dir2/dir3/sushi
dir1/dir2/dir3/neomutt/developer
```

By setting `sidebar_component_depth=2` in one's rc file, the sidebar will
display these maildirs in the following way:
```
Inbox
dir2
dir3/sushi
dir3/neomutt/developer
```

As seen in this example, the two highest directories have been truncated from
each path except in the case when there are `<= N` subdirectories, in which
case the bottom-most directory is displayed.

This parameter has a default value of 0, which disables the feature.

An existing option, `sidebar_short_path`, provides similar behavior,
but truncates all except the bottom-most directories. When that option is
enabled, `sidebar_component_depth` is ignored.

I think I made the relevant documentation changes, but it's possible I missed something. I'm new here so sanity checks are appreciated.

Resolves: #813

### DIFF
--- a/sidebar.c
+++ b/sidebar.c
@@ -900,6 +900,18 @@ static void draw_sidebar(int num_rows, int num_cols, int div_width)
         }
       }
     }
+    else if ((SidebarComponentDepth > 0) && SidebarDelimChars)
+    {
+      sidebar_folder_name = b->path + maildir_is_prefix * (maildirlen + 1);
+      for (int i = 0; i < SidebarComponentDepth; i++)
+      {
+        char *chars_after_delim = strpbrk(sidebar_folder_name, SidebarDelimChars);
+        if (!chars_after_delim)
+          break;
+        else
+          sidebar_folder_name = chars_after_delim + 1;
+      }
+    }
     else
       sidebar_folder_name = b->path + maildir_is_prefix * (maildirlen + 1);
 
--- a/doc/manual.xml.head
+++ b/doc/manual.xml.head
@@ -748,11 +748,12 @@ mailboxes =water/ocean/atlantic =water/ocean/pacific   =water/ocean/arctic
 </screen>
               <para>Shorten the names:</para>
               <screen>
-set sidebar_short_path                  <emphasis role="comment"># Shorten mailbox names</emphasis>
-set sidebar_delim_chars="/"             <emphasis role="comment"># Delete everything up to the last / character</emphasis>
+set sidebar_short_path                  <emphasis role="comment"># Shorten mailbox names (truncate all subdirs)</emphasis>
+set sidebar_component_depth=1           <emphasis role="comment"># Shorten mailbox names (truncate 1 subdirs)</emphasis>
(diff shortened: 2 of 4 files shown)


flatcap:
Not bad! (my first test failed :-)

```
mailboxes "=neo"
mailboxes "=neo/devel"
mailboxes "=neo/users"
mailboxes "=mutt"
mailboxes "=mutt/devel"
mailboxes "=mutt/users"
```

with `sidebar_component_depth=1` displays:

```
eo
github
devel
users
utt
devel
users
```

Folders without any delimiters are being beheaded.

justinvasel:
Thanks for finding that. I'll take a look and make it a bit smarter.

justinvasel:
Added a fix to handle the case when no delimiter characters are present.

flatcap:
OK, this works nicely, but the code...  I think it could be improved.
You iterate through all the characters of the path looking matching them against `SidebarDelimChars` using `strchr()`.  It might be better to use `strpbrk()` and loop through `SidebarComponentDepth` times.

justinvasel:
Modified to use `strpbrk()` in place of `strchr()`.

lrosenman:
What if I want to always display the *LAST* 2 components?

sidebar.c
```diff
@@ -900,6 +900,27 @@ static void draw_sidebar(int num_rows, int num_cols, int div_width)
         }
       }
     }
+    else if (SidebarComponentDepth > 0)
+    {
+      sidebar_folder_name = b->path + maildir_is_prefix * (maildirlen + 1);
+      for (int i = 0; i < SidebarComponentDepth; i++)
+      {
+        /* if sidebar_delim_chars was set as an empty string, break to avoid seg fault */
+        if (!SidebarDelimChars)
```
flatcap: I'd combine this check with the `else if (SidebarComponentDepth > 0)`

sidebar.c
```diff
@@ -900,6 +900,27 @@ static void draw_sidebar(int num_rows, int num_cols, int div_width)
...
+
+        char* chars_after_delim = strpbrk(sidebar_folder_name, SidebarDelimChars);
+        if (!chars_after_delim)
+        {
+          break;
+        }
+        else
+        {
+          sidebar_folder_name = chars_after_delim;
+          sidebar_folder_name += 1;
```
flatcap: These can be combined `= cad + 1`
Then you can (optionally) drop the `{}`s

justinvasel:
Thanks, @flatcap. Good suggestions; it looks much cleaner now. They've been checked in.

PR #979, opened 2017-12-03
restore folder settings after folder-hook

### LINKED ISSUE #977: Folder-hooks for outgoing folder are run upon sending mail, but after return to index, hooks for original folder are not.
*  Actual behavior
When an email is sent (with "record" set), folder-hooks for the outoing mail folder are executed, but upon return to the index of the currently viewed folder, neither the default folder-hooks nor the folder-hooks for the currently viewed folder are executed.

When the email is not sent, but postponed instead, the outoing folder-hooks do not seem to be executed (or at least nothing +INBOX-related seems to be changed afterwards).

    *  Expected behavior
From a user's point of view, I expect that sending an email doesn't change anything regarding the currently viewed folder.

The manual states:

> It is often desirable to change settings based on which mailbox you are
> **reading**. The folder-hook command provides a method by which you can execute any
> configuration command. regex is a regular expression specifying in which
> mailboxes to execute command before **loading**.

So, if folder-hooks are supposed to be just for _reading_, then don't execute the outgoing folder-hooks at all when sending mail. If they are supposed to be for _loading_, then running the outgoing folder-hooks when sending mail can at least be considered, but then the default folder-hooks as well as the folder-hooks for the currently viewed folder need to be executed afterwards.

    * Steps to reproduce

As an example, define different folder hooks for +INBOX$ and +out$, and set record to "+out":

```
set record="+out"
folder-hook +INBOX$ "set sort=threads"
folder-hook +out$   "set sort=reverse-date"
```
_(See attachment for full .muttrc, sensitive information replaced by "XXX")_

- Start `mutt -f+INBOX`
- Check in the status line that the sort order is "threads"
- Compose a mail, and send it (default key "y")
- mutt will return to the index for +INBOX
- The status line now shows that the sort order is "reverse-date"

```
BUG in markdown?
This code block seems to be needed, as otherwise,
the items below appear as subitems of the list above.
```

    *  Used program versions
neomutt 20171027+dfsg.1-1

    *  Operating System and its version
Debian testing (amd64)

**Attachment**
[muttrc.txt](https://github.com/neomutt/neomutt/files/1523330/muttrc.txt)

_{ EDIT 2017-12-02: Minor typo }_

flatcap:
Issue #866 (4aaf1ef7) fixed a problem where the folder-hook wasn't being called for the 'fcc' (sent mail folder).

If we run the folder-hook for the fcc folder, we must run the folder-hook for the current folder afterwards.

Fixes #977

**Notes**:
- `mutt_write_fcc()` doesn't have a pointer to the context, so the global `Context` is used
- `mutt_write_fcc()` can be called multiple times by `mutt_write_multiple_fcc()`.
  This means the folder-hook will be called more than necessary: fcc1, cur dir, fcc2, cur dir, ... 
- If the restore was outside of `mutt_write_fcc()` it would need to be in several places.

### DIFF
--- a/init.c
+++ b/init.c
@@ -133,6 +133,36 @@ static char **nm_tags;
 
 extern char **envlist;
 
+#ifndef DOMAIN
+/**
+ * getmailname - Try to retrieve the FQDN from mailname files
+ * @retval ptr Heap allocated string with the FQDN
+ * @retval NULL if no valid mailname file could be read
+ */
+static char *getmailname(void)
+{
+  char *mailname = NULL;
+  static const char *mn_files[] = { "/etc/mailname", "/etc/mail/mailname" };
+
+  for (size_t i = 0; i < mutt_array_size(mn_files); i++)
+  {
+    FILE *f = mutt_file_fopen(mn_files[i], "r");
+    if (!f)
+      continue;
+
+    size_t len = 0;
+    mailname = mutt_file_read_line(NULL, &len, f, NULL, 0);
+    mutt_file_fclose(&f);
+    if (mailname && *mailname)
+      break;
+
+    FREE(&mailname);
+  }
+
+  return mailname;
+}
+#endif
+
 static void toggle_quadoption(int opt)
 {
   int n = opt / 4;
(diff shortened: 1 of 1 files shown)


toogley:
@flatcap what was the reason for closing this?

flatcap:
@toogley 
The fix got merged (7d44082d).
I tend to merge from the command line.
Occasionally, I push things too quickly and github doesn't keep up.

PR #988, opened 2017-12-10
man: "mutt" --> "neomutt"

toogley:
* **What does this PR do?**

renaming mutt-related things in the manpage to the corresponding neomutt things. (including website and mailinglist).

No.

### DIFF
--- a/doc/neomutt.man
+++ b/doc/neomutt.man
@@ -89,7 +89,7 @@ Specify a blind-carbon-copy (BCC) recipient
 .IP "-c \fIaddress\fP"
 Specify a carbon-copy (CC) recipient
 .IP "-d \fIlevel\fP"
-If neomutt was compiled with +DEBUG log debugging output to ~/.neoneomuttdebug0.
+log debugging output to ~/.neomuttdebug0.
 \fILevel\fP can range from 1-5 and effects verbosity. A value of 2 is
 recommended.
 .IP "-D"
@@ -209,7 +209,8 @@ The help line for the index menu is not updated if you change the bindings
 for one of the functions listed while Neomutt is running.
 .PP
 For a more up-to-date list of bugs, errm, fleas, please visit the
-neomutt project's bug tracking system under http://bugs.neomutt.org/.
+neomutt project's bug tracking system under https://github.com/neomutt/neomutt/issues.
+Keep in mind that some very hard to solve bugs are closed for now.
 .SH NO WARRANTIES
 This program is distributed in the hope that it will be useful,
 but WITHOUT ANY WARRANTY; without even the implied warranty of
--- a/doc/neomutt.man
+++ b/doc/neomutt.man
@@ -17,16 +17,16 @@
 .\"     along with this program; if not, write to the Free Software
 .\"     Foundation, Inc., 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301, USA.
 .\"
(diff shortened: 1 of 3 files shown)


doc/neomutt.man
```diff
@@ -181,12 +181,12 @@ Default Reply-To address.
 Specifies the editor to use when composing messages.
 .SH FILES
 .PP
-.IP "~/.muttrc or ~/.mutt/muttrc"
+.IP "~/.neomuttrc or ~/.neomutt/neomuttrc"
 User configuration file.
-.IP "@sysconfdir@/Muttrc"
+.IP "@sysconfdir@/Neomuttrc"
```
toogley: I guess it makes sense to lower-case that?

doc/neomutt.man
```diff
@@ -209,7 +209,8 @@ The help line for the index menu is not updated if you change the bindings
 for one of the functions listed while Neomutt is running.
 .PP
 For a more up-to-date list of bugs, errm, fleas, please visit the
-neomutt project's bug tracking system under http://bugs.neomutt.org/.
+neomutt project's bug tracking system under https://github.com/neomutt/neomutt/issues.
+Keep in mind that some very hard to solve bugs are closed for now.
```
nicolasbock: I don't think the last sentence is adding a lot of information and has a bit of a negative ring to it. We should encourage user contributions even in the form of new issues, and the qualifier "some very hard to solve bugs" is nebulous at best. What's "very hard to solve" is something that needs to be determined on a case by case basis and hopefully commented on in the the issue itself. I would prefer if the the sentence was removed.

toogley: Hm. Well. @flatcap likes to keep the list of issues short and "solvable". That means, relevant issues are closed and maybe reopened some day. That means, i want to make sure users don't think of the open bugs as "all known issues", as that leaves the hard to solve out.

Keeping that in mind, do you have a better idea for expressing that?

mschilli87: @toogley: You could link both, https://github.com/neomutt/neomutt/issues and https://github.com/neomutt/neomutt/issues?q=is%3Aissue+label%3Adifficulty%3Ahard.

Once https://github.com/isaacs/github/issues/660 is addressed, one could go for sth like https://github.com/neomutt/neomutt/issues?q=is%3Aissue+is%3Aopen+OR+label%3Adifficulty%3Ahard.

For clarity, maybe a new label (wontfix/postponed/?) would be even more clear than just 'difficulty:hard'.

toogley: @mschilli87 Hm, thanks :) So, @flatcap I'd like to have your opinion here.

flatcap:
Thanks, @toogley.  This is long overdue (read: I've been too lazy to do it myself).

> @flatcap likes to keep the list of issues short and "solvable"

Beyond a 100 issues/PRs it gets quite depressing, so I feel the need to cull them.
This helps to keep developers focussed on the issues I'd like to see fixed.
We don't have a roadmap; a lot depends on what developers are prepared to work ok.

This culling means plenty of enhancement requests get closed.
Some of them are great ideas, but we don't have enough developers to work in them.
(I keep notes of **ALL** ideas)

I don't close bugs, except in extreme cases like "the colours don't work".

> @mschilli87: For clarity, maybe a new label (wontfix/postponed/?)

Yes, that's a better idea.
I'll have a think about the wording, then retro-tag all the issues in the "closed discussions" list.
(and probably drop the list)

PR #989, opened 2017-12-10
[WIP/RFC] remove doc folder, as docs have their own repo

toogley:
* **What does this PR do?**

    * remove the doc folder, since they will live in https://github.com/neomutt/neomutt-docs
        * depends on finishing https://github.com/neomutt/neomutt-docs/pull/12
    * (to-be-done):  update Makefiles of this repo, so that docs from https://github.com/neomutt/neomutt-docs are included in the main build. That includes, that e.g. manpages are correctly distributed.
    * What else needs to be done before this PR can be merged?


toogley:
This PR might have been useful initially (if i worked a lot on those things), however now it certainly isn't useful anymore. Therefore i close it, sorry for the noise.

PR #771, opened 2017-09-21
build: add -lidn when idn is enabled

andir:
**What does this PR do?**
Add a missing `-lidn` when building with `idn` support to fix neomutt on NixOS.

Neomutt didn't have access to `libidn.so` (during runtime) since it was not specified during linking.

It seems that https://github.com/neomutt/neomutt/commit/abfe0f14d2c98d17a2ce39e14a377782e1aa732d#diff-67e997bcfdac55191033d57a16d1408aL887 introduced that issue and only appeared when I did run the 20170912 release on my NixOS machine.

PR #992, opened 2017-12-11
build: prevent omission of adding -lidn to linker flags when notmuch is enabled

toogley:
* **What does this PR do?**
    * see https://github.com/neomutt/neomutt/pull/771 - which fixed this problem already for autotools
    * enables removal of [this line](https://github.com/NixOS/nixpkgs/blob/f16bd1c72d32bba164d5bee395e576999c4647e3/pkgs/applications/networking/mailreaders/neomutt/default.nix#L73) for neomutt's autosetup build.

    * https://github.com/neomutt/neomutt/pull/771
    * https://github.com/NixOS/nixpkgs/pull/32546

### DIFF
--- a/auto.def
+++ b/auto.def
@@ -105,7 +105,7 @@ if {1} {
   # relative --enable-opt to true. This allows "--with-opt=/usr" to be used as
   # a shortcut for "--opt --with-opt=/usr".
   foreach opt {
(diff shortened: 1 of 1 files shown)


toogley:
closing, as the described problem doesn't occur without this PR.

PR #1002, opened 2017-12-14
Add spaces after "References:" and "In-Reply-To:"

jwilk:


flatcap:
I changed your fix slightly altering `mutt_write_references()`.
The bug was introduced there when the function was refactored.

Issue #1001, opened 2017-12-14
optional construct acts weirdly when it sees ">"


cyphar:
####  Expected behavior ####

If you try doing something like

```
set status_format="%?d?%d deleted >abc?"
```

You get a status line of 

```
>abc
```

If "%d" is zero. This also happens if you put this construct anywhere in a statusline as far as I can tell (and I believe it also happens in other situations where you have optional formatting).

#### Actual behavior ####

One would expect an empty staus line.

####  Steps to reproduce ####

See description.

<details>
<summary><b>System Information</b></summary>

```
% mutt -v
NeoMutt 20170912 (1.9.0)
Copyright (C) 1996-2016 Michael R. Elkins and others.
Mutt comes with ABSOLUTELY NO WARRANTY; for details type `mutt -vv'.
Mutt is free software, and you are welcome to redistribute it
under certain conditions; type `mutt -vv' for details.

System: Linux 4.14.2-1-default (x86_64)
ncurses: ncurses 6.0.20171118 (compiled with 6.0)
libidn: 1.33 (compiled with 1.33)
hcache backends: kyotocabinet, gdbm

Compiler:
Using built-in specs.
Reading specs from /usr/lib64/gcc/x86_64-suse-linux/7/defaults.spec
COLLECT_GCC=gcc
COLLECT_LTO_WRAPPER=/usr/lib64/gcc/x86_64-suse-linux/7/lto-wrapper
OFFLOAD_TARGET_NAMES=hsa:nvptx-none
Target: x86_64-suse-linux
[...]

PR #1009, opened 2017-12-17
nested-if: correctly handle "<" and ">" with %?

cyphar:
As part of the implementation of [nested-if][1], we translate old-style
conditional expandos into nested-if expandos. However, this translation
did not account for the metacharacters that need to be escaped in
nested-if expandos that are valid in old-style expandos (in particular
"<" and ">"). Correct this by escaping the relevant metacharacters.

With this patch, it is now possible to do something like

  :set status_format=" %rMail%r >%?u? +%u >?"

Without causing an ">" to be included if %u is 0 (which used to be the
case).

[1]: https://www.neomutt.org/feature/nested-if

Fixes: neomutt/neomutt#1001
Signed-off-by: Aleksa Sarai <cyphar@cyphar.com>


flatcap:
Thanks @cyphar, nice job.  Merged in 39a488df9

PR #1010, opened 2017-12-17
pgppubring: fix signed and unsigned comparison

jelly:
Fix size_t and int comparison in pgppubring.

* **What does this PR do?**

Fix one -Wsign-compare compiler warning


flatcap:
Thanks @jelly.  Merged in b931a3b15

Issue #998, opened 2017-12-14
[Regression] Cannot send mail with PGP sign/encrypt; fails with `gpg: conflicting commands`


protist:
* Expected behavior: sign and encrypt outgoing mail.
* Actual behavior: sending fails. After attempting to `send-message` from the `Compose` window, the error message is shown: `gpg: conflicting commands`.
* Steps to reproduce: sign and encrypt outgoing mail (PGP/MIME), then attempt to send.
* Used program versions: neomutt 20171208-1 (bug not present in 20171013-2)
* Operating System and its version: Arch Linux (up-to-date)

Additional details
-----
The following security modes fail:
* Sign, Encrypt (PGP/MIME) (OppEnc mode)
* Sign, Encrypt (PGP/MIME)

The following security modes work:
* Encrypt (PGP/MIME) (OppEnc mode)
* Encrypt (PGP/MIME)
* Sign (PGP/MIME)

PGP, so possibly related to bug #991?

gahr:
Are you using GPGME? If not, could the changes introduced by #949 to the gpg.rc file be related?

protist:
I'm not 100% if I'm using gpgme, but I do have it installed. I also have the following set:
```
pgp_use_gpg_agent = yes
source /usr/share/doc/neomutt/samples/gpg.rc
```

I tried reverting the `gpg.rc` file to the version from neomutt 20171013-2, and that fixes the problem. I also tried reverting the following line alone, which also worked.
```
--- gpg.rc.factory_default	2017-12-14 21:13:47.907719361 +1100
+++ gpg.rc	2017-12-14 21:20:37.208184794 +1100
@@ -68,7 +68,7 @@
 # 
 # set pgp_encrypt_sign_command="/usr/lib/neomutt/pgpewrap gpg-2comp %?p?--passphrase-fd 0? -v --batch --output - --encrypt --sign %?a?-u %a? --armor --always-trust -- -r %r -- %f"
 #
-set pgp_encrypt_sign_command="/usr/lib/neomutt/pgpewrap gpg %?p?--pinentry-mode loopback --passphrase-fd 0? --batch --quiet --no-verbose --textmode --output - %?a?--local-user %a? --armor -- --recipient %r -- --sign --encrypt %f"
[...]

PR #1023, opened 2018-01-16
fix for pgpewrap in default gpg.rc

makoshark:
I believe this fix an bug introduced by changes introduced in Commit 465fad6d21ac3b9d2ae5d12bae7705f5f04323e7 which makes a change to `contrib/gpg.rc` that breaks signed and encrypted mail. The issues is just a typo related to way that `pgpewrap` works. This should fix the bug discussed in issue #998 and seems to do so on my system.

**Please review this carefully.** This is my first commit to mutt/neomutt and I know nothing about `pgpewrap` other than I what I just figured by experimenting with it on the command line. Since this i just a change to a configuration, I've not tested this more carefully than running the new configuration in my version of neomutt (the most recent version in debian).

### DIFF
--- a/contrib/gpg.rc
+++ b/contrib/gpg.rc
@@ -62,13 +62,13 @@ set pgp_clearsign_command="gpg %?p?--pinentry-mode loopback --passphrase-fd 0? -
 #
 # set pgp_encrypt_only_command="/usr/lib/neomutt/pgpewrap gpg-2comp -v --batch --output - --encrypt --textmode --armor --always-trust -- -r %r -- %f"
 #
-set pgp_encrypt_only_command="/usr/lib/neomutt/pgpewrap gpg --batch --quiet --no-verbose --output - --textmode --armor -- --recipient %r -- --encrypt %f"
+set pgp_encrypt_only_command="/usr/lib/neomutt/pgpewrap gpg --batch --quiet --no-verbose --output - --textmode --armor --encrypt -- --recipient %r -- %f"
 
 # Create an encrypted and signed attachment (note that some users include the --always-trust option here)
 # 
 # set pgp_encrypt_sign_command="/usr/lib/neomutt/pgpewrap gpg-2comp %?p?--passphrase-fd 0? -v --batch --output - --encrypt --sign %?a?-u %a? --armor --always-trust -- -r %r -- %f"
 #
-set pgp_encrypt_sign_command="/usr/lib/neomutt/pgpewrap gpg %?p?--pinentry-mode loopback --passphrase-fd 0? --batch --quiet --no-verbose --textmode --output - %?a?--local-user %a? --armor -- --recipient %r -- --sign --encrypt %f"
+set pgp_encrypt_sign_command="/usr/lib/neomutt/pgpewrap gpg %?p?--pinentry-mode loopback --passphrase-fd 0? --batch --quiet --no-verbose --textmode --output - %?a?--local-user %a? --armor --sign --encrypt -- --recipient %r -- %f"
 
 # Import a key into the public key ring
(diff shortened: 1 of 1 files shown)


gahr:
@libBletchley thoughts?

a3nm:
I don't know anything about neomutt/pgpewrap either, but this fixes the problem on my system too, and it looks reasonable to me based on my understanding of https://www.neomutt.org/man/pgpewrap.

makoshark:
`pgpgewrap` is merely doing command line munging and is trivial to test by replacing `gpg` with `echo` in the commands in question.

If you do so, it's easy to see why the changed `gpg.rc` causes errors which will keep anybody using the latest version of neomutt to be unable to sign and encrypt and it's easy to see why this PR fixes it.

libBletchley:
I just commented here:

https://github.com/neomutt/neomutt/issues/998#issuecomment-358935662

I think there needs to be a bug reported against gpg if in fact the syntax-conforming commandline fails and a syntax-violating commandline succeeds.  It's possible that the gpg code is doing the correct thing, and that the GPG man page needs an update to clarify what "[args]" is, because @makoshark's fix supplies an "option" as an argument.  Since `--recipient` is not by nature _optional_, it's perhaps supposed to be an argument to the `--encrypt` command.

Note that there is some sloppiness with the gpg manpage.  E.g. it says there can only be _one_ command, but obviously `--encrypt --sign` is an exception to that.

libBletchley:
For the record, there is still a bug in pgpwrap:
```
$ pgpewrap /bin/echo command arg1 arg2 -- -o opt1 opt2 -- foo bar
command arg1 arg2 -o opt1 -o opt2 foo bar
$ pgpewrap /bin/echo command arg1 arg2 -- -o opt1 opt2 -- foo bar baz
command arg1 arg2 -o opt1 -o opt2 foo bar foo baz
```
Notice `foo` is duplicated on the second instance.

flatcap:
No, that's by design.
NeoMutt can't generate command lines for multiple recipients in a generic way.
gpg expects something like `-r Jim -r Bob`
`pgpewrap` transforms the last params (after `--`) from
`FLAG ARG1 ARG2 ARG3 ARG4`
to
`FLAG ARG1 FLAG ARG2 FLAG ARG3 FLAG ARG4`

This is so that the simple format string can be transformed for the user

`pgpewarp` is an ugly workaround that we'd love to get rid of.

libBletchley:
@flatcap 
That doesn't explain why the arguments *after* the *second* `--` delimiter are getting duplicated.  It's also inconsistent.  Why would there be repetition if there are 3 tokens, but not 2?

flatcap:
It does behave _consistently_, but not _obviously_ :-)

Considering the options after the last `--` ...
The first is the option flag.  This is repeated before all the others
The second, [third, fourth...] are parameters to go with the option flag

| Options                       | Expanded to                       | 
| :---------------------------- | :-------------------------------- |
| -- -r Adam                    | -r Adam                           |
| -- -r Bob Charlie             | -r Bob -r Charlie                 |
| -- -r Dave Eric Fred          | -r Dave -r Eric -r Fred           |
| -- -r George Harry Ian John   | -r George -r Harry -r Ian -r John |

Hopefully that makes sense.

libBletchley:
> Considering the options after the last -- ...
> The first is the option flag. 

After the last `--`, there is no option flag.  The option flag is `-r`, correct?  That's after the **first** occurrence of `--`, and that's being repeated correctly.  The bug is the repetition of tokens that are **not** option flags (in this case, gpg _commands_).  The bug only occurs if there are 3 tokens that are not options, to the right of the last `--`.

Issue #960, opened 2017-11-20
Replace mutt_mktemp() with mkstemp()


nobrowser:
Ultimately all the ways in mutt to create a temporary file come down to `mutt_mktemp_full()` in `muttlib.c`.  The varying part of the name that function creates, not counting totally predictable things like the PID, is a 64 bit random number.  That is less randomness than I have in freaking _shell scripts_, by using the `mktemp(1)` program with a suffix of 12 Xs (roughly 6*12 = 72 bits), as the Turtle Book [1] recommends.
Shouldn't we use the libc interfaces for making temporary files, which are similar to `mktemp(1)` and thus allow for more randomness?

[1] Classic Shell Scripting by Robbins & Beebe, O'Reilly Media 2005.  See pages 274-278.

gahr:
One potential issue is that mktemp is legacy in POSIX 2001 and is not mentioned in 2008. The newer variants mkstemp and mkdtemp do not exist in POSIX-2001.

flatcap:
> mkstemp

Actually, this **does** exist in POSIX-2001 (and 2008).
http://pubs.opengroup.org/onlinepubs/009695399/functions/mkstemp.html

As for why we're not using it... who knows!
There's a lot of filesystem code in NeoMutt that works around problems that probably don't exist any more.

Would you like to make us a proposal?

gahr:
I stand corrected - shall we turn this issue into "switch to using mkstemp" ?

flatcap:
There are nearly 100 calls to `mutt_mktemp()`.
Some of NeoMutt's temporary files are immediately deleted (just an open handle is kept).
There may be more common behaviour that can be refactored.

nobrowser:
Well, hmm.  Is there an assumption that the files will be created in a safe private directory?  The fact that `mutt_mktemp()` just constructs the name and leaves the caller to actually `open()` it suggests that there is, because otherwise we suffer from the classic _symlink-to-passwd_ vulnerability.  OTOH `/tmp` is the ultimate default, and it is not safe this way.  So I am confused.  Either this is not worth doing (we're already safe), or we have a much bigger problem than just the form of the name.

flatcap:
> Is there an assumption that the files will be created in a safe private directory?

Yes.  We'll still create the directory.  We'll just let `mkstemp()` do the 'random' bit.

> suffer from the classic symlink-to-passwd vulnerability

Possibly for the situation where we have a single db file in a cache dir that we don't own.
[...]

PR #1025, opened 2018-01-18
Issue #960: Added new function mutt_mkstemp_full and macro that uses mkstemp()

wcurry:
Also refactored one caller in attach.c.


muttlib.c
```diff
@@ -488,6 +488,30 @@ void mutt_mktemp_full(char *s, size_t slen, const char *prefix,
                strerror(errno), errno);
 }
 
+FILE *mutt_mkstemp_full(char *s, size_t slen, const char *src, int line)
+{
+  int fd;
+  FILE *fp;
+
+  size_t n = snprintf(s, slen, "%s/neomutt-XXXXXX", NONULL(Tmpdir));
```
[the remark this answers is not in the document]
wcurry: I see the following in `init.c`:

`4038   Tmpdir = mutt_str_strdup((p = getenv("TMPDIR")) ? p : "/tmp");`

Are you saying we should check each time a tmp file is created?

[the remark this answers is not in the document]
flatcap: Yes, `Tmpdir` is the code variable that backs `$tmpdir` the NeoMutt config variable.
The user could change it at any time.

If we ignore the possibility of it being empty, for now, and make sure that if we can't create the file that the user gets a message.

The association is made here:
https://github.com/neomutt/neomutt/blob/94903f3227eacfa0ae34517e0ab2167d5da302fb/init.h#L4122-L4129

attach.c
```diff
@@ -180,7 +180,7 @@ int mutt_compose_attachment(struct Body *a)
              * copying the file back */
             fseeko(fp, b->offset, SEEK_SET);
             mutt_free_body(&b);
-            tfp = mutt_mkstemp(tempfile, sizeof(tempfile));
+            tfp = mutt_mkstemp();
```
flatcap: That looks a lot nicer :-)

protos.h
```diff
@@ -192,8 +192,8 @@ void mutt_message_to_7bit(struct Body *a, FILE *fp);
 void mutt_mktemp_full(char *s, size_t slen, const char *prefix, const char *suffix, const char *src, int line);
 #define mutt_mktemp_pfx_sfx(a, b, c, d) mutt_mktemp_full(a, b, c, d, __FILE__, __LINE__)
 #define mutt_mktemp(a, b) mutt_mktemp_pfx_sfx(a, b, "neomutt", NULL)
-FILE *mutt_mkstemp_full(char *s, size_t slen, const char *src, int line);
-#define mutt_mkstemp(a, b) mutt_mkstemp_full(a, b, __FILE__, __LINE__)
+FILE *mutt_mkstemp_full(const char *src, int line);
+#define mutt_mkstemp() mutt_mkstemp_full(__FILE__, __LINE__)
```
flatcap: Change this to:
`#define mutt_mkstemp() mutt_mkstemp_full(__func__, __FILE__, __LINE__)`

muttlib.c
```diff
@@ -488,26 +488,30 @@ void mutt_mktemp_full(char *s, size_t slen, const char *prefix,
                strerror(errno), errno);
 }
 
-FILE *mutt_mkstemp_full(char *s, size_t slen, const char *src, int line)
+FILE *mutt_mkstemp_full(const char *src, int line)
```
flatcap: I think we need the function in here, too (see comment for protos.h)

muttlib.c
```diff
@@ -488,26 +488,30 @@ void mutt_mktemp_full(char *s, size_t slen, const char *prefix,
...
   int fd;
   FILE *fp;
 
-  size_t n = snprintf(s, slen, "%s/neomutt-XXXXXX", NONULL(Tmpdir));
+  int n = snprintf(s, sizeof(s), "%s/neomutt-XXXXXX", NONULL(Tmpdir));
 
   if (n < 0)
-    mutt_debug(1, "%s:%d: snprintf failed! returns \"%s\".\n", src, line, n);
+    mutt_debug_real(__func__, __FILE__, __LINE__, 1,
+                    "%s:%d: snprintf failed! returns \"%s\".\n", src, line, n);
```
flatcap: Hmm...  Sorry, I've made things worse.
How about:
- After the `snprintf()` we `mutt_debug_real()` with the name of the file.
  This means the log shows the caller and the filename.
- The rest of the function can use `mutt_debug()` as before.
  If anything goes wrong, we see the line in `mutt_mkstemp_full()`

gahr:
Do we really need this to be a macro and use func, file and line in the logs? Wouldn’t it be sufficient to check for null at caller side and log there? After all, not many different things can practically go wrong that would prevent a tmp file from being created.

flatcap:
We _could_, but there's one nice feature...
We automatically get a log of the caller -- **who** created **which** temporary file.
That might be useful.

The original function used the caller's file:line in the debug messages.

gahr:
> We automatically get a log of the caller -- who created which temporary file.

Uhm.. it doesn't look like we do. What we get is a log of who *failed* to create a temporary file. Since you probably check for NULL at caller site and log there, we'd end up with duplicated logs. We should try to push logging up the stack, not down the stack.

flatcap:
> Uhm.. it doesn't look like we do.

You're right, it doesn't at the moment -- I'd like it to.
And yes, it's unlikely to fail.

In this refactoring, the caller doesn't know the name of the file any more.
I think this sort of debug log line would be useful:
`[TIMESTAMP] caller_function() created '/tmp/neomutt-XXX'`

flatcap:
@wcurry Sorry if you feel like you're adrift on an ocean of opinion :-)

When this function is done, it'll be moved to the library (`/mutt`).
But... the library is quite new and @gahr and I are still deciding how we'd like it to look.

We're planning to move the (debug) logging from the library to the callers.
However, `mutt_mkstemp()` is a bit of an exception.

Please can you:
- Leave the macro in place
- Add a log line: `[TIMESTAMP] caller_function() created '/tmp/neomutt-XXX'`
- Remove the other log lines (just return)

This function's unlikely to fail and if it does, then `errno` will be set, so the callers can run `mutt_perror()`.

Thanks.

PR #1030, opened 2018-01-21
Fix reply_regexp

rawtaz:
* **What does this PR do?**
Fixes the incorrectly formatted `reply_regexp` in `init.h`, and also adds the Swedish "Sv:" prefix to it.

That things still builds, I haven't set up the entire dev environment as this was such a simple fix. Perhaps also some documentation needs updating, e.g. Doxygen run or similar?

N/A

It should, as I expect it to build just fine. I couldn't find any places in the manual touching on this feature/setting. Doxygen format should be fine, just added some words.

None, AFAIK.

* **Other comments**
Please note that the regex still allows the following subject prefixes which I'd argue are not proper. Since it's the current state of things I was reluctant to remove the matching for these cases, even though I think they're wrong: `Re[n][n]:`, `Re[0]:`

### DIFF
--- a/init.h
+++ b/init.h
@@ -2917,12 +2917,12 @@ struct Option MuttVars[] = {
   ** .pp
   ** Also see $$wrap.
   */
-  { "reply_regexp",     DT_REGEX,   R_INDEX|R_RESORT, UL &ReplyRegexp, UL "^(re([\\[0-9\\]+])*|aw):[ \t]*" },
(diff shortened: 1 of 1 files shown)


rawtaz:
Just noticed the setting I modified is named `reply_regexp` and not `reply_regex`. Sorry, doesn't affect anything practical though.

flatcap:
Yes, the 'regex' vs 'regexp' is a bit of a pain.
You can edit your comment above and if you "git commit --amend" you can fix the commit message.
Then you can "git push --force" to update the PR.

rawtaz:
You're right, better get it done.

PR #1031, opened 2018-01-22
Unit test framework

gahr:
* **What does this PR do?**

This PR is a first step towards having our C library functions unit-tested.

A new binary `neomutt-test` is generated under the `test` directory. This directory contains a `main.c` function which lists the test cases, represented by `void(*)(void)` functions. Test cases are ran through [acutest](https://github.com/mity/acutest), which is a single-header unit-test framework. It provides the scaffolding for running tests, provides a report, and implements the test macros `TEST_CHECK` and `TEST_CHECK_`.

I have already added a couple of test suites for base64 and rfc2047, to give people a sense of how it looks like.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -25,6 +25,7 @@ neomutt
 pgpewrap
 pgpring
 txt2c
+test/neomutt-test
 
 # ./configure
 config.h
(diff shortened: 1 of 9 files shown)


toogley:
I'm curious, how good is the isolation? I guess isolated memory is "indirectly" given, as the compiler complains for memory not being freed and such things. So, are there things we have to be cautious about, to make sure some test doesn't infer with other tests?

(i mean beyond mocking external things like smtp)

gahr:
> I'm curious, how good is the isolation?

Acutest by default forks for each test case.

Issue #1028, opened 2018-01-19
[browser] <change-dir>.. Resolve Parent Directory


bazzinotti:
## bug reports

Using `<change-dir>..` does not resolve the parent directory. Example: it uses `/some/folder/../../another/folder` instead of `/another/folder`.

It is not user friendly.

It might cause adverse behavior. (?)

### Steps to reproduce

* First get into the browser. 
  * Here is one way to get into the browser. Open neomutt. From a folder with >0 messages, press 's' to invoke `<save-message>` and then press '?' to enter the browser.
  * Another way is by attaching a file from the compose menu. Open neomutt. Use the default shortcut 'm' to compose a mail. After exiting your editor, neomutt will be in the compose menu. Use default shortcut 'a' to attach file, then press '?' to enter the browser.

From the browser, use the default shortcut 'c' to invoke `<change-dir>`. Then, press `..<enter>` to access the parent directory. The browser updates to the parent directory, but the directory used by `<change-dir>` is not resolved.

* Example (cwd: `/some/folder`)

With default shortcuts, you would press
`c..<enter>`
You enter the parent directory successfully, but press 'c' again to see the new path is unresolved.

* Expected Directory Output
`/some/`
* Actual Directory Output
`/some/folder/../`

* Used program versions
20171215

* Operating System and its version
Linux 4.14.8-gentoo-r1+ (x86_64)

# Other
`macro browser p "<change-dir>..<enter>" "Goto Parent Directory"`

flatcap:
Thanks for the bug report.

It should be a simple job to insert a few `realpath()`s into the code.
In addition, we could add a new browser function `<go-to-parent>`

bazzinotti:
Dear flatcap, let me give the realpath() a shot :smiley:

I'm new to neomutt, so please be kind.

I've narrowed down the location for `realpath` insertion to https://github.com/neomutt/neomutt/blob/da6c0aadaec4021a584713b7749db864d25d2392/browser.c#L1730-L1754

Particularly L1743-L1744 or L1734-L1736 (to make use of the tmp variable)

I'll add a realpath check around there.

# Symlink Resolution

Neomutt's normal behavior is to not resolve symlinks (tested from the browser). (It also does not show hidden files, and I couldn't see a config setting for that. But it's a little off-topic so I digress).

[...]

PR #1037, opened 2018-01-25
[browser] resolve dir on <change-dir> using realpath()

bazzinotti:
This PR is an ongoing solution to #1028.
Part 2 is not yet completed. It will add a `<goto-parent>` function.

### DIFF
--- a/browser.c
+++ b/browser.c
@@ -1509,16 +1509,22 @@ void mutt_select_file(char *f, size_t flen, int flags, char ***files, int *numfi
             }
             else
 #endif
-                if (examine_directory(menu, &state, LastDir, prefix) == -1)
             {
-              /* try to restore the old values */
-              mutt_str_strfcpy(LastDir, OldLastDir, sizeof(LastDir));
               if (examine_directory(menu, &state, LastDir, prefix) == -1)
               {
-                mutt_str_strfcpy(LastDir, NONULL(HomeDir), sizeof(LastDir));
-                goto bail;
+                /* try to restore the old values */
+                mutt_str_strfcpy(LastDir, OldLastDir, sizeof(LastDir));
+                if (examine_directory(menu, &state, LastDir, prefix) == -1)
+                {
+                  mutt_str_strfcpy(LastDir, NONULL(HomeDir), sizeof(LastDir));
+                  goto bail;
+                }
               }
+              /* resolve paths navigated from GUI */
+              if (mutt_realpath(LastDir) == 0)
+                break;
             }
+
             browser_highlight_default(&state, menu);
             init_menu(&state, menu, title, sizeof(title), buffy);
             if (GotoSwapper[0])
@@ -1686,6 +1692,7 @@ void mutt_select_file(char *f, size_t flen, int flags, char ***files, int *numfi
         break;
 #endif
 
+      case OP_GOTO_PARENT:
       case OP_CHANGE_DIRECTORY:
 
 #ifdef USE_NNTP
@@ -1707,7 +1714,16 @@ void mutt_select_file(char *f, size_t flen, int flags, char ***files, int *numfi
           }
         }
 
-        if (mutt_get_field(_("Chdir to: "), buf, sizeof(buf), MUTT_FILE) == 0 && buf[0])
+        if (i == OP_CHANGE_DIRECTORY)
+        {
+          int ret = mutt_get_field(_("Chdir to: "), buf, sizeof(buf), MUTT_FILE);
+          if (ret != 0)
+            break;
+        }
+        else if (i == OP_GOTO_PARENT)
+          mutt_get_parent_path(buf, buf, sizeof(buf));
+
+        if (buf[0])
         {
           buffy = 0;
           mutt_expand_path(buf, sizeof(buf));
@@ -1735,6 +1751,12 @@ void mutt_select_file(char *f, size_t flen, int flags, char ***files, int *numfi
               mutt_file_concat_path(tmp, LastDir, buf, sizeof(tmp));
               mutt_str_strfcpy(buf, tmp, sizeof(buf));
             }
+            /* Resolve path from <chdir>
+             * Avoids buildup such as /a/b/../../c
+             * Symlinks are always unraveled to keep code simple */
+            if (mutt_realpath(buf) == 0)
+              break;
+
             if (stat(buf, &st) == 0)
             {
               if (S_ISDIR(st.st_mode))
--- a/functions.h
+++ b/functions.h
@@ -472,6 +472,7 @@ const struct Binding OpAlias[] = { /* map: alias */
 /* The file browser */
 const struct Binding OpBrowser[] = { /* map: browser */
   { "change-dir",            OP_CHANGE_DIRECTORY,            "c" },
+  { "goto-parent",           OP_GOTO_PARENT,                 "p" },
   { "display-filename",      OP_BROWSER_TELL,                "@" },
   { "enter-mask",            OP_ENTER_MASK,                  "m" },
   { "sort",                  OP_SORT,                        "o" },
--- a/muttlib.c
+++ b/muttlib.c
@@ -1466,6 +1466,10 @@ void mutt_get_parent_path(char *output, char *path, size_t olen)
     mutt_str_strfcpy(output, path, olen);
     int n = mutt_str_strlen(output);
 
+    /* remove any final trailing '/' */
+    if (output[n - 1] == '/')
+      output[n - 1] = '\0';
+
     /* Remove everything until the next slash */
     for (n--; ((n >= 0) && (output[n] != '/')); n--)
       ;
@@ -1478,6 +1482,26 @@ void mutt_get_parent_path(char *output, char *path, size_t olen)
       output[1] = '\0';
     }
   }
+}
+
+/**
+ * mutt_realpath - resolve path, unraveling symlinks
+ * @param buf Buffer containing path
+ * @retval len String length of resolved path
+ * @retval 0   Error, buf is not overwritten
+ *
+ * Resolve and overwrite the path in buf.
+ *
+ * @note Size of buf should be at least PATH_MAX bytes.
+ */
+size_t mutt_realpath(char *buf)
+{
+  char s[PATH_MAX];
+
+  if (realpath(buf, s) == NULL)
+    return 0;
+
+  return mutt_str_strfcpy(buf, s, PATH_MAX);
 }
 
 char debugfilename[_POSIX_PATH_MAX];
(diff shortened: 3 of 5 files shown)


browser.c
```diff
@@ -1735,6 +1735,21 @@ void mutt_select_file(char *f, size_t flen, int flags, char ***files, int *numfi
               mutt_file_concat_path(tmp, LastDir, buf, sizeof(tmp));
               mutt_str_strfcpy(buf, tmp, sizeof(buf));
             }
+            /* resolve path. Avoids buildup such as /a/b/../../c
+             * Symlinks are always unraveled to keep code simple */
+            {
+              char *real_path = NULL;
```
gahr: how about const?

bazzinotti: Will do

browser.c
```diff
@@ -1735,6 +1735,21 @@ void mutt_select_file(char *f, size_t flen, int flags, char ***files, int *numfi
               mutt_file_concat_path(tmp, LastDir, buf, sizeof(tmp));
               mutt_str_strfcpy(buf, tmp, sizeof(buf));
             }
+            /* resolve path. Avoids buildup such as /a/b/../../c
+             * Symlinks are always unraveled to keep code simple */
+            {
+              const char *real_path = NULL;
+              if ((real_path = realpath(buf, NULL)) == NULL)
```
gahr: Also, I think we're more into style like this, right @flatcap ?
```c
const char *real_path = = realpath(buf, NULL);
if (real_path == NULL)
{
   ...
}
```

flatcap: Yes, please.
This is a simple case, where the code is legible, but this style can lead to some very hard-to-read conditionals.

bazzinotti:
I coded the `<goto-parent>` function in 2 different manners, in search of an elegant solution (not yet pleased personally).

Mutt Functions are accessed via a switch statement, eg `case OP_CHANGE_DIRECTORY:`

The `<goto-parent>` mostly uses the code from `<chdir>` with minor differences assessed by a new `bool goto_parent`

---

The first solution can be summarized with the following:
```
      case OP_GOTO_PARENT:
        {
          bool goto_parent = 1;

          goto _chdir_parent;
 

        case OP_CHANGE_DIRECTORY:
          goto_parent = 0;

        _chdir_parent:
          ...
```

Ew, right! So here's my next attempt:

Moved the `<change-dir>` code into a subroutine that facilitates access to needed vars.

`static int _chdir(char *buf, int bufsize, char *title, int titlesize, char *prefix,
                  struct BrowserState *state, struct Menu *menu, int *buffy,
                  bool goto_parent)`

```
      case OP_GOTO_PARENT:                                                
        {                                                                 
          int ret = _chdir(buf, sizeof(buf), title, sizeof(title), prefix,
                           &state, menu, &buffy, true);                      
          if (ret == 1)                                                   
            goto bail;                                                    
          break;                                                          
        }                                                                 
                                                                          
      case OP_CHANGE_DIRECTORY:                                           
        {                                                                 
          int ret = _chdir(buf, sizeof(buf), title, sizeof(title), prefix,
              &state, menu, &buffy, false);                                   
          if (ret == 1)                                                   
            goto bail;                                                    
          break;                                                          
        }
```

Much better formatting, right?! The only thing that's a drag is having to pass so many variables just to be transparent for the new function as if it was inline.

What are your thoughts?

flatcap:
The switch statements in the browser, the index and the pager are a mess,
so nothing you propose can make them worse :-)

Your second solution looks cleaner, but you're right, that's a lot of parameters.

There is an third possibility:

```c
case OP_CHANGE_DIRECTORY:
case OP_GOTO_PARENT:
  shared(code);
  if (i == OP_CHANGE_DIRECTORY)
  {
    change(dir);
  }
  else
  {
    goto(parent);
  }
  more(shared);
  break;
```

Existing example:
https://github.com/neomutt/neomutt/blob/032aca8048df826ccf349766befe6bd7ac641b53/browser.c#L1899-L1906

browser.c
```diff
@@ -1145,6 +1145,27 @@ void mutt_browser_select_dir(char *f)
...
+ * get_realpath - private helper to call realpath() and do error checking.
+ *
+ * You may pass the same variable to srcdir and dest.
+ *
+ * Returns 1 if real_path failed. dest is untouched
+ * Returns 0 on success. dest has resolved path
+ */
+static int get_realpath(const char *srcdir, char *dest, int destlen)
+{
+  const char *real_path = realpath(srcdir, NULL);
```
gahr: Why not store real_path into dest directly, then?

bazzinotti:
`realpath()` is now impl'd for

* `<chdir>` and the new `<goto-parent>` function
* GUI browser

## Idea

Stub the new `get_realpath` helper/wrapper with a `bool resolve_symlinks` parameter for future customization.

## Testing Phase
* I have tested navigating my local filesystem. Seems to work nicely. Throughout many tests, there was a time or two I thought pressing 'q' pulled up `<chdir>` but I could never cause this to happen at will, so I'm not sure about that.
* I have __not__ yet tested the effects of IMAP browsing. The worst case scenario is that finer-grained positioning of `realpath()` becomes necessary.

I don't have an active neomutt IMAP configuration, so IMAP testing will be postponed until my next dev session. Testers who are willing are welcome to try out this PR and lemme know if navigating IMAP folders is working as per normal.

browser.c
bazzinotti: @gahr I'm allowing `srcdir == dest` and avoiding potential clobbering in `realpath()`

bazzinotti: @gahr I forgot, the pointer returned from `real_path` has to be freed as well, since it's `malloc()`'d by `realpath`

bazzinotti: I guess it's a design decision at that point, but the whole reason I wrote this wrapper was to refactor my identical code to `realpath()`, namely the error code, and working with the `malloc()`'d pointer.

bazzinotti: Heh, I'll try to take more time and hit "comment" less often next time. I still consider myself a student at proper design. My brain still tells me there may be a facet to it that I don't understand completely. So in that regards, I'm listening of my coding "angle" is not coming from where you see fit.

gahr: The pointer returned by realpath needs to be freed only if the second argument is NULL.

gahr: Yes, `srcdir == dest` is an issue for using the second argument. I guess it's gonna be a trade off between ease of use and avoiding a superfluous heap allocation.

bazzinotti:
What do you recommend, gahr?

gahr:
How about

```c
size_t mutt_realpath(char *buf)
{
  char s[PATH_MAX];
  size_t len = 0;
  const char *res = realpath(buf, s);
  if (res != NULL)
  {
    len = strlen(s);
    memcpy(buf, s, len);
  }
  return len;
}
```
  
Or some improved veraiin

gahr:
Incidentally, we could have a version of strcpy that returns the length of the source string..

bazzinotti:
I pushed a new commit based on @gahr's last comment here, and reviewing related discourse between @gahr and @flatcap on IRC this morning (EST). 

> size_t mutt_realpath(char *buf)

* I interpreted the use of the `mutt_` prefix to mean you wanted a new muttlib function, so I "publicized" the function in my latest commit. Was I correct in my interpretation?
* I interpreted the lack of a `buflen` such as `mutt_realpath(char *buf, size_t buflen)` to signify it's been declared unnecessary to ensure `buflen > strlen(s)` in code. I also got this impression from reading your brief discussion in IRC. That is, we are leaving it to the programmer's discretion. Is this correct?

It's been awhile since I've done doxygen style comments, so please look at the comments if you could.

Thanks!

gahr:
Also, now that we have 74922c7, we could further compress that function a bit :)

gahr:
I still think the len argument is not needed, and that `mutt_realpath` should just proxy the contract or POSIX’ `realpath`. In other words, I don’t see a use case for passing any buffers smaller than PATH_MAX, as this could result in an unusable truncated string.

gahr:
Oh, even worse, such truncation would go unnoticed, as there would be no way for client code to know whether the full resolved path fit in the provided buffer.

bazzinotti:
@gahr You're right. My confusion from `mutt_str_strnfcpy` led me to think that I could have added error checking at that point for the programmer, in case they erroneously passed a buffer that was too small (in this case less than the string length returned by `realpath` or PATH_MAX). But, now I see that `strnfcpy` is designed not to do such error checking. It is designed to fill buffers smaller than the max `size` or truncate buffers to that `size` if they are bigger. Though even at that realization I wondered if it would not be suitable to have a warning when the `dlen` (or my prefered term) `dsize` < `size`. something like

```
if (dsize < size)
{
  mutt_debug(2, "mutt_str_strnfcpy: Warning, dsize < size");
}
```

bazzinotti:
Anyways, I'll leave that decision to you. I've already made the PR to at least fix the off-by-1 error in `mutt_str_strnfcpy` and brush up the terminology (in separate commits).

Back to the actual topic at hand: I shall remove the `len` argument from `mutt_realpath`

bazzinotti:
## IMAP

As I suspected, IMAP folders don't play nicely with `mutt_realpath` since they're not real folders on the system. Then again, the pre-existing `<change-dir>` behavior isn't polished for IMAP either. Here's what I learned from 5 minutes of playing around with IMAP in neomutt for the first time.

* `<chdir> imaps://imap.gmail.com/[Gmail]/..` stays in [Gmail] folder, rather than going to root dir.
* in previous example, trailing slash disappears.

At least, my next steps are to refine the positioning of the calls to `mutt_realpath` or add logic so they are not called for IMAP.

Optionally I might attempt to enhance the `<change-dir>` behavior for IMAP, though I consider it beyond my initial goal.

gahr:
I'd leave the error handling to the caller. Other than that, I think this is fine.

bazzinotti:
I updated my previous comment to distinguish the problems between "stock" neomutt IMAP browser functionality and after adding realpath(). It sure makes it obvious that IMAP in the browser is inadequate and I hope someone brushes it up. 

Getting IMAP browsing to be cleaner would be great but beyond the scope of this PR. 

My next commits allow the new 'p' `<goto-parent>` shortcut to work from IMAP, at least :), simply by using the "universal" `mutt_get_parent_path()` pre-existing muttlib function for the `<goto-parent>` function. But, it can't be useful until IMAP in the browser is fixed. The universal call supports notmuch as well, so yay!

The conflict between `mutt_realpath` and IMAP by fixing the scope of `realpath` call.
Other comments and dox touchups are in there.

bazzinotti:
Let's wrap up this PR. Let me squash my commits when you feel she's ready to be merged.

bazzinotti:
I may have found a dumbed-down impl of `realpath` in the muttlib function `mutt_pretty_mailbox` that doesn't unravel symlinks. This means that it may be possible to refactor that dumb `realpath` as a sub-function of the new `mutt_realpath`, adding a boolean flag to resolve symlinks to `mutt_realpath`. The dumb realpath code can be invoked when symlinks are not resolved. This in turn can set the stage for add a neomuttrc option to resolve symlinks in the browser. Something like `browser_resolve_symlinks` or better.

The above idea would postpone the PR a bit, so I'm fine with making a separate PR, allowing this one to be accepted as-is. I'd like to squash some commits first.

bazzinotti:
ignore the above, it's just Github sticky'ing the same commit that I amended (several times in attempt to stop this sticky behavior!).

This PR is ready for merging. :dancer:

PR #1040, opened 2018-01-27
[mutt_str_strnfcpy] fix off-by-1 bug & clarify parameter meaning

bazzinotti:
discussion: https://github.com/neomutt/neomutt/commit/74922c787a4efae9b8db9b497b680f073050c217

### DIFF
--- a/mutt/string.c
+++ b/mutt/string.c
@@ -723,7 +723,7 @@ int mutt_str_is_email_wsp(char c)
 size_t mutt_str_strnfcpy(char *dest, char *src, size_t size, size_t dlen)
 {
   if (dlen > size)
-    dlen = size - 1;
+    dlen = size;
   return mutt_str_strfcpy(dest, src, dlen);
 }
 
--- a/mutt/string.c
+++ b/mutt/string.c
@@ -674,13 +674,13 @@ void mutt_str_remove_trailing_ws(char *s)
  * mutt_str_strfcpy - Copy a string into a buffer (guaranteeing NUL-termination)
  * @param dest Buffer for the result
  * @param src  String to copy
- * @param dlen Length of buffer
- * @retval len Destination string length
+ * @param dsize size of destination buffer (includes null byte)
+ * @retval len Destination string length (excludes null byte)
  */
-size_t mutt_str_strfcpy(char *dest, const char *src, size_t dlen)
+size_t mutt_str_strfcpy(char *dest, const char *src, size_t dsize)
 {
   char *dest0 = dest;
-  while ((--dlen > 0) && (*src != '\0'))
+  while ((--dsize > 0) && (*src != '\0'))
     *dest++ = *src++;
 
   *dest = '\0';
@@ -716,15 +716,15 @@ int mutt_str_is_email_wsp(char c)
  * mutt_str_strnfcpy - Copy a limited string into a buffer (guaranteeing NUL-termination)
  * @param dest Buffer for the result
  * @param src  String to copy
- * @param size Maximum number of characters to copy
- * @param dlen Length of buffer
- * @retval len Destination string length
+ * @param size Maximum number of characters to copy (includes null byte)
+ * @param dsize size of destination buffer (includes null byte)
+ * @retval len Destination string length (excludes null byte)
  */
-size_t mutt_str_strnfcpy(char *dest, char *src, size_t size, size_t dlen)
+size_t mutt_str_strnfcpy(char *dest, char *src, size_t size, size_t dsize)
 {
-  if (dlen > size)
-    dlen = size;
-  return mutt_str_strfcpy(dest, src, dlen);
+  if (dsize > size)
+    dsize = size;
+  return mutt_str_strfcpy(dest, src, dsize);
 }
 
 /**


gahr:
I pushed a test case for `mutt_str_strnfcpy` on a branch, according to my expectations given the function signature. It currently fails, but it requires a different change to pass, namely `dsize = size + 1;`.

Please have a look at it and see whether your expectations would be different for any of it.

I don't like the "including / excluding null byte" in the comments. There really are three concepts here:

1. a buffer size, i.e., how many bytes are allocated for a certain object. This is `dlen`, and I agree it should be called `dsize`
2. a number of bytes to copy, i.e., the number of characters of the substring one's interested in. This is `size`. Perhaps `mutt_str_strncasecmp` has a better take at it, both in the comments and in the argument name. I think a simple `n` for "number" would do, since this is really a count, not a size (accidentally, sizeof(char) == 1, but if we were talking about doubles, I would still want to say "copy *1* double", no matter how many bytes that would take. That *1* is still a count, or number of items. Not a length, not a size.).
3. the resulting string length, as return value. "String length" is a widely accepted concept, and I don't think adding any information about it including or not including the terminator helps. The function's doxy comment already states that it guarantees null termination anyway.

bazzinotti:
I agree with you on all of that. Let me look at the test case when I have
some free time.

On Jan 28, 2018 6:48 AM, "Pietro Cerutti" <notifications@github.com> wrote:

I pushed a test case for mutt_str_strnfcpy on a branch, according to my
expectations given the function signature. It currently fails, but it
requires a different change to pass, namely dsize = size + 1;.

Please have a look at it and see whether your expectations would be
different for any of it.

I don't like the "including / excluding null byte" in the comments. There
really are three concepts here:

   1. a buffer size, i.e., how many bytes are allocated for a certain
   object. This is dlen, and I agree it should be called dsize
   2. a number of bytes to copy, i.e., the number of characters of the
   substring one's interested in. This is size. Perhaps mutt_str_strncasecmp
   has a better take at it, both in the comments and in the argument name. I
   think a simple n for "number" would do, since this is really a count,
   not a size (accidentally, sizeof(char) == 1, but if we were talking about
   doubles, I would still want to say "copy *1* double", no matter how many
   bytes that would take. That *1* is still a count, or number of items.
   Not a length, not a size.).
   3. the resulting string length, as return value. "String length" is a
   widely accepted concept, and I don't think adding any information about it
   including or not including the terminator helps. The function's doxy
   comment already states that it guarantees null termination anyway.

—
You are receiving this because you were assigned.
Reply to this email directly, view it on GitHub
<https://github.com/neomutt/neomutt/pull/1040#issuecomment-361056911>, or mute
the thread

bazzinotti:
That is interesting indeed! I added my test to it, and it causes a question to pop into my head - If you pass `strfcpy` `dsize=4` you will get a resultant string with `len` 3. So, shouldn't `strnfcpy` work the same way? Meaning if you have a dsize `n` you get a string len `n-1`.

Well I went forward with that logic, and changed the code back to `dsize = size`. And it only challenged one of your tests:

```
  const char src[] = "One Two Three Four Five";                           
  char dst[10];

...
  { /* copy a substring */                                                
    size_t len = mutt_str_strnfcpy(dst, src, 3, sizeof(dst));             
    if (!TEST_CHECK(len == 3))                                            
    {                                                                     
      TEST_MSG("Expected: %zu", 3);                                       
      TEST_MSG("Actual  : %zu", len);                                     
    }                                                                     
    if (!TEST_CHECK(strcmp(dst, "One") == 0))                             
    {                                                                     
      TEST_MSG("Expected: %s", "One");                                    
      TEST_MSG("Actual  : %s", dst);                                      
    }                                                                     
  }
```

With regards to size, The code is saying "size = 3 chars (Max number of chars to copy)". Your test checks to receive "One", but that is a size of 4 chars, and that is why it fails when `strnfcpy()` contains `dsize = size`. Should it provide "On" which is `size` 3, the same size in the test? I thought so, and that change will cause all tests to pass.

Now there's no spec for `strnfcpy()` that I could find in or out of neomutt, and it's only used once in neomutt. So it's not a big deal what we decide. Personally, I vote for having `strfcpy()` and `strnfcpy()` follow the same logic.

I'm putting the commit up to to the same devel/test-strnfcpy branch you started. I also added some new tests with brief explanations. Then I removed them for being superfluous, but that way you can have a look.

So the question to be answered here is: 

* is `strnfcpy()``size` param actually a string length as you treated it in one of your tests?
* Or is it still a `size` that includes null byte, which is identical to how `strfcpy()` treats it. ?

bazzinotti:
Btw, nice testing framework :)
Note to anyone who wants to try it - you might need to rerun `./configure` before `make test` can be invoked :)

bazzinotti:
Here's a reference I found for `strfcpy()` that has a nice blurb about it at the top. https://github.com/jonathangray/freebsd-1.x-ports/blob/master/mail/elm/lib/strfcpy.c

gahr:
> is strnfcpy()``size param actually a string length as you treated it in one of your tests?
Or is it still a size that includes null byte, which is identical to how strfcpy() treats it. ?

The 3rd param to strnfcpy is how many characters you want to copy from your source into your destination. `int ret = strnfcpy(dst, "ABCDEFG", 3, sizeof(dst);` says I want to copy three characters, which is obviously the substring "ABC". The fact that `dst` gets null-terminated has little to do with my willing to copy three characters. In this case `ret == 3 == strlen(dst)` must hold.

gahr:
I have applied some modifications to the functions' signatures, fixed the off-by-one error, and restored the test cases in [devel/test-strnfcpy](https://github.com/neomutt/neomutt/tree/devel/test-strnfcpy) . It should be mergeable as is.

bazzinotti:
Nice it looks great!

PR #1041, opened 2018-01-29
Fix off-by-1 bug in mutt_str_strnfcpy, add test cases

gahr:
Spin off of #1040.


bazzinotti:
@gahr it occurred to me yesterday that we should test the 1 instance of `mutt_strnfcpy` in neomutt for differences in its execution now that the function has been modified.

gahr:
The last commit fix the expansion of a nametemplate in an RFC1524 (mailcap) entry.

With a `nametemplate=temporary_%s` and an HTML attachment with `filename=spam.html` as parameter for the `Content-Disposition` header, the expanded filename is:

- before the patch: `/tmp/mutt/temporaryspam.html` (luck has it it still created something ok-ish)
- after the patch: `/tmp/mutt/temporary_spam.html`

PR #1044, opened 2018-01-30
fix parsing of urls containing '?'

### LINKED ISSUE #1043: smtp_url: password cannot contain a question mark
## bug report

* Expected behavior
Neomutt says the smtp_url is incorrect when sending an e-mail.
* Actual behavior
The url is correct.
* Steps to reproduce
Set in `.muttrc`:

```set smtp_url=smtps://"aaa":"AA?"@server```

Nothing else is needed, this is the whole config file causing the problem.

* Used program versions
Bisected, the first bad commit is https://github.com/neomutt/neomutt/commit/4ac1c81ed710c8b27dc23821330b2005f7122214
Archlinux, x86_64

flatcap:
The URL schemes come in two patterns
- smtp://user:pass@host
- notmuch:///path?query

The 'host' types don't take a query string.
'notmuch' and 'mailto' don't use a 'user:pass' string.

This fix will still be confused by a notmuch URL that has a '?' in the path component, but no query string.  (but I don't think the code ever coped with that).  e.g.

- notmuch:///dir?name/

Fixes #1043

### DIFF
--- a/mutt/rfc2047.c
+++ b/mutt/rfc2047.c
@@ -658,6 +658,9 @@ void mutt_rfc2047_encode(char **pd, const char *specials, int col, const char *c
  */
 void mutt_rfc2047_decode(char **pd)
 {
+  if (!pd || !*pd)
+    return;
+
   struct Buffer buf = { 0 }; /* Output buffer                          */
   char *s = *pd;             /* Read pointer                           */
   char *beg;                 /* Begin of encoded word                  */
@@ -745,5 +748,4 @@ void mutt_rfc2047_decode(char **pd)
 
   mutt_buffer_addch(&buf, '\0');
   *pd = buf.data;
-  return;
 }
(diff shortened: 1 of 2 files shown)


gahr:
Although admittedly this fix resolves the originating issue, I'd rather have the URI parser reject an invalid URI. If I read [RFC3987, Appendix A](https://tools.ietf.org/html/rfc3986#appendix-A) correctly, a question mark is not allowed in the `userinfo` production.

How about we make `url_parse` compliant instead? The user can still set its username in the url and its password in `smtp_pass`.

flatcap:
Yes, this just fixes us up to 'mutt' level; the 'neomutt' solution would be a RFC-compliant parser.
If you'd like to tackle that, great.
For now, though, I'd like to merge this.

gahr:
#1045 is also a production of mutt_rfc2047_decode not checking for NULL.

flatcap:
squash commits

PR #1050, opened 2018-02-07
configure: Check for Docbook DTDs

flatcap:
Run xsltproc with a minimal Docbook file.
If this succeeds, then the docs will probably build correctly.


flatcap:
Rather than use the `xsl` file from `doc/`, I've used a simple inline version, too.
I've gone back to the temp files, though.  However I worked it, I couldn't get xsltproc to accept **both** via piping/redirection.  Some combinations reported success, but didn't actually do anything.
There's probably some clever shell technique to do it, but then I worried about portability.

PR #1065, opened 2018-02-16
disable the folder-hook for $record

flatcap:
It's possible to use `$record` to save an email to a different account.
The right thing to do is to call the `folder-hook` for the save folder, save the email, then call the `folder-hook` for the 'current' folder.

Calling the hook this way causes problems for lots of people who have commands in their hooks that are called when they didn't expect.

The hooks are left in the code wrapped in: `#ifdef RECORD_FOLDER_HOOK`


flatcap:
Closed by 9b1f80004

PR #1066, opened 2018-02-17
Add tidy_path() with configurable symlink resolution

bazzinotti:
* **What does this PR do?**

    * Move `mutt_realpath` to more appropriate location mutt/file.c

    * Rename `mutt_realpath` to `tidy_path`;  The original `realpath` does not provide optional symlink resolution, so a different name is less confusing.

    * Add the config option `$browser_resolve_symlinks` with a default __unset__ to reflect original mutt behavior.

    * Add a foundation of test cases for `tidy_path`

Everything :sweat_smile: 

#1028


mutt/file.c
```diff
@@ -1397,7 +1397,13 @@ size_t mutt_file_tidy_path(char *buf, bool rsym)
...
 
   if (rsym)
-    return mutt_str_strfcpy(buf, NONULL(realpath(buf, s)), PATH_MAX);
+  {
+    char *rp = realpath(buf, s);
+    if (rp)
+      return mutt_str_strfcpy(buf, rp, PATH_MAX);
+    else
+      return mutt_str_strlen(buf);
+  }
```
flatcap: @bazzinotti 
This breaks one of your test cases.
If the path doesn't exist, `realpath()` returns `NULL` ...
Before: and the original path was killed
After: the original path is left untouched

I _think_ this is a saner thing to do.

bazzinotti: I think you are right. Btw nice job retouching the code :bowing_man: 
I've modified the test so that if `realpath` fails, it expects the output path to match the input path.

gahr:
Could we please define `BrowserResolveSymlinks` in `mutt/file.c` and avoid the `bool rsym` arg to `mutt_file_tidy_path` ?

bazzinotti:
@gahr, wouldn't that implicitly tie BrowserResolveSymlinks to the behavior of tidy_path? But there is possiblity that a developer will want to call this function independent of that setting because it is not explicitly a browser function. I will hold changes until further response from @flatcap @gahr

flatcap:
@bazzinotti 
Yeah, that's a good point.

flatcap:
## mutt_file_tidy_path()

Problems:
- Doesn't eliminate repeated cases `///` or `/././`
- Uses large buffers
- Needs large array of pointers

The output string will always be the same size or shorter than the original.
So we can work in-place, or better still duplicate the string and work on the copy.

## Stage 1 - Filter out `//+` and `(/.)+`

- Two buffers: (R)ead, (W)rite
- `!` means `NUL`
- `$` means end of string

```
(R)     /apple//banana!
(W)     !
```

Loop through R, writing to W
- Copy each char from R to W
- If you find `//` skip ahead of all the `/`s
- If you find `/./` or `/.$` skip ahead

Note: Make sure the string doesn't end in `/` or `/.`

## Stage 2 - Strip out `/..`

Duplicate the string and work with the copy.

```
(W)     /apple/banana/../cherry!
```

Loop while there are matches for `/..` (strstr)
1. Check the match ends with `/` or NUL
2. Where the match starts, write a NUL
3. Find the previous `/` in W (strrchr)
4. Copy the remaining string back (memmove)

```
                        1
(W)     /apple/banana/../cherry!
                     2   
(W)     /apple/banana!../cherry!
              3          
(W)     /apple/banana!../cherry!
              4          
(W)     /apple/cherry!../cherry!
```

---

Hopefully this makes sense.

gahr:
I’m almost convinced that it makes absolutely no sense to reimplement this logic in neomutt, since it’s bound to exist in tons of projects already - scripting and higher level languages, shells, and file browser applications come to mind.

As an example, see TclFSNormalizeAbsolutePath in https://core.tcl.tk/tcl/artifact/4465c9e9cdd0ab3f.

bazzinotti:
fwiw the // and /./ code I used came from (neo)mutt. I only derived a
parent traversal. There are some counter arguments I could form, but you
seem keen on your impl idea.

If you want to change non-symlink impl, please, be my guest. I want you to
be happy with the final result.

Consider everything else finished by my part :)

If you are hoping for more from me, I'm sorry I cannot make myself more
available at present.

flatcap:
Thanks, but this PR has been superseded by the functions:

- `mutt_path_tidy_slash()` - Remove unnecessary slashes and dots                                                                                                   
- `mutt_path_tidy_dotdot()` - Remove dot-dot-slash from a path                                                                                                     
- `mutt_path_tidy()` - Remove unnecessary parts of a path

Issue #1067, opened 2018-02-21
Bug detecting mail addressed to me with alternates


theojepsen:
### Expected behavior
`alternates` should be used for determining `to_chars` and showing mail addressed to me in the index.

### Actual behavior
`alternates` is not being used, so mutt doesn't know who I am. This results in the index not showing mail addressed to me (`T`, `+`). Also, when I reply-to-all, it replies to my e-mail address, even though it should be omitted.

If I set `from`, then it detects the from address in mail addressed to me: `set from = 'uname@domain.com'`.

### Steps to reproduce
I built the latest commit from master: 9666da1f13e8bf4bb32f7a0190bf63706a28dda5
```
Configure options: --prefix=/usr/local --ssl --tokyocabinet --sasl
```
I used the following config:

```
source 'gpg -d ~/.mutt/passwords.gpg |'
set imap_pass = $my_pass

set header_cache = ~/.mutt/tmp_headers
set folder    = "imaps://uname@domain.com@imap.domain.com/"
set spoolfile = "imaps://uname@domain.com@imap.domain.com/Archive"

alternates user.name@domain.com uname@domain.com
```

I ran neomutt: `./neomutt -F configfile`

The index should show `+` or `T` next to mail addressed to me. However, this is not displayed.

### Used program versions
Commit 9666da1f13e8bf4bb32f7a0190bf63706a28dda5 was used to reproduce this bug.
The latest commit for which this bug is not manifested is 8872e39a91236266a6052ce2766448cab323c7df
The next commit b4b7fc9939888725c8f86eeeb04544ca9badc11b seems to have introduced the bug.

### Operating System and its version
Ubuntu 16.04

theojepsen:
On master commit 9666da1f13e8bf4bb32f7a0190bf63706a28dda5 I commented out the following lines on in `mutt/regex.c`:
```
218   for (; rl; rl = rl->next)
219   {
220     //if (!rl->regex || rl->regex->regex)
221     //  continue;
222     if (regexec(rl->regex->regex, str, (size_t) 0, (regmatch_t *) 0, (int) 0) == 0)
223     {
```
After that, the index displays `T` and `+` for mail addressed to me (alternates).

PR #1068, opened 2018-02-22
fix: missing ! when checking regex pointer

theojepsen:
* **What does this PR do?**
A `!` was missing when checking regex pointers. This caused some regexes to be ignored.

#1067


flatcap:
Thanks @theojepsen

PR #1069, opened 2018-02-23
Fix docbook dtd

rtlanceroad:
Sorry for not following the contributing guidance. 
This fixes the docbook dtd detection. This is case-sensitive I think.

### DIFF
--- a/auto.def
+++ b/auto.def
@@ -327,7 +327,7 @@ if {[get-define want-doc]} {
   }
 
   msg-checking "Checking for DocBook DTDs..."
-  set res [exec xmlcatalog --shell << {public "-//OASIS//DTD DocBook XML V4.2//EN"}]
+  set res [exec xmlcatalog --shell << {PUBLIC "-//OASIS//DTD DocBook XML V4.2//EN"}]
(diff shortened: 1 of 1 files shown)


auto.def
```diff
@@ -327,7 +327,7 @@ if {[get-define want-doc]} {
   }
 
   msg-checking "Checking for DocBook DTDs..."
-  set res [exec xmlcatalog --shell << {public "-//OASIS//DTD DocBook XML V4.2//EN"}]
+  set res [exec xmlcatalog --shell << {PUBLIC "-//OASIS//DTD DocBook XML V4.2//EN"}]
```
flatcap: Ah, no.
If it were part of the DTD definition, then yes it would be uppercase.
Here, it's a **command** `public` with argument `"-//OASIS//DTD DocBook XML V4.2//EN"`
which is fed into `xmlcatalog --shell`

rtlanceroad: Ok I thought public was a part of the DTD contents. Yesterday I saw public was uppercased in 4.2's docbook.cat, at least for my distro CRUX. If it's not the case for all distros, please close this, I'll fix it on my side.

Issue #1073, opened 2018-02-24
Use Portable directory names


fekir:
## feature requests

* Why do you not like the current state?
The directories used by `header_cache` and `message_cachedir` are not "portable"
For example, given following configuration:
````
set imap_user = "foo.bar"
set hostname = "gmail.com"
set folder = "imaps://imap.gmail.com:993"
set header_cache = ~/.cache/neomutt/cache
set message_cachedir = ~/.cache/neomutt/message
````
Following folder is created: `~/.cache/neomutt/message/imaps:foo.bar@imap.gmail.com:993`

* What would you like to see to change?
Use by default "portable" names.

The biggest issue is that `:` is not part of a valid filename under windows. On this environment it is used for file [streams](https://msdn.microsoft.com/en-us/library/windows/desktop/aa364404(v=vs.85).aspx), and can therefore cause difficulties and incompatibilities with other programs.

On most GNU/Linux system any character except '/' will be valid for a filename.
Unfortunately even on those systems problems might arise when using different encoding and special characters, like `?` or `*`.
It would probably be best to avoid those [reserved characters](https://en.wikipedia.org/wiki/Filename#Reserved_characters_and_words) at all to ease supporting other platforms and interoperability with other tools.

Even POSIX defines a ["Portable Filename Character Set"](http://pubs.opengroup.org/onlinepubs/9699919799/basedefs/V1_chap03.html#tag_03_282), which is hopefully supported on all systems.

There are some downsides if restricting the current schema for creating directories on the posix character set (or some other subset of characters): name collisions.
On the other hand name collisions are already possible since some filesystem+OS combinations (mainly NTFS+windows and HFS+mac) use case insensitive filenames, so it might not be such a blocker, considering that the user can still set `header_cache` and `message_cachedir`.

flatcap:
That sounds like a great idea.

Fixing a single place, like the header cache, will be easy enough.
Making a general solution will be much harder -- there are over 80 calls to `fopen(3)`.

fekir:
Probably a general solution will not be possible, since the user could always use non-portable characters when setting directories.
I think the default settings are the most important, so fixing for example header cache would be enough for "fixing" this feature request.
There will still be use-cases that are not really fixable, if for example the user home directory contains special characters, so it does not have to be a strict rule.

gahr:
Merged to master.

PR #1080, opened 2018-02-27
Make sure paths created by header/body/news cache are POSIX-compliant

gahr:
Issue #1073

Please double check and test - I have tested it with my configuration, but header cache is subtle.


gahr:
I have checked this with IMAP and empirically followed the code path that would be taken by POP (...), and the changes look ok to me now. I would appreciate a test with POP / NNTP, though.

gahr:
Checked nntp - works fine.

gahr:
Rebased.

PR #1085, opened 2018-03-01
Devel/cppcheck warning

fekir:
* **What does this PR do?**
Clean up warnings issued by static analyzers, I want to get more confident with the code base.
Most commits simply reduce the scope of variable, eventually remove unused variables and sprinkles `const` here and there. Mainly minor changes with a low chance to break something.

That no behavior has been accidentally changed, for example changing
`````
void f(int x)
{
  int i = 0;
  for (int n = 0; n < 10; ++n)
  {
    do_something(&i);
  }
}
````` 
to 
`````
void f(int x)
{ 
  for (int n = 0; n < 10; ++n)
  {  
    int i = 0;
    do_something(&i);
  }
}

````` 
would be a possibly invalid change.

I think so, and I would also propose to add to the  [style guide](https://www.neomutt.org/dev/coding-style) that variables, if possible, should be declared as `const`.

### DIFF
--- a/pgppubring.c
+++ b/pgppubring.c
@@ -458,8 +458,6 @@ static int pgp_parse_pgp2_sig(unsigned char *buf, size_t l,
 static int pgp_parse_pgp3_sig(unsigned char *buf, size_t l,
                               struct PgpKeyInfo *p, struct PgpSignature *s)
 {
-  time_t sig_gen_time = -1;
-  long validity = -1;
   long key_validity = -1;
   unsigned long signerid1 = 0;
   unsigned long signerid2 = 0;
@@ -507,22 +505,11 @@ static int pgp_parse_pgp3_sig(unsigned char *buf, size_t l,
       switch (skt & 0x7f)
       {
         case 2: /* creation time */
-        {
-          if (skl < 4)
-            break;
-          sig_gen_time = 0;
-          for (int i = 0; i < 4; i++)
-            sig_gen_time = (sig_gen_time << 8) + buf[j++];
-
-          break;
-        }
         case 3: /* expiration time */
         {
           if (skl < 4)
             break;
-          validity = 0;
-          for (int i = 0; i < 4; i++)
-            validity = (validity << 8) + buf[j++];
+          j += 4;
           break;
         }
         case 9: /* key expiration time */

### CODE REVIEW

**Comment by gahr** [comment]
>> @fekir
>> I would also propose to add to the style guide that variables,
if possible, should be declared as const.

> @flatcap 
> Hmm... I hadn't considered that before. I don't mind that.
We'll see what the others think.

Yes, please!

File: alias.c [hunk 1]
```diff
@@ -520,6 +519,7 @@ int mutt_alias_complete(char *s, size_t buflen)
                            MIN(mutt_str_strlen(a->name) + 1, sizeof(bestname)));
         else
         {
+          int i;
           for (i = 0; a->name[i] && a->name[i] == bestname[i]; i++)
```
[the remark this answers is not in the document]
fekir: Unfortunately not, it is used in `bestname[i] = '\0';`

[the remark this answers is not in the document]
flatcap: Ah, sorry.  Of course.

fekir:
@flatcap: in the sample I've provided code would still compile, and there is no shadowing, So there might be some tiny chance to break something. I pad attention, but it is no impossible that I missed something.

We should also eventually add that `FILE *fp` is preferred to the guidelines (personally I dislike it, but NP).

Some good reasoning for using const: https://github.com/isocpp/CppCoreGuidelines/blob/master/CppCoreGuidelines.md#S-const (yes, I know that C and C++ are two different languages, but a lot of things written in the guidelines are language agnostic)

color.c
```diff
@@ -702,7 +701,7 @@ static int parse_object(struct Buffer *buf, struct Buffer *s, int *o, int *ql,
     if (buf->data[6])
     {
       *ql = strtol(buf->data + 6, &eptr, 10);
-      if (*eptr || q_level < 0)
+      if (*eptr)
       {
```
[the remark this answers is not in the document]
fekir: Ok, I could have looked myself

fekir:
One question, in the function `decode_xbit` (see `handler.c`), all variables are used only if the `if`-statement evaluates to true.

I think changing the code to:
````
decode_xbit(){
if(!istext)
{
    mutt_file_copy_bytes(s->fpin, s->fpout, len);
    return;
}
(A)
}
````
would be more readable than the original form:
````
decode_xbit(){
if(istext)
{
    (A)
}
else
   mutt_file_copy_bytes(s->fpin, s->fpout, len);
}
````
since (A) is approx 25 lines long, we would have a nested scope less (inside (A) there is a while loop with if statements too) and (IMHO), in its current form, the else branch is "hidden" since it is so short compared to the first branch.
I think it would be a good case where to have an early exit, but since I'm not changing anything (except the scope of variables/readability), I wanted to ask if it is OK.

**Comment by gahr** [comment]
I agree with you

**Comment by flatcap** [comment]
> We should also eventually add that FILE *fp is preferred to the guidelines

Yes, will do.
 
> Some good reasoning for using const

Wow, I could spend a week reading that!

> I think it would be a good case where to have an early exit

Definitely

> but since I'm not changing anything (except the scope of variables/readability),
> I wanted to ask if it is OK.

These sorts of changes will, may cause large diffs,
but as long as each is in a separate commit, that'll be fine.

**Comment by fekir** [comment]
I've edited all files until letter 'i' where I could see that the scope of the variable could be reduced (I grepped for that warning in cppcheck).
I noticed that he did not warn for some variables that I changed too since they were next to the one I just changed, so there might be other places were we could reduce the scope.

I think it might be a good moment to merge it back, in order to avoid having a giant PR that changes every file, with possible conflicts with the other feature branches.

I kept the commit small, most of the times a commit per file, and eventually a separate commit if I did not only change the scope of a variable.

**Comment by flatcap** [comment]
> I think it might be a good moment to merge it back

OK, thanks.  I'll have another close look shortly.
(meanwhile, you can keep pushing commits to this branch)

> there might be other places were we could reduce the scope.

There are also a few degenerate cases such as `mutt_index_menu()` and `mutt_pager()` where one or two function-scope variables are used for _so many_ different things.  Keep that in mind for when you want a challenge :-)

**Comment by flatcap** [comment]
I've re-ordered, renamed and merged your 'reducing scope' commits.
I've added the filename to messages so we can tell them apart.

**Comment by fekir** [comment]
@flatcap: thank you for merging it back, I'm a little confused about the status of this pull request.

GitHub shows me that there are still 4 commits in this feature branch, I guess they were not merged back, right?

Should I also work on this feature branch/PR or better make a new one?

**Comment by flatcap** [comment]
Yes, I merged the large block of simple changes.  Thanks.

I've checked the 4 more complicated commits -- they're great too, but...
Had I merged these 4, github would have auto-closed the issue.
Once before, we've had trouble re-opening an PR.

As you've still got lots of changes to make (some simple, some complicated)
I'll bundle them up when you're done.

Separating the simple from the complex changes is more than cosmetic.
When tracking down bugs, knowing that a block of 30 commits is 'safe' could save a lot of time.

So, please keep adding commits to this PR.

**Comment by flatcap** [comment]
I've checked through the latest batch -- all good.

I was a little worried by d3d5657c, though!
I checked and the bug's there in upstream, too.

If you have a second, I noticed three functions that could do with an early return:
`maildir_open_mailbox_append()`, `mh_open_mailbox_append()` and `mutt_file_unlink()`

Do you have any more changes ready to push?
If not, I'll bundle up this set and merge it.

Thanks.

**Comment by fekir** [comment]
> Do you have any more changes ready to push?

Not yet, maybe tomorrow.

> `maildir_open_mailbox_append()`, `mh_open_mailbox_append()` and `mutt_file_unlink()`

I'll give a look at them too

File: rfc3676.c [hunk 3]
```diff
@@ -322,7 +322,7 @@ int rfc3676_handler(struct Body *a, struct State *s)
 
     /* for DelSp=yes, we need to strip one SP prior to CRLF on flowed lines */
     if (delsp && !fixed)
-      buf[--buf_len] = '\0';
+      buf[buf_len - 1] = '\0';
```
flatcap: Hmm... that's an interesting change.
Did `cppcheck` suggest this, or is it your work?

query.c
```diff
@@ -343,12 +341,15 @@ static void query_menu(char *buf, size_t buflen, struct Query *results, int retb
 
     menu->data = QueryTable = mutt_mem_calloc(menu->max, sizeof(struct Entry));
 
+    int i;
```
flatcap: This `i` is used in 6 different `for` loops.
I'm confident that each one isn't used beyond the scope of the `for`.

rfc3676.c
fekir: No, cppechek suggested that the scope of `buf_len` could be reduced.
I noticed that only there the value got was mutated. I replace it so that `buf_len` can be declared as `const`.
Since the value is not used afterwards, it should not cause problems.

query.c
fekir: Unfortunately all those loops begin with `for (i = 0, queryp = results;`.

Since `queryp` is not an `int`, I cannot change it to `for (i = 0, queryp = results;`.

rfc3676.c
flatcap: nice :-)

query.c
flatcap: ah, fair enough

fekir:
Except for `jimsh0.c` (I guess it's autogenerated), I should have fixed all cppcheks warnings about variable scopes (plus some other things like improve const correctness, remove unused variables and so on, when I noticed it).

Of course there are other categories of warnings (redundant code, variable reassigned before reading its old value, conditions that are always true or false....) that should get checked.

gahr:
Thanks for the huge work @fekir! In addition to jimsh0.c, please avoid touching anything in autosetup/ and test/acutest.h. Those are imported from upstream projects. Thanks!

fekir:
@gahr: Ok, I've removed those changes and added a couple of commits since I noticed that I also missed a couple of warning.

flatcap:
> I should have fixed all cppcheck warnings

Brilliant.

While I'm checking through the last batch, please can you document what you've done.
In the [website repo](https://github.com/neomutt/neomutt.github.io):
- Add a sentence about `const`-ness in ["Coding Style"](https://www.neomutt.org/dev/coding-style)
  I suggest putting it in the "Top Tips" section after "Use ! to test booleans and pointers only"
- Add a new page, "Analysis" in the `_dev` directory
  - Cut'n'paste `debug.md` to get you started
  - Write sentence describing `cppcheck`
  - The config file you used
  - The command line you used

That way others can learn from your experience and anyone can test new code.

Thanks.

flatcap:
OK, I've checked and merged all but one of your changes.
The remaining commit gave me a real WTF? moment, but it's nothing you've done.

`pgppubring.c` has several functions for dealing with (now archaic) versions of PGP.
Some of the code in these functions appears to have been cut'n'pasted from elsewhere.
The clue is that several calculations are performed whose results aren't used.
This code hasn't changed since Mutt's first commit.

Rather than optimising obsolete code, I'd rather investigate if we can just kill it.

flatcap:
On a happier note...  Well done! :+1: 

Stats:
- 84 commits
- 68 files changed
- 209 fewer lines (+668,-877)

PR #1087, opened 2018-03-04
Add support for alert (blink) colors.

parazyd:
* **What does this PR do?**
Add support for `alert$color` as neomutt colors. It will make text blink.

Not really, no.

https://parazyd.org/pub/dev/random/alert.webm

None. This is a new feature.

### DIFF
--- a/color.c
+++ b/color.c
@@ -384,12 +384,18 @@ void mutt_free_color(int fg, int bg)
 static int parse_color_name(const char *s, int *col, int *attr, int is_fg, struct Buffer *err)
 {
   char *eptr = NULL;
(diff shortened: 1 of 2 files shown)


parazyd:
@flatcap 
Should I rename it to 'blink' then?

If so, how should I amend? Rebase the patch or just commit on top?

flatcap:
Leave it for now.  We'll see what the others think.

> If so, how should I amend? Rebase the patch or just commit on top?

Here at NeoMutt we're happy for either:
- Amend your commit and force-push
- Add a new commit on top (and we'll squash them before merging)

PR #1094, opened 2018-03-11
support for multipart/alternative and multipart/mixed

walkerlala:
* **What does this PR do?**

This work is based on commit 5b4f7a1342 from @somini and was discussed at <https://github.com/neomutt/neomutt/issues/587>. As of now, one can tag entries in the compose menu and use <group-alternatives> to group those entries into one. If all the entries are tagged (including the
fundamental part), then the email become a multipart/alternative one, otherwise it becomes a multipart/mixed one.

There is some weird things to be noted, though: when postponning a multipart/alternative mail and then recall it, those entries get splitted up. This does not occurred for multipart/mixed mails.

Also, please note line 1160 in compose.c


flatcap:
Thanks @walkerlala.

To keep the development in one place, I've applied your work on top of @somini's.
It's now in branch [devel/multipartalternative]

I've:
- rebased the original commit on top of master
- clang-format'd the code
- applied your code on top
- copied your comment to #734
- invited you to join NeoMutt

As a member, you'll be able to commit directly to that branch.

Issue #980, opened 2017-12-03
add the correct buildflags when building makedoc


aradici:
Currently, when building doc/makedoc, a CC_FOR_BUILD command is used; that command does not include CFLAGS and CPPFLAGS from the environment, which in Debian, will cause the hardening flags not to be included.

This bug is not tagged distro:debian because it's not really debian specific. This patch should fix the issue:
https://anonscm.debian.org/cgit/pkg-mutt/neomutt.git/tree/debian/patches/neomutt-devel/0005-add-flags-to-cc-for-build.patch?id=20007531ada69e289d95e8dad2ac37e3e2075b32

aradici:
It seems taht I forgot to add @LDFLAGS@ to the list :( that should be there too. I'll prepare a better patch

aradici:
This should be correct now and does not generate any warnings:
https://anonscm.debian.org/cgit/pkg-mutt/neomutt.git/tree/debian/patches/neomutt-devel/0005-add-flags-to-cc-for-build.patch

gahr:
The reason for having a separate `CFLAGS_FOR_BUILD` variable is to be able to differentiate between flags used to compile programs to be used on the host system vs. target system.  `CFLAGS_FOR_BUILD` are used for the former, while `CPPFLAGS` and `CFLAGS` are used for the latter.

If you need to specify additional flags for the former case, I'd suggest augmenting the `CC_FOR_BUILD` definition with an additional `EXTRA_CFLAGS_FOR_BUILD` variable. That could be populated from the command line.

Please do not pollute `CFLAGS_FOR_BUILD` with `CFLAGS` and `CPPFLAGS`, as that is breaking the logic because of which `CFLAGS_FOR_BUILD` was originally created.

gahr:
@aradici 

> in Debian, will cause the hardening flags not to be included.

Does Debian really care about having those flags used for binaries used only at build time?

aradici:
yes we need the output of dpkg-buildflags, it also apply to the linker flags

gahr:
@aradici in #1098 I have added explicit `EXTRA_CFLAGS_FOR_BUILD` and `EXTRA_LDFLAGS_FOR_BUILD` that your distro can set so that anything from dpkg-buildflags is used when building executables for the host system. Will this ease your porting work?

gahr:
#1098 was merged, thanks.

PR #1098, opened 2018-03-14
Expose EXTRA_CFLAGS_FOR_BUILD and EXTRA_LDFLAGS_FOR_BUILD

gahr:
This allows distros to set flags that must be used both for building
target and host executables.

Issue #980


aradici:
That can work, thanks a lot!

PR #1102, opened 2018-03-16
Update Turkish translation

rypervenche:
Update Turkish translation.

This is not a complete translation, but I fixed up some of the mistakes and as many fuzzy translations as I could. I left the word "fuzzy" in some of the change that I made as I cannot be sure if they are 100% correct or not, but they are at least improvements.

I hope we can get a native speaker to help with the rest. :)

### DIFF
--- a/po/tr.po
+++ b/po/tr.po
@@ -92,7 +92,7 @@ msgstr "Dosyaya kaydet: "
 #: alias.c:413
 #, fuzzy
 msgid "Error reading alias file"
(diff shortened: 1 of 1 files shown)


flatcap:
Thanks for the translations, @rypervenche 

Also for enlightening me about GitHub -- I didn't know you could raise a pull request against specific branches!

Note that travis spotted a mistake in the translation: one string ended `\n` and yours ended `\n ` (an extra space).  I've fixed it and updated your branch.

PR #1105, opened 2018-03-18
Re-sort resources for better documents

floand:
* **What does this PR do?**

  The main goal is to build sorted lists of commands, functions and variables
  for the documentation (`manual.xml`, `neomuttrc`, `neomuttrc.man`).

  The latter needs further decisions, because the commands section in
  `neomuttrc(5)` differs, due to usage of file `doc/neomuttrc.man.head`, from
  the generated content of `manual.xml`. Also it is probably about the time
  for timestamp updates within the manual pages:

    - `doc/neomutt.man:20`               ... neomutt 1 "January 2009" Unix "User Manuals"
    - `doc/neomuttrc.man.head:29` ... neomuttrc 5 "September 2002" Unix "User Manuals"

  with the release version date, maybe?

  If the code re-sort is OK, nothing special yet, I think. The most changes are
  code line moves. Some real changes are related to indentation fixes and
  splitting up some of the preprocessor directives.

  Compiling with `--full-doc` works as usual.

### DIFF
--- a/auto.def
+++ b/auto.def
@@ -177,6 +177,29 @@ if {1} {
   proc yesno val {
     expr {$val ? "yes" : "no"}
   }
+
+  # Recent author timestamp of file(s) as UTC, use PACKAGE_VERSION as fallback
+  proc get-author-date {format args} {
+    set oldpwd [pwd]
+    cd $::autosetup(srcdir)
+
+    if {$args eq "" || [catch {
+        set fd [open "|git log -1 --format=%at master -- $args" r]
+        gets $fd epoch
+        close $fd}] || $epoch eq ""
+    } {
+      #user-notice "No author timestamp found, falling back to PACKAGE_VERSION..."
+      if {![set pkgdate [get-define PACKAGE_VERSION]] || [catch {
+          set epoch [clock scan $pkgdate -format {%Y%m%d} -timezone :UTC]}]
+      } {
+        autosetup-error "Cannot get or parse PACKAGE_VERSION as date!"
+      }
+    }
+    cd $oldpwd
+
+
+    return [clock format $epoch -format $format -timezone :UTC]
+  }
 }
 ###############################################################################
 
@@ -916,6 +939,12 @@ foreach dir $subdirs {
   define-append UNINSTALL_TARGETS uninstall-$dir
   make-template $dir/Makefile.autosetup $dir/Makefile
 }
+
+###############################################################################
+# Provide timestamps (UTC) based on most recent author date or PACKAGE_VERSION
+# doc/neomuttrc.man, doc/neomutt.1
+define TS_MAN_CONF [get-author-date "%Y-%m-%d" init.h doc/neomuttrc.man.head doc/neomuttrc.man.tail]
+define TS_MAN_MAIN [get-author-date "%Y-%m-%d" doc/neomutt.man]
 
 ###############################################################################
 # Generate Makefile and config.h
--- a/auto.def
+++ b/auto.def
@@ -189,8 +189,8 @@ if {1} {
         close $fd}] || $epoch eq ""
     } {
       #user-notice "No author timestamp found, falling back to PACKAGE_VERSION..."
-      if {![set pkgdate [get-define PACKAGE_VERSION]] || [catch {
-          set epoch [clock scan $pkgdate -format {%Y%m%d} -timezone :UTC]}]
+      if {[catch {clock scan [get-define PACKAGE_VERSION] \
+                  -format {%Y%m%d} -timezone :UTC} epoch]
       } {
         autosetup-error "Cannot get or parse PACKAGE_VERSION as date!"
       }
--- a/auto.def
+++ b/auto.def
@@ -930,11 +930,6 @@ make-template Makefile.autosetup Makefile
 make-config-header config.h -auto $auto_rep -bare $bare_rep -str $str_rep
 
 ###############################################################################
-# Generate doc/neomutt.1
-define bindir [get-define BINDIR]
-define docdir [get-define PKGDOCDIR]
-
-###############################################################################
 # Generate .clang_complete
 define cflags-one-per-line [string map {" " "\n"} [get-define CFLAGS]]
 make-template .clang_complete.in
--- a/auto.def
+++ b/auto.def
@@ -177,24 +177,6 @@ if {1} {
   proc yesno val {
     expr {$val ? "yes" : "no"}
   }
-
-  # Recent author timestamp of file(s) as UTC, use PACKAGE_VERSION as fallback
-  proc get-author-date {format args} {
-    set oldpwd [pwd]
-    cd $::autosetup(srcdir)
-
-    if {$args eq "" || [catch {
-        set fd [open "|git log -1 --format=%at master -- $args" r]
-        gets $fd epoch
-        close $fd}] || $epoch eq ""
-    } {
-      set epoch [clock scan \
-                   [get-define PACKAGE_VERSION] -format {%Y%m%d} -timezone :UTC]
-    }
-    cd $oldpwd
-
-    return [clock format $epoch -format $format -timezone :UTC]
-  }
 }
 ###############################################################################
 
@@ -938,8 +920,8 @@ foreach dir $subdirs {
 ###############################################################################
 # Provide timestamps (UTC) based on most recent author date or PACKAGE_VERSION
 # doc/neomuttrc.man, doc/neomutt.1
-define TS_MAN_CONF [get-author-date "%Y-%m-%d" init.h doc/neomuttrc.man.head doc/neomuttrc.man.tail]
-define TS_MAN_MAIN [get-author-date "%Y-%m-%d" doc/neomutt.man]
+define PACKAGE_DATE \
+  [regsub {(....)(..)(..)} [get-define PACKAGE_VERSION] {\1-\2-\3}]
 
 ###############################################################################
 # Generate Makefile and config.h
@@ -951,7 +933,6 @@ make-config-header config.h -auto $auto_rep -bare $bare_rep -str $str_rep
 # Generate doc/neomutt.1
 define bindir [get-define BINDIR]
 define docdir [get-define PKGDOCDIR]
-make-template doc/neomutt.man doc/neomutt.1
 
 ###############################################################################
 # Generate .clang_complete
--- a/auto.def
+++ b/auto.def
@@ -918,8 +918,8 @@ foreach dir $subdirs {
 }
 
 ###############################################################################
-# Provide timestamps (UTC) based on most recent author date or PACKAGE_VERSION
-# doc/neomuttrc.man, doc/neomutt.1
+# Define package timestamp (UTC) based on PACKAGE_VERSION for:
+# doc/neomuttrc.5, doc/neomutt.1
 define PACKAGE_DATE \
   [regsub {(....)(..)(..)} [get-define PACKAGE_VERSION] {\1-\2-\3}]
 
--- a/auto.def
+++ b/auto.def
@@ -188,15 +188,10 @@ if {1} {
         gets $fd epoch
         close $fd}] || $epoch eq ""
     } {
-      #user-notice "No author timestamp found, falling back to PACKAGE_VERSION..."
-      if {[catch {clock scan [get-define PACKAGE_VERSION] \
-                  -format {%Y%m%d} -timezone :UTC} epoch]
-      } {
-        autosetup-error "Cannot get or parse PACKAGE_VERSION as date!"
-      }
+      set epoch [clock scan \
+                   [get-define PACKAGE_VERSION] -format {%Y%m%d} -timezone :UTC]
     }
     cd $oldpwd
-
 
     return [clock format $epoch -format $format -timezone :UTC]
   }
--- a/Makefile.autosetup
+++ b/Makefile.autosetup
@@ -46,6 +46,7 @@ mandir=		@mandir@
 libdir=		@libdir@
 libexecdir=	@libexecdir@
 sysconfdir=	@SYSCONFDIR@
+textdomaindir=	@MUTTLOCALEDIR@
 
 # targets for specific subdirectories
 ALL_TARGETS=		@ALL_TARGETS@
--- a/Makefile.autosetup
+++ b/Makefile.autosetup
@@ -3,6 +3,7 @@
 
 PACKAGE=	@PACKAGE@
 PACKAGE_VERSION=@PACKAGE_VERSION@
+PACKAGE_DATE=	@PACKAGE_DATE@
 
 # Build-time compiler and flags. These are used for building executables that
 # are only used at build-time, e.g., doc/makedoc. These are different from CC /
@@ -55,10 +56,6 @@ UNINSTALL_TARGETS=	@UNINSTALL_TARGETS@
 VPATH=		$(SRCDIR)
 
 ALL_FILES!=	(cd $(SRCDIR) && git ls-files 2>/dev/null) || true
-
-# timestamps
-TS_MAN_CONF=	@TS_MAN_CONF@
-TS_MAN_MAIN=	@TS_MAN_MAIN@
 
 ###############################################################################
 # neomutt
--- a/Makefile.autosetup
+++ b/Makefile.autosetup
@@ -56,6 +56,10 @@ VPATH=		$(SRCDIR)
 
 ALL_FILES!=	(cd $(SRCDIR) && git ls-files 2>/dev/null) || true
 
+# timestamps
+TS_MAN_CONF=	@TS_MAN_CONF@
+TS_MAN_MAIN=	@TS_MAN_MAIN@
+
 ###############################################################################
 # neomutt
 NEOMUTT=	neomutt$(EXEEXT)
--- a/Makefile.autosetup
+++ b/Makefile.autosetup
@@ -43,7 +43,7 @@ datadir=	@PKGDATADIR@
 docdir=		@PKGDOCDIR@
 mandir=		@mandir@
 libdir=		@libdir@
-libexecdir=		@libexecdir@
+libexecdir=	@libexecdir@
 sysconfdir=	@SYSCONFDIR@
 
 # targets for specific subdirectories
--- a/doc/Makefile.autosetup
+++ b/doc/Makefile.autosetup
@@ -68,10 +68,11 @@ doc/neomuttrc.5:	doc/makedoc$(EXEEXT) \
 		$(SRCDIR)/init.h \
 		$(SRCDIR)/doc/neomuttrc.man.head \
 		$(SRCDIR)/doc/neomuttrc.man.tail
-	( sed -e "/^\.TH/s/@MAN_DATE@/$(PACKAGE_DATE)/" \
+	( sed -e "/^\.TH/s|@MAN_DATE@|$(PACKAGE_DATE)|" \
 	    $(SRCDIR)/doc/neomuttrc.man.head && \
-	    $(MAKEDOC_CPP) $(SRCDIR)/init.h | doc/makedoc$(EXEEXT) -m | \
-	    cat - $(SRCDIR)/doc/neomuttrc.man.tail \
+	    $(MAKEDOC_CPP) $(SRCDIR)/init.h | doc/makedoc$(EXEEXT) -m && \
+	    sed -e "s|@MAN_DOCDIR@|$(docdir)|g" \
+	      $(SRCDIR)/doc/neomuttrc.man.tail \
 	) > $@
 
 doc/neomutt.1:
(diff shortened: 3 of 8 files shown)


flatcap:
hehe, very nice.  It's something I've been meaning to do for ages.  It all looks good.

I made one change in `opcodes.h`; I moved the `OP_NULL` operation back to the start of the array.
Just in case some code is expecting `OP_NULL` to be the 0th element in the array.
It probably doesn't matter...

(I've also tidied your commits).

Note: I don't run clang-format on the header files because of all the manual formatting that we do,
but I've left in your `//clang-format` markers.

> `neomuttrc.man` needs further decisions,
> because the commands section in neomuttrc(5) differs,

What differs?  What needs deciding?

> timestamp updates within the manual pages

Would you like to try fixing that?
Currently, most of the man pages are just copied into the man dir.
You'd need to rename them to `*.man` and insert a `sed` build step, like `neomutt.man`

btw, if you haven't seen it, [Making NeoMutt](https://www.neomutt.org/dev/make) has diagrams of the build.

flatcap:
Slight change of plan, I've merged your commits.
Large commits like these become very hard to merge, if not done immediately.
(I know from experience)

I've left a dummy commit to keep the PR open.

floand:
> I made one change in opcodes.h; I moved the OP_NULL operation back to the start of the array.
> Just in case some code is expecting OP_NULL to be the 0th element in the array.
> It probably doesn't matter...

Good point. I saw it but after successful compiling and testing some  
Vim-like key bindings (with prior `noop` in front), I left it sorted.

>(I've also tidied your commits).

That’s fine, I don’t wanted to squash the commits too early.

>Note: I don't run clang-format on the header files because of all the manual formatting that we do,
>but I've left in your `//clang-format` markers.

From my point they can go away, but I found it in `functions.h`. That and the 
result of running clang-format against `init.h` and `opcodes.h` let me put it 
in to prevent future surprises while sanitising.

>> `neomuttrc.man` needs further decisions,
>> because the commands section in neomuttrc(5) differs,
>
>What differs?

Mostly the command definitions. Sorry, I’ve had to be more clearer about
that. I don’t dig deeply into it (because it causes headaches) but here
are some examples:

*   _account-hook_
    -   `doc/neomuttrc.man.head` points out the ability of negation
        (`[!]regex`) while `doc/manual.xml.head` doesn’t (in opposite to
        _folder-hook_ or _mbox-hook_)

*   _alias_, _bind_, _ignore_
    -   notice the comma separations in their definitions
    -   _unalias_ in `doc/neomuttrc.man.head` lacks the `[-group name]` part
    -   in `doc/neomuttrc.man.head` _unignore_ / _ignore_ are merged 
        into one definition

*   _open-hook_, _close-hook_, _append-hook_
    -   are separate list items in `doc/manual.xml.head`
    -   compare definition `open-hook regex "command"` and
        `open-hook pattern shell-command`; notice the absent `[!]` token
        (before regex/pattern), that is used in other but similar definitions

*   _timeout-hook_
    -   is only mentioned in `doc/neomuttrc.man.head` not in commands
        section of `doc/manual.xml.head`, see [Configuration Commands](https://www.neomutt.org/guide/reference#2-%C2%A0configuration-commands)

*   _unhdr_order_
    -   is only mentioned in `doc/manual.xml.head`

*   ...   

>What needs deciding?

For a consistent documentation across all formats we need a ‘master’ reference,
that is, I think, the generated `manual.xml`, because it includes the newest source
code changes, *.head and *.tail files.

But some command definitions in `doc/neomuttrc.man.head` are nicer than in
`doc/manual.xml.head`, e.g. `type[/subtype]` instead of just `mimetype`.

Maybe the command section/list in both files needs a reorganisation, e.g. by
grouping (common, hook, address, etc. related) commands into sub-lists or under
sub-headers.

Where I’m sure about and what I will do, is to re-sort the command section in
`doc/neomuttrc.man.head` by command names.

>
>> timestamp updates within the manual pages
>
>Would you like to try fixing that?

Sure, but please be patient. I have to figure out the magic stuff first. ;-)

>I've left a dummy commit to keep the PR open.

OK, I’ll use it for the above.

floand:
Hi @flatcap,

if my Tcl procedure and other changes stand up your scrutiny, I think the
manual page timestamps for `doc/neomuttrc.man` and `doc/neomutt.1`
are fixed now.

I think implementing it via Tcl is more suitable (and reusable) than multiple
shell commands in a Makefile, so manual pages will be stamped while they
are build and not when they get installed.

auto.def
```diff
@@ -176,6 +176,29 @@ if {1} {
...
+    set oldpwd [pwd]
+    cd $::autosetup(srcdir)
+
+    if {$args eq "" || [catch {
+        set fd [open "|git log -1 --format=%at master -- $args" r]
+        gets $fd epoch
+        close $fd}] || $epoch eq ""
+    } {
+      #user-notice "No author timestamp found, falling back to PACKAGE_VERSION..."
+      if {![set pkgdate [get-define PACKAGE_VERSION]] || [catch {
```
gahr: I don't think `set pkgdate [get-define PACKAGE_VERSION` can ever fail. How about just using `[get define PACKAGE_VERSION]` in your `[clock scan]` command?

floand: I wanted to avoid nested evaluations to keep it more readable and to prevent
a potential free form scan resulting in current day start when the `-format`
constraint is not given.

But it is and readability doesn’t really differ with syntax highlighting on,
so you are right and I’ve changed it, thanks.

What do you think about an explicit `then` keyword for `if` conditions? I do
not use it, because I found only two occurrences (`autosetup/cc-lib.tcl`) in
all Tcl scripts.

floand:
Hi @flatcap,

can you enlighten me please? I miss my last two commits:
*   [a33258ec](https://github.com/neomutt/neomutt/commit/a33258ecd5304d980ec57234d7815318eeb3505d)
*   [9ef9e3e6](https://github.com/neomutt/neomutt/commit/9ef9e3e64fce6127a13177994a815530719bb047)

in this [PR #1105](https://github.com/neomutt/neomutt/pull/1105) (pushed to branch [devel/docs_sort-resources](https://github.com/neomutt/neomutt/tree/devel/docs_sort-resources) this morning).

The former is unimportant because it [fails to build](https://travis-ci.org/neomutt/neomutt/builds/357684832) but the latter [corrected](https://travis-ci.org/neomutt/neomutt/builds/357693768)
my mistake and applies a suggestion of @gahr.

Also all hashes of the remaining commits are changed now like after a
rebase. Should I also rebase locally and push again?

gahr:
On a second thought, I think it’s fair to assume that

* `PKG_VERSION` is defined
* it has a fixed format
* clock scan on it won’t fail

so, I would not expose format as an argument and I would not wrap scannig `PKG_FORMAT` in `catch`.

What do you think?

flatcap:
@floand Argh! Sorry.
I rebased all the branches to the new release.
You must have committed in the small window between pulling and pushing.
Please push then again.

floand:
Hi @gahr,

thanks for your responses. I have to admit I’m a little bit confused, sorry. With “`else`
clauses” did you mean my question about what you think about usage of “`then`
keyword”? If yes, I just asked it only to learn if:
```tcl
# it is better to write:
if { ... } then { ... }
# or just:
if { ...} { ... }

# because I used, for instance:
if {$args eq "" || [catch {
  set fd [open "|git log -1 --format=%at master -- $args" r]
  gets $fd epoch
  close $fd}] || $epoch eq ""
} {
  ...
}
```
and I’m not sure `} {` can be clearly determined as start of the True branch
of the `if` condition like a `} then {` notation.

---

> On a second thought, I think it’s fair to assume that
>
> * PKG_VERSION is defined

I assume here and below you mean `PACKAGE_VERSION`.

> * it has a fixed format
> * clock scan on it won’t fail

Yes, I think also it always outputs something. But is it the expected result, when
it is used without the `-format` constraint? When I locally use `tclsh` (version 8.6),
I get:
```tcl
clock scan 20180323
1521759600
clock format 1521759600 -format "%Y%m%d"
20180323    # This looks OK, but also if you live in australia and it is a commit date?
clock format 1521759600 -format "%Y%m%d" -timezone :UTC
20180322    # This is wrong now, it is != 20180323
```

>
> so, I would not expose format as an argument and I would not wrap scannig PKG_VERSION > in catch.
>
> What do you think?

As an argument to **a.)** `proc get-author-date` or **b.)** to `clock scan`?
**a.)** I thought it could be useful for future stuff, e.g. replacing “Unix” on bottom left
of the manual pages with “NeoMutt 20180323” or similar. 
**b.)** I read about [clock](https://www.tcl.tk/man/tcl8.6/TclCmd/clock.htm#M80) and a free form scan is mentioned as “This form of scan is deprecated”
there, so I used `-format`.

I’m really hope not to raise confusion with this longer reply. ;-)

floand:
Hi @flatcap,

>I rebased all the branches to the new release.
>You must have committed in the small window between pulling and pushing.

thanks for the insight.

> Please push then again.

I’ll do when confusion is gone. ;-)

gahr:
Oh, sorry, I was the one confused. Writing from mobile while shopping :-/

* Please forget everything I said about `format`. I see you’re using the param for the final formatting, not for scanning. What you’re doing is fine :)

* I personally have never used `then`, which is probably why my mind read your comment and interpreted it as meaning `else`.. I don’t see `then` used (much) anywhere in Tcl code, so I’d just do without. 

Sorry for the confusion.

floand:
Hi @gahr,

thank you for sharing your opinion about Tcl syntax. This is all quite new to me.
Learning by reading docs is good, from practical experiences is better.

I’ll rebase my local branch and re-push my last commit tomorrow with your
`[catch {clock scan [get-define PACKAGE_VERSION] -format {%Y%m%d} -timezone :UTC} epoch]` suggestion applied.

>Sorry for the confusion.

All is fine and I’m happy now.

gahr:
Thanks, I still think that `[clock scan]` cannot fail in this particular instance, so I'd even go without a `[catch]`.

floand:
>I still think that [clock scan] cannot fail

Just to be sure, without `catch` and presumed we can really be sure (could we?),
`PACKAGE_VERSION` is:

*   always defined
*   interpretable as a date – `23.03.2018` for example, will fail

we’re back using `set` then, though even simpler:

```tcl
if {$args eq "" || ...
} {
  #user-notice "No author timestamp found, falling back to PACKAGE_VERSION..."
  set epoch [clock scan [get-define PACKAGE_VERSION] -timezone :UTC] 
}
cd $oldpwd
return [clock format $epoch -format $format -timezone :UTC] 
```

because we assume `clock scan` **and** `set` wont fail, we can also remove the
`autosetup-error` statement. But this again, is the earlier mentioned deprecated
free-form scan. So we probably better do:

`set epoch [clock scan [get-define PACKAGE_VERSION] -format {%Y%m%d} -timezone :UTC]`

with a now useless (after presumption above) `-format {%Y%m%d}` constraint?

On the other side, I don’t know what to believe, because [1] and [2] also say:

>“If free-form scan is used, only the **-base** and **-gmt** options are accepted.
>The **-timezone** and **-locale** options will result in an error if **-format** is not supplied.”

This seems only be true for the **-locale**, not for **-timezone** option – at least
on my box. I’ve tested several date formats and time zones, e.g.: 

```tcl
clock scan "03/23/2018" -locale en_AU 
legacy [clock scan] does not support -locale
clock scan "03/23/2018" -timezone :America/Vancouver
1521788400
```

What do you think?


References:

[1] <https://www.tcl.tk/man/tcl8.6/TclCmd/clock.htm#M80>
[2] <https://core.tcl.tk/tips/doc/trunk/tip/173.md>

floand:
@gahr, tested and done, thanks again.

gahr:
Thanks, autosetup stuff looks good :)

gahr:
I have had a closer look at this, not from mobile this time.. 

unfortunately there are a number of problems in this implementation, none of which are your fault. The bundled jimsh implementation doesn't support `[open "|..."]` nor `[clock scan`].

Let me have a thought and get back with ideas.

gahr:
Anyway, I was about to propose the following simplification to your get-author-date function:

```tcl
proc get-author-date {format args} {
  set epoch [readfile "|git -C $::autosetup(srcdir) log -1 --format=%at master -- $args"]
  if {$epoch eq {}} {
    set epoch [clock scan [get-define PACKAGE_VERSION] -format {%Y%m%d} -timezone :UTC]
  }
  clock format $epoch -format $format -timezone :UTC
}
```

With the difference that not passing any files gets the latest commit to the repository instead of `PACKAGE_VERSION`, which is imho fine. You'd revert to `PACKAGE_VERSION` only when running `./configure` off a non-git-repo directory.

floand:
>The bundled jimsh implementation doesn't support `[open "|..."]` nor `[clock scan]`.

Oops, I wasn’t aware of that and that I have to make my locally installed `tclsh` invisible
for `autosetup/autosetup-find-tclsh` to use the bundled jimsh. Did I understand it right,
when strptime (time.h) is available, we can use `[clock scan]`?

Your snippet looks nice – a week more and we’ve squashed it to a one-liner. ;-)
I’ll do some tests tonight and hopefully the final solution comes closer.

flatcap:
Hi @floand 
What's the status on this PR?

floand:
Hi @flatcap,
>What's the status on this PR?

I use @gahr’s suggestion from #1117, made some changes to comments,
adapted the configure command definitions and rewrote the manual pages
for `neomutt(1)`and `neomuttrc(5)`. I hope manual pages are more up to date
and consistent to the full manual now.

What is not finished yet, is to complete section “Command-Line Options” in
full manual and sorting of the `neomutt -h` output – similar to `neomutt(1)`.

Also, I would like to reformat the manual (probably in an other branch/PR)
so it is more eye candy, easier to read and visually ‘searchable’.

flatcap:
Sounds brilliant. :+1:

floand:
Then I have a question because I saw different syntax in the manual.
For instance in section [“Using Hooks”](../blob/1d811a736c4e01307aeaa17526375b8a9f8924ef/doc/manual.xml.head#L7285):
```xml
<link linkend="account-hook">
  <command>account-hook</command>
</link>
```
and [“Configuration Commands”](../blob/1d811a736c4e01307aeaa17526375b8a9f8924ef/doc/manual.xml.head#L15292):
```xml
<command>
  <link linkend="account-hook">account-hook</link>
</command>
```
Which one to prefer? Can we do just:
```xml
<command linkend="account-hook">account-hook</command>
```
or will it break the website generation?

flatcap:
> I saw different syntax in the manual

Well spotted.

> Can we do just:
> `<command linkend="account-hook">account-hook</command>`

Unfortunately, DocBook doesn't allow that.

> Which one to prefer?

After a quick bit of vim-fu, I've counted the examples:
- 73 `<command><link>...</link></command>`
- 53 `<link><command>...</command></link>`

That's quite a lot of work to unify them.  Leave them for now.

Much as I like DocBook, it's an unfriendly language for beginners.
If the docs were in markdown, then ordinary users could contribute more easily.

I have some **big** ideas on how to make the docs friendly and accessible.
(not yet discussed, so they may just be crazy :-)

floand:
@flatcap,
thanks for the rebase.

I think old code came back with commit 9095c439 previously deleted in 4c04c3ac.

floand:
>I have some big ideas on how to make the docs friendly and accessible.
(not yet discussed, so they may just be crazy :-)

CC me when time has come.

flatcap:
Sorry about the merge mistake.
I tried using GitHub's "Resolve Conflict" button -- I shalln't to it again :-)

> CC me when time has come.

Will do.  I'll type up my notes and open a discussion issue.

floand:
Hi @flatcap,

I reformatted doc/manual.xml.{head,tail} in branch
[doc/reformat_xml-manual](https://github.com/neomutt/neomutt/tree/doc/reformat_xml-manual)
based on current master (785ddc222), but don’t open a pull request for it to not
pollute the PR list. 

I would like to know if that branch has a chance for a merge into master, so
I know also if _this PR_ beast (that I think is ready from my part) needs further
modifications, because currently both branches conflicts to each other.

Thanks

flatcap:
> I would like to know if that branch has a chance for a merge into master

I've just been waiting for you to stop moving -- you've been a fast-moving target.
I shall take a good look shortly.

> both branches conflicts

I've made changes to the `init.h` docs for logging and config precendence.
They should be minor.  I'll fix them.

flatcap:
The xml in [doc/reformat_xml-manual] looks much better. How did you do it?
(The existing formatting was done with [HTML Tidy](http://www.html-tidy.org/))

floand:
> I've just been waiting for you to stop moving

I will do so ;-) to think about xml -> md conversion for [#1133](https://github.com/neomutt/neomutt/issues/1133#issuecomment-382698161).
Having a read-only mailbox for help/manual is a really nice idea.

> I've made changes to the `init.h` docs for logging and config precendence.

Yes, `init.h` is minor – your changes are simply preferred because I added only
format macros for manual pages and a one-liner about `-l` option. But I mean
the not so simple conflicts in `doc/manual.xml.head` – one branch has the tags
reformatted while this PR uses currently old/different tag format, re-sorts and
has some additional changes.

My idea is that I modify this PR shall branch [doc/reformat_xml-manual](https://github.com/neomutt/neomutt/tree/doc/reformat_xml-manual)
be merged, so you don’t have to bother and wasting time with my conflicts.

floand:
> The xml in [doc/reformat_xml-manual] looks much better. How did you do it?

Vim with regular expresssions and recorded key-macros. I tested `xmllint`,
`xml_pp` and `HTML tidy` but could not get the control I want – maybe
I spend not enough time to find _the_ way.

flatcap:
> shall branch doc/reformat_xml-manual be merged

Done

floand:
> Ooh! Is it finished?

Normally I would say yes (finished in the sense ready for review), but unfortunately
I changed the `puts(_("some string"));` without thinking (and knowing about `_()` and
`N_()` at that time) in:

*   https://github.com/neomutt/neomutt/commit/cdad573e9b01c0dba21d87fd4534a5b78603c871#diff-2045016cb90d1e65d71c2407a2570927

How stupid! Now I have to figure out the whole translation and `*.po`-updating
machinery first, before it's really finished. Kicks in the right direction are welcome. ;-)

flatcap:
Generally speaking: `make update-po` will scan the source and update the message strings.
**Note**: it also updates all the line numbers which makes the diffs huge and hard to merge.

@reisradomil has made huge changes in another branch (#1156), so I don't want any conflicts in that dir (if it can be helped).

In your case, you've rearranged a lot of existing info, so you'd need to use some vim-fu on all 30 of the po files.
**I'll sort this out** after I've merged reisradomil's branch (which is also ready).

floand:
> `make update-po` will scan the source ...

Thanks a lot, that are good news because I saw the gettext branch and its thousands
of changes and thought by me there must be a kind of machinery in the background.

I'll change back the `puts(_(` and `));` in `main.c` to their old positions first and try
your mentioned command before looking for the po-files. Thanks.

flatcap:
I've rebased onto master.
These commits are all your work except the `_()` changes to `main.c` that you reverted.

floand:
> I've rebased onto master.

Thanks, I have a basic question about that because I was in (inner) conflict how
to solve the file conflict. Firstly, I don’t wanted the second merge bubble in three
days, secondly, I don’t trust the “Resolve Conflict” button.

Is it OK to rebase a branch onto master whenever a conflict comes up?

flatcap:
> Is it OK to rebase a branch onto master whenever a conflict comes up?

Yes.  It's your branch (until the point of merging), so you're welcome to rebase, reorder commits, etc.

NeoMutt's development speed is slow enough that I can "massage" the commits into neat side-branches.
(Some day, I may have to accept chaos :-)

flatcap:
If you'd like to push your changes to main into a branch, we can get them merged too.
Then I can update the translators' branch.
Thanks.

floand:
> so you're welcome to rebase, reorder commits, etc.

That will make things easier and I wont steal your time.

> If you'd like to push your changes to main into a branch, we can get them merged too.
> Then I can update the translators' branch.

Not sure what you exactly mean with _my changes_. Those for `neomutt -h` output
in `main.c` that you’ve _not_ rebased ([cdad573e9b](https://github.com/neomutt/neomutt/commit/cdad573e9b01c0dba21d87fd4534a5b78603c871#diff-2045016cb90d1e65d71c2407a2570927))?

flatcap:
Yes, they're the ones: `usage()`

If you'd like to commit those, now that everything's calmed down.
(and see if there's anything in Reis' changes you like).

flatcap:
But, please don't edit the `.po` files.
We'll do that when we're happy with the wording.

floand:
> If you'd like to commit those […]

Yes, I will to close the discrepance between `neomutt -h` and the guide/**neomutt**(1)
for the command line options.

> But, please don't edit the .po files.

So I create a new branch and will commit my changed help string (`usage()`) and the
raw result of `make update-po` to it for wording review!?

flatcap:
Just the code, please.  Nothing else.

floand:
Done in: [devel/usage-string](https://github.com/neomutt/neomutt/tree/devel/usage-string).

PR #1106, opened 2018-03-21
use sleep_time as error message wait time instead of hardcoding it to 1s

royger:
Also fixes an error in the conversion from us to ns.


flatcap:
I made a small fix.
`nanosleep()` needs the time split into seconds and nanoseconds, otherwise times >1sec fail.
Merged into #1109.

PR #1107, opened 2018-03-21
Add support for direct color values when using slang

royger:
slang supports using direct color values in order to create color pairs, this means that neomutt is not limited to the console color palette.

The same is not possible with ncurses, since it doesn't support direct-color values ATM. The only way to have custom colors in ncurses is to set them in the terminal palette, which is very antisocial because there's no way to revert back to the original palette afterwards, leaving the console after quitting neomutt in a weird state (color-wise).

This series adds supports for RGB color values in the neomutt config file, like the following example:

https://gist.github.com/royger/80c8aeff38567aa47274165f2da19117

This is the solarized dark color scheme from:

https://github.com/altercation/mutt-colors-solarized/blob/master/mutt-colors-solarized-dark-256.muttrc

This is a minimal change in order to get some RGB color support, in order to get something similar with ncurses neomutt will have to print it's own escape sequences interleaved with ncurses, which is far from trivial AFAICT.

### DIFF
--- a/auto.def
+++ b/auto.def
@@ -592,6 +592,7 @@ switch [opt-val with-ui ncurses] {
     }
     define USE_SLANG_CURSES
     define-feature COLOR
+    define-feature DIRECTCOLOR
   }
 
   default {
--- a/color.c
+++ b/color.c
@@ -66,14 +66,16 @@ static int ColorQuoteSize;
 #ifdef HAVE_COLOR
 
 #define COLOR_DEFAULT (-2)
+#define COLOR_UNSET UINT32_MAX
 
 /**
  * struct ColorList - A set of colors
  */
 struct ColorList
 {
-  short fg;
-  short bg;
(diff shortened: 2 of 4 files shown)


flatcap:
This pull request **introduces 2 alerts** when merging 06229fce342f3249b87711addbc766a069d0c994 into 23d1bfb2e51c9db929adceb7a1970962b263a81b - [view on LGTM.com](https://lgtm.com/projects/g/neomutt/neomutt/rev/pr-ad2aa52a256b01fb808245fc6c03984ec76d4186)

**new alerts:**

* 2 for Unsigned comparison to zero

---

*Comment posted by [LGTM.com](https://lgtm.com)*

flatcap:
Hi @royger.
Sorry for sitting on this forever.

I wasn't happy, initially, because this code only works for slang
(and I'd rather try to concentrate our users onto ncurses).

That stopped me merging a perfectly good piece of code.

Sorry and thanks.

PR #1110, opened 2018-03-24
fix Notmuch deprecation warning

flatcap:
We need to distinguish between two versions of `notmuch.h`.
Both declare that they are v5.0.0.  They are:

- libnotmuch 5.0 (Notmuch 0.25)
  - `notmuch_database_add_message()` is a normal function
  - `notmuch_database_index_file()` doesn't exist

- libnotmuch 5.1 (Notmuch 0.26)
  - `notmuch_database_add_message()` is **deprecated**
  - `notmuch_database_index_file()` exists

To work around this, we check for `notmuch_database_index_file()` specifically.

<sub>Edited to match new commit</sub>

### DIFF
--- a/auto.def
+++ b/auto.def
@@ -429,17 +429,14 @@ if {[get-define want-notmuch]} {
   }
   define USE_NOTMUCH
   msg-checking "Checking for Notmuch API version 3..."
-  if {[cctest -includes {notmuch.h} \
-              -code {
-                notmuch_database_open("/path",
-                                      NOTMUCH_DATABASE_MODE_READ_ONLY,
-                                      (notmuch_database_t**)NULL);
-              }]} {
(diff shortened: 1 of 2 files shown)


flatcap:
@jelly 
Yeah, you're right.  Both 0.25 and 0.26 claim to be v5.0.0, but have different functions.
This new commit checks for the new function specifically.

@gahr 
I've also made a change to one of the Notmuch `auto.def` checks.
I've added `-libs {-lnotmuch} -link 1` to the `cc-test`.

"**Checking for Notmuch API version 3...**"
The test only **compiled** the code, so it always succeeded.
If the function was missing, there'd be warnings, but it was still successful.

The alternative is to add `-nooutput 1` to make any warnings fatal.
But... I wondered if the `NULL`s might cause trouble and need `#include`s of their own.

jelly:
@flatcap Looks good to me.

PR #1113, opened 2018-03-25
fix return code for 'neomutt -Z'

### LINKED ISSUE #1112: Regression: `neomutt -Z` no longer gives non-zero exit codes
## bug reports

* Expected behavior: If `neomutt -Z` fails to launch due to there being no new mail, it should return a non-zero exit code.
* Actual behavior: If `neomutt -Z` fails to launch due to there being no new mail, it returns zero as an exit code.
* Steps to reproduce: Have no new mail. Run `neomutt -Z`.
* Used program versions: regression observed in `20180323`. Previous version `20180223` acted as expected.
* Operating System and its version: up-to-date Arch Linux.

## Rationale

`neomutt -Z` is very useful. If you have unread mail, neomutt will launch and automatically select the new mail. However, if there is no new mail, neomutt will fail to launch, and instead give the following output:
```
$ neomutt -Z
No mailbox with new mail.
```
Previously in neomutt `20180223` and earlier, if neomutt failed to launch (due to no unread mail), the exit status would be `1`. This meant that users could launch neomutt with an alias of `neomutt -Z || neomutt`. i.e. if there were no unread mail, then the vanilla `neomutt` command would run anyway.

However, from version `20180323`, after neomutt fails to launch (due to no unread mail), the exit status in now `0`, which falsely indicates success. This makes the previous alias unusable.

Ideally neomutt should return a non-zero exit code when it does not launch. In addition, there could be an argument for `neomutt -Z` to launch in a vanilla sense anyway, even if there are no new emails.

## EDIT

Also, I can't even use a workaround like `neomutt -Z | grep -q "No mailbox with new mail\." && neomutt`. This fails because whenever neomutt gets piped somewhere else, there is no inferface shown at all.

flatcap:
`neomutt -Z` should return `1` if there's no new mail.
This allows users to use the check in a script.

Fixes #1112

### DIFF
--- a/main.c
+++ b/main.c
@@ -532,8 +532,7 @@ int main(int argc, char **argv, char **env)
   }
 
   /* set defaults and read init files */
-  rc = mutt_init(flags & MUTT_NOSYSRC, &commands);
-  if (rc != 0)
+  if (mutt_init(flags & MUTT_NOSYSRC, &commands) != 0)
     goto main_curses;
 
   /* The command line overrides the config */


flatcap:
I've tested all 35 routes through the code.
They all return the same values as they did before the startup changes.

PR #1117, opened 2018-03-30
 Simplify generation of man date

gahr:
In case we want a simpler solution by which the dates in man pages is always consistent with the release date...

### DIFF
--- a/auto.def
+++ b/auto.def
@@ -176,24 +176,6 @@ if {1} {
   proc yesno val {
     expr {$val ? "yes" : "no"}
   }
-
-  # Recent author timestamp of file(s) as UTC, use PACKAGE_VERSION as fallback
-  proc get-author-date {format args} {
-    set oldpwd [pwd]
-    cd $::autosetup(srcdir)
-
-    if {$args eq "" || [catch {
-        set fd [open "|git log -1 --format=%at master -- $args" r]
-        gets $fd epoch
-        close $fd}] || $epoch eq ""
-    } {
-      set epoch [clock scan \
-                   [get-define PACKAGE_VERSION] -format {%Y%m%d} -timezone :UTC]
-    }
-    cd $oldpwd
-
-    return [clock format $epoch -format $format -timezone :UTC]
-  }
 }
 ###############################################################################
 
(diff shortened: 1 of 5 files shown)


gahr:
@floand @flatcap thoughts?

flatcap:
@floand 
Thanks for your work figuring out tcl.
It's a good solution, but I think it's overkill for the few man pages that we have.
Sorry that it's taken so long to decide that.

I think that @gahr's solution is better for this case.
A simple timestamp that matches the latest release.

floand:
Hi @gahr,

>In case we want a simpler solution by which the dates in man pages is 
>always consistent with the release date...

It works, I like it and it should be used. 

Because we have to be aware of the lowest common denominator regarded 
to installed Tcl versions and the bundled Jim Tcl interpreter, the latter 
might lead to rather ugly and complicated workarounds than clean helper 
procedures, for instance:

*   `readfile "|git ..."` cannot be used – it always return the empty 
    default string because `popen` (exec/pipe wrapper for `open "|..."`) 
    always fails with `invalid command name "socket"` at line 123 of 
    file tclcompat.tcl [1]
*   `clock scan` isn’t available and even it were, it wouldn’t support 
    the `-timezone` option – same for `clock format` [2]
*   `clock add` (as workaround for `-timezone` calculation) isn’t implemented

References:
[1] <https://github.com/msteveb/jimtcl/blob/ba3431e8a4/tclcompat.tcl#L123>
[2] <http://jim.tcl.tk/fossil/doc/trunk/Tcl_shipped.html#_clock>

---
Hi @flatcap,

>Thanks for your work figuring out tcl.
>It's a good solution, but I think it's overkill for the few man pages 
>that we have.

I don’t think so, because my approach was to provide one procedure for 
all future timestamp related tasks with a fallback to PACKAGE_VERSION.
But it’s hard to achieve (see above), so I’m convinced now to use a 
simple solution instead of fighting against limitations of not fully Tcl 
compatible implementations.

That being said, I would like to merge @gahr's commits and continue, e.g. 
to complete `neomutt.1` by adding missing options, environment variables, 
etc. and sorting the man page sections itself.

auto.def
```diff
@@ -902,8 +884,8 @@ foreach dir $subdirs {
 ###############################################################################
 # Provide timestamps (UTC) based on most recent author date or PACKAGE_VERSION
 # doc/neomuttrc.man, doc/neomutt.1
```
floand: I would also change the comment to something like:

Define package timestamp (UTC) based on PACKAGE_VERSION for:
doc/neomuttrc.man, doc/neomutt.1

auto.def
```diff
@@ -915,7 +897,6 @@ make-config-header config.h -auto $auto_rep -bare $bare_rep -str $str_rep
 # Generate doc/neomutt.1
 define bindir [get-define BINDIR]
 define docdir [get-define PKGDOCDIR]
```
floand: Aren’t `bindir` and `docdir` obsolete now, since `neomutt.1` is no longer generated here and
`Makefile.autosetup` defines them?

gahr:
> Aren’t bindir and docdir obsolete now, since neomutt.1 is no longer generated here and
Makefile.autosetup defines them?

yep, pushed thanks

Issue #1118, opened 2018-03-30
Segmentation fault when starting neomutt


Herny64:
## bug reports

* Expected behavior
neomutt should not crash
* Actual behavior
it crashes with segfault 
* Steps to reproduce
just start neomutt
* Used program versions
neomutt -v output:
```
System: Linux 4.15.14-gentoo (x86_64)
ncurses: ncurses 6.1.20180127 (compiled with 6.1)
hcache backends: gdbm, bdb

Compiler:
x86_64-pc-linux-gnu-gcc (Gentoo 7.3.0 p1.0) 7.3.0
Copyright (C) 2017 Free Software Foundation, Inc.
This is free software; see the source for copying conditions.  There is NO
warranty; not even for MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.

Configure options: --prefix=/usr --build=x86_64-pc-linux-gnu --host=x86_64-pc-linux-gnu --mandir=/usr/share/man --infodir=/usr/share/info --datadir=/usr/share --sysconfdir=/etc --localstatedir=/var/lib --docdir=/usr/share/doc/neomutt-20180223 --libdir=/usr/lib64 --disable-doc --disable-nls --enable-notmuch --enable-gpgme --enable-pgp --disable-smime --enable-bdb --enable-gdbm --disable-kyotocabinet --disable-qdbm --disable-tokyocabinet --disable-idn --disable-gss --disable-lmdb --disable-sasl --with-ui=ncurses --sysconfdir=/etc/neomutt --enable-ssl --disable-gnutls

Compilation CFLAGS: -march=haswell -O2 -pipe -std=c99 -fno-delete-null-pointer-checks -D_ALL_SOURCE=1 -D_GNU_SOURCE=1 -D__EXTENSIONS__ -I/usr/include -DNCURSES_WIDECHAR  -I/include -I/usr/include/db5.3

Default options:
  +attach_headers_color +compose_to_sender +compress +cond_date +debug
  +encrypt_to_self +forgotten_attachments +forwref +ifdef +imap +index_color
  +initials +limit_current_thread +multiple_fcc +nested_if +new_mail +nntp +pop
  +progress +quasi_delete +regcomp +reply_with_xorig +sensible_browser +sidebar
  +skip_quoted +smtp +status_color +timeout +tls_sni +trash

Compile options:
  +bkgdset +color +curs_set +fcntl -flock -fmemopen +futimens +getaddrinfo
[...]

PR #1119, opened 2018-03-31
Build: restructure ncurses/tinfo checks

gahr:
Issue #1118


gahr:
I have tested this on FreeBSD with curses 5.9 and 6.1, and on NetBSD with curses (not ncurses) from base. They both work.

@WPettersson @Herny64 could you please give this branch a try? Thanks!

WPettersson:
This branch compiles+seems to run fine on my system

PR #1122, opened 2018-04-01
Describe <show-messages> more precisely

reisradomil:
The term 'message' is used for log-messages as well as for ordinary
mails. In the help 'message' usually means an email. Adjust the
description of <show-message>/OP_SHOW_MESSAGE to clarify that log and
debug messages are meant.

### DIFF
--- a/curs_main.c
+++ b/curs_main.c
@@ -1408,7 +1408,7 @@ int mutt_index_menu(void)
         break;
 #endif /* USE_POP */
 
-      case OP_SHOW_MESSAGES:
+      case OP_SHOW_LOG_MESSAGES:
       {
         char tempfile[PATH_MAX];
         mutt_mktemp(tempfile, sizeof(tempfile));
--- a/doc/manual.xml.head
+++ b/doc/manual.xml.head
@@ -2964,6 +2964,47 @@ color sidebar_divider   color8  default
       <para>See also the
       <link linkend="postpone">$postpone</link> quad-option.</para>
     </sect1>
+
+    <sect1 id="logging">
+      <title>Logging</title>
+      <para>NeoMutt has different types of logging/error messages</para>
+      <itemizedlist>
+        <listitem>Primitive Errors:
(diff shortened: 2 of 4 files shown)


reisradomil:
Hi guys,

The second commit renames the function and is thus an UI breakage. Feel free to reject this if you don't seem this is worth the gain.

Best
Reis

flatcap:
Thanks.  I'm tempted to change it slightly to:
- "show log (and debug) messages"
making the "and debug" a sort of after-thought.
What do you think?

As for changing the name of the command: Yes, ok.
It's a new command, few people will be using it, fewer still will have rebound the default key.

reisradomil:
> Thanks. I'm tempted to change it slightly to: "show log (and debug) messages"

Good idea. I updated the pull request accordingly.

Initially I wanted to document the new function `<show-messages>` in the manual/Neomutt Guide since a one line documentation seems to me very thin (for example the commit command mentions that only the last 100 log entries are shown and debug is only shown if it was enabled at compile time).
However, it seems as the part with the function reference [0] is generated automatically. Is there any place where functions are documented (besides for individual features like [1])?

[0] https://www.neomutt.org/guide/reference
[1] https://www.neomutt.org/feature/limit-current-thread

flatcap:
> one line documentation seems to me very thin

Yes, you're right.  For many commands, that one line is all there is.
I've added an empty section "Logging" to the "Getting Started" chapter.
If you'd like to add a few sentences there.

Things we might want to mention (and link to):
- Messages vs Errors
- Where they appear
- How to list the latest messages (and how many)
- How to colour the messages / errors
- How to enable debugging (file and level)
- Where debugging is saved

Thanks

reisradomil:
I've written a small section. Please review. Thanks. In particular the description of the perror. I've found no documentation about it and that's what I've deduced from the source.

flatcap:
That's a good start, thanks.
As you've asked for a review of the logging, we'll close this PR.

PR #1130, opened 2018-04-06
message.c: retry message parse if first attempt is zero-length

alyptik:
On IMAP failure the message body can sometimes be truncated, which
leaves an empty message in the cache that must be manually deleted
to be properly re-fetched.

If the cached message body is zero-length, retry the message parse
once to be sure that the message body actually *is* empty.

### DIFF
--- a/imap/message.c
+++ b/imap/message.c
@@ -989,6 +989,7 @@ int imap_fetch_message(struct Context *ctx, struct Message *msg, int msgno)
   unsigned int uid;
   int cacheno;
   struct ImapCache *cache = NULL;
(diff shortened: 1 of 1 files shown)


flatcap:
Hi @alyptik, thanks for the patch.
That seems like a reasonable thing to do.

Of course, if you can find and fix the cause (as well), that'd be even better :-)

flatcap:
Hi @alyptik 
Sorry it's taken a while to check this.
Merged.  Thanks.

alyptik:
No worries, cheers.

PR #1134, opened 2018-04-09
minor IMAP string handling fixes

alyptik:
While looking for the cause of my IMAP truncation bug I noticed a few minor places that needed minor fixes along the way (mostly snprintf related) that were a little sparse on error checking and other small things like nesting confusing expressions a bit too deeply and incorrect return types.

The commits are all fairly small and simple to read over except for maybe:

* commit [f8385f5](https://github.com/neomutt/neomutt/commit/f8385f5bbd22d563a5390183b45ff9147994452b) (message.c: replace nested conditional operators with static string table)

which I couldn't really figure out any good ways to simplify without drastically affecting semantics.


bcache.c
```diff
@@ -70,7 +70,7 @@ static int bcache_path(struct Account *account, const char *mailbox, char *dst,
   }
 
   size_t mailboxlen = mutt_str_strlen(mailbox);
-  len = snprintf(dst, dstlen - 1, "%s/%s%s%s", MessageCachedir, host, NONULL(mailbox),
```
gahr: Using `dstlen` is correct. The only caller uses `sizeof(bcache->path)`, so this argument is really the number of bytes available in the buffer, not the length of the longest C string that would fit in there.

imap/message.c
```diff
@@ -693,7 +694,11 @@ int imap_read_headers(struct ImapData *idata, unsigned int msn_begin, unsigned i
     mutt_progress_init(&progress, _("Evaluating cache..."), MUTT_PROGRESS_MSG,
                        ReadInc, msn_end);
 
-    snprintf(buf, sizeof(buf), "UID FETCH 1:%u (UID FLAGS)", uidnext - 1);
+    if (snprintf(buf, sizeof(buf), "UID FETCH 1:%u (UID FLAGS)", uidnext - 1) < 0)
```
gahr: Is there any reasonable situation in which this snprintf could fail, that you can think of?

imap/message.c
```diff
@@ -1051,10 +1061,21 @@ int imap_fetch_message(struct Context *ctx, struct Message *msg, int msgno)
    * command handler */
   h->active = false;
 
-  snprintf(buf, sizeof(buf), "UID FETCH %u %s", HEADER_DATA(h)->uid,
-           (mutt_bit_isset(idata->capabilities, IMAP4REV1) ?
-                (ImapPeek ? "BODY.PEEK[]" : "BODY[]") :
-                "RFC822"));
+  /* the IMAP4REV1 predicate being true will always result in one of our two
```
gahr: Sorry, but I find the original code clearer.

imap/message.c
alyptik: That's a good point, I'll revert this one.

imap/message.c
alyptik: Figured this might be a bit too weird; will revert this as well.

alyptik:
Dropped the requested commits.

Although I wasn't sure if the comment on:

* de2bc27 (bcache.c: fix off-by-one sprintf() dest buffer size error)

was requesting any changes?

gahr:
Yes it was :) the code is correct as it is. I would also ask you to revert the reindentation of the header file. Thank you!

alyptik:
No problem; header change has been reverted.

alyptik:
Whoops, I reverted the prototype as well by accident, fixed.

gahr:
@alyptik sorry for the confusion about your change to `bcache_path`. I read the diff the other way around: your change is correct.

alyptik:
No worries, cheers.

PR #1142, opened 2018-04-15
init.c: Ensure more consistent variable dumps

floand:
* **What does this PR do?**

This avoids blank lines and use double quotes for the `neomutt -D -S`
variable dump as in `neomutt -D`.

It also gives a little bit more detail on sensitive variable values and
should be unique enough to allow filtered results easily.

### DIFF
--- a/init.c
+++ b/init.c
@@ -3669,7 +3669,7 @@ int mutt_dump_variables(int hide_sensitive)
 
     if (hide_sensitive && IS_SENSITIVE(MuttVars[i]))
     {
-      mutt_message("%s='***'\n", MuttVars[i].name);
+      mutt_message("%s='***'", MuttVars[i].name);
       continue;
     }
     snprintf(command, sizeof(command), "set ?%s\n", MuttVars[i].name);


floand:
> […] that is the only reliable way to differentiate between a concealed and a non concealed option

You are right, after a second thinking, I get it.

> I'm kind of neutral towards what is inside the single quotes: *** or *** concealed *** are both fine.

Me too, I use it only to get a more unique and visual identifier after changing
quotes from single to double.

I have a question, because I would like to change back to single quotes. I saw
different syntax:
```c
mutt_message("format sample %s", any->data);
```
and also:
```c
mutt_message(_("format sample %s"), any->data);
```
Which one is preferred? Thanks.

gahr:
The second form is used when the string is going to be translated. Ideally all messages with real-language content (e.g., not a variable name and its value) would have a translation, but I don’t think that’s the case in our code base.

flatcap:
`_()` and `N_()` are macros for `gettext()` -- [mutt/message.h](https://github.com/neomutt/neomutt/blob/master/mutt/message.h)
It translates strings at runtime.

floand:
I searched for it with a wrong regex, so couldn’t found and overlooked it. :-/
Thank you both for the explanation. Now I know for what to look for I also get
#376 from ChangeLog.md file. Thanks again.

flatcap:
I've chosen to drop the word 'concealed'.
Using just stars means we don't need to bother translating anything.

PR #1143, opened 2018-04-15
Alternative OP_JUMP

floand:
Hi @flatcap,

I push this for an easier review and just as a reminder for my patch
starting at neomutt-devel mailing list thread [1].

References:
  [1] <http://mailman.neomutt.org/pipermail/neomutt-devel-neomutt.org/2018-March/000583.html>

### DIFF
--- a/curs_main.c
+++ b/curs_main.c
@@ -1412,11 +1412,14 @@ int mutt_index_menu(void)
         if (isdigit(LastKey))
           mutt_unget_event(LastKey, 0);
         buf[0] = 0;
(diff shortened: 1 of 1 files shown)


flatcap:
I've just had a close look at this.  Very nice.
If you've nothing more to add, I'll merge it.

floand:
> If you've nothing more to add

It works for me without issues, the newly added error message “Nothing to do.”
already exists in translations, so I shan’t add/change something.

> I'll merge it.

I shall be glad then, thanks.

PR #1157, opened 2018-04-22
mutt/file.c: split mutt_file_{lock,unlock} into flock/fcntl versions

### LINKED ISSUE #1147: Split up mutt_file_lock() and mutt_file_unlock()
This is an "easy" task ([find more](https://github.com/neomutt/neomutt/issues?q=label:"good%20first%20issue")) and it's open to everyone.  [_read more..._](https://www.neomutt.org/dev/easy-tasks)

## Description

NeoMutt has two file-locking systems: `fcntl()` and `flock()`.  
It's not recommended to use both at once.

### Part 1: Split `mutt_file_lock()` into two functions

There's very little overlap between the two locking mechanisms, so create two versions of `mutt_file_lock()`.  
Put a `#ifdef` block around each one.

### Part 2: Split `mutt_file_unlock()` into two functions

Same as part 1 for `mutt_file_unlock()`

### Part 3: Disallow building with both options

Change `configure.autosetup` to only allow one locking system, which defaults to `fcntl()`, e.g.

```
./configure --locking=[flock|fcntl]
```

This means changing `auto.def`.  There are already examples of this sort of option which can be cut'n'pasted.

larsks:
this splits mutt_file_lock and mutt_file_unlock into separate
fcntl() and flock() based versions.  It teaches ./configure
about the new --with-lock option (--with-lock=fcntl or
--with-lock=flock).

Includes updates to INSTALL and AUTOSETUP.md that reference the
removed --enable-fcntl and --enable-flock options.

Closes #1147

### DIFF
--- a/AUTOSETUP.md
+++ b/AUTOSETUP.md
@@ -55,7 +55,6 @@ section (it's actually a proc under the hood):
 ```
 options {
    smime=1                    => "Disable S/Mime"
-   flock=0                    => "Enable flock(1)"
    gpgme=0                    => "Enable GPGME"
    with-gpgme:path            => "Location of GPGME"
    with-mailpath:=/var/mail   => "Location of the spool mailboxes"
@@ -64,7 +63,7 @@ options {
 
 A user can configure the build to suit his needs by modifying the default
 values, e.g.,
-`./configure --disable-smime --enable-flock --gpgme --with-gpgme=/usr/local`.
+`./configure --disable-smime --gpgme --with-gpgme=/usr/local`.
 
 Within `auto.def`, option can be easily queried with `[opt-bool smime]` and
 `[opt-val with-gpgme $prefix]`, with the latter using `$prefix` if not value
--- a/INSTALL
+++ b/INSTALL
@@ -42,6 +42,12 @@ to ``configure'' to help it out, or change the default behavior:
 	will automatically look in /usr/include/ncurses for the include
 	files.
 
+--with-lock=(fcntl | flock)
+
+	select between fcntl() and flock() for file locking. Neomutt will
+	use fcntl() by default, but this can result in poor performance 
+	on NFS.
+
 --with-slang[=DIR]
 	use the S-Lang library instead of ncurses.  This library seems to
 	work better for some people because it is less picky about proper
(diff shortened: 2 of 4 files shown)


flatcap:
Welcome to NeoMuut, @larsks.
Thanks for contributing.

flatcap:
Lovely job!

I've just tested with:
- `./configure`
  uses `fcntl()`
- `./configure --with-lock`
  uses `fcntl()`
- `./configure --with-lock=fcntl`
  uses `fcntl()`
- `./configure --with-lock=flock`
  uses `flock()`
- `./configure --with-lock=fcntl --with-lock=flock`
  configure fails

We'll just see if @gahr has any comments on `auto.def`.

larsks:
Whoops, sorry for leaving those `#pragma` lines in there.  I meant to remove those.

gahr:
`auto.def` looks good, thanks! I'm wondering whether we want the double conditionals `#if defined(USE_FCNTL) && !defined(USE_FLOCK)` in the source files, though.

larsks:
> I'm wondering whether we want the double conditionals #if defined(USE_FCNTL) && !defined(USE_FLOCK) in the source files, though.

Would this be better?

```
#if defined(USE_FLOCK)

/* flock stuff goes here */

#else

/* fcntl stuff goes here

#endif
```

If `USE_FLOCK` is defined, you get `flock()`; in all other cases you get `fcntl()`.

gahr:
This makes me think: we may want to support a `--with-lock=none` for people who know what they are doing (TM). In that case:
```c
#if defined(USE_FLOCK)
..
#elif defined(USE_FCNTL)
...
#else
// lock and unlock are no-ops here
#endif
```

flatcap:
Hmm... I need to refresh my page more frequently.
I changed it to #ifdef/#else/#endif

larsks:
@gahr changed mutt/file.c per your comments.

flatcap:
Hi @larsks.

Well that seemed too easy for you.
What would you like to tackle next?

There's one (admittedly easy) but BIG task that needs tackling: #1165 
I also have some "medium" tasks in mind, but they're less well-defined.

### Merge 1

Investigate.  Merge if practical/possible.
May be simple.

Two nearly identical functions:
- `mutt_file_concat_path()` 14 uses
- `mutt_file_concatn_path()` 1 use

### Merge 2

Could be tricky to test.

In `mutt_mb_width()` we use `mbtowc()`
All the other code uses `mbrtowc()`

The man page for `mbtowc()` says:
> The function mbrtowc(3) provides a better interface to the same functionality

PR #1158, opened 2018-04-22
init.c: convert MyVars to use TAILQ

### LINKED ISSUE #1154: Convert Myvars to use STAILQ
This is an "easy" task ([find more](https://github.com/neomutt/neomutt/issues?q=label:"good%20first%20issue")) and it's open to everyone.  [_read more..._](https://www.neomutt.org/dev/easy-tasks)

## Description

Many NeoMutt structures implement their own linked-list type.  
The Myvars are simply stored as a long linked list.

`STAILQ` is a singly-linked list with pointers to both the head and the tail.  
It's defined in `mutt/queue.h`

### Part 2: Change all users to match

There are plenty of examples in NeoMutt's code of how to iterate through a list.  
**Note**: When deleting items, use `STAILQ_FOREACH_SAFE()`

z2v:
Fixes #1154

### DIFF
--- a/init.c
+++ b/init.c
@@ -99,44 +99,44 @@ struct MyVar
 {
   char *name;
   char *value;
-  struct MyVar *next;
+  TAILQ_ENTRY(MyVar) entries;
 };
 
-static struct MyVar *MyVars;
+static TAILQ_HEAD(, MyVar) MyVars = TAILQ_HEAD_INITIALIZER(MyVars);
 
 void myvar_set(const char *var, const char *val)
 {
-  struct MyVar **cur = NULL;
+  struct MyVar *myv = NULL;
 
-  for (cur = &MyVars; *cur; cur = &((*cur)->next))
-    if (mutt_str_strcmp((*cur)->name, var) == 0)
-      break;
+  TAILQ_FOREACH(myv, &MyVars, entries)
+  {
+    if (mutt_str_strcmp(myv->name, var) == 0)
+    {
+      mutt_str_replace(&myv->value, val);
+      return;
+    }
+  }
 
-  if (!*cur)
-    *cur = mutt_mem_calloc(1, sizeof(struct MyVar));
-
-  if (!(*cur)->name)
-    (*cur)->name = mutt_str_strdup(var);
-
-  mutt_str_replace(&(*cur)->value, val);
+  myv = mutt_mem_calloc(1, sizeof(struct MyVar));
+  myv->name = mutt_str_strdup(var);
+  myv->value = mutt_str_strdup(val);
+  TAILQ_INSERT_TAIL(&MyVars, myv, entries);
 }
 
 static void myvar_del(const char *var)
 {
-  struct MyVar **cur = NULL;
-  struct MyVar *tmp = NULL;
+  struct MyVar *myv = NULL;
 
-  for (cur = &MyVars; *cur; cur = &((*cur)->next))
-    if (mutt_str_strcmp((*cur)->name, var) == 0)
-      break;
-
-  if (*cur)
+  TAILQ_FOREACH(myv, &MyVars, entries)
   {
-    tmp = (*cur)->next;
-    FREE(&(*cur)->name);
-    FREE(&(*cur)->value);
-    FREE(cur);
-    *cur = tmp;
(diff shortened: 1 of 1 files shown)


init.c
```diff
@@ -99,44 +99,49 @@ struct MyVar
...
+  struct MyVar *myv = NULL;
   struct MyVar *tmp = NULL;
+  mutt_debug(2, " * myvar_del(%s)\n", NONULL((char *) var));
 
-  for (cur = &MyVars; *cur; cur = &((*cur)->next))
-    if (mutt_str_strcmp((*cur)->name, var) == 0)
-      break;
-
-  if (*cur)
+  STAILQ_FOREACH_SAFE(myv, &MyVars, entries, tmp)
```
gahr: You don’t need the _SAFE version of the macro because you’re not iterating after deletion.

init.c
```diff
@@ -99,44 +99,49 @@ struct MyVar
...
-  struct MyVar **cur = NULL;
-
-  for (cur = &MyVars; *cur; cur = &((*cur)->next))
-    if (mutt_str_strcmp((*cur)->name, var) == 0)
-      break;
+  struct MyVar *myv = NULL;
 
-  if (!*cur)
-    *cur = mutt_mem_calloc(1, sizeof(struct MyVar));
+  mutt_debug(2, " * myvar_set(%s, '<hidden>')\n", NONULL((char *) var));
```
gahr: I’m not sure we want this debug message

init.c
```diff
@@ -4308,11 +4315,16 @@ static int parse_unsubscribe_from(struct Buffer *b, struct Buffer *s,
...
+  struct MyVar *myv = NULL;
 
-  for (cur = MyVars; cur; cur = cur->next)
-    if (mutt_str_strcmp(cur->name, var) == 0)
-      return NONULL(cur->value);
+  STAILQ_FOREACH(myv, &MyVars, entries)
+  {
+    if (mutt_str_strcmp(myv->name, var) == 0)
+    {
+      mutt_debug(2, " * myvar_get(%s)\n", NONULL((char *) var));
```
gahr: Same as above: do we want this debug message?

z2v:
@gahr @flatcap what is the preferred way to make changes in PR? Should I add new commits with changes or squash them in initial commit?

flatcap:
Adding more commits is probably simpler, for now.
That way GitHub will let us see the individual commits, or the combined view.

Ultimately though, it's your branch, so however you want to work is fine.

z2v:
Ok, while reviewing I'll be adding commits. Before merge I'll squash them in single commit. Anyway, replacing STAILQ with TAILQ should be reflected in commit message.

init.c
```diff
@@ -99,44 +99,46 @@ struct MyVar
...
-  struct MyVar **cur = NULL;
-
-  for (cur = &MyVars; *cur; cur = &((*cur)->next))
-    if (mutt_str_strcmp((*cur)->name, var) == 0)
-      break;
-
-  if (!*cur)
-    *cur = mutt_mem_calloc(1, sizeof(struct MyVar));
+  struct MyVar *myv = NULL;
+  struct MyVar *new = NULL;
```
gahr: How about doing without the second variable and just reusing `myv` later one?

z2v: This was the previous version of this code. I've introduced another variable just for clarity. No other reasons.

z2v: Removed.

z2v:
I've squashed commits and fixed commit message.

init.c
```diff
@@ -99,44 +99,45 @@ struct MyVar
 {
   char *name;
   char *value;
-  struct MyVar *next;
+  TAILQ_ENTRY(MyVar) entries;
 };
 
-static struct MyVar *MyVars;
+TAILQ_HEAD(MyVarList, MyVar);
+static struct MyVarList MyVars = TAILQ_HEAD_INITIALIZER(MyVars);
```
gahr: Oh, btw, it's also possible to have the head as an anonymous struct..
```c
static TAILQ_HEAD(, MyVar) MyVars = TAILQ_HEAD_INITIALIZER(MyVars);
```

z2v: Done.

gahr: Thanks!

Issue #1133, opened 2018-04-07
Logging: Review all logging messages and their log level


reisradomil:
(This is more a code clean-up thing than really an issue)

I noticed that `mutt_warrning()` isn't used by any code. On the other hand some log messages are flagged as errors (e.g. `key not bound` or `Top of message is shown`) which -- I would say -- are more warnings (or even informational messages) than errors.

It would be an improvement to make this consistent and develop (and document!) some guidelines what type of messages are debug5/.../debug1/messages/warning/errors.

(Note: there is also a perror level which I think are errors from C library functions)

flatcap:
> make this consistent ... (and document!) some guidelines
  
That is a **BIG** job, but one that we can do piecemeal.

Here are the current uses:
- 201 `mutt_perror()`
- 565 `mutt_error()`
- 216 `mutt_message()`
-   0 `mutt_warning()`
- 322 `mutt_debug(1)`
- 243 `mutt_debug(2)`
-  73 `mutt_debug(3)`
-  15 `mutt_debug(4)`
-  49 `mutt_debug(5)`

> I noticed that mutt_warning() isn't used by any code

Yes, that's a new level that I introduced when I updated the logging system.

> there is also a perror level which I think are errors from C library functions

Yes, many of the standard library functions return `-1` on failure and set a global variable `errno`.
`perror(3)` turns that into a localised string.

We need a separate level for this in the code, but the user shouldn't be aware of it.

> guidelines what type of messages are debug5/.../debug1/messages/warning/errors.

Hmm... tricky.

There are a couple of things we can do to improve the logging.
Filter out:
- Progress bar messages
  Having a separate level for these, means they could be displayed, but not saved
- UI messages, such as 'You're on the last message'
  These 'errors' probably need downgrading, too

What are your thoughts?

flatcap:
@reisradomil 
Are you ready for another coding challenge?

reisradomil:
> > make this consistent ... (and document!) some guidelines
>
> That is a **BIG** job, but one that we can do piecemeal.

I'm aware of that. This issue was more intended to point this out and to have a place to discuss it and to keep track of this issue.

> > I noticed that mutt_warning() isn't used by any code
>
> Yes, that's a new level that I introduced when I updated the logging system.

I thought something like that.

> >  guidelines what type of messages are debug5/.../debug1/messages/warning/errors.
>
> Hmm... tricky.

[...]

PR #1159, opened 2018-04-22
mutt/logging.c: set line buffering for logging stream

z2v:
Flush stream to the logging file after adding a line.

Without flushing logging messages are saved to file with delay due to stream buffering.

### DIFF
--- a/mutt/logging.c
+++ b/mutt/logging.c
@@ -128,6 +128,7 @@ int log_file_open(bool verbose)
   LogFileFP = mutt_file_fopen(LogFileName, "a+");
   if (!LogFileFP)
     return -1;
+  setvbuf(LogFileFP, NULL, _IOLBF, 0);
 
   fprintf(LogFileFP, "[%s] NeoMutt%s debugging at level %d\n", timestamp(0),
           NONULL(LogFileVersion), LogFileLevel);


flatcap:
Hmm... I'm not sure about this.
The buffering improves performance (but I haven't measured it).
Perhaps we're not writing enough for this to be a problem.

I have a few other ideas:
- Flush the lines after an error message
- Flush at most once per second
- Flush after 'n' messages

If NeoMutt were to crash, we wouldn't lose too much logging info.

gahr:
I could have sworn that the output was flushed anyway because of the fact that log messages end with a newline.

z2v:
@gahr Possibly this is OS-specific. On FreeBSD-CURRENT I definitely get partial log messages without flushing.

Run `neomutt -d 2` without flushing:
```
$ tail -n 3 .neomuttdebug0
[2018-04-23 13:52:26]<2> source_rc() Reading configuration file '/home/vovka/.mutt/hooks/accounthooks/gmail-account-hook'.
[2018-04-23 13:52:26]<2> source_rc() Reading configuration file '/home/vovka/.mutt/hooks/accounthooks/default-account-hook'.
[2018-04-23 13:52:26]<2> source_rc() Reading configuration file '/home/vovka/.mutt/hooks/accounthooks/default-acco$
```

With flushing:
```
$ tail -n 3 .neomuttdebug0
[2018-04-23 13:53:09]<2> source_rc() Reading configuration file '/home/vovka/.mutt/hooks/accounthooks/default-account-hook'.
[2018-04-23 13:53:09]<2> source_rc() Reading configuration file '/home/vovka/.mutt/hooks/accounthooks/gmail-account-hook'.
[2018-04-23 13:53:09]<M> mutt_version() NeoMutt 20180323-114-bc8636
$
```

@flatcap Buffering makes 'live' log monitoring (with `tail -f` for example) much less usable. Batching and timer raise question how to select good default values. What about additional config option, for example `debug_buffered` with default value equals to `true` for backward compatibility or enviroment variable like Python does?

flatcap:
I too have just done a bit of reading and experimenting :-)

My docs for `setvbuf()` say:
- Normally all files are block buffered.
- ... a terminal (as stdout normally does), it is line buffered
- standard error stream stderr is always unbuffered by default

> 'live' log monitoring

Yes, that's a reasonable case for flushing/not-buffered

As it's a text file that isn't going to the _that_ big, we don't need to buffer.

There are some other changes planned to reduced the amount of junk getting logged (#1133)

@z2v Please change the `fflush` to `setvbuf(fp, NULL, 0, 0);` in `log_file_open()`

gahr:
> @z2v Please change the fflush to setvbuf(fp, NULL, 0, 0); in log_file_open()

Did you mean `setvbuf(fp, NULL, _IOLBF, 0)` ?

flatcap:
Hmm...  The only place that would make a difference is some `mutt_debug()`s which log imcomplete lines.
I don't think there are many, but you're right `_IOLBF` would be better.

@z2v Please change the `fflush()` to `setvbuf(fp, NULL, _IOLBF, 0);` in `log_file_open()`

z2v:
Done.

PR #1163, opened 2018-04-24
Respect environment VISUAL and EDITOR variables

### LINKED ISSUE #1162: Neomutt uses Vi instead of EDITOR (Vim/Neovim)
## bug reports

`git bisect` revealed 0ef255ecd838c52dbbe0b35e954a932e017fc9fa as the offending commit

* Expected behavior
When composing/replying/forwarding email, Neomutt should open preferred editor.

* Actual behavior
Neomutt opens `vi` instead of preferred editor (`vi`/`nvim`)

* Steps to reproduce
  - Navigate to email composition/reply/forward window
  - Editor should be Vi. Use `:ve` to verify.

* Used program versions
neomutt-git

Arch 4.15.14-1

Austin-Ray:
#### **What does this PR do?**

Commit 0ef255e repositioned where Neomutt set its standard defaults in
init.c to after the logic for loading the VISUAL and EDITOR variables.
As a result, Neomutt overrides the values it retrieved from the
environment with the ones from the configuration file. This matches the
desired precedence of config->environment->code.

However, if `editor` is not set in the configuration file then Neomutt
does not retrieve EDITOR and VISUAL again. This results in Neomutt
falling back to the editor specified in the code. Thus the precedence is
config->code.

This commit returns the standard default set loop to its initial
location allowing for the environment variables to be usable.

This fixes #1162 

#### Are there points in the code the reviewer needs to double check?**
`MailcapPath` and `Tmpdir` were affected by commit 0ef255e as well; however, I did not know it when I made my commit since they did not show any errors (they were set in the config file.) Please verify they conform to the desired precedence.

#### What are the relevant issue numbers?**
#1162

### DIFF
--- a/init.c
+++ b/init.c
@@ -3905,6 +3905,13 @@ int mutt_init(int skip_sys_rc, struct ListHead *commands)
 
   Matches = mutt_mem_calloc(MatchesListsize, sizeof(char *));
 
+  /* Set standard defaults */
+  for (int i = 0; MuttVars[i].name; i++)
+  {
+    set_default(&MuttVars[i]);
+    restore_default(&MuttVars[i]);
+  }
+
   /* "$mailcap_path" precedence: config file, environment, code */
   MailcapPath = mutt_str_strdup(mutt_str_getenv("MAILCAPS"));
 
@@ -3919,13 +3926,6 @@ int mutt_init(int skip_sys_rc, struct ListHead *commands)
   {
     mutt_str_replace(&Editor, env_ed);
     mutt_str_replace(&Visual, env_ed);
-  }
-
-  /* Set standard defaults */
-  for (int i = 0; MuttVars[i].name; i++)
-  {
-    set_default(&MuttVars[i]);
-    restore_default(&MuttVars[i]);
   }
 
   CurrentMenu = MENU_MAIN;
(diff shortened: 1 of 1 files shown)


flatcap:
Thanks for the PR.
If I stare at this code too long my brain melts.

Austin-Ray:
Using an empty `.neomuttrc` is the easiest way to demonstrate that Neomutt ignores the environment variables. 

On a partially related note, if the variable `Tmpdir` is neither set in the configuration file nor as an environmental variable, Neomutt tries to use `/` as the temp directory.

gahr:
>  Neomutt tries to use / as the temp directory.

How comes? The default is hardcoded as "/tmp"

Austin-Ray:
@gahr I initially did not mean with my patch, but you found the source of the error before I could. 

So it seems that moving the set standard defaults loop to its initial location prioritizes environment variables over the hard-coded defaults, even if the variables are empty.

However, in the case of `VISUAL` and `EDITOR`, it is desirable for the default, `vi` in this case, to be overridden by the environment variable.

I believe the removed if statements in commit 0ef255e prevented empty environment variables from overriding the defaults.

Austin-Ray:
My latest commit restores the empty env var checks so that the precedence becomes:
config->env var (if not empty)->defaults

This prevents an `unset TMPDIR` from setting `/` as the temporary directory

PR #1170, opened 2018-04-27
Improve the usage() string

flatcap:
- ef8f5f92c Rewrite usage help string
- e448fe6a5 l10n comments
- 8b552589a re-wording
- b08c7f13d drop dots

I've done a little re-wording, I hope it's OK.

Also, I've dropped the dots.
Yes, most of the strings are sentences, but to me it looks a bit odd.
(I can kill this commit if you'd like to keep them).


mschilli87:
re-wording LGTM

floand:
Hi @flatcap,

thanks for looking at it – I like it. Two things, after reading it so many times in the
past few days. What do you think about putting `--` description at first above `-A`?
It would match my first intention to sort commands, functions, variables etc. in
our docs. Also, `--` is only mentioned in `-a` right below then.

The second thing, I would like to adopt the final changes in `neomutt(1)` and
the guide, before you commit this. So all is in sync. :-)

PR #1171, opened 2018-04-28
Eliminating static variable in mutt_file_dirname

yuzibo:
* **What does this PR do?**
Eliminate static variable in mutt_file_dirname
Yes

     for that)

   - All builds are passing

Malloc a char *buf and free it in mutt_file_to_absolate_path() and mutt_save_confirm()

### DIFF
--- a/mutt/file.c
+++ b/mutt/file.c
@@ -927,12 +927,10 @@ time_t mutt_file_decrease_mtime(const char *f, struct stat *st)
  * Unlike the IEEE Std 1003.1-2001 specification of dirname(3), this
  * implementation does not modify its parameter, so callers need not manually
  * copy their paths into a modifiable buffer prior to calling this function.
- *
- * @warning mutt_file_dirname() returns a static string which must not be free()'d.
  */
-const char *mutt_file_dirname(const char *p)
+char *mutt_file_dirname(const char *p)
 {
-  static char buf[_POSIX_PATH_MAX];
+  char *buf = mutt_mem_malloc(_POSIX_PATH_MAX);
   mutt_str_strfcpy(buf, p, sizeof(buf));
   return dirname(buf);
 }
@@ -1312,7 +1310,6 @@ int mutt_file_rename(char *oldfile, char *newfile)
  */
 int mutt_file_to_absolute_path(char *path, const char *reference)
 {
-  const char *dirpath = NULL;
   char abs_path[PATH_MAX];
   int path_len;
 
@@ -1322,8 +1319,9 @@ int mutt_file_to_absolute_path(char *path, const char *reference)
     return true;
   }
 
-  dirpath = mutt_file_dirname(reference);
+  char *dirpath = mutt_file_dirname(reference);
   mutt_str_strfcpy(abs_path, dirpath, PATH_MAX);
+  FREE(&dirpath);
   mutt_str_strncat(abs_path, sizeof(abs_path), "/", 1); /* append a / at the end of the path */
 
   path_len = PATH_MAX - strlen(path);
--- a/mutt/file.c
+++ b/mutt/file.c
@@ -921,18 +921,18 @@ time_t mutt_file_decrease_mtime(const char *f, struct stat *st)
 
 /**
  * mutt_file_dirname - Return a path up to, but not including, the final '/'
- * @param  p    Path
+ * @param  path Path
  * @retval ptr  The directory containing p
  *
  * Unlike the IEEE Std 1003.1-2001 specification of dirname(3), this
  * implementation does not modify its parameter, so callers need not manually
  * copy their paths into a modifiable buffer prior to calling this function.
  */
-char *mutt_file_dirname(const char *p)
+char *mutt_file_dirname(const char *path)
 {
-  char *buf = mutt_mem_malloc(_POSIX_PATH_MAX);
-  mutt_str_strfcpy(buf, p, sizeof(buf));
-  return dirname(buf);
+  char buf[PATH_MAX] = { 0 };
+  mutt_str_strfcpy(buf, path, sizeof(buf));
+  return mutt_str_strdup(dirname(buf));
 }
 
 /**
@@ -1388,3 +1388,4 @@ int mutt_file_check_empty(const char *path)
 
   return ((st.st_size == 0));
 }
+
--- a/mutt/file.h
+++ b/mutt/file.h
@@ -46,7 +46,7 @@ char *      mutt_file_concat_path(char *d, const char *dir, const char *fname, s
 int         mutt_file_copy_bytes(FILE *in, FILE *out, size_t size);
 int         mutt_file_copy_stream(FILE *fin, FILE *fout);
 time_t      mutt_file_decrease_mtime(const char *f, struct stat *st);
-char *      mutt_file_dirname(const char *p);
+char *      mutt_file_dirname(const char *path);
 int         mutt_file_fclose(FILE **f);
 FILE *      mutt_file_fopen(const char *path, const char *mode);
 int         mutt_file_fsync_close(FILE **f);
--- a/mutt/file.h
+++ b/mutt/file.h
@@ -46,7 +46,7 @@ char *      mutt_file_concat_path(char *d, const char *dir, const char *fname, s
 int         mutt_file_copy_bytes(FILE *in, FILE *out, size_t size);
 int         mutt_file_copy_stream(FILE *fin, FILE *fout);
 time_t      mutt_file_decrease_mtime(const char *f, struct stat *st);
-const char *mutt_file_dirname(const char *p);
+char *      mutt_file_dirname(const char *p);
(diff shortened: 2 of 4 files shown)


mutt/file.c
```diff
@@ -930,11 +930,16 @@ time_t mutt_file_decrease_mtime(const char *f, struct stat *st)
  *
  * @warning mutt_file_dirname() returns a static string which must not be free()'d.
  */
-const char *mutt_file_dirname(const char *p)
+char *mutt_file_dirname(const char *p)
 {
-  static char buf[_POSIX_PATH_MAX];
-  mutt_str_strfcpy(buf, p, sizeof(buf));
-  return dirname(buf);
+  char *buf = malloc(sizeof(_POSIX_PATH_MAX));
```
flatcap: Hmm... that doesn't allocate the amount you think it does!

Also, please use `mutt_mem_malloc()`.
It cannot fail, so you don't need to test the return value.

mutt/file.c
```diff
@@ -1324,6 +1329,8 @@ int mutt_file_to_absolute_path(char *path, const char *reference)
 
   dirpath = mutt_file_dirname(reference);
   mutt_str_strfcpy(abs_path, dirpath, PATH_MAX);
+  /*mutt_file_dirname malloc char *dirpath, so here free it*/
+  free(dirpath);
```
flatcap: Please use our macro `FREE()` which takes a double pointer.
It sets the variable to NULL, too.

mutt/file.h
```diff
@@ -46,7 +46,7 @@ char *      mutt_file_concat_path(char *d, const char *dir, const char *fname, s
 int         mutt_file_copy_bytes(FILE *in, FILE *out, size_t size);
 int         mutt_file_copy_stream(FILE *fin, FILE *fout);
 time_t      mutt_file_decrease_mtime(const char *f, struct stat *st);
-const char *mutt_file_dirname(const char *p);
+char *mutt_file_dirname(const char *p);
```
flatcap: Would you mind aligning this function with the others :-)

muttlib.c
```diff
@@ -1313,12 +1314,14 @@ int mutt_save_confirm(const char *s, struct stat *st)
...
-        ret = -1;
+	/* Maybe waste some spaces here */
+        snprintf(tmp, sizeof(tmp), _("Append messages to %s?"), s);
+        rc = mutt_yesorno(tmp, MUTT_YES);
+        if (rc == MUTT_NO)
+          ret = 1;
+        else if (rc == MUTT_ABORT)
+          ret = -1;
+
     }
```
flatcap: Please can you replace all the `<tab>`s with spaces.
You can run `clang-format -i FILE` to fix most formatting issues.

muttlib.c
```diff
@@ -1376,6 +1387,7 @@ int mutt_save_confirm(const char *s, struct stat *st)
   }
 
   mutt_window_clearline(MuttMessageWindow, 0);
+  free(tmp_path);
```
flatcap: `FREE()`

muttlib.c
```diff
@@ -1347,26 +1350,34 @@ int mutt_save_confirm(const char *s, struct stat *st)
...
+            ret = -1;
+        }
 
       /* user confirmed with MUTT_YES or set Confirmcreate */
       if (ret == 0)
       {
         /* create dir recursively */
-        if (mutt_file_mkdir(mutt_file_dirname(s), S_IRWXU) == -1)
-        {
+	if(tmp_path != NULL)
```
flatcap: For pointers and booleans, it's OK to just: `if (variable)` or `if (!variable)`

mutt/file.c
```diff
@@ -930,11 +930,11 @@ time_t mutt_file_decrease_mtime(const char *f, struct stat *st)
  *
  * @warning mutt_file_dirname() returns a static string which must not be free()'d.
  */
-const char *mutt_file_dirname(const char *p)
+char *mutt_file_dirname(const char *p)
 {
-  static char buf[_POSIX_PATH_MAX];
+  char *buf = mutt_mem_malloc(sizeof(_POSIX_PATH_MAX));
```
flatcap: `sizeof(_POSIX_PATH_MAX)` == 4
You just want `_POSIX_PATH_MAX`

yuzibo: Thank you again!

mutt/file.c
```diff
@@ -1324,6 +1323,8 @@ int mutt_file_to_absolute_path(char *path, const char *reference)
 
   dirpath = mutt_file_dirname(reference);
   mutt_str_strfcpy(abs_path, dirpath, PATH_MAX);
+  /*mutt_file_dirname malloc char *dirpath, so here free it*/
+  FREE(dirpath);
```
gahr: `FREE(&dirpath)`

flatcap:
I've undone all the unnecessary formatting.
clang-format shouldn't have made such a mess, it should have picked up the config file in the source directory.

Anyway, now we can see your changes again.

flatcap:
OK, so the easy bit is done. The code is changed and it compiles.

Unfortunately, even without running it, I can see some problems:
1. The function doesn't work any more
   You should be able to see this if you look carefully at the code before and after your changes.
2. Both callers are likely to crash when used
   gahr and I have both commented about the cause of this.
3. One of the callers leaks memory
   When a function allocates memory, it's important that it's freed in **every** path through the code.

I'd like you to find (and fix) these problems.

flatcap:
OK, we're getting close.
Both callers still have paths that leak, though.

https://github.com/neomutt/neomutt/blob/fcd92d5413061995338400accaa5aadfeaa8535e/mutt/file.c#L1311-L1342

First let's combine lines 1313 and 1323.
Define the variable when we first use it.

`dirpath` gets created on line 1323, used on 1324 and doesn't get used again.

After it's been created, then function can end at line 1336, or line 1341.
They both need to free the memory.

**OR**

We move the `FREE()` to **immediately** after we use it.

The same is happening in `mutt_save_confirm()`, so that needs fixing too.

muttlib.c
```diff
@@ -1295,6 +1295,7 @@ FILE *mutt_open_read(const char *path, pid_t *thepid)
 int mutt_save_confirm(const char *s, struct stat *st)
 {
   char tmp[_POSIX_PATH_MAX];
+  char *tmp_path = mutt_mem_malloc(_POSIX_PATH_MAX);
```
flatcap: This alloc isn't used.
Move the variable definition to where it's used.

muttlib.c
```diff
@@ -1360,12 +1361,15 @@ int mutt_save_confirm(const char *s, struct stat *st)
       if (ret == 0)
       {
         /* create dir recursively */
-        if (mutt_file_mkdir(mutt_file_dirname(s), S_IRWXU) == -1)
+        tmp_path = mutt_file_dirname(s);
+        if (mutt_file_mkdir(tmp_path, S_IRWXU) == -1)
         {
           /* report failure & abort */
           mutt_perror(s);
+	  FREE(&tmp_path);
```
flatcap: Can you check your editor.
There are \<tab\>s here.

muttlib.c
```diff
@@ -1360,12 +1361,15 @@ int mutt_save_confirm(const char *s, struct stat *st)
...
           return 1;
         }
+	FREE(&tmp_path);
```
flatcap: There are \<tab\>s here.

mutt/file.c
```diff
@@ -1335,9 +1333,12 @@ int mutt_file_to_absolute_path(char *path, const char *reference)
...
   {
     printf("Error: issue converting path to absolute (%s)", strerror(errno));
+    /* Even fail to get absolute path, FREE() dirpath also */
+    FREE(&dirpath);
     return false;
   }
-
+  /*Beforing return value successfully, first FREE() dirpath */
+  FREE(&dirpath);
   return true;
```
flatcap: These two `FREE()`s stop the leaks, but it would be better if they were closer to the creation.
Make the scope of the variable (and memory allocation) as small as possible.

```c
char *var = alloc_function();
work_function(var);
FREE(&var);
```

flatcap:
muttlib.c still contains tabs.
the FREE()s in mutt_file_to_absolute_path() still need moving to immediately after we've finished with dirpath.

flatcap:
YEAH! Now the code looks good.

Time to get things in a state that's ready for commit.
(make a backup before, just in case :-)

- Squash the changes into one commit, e.g. `git reset master`
- Create a commit message
  - Short subject (max 50 chars)
  - Description of changes (very little is needed for this change)
- Rebase the changes onto the latest master
  - Update master: `git fetch origin master:master`
  - Rebase: `git rebase master`
- Check your changes one last time
- Push to github: `git push --force origin devel/eliminate-static-variable`
- Ask for a review from "neomutt/reviewers" (button top right)

yuzibo:
I hope i am not mess up the commits and code.
But i am worried about it a little.
Thanks.

PR #1174, opened 2018-04-28
Convert Aliases to use TAILQ

### LINKED ISSUE #1153: Convert Aliases to use STAILQ
This is an "easy" task ([find more](https://github.com/neomutt/neomutt/issues?q=label:"good%20first%20issue")) and it's open to everyone.  [_read more..._](https://neomutt.org/dev/easy-tasks)

## Description

Many NeoMutt structures implement their own linked-list type.  
The Aliases are simply stored as a long linked list.

~`STAILQ` is a singly-linked~
`TAILQ` is a doubly-linked list with pointers to both the head and the tail.  
It's defined in `mutt/queue.h`

### Part 2: Change all users to match

There are plenty of examples in NeoMutt's code of how to iterate through a list.  
~**Note**: When deleting items, use `STAILQ_FOREACH_SAFE()`~

@flatcap: updated:STAILQ -> TAILQ

z2v:
* **What does this PR do?**
Convert Aliases to use TAILQ

Fixes #1153.

### DIFF
--- a/addrbook.c
+++ b/addrbook.c
@@ -201,20 +201,19 @@ static int alias_sort_address(const void *a, const void *b)
  * @param buflen Length of buffer
  * @param aliases Alias List
  */
-void mutt_alias_menu(char *buf, size_t buflen, struct Alias *aliases)
+void mutt_alias_menu(char *buf, size_t buflen, struct AliasList *aliases)
 {
-  struct Alias *aliasp = NULL;
+  struct Alias *a = NULL, *last = NULL;
   struct Menu *menu = NULL;
   struct Alias **AliasTable = NULL;
   int t = -1;
   int i;
   bool done = false;
-  int op;
   char helpstr[LONG_STRING];
 
   int omax;
 
-  if (!aliases)
+  if (TAILQ_EMPTY(aliases))
   {
     mutt_error(_("You have no aliases!"));
     return;
@@ -228,14 +227,13 @@ void mutt_alias_menu(char *buf, size_t buflen, struct Alias *aliases)
   mutt_menu_push_current(menu);
 
 new_aliases:
-
   omax = menu->max;
 
   /* count the number of aliases */
-  for (aliasp = aliases; aliasp; aliasp = aliasp->next)
+  TAILQ_FOREACH_FROM(a, aliases, entries)
   {
-    aliasp->del = false;
-    aliasp->tagged = false;
+    a->del = false;
+    a->tagged = false;
     menu->max++;
   }
 
@@ -244,30 +242,36 @@ new_aliases:
   if (!AliasTable)
     return;
 
-  for (i = omax, aliasp = aliases; aliasp; aliasp = aliasp->next, i++)
+  if (last)
+    a = TAILQ_NEXT(last, entries);
+
+  i = omax;
+  TAILQ_FOREACH_FROM(a, aliases, entries)
   {
-    AliasTable[i] = aliasp;
-    aliases = aliasp;
+    AliasTable[i] = a;
+    i++;
   }
 
   if ((SortAlias & SORT_MASK) != SORT_ORDER)
   {
-    qsort(AliasTable, i, sizeof(struct Alias *),
+    qsort(AliasTable, menu->max, sizeof(struct Alias *),
           (SortAlias & SORT_MASK) == SORT_ADDRESS ? alias_sort_address : alias_sort_alias);
   }
 
   for (i = 0; i < menu->max; i++)
     AliasTable[i]->num = i;
 
+  last = TAILQ_LAST(aliases, AliasList);
+
   while (!done)
   {
-    if (aliases->next)
+    int op;
+    if (TAILQ_NEXT(last, entries))
     {
       menu->redraw |= REDRAW_FULL;
-      aliases = aliases->next;
+      a = TAILQ_NEXT(last, entries);
       goto new_aliases;
     }
-
     switch ((op = mutt_menu_loop(menu)))
     {
       case OP_DELETE:
--- a/alias.c
+++ b/alias.c
@@ -40,11 +40,13 @@
 
 struct Address *mutt_alias_lookup(const char *s)
 {
-  struct Alias *t = Aliases;
+  struct Alias *a = NULL;
 
-  for (; t; t = t->next)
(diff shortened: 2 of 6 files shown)


globals.h
```diff
@@ -95,7 +95,7 @@ WHERE SIG_ATOMIC_VOLATILE_T SigWinch;
 
 WHERE int CurrentMenu;
 
-WHERE struct Alias *Aliases;
+WHERE TAILQ_HEAD(AliasList, Alias) Aliases INITVAL(TAILQ_HEAD_INITIALIZER(Aliases));
```
flatcap: Can you split this up, please.
`mutt/list.h` defines both the ListNode (Alias) and ListHead (AliasList)

alias.c
```diff
@@ -681,17 +648,21 @@ bool mutt_addr_is_user(struct Address *addr)
...
+    return;
+
+  mutt_alias_delete_reverse(p);
+  FREE(&p->name);
+  mutt_addr_free(&p->addr);
+  FREE(&p);
+}
 
-  while (*p)
+void mutt_alias_free_all() {
```
flatcap: I'd prefer if this took the AliasList to be freed.
There's only one, atm, but that may change.

z2v:
@flatcap done. Also `mutt_alias_free_all` renamed to `mutt_aliaslist_free` to conform with `mutt_regexlist_free`.

alias.c
```diff
@@ -681,17 +648,21 @@ bool mutt_addr_is_user(struct Address *addr)
...
+  struct Alias *a = NULL, *tmp = NULL;
+  TAILQ_FOREACH_SAFE(a, a_list, entries, tmp)
   {
-    t = *p;
-    *p = (*p)->next;
-    mutt_alias_delete_reverse(t);
-    FREE(&t->name);
-    mutt_addr_free(&t->addr);
-    FREE(&t);
+    mutt_alias_free(a);
```
flatcap: I think we need a `REMOVE` here, otherwise we'll end up with a list of dead pointers.

init.c
```diff
@@ -1776,36 +1776,31 @@ static int parse_unalias(struct Buffer *buf, struct Buffer *s,
...
-            break;
           }
-
-          if (last)
-            last->next = tmp->next;
           else
-            Aliases = tmp->next;
-          tmp->next = NULL;
-          mutt_alias_free(&tmp);
+            mutt_alias_free(a);
```
flatcap: I think we need a `REMOVE` here

addrbook.c
```diff
@@ -227,15 +224,11 @@ void mutt_alias_menu(char *buf, size_t buflen, struct Alias *aliases)
   menu->help = mutt_compile_help(helpstr, sizeof(helpstr), MENU_ALIAS, AliasHelp);
   mutt_menu_push_current(menu);
 
-new_aliases:
```
flatcap: We need this loop.
If you go to the alias selection menu, then add a new alias `:alias ...` it should show up in the menu.

alias.c
z2v: Fixed.

init.c
z2v: Fixed. Also eliminated `tmp` variable, because we don't need `_SAFE` macro here (looping stops after element removal).

addrbook.c
z2v: Dynamic adding new alias works only when *all* aliases are shown in the menu. But if the menu includes aliases selected by the completion than new alias will not be shown. It seems, this is the wrong place for this loop.

I'll restore functionality as it is in current `master`.

z2v:
@flatcap I've managed to restore alias adding in alias menu without traversing full list. Also squashed all commits and rebased on actual `master`.

alias.c
```diff
@@ -681,17 +648,23 @@ bool mutt_addr_is_user(struct Address *addr)
   return false;
 }
 
-void mutt_alias_free(struct Alias **p)
+void mutt_alias_free(struct Alias *p)
```
flatcap: Please can you change this back to `void mutt_alias_free(struct Alias **p)`
(to keep it the same as all our other 'free' functions -- free and NULL)

alias.c
```diff
@@ -509,17 +503,17 @@ void mutt_alias_delete_reverse(struct Alias *t)
...
 
   if (s[0] != 0) /* avoid empty string as strstr argument */
   {
     memset(bestname, 0, sizeof(bestname));
 
-    while (a)
+    TAILQ_FOREACH(a, &Aliases, entries)
     {
-      if (a->name && strstr(a->name, s) == a->name)
+      if (a->name && a->name[0] == s[0] && strstr(a->name, s) == a->name)
```
flatcap: If `(strstr(a->name, s) == a->name)`, then we know the two strings match at the beginning,
so I'm not sure I see the need for the check `a->name[0] == s[0]`

alias.c
```diff
@@ -545,63 +538,37 @@ int mutt_alias_complete(char *s, size_t buflen)
       }
 
       /* build alias list and show it */
-
-      a = Aliases;
-      while (a)
+      TAILQ_FOREACH(a, &Aliases, entries)
       {
-        if (a->name && (strstr(a->name, s) == a->name))
+        if (a->name && a->name[0] == s[0] && strstr(a->name, s) == a->name)
```
flatcap: `a->name[0] == s[0]` here, too

alias.c
z2v: It's a micro-optimization. `strstr(str, sub)` scans full string `str` to locate `sub`. This avoids scanning at all if first characters do not match.

gahr: How about  `strncmp(a->name, s, strlen(s)) == 0` ?

flatcap: hmm, the `strstr()` implies that `mutt_alias_complete()` might match on any partial string,
but I can only get it to complete if the beginning matches.

flatcap: as gahr points out, `strstr(a->name, s) == a->name` always anchors the expression to the beginning, so `strncmp()` is much better.

alias.c
z2v: @flatcap done.

alias.c
z2v: @flatcap, @gahr done.

alias.c
z2v: Replaced with `strncmp(..) == 0` here too.

PR #1183, opened 2018-05-01
Translate

yuzibo:
* **What does this PR do?**
Update Chinese(Simplified) translate

     for that)

   - All builds are passing

In my commit, i have expressed my opinion about Chinese translate.If it is ok, i will uodate it next time
Thanks,


flatcap:
I see the problem.  You don't need the `msgstr[1]` at all.

In the po file header, it defines how many plural forms a language has and when to use them.

- German: 2 forms, use the second when the number is not equal to 1
  "Plural-Forms: nplurals=2; plural=(n != 1);\n"

- Chinese: 1 form, always
  "Plural-Forms: nplurals=1; plural=0;\n"

- Polish: 3 forms, complicated!
  "Plural-Forms: nplurals=3; plural=(n==1 ? 0 : n%10>=2 && n%10<=4 && (n%100<10 || n%100>=20) ? 1 : 2);\n"

I've fixed the problem.
I'll merge-in the translate branch closer to the release.

PR #1184, opened 2018-05-02
Fix static addr mbox to udomain

### LINKED ISSUE #1150: Eliminate static variable in mutt_addr_mbox_to_udomain()
This is an "easy" task ([find more](https://github.com/neomutt/neomutt/issues?q=label:"good%20first%20issue")) and it's open to everyone.  [_read more..._](https://neomutt.org/dev/easy-tasks)

## Description

`mutt_addr_mbox_to_udomain()` returns pointers into a static buffer.

### Part 2: Change the callers to `free()` the strings

There are three callers, all in `mutt/address.c` which will need changing.

psyomn:
Remove static function variable from function.

Fixes #1150

### DIFF
--- a/test/address.c
+++ b/test/address.c
@@ -1,0 +1,94 @@
+#define TEST_NO_MAIN
+#include "acutest.h"
+#include "acutest.h"
+#include "email/address.h"
+#include "mutt/memory.h"
+#include <string.h>
+
+#define TEST_CHECK_STR_EQ(expected, actual)                                    \
+  do                                                                           \
+  {                                                                            \
+    if (!TEST_CHECK(strcmp(expected, actual) == 0))                            \
+    {                                                                          \
+      TEST_MSG("Expected: %s", expected);                                      \
+      TEST_MSG("Actual  : %s", actual);                                        \
+    }                                                                          \
+  } while (false)
+
+void test_addr_mbox_to_udomain(void)
+{
+  { /* edge cases */
+    char *user = NULL;
+    char *domain = NULL;
+
+    if (!TEST_CHECK(mutt_addr_mbox_to_udomain("bobnodomain@", &user, &domain) == -1) ||
+        !TEST_CHECK(mutt_addr_mbox_to_udomain("bobnodomain", &user, &domain) == -1) ||
+        !TEST_CHECK(mutt_addr_mbox_to_udomain("@nobobohnoez", &user, &domain) == -1) ||
+        !TEST_CHECK(mutt_addr_mbox_to_udomain("", &user, &domain) == -1))
+    {
+      TEST_MSG("bad return");
+    }
+  }
+
+  { /* common */
+    int ret = 0;
+    const char *str = "bob@bobsdomain";
+
+    char *user = NULL;
+    char *domain = NULL;
+    ret = mutt_addr_mbox_to_udomain(str, &user, &domain);
+
+    if (!TEST_CHECK(ret == 0))
+    {
+      TEST_MSG("bad return");
+      TEST_MSG("Expected: %d", 0);
+      TEST_MSG("Actual  : %d", ret);
+    }
+
+    TEST_CHECK_STR_EQ("bob", user);
+    TEST_CHECK_STR_EQ("bobsdomain", domain);
+
+    FREE(&user);
+    FREE(&domain);
+  }
+
+  { /* integration */
+    char buf[256] = { 0 };
+    char *per = "bobby bob";
+    char *mbx = "bob@bobsdomain";
+
+    struct Address addr = {
+      .personal = per,
+      .mailbox = mbx,
+      .group = 0,
+      .next = NULL,
+      .is_intl = 0,
+      .intl_checked = 0,
+    };
+
+    mutt_addr_write_single(buf, sizeof(buf), &addr, false);
+
+    const char *expected = "bobby bob <bob@bobsdomain>";
+
+    TEST_CHECK_STR_EQ(expected, buf);
+  }
+
+  { /* integration */
+    char *per = "bobby bob";
+    char *mbx = "bob@bobsdomain";
+
+    struct Address addr = {
+      .personal = per,
+      .mailbox = mbx,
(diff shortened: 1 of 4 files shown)


psyomn:
I'm not sure how you folk feel about the tests I added, but I could go ahead and remove them if they're not unwanted. Read the docs, tried to stick to things as much as possible.

mutt/address.c
```diff
@@ -930,23 +930,33 @@ bool mutt_addr_is_local(struct Address *a)
...
 {
-  static char *buf = NULL;
+  char *buf = NULL;
   char *p = NULL;
 
   mutt_str_replace(&buf, mbox);
-  if (!buf)
-    return -1;
+  if (!buf) goto fail;
+  if (strchr(buf, '@') - buf == 0) goto fail_cleanup;
```
gahr: I would prefer a solution where `strchr` is called only once. Once you get the result, you should be able to discriminate the four important cases:
1. `@` is not part of the string -> fail
2. the string starts with `@` -> fail
3. the string ends with `@` -> fail
3. the string contains `@` -> ok, split into `user` and `domain` part

Also, you should be able to exploit `mutt_str_substr_dup` to avoid copying the original `mbox` string twice, and get rid of `buf` altogether.

mutt/address.c
```diff
@@ -998,11 +1008,14 @@ const char *mutt_addr_for_display(struct Address *a)
     return a->mailbox;
 
   local_mailbox = mutt_idna_intl_to_local(user, domain, MI_MAY_BE_IRREVERSIBLE);
+
   if (!local_mailbox)
     return a->mailbox;
```
gahr: You're leaking `user` and `domain` here. Please free them as soon as you're done using them.

psyomn:
Thanks for the feedback, I've applied the changes you requested. I'll try and go over this one more time tomorrow.

mutt/address.c
```diff
@@ -930,23 +930,22 @@ bool mutt_addr_is_local(struct Address *a)
...
+  char *ptr = strchr(mbox, '@');
+
+  if (ptr - mbox == 0) goto fail;
+  if (!ptr || !ptr[1]) goto fail;
 
-  mutt_str_replace(&buf, mbox);
-  if (!buf)
-    return -1;
+  *user = mutt_str_substr_dup(mbox, ptr);
+  *domain = mutt_str_substr_dup(ptr + 1, NULL);
```
gahr: `mutt_str_strdup(ptr + 1)` perhaps?

mutt/address.c
```diff
@@ -930,23 +930,22 @@ bool mutt_addr_is_local(struct Address *a)
...
+ * @warning you must free user and domain
  */
 int mutt_addr_mbox_to_udomain(const char *mbox, char **user, char **domain)
 {
-  static char *buf = NULL;
-  char *p = NULL;
+  char *ptr = strchr(mbox, '@');
+
+  if (ptr - mbox == 0) goto fail;
+  if (!ptr || !ptr[1]) goto fail;
```
gahr: Can we please make it a bit clearer what the fail cases are? How about:

```c
/* Fail if '@' is nowehere, at start, or at end */
if (ptr == NULL || ptr == mbox || ptr[1] == '\0')
{
  goto fail;
}
```

mutt/address.c
```diff
@@ -998,11 +997,17 @@ const char *mutt_addr_for_display(struct Address *a)
     return a->mailbox;
 
   local_mailbox = mutt_idna_intl_to_local(user, domain, MI_MAY_BE_IRREVERSIBLE);
-  if (!local_mailbox)
```
gahr: I would prefer `user` and `domain` to be freed as soon as they're not needed anymore, no matter whether `mutt_idna_intl_to_local` succeeded or not. What you do in `mutt_addrlist_to_intl` is better in my opinion.

test/address.c
```diff
@@ -0,0 +1,96 @@
...
+        TEST_CHECK(mutt_addr_mbox_to_udomain("@nobobohnoez", &user, &domain) == -1) ||
+        TEST_CHECK(mutt_addr_mbox_to_udomain("", &user, &domain) == -1)) {
+      TEST_MSG("bad return");
+    }
+  }
+
+  { /* common */
+    int ret = 0;
+    char *str = calloc(256, sizeof(*str));
+    strcat(str, "bob@bobsdomain");
```
gahr: How about `const char *str = "bob@bobsdomain"` ?

test/address.c
```diff
@@ -0,0 +1,96 @@
...
+    TEST_CHECK_STR_EQ("bobsdomain", domain);
+
+    FREE(&user);
+    FREE(&domain);
+    FREE(&str);
+  }
+
+  { /* integration */
+    size_t buflen = 256;
+    char *buf = calloc(buflen, sizeof(*buf));
```
gahr: How about `char buf[256]` ?

mutt/address.c
psyomn: Done!

psyomn:
Thanks for the feedback! I tried my best to apply the changes you suggested. Cheers!

flatcap:
(@gahr's on holiday)

The code looks good.  Please can you run `clang-format -i` on the source files you've changed.
It'll move some `{}`s, and reindent, to match the other NeoMutt code.

I see a problem with one of the test cases, though.
```c
if (TEST_CHECK(addr_mbox_to_udomain("bobnodomain@", &user, &domain) == -1) ||
    TEST_CHECK(addr_mbox_to_udomain("bobnodomain", &user, &domain) == -1) ||
    TEST_CHECK(addr_mbox_to_udomain("@nobobohnoez", &user, &domain) == -1) ||
    TEST_CHECK(addr_mbox_to_udomain("", &user, &domain) == -1))
```

Do you see it?
Try changing the second line to a **valid** address and run the test.

psyomn:
Hey sorry for not getting back to you sooner. It's been a little busy on my side. I'll try and contribute again to this repo; thank you for all the feedback you guys gave me along the way!

Issue #1177, opened 2018-04-30
Convert GroupContexts to use STAILQ


flatcap:
This is an "easy" task ([find more](https://github.com/neomutt/neomutt/issues?q=label:"good%20first%20issue"+is:open)) and it's open to everyone.  [_read more..._](https://neomutt.org/dev/easy-tasks)

## Description

Many NeoMutt structures implement their own linked-list type.  
The GroupContexts are simply stored as a long linked list.

`STAILQ` is a singly-linked list with pointers to both the head and the tail.  
It's defined in [mutt/queue.h](https://github.com/neomutt/neomutt/blob/master/mutt/queue.h),

### Part 2: Change all users to match

There are plenty of examples in NeoMutt's code of how to iterate through a list.  
For example: [tags.c](https://github.com/neomutt/neomutt/blob/master/tags.c), [tags.h](https://github.com/neomutt/neomutt/blob/master/tags.h)

**Note**: When deleting items, use `STAILQ_FOREACH_SAFE()`

yuzibo:
Hello,
Sorry, but i want to ask clearly question here.
`.struct GroupContext
{
    struct *g;
    STAILQ_ENTRY(GroupContext) entries;
};
STAILQ_HEAD(GroupHead, GroupNode);`
Could you have a glance at the the first step for the issue?
Is this right the first step to do that?
Thank you

flatcap:
Yes, that's right.

flatcap:
This issue is now finished successfully!  Thanks, @yuzibo.
PR #1187 will stay open whilst we do some other work to add tests for the GroupContext code.

PR #1187, opened 2018-05-04
Convert GroupContext to use STAILQ

yuzibo:
* **What does this PR do?**
Convert GroupContext to use STAILQ
YES. The code is very bad

     for that)

   - All builds are passing


yuzibo:
#1177

yuzibo:
Hello,   
I know the question may raise where:
I don't initializer GroupContext?
But which function can play the role in group.h?

gahr:
I think `GroupContextHead` should be passed around by pointer, not by pointer-to-pointer. Also, it should be defined by value inside functions, and properly initialized. I think the current implementation will segfault, as TAILQs are not properly initialized.

gahr:
Oh, I see you have already discovered what I pointed out. Please take some time to see how the other instances of TAILQ work around in the code, to get an idea of the paradigm. I'll be happy to answer specific questions or to discuss your design. Thanks!

yuzibo:
@gahr 
Hi, Thank you very much!
`void mutt_group_context_add(struct GroupContextHead **head, struct Group *group);
void mutt_group_context_destroy(struct GroupContextHead **head);
void mutt_group_context_add_addrlist(struct GroupContextHead *head, struct Address *a);
...
`
I think you are right what you said 

> it should be defined by value inside functions, and properly initialized. 
Each function about mutt_group_xx_ in group.h have owner GroupContext?
In my view, it seems to like a global GroupContext in there.But i don't know.
Thanks again, i will take time to research other code.

yuzibo:
@flatcap  @gahr 
Hello,ready for review :)

gahr:
> Each function about mutt_group_xx_ in group.h have owner GroupContext? In my view, it seems to like a global GroupContext in there.But i don't know.

I see you're passing a `struct GroupContest **head` around which isn't used anywhere. If you want to switch to a global variable, please explain why and perhaps expect some slight resistance :)

yuzibo:
Hi,
yes. Arguments in those functions are not used anymore.
In init.c Maybe the mutt_group_context_add_regex() is called first time in those functions.Here, if we use argument like GroupContext *gc, how we initialize the *gc*?
I don't like global variable,but as Rich mentioned, in tags.[ch],i see the usage.
Thank your help

gahr:
It looks to me like the `mutt_group_context_*` family of functions is consistently used like this within other functions:

1. initialize with `parse_group_context`
2. call various variations of `mutt_group_context_*` functions
3. clean up with `mutt_group_context_destroy`

So it looks to me like the contexts should not be global, but local to a function doing these steps. Each one has its own context that needs not be shared with the others.

yuzibo:
Hello,
I am stuck here when dealing with the problem.
As @gahr said, I know, in init.c:
```c
    if (parse_group_context(gc, buf, s, data, err) == -1)
      goto bail;
...
    if (mutt_group_context_add_regex(gc, buf->data, REG_ICASE, err) != 0)
      goto bail;
  } while (MoreArgs(s));

  mutt_group_context_destroy(gc);
  return 0;

bail:
  mutt_group_context_destroy(gc);
  return -1;
```
In fact, the *parse_group_context* call *mutt_group_context_add* first:

```c
static int parse_group_context(struct GroupContextHead *ctx, struct Buffer *buf,
                               struct Buffer *s, unsigned long data, struct Buffer *err)
{
...
    mutt_group_context_add(ctx, mutt_pattern_group(buf->data));
...
bail:
  mutt_group_context_destroy(ctx);
  return -1;
}
```
Now, i receoved the segfault in mutt_group_context_add_regex function. Does it mean function *mutt_group_context_add* is ok?
my lasted code:
`void mutt_group_context_add(struct GroupContextHead *head, struct Group *group)
{
  struct GroupContext *np = NULL;
  STAILQ_FOREACH(np, head, entries)
  {
    if (np->g == group)
      return;
  }
  np = mutt_mem_calloc(1, sizeof(struct GroupContext));
  np->g = group;
  STAILQ_INIT(head);
}
`

group.c
```diff
@@ -77,26 +78,31 @@ static int empty_group(struct Group *g)
...
-    if ((*ctx)->g == group)
+    if (np->g == group)
       return;
   }
-
-  *ctx = mutt_mem_calloc(1, sizeof(struct GroupContext));
-  (*ctx)->g = group;
+  np = mutt_mem_calloc(1, sizeof(struct GroupContext));
+  np->g = group;
+  STAILQ_INIT(head);
```
gahr: you want to insert the newly created node here, not reinitialize the whole list.

gahr:
Please try to step back and understand the *logic* of the code, not only the *mechanics* of how a node is added / removed to / from a list.

For example, in your proposed `mutt_group_context_add`, you are creating a new node `np`, which you are then not using anywhere and just leaking. Also, you are clearing the whole list for no apparent reason. Make sure you do understand and can explain:
1. what *the previous code* was doing
2. what *you code* is doing
3. the *reason* for any differences between 1. and 2.

yuzibo:
Hello,
Thank your help. I help my terrible questions not enrage you :)
```c
void mutt_group_context_add(struct GroupContextHead *head, struct Group *group)
{
  struct GroupContext *np = NULL;
  STAILQ_FOREACH(np, head, entries)
  {
    if (np->g == group)
      return;
  }
  np = mutt_mem_calloc(1, sizeof(struct GroupContext));
  np->g = group;
  STAILQ_INSERT_TAIL(head, np, entries);
}
```
The functions's goal is put group into (GroupContext)head.First, making match,if found,over the func.
else put group into head.As i know.
Is it(functions) ok?I suspect  it is not ok
.I think that i  only deal with the function well,then can process 
 *mutt_group_context_add_regex*.
Thanks

yuzibo:
Hello,
After parse_group_context(including mutt_group_context_add),the *mutt_group_context_add_regex* is called.
This is origin code:
```c
int mutt_group_context_add_regex(struct GroupContext *ctx, const char *s,
                                 int flags, struct Buffer *err)
{
  int rc = 0;

  for (; (!rc) && ctx; ctx = ctx->next)
    rc = group_add_regex(ctx->g, s, flags, err);

  return rc;
}
```
This is my indent code:
```c
int mutt_group_context_add_regex(struct GroupContextHead *head, const char *s,
                                 int flags, struct Buffer *err)
{
  int rc = 0;
  struct GroupContext *np;
  if (!rc)
  {
    STAILQ_FOREACH(np, head, entries)/*Segfault here*/
    {
      if (!rc)
      {
        rc = group_add_regex(np->g, s, flags, err);
      }
    }
  }
  STAILQ_INSERT_TAIL(head, np, entries);

  return rc;
}
```
The routine is easy to understand:match s in head->g, right?But i dont know **why** segfault here. 
gdb neomutt:
```c
Program received signal SIGSEGV, Segmentation fault.
mutt_group_context_add_regex (head=head@entry=0x0, 
    s=0x5555558d4a60 "linux-kernel@vger.kernel.org", flags=flags@entry=2, 
    err=err@entry=0x7fffffffb2a0) at group.c:190
190	    STAILQ_FOREACH(np, head, entries)
```
Thanks for your help.

gahr:
> Just make sure you pass a pointer to a properly initialized GroupContextHead structure, as you cannot iterate over a NULL or invalid list.

This. You are passing a NULL `head`, as can be seen in your crash output: `head=head@entry=0x0` .

yuzibo:
Hello,
Now, maybe i understand the functions family.
In init.c, original code:
```c
static int parse_alternates(struct Buffer *buf, struct Buffer *s,
                            unsigned long data, struct Buffer *err)
{
  struct GroupContext *gc = NULL;
  alternates_clean();
  do
  {
    mutt_extract_token(buf, s, 0);
    if (parse_group_context(&gc, buf, s, data, err) == -1)
      goto bail;
    mutt_regexlist_remove(&UnAlternates, buf->data);
    if (mutt_regexlist_add(&Alternates, buf->data, REG_ICASE, err) != 0)
      goto bail;
    if (mutt_group_context_add_regex(gc, buf->data, REG_ICASE, err) != 0)
      goto bail;
  } while (MoreArgs(s));
  mutt_group_context_destroy(&gc);
  return 0;
bail:
  mutt_group_context_destroy(&gc);
  return -1;
}
```
Here, parse_group_context() shares gc defined in parse_alternates  with address of pointer, so other functions, for example, mutt_group_context_add_regex deals with it.Obviously, it is convenient the using pointer-to-pointer in such circumstances.
DO STAILQ_xx not deal with pointer-to-pointer, right?
If  continue to use STAILQ_XX,maybe we have to define 
```c
struct GroupContext *parse_group_context(struct GroupContext *head,struct Buffer *buf, struct Buffer *s, unsigned long data, struct Buffer *err)
```
But this changes static attr of the function, is it ok?

yuzibo:
Hello,
Sorry again!
In init.c, i know the below functions does not call parse_group_context at all(maybe just in init step) and call mutt_group_add_regex() directly. I have tried to look into how to initialize the var gc.
```c
static int parse_alternates(struct Buffer *buf, struct Buffer *s,
                            unsigned long data, struct Buffer *err)
{
  struct GroupContextHead gc = STAILQ_HEAD_INITIALIZER(gc);
  alternates_clean();
  do
  {
    mutt_extract_token(buf, s, 0);
    if (parse_group_context(&gc, buf, s, data, err) == -1)
      goto bail;
    mutt_regexlist_remove(&UnAlternates, buf->data);
    if (mutt_regexlist_add(&Alternates, buf->data, REG_ICASE, err) != 0)
      goto bail;
    if (mutt_group_context_add_regex(&gc, buf->data, REG_ICASE, err) != 0)
      goto bail;
  } while (MoreArgs(s));
mutt_group_context_destroy(&gc);
  return 0;
bail:
  mutt_group_context_destroy(&gc);
  return -1;
}
```
Or here is some wrong to pass gc var to functions?
Thank you very much!

gahr:
@yuzibo, sorry I'm a bit lost here - what is your question?

To summarize: there are a few functions which use the `GroupContext` structure to keep state. These are `parse_alternates`, `parse_lists`, `parse_group`, `parse_subscribe`, and `parse_alias`. When moving from a custom linked-list to a STAILQ, these functions will need to properly initialize the STAILQ, then pass it down to any callees, like `parse_group_context`, and eventually destroy it.

yuzibo:
@gahr 
Oh, thank you.
I just initialize **GroupContextHead** in **parse_alternates**.
Now, the compiling seem to like good and i will adjust code then commit them.

yuzibo:
@gahr  @flatcap 
Ready for review,
Thanks

group.c
```diff
@@ -58,14 +58,34 @@ static void group_remove(struct Group *g)
...
+  struct GroupContext *np = NULL, *tmp = NULL;
+  STAILQ_FOREACH_SAFE(np, head, entries, tmp)
   {
-    group_remove((*ctx)->g);
-    t = (*ctx)->next;
-    FREE(ctx);
+    group_remove((np)->g);
+    STAILQ_REMOVE(head, np, GroupContext, entries);
+    FREE(&(np));
   }
```
gahr: Please use the pattern used in `mutt_list_free` in `mutt/list.c` for a quicker removal of all elements in the list.

Please remove the parentheses around single tokens, as in `(np)->g` --> `np->g` and `FREE(&(np))` --> `FREE(&np)`.

group.c
```diff
@@ -77,26 +97,30 @@ static int empty_group(struct Group *g)
...
-  struct GroupContext *p = NULL;
-  for (; *ctx; *ctx = p)
+
+  struct GroupContext *np = NULL, *tmp = NULL;
+  STAILQ_FOREACH_SAFE(np, head, entries, tmp)
   {
-    p = (*ctx)->next;
-    FREE(ctx);
+    STAILQ_REMOVE(head, np, GroupContext, entries);
+    FREE(&np);
```
gahr: Please use the pattern used in mutt_list_free in mutt/list.c for a quicker removal of all elements in the list.

group.c
```diff
@@ -141,48 +165,77 @@ static int group_remove_regex(struct Group *g, const char *s)
...
-      group_remove(ctx->g);
+    STAILQ_FOREACH_SAFE(np, head, entries, tmp)
+    {
+      if (!rc)
+      {
+        rc = group_remove_addrlist(np->g, a);
+        if (empty_group(np->g))
+          group_remove(np->g);
+      }
+      STAILQ_REMOVE(head, np, GroupContext, entries);
```
gahr: The original function doesn't seem to be touching the elements of the list - `ctx` remains linked as it was before.

group.c
```diff
@@ -141,48 +165,77 @@ static int group_remove_regex(struct Group *g, const char *s)
...
-int mutt_group_context_add_regex(struct GroupContext *ctx, const char *s,
+int mutt_group_context_add_regex(struct GroupContextHead *head, const char *s,
                                  int flags, struct Buffer *err)
 {
   int rc = 0;
 
-  for (; (!rc) && ctx; ctx = ctx->next)
-    rc = group_add_regex(ctx->g, s, flags, err);
+  struct GroupContext *np  = NULL;
+  if (!rc)
```
gahr: This first check is unnecessary

group.c
```diff
@@ -141,48 +165,77 @@ static int group_remove_regex(struct Group *g, const char *s)
...
 }
 
-int mutt_group_context_remove_regex(struct GroupContext *ctx, const char *s)
+int mutt_group_context_remove_regex(struct GroupContextHead *head, const char *s)
 {
   int rc = 0;
-
-  for (; (!rc) && ctx; ctx = ctx->next)
+  struct GroupContext *np = NULL, *tmp = NULL;
+  if (!rc)
```
gahr: This first check is unnecessary

group.c
```diff
@@ -141,48 +165,77 @@ static int group_remove_regex(struct Group *g, const char *s)
...
-      group_remove(ctx->g);
+    STAILQ_FOREACH_SAFE(np, head, entries, tmp)
+    {
+      if (!rc)
+      {
+        rc = group_remove_regex(np->g, s);
+        if (empty_group(np->g))
+          group_remove(np->g);
+      }
+      STAILQ_REMOVE(head, np, GroupContext, entries);
```
gahr: Again, the original code did not touch the linked list

yuzibo:
Hello,
@gahr  @flatcap 
Ready for review.
Thanks

yuzibo:
How about implement  mutt_group_context_destroy() below:
```c
void mutt_group_context_destroy(struct GroupContextHead *head)
{
  struct GroupContext *np = STAILQ_FIRST(head), *next = NULL;
  STAILQ_FOREACH_SAFE(np, head,entries, next)
  {
    STAILQ_REMOVE(head,np,　GroupContext,  entries);
    FREE(&np->g);
    FREE(&np);
  }
  STAILQ_INIT(head);
}
```

flatcap:
Ooh, function comment blocks!  Thanks :-)

group.c
```diff
@@ -58,15 +58,20 @@ static void group_remove(struct Group *g)
...
+  STAILQ_FOREACH_SAFE(np, head, entries, next)
   {
-    group_remove((*ctx)->g);
-    t = (*ctx)->next;
-    FREE(ctx);
+    group_remove(np->g);
+    STAILQ_REMOVE(head, np, GroupContext, entries);
+    FREE(&np);
   }
+  STAILQ_INIT(head);
```
gahr: You don't need a STAILQ_INIT if you call STAILQ_REMOVE inside the loop. Please note that STAILQ_REMOVE runs in O(n), which is why I was suggesting to follow the approach used in `mutt_list_free`.

group.c
```diff
@@ -77,26 +82,39 @@ static int empty_group(struct Group *g)
...
 {
-  struct GroupContext *p = NULL;
-  for (; *ctx; *ctx = p)
+  struct GroupContext *np = STAILQ_FIRST(head), *next = NULL;
+  while (np)
   {
-    p = (*ctx)->next;
-    FREE(ctx);
+    next = STAILQ_NEXT(np, entries);
+    FREE(&np->g);
```
gahr: The original version did not free `np->g`. Why the change?

group.c
```diff
@@ -141,48 +159,87 @@ static int group_remove_regex(struct Group *g, const char *s)
...
   {
-    rc = group_remove_addrlist(ctx->g, a);
-    if (empty_group(ctx->g))
-      group_remove(ctx->g);
+    if (!rc)
+    {
+      rc = group_remove_addrlist(np->g, a);
+      if (empty_group(np->g))
+        group_remove(np->g);
+    }
```
gahr: How about quitting the function earlier as soon as `rc == 1`?

```
STAILQ_FOREACH(....)
{
    rc = group_remove_addrlist(np->g, a);
    ....
    if (rc)
        return rc;
}
```

group.c
```diff
@@ -141,48 +159,87 @@ static int group_remove_regex(struct Group *g, const char *s)
...
-
+  struct GroupContext *np = NULL;
+  STAILQ_FOREACH(np, head, entries)
+  {
+    if (!rc)
+    {
+      rc = group_add_regex(np->g, s, flags, err);
+    }
+  }
   return rc;
```
gahr: Same as above

group.c
```diff
@@ -141,48 +159,87 @@ static int group_remove_regex(struct Group *g, const char *s)
...
-    rc = group_remove_regex(ctx->g, s);
-    if (empty_group(ctx->g))
-      group_remove(ctx->g);
+    if (!rc)
+    {
+      rc = group_remove_regex(np->g, s);
+      if (empty_group(np->g))
+        group_remove(np->g);
+    }
   }
```
gahr: Same as above

init.c
```diff
@@ -3817,10 +3823,11 @@ static bool get_hostname(void)
 
   if (!Hostname)
   {
-    /* now get FQDN.  Use configured domain first, DNS next, then uname */
+/* now get FQDN.  Use configured domain first, DNS next, then uname */
```
gahr: Original was probably better

yuzibo:
Hello,
@gahr @flatcap 
Ready for review. Line-splits in init.c was introduced by clang-format.I think we can ignore it now.
Thanks

group.c
```diff
@@ -58,15 +58,21 @@ static void group_remove(struct Group *g)
...
 
-int mutt_group_context_clear(struct GroupContext **ctx)
+/**
+ * mutt_group_context_clear: Clear groups of GroupContext list
+ * @param head Head of the GroupContext list
+ */
+int mutt_group_context_clear(struct GroupContextHead *head)
 {
-  struct GroupContext *t = NULL;
-  for (; ctx && *ctx; (*ctx) = t)
```
yuzibo: I suspect here i missed a test for(ctx), right?

yuzibo:
Hello,
@gahr  @flatcap  
Ready for review. 
Thanks,

flatcap:
I've:
- squashed and rebased the commits
- removed a few unexpected reformatting changes
- tidied a few doxygen comment blocks

yuzibo:
@flatcap 
Thank you.
I am looking forward to merge the RP and want to try other easy task.
I have a question:In fact, i make an original  plan to squash these commit only if the code is ok. Now,you help me to squash it and i am wonder how i update my local devel/convert-groupcontext-stailq branch?The way one i thought is i delete the Neomutt git repositories(sure, the repositories that i want to solve GroupContext issue) then git clone -b devel/convert-groupcontext-stailq. But this way wastes time and internet resources.git pull maybe ok, but it looks terrible,right?

gahr:
git reset --hard origin/devel/convert-groupcontext-stailq

flatcap:
> i make an original plan to squash these commit only if the code is ok.

That's fine.

> Now,you help me to squash it

Yes.  I had made a lot of changes to `init.c` so your PR had conflicts.
When I cause trouble like that, I'm happy to help. 

> how i update my local devel/convert-groupcontext-stailq branch?

as @gahr says

> git pull maybe ok, but it looks terrible,right?

`git pull` is actually two commands.  It does a `git fetch` to get the changes, then a `git merge`.
The way I use git requires a bit more understanding, but it leaves us a very tidy history.
(This is why I'm happy to help rebase)

> wastes time and internet resources

As you can probably imagine, I have a **LOT** of separate copies of the source tree.

To keep them small I have a master copy of the repo `~/.neomutt.git` (a bare repo)
I keep this up-to-date with `cron` (`git fetch --all`).

When I clone a new working directory, I use `git clone --reference ~/neomutt.git https://github...`
Git will only clone the bits it hasn't got locally.

**IMPORTANT**: You must not delete the master copy, or all the other repos will break.

yuzibo:
Wow,
Thanks all.
Delicious git food for me :)

yuzibo:
@flatcap
Thanks!
I think it is very important be able to test new code.Especially during in development.
Anyway, i will keep eye on the issue to improve it as soon as get feedback.

yuzibo:
Hello, 
Suddenly, i think 
```c
STAILQ_HEAD(GroupContextHead, GroupContext);
```
would be better for:

```c
STAILQ_HEAD(GroupContextList, GroupContext);
```
Did you think so?

yuzibo:
Oh,
I did not notice @gahr  recently commit,
Thanks.

flatcap:
I've rebased this branch against master - the result is in [devel/convert-groupcontext-stailq2](https://github.com/neomutt/neomutt/tree/devel/convert-groupcontext-stailq2)

It builds and the diffs _look_ OK (similar to the un-rebased diffs).

`group.c` has a dependency on `struct Address` which is now in libemail, so it probably needs moving there.

If you're happy, please reset this branch and delete mine.  Thanks.

flatcap:
> > group.c has a dependency on struct Address which is now in libemail,
> > so it probably needs moving there.

> why we need to do it.
> In group.c, which had included "email/address.h", not Ok?

The goal for each library is to wrap up a functional piece of the code.
If the code is well-designed, each library shouldn't have any dependencies on NeoMutt.
The libraries can depend on other libraries.

The 'mutt' library is slightly special.  It **only** depends on the C library.
The `GroupContext` uses `struct Address` which is in 'email' library, so it should live there, too.

Issue #1189, opened 2018-05-07
neomutt fails to decode (some?) RFC2047-encoded headers


otizonaizit:
Since version 20180223 neomutt fails to properly decode headers encoded following RFC2047. For example, an email with Subject header:

```
Subject: =?UTF-8?Q?Sicherheitsl=C3=BCcke in praktisch allen IT-Systemen?=
```
should be shown in the message index and in the pager as:

```
Subject: Sicherheitslücke in praktisch allen IT-Systemen
```

The encoded version is shown instead. Version 20171215 works fine on the same mailbox. Also mutt 1.9.4 works on the same mailbox. My conclusion is that there must be something in neomutt that happened between 20171215 and 20180223.
By reading the RFC2047 I am not sure that such headers are actually standard conform, but they were working fine until now, so I'm guessing this is a bug in neomutt?

I attached an example mailbox (with a `.txt` extension to be able to attach it to the GitHub issue)
I reported this bug also in [Debian](https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=893116), but I guess it is more relevant here.

Thank you for your work on neomutt :)))

[test.txt](https://github.com/neomutt/neomutt/files/1979501/test.txt)

gahr:
I'll look into this one

gahr:
Oh - simply by looking at the mbox you have attached it's clear what's wrong: the subject contains spaces within an `encoded-text` production, which is against [RFC2047, §2](https://tools.ietf.org/html/rfc2047#section-2).

I think there was an option in previous versions that allowed NeoMutt to deal with these invalid representations. I'll see whether we can accept those in a non-breaking way.

otizonaizit:
Ah OK, I had no doubt the headers were not RFC compliant: they come from a Microsoft client ;)

But, version 20171215 of neomutt displays the mailbox "correctly", in the sense that the headers are decoded and can be read by humans, so I think that some recent change in neomutt is causing these invalid headers to not be displayed the way they were before...

otizonaizit:
By the way, option `rfc2047_parameters` still exists and it was set to `yes`. But I guess that option has nothing to do with the issue we are discussing here...

gahr:
> But, version 20171215 of neomutt displays the mailbox "correctly"

Yes, we used to be less strict, see

https://github.com/neomutt/neomutt/blob/ae61285170533b4be544e738cdd2eedefd1856ff/rfc2047.c#L768-L769

[...]

PR #1190, opened 2018-05-07
Accept whitespace in RFC2047 encoded-text productions

gahr:
Let's be more liberal and accept RFC2047 encoded-test productions that include white space. This is explicitly forbidden by RFC2047, but Mutt has been accepting them since (almost) forever, so let's not break existing clients.

#1189


mschilli87:
Maybe we could log a warning to raise awareness for this? I know it's a long shot but if Neomutt accepts non-standard headers while encouraging people to report the issue to the sender's client's maintainer as an RFC violation, we might eventually be able to drop this again. :wink:

PR #1191, opened 2018-05-08
Update zh_CN translation

l2dy:
* **What does this PR do?**
Fix some Chinese (Simplified) translation.

@flatcap Should translation PRs be made on the translate branch?

     for that)

   - All builds are passing

### DIFF
--- a/po/zh_CN.po
+++ b/po/zh_CN.po
@@ -525,7 +525,7 @@ msgid ""
 "s(p)am/(l)abel?: "
 msgstr ""
 "按日期(d)/发信人(f)/收信时间(r)/标题(s)/收信人(o)/线索(t)/不排(u)/大小(z)/分数(c)/"
-"垃圾邮件(p)/标签()l反向排序?: "
+"垃圾邮件(p)/标签(l)反向排序?: "
 
 #. L10N: The highlighted letters must match the "Rev-Sort" options
 #: commands.c:564
@@ -535,7 +535,7 @@ msgid ""
 "(l)abel?: "
 msgstr ""
 "按日期(d)/发信人(f)/收信时间(r)/标题(s)/收信人(o)/线索(t)/不排(u)/大小(z)/分数(c)/"
-"垃圾邮件(p)/标签()l反向排序?: "
+"垃圾邮件(p)/标签(l)反向排序?: "
 
 #. L10N: These must match the highlighted letters from "Sort" and "Rev-Sort"
 #: commands.c:568
@@ -4631,7 +4631,7 @@ msgstr "储存信件以便稍后寄出"
 #: opcodes.h:70
 #, fuzzy
 msgid "send attachment with a different name"
-msgstr "编辑附件的传输编码"
+msgstr "使用另一名字发送附件"
 
 #: opcodes.h:71
 msgid "rename/move an attached file"
@@ -5082,9 +5082,8 @@ msgid "mark the current thread as read"
 msgstr "标记当前线索为已读取"
 
 #: opcodes.h:180
-#, fuzzy
 msgid "jump to root message in thread"
-msgstr "跳到本线索中的父信件"
+msgstr "跳到本线索中的根信件"
 
 #: opcodes.h:181
 msgid "set a status flag on a message"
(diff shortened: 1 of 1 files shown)


l2dy:
PR builds shouldn't attempt deploy to website.

yuzibo:
Hello,
１.
"Decode-save tagged to mailbox"
"解码保存到信箱" 是否为"被标记的解码保存到信箱"更好一些？　：）
２.
Maybe dropping the "fuzzy" string to update translate page:
https://neomutt.org/translate

l2dy:
1. I was using Poedit and it didn't show me the plural source string... I'll leave it alone since I don't know how to handle plural translation.
2. I'm keeing `fuzzy` unless I'm sure it's translated correctly.

yuzibo:
Hi,
1. The plural string does not effect Chinese translate.
In po/zh_CN.po head section, here is 
```bash
...
Language-Team: i18n-zh <i18n-zh@googlegroups.com>\n"
"Language: zh_CN\n"
"MIME-Version: 1.0\n"
"Content-Type: text/plain; charset=UTF-8\n"
"Content-Transfer-Encoding: 8bit\n"
"Plural-Forms: nplurals=1; plural=0;\n"
```
I think Chinese version is "nplurals=1"
2. You are right.But after Rich merging,you have to commit again.

l2dy:
Problem is this isn't about plural words. The singular msgid doesn't involve tagging but the plural one does.

```
msgid "Decode-save to mailbox"
msgid_plural "Decode-save tagged to mailbox"

msgid "Decrypt-save to mailbox"
msgid_plural "Decrypt-save tagged to mailbox"
```

yuzibo:
Maybe you can refer to ga.po to modify the "Plural-forms" :)

l2dy:
>In Poedit, I see [...]

Yes, I noticed that later and edited my comment.

Poedit formats the .po file every time I save it, so I'm committing the formatting change in a944a159043df29b9360dff1853216b06e5844ba. I can discard this commit if anyone objects it, but it will be difficult to contribute with Poedit unless Poedit can be configured to preserve the original formatting.

l2dy:
Can I merge this pull request now?

flatcap:
I've:
- merged all your changes
- reverted the bogus plurals
- changed the build to prevent unnecessary wrapping

@l2dy 
Now that the strings are unwrapped, you'll need to turn off wrapping in Poedit (Edit, Preferences, Advanced).
All the other changes were valid (and would have happened when I sync'd the translations)

l2dy:
@flatcap

Thanks. Should I always make pull requests or directly push to the translate branch?

flatcap:
@l2dy 
Replied on the Translator Team discussion board: https://github.com/orgs/neomutt/teams/translators/discussions/8

reisradomil:
> @reisradomil Do you have time to undo those two uses of ngettext()?

@flatcap Sorry for being late to the party. As I see you reverted that change yourself.
After reverting we now have this:
```
if (h)
  mutt_str_strfcpy(prompt, _("Bounce message to: "), sizeof(prompt));
else
  mutt_str_strfcpy(prompt, _("Bounce tagged messages to: "), sizeof(prompt));
````
Would it be worth to have something like
```
if (h)
  mutt_str_strfcpy(prompt, _("Bounce message to: "), sizeof(prompt));
else
  mutt_str_strfcpy(prompt, ngettext("Bounce tagged message to", "Bounce tagged messages to: ", msgcount), sizeof(prompt));
````
i.e. formulate the question according to the actual number of tagged messages? Or should we stay with the "generic plural" phrase?

flatcap:
@reisradomil 

> Sorry for being late to the party

No worries.

> As I see you reverted that change yourself.

Yes, I was fixing a bunch of other things, anyway.
I was keen to give the translators as much time as possible before the release.

> Would it be worth to have something like...
> i.e. formulate the question according to the actual number of tagged messages?
> Or should we stay with the "generic plural" phrase?

Hmm... interesting.
The number of tagged messages will most likely be plural.
Perhaps for the singular-tagged-case, we should use the singular-non-tagged message.
Perhaps that's more confusing.

Leave it for now.

Thanks.

PR #1193, opened 2018-05-08
Autocrypt support for incoming messages

Valodim:
This PR adds some support for [Autocrypt](https://autocrypt.org), in the form of an autocrypt_cmd setting. The setting contains a command name, which is called at message view time and receives Autocrypt-related headers on its stdin. Specifically, it receives the `From`, `To`, `Cc`, `Autocrypt` headers, and also the `Autocrypt-Gossip` header from the outermost decrypted mime part. This information is sufficient to maintain a database of keys in an Autocrypt-compliant manner.

This isn't really camera-ready yet, in particular I didn't document much. But I would be interested in feedback on the concept and implementation.

I had to add a parameter to a bunch of functions, which doesn't feel too elegant. If anyone has a better approach in mind, I'd be all ears :) otherwise this works quite well and isn't too intrusive on other matters.

### DIFF
--- a/commands.c
+++ b/commands.c
@@ -161,7 +161,26 @@ int mutt_display_message(struct Header *cur)
   if (Context->magic == MUTT_NOTMUCH)
     chflags |= CH_VIRTUAL;
 #endif
-  res = mutt_copy_message_ctx(fpout, Context, cur, cmflags, chflags);
+
+  FILE *fpout_autocrypt = NULL;
+  pid_t autocryptpid = -1;
+  if (AutocryptCmd && *AutocryptCmd)
+  {
+    autocryptpid = mutt_create_filter(AutocryptCmd, &fpout_autocrypt, NULL, NULL);
+    if (autocryptpid < 0)
+    {
+      mutt_perror(_("Can't create Autocrypt cmd process"));
+      return 0;
+    }
+  }
+
+  res = mutt_copy_message_ctx_autocrypt(fpout, Context, cur, cmflags, chflags, fpout_autocrypt);
+
+  if (fpout_autocrypt)
+  {
+    mutt_file_fclose(&fpout_autocrypt);
+    mutt_wait_filter(autocryptpid);
+  }
 
   if ((mutt_file_fclose(&fpout) != 0 && errno != EPIPE) || res < 0)
   {
--- a/copy.c
+++ b/copy.c
@@ -72,11 +72,12 @@ static int copy_delete_attach(struct Body *b, FILE *fpin, FILE *fpout, char *dat
  * wrap headers much more aggressively than the other one.
  */
 int mutt_copy_hdr(FILE *in, FILE *out, LOFF_T off_start, LOFF_T off_end,
-                  int flags, const char *prefix)
+                  int flags, const char *prefix, FILE *out_autocrypt)
 {
   bool from = false;
   bool this_is_from = false;
   bool ignore = false;
+  bool is_autocrypt = false;
   char buf[LONG_STRING]; /* should be long enough to get most fields in one pass */
   char *nl = NULL;
   char **headers = NULL;
@@ -215,6 +216,7 @@ int mutt_copy_hdr(FILE *in, FILE *out, LOFF_T off_start, LOFF_T off_end,
       }
 
       ignore = true;
+      is_autocrypt = false;
       this_is_from = false;
       if (!from && (mutt_str_strncmp("From ", buf, 5) == 0))
       {
@@ -225,10 +227,23 @@ int mutt_copy_hdr(FILE *in, FILE *out, LOFF_T off_start, LOFF_T off_end,
       else if (buf[0] == '\n' || (buf[0] == '\r' && buf[1] == '\n'))
         break; /* end of header */
 
+      if (out_autocrypt &&
+          ((mutt_str_strncasecmp("Autocrypt:", buf, 10) == 0) ||
+            (mutt_str_strncasecmp("From:", buf, 5) == 0) ||
+            (mutt_str_strncasecmp("Date:", buf, 5) == 0) ||
+            (mutt_str_strncasecmp("To:", buf, 3) == 0) ||
+            (mutt_str_strncasecmp("Cc:", buf, 3) == 0))) {
+        is_autocrypt = true;
+      }
+
       /* note: CH_FROM takes precedence over header weeding. */
       if (!((flags & CH_FROM) && (flags & CH_FORCE_FROM) && this_is_from) &&
           (flags & CH_WEED) && mutt_matches_ignore(buf))
       {
+        if (is_autocrypt)
+        {
+            fputs(buf, out_autocrypt);
+        }
         continue;
       }
       if ((flags & CH_WEED_DELIVERED) &&
@@ -279,6 +294,11 @@ int mutt_copy_hdr(FILE *in, FILE *out, LOFF_T off_start, LOFF_T off_end,
 
       ignore = false;
     } /* If beginning of header */
+
+    if (is_autocrypt)
+    {
+      fputs(buf, out_autocrypt);
+    }
 
     if (!ignore)
     {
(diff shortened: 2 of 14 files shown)


Valodim:
I used a simple script `cat >> ~/test.log` to see the output. That should write out From/To/Date/Cc headers for all messages, plus the Autocrypt header if present. Since the code doesn't care, you can use any mail to test this - newer versions of K-9 and Enigmail include this header, so if you receive mail from people who use that at all you probably have some test messages.

The autocrypt-gossip header is a bit more tricky: It is output only if it's in the mime header of the inner decrypted part, and only if that part has a valid signature. I can send such a message to you if you have a pgp key.

flatcap:
> I can send such a message to you if you have a pgp key.

I do, but that'll only help me.

I've just created a new gpg key: [neo-test-key](https://github.com/neomutt/neomutt/files/1984919/neo-test-keys.txt) (with a temporary email address)
Fingerprint: **F83B 8E64 0F87 04A5 E819 1E36 4FBF 54B7 22AC 427F**
That's both the public and private parts, with no passphrase.

Please can you send me a couple of Autocrypted emails:
- Simple: short message
- Complex: a couple of small attachments

If I publish them too, then any developer can test your code.

If this is a crazy idea, let me know :-)

flatcap:
Message received and understood!

Valodim:
Found a bug :) fixing in a second

flatcap:
For anyone wishing to test this feature:

- Save [this gpg key](https://github.com/neomutt/neomutt/files/1984919/neo-test-keys.txt)
- Import it into gpg, e.g. `gpg --import neo-text.txt`
- Save [this email](https://github.com/neomutt/neomutt/files/1985214/autotest.mbox.txt) as "autotest.mbox"
- Run `neomutt -f autotest.mbox`

You should see an email from Vincent with text/html multiparts and two attachments.

Valodim:
I fixed a bug with missing newlines. I'd still like to unfold the regular headers, to be consistent with the gossip ones. Is there a good way to fputs a string without its trailing newline?

gahr:
So if I understand this correctly, the idea would be to dump part of the message to some 3rd party external program, that would then maintain a database using this information.

Wouldn't it be easier and more generic to just use pipe-message and a filter, like this one, instead of trying to fit mutt's to this particular use case?

```awk
#!/usr/bin/awk -f

BEGIN { FS = ":"; }

# Terminate after headers
/^$/ { exit; }

/^[^ \t]/ {
    in_hdr = 0;
}

/^[ \t]/ {
    if (in_hdr == 1) {
        print;
    }
}

tolower($1) ~ /^(subject|from|date|to|cc|autocrypt)$/ {
    print;
    in_hdr=1;
}
```

gahr:
How about redefining <enter> as something like this:

```
 macro index <return> '<enter-command>set my_wait_key=$wait_key nowait_key<enter>\
<pipe-entry>/path/to/script.awk | /path/to/autocrypt/external/cmd<enter>\
<enter-command>set wait_key=$my_wait_key<enter><display-message>'
````

Valodim:
I thought about that, but in the end any solution that uses hooks or macros is a hack. The main thing that's a problem is that gossip headers can't be accessed that way, since they are part of the encrypted payload. But even without that, hooks and macros introduce delay and error cases, and just general friction in the UI.

gahr:
I disagree. Macros and hooks are only of the few ways we have to customize and extend the hard-coded NeoMutt workflow. Do you envision having to implement a dedicated `mutt_copy_message_ctx_FOO` and similar for any slightly related functionality that we might want to add in the future? I fear AutoCrypt won't be the last to come up.

I think Autocrypt support and anything similar (see Keybase stuff in contrib) should definitely be provided as an extension. I simply don't see it in pair with the core RFCs implemented hard-coded.

Valodim:
Ah, I see where you are coming from :) There is of course always the question of maintenance and feature creep.

As I mentioned, I tried with the existing hooks to implement this, but was ultimately unable to get to a satisfying solution. If there are other ways I missed, I would be interested in trying them. In particular there doesn't seem to be a way with the existing hooks to get to the gossip header, without redundantly decrypting *and* mime-parsing the message.

I had considered making this a more generic solution (i.e. pipe the whole raw message upon view instead of just the Autocrypt related parts), but in the end it would still mean duplicating mime parsing (and gosh mime is complex!) in the autocrypt backend that mutt as a MUA already does.

In regards to `mutt_copy_message_ctx_FOO`: In the interest of keeping the patch simple and small I went for the more trivial but specialized solution here, but that's an implementation detail. If more of these come up it might make sense to use a more abstract pattern, similar to the `State` struct that is used for the mime handler functions.

gahr:
Yeah, if anything needs to be done within NeoMutt, I would probably be more in favour of a self-contained solution that just calls out to a dedicated API to do the AutoCrypt stuff before going into the pager or something. Ideally, I'd love to see the autocrypt machinery in mutt/ and unit-testable :)

Am I missing where you decrypt the autocrypt header to extract the gossip, or is it still coming?

gahr:
I am sorry, I'm sure I'm missing something here :) Using `autocrypt_cmd=cat > /path/to/some/file.txt` with the sample email above gives me a file containing the Date, Autocrypt, To, and From headers. The gossip is not there.

gahr:
You could still get the gossip using a simple pipe by using `gpg --decrypt` and pipe the output to the awk script I pasted above.

Valodim:
> Using autocrypt_cmd=cat > /path/to/some/file.txt with the sample email above gives me a file containing the Date, Autocrypt, To, and From headers. The gossip is not there.

Oh. Yeah that's a bug. I do get Autocrypt-Gossip in my output, for this message too. Sorry about that, I'm not very used to writing C code. I'll investigate.

> You could still get the gossip using a simple pipe by using gpg --decrypt and pipe the output to the awk script I pasted above.

I'm not a fan of this approach at all: It requires decrypting the message twice (with all that implies: passphrase cache, validity considerations, ...), an additional coupling with the decryption mechanism, and requires parsing a header out of a mime structure independently from the MUA that is just just about to parse and display the same data. And that still disregards the additional friction from doing pipe-message calls in a message-hook or similar somewhere in the UI, and asking the user to put longish configuration lines into their muttrc that users shouldn't be required to understand to use autocrypt. I'd very much prefer to have a proper and purposeful mechanism (at +108/-25 lines) here than something that is held together by duct tape and pipes.

Valodim:
I think the bug about gossip headers was with "good" signatures: I thought it meant just valid ones, but it's actually "trusted" ones. Fixed in the latest commit.

dkg:
thanks for considering this patch series!

The autocrypt spec is really designed with tight integration with MUAs in mind.  To the extent that neomutt can have autocrypt logic as a first-class citizen (instead of requiring users to adopt a multi-line configuration change that incurs additional decryption operations and other stumbling blocks), that will be much more useful to neomutt users.

Valodim:
Fixed a bug, rebased this on current master, and I'm also now folding headers into one line during output. Please review, I'll add documentation when you're happy with this technically :)

tshirtman:
Just learned about autocrypt, and wondered if neomutt had something for it, happy to see there is this PR! Is there anything preventing merging it?

flatcap:
I'm closing this PR, which will come as a disappointment, but not a surprise.

The problems are:
- It's a very complicated feature
- The code is complex (and we're making big changes underneath it)
- NeoMutt has very few developers

When this is merged it will require time and effort to maintain, bug-fix and help the users.
Until there are more active developers, this will be too much of a burden.

I've archived the feature to the [neomutt-old](https://github.com/neomutt/neomutt-old) repo until we have the developers to work on it.

opnsrcr:
This is too bad. Thanks for your work @Valodim and consideration @flatcap 

I have no tried to test this patch, but @Valodim is it possible you make this optional for users to configure separately despite the less than ideal concerns @dkg mention

> ... requiring users to adopt a multi-line configuration change that incurs additional decryption operations and other stumbling blocks ...
 
being a mutt user I am used to this =)

Valodim:
> I have no tried to test this patch, but @Valodim is it possible you make this optional for users to configure separately despite the less than ideal concerns @dkg mention

I guess it's already configurable with a macro similar to the one mentioned above. But that's not the user experience I would go for here.

I definitely still believe it's a useful feature, seeing the steadily increasing number of autocrypt headers in the wild and active interest in this PR. However I'm not using mutt anymore myself, so I won't be able to maintain this patch. I'm happy to provide  guidance and testing though, if anyone else wants to pick this up.

Hopefully the "big changes underneath" mentioned above will make this feature easier to implement in the future?

Issue #1152, opened 2018-04-20
Script the deployment of Code Docs


flatcap:
This is an "easy" task ([find more](https://github.com/neomutt/neomutt/issues?q=label:"good%20first%20issue")) and it's open to everyone.  [_read more..._](https://neomutt.org/dev/easy-tasks)

## Description

NeoMutt's uses Doxygen to generate documentation for its code.  
Specially-formatted comment blocks are turns into HTML and linked together.

- https://neomutt.org/code/pages.html

Currently, updating the docs is done by hand, but all the steps could easily be automated and given to Travis.

### Part 1: Clone necessary repos

There are three repos:

- https://github.com/neomutt/neomutt
  Source code.  Check out branch 'doxygen'

- https://github.com/neomutt/neomutt.github.io
  Website.  The `code` dir is a git submodule 

- https://github.com/neomutt/doxygen
  HTML docs.

### Part 2: Generate the docs

There are a few steps to generate the HTML docs (see `.travis.yml`) 
Scripts can be saved in the `.travis` directory.

### Part 3: Update the repos

- Delete all the files in the doxygen repo.
- Copy the docs from the neomutt repo's `html` directory to the doxygen directory.
- Commit those changes
- Commit the submodule change to the website repo

### Part 4: Push the results

To allow automatic pushing to GitHub, it's necessary to create an SSH key and give it permissions.  
This is documented, here:

- https://neomutt.org/dev/deploy

- Push the doxygen changes
- Push the website changes
- Wait patiently
- Check the results on the website

### Part 5: Test on Travis

Once everything's working, change `.travis.yml` to run the script and push everything to GitHub.

- https://travis-ci.org/neomutt/neomutt

jamesbvaughan:
Hey I'm new and interested in working on this! I'm on vacation for the weekend so I probably won't get started before Monday.

flatcap:
Welcome to NeoMutt, @jamesbvaughan 

Great!

This should be a fun one!
All the pieces are there, it just needs a bit of magic.

Feel free to ask lots of questions.

Thanks.

jamesbvaughan:
@flatcap For something like this, is it better to put everything into a script (like "deploy-code-docs.sh") and call that from .travis.yml or to put all the steps directly in .travis.yml?

flatcap:
I think a script (or family of scripts) would be best.
Travis works well when you have a matrix of test cases, but this situation is pretty linear.
[...]

PR #1201, opened 2018-05-11
[WIP] Script the deployment of Code Docs

jamesbvaughan:
* **What does this PR do?**
  - adds a new step in the script in `.travis.yml` to deploy updated code docs

  - all of `.travis/push-docs.sh`
  - my changes in `.travis.yml`

  - #1152


jamesbvaughan:
@flatcap I just finished midterm exams and had some time this morning to work on this. Let me know if I'm headed in the right direction and going about this the right way.

.travis.yml
```diff
@@ -29,4 +29,5 @@ script:
   - doxygen doxygen/doxygen.conf
   - grep -v "$IGNORE" doxygen-build.txt | tee tmp.txt
   - test ! -s tmp.txt
+  - .travis/push-docs.sh
```
toogley: As far as i understand, this is only generating the doxygen related things, right? I think the name `push-docs.sh` is misleading then, as it doesn't touch our [our neomutt guide](https://neomutt.org/guide/index.html).

.travis.yml
```diff
@@ -29,4 +29,5 @@ script:
   - doxygen doxygen/doxygen.conf
```
toogley: can we replace `addrbook.c body.c complete.c compress.c [...]`  with `./*.c` ? 

Might have nothing to do with your PR, but maybe we could fix that here.

.travis/push-docs.sh
```diff
@@ -0,0 +1,21 @@
...
+# exit early if any command fails
+set -e
+
+git clone --recursive https://github.com/neomutt/neomutt.github.io
+
+(
+cd neomutt.github.io/code
+
+# Remove the old docs
+git rm -r ./*
```
toogley: As that makes the travis log currently hard to read, you might want to silence the output of this line.

.travis.yml
flatcap: This list will become `*.c` in the future.
For now, it's the files that I've tidied.

flatcap:
I've just created two new repos:
- https://github.com/neomutt/test-neomutt.github.io
  An exact copy of the website
- https://github.com/neomutt/test-doxygen
  Empty repo (the original's a bit big for playing around with)

The two repos have deployment keys on them
- travis-deploy-doxygen to push to test-doxygen
- travis-deploy-website to push to test-neomutt.github.io

I've put the encrypted keys in `.travis/`

Using these repos and keys you should be able to push.

flatcap:
Hi @jamesbvaughan 
Thanks for your help on this.
I finished off the last few details, and now it's up and ~running~ automating :-)

PR #1204, opened 2018-05-12
sync with doc tidying in 77155bb

sthen:
77155bb renamed some doc files but doc/Makefile.autosetup wasn't changed to match, this fixes it.


flatcap:
Thanks.  Sorry about that.
I tested the install/uninstall and after a flurry of text, I saw a successful return code!

PR #1206, opened 2018-05-13
ci_send_message: small memory leak fix in x_original_to part

P-EB:
@flatcap this should do it.


gahr:
As mentioned on IRC, how about copying a single address by replacing `mutt_addr_copy_list` with `mutt_addr_copy` a few lines above?

P-EB:
@gahr right, I don't know why I discarded this alternative. I'll give it a second thought.

P-EB:
Ok, indeed I missed that.

Issue #1213, opened 2018-05-17
$hostname is empty during the configuration files reading process


doronbehar:
## Expected Behaviour

I want to `source /home/doron/.neomutt/$hostname.rc`. `$hostname.rc` is supposed to be a machine specific rc file which can be synced across different machines along side other machine dependent configuration files.

## Actual Behaviour

I get an error:
```
Error in /home/doron/.neomutt/neomuttrc, line 1: source: file /home/doron/.neomutt/.rc could not be sourced
```

## Steps to Reproduce

Just write something like this in your `neomuttrc`:
```
source ~/$hostname.rc
```
And put a real file in the specified location.

## How often does this happen?

- Always

## When did it start to happen?

- When I upgraded to version `20180512`. This bug was not present in the previous version.

## NeoMutt Version

```
NeoMutt 20180512
Copyright (C) 1996-2016 Michael R. Elkins and others.
NeoMutt comes with ABSOLUTELY NO WARRANTY; for details type 'neomutt -vv'.
NeoMutt is free software, and you are welcome to redistribute it
under certain conditions; type 'neomutt -vv' for details.

System: Linux 4.16.8-1-ARCH (x86_64)
ncurses: ncurses 6.1.20180127 (compiled with 6.1.20180127)
libidn: 1.34 (compiled with 1.34)
hcache backends: gdbm

Compiler:
Using built-in specs.
COLLECT_GCC=cc
COLLECT_LTO_WRAPPER=/usr/lib/gcc/x86_64-pc-linux-gnu/8.1.0/lto-wrapper
Target: x86_64-pc-linux-gnu
Configured with: /build/gcc/src/gcc/configure --prefix=/usr --libdir=/usr/lib --libexecdir=/usr/lib --mandir=/usr/share/man --infodir=/usr/share/info --with-bugurl=https://bugs.archlinux.org/ --enable-languages=c,c++,ada,fortran,go,lto,objc,obj-c++ --enable-shared --enable-threads=posix --enable-libmpx --with-system-zlib --with-isl --enable-__cxa_atexit --disable-libunwind-exceptions --enable-clocale=gnu --disable-libstdcxx-pch --disable-libssp --enable-gnu-unique-object --enable-linker-build-id --enable-lto --enable-plugin --enable-install-libiberty --with-linker-hash-style=gnu --enable-gnu-indirect-function --enable-multilib --disable-werror --enable-checking=release --enable-default-pie --enable-default-ssp
Thread model: posix
[...]

PR #1215, opened 2018-05-17
Determine hostname before loading RC file

Austin-Ray:
Moved `get_hostname()` to before of the logic for loading the RC file.

If the RC file uses `$hostname`, but lacks `set hostname` then Neomutt
expands it to an empty string. Moving `get_hostname()` to before the RC
file logic allows hostname to have a pseudo-default value, which the RC
file can override. This prevents Neomutt from substituting an empty
string for `$hostname`.

#### What does this PR do?

Loads the hostname before trying to parse the RC file to prevent the substitution of an empty string for `$hostname`

#### What are the relevant issue numbers?
Issue #1213

### DIFF
--- a/init.c
+++ b/init.c
@@ -4063,6 +4063,9 @@ int mutt_init(int skip_sys_rc, struct ListHead *commands)
     }
   }
 
+  if (!get_hostname())
+    return 1;
+
   /* Read the user's initialization file.  */
   struct ListNode *np;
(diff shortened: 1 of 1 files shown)


flatcap:
@Austin-Ray 
Thanks for the bisect and PR, but I'd rather keep the new behaviour.

Issue #1203, opened 2018-05-12
Fix plural messages


flatcap:
@mgedmin pointed out some more messages that may need the `ngettext()` treatment.

```c
mx.c:634:          mutt_error(_("message(s) not deleted"));
ncrypt/crypt.c:95: mutt_message(_("Passphrase(s) forgotten."));
recvattach.c:841:  if (query_quadoption(Print, tag ? _("Print tagged attachment(s)?") :
        
curs_main.c:1377:  CHECK_ACL(MUTT_ACL_DELETE, _("Cannot delete message(s)"));
curs_main.c:1606:  CHECK_ACL(MUTT_ACL_DELETE, _("Cannot undelete message(s)"));
curs_main.c:2772:  CHECK_ACL(MUTT_ACL_DELETE, _("Cannot delete message(s)"));
curs_main.c:3006:  CHECK_ACL(MUTT_ACL_SEEN, _("Cannot mark message(s) as read"));
curs_main.c:3203:  CHECK_ACL(MUTT_ACL_DELETE, _("Cannot undelete message(s)"));
pager.c:2818:      CHECK_ACL(MUTT_ACL_DELETE, _("Cannot delete message(s)"));
pager.c:3137:      CHECK_ACL(MUTT_ACL_DELETE, _("Cannot undelete message(s)"));

opcodes.h:47:      _fmt(OP_COMPOSE_ATTACH_FILE, N_("attach file(s) to this message")) \
opcodes.h:48:      _fmt(OP_COMPOSE_ATTACH_MESSAGE, N_("attach message(s) to this message")) \
opcodes.h:49:      _fmt(OP_COMPOSE_ATTACH_NEWS_MESSAGE, N_("attach news article(s) to this message")) \
opcodes.h:255:     _fmt(OP_FORGET_PASSPHRASE, N_("wipe passphrase(s) from memory")) \
```

and also 

`"[-- This %s/%s attachment (size %s bytes) has been deleted --]\n"` (handler.c lines 1611 and 1620)

mgedmin:
Another one:
```
#: mx.c:754
msgid "Move %d read messages to %s?"
```

flatcap:
```c
pop.c:954:
snprintf(msgbuf, sizeof(msgbuf), _("Reading new messages (%d bytes)..."), bytes);
```

reisradomil:
I must have missed these. I'll have a look. If you find more do not hesitate to post them here.

mgedmin:
[...]

PR #1219, opened 2018-05-21
Fix plural messages (#1203)

reisradomil:
Revisit plural messages as noted in issue #1203.

Some messages cannot be fixed, some only if we completely reword them or rewrite the programming logic. For those cases I've added a note for the translator. Others were straight forward.

Here is a quick recap of the details:

```        
curs_main.c:1377:  CHECK_ACL(MUTT_ACL_DELETE, _("Cannot delete message(s)"));
curs_main.c:1606:  CHECK_ACL(MUTT_ACL_DELETE, _("Cannot undelete message(s)"));
curs_main.c:2772:  CHECK_ACL(MUTT_ACL_DELETE, _("Cannot delete message(s)"));
curs_main.c:3006:  CHECK_ACL(MUTT_ACL_SEEN, _("Cannot mark message(s) as read"));
curs_main.c:3203:  CHECK_ACL(MUTT_ACL_DELETE, _("Cannot undelete message(s)"));
pager.c:2818:      CHECK_ACL(MUTT_ACL_DELETE, _("Cannot delete message(s)"));
pager.c:3137:      CHECK_ACL(MUTT_ACL_DELETE, _("Cannot undelete message(s)"));
```
Cannot be fixed unless we either reword them or change the logic to get a the precise number of messages. Note that these strings are part of the expression "%s: Operation not permitted by ACL" (with %s the above strings). I added a comment for the translator to it.

```
opcodes.h:47:      _fmt(OP_COMPOSE_ATTACH_FILE, N_("attach file(s) to this message")) \
opcodes.h:48:      _fmt(OP_COMPOSE_ATTACH_MESSAGE, N_("attach message(s) to this message")) \
opcodes.h:49:      _fmt(OP_COMPOSE_ATTACH_NEWS_MESSAGE, N_("attach news article(s) to this message")) \
opcodes.h:255:     _fmt(OP_FORGET_PASSPHRASE, N_("wipe passphrase(s) from memory")) \
```
These are part of a global macro and GNU gettext cannot use ngettext inside of it (we use `N_` not `_` here). So unless we rewrite that code or do other fancy stuff we cannot fix it.

```
handler.c:1611: "[-- This %s/%s attachment (size %s bytes) has been deleted --]\n"
handler.c:1620: "[-- This %s/%s attachment (size %s bytes) has been deleted --]\n"
```
We use ngettext here to distinguish wrt the size. However, %s will be replaced by the prettified version of size, e.g. `2K`. Some languages might be sensitive to such suffixes, e.g. `1K bytes` and `1024 bytes` might inflect 'byte' differently. I added a note to make the translators aware of that.

```
mx.c:754: "Move %d read messages to %s?"
pop.c:954: snprintf(msgbuf, sizeof(msgbuf), _("Reading new messages (%d bytes)..."), bytes);
```
Straight forward

```
mx.c:634:          mutt_error(_("message(s) not deleted"));
recvattach.c:841:  if (query_quadoption(Print, tag ? _("Print tagged attachment(s)?") :
```
We count the messages and use `ngettext()` but we don't use the number in the user messages.
We noted that for the translators.

```
ncrypt/crypt.c:95: mutt_message(_("Passphrase(s) forgotten."));
```
We stick with the "generic plural" of "Passphrase(s)". There is currently no way to get the number of passphrases due to implementation details, e.g. sometimes gpg-agent handles the passphrases.
We added a comment so that the translators are aware of it.

### DIFF
--- a/curs_main.c
+++ b/curs_main.c
@@ -1374,7 +1374,11 @@ int mutt_index_menu(void)
         CHECK_VISIBLE;
         CHECK_READONLY;
         /* L10N: CHECK_ACL */
-        CHECK_ACL(MUTT_ACL_DELETE, _("Cannot delete message(s)"));
+        /* L10N: Due to the implementation details we do not know whether we
+           delete zero, 1, 12, ... messages. So in English we use
+           "messages". Your language might have other means to express this.
+         */
+        CHECK_ACL(MUTT_ACL_DELETE, _("Cannot delete messages"));
 
         CHECK_ATTACH;
         mutt_pattern_func(MUTT_DELETE, _("Delete messages matching: "));
@@ -1603,7 +1607,11 @@ int mutt_index_menu(void)
         CHECK_VISIBLE;
         CHECK_READONLY;
         /* L10N: CHECK_ACL */
-        CHECK_ACL(MUTT_ACL_DELETE, _("Cannot undelete message(s)"));
+        /* L10N: Due to the implementation details we do not know whether we
+           undelete zero, 1, 12, ... messages. So in English we use
+           "messages". Your language might have other means to express this.
+         */
+        CHECK_ACL(MUTT_ACL_DELETE, _("Cannot undelete messages"));
 
         if (mutt_pattern_func(MUTT_UNDELETE,
                               _("Undelete messages matching: ")) == 0)
@@ -2769,7 +2777,11 @@ int mutt_index_menu(void)
         CHECK_VISIBLE;
         CHECK_READONLY;
         /* L10N: CHECK_ACL */
-        CHECK_ACL(MUTT_ACL_DELETE, _("Cannot delete message(s)"));
+        /* L10N: Due to the implementation details we do not know whether we
+           delete zero, 1, 12, ... messages. So in English we use
+           "messages". Your language might have other means to express this.
+         */
+        CHECK_ACL(MUTT_ACL_DELETE, _("Cannot delete messages"));
 
         {
           int subthread = (op == OP_DELETE_SUBTHREAD);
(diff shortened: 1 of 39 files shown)


flatcap:
Very nice.  Thanks.

I propose a rule to distinguish "types" of plurals.

**Positive events** (About to do / Have just done something)

- Show the number, if at all possible, e.g.
  "4 messages deleted"
- It's reassuring to get the exact feedback you expected.

**Negative events** (Something has **not** happened)
- Don't show the number, it's not important, e.g.
  "Messages not deleted"

**Future unspecified** (For menus and help)
- In English, it's common to use the plural in this situation, e.g.
  "Delete messages"
  The implication is that the user might perform:
  - One action on multiple objects
  - Repeated actions on a single object
- Using a suffix of "(s)" is common, but ugly.
  Let's avoid them.

---

> `curs_main.c:1377: CHECK_ACL(MUTT_ACL_DELETE, _("Cannot delete message(s)"));`

I'd like to eliminate these ugly macros, but for now,
I suggest replacing: "message(s)" with "messages", etc.

> `opcodes.h:47: _fmt(OP_COMPOSE_ATTACH_FILE, N_("attach file(s) to this message"))`

I suggest we turn "file(s)" into "files" for each occurrence.

---

- 39fe98413 ncrypt/ncrypt.c: Comment on plural message
    - `mutt_message(_("Passphrase(s) forgotten."));`
      Suggest: "Passphrases forgotten."

- 9eb074fd0 recvattach.c: Fix plural message "Print tagged
    - `"Print tagged attachments?", tagmsgcount)`
      Suggest: "Print %d tagged attachments?"

- 9e13511dc pager.c: Comment on plural messages with "message(s)"
    - `CHECK_ACL(MUTT_ACL_DELETE, _("Cannot delete message(s)"));`
      Suggest: "Cannot delete messages"

- fed719d36 curs_main.c: Comment on plural messages with
    - Same


/cc @neomutt/translators

reisradomil:
I've updated the pull request accordingly.

flatcap:
Thanks, merged into [translate].

PR #1227, opened 2018-05-26
Update zh_CN translation

l2dy:
* **What does this PR do?**
@yuzibo "个" is needed here.

### DIFF
--- a/po/zh_CN.po
+++ b/po/zh_CN.po
@@ -5918,10 +5918,10 @@ msgstr "我不知道要如何打印 %s 类型的附件！"
 #. do not show it to the user.  So feel free to use a "generic
 #. plural" as plural translation if your language has one.
 #: recvattach.c:852
-#,  c-format
+#, c-format
 msgid "Print tagged attachment?"
 msgid_plural "Print %d tagged attachments?"
-msgstr[0] "打印已标记的 %d 附件？"
+msgstr[0] "打印已标记的 %d 个附件？"
 
 #: recvattach.c:853
 #, c-format


yuzibo:
Sorry for later reply.
Maybe you have known my worry about the translate.　
I just  am not clear the plural translate for here.
Do you think that there is never path to translate  "Print tagged attachment?"
If so, i am sure that the translate is no problem.
Please correct me.

l2dy:
With `Plural-Forms: nplurals=1; plural=0;`, I don't think there's a way to omit the " 1 " without changing the code. See https://www.gnu.org/software/gettext/manual/html_node/Plural-forms.html for details.

```
snprintf(prompt, sizeof(prompt),
         /* L10N: Although we now the precise number of tagged messages, we
            do not show it to the user.  So feel free to use a "generic
            plural" as plural translation if your language has one. */
         tag ? ngettext("Print tagged attachment?", "Print %d tagged attachments?", tagmsgcount) :
               _("Print attachment?"),
         tagmsgcount);
```

Here the `ngettext` function returns `msgstr[0]` for zh_CN and snprintf takes that as the format string. snprintf discards excess arguments, but in our case we can only have one format string so we can't have one with `%d` and another without `%d`.

yuzibo:
Ok,
Only the program run here,it tagged more than zero attachment that need %d + quantifier
You are right.
Could you have glance at line 1745&2628.

l2dy:
@yuzibo Line 2628 already has "个" in the translation; for line 1745 see #1230.

yuzibo:
2628..
I mean %2$s and %1$d. The '$' i don know it from where :)
BTW,if "在％S中..."　would be better?

l2dy:
Right, see https://www.gnu.org/software/gettext/manual/html_node/c_002dformat.html, it's for reordering arguments.

No, "在" is unnecessary here.

yuzibo:
Thanks!

Issue #1149, opened 2018-04-20
Split up mutt_resize_screen()


flatcap:
This is an "easy" task ([find more](https://github.com/neomutt/neomutt/issues?q=label:"good%20first%20issue")) and it's open to everyone.  [_read more..._](https://neomutt.org/dev/easy-tasks)

## Description

This function has both ncurses and slang code in it.  
As only one can be used at a time, this would be better as two functions surrounded by `#ifdef`.

### Part 1: Create ncurses version

Also, rename all the strange variables.

### Part 2: Create slang version

The variables here are globals from slang, so they need to keep their names.

parazyd:
Do you really think it helps? Splitting up this function duplicates most of the code.

flatcap:
Yes.  The two versions are mutually exclusive and behave differently.
The slang version uses **global variables** from slang,
and the curses version uses **local variables** with slang names.

flatcap:
@parazyd 
Please can you create a pull-request with you changes, so far.
Thanks

parazyd:
I haven't had time to test it yet. But I hope to soon.

flatcap:
Ping?

flatcap:
Please can we see some code?

parazyd:
I'll rebase and open a pull request tomorrow :)

PR #1233, opened 2018-05-28
Refactor resize.c into separate functions.

parazyd:
* **What does this PR do?**

Splits up mutt_resize_screen() into its separate function to improve readability.

https://github.com/neomutt/neomutt/issues/1149

### DIFF
--- a/resize.c
+++ b/resize.c
@@ -4,6 +4,7 @@
  *
  * @authors
  * Copyright (C) 1996-2000 Michael R. Elkins <me@mutt.org>
+ * Copyright (C) 2018      Ivan J. <parazyd@dyne.org>
  *
  * @copyright
  * This program is free software: you can redistribute it and/or modify it under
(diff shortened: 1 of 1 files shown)


resize.c
```diff
@@ -42,50 +43,95 @@
 #endif
 #endif
 
-/**
- * mutt_resize_screen - Called after receiving SIGWINCH
- */
-void mutt_resize_screen(void)
+int mutt_resize_getrows(int rows)
```
flatcap: We don't need the `rows` parameter.
The old (invalid) value we're passed is just overwritten anyway.

resize.c
```diff
@@ -42,50 +43,95 @@
...
 {
   const char *cp = NULL;
-  int fd;
-  struct winsize w;
-#if !defined(USE_SLANG_CURSES)
-  int SLtt_Screen_Rows, SLtt_Screen_Cols;
-#endif
+  cp = mutt_str_getenv("LINES");
```
flatcap: These two lines can be combined

resize.c
```diff
@@ -42,50 +43,95 @@
...
+  if (cp && (mutt_str_atoi(cp, &rows) < 0))
+    return 24;
+  return -1;
```
flatcap: The default of `-1` **is** the original behaviour,
but I think returning a sensible value (24) is a better idea.

resize.c
```diff
@@ -42,50 +43,95 @@
...
+int mutt_resize_getrows(int rows)
 {
   const char *cp = NULL;
-  int fd;
-  struct winsize w;
-#if !defined(USE_SLANG_CURSES)
-  int SLtt_Screen_Rows, SLtt_Screen_Cols;
-#endif
+  cp = mutt_str_getenv("LINES");
+  if (cp && (mutt_str_atoi(cp, &rows) < 0))
```
flatcap: `atoi` saves the value into `rows` (a local copy of a variable), but it's not returned to the caller.

resize.c
```diff
@@ -42,50 +43,95 @@
...
+    return 24;
+  return -1;
+}
 
-  SLtt_Screen_Rows = -1;
-  SLtt_Screen_Cols = -1;
+int mutt_resize_getcols(int cols)
```
flatcap: Same comments as for `getrows()`

resize.c
```diff
@@ -42,50 +43,95 @@
...
+{
+  const char *cp = NULL;
+  cp = mutt_str_getenv("COLUMNS");
+  if (cp && (mutt_str_atoi(cp, &cols) < 0))
+    return 80;
+  return -1;
+}
+
+int mutt_resize_getwinsize(winsize w)
```
flatcap: You're missing a `struct`

resize.c
```diff
@@ -42,50 +43,95 @@
...
+{
+  int fd;
```
flatcap: We can combine this with the assign, below

resize.c
flatcap: This needs to be a pointer, otherwise we're working with a local copy of the struct

resize.c
```diff
@@ -42,50 +43,95 @@
...
-    cp = mutt_str_getenv("LINES");
-    if (cp && (mutt_str_atoi(cp, &SLtt_Screen_Rows) < 0))
-      SLtt_Screen_Rows = 24;
+  return ret;
+}
+
+#if defined(USE_SLANG_CURSES)
+void mutt_resize_screen_slang(void)
+{
+  struct winsize w;
```
flatcap: Making this `w = { 0 }` will mean the members are initialised

resize.c
```diff
@@ -42,50 +43,95 @@
...
   stdscr = newwin(0, 0, 0, 0);
   keypad(stdscr, true);
+  mutt_window_reflow();
+}
 #else
-  resizeterm(SLtt_Screen_Rows, SLtt_Screen_Cols);
-#endif
+void mutt_resize_screen_ncurses(void)
+{
+  struct winsize w;
```
flatcap: `w = { 0 }`

resize.c
```diff
@@ -42,50 +43,95 @@
...
+
+  resizeterm(screenrows, screencols);
   mutt_window_reflow();
 }
+#endif
+
+/**
+ * mutt_resize_screen - Called after receiving SIGWINCH
+ */
+void mutt_resize_screen(void)
```
flatcap: I don't think we really need this extra level of abstraction.
We can simply rename the two functions to `mutt_resize_screen()` and `#ifdef` them.

parazyd:
I've updated my branch with a little different way of patching it. I think this is better.

flatcap:
I've just made a small tweak (which saves a few lines).
Unless you've anything to add, I think we've spent more than enough time on this :-)

parazyd:
Looks alright. Sorry for all the confusion :D

Issue #1236, opened 2018-05-30
IMAP auth_plain w/o initial response hangs on BAD response


gahr:
# Expected behaviour

- neomutt attempts `LOGIN ...` after `BAD` response to `AUTHENTICATE PLAIN`

# Actual behaviour

- login process hangs

# What might be happening?

From imap/auth_plain.c :

```c
    do
      rc = imap_cmd_step(idata);
    while (rc == IMAP_CMD_CONTINUE);
```

neomutt is expecting a `+ ...` response from the server (prompting command continuation from the client) perhaps?

(Thanks @edwinbalani for the report)

PR #1237, opened 2018-05-30
Handle a BAD response in AUTH PLAIN w/o initial response

gahr:
The incorrect assumption was that the first line of an AUTH PLAIN w/o
I-R was always successful, so the client could write the second line and
end up with a readable socket that would be activated by the server
sending the final OK or BAD response.
If however the first line results in a BAD response already, the server
wouldn't write anything else and the client would be left polling for an
additional response that never arrives, effectively hanging the process.

Issue #1236

* **What does this PR do?**

     for that)

### DIFF
--- a/imap/auth_plain.c
+++ b/imap/auth_plain.c
@@ -45,7 +45,7 @@
  */
 enum ImapAuthRes imap_auth_plain(struct ImapData *idata, const char *method)
 {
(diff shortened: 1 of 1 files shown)


gahr:
OP confirmed that this solves his issue.

PR #1241, opened 2018-05-31
notmuch: use global settings for vfolder-from-query

jpoimboe:
When the <vfolder-from-query> command is run while already in a vfolder,
it takes the current vfolder as context.  That means the global settings
(nm_query_type, nm_db_limit, nm_default_uri) are ignored in favor of the
current vfolder's settings.  This is confusing behavior because the new
vfolder is otherwise independent from the old one.

Set the nm_uri_from_query() context to NULL so that it uses the global
settings instead.

* **What does this PR do?**

     for that)

### DIFF
--- a/curs_main.c
+++ b/curs_main.c
@@ -2068,7 +2068,7 @@ int mutt_index_menu(void)
           mutt_message(_("No query, aborting"));
           break;
         }
-        if (!nm_uri_from_query(Context->mailbox, buf, sizeof(buf)))
+        if (!nm_uri_from_query(NULL, buf, sizeof(buf)))
           mutt_message(_("Failed to create query, aborting"));
         else
           main_change_folder(menu, op, buf, sizeof(buf), &oldcount, &index_hint);
--- a/notmuch/mutt_notmuch.c
+++ b/notmuch/mutt_notmuch.c
@@ -201,6 +201,22 @@ static void free_mboxdata(void **data)
 }
 
 /**
+ * string_to_query_type - Lookup a query type
+ * @param str String to lookup
+ * @retval num Query type, e.g. #NM_QUERY_TYPE_MESGS
+ */
+static enum NmQueryType string_to_query_type(const char *str)
+{
+  if (mutt_str_strcmp(str, "threads") == 0)
+    return NM_QUERY_TYPE_THREADS;
+  else if (mutt_str_strcmp(str, "messages") == 0)
+    return NM_QUERY_TYPE_MESGS;
+
+  mutt_error(_("failed to parse notmuch query type: %s"), NONULL(str));
+  return NM_QUERY_TYPE_MESGS;
+}
+
+/**
  * new_mboxdata - Create a new NmMboxData object from a query
  * @param uri Notmuch query string
  * @retval ptr New NmMboxData struct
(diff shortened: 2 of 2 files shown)


flatcap:
hi @jpoimboe 
Thanks for the PR.

Austin-Ray:
@flatcap I applied the proposed change and nothing appeared to break, but I'll continue testing.

I agree with @jpoimboe that the current behavior is confusing. The current behavior makes `vfolder-from-query` unpredictable.

flatcap:
Thanks @Austin-Ray 

> The current behavior makes vfolder-from-query unpredictable.

Suggestions for fixes, or improved commands, are welcome.

Austin-Ray:
@flatcap: I believe @jpoimboe's proposed behavior is the way to go. 

Here is are two examples of different behavior based on folder:

1. Suppose I have two identical `vfolders`, but one has a limit of 100 emails. If I execute an arbitrary `vfolder-from-query` from within the limited `vfolder`, the query will return 100 emails that match. The non-limited `vfolder` will return all emails that match.

2. Suppose you had two `notmuch` databases configured and two `vfolder` setup corresponding to the different databases. If you executed `vfolder-from-query` from one `vfolder`, it will only search the corresponding database. 

Example two reveals a gap in our `notmuch` query syntax in relation to `vfolder-from-query`: there's no way to change `notmuch` databases in `vfolder-from-query` without changing `$nm_default_uri` or `$folder` [0][1].

### Issues with PR

There are a few issues with passing `NULL` for the `Context`

1. **Only tags can be specified**
 Since the `Context` is `NULL` so will `NmCtxData`, which means the below block of code cannot be executed from `vfolder-from-query`.
https://github.com/neomutt/neomutt/blob/1eaedd6f10a978dd55649dd05e48de1bc163d210/notmuch/mutt_notmuch.c#L1904-L1918
 As such, there is no valid object to store parsed `type=` statements.

2. You can only search one database, which must be set by either `$nm_default_uri` or `$folder`[1]. Changing the database is possible, but requires using `set ...` beforehand.

### Potential resolutions

1. On `Context == NULL`, create an empty `NmCtxData` to stored parsed values.

However, the issue regarding different `notmuch` databases exists. If we pass `NULL`, we're stuck to one database unless we execute more commands. If we pass `Context` then we can search different databases (based on `vfolder`), but will be stuck with any limitations of the current `vfolder`.

I hope you found this analysis to be satisfactory.

### Citations
[0] https://neomutt.org/feature/notmuch#3-2-%C2%A0items
[1] https://neomutt.org/feature/notmuch#3-1-%C2%A0folders-uri

Austin-Ray:
@flatcap I have rebased @jpoimboe's changes onto master and built on them.

I found a way to reduce `nm_uri_from_query(...)`'s `if` branch by implementing logic to build a `NmMboxData` with default values when no existing `NmMboxData` exists within the passed `Mailbox`. `nm_get_default_data(...)` still ensures that either `nm_default_uri` or `folder` is set so we won't attempt to build an impossible URI.

flatcap:
I've done a little tidying:
- Fix doxygen headers
  - function_name - Description
  - Drop `.`s from the end of `@param` lines (it confuses doxygen)
- `string_to_query_type()`
  - make it static (it's not used elsewhere)
  - move the function to avoid the need for a prototype

If you're happy, I'll merge it.  Thanks.

Austin-Ray:
@flatcap It looks good to me!

PR #1245, opened 2018-06-01
Multi-Column Pager

flatcap:
**Config options**:
- INT: pager_multi_width
- BOOL: pager_multi_wrap
  If "markers" is set, perhaps display a ">" at the end of truncated lines (like vim)

**Behaviour**:
- Scroll up/down one line
- Scroll up/down one page
  Should this move one page (shift columns right/left), or move down 'n' pages?
- Home
- End

**Graphics**:
- Divider chars -- use logic from sidebar (ASCII vs Unicode)
- Colour: use Sidebar divider colour?  Or a new colour?

**Quirks**::
- Sometime the headers aren't all displayed
- The headers seem to affect the layout when moving up/down one line

**Tests**:
- Toggle sidebar
- Resize window vertically
- Resize window horizontally
- Toggle wrap setting
- Change wrap width

Currently this feature isn't well isolated from the main pager code.
I suggest adding a `./configure --experimental` option.

### DIFF
--- a/pager.c
+++ b/pager.c
@@ -1221,7 +1221,8 @@ static int format_line(struct Line **line_info, int n, unsigned char *buf, int f
   int wrap_cols =
       mutt_window_wrap_cols(pager_window, (flags & MUTT_PAGER_NOWRAP) ? 0 : Wrap);
 
-  mutt_debug(2, "pager: fmt wrap? %d Wrap %d wc %d\n", (flags & MUTT_PAGER_NOWRAP), Wrap, wrap_cols);
+  mutt_debug(2, "pager: fmt wrap? %d Wrap %d wc %d\n",
+             (flags & MUTT_PAGER_NOWRAP), Wrap, wrap_cols);
 
   if (check_attachment_marker((char *) buf) == 0)
     wrap_cols = pager_window->cols;
@@ -1965,29 +1966,26 @@ static void pager_menu_redraw(struct Menu *pager_menu)
       int col_max = 2;
 
       /* recompute in place for WINCH */
-      if ((COLS
 #ifdef USE_SIDEBAR
-                               - SidebarWidth
+      if ((COLS - SidebarWidth) > col_min_width)
+        col_max = (COLS - SidebarWidth) / col_min_width;
+#else
+      if (COLS > col_min_width)
+        col_max = COLS / col_min_width;
 #endif
-                               ) > col_min_width)
-        col_max = (COLS
-#ifdef USE_SIDEBAR
-                         - SidebarWidth
-#endif
-                         ) / col_min_width;
 
 #ifdef USE_SIDEBAR
-       /* For header */
+      /* For header */
       if (SidebarVisible)
         Wrap -= SidebarWidth;
 #endif
 
       bodylen2 = col_max * bodylen;
-      col_shift = (COLS
 #ifdef USE_SIDEBAR
-                       - SidebarWidth
+      col_shift = (COLS - SidebarWidth) / col_max;
+#else
+      col_shift = COLS / col_max;
 #endif
-                       ) / col_max;
 
       mutt_window_move(rd->pager_window, 0, 0);
       rd->curline = rd->oldtopline = rd->topline;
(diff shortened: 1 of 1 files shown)


kdave:
Regarding the logic to decide how many columns to show, there are 2 main approaches:

* set max multicol width and show as many columns as the screen allows (currently implemented)
* set max column number and use the confugred line wrap width
* maybe combination of both

I personally don't have use for the text wrapping, as I want to see the whole
mail and would have to do something else to also display the missing text.

The scrolling by line or more lines depends on how the headers are displayed:

* full width (implemented)
* inside the column, wrapping etc

The first option screws up scrolling as column N will jump by N lines, but I
find it bearable.

Page scrolling looks better when it shifts by all columns, the multicolumn
setup is really great for a quick skim of the mail but that obviously depends
on the mail content. It does work great for patches.

'End' behaviour is now maybe inconsistent and just shifts by the N * ROWS
lines, but ideally it could show the last screen of mail (ie. all columns
filled).

I use divider different from sidebar, it IMO looks better if there's no
background color, just a different char color.

Maybe a few more quirks or things to check:

* help screen is also multicolumn but it totally screws up the output and it's
  unreadable, this might be different with the windowing rework. also who needs
  help ...

* line colors could get out of sync if the line is wrapped inside the column
  (again observed on the ancient version)

flatcap:
Hi @kdave, (sorry for the slow reply, I've been hiding this week)

> * set max multicol width and show as many columns as the screen allows (currently implemented)
> * set max column number and use the confugred line wrap width
> * maybe combination of both

Hmm... I hadn't considered the `wrap` option.
Generally, we want to add as few config options as we can (we've over 400 already :-)

Perhaps:
- `set wrap=80 multicol=yes`
- `set wrap=0  multicol=no`

(name to be decided)

> ... text wrapping
> I want to see the whole mail

Yeah, ignore my suggestion.
Altering the pager this much would be risky.

> The scrolling by line or more lines depends on how the headers are displayed:
> * full width (implemented)

I think that's the only sensible solution

> ... screws up scrolling as column N will jump by N lines

So the calculations need to ignore the header lines
/me feels the urge to start creating diagrams.

> Page scrolling looks better when it shifts by all columns

So after `<page-down>`: '1|2|3' becomes '2|3|4'

OK, good

(as opposed to '1|2|3' becomes '4|5|6')

> 'End' behaviour ... ideally ... show the last screen of mail
> (ie. all columns filled).

Yes, that sounds nice.

> I use divider different from sidebar,

OK.  So we'll need a `color` name for it.

> it IMO looks better if there's no background color, just a different char color.

Yes, but that'll depend on people's colour scheme.

> Maybe a few more quirks or things to check:

> * help screen ... unreadable

Yes, multicol needs to be disabled for help (and our upcoming summary) pages.

> * line colors could get out of sync if the line is wrapped inside the column
>   (again observed on the ancient version)

Hmm... we'll worry about that when we find an example.
The pager's colouring of wrapped lines isn't that great anyway.

kdave:
On Sat, Jun 09, 2018 at 04:43:06AM -0700, Richard Russon wrote:
> Perhaps:
> - `set wrap=80 multicol=yes`
> - `set wrap=0  multicol=no`

Does that mean I have to set wrap in order to get wrapping for multicol?

> > The scrolling by line or more lines depends on how the headers are displayed:
> > * full width (implemented)
> 
> I think that's the only sensible solution
> 
> > ... screws up scrolling as column N will jump by N lines
> 
> So the calculations need to ignore the header lines
> /me feels the urge to start creating diagrams.

When the headers are visible and scrolled up by one line, this line shifts up the leftmost column, that takes the 1st line of the 2nd column.

The 2nd column has now 1 line less and also shifts up by the line made available by the headers. And now has 2 lines empty at the bottom that it takes from the 3rd column.

The rest follows by induction. It's not that complicated :)

> > Page scrolling looks better when it shifts by all columns
> 
> So after `<page-down>`: '1|2|3' becomes '2|3|4'
> 
> OK, good
> 
> (as opposed to '1|2|3' becomes '4|5|6')

My idea is rather 123 -> 456. From the usability POV it takes fewer PgDn's to go through the mail.

flatcap:
Though this is a interesting idea, it really needs someone to drive it forward.
For now, I'm closing this PR and I have archived the branch to the [neomutt-old](https://github.com/neomutt/neomutt-old) repo.

PR #1246, opened 2018-06-02
Update ja translation

l2dy:
* **What does this PR do?**
Adapted from translation of "Mail not sent." and "Mail sent.".


l2dy:
@flatcap Do we have an active Japanese translator?

flatcap:
No.  Nobody's admitted to speaking Japanese, so you may as well merge it.

PR #1247, opened 2018-06-04
Remove pgpring.1 from doc/Makefile.autosetup

Austin-Ray:
Since commit e132075a6f9694878e644aa6683665490f2746d5 removes `pgpring.1`,
the reference to `pgpring.1` must be removed from `doc/Makefile.autosetup`
otherwise `make install` will throw an error, which can prevent installation.

The error thrown by `make install` prevented installation on two devices I use regularly.


gahr:
Thanks, I pushed that in a different commit to also remove pgpring from uninstall.

Issue #1248, opened 2018-06-06
During IMAP reconnection neomutt sends LIST but not SELECT


EdwardBetts:
If the connection to the IMAP server is lost neomutt will reconnect, but the reconnection is incomplete. When first opening a connection mutt sends the LIST and SELECT commands, during the reconnection it only sends the LIST command. 

After a reconnect the neomutt status bar says `---NeoMutt: (no mailbox) [Msgs:0]`. This happens every time.

Expected behaviour is for neomutt to reconnect and display the current mailbox.

This is the muttrc config file I used for testing.
```rc
set folder=imap://username:password@localhost/

set mail_check=2
set timeout=2
unset help
unset ssl_starttls
```

To reproduce run test_imap.py (included below), then use neomutt to connect with these parameters:
```shell
neomutt -F muttrc -f imap://user:password@localhost:1143/
```

Neomutt will connect to the IMAP server and show one message, after 3 seconds the test server will drop the connection, neomutt will attempt to reconnect but the reconnection is incomplete.

The cmd_handle_fatal function is responsible for handling the disconnection, it displays the "Mailbox closed" error, then calls imap_conn_find to reconnect.

``` c
static void cmd_handle_fatal(struct ImapData *idata)
{
  idata->status = IMAP_FATAL;

  if ((idata->state >= IMAP_SELECTED) && (idata->reopen & IMAP_REOPEN_ALLOW))
  {
    mx_fastclose_mailbox(idata->ctx);
    mutt_socket_close(idata->conn);
    mutt_error(_("Mailbox %s@%s closed"), idata->conn->account.login,
               idata->conn->account.host);
    idata->state = IMAP_DISCONNECTED;
  }

  imap_close_connection(idata);
  if (!idata->recovering)
  {
    idata->recovering = true;
    if (imap_conn_find(&idata->conn->account, 0))
      mutt_clear_error();
    idata->recovering = false;
  }
}
```

It isn't clear who is responsible for sending the SELECT command to the IMAP server, maybe it should be in imap_conn_find or cmd_handle_fatal. The following is the relevant part of imap_conn_find. 
 
``` c
  if (new && idata->state == IMAP_AUTHENTICATED)
  {
    /* capabilities may have changed */
    imap_exec(idata, "CAPABILITY", IMAP_CMD_QUEUE);
    /* enable RFC6855, if the server supports that */
    if (mutt_bit_isset(idata->capabilities, ENABLE))
[...]

PR #1249, opened 2018-06-06
WIP: Send SELECT to IMAP server after reconnect

EdwardBetts:
This is an attempt to do something about bug #1248

New imap_select_mailbox function split out of imap_open_mailbox to be called during reconnect.

There is a bug in this implementation, it crashes during the reconnection stage with this error:

terminated by signal SIGSEGV (Address boundary error)

Somebody with a better grasp of C might figure out why it is failing.

### DIFF
--- a/imap/command.c
+++ b/imap/command.c
@@ -164,8 +164,10 @@ static void cmd_handle_fatal(struct ImapData *idata)
   if (!idata->recovering)
   {
     idata->recovering = true;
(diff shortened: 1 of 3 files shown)


EdwardBetts:
Needs more work. Closing for now.

gahr:
> terminated by signal SIGSEGV (Address boundary error)

`ctx` is NULL after `mx_fastclose_mailbox(idata->ctx);`. That might be the cause.

PR #1254, opened 2018-06-10
fix trash mailbox

### LINKED ISSUE #1253: trash folder doesn't work
## Expected Behaviour
When `trash` is set, mail is moved to trash folder after marked for deletion.

## Actual Behaviour
When `trash` is set, mail is directly deleted.

## Steps to Reproduce
1. Delete a mail
2. Open the trash folder

- Always

- When I upgraded
  Which version did you use to use?
20180512

## NeoMutt Version
1c3dbec59739c1db601d667920a14a3b082e6303

## Extra Info

```
ProductName:	Mac OS X
ProductVersion:	10.13.5
BuildVersion:	17F77
```
No.
No.
* Is your email local (maildir) or remote (IMAP)?
Local.

flatcap:
The second `for` loop was using the value of `i` as a starting place.
By changing the first loop, the second loop wasn't doing any work.

Fixes #1253

### DIFF
--- a/mx.c
+++ b/mx.c
@@ -617,10 +617,18 @@ static int trash_append(struct Context *ctx)
     return 0;
 
   int delmsgcount = 0;
+  int first_del = -1;
   for (i = 0; i < ctx->msgcount; i++)
+  {
     if (ctx->hdrs[i]->deleted && (!ctx->hdrs[i]->purge))
+    {
+      if (first_del < 0)
+        first_del = i;
       delmsgcount++;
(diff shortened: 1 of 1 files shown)


flatcap:
> save the list of indexes somewhere during the first iteration

That should be fine as long as we're not interrupted.

> iterate twice, as it's not going to be noticeable anyway

Yeah, even with tens of thousands of emails, the loop will take less time than the file operations.

PR #1256, opened 2018-06-13
fcc_attach: Don't ask every time

BertramScharpf:
Option fcc_attach used to ask in any case, if there were attachment nad if there weren't.
In case there is no attachment, there is no need to ask.

I'm not very sure whether the actual behaviour with signatures is right: If you don't fcc the attachments, two signatures will be calculated: one for the bare mail and one for the whole multipart bundle.

### DIFF
--- a/send.c
+++ b/send.c
@@ -2052,8 +2052,8 @@ int ci_send_message(int flags, struct Header *msg, char *tempfile,
       msg->content = clear_content;
 
     /* check to see if the user wants copies of all attachments */
-    if (query_quadoption(FccAttach, _("Save attachments in Fcc?")) != MUTT_YES &&
-        msg->content->type == TYPEMULTIPART)
+    if (msg->content->type == TYPEMULTIPART &&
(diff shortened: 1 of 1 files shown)


flatcap:
Hi @BertramScharpf, welcome to NeoMutt.
Thanks for the PR.  I'll have a look and a think about it tomorrow.

BertramScharpf:
> >  two signatures will be calculated: one for the bare mail and one for the whole multipart bundle

> I'm not sure I understand. Can you give an example?

Here is what I do:

  - start my patched version
  - set fcc_attach=ask-no
  - create a message with an attachment and with myself in 'To:'
  - choose 'sign' from the pgp menu
  - send the message
  - enter my pgp-passphrase
  - answer 'no' to the fcc-attach-question

The folder now contains two messages: one without the attachment and one with.
They both have different signatures. The former is a simple multipart: [mail text, signature]. The latter is a nested multipart [ [ mail text, attachment], signature].

PR #1259, opened 2018-06-16
Fixes in style

BertramScharpf:
This branch doesn't change the behaviour of one single function.
I just detected some problems in style like missing "extern", or trailing spaces.

### DIFF
--- a/COPYRIGHT.md
+++ b/COPYRIGHT.md
@@ -8,12 +8,12 @@ GNU General Public License.  See [LICENSE.md](LICENSE.md).
 > the terms of the GNU General Public License as published by the Free Software
 > Foundation, either version 2 of the License, or (at your option) any later
 > version.
-> 
+>
 > This program is distributed in the hope that it will be useful, but WITHOUT
 > ANY WARRANTY; without even the implied warranty of MERCHANTABILITY or FITNESS
 > FOR A PARTICULAR PURPOSE.  See the GNU General Public License for more
 > details.
-> 
+>
 > You should have received a copy of the GNU General Public License along with
 > this program.  If not, see http://www.gnu.org/licenses/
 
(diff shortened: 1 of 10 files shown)


flatcap:
Hi @BertramScharpf.

I try to keep the code tidy as we go along (I like tidying), but mistakes sneak
in.  To get around this I run `clang-format` on all the source just before
every release.

A mixed bag here, some good bits, some slightly problematic.

- a58676e38 Separate line for ";"

  Ah, that pesky semicolon.  I'd noticed that once before.
  Just moving it won't work, because `clang-format` will put it back.
  Please can you insert `{}`s as you suggested.

- 5be9a573a Trailing spaces

  The changes to the docs are good, but `queue.h` and `acutest.h` are 3rd party source.
  As such, we (the devs) decided that we'd keep that exactly as they were (for ease of merging updates).
  Yesterday, I created a PR (merged) to remove the whitespace from upstream's `acutest.h`.

- 220d6a724 Insert extern

  Good

- ed5c7424c One definition is enough

  Ah!  All the globals **used** to live in `globals.h`, but I'm starting to break it up.
  So, your changes to `HiddenTags` and `TagTransforms` are in the wrong direction.
  Please can you remove the `globals.h` versions and put them back in `tags.c`.

- 91bcbc335 Internal linkage was not intended

  Good

- 188005552 Removed static qualifier from extern functions

  I don't think we want this.
  All the header cache functions are hidden behind the [hcache API](https://neomutt.org/code/hcache.html).
  Each of `bdb.c`, `lmdb.c`, etc exposes a [`struct HcacheOps`](https://neomutt.org/code/structHcacheOps.html).

If your 'git-fu' is strong, you can update the commits, rebase and git force-push to update the PR.
(Or you can add more new commits to the PR).

Thanks.

BertramScharpf:
> - 1880055 Removed static qualifier from extern functions
> I don't think we want this.
> All the header cache functions are hidden behind the hcache API.
> Each of bdb.c, lmdb.c, etc exposes a struct HcacheOps.

Oops. I confused *_open and *_ops. The one is static and the other extern. I withdraw.

PR #1260, opened 2018-06-17
Fixes in style 

BertramScharpf:
Arrgh! I deleted the old branch, and the pull request was closed automatically.

This is the continuation of #1259.

### DIFF
--- a/bcache.c
+++ b/bcache.c
@@ -58,7 +58,7 @@ struct BodyCache
 static int bcache_path(struct Account *account, const char *mailbox, char *dst, size_t dstlen)
 {
   char host[STRING];
-  struct Url url = { 0 };
+  struct Url url = { U_UNKNOWN };
   int len;
 
   if (!account || !MessageCachedir || !*MessageCachedir || !dst || (dstlen == 0))
--- a/mutt/address.c
+++ b/mutt/address.c
@@ -989,8 +989,6 @@ const char *mutt_addr_for_display(struct Address *a)
   static char *buf = NULL;
   char *local_mailbox = NULL;
 
-  FREE(&buf);
-
   if (!a->mailbox || mutt_addr_is_local(a))
     return a->mailbox;
 
--- a/mutt_notmuch.c
+++ b/mutt_notmuch.c
@@ -2173,7 +2173,7 @@ int nm_update_filename(struct Context *ctx, const char *old, const char *new,
 int nm_nonctx_get_count(char *path, int *all, int *new)
 {
   struct UrlQueryString *item = NULL;
-  struct Url url = { 0 };
+  struct Url url = { U_UNKNOWN };
   char *url_holder = mutt_str_strdup(path);
   char *db_filename = NULL, *db_query = NULL;
   notmuch_database_t *db = NULL;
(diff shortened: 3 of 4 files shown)


BertramScharpf:
> @BertramScharpf: You could still ceckout 1880055, then git checkout -b a new branch, ...

No, I can't, because I was so angry that I deleted my whole local material, too.

But for the next time, I will remember that.

flatcap:
Having fun? :-)

> I propose to redefining the constants, letting U_UNKNOWN be zero

Yes, that's a good idea.
I've just checked and all URL code is explicitly checking against `U_UNKNOWN`.
Note: there's another `struct Url url = { 0 };` in `mutt_notmuch.c`

BertramScharpf:
> Having fun? :-)

It's rather an urge.

I want to understand the code before I propose my self_encrypt enhancements that will be a deeper change.

flatcap:
> It's rather an urge.

hehe, I know that feeling.
After two years, I've still got a long list of things to fix.

Keep 'em coming :-)

I've merged all the changes except the Url enum commit.
Also, I've taken the liberty of rebasing your branch.
This should realign you:
- `git fetch origin`
- `git reset --hard origin/bsch_nitpick`

> ... my self_encrypt enhancements ...

Sounds interesting.  What do you have in mind?

BertramScharpf:
> > ... my self_encrypt enhancements ...
> Sounds interesting. What do you have in mind?

I wrote a self-encrypt patch in 2009. The Mutt mailing list just told me that this job could be done by configuring GnuPG, and then the whole first-read-this-and-that-documentation stuff that's only intention is to get rid of the troublemaker. I decided not to start a discussion with these people and patched Mutt all the years for myself. Now, doing an upgrade on my machines, I was not too much surprised to see that there is a self-encrypt feature integrated in Mutt since 2017. I once more turned to the Mutt mailing list just to drop some sarcasm. They told me the same rubbish as nine years ago: There is no need for this feature. Although it's already accepted! Then I found NeoMutt.

My self-encrypt patch from 2009 did the following:
  - *_self_encrypt is a quadoption,
  - if *_default_key is not set, the key is guessed from the "From:" header,
  - do nothing at all if the "Fcc:" field is empty.

I applied my patch to NeoMutt last week. In case you're interested, I will make a PR out of it.

flatcap:
> ... Mutt mailing list ...

They wouldn't accept a bucket of water if they were on fire.

> In case you're interested, I will make a PR out of it.

Yes please.

BertramScharpf:
> While U_FILE makes sense for the first member, it doesn't really for any other. I think 0 is more semantically correct in this case.

That's what I meant. I inserted U_FILE because it is identical to 0 according to the enum definition. At least, one should use U_UNKNOWN, and, once when we are on the subject, the enum elements should be reordered so that U_UKNOWN will be 0.

Yet, I could not do that because, in the description of this PR I promised these changes wouldn't change one single byte in the compiled result. If you agree with me, I will be glad to be allowed to make the change.

flatcap:
> I promised these changes wouldn't change one single byte in the compiled result

I release you from your promise :-)

PR #1264, opened 2018-06-20
Improvements to self-encrypt

BertramScharpf:
This patch adds the following features to self-encrypt:

  - *_self_encrypt is a quadoption,
  - if *_default_key is not set, the key is guessed from the "From:" header,
  - do nothing at all if the "Fcc:" field is empty.

The last feature needs an extension of struct Header. In the other case I had to add an fcc parameter to a _lot_ of functions. I put the redesign into a separate branch which I merged before I added the fcc feature.

A subject to discussion could be the name or meaning of the *_default_key variables. The current behaviour is:

  - look at the *_pgp_default variable
  - if empty, look at the From: header
  - if empty as well, ask the user.

Yet, the meaning of the word "default" is what should be done if no value is provided. Therefore, in my opinion, the proceeding should rather be something like:

  - look at the From: header
  - if empty, look at the *_pgp_default variable
  - if empty as well, bother the user with a menu.

### DIFF
--- a/compose.c
+++ b/compose.c
@@ -453,7 +453,7 @@ static void draw_envelope_addr(int line, struct Address *addr)
  * @param msg Header of the message
  * @param fcc Fcc field
  */
-static void draw_envelope(struct Email *msg, char *fcc)
+static void draw_envelope(struct Email *msg)
 {
   draw_envelope_addr(HDR_FROM, msg->env->from);
 #ifdef USE_NNTP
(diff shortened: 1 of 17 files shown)


flatcap:
Rebased to neomutt-20180716

flatcap:
Sorry for not making a decision sooner.
I'm not going to merge these changes.
I'd rather not change the behaviour of the crypto options, for now.

I've archived the code to the [neomutt-old](https://github.com/neomutt/neomutt-old) repo.

PR #1267, opened 2018-06-20
Put bit fields together

BertramScharpf:
I don't know whether you plan to keep the bit fields at all. But if you do, they should be arranged so that they make sense.


mschilli87:
superceeded by https://github.com/neomutt/neomutt/pull/1268(?)

PR #1268, opened 2018-06-20
Put bit fields together

BertramScharpf:
I don't know whether you plan to keep the bit fields at all. But if you do, they should be arranged so that they make sense.

### DIFF
--- a/browser.h
+++ b/browser.h
@@ -54,11 +54,11 @@ struct FolderFile
   bool inferiors : 1;
 #endif
   bool has_buffy : 1;
+  bool local : 1; /**< folder is on local filesystem */
+  bool tagged : 1;
 #ifdef USE_NNTP
   struct NntpData *nd;
 #endif
-  bool local : 1; /**< folder is on local filesystem */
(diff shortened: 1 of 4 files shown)


gahr:
How about turning them all into bools as well?

flatcap:
I hadn't thought about _rearranging_ the structs.
**Note**: Every time we change any of the main structs, we invalidate everyone's header cache.
(but that tends to happen every release).

Yes, these changes look sensible and I like @gahr's suggestion of them being bools.

---

I had considered some renaming, though.

In `struct Context` we have some examples that annoy me:
- `changed`
- `deleted`

Two _past participles_, but reading the code you've no idea what they mean?
- `bool changed` - The **mailbox** has changed in some way
- `int deleted` - The **number** of deleted **emails**

Perhaps we should create a naming rule:
- `bool has_changed`
- `int num_deleted`

(I'm not suggesting that **you** should make these changes -- just ideas)

---

If you want to discuss any ideas, we hang out in `#neomutt` on irc.freenode.net

PR #1272, opened 2018-06-22
Simple replacements from mutt_mktemp() to mutt_file_mkstemp() in ncrypt

mjsir911:
* **What does this PR do?**

Replace `mutt_mktemp()` to `mutt_file_mkstemp()` where it is easily done

#960

### DIFF
--- a/ncrypt/pgp.c
+++ b/ncrypt/pgp.c
@@ -799,8 +799,8 @@ int pgp_class_check_traditional(FILE *fp, struct Body *b, bool just_one)
  */
 int pgp_class_verify_one(struct Body *sigbdy, struct State *s, const char *tempfile)
 {
-  char sigfile[PATH_MAX], pgperrfile[PATH_MAX];
-  FILE *pgpout = NULL, *pgperr = NULL;
+  char sigfile[PATH_MAX];
+  FILE *pgpout = NULL;
   pid_t thepid;
   int badsig = -1;
 
@@ -817,11 +817,10 @@ int pgp_class_verify_one(struct Body *sigbdy, struct State *s, const char *tempf
   mutt_file_copy_bytes(s->fpin, fp, sigbdy->length);
   mutt_file_fclose(&fp);
 
-  mutt_mktemp(pgperrfile, sizeof(pgperrfile));
-  pgperr = mutt_file_fopen(pgperrfile, "w+");
+  FILE *pgperr = mutt_file_mkstemp();
   if (!pgperr)
   {
-    mutt_perror(pgperrfile);
+    mutt_perror("mutt_file_mkstemp() failed!");
     unlink(sigfile);
     return -1;
   }
@@ -853,7 +852,6 @@ int pgp_class_verify_one(struct Body *sigbdy, struct State *s, const char *tempf
   state_attach_puts(_("[-- End of PGP output --]\n\n"), s);
 
   mutt_file_unlink(sigfile);
-  mutt_file_unlink(pgperrfile);
 
   mutt_debug(1, "returning %d.\n", badsig);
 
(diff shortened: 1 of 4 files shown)


mjsir911:
The little testing that I have done:
```
$ grep 'created temp file' ~/.neomuttdebug*
[2018-06-22 13:50:16]<1> pgp_class_encrypt_message() created temp file '/tmp/neomutt-KQ1v5c'
[2018-06-22 13:50:32]<1> pgp_class_encrypted_handler() created temp file '/tmp/neomutt-WBjZvw'
[2018-06-22 13:50:32]<1> pgp_decrypt_part() created temp file '/tmp/neomutt-mxazwv'
```

The only commit I'm really concerned about not working as expected is 3a9d7bd5eea5687dd355062dc0d7d3a49d7afd48, the rest follow pretty much the same pattern

mjsir911:
I have a few questions:

currently the codebase mixes:
```c
FILE *fp;
fp = mutt_file_mkstemp();
```
and
```c
FILE *fp = mutt_file_mkstemp();
```

or even
```c
FILE *fp;
if ((fp = mutt_file_mkstemp()) == NULL) { // what about (!(fp = mutt_file_mkstemp()))?
  // do something
}
```

Which one is preferred and should I fix the changed lines in this PR?

Additionally, I'm writing `mutt_perror("mutt_file_mkstemp() failed!")` a lot, could I move all the cases of that to inside the `mutt_file_mkstemp()` file? Would get rid of a lot of repetition.

flatcap:
Wow!  That's fast work, thanks.
I'll have a look in the morning.

> I'm writing mutt_perror("mutt_file_mkstemp() failed!") a lot

OK, point taken.
I'll have another think about libraries and logging.

flatcap:
> currently the codebase mixes...

I've been working hard to bring some consistency, but the job's hard to automate.
So far, I've been using vim-fu and [coccinelle](https://neomutt.org/dev/analysis#coccinelle).

Too much of the code sacrifices legibility to save a line of text.

> Which one is preferred

C99 allows variables to be declared where they are used.
This means we can reduce the scope of variables, which often makes the code easier to read.
Tools like [cppcheck](https://neomutt.org/dev/analysis#cppcheck) can show where variables can have their scope reduced.

```c
/* Prefer: Variable declared where it is first set */
FILE *fp = fopen(...);
```

> should I fix the changed lines in this PR?

Yes, please (optional, but appreciated :-)
If you're fixing a specific bug, then it might be better to put the tidying in a separate commit.
In this PR, the same commit will be fine.

> `if ((fp = mutt_file_mkstemp()) == NULL)`

I dislike this formation and have eliminated hundreds of them.
Having separate statements has two advantages.
- Easier to read
- Easier to debug

```c
/* Prefer: Separate assignment from tests, where possible */
FILE *fp = fopen(...);
if (!fp)
{
}
```

- `!` is preferred for pointers and booleans
- Numeric types should be compared to `0` including functions that return int.
  This help to reinforce the type in the reader's mind.

---

Some of the worst examples are in `mutt_index_menu()` and `mutt_pager()`.
2000 line functions that keep re-using the same variables over and over.
You're never quite sure which paths through the code could set the variable before it's used.

ncrypt/pgp.c
```diff
@@ -1663,22 +1641,21 @@ struct Body *pgp_class_traditional_encryptsign(struct Body *a, int flags, char *
...
   {
-    mutt_perror(pgpout ? pgperrfile : pgpoutfile);
+    mutt_perror(pgpout ? "mutt_file_mkstemp() failed!" : pgpoutfile);
     unlink(pgpinfile);
     if (pgpout)
     {
       mutt_file_fclose(&pgpout);
       unlink(pgpoutfile);
     }
+    mutt_file_fclose(&pgperr);
```
flatcap: I added this line.
If the `pgpoutfile` failed, we could leak the `pgperr` file handle.

Issue #1182, opened 2018-04-30
Relative Source Files


flatcap:
Looking for something easier?  ([try these](https://github.com/neomutt/neomutt/issues?q=label:"good%20first%20issue"+is:open))

## Description

NeoMutt's `source` command works relative to the file it's in.
Extend this behaviour to four more config variables:
- `alias_file`
- `certificate_file`
- `history_file`
- `ssl_ca_certificates_file`

### Part 1: Make them relative

We should store the absolute path for these files.

If they're specified as relative paths, then expand them using
- Current config file's location (if reading from a config file)
- Current working directory of NeoMutt (if the user enters a command)

PR #1274, opened 2018-06-24
Allow variables that represent filenames to be relative

mjsir911:
* **What does this PR do?**

Allow variables such as alias_file to be relative and point to the correct locations

#1182

### DIFF
--- a/doc/makedoc.c
+++ b/doc/makedoc.c
@@ -959,7 +959,8 @@ enum DataType
   DT_MAGIC,
   DT_SYNONYM,
   DT_ADDRESS,
-  DT_MBTABLE
+  DT_MBTABLE,
+  DT_COMMAND,
 };
 
 struct VariableTypes
@@ -980,6 +981,7 @@ struct VariableTypes
   { "DT_SYNONYM", NULL },
   { "DT_ADDRESS", "e-mail address" },
   { "DT_MBTABLE", "string" },
+  { "DT_COMMAND", "command" },
   { NULL, NULL },
 };
 
@@ -1051,6 +1053,7 @@ static void pretty_default(char *t, size_t l, const char *s, int type)
     case DT_REGEX:
     case DT_ADDRESS:
     case DT_PATH:
+    case DT_COMMAND:
     case DT_MBTABLE:
     {
       if (strcmp(s, "0") == 0)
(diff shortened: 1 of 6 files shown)


mjsir911:
There's a single issue that needs addressing:

Currently all variables of the type `DT_PATH` are being translated. I'm not sure if this is wanted, but it seemed logical to expand all `DT_PATH` variables to absolute paths rather than relative paths.

Variables that are commands eg `editor`, `display_filter`, `inews`, `ispell`, `mixmaster`, `new_mail_command`, `pager`, `print_command`, `query_command`, `sendmail`, `shell`, & `visual` also get expanded to relative paths, so if I set my editor to `set editor = "vim +startinsert! +normal!ggA"`, then
```
:set ?editor
editor="~/.mutt/vim +startinsert! +normal!ggA"
```
Which is not preferable.

Is there a a way to differentiate between command paths and real paths? Maybe a new type should be added? `DT_COMMAND`?

flatcap:
Thanks for tackling this one.
It's one of those problems that when fixed, everything will work as people expect.

> Currently all variables of the type DT_PATH are being translated

Ah, good.

> Maybe a new type should be added? DT_COMMAND?

Please, do.  That's been on my ideas list for a while.
So, a `DT_COMMAND` wouldn't need expansion; we'd assume that `$PATH` would be sufficient.

Looking further ahead, I also wondered about `DT_MAILBOX` for: `$record`, `$spoolfile`, `$trash`.

flatcap:
I've had a quick look at the code (I haven't run it yet).
When source-ing a file the paths are relative to the config file they're read from.

When a user types: `:set alias_file = 'a.rc'` should that be relative to:
- `~/.muttrc` (wherever the user's config is)
- `.` (current directory, where NeoMutt was started)

mjsir911:
Currently it is using the same logic as `source_rc`

https://github.com/mjsir911/neomutt/blob/ca3d7a98ff9b211c8967fb02f04ef07083c6cb8d/init.c#L1229-L1230

`MuttrcStack` is non-empty during the initialization when config files are sourced, so any `set` would be relative to the current config file, but during use when a user were to just type `:set alias_file = 'a.rc'` it would be relative to the current directory (same as `source`)

flatcap:
Please can you convert the "*_command" config items, too.

mjsir911:
@flatcap, turned some `DT_STRING` variables ending with `_command` to `DT_COMMAND`

flatcap:
Lovely job, thanks.

flatcap:
_ahem_.  slight correction.
I've just pushed f2dfd2a4a ;-)

gahr:
Hi, sorry for jumping in so late but I was on vacation. I think we have a regression here:

My signature is
```
set signature       = "echo 'Pietro Cerutti'|"
````

Previously this resulted in my name, now `:set ?signature` gives me
````
signature="/usr/home/gahr/.config/neomutt/echo 'Pietro Cerutti'|"
````

Issue #1275, opened 2018-06-26
Kill ! (exclamation marks) in mutt_message() and mutt_error()


flatcap:
There's! No! Need! For! Them!
Error messages can be coloured if the user needs a sense of urgency.

Replace the `!`s with `.`s in:
- 6 `mutt_message()`
- 73 `mutt_error()`

Also, where possible, make the replacements in the `.po` files.
That will avoid 79 new fuzzy translations.

---

Please make a branch from **translate** for this work.

flatcap:
Fixed in df29bc8d2

PR #1281, opened 2018-06-27
Improvements in style II

BertramScharpf:
These are only suggestion that I didn't test too thourough. My main intention was to make the code more readable.

### DIFF
--- a/hcache/hcache.c
+++ b/hcache/hcache.c
@@ -132,7 +132,7 @@ static bool crc_matches(const char *d, unsigned int crc)
 
   unsigned int mycrc = *(unsigned int *) (d + sizeof(union Validate));
 
-  return (crc == mycrc);
+  return crc == mycrc;
 }
 
 /**
@@ -485,5 +485,5 @@ const char *mutt_hcache_backend_list(void)
  */
 bool mutt_hcache_is_valid_backend(const char *s)
 {
-  return (hcache_get_backend_ops(s) != NULL);
+  return hcache_get_backend_ops(s) != NULL;
 }
--- a/addrbook.c
+++ b/addrbook.c
@@ -145,7 +145,7 @@ static int alias_tag(struct Menu *menu, int n, int m)
 
   cur->tagged = (m >= 0 ? m : !cur->tagged);
 
-  return (cur->tagged - ot);
+  return cur->tagged - ot;
 }
 
 /**
--- a/addrbook.c
+++ b/addrbook.c
@@ -48,7 +48,7 @@
 char *AliasFormat;
 short SortAlias;
 
-#define RSORT(x) (SortAlias & SORT_REVERSE) ? -x : x
+#define RSORT(x) ((SortAlias & SORT_REVERSE) ? -x : x)
 
 static const struct Mapping AliasHelp[] = {
   { N_("Exit"), OP_EXIT },      { N_("Del"), OP_DELETE },
@@ -162,7 +162,7 @@ static int alias_sort_alias(const void *a, const void *b)
   struct Alias *pb = *(struct Alias **) b;
   int r = mutt_str_strcasecmp(pa->name, pb->name);
 
-  return (RSORT(r));
+  return RSORT(r);
 }
 
 /**
@@ -196,7 +196,7 @@ static int alias_sort_address(const void *a, const void *b)
     r = -1;
   else
     r = mutt_str_strcasecmp(pa->mailbox, pb->mailbox);
-  return (RSORT(r));
+  return RSORT(r);
 }
 
 /**
--- a/alias.c
+++ b/alias.c
@@ -288,7 +288,7 @@ struct Address *mutt_expand_aliases(struct Address *a)
   STAILQ_INIT(&expn);
   t = expand_aliases_r(a, &expn);
   mutt_list_free(&expn);
-  return (mutt_remove_duplicates(t));
+  return mutt_remove_duplicates(t);
 }
 
 /**
--- a/browser.c
+++ b/browser.c
@@ -138,7 +138,7 @@ static int browser_compare_subject(const void *a, const void *b)
   int r = mutt_inbox_cmp(pa->name, pb->name);
   if (r == 0)
     r = mutt_str_strcoll(pa->name, pb->name);
-  return ((SortBrowser & SORT_REVERSE) ? -r : r);
+  return (SortBrowser & SORT_REVERSE) ? -r : r;
 }
 
 static int browser_compare_desc(const void *a, const void *b)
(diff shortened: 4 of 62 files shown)


hcache/hcache.c
```diff
@@ -343,7 +343,7 @@ static void restore_address(struct Address **a, const unsigned char *d, int *off
     restore_char(&(*a)->personal, d, off, convert);
     restore_char(&(*a)->mailbox, d, off, false);
     restore_int(&g, d, off);
-    (*a)->group = g;
+    (*a)->group = g ? true : false;
```
mschilli87: Why not just add a comment here instead of introducing a noop and relying on the compiler to optimize it away?
even `(*a)->group = g /* ? true : false */;` would be possible.

flatcap:
> 8a3918e Remove parentheses from return statements

Here's an area that could do with some attention (if you're interested).
There are 4126 `return` statements.  Most of them are pretty simple.
However there are 80 whose line length > 60 chars (the worst is 190 chars).

Besides the length, conditionals make things hard to read, e.g.
```c
return (ib - d == dlen) ? dlen : ib - d + 1;
return (pat->not ^ ((h->security & SIGN) ? 1 : 0));
```

flatcap:
- 672a24198 Remove doubled function call in nested if statement
  very nice

- a75fcf0e3 Shorter if
- dc9d9f9bb Calculate mutt_fqdn not until needed
  very nice

send.c
```diff
@@ -1082,13 +1081,14 @@ struct Address *mutt_default_from(void)
     addr = mutt_addr_new();
     if (UseDomain)
     {
-        addr->mailbox =
-            mutt_mem_malloc(mutt_str_strlen(Username) + mutt_str_strlen(fqdn) + 2);
-        sprintf(addr->mailbox, "%s@%s", NONULL(Username), NONULL(fqdn));
+      const char *fqdn = mutt_fqdn(1);
```
flatcap: Since you branched, `mutt_fqdn()` now takes a bool -- this got lost in your merge.

BertramScharpf:
> There are 4126 return statements. Most of them are pretty simple. However there are 80 whose line length > 60 chars (the worst is 190 chars).

> Besides the length, conditionals make things hard to read, e.g.

What should happen to them? Parentheses around the before-`?` part or not? Line breaks after `?` and `:` or before them?

flatcap:
> Parentheses ... Line breaks

I like the parentheses around the tertiary's conditional: `(a > b) ? apple : banana`

> What should happen to them?

Perhaps factoring them out completely?

```c
return (ch >= 0) ? ch : EOF;
```

becomes

```c
if (ch >= 0)
 return ch;
return EOF;
```

or for a complicated example, just factor out the conditional part:

```c
return pgp_invoke(pgpin, pgpout, pgperr, keyring == PGP_SECRING ? PgpListSecringCommand : PgpListPubringCommand);
```

```c
int param = (keyring == PGP_SECRING) ? PgpListSecringCommand : PgpListPubringCommand
return pgp_invoke(pgpin, pgpout, pgperr, param);
```

and then there's **** like these :-)

```c
return ((old && new) ? new : (old ? old : new));
return occult ? MUTT_REOPENED : (ctx->msgcount > data->oldmsgcount) ? MUTT_NEW_MAIL : new_flags ? MUTT_FLAGS : 0;
```

We _do_ have a [style guide](https://neomutt.org/dev/coding-style), but I intentionally kept it short.

BertramScharpf:
I like the ternary operator very much in return statements. But as a function parameter, though the ternary operator takes precedence over the comma, it's unaestethic.

Nevertheless, I dislike function calls that are written out twice, whilst only one parameter is different (`mutt/address.c`):

```c
        if (display)
            mutt_str_strfcpy(pbuf, mutt_addr_for_display(addr), buflen);
        else
            mutt_str_strfcpy(pbuf, addr->mailbox,               buflen);
```

Are you worried about a ternary operator here?

```c
        const char *astr = display ? mutt_addr_for_display(addr) : addr->mailbox;
        mutt_str_strfcpy(pbuf, astr, buflen);
```

flatcap:
> I like the ternary operator very much in return statements.

OK, for the simple cases.

> But as a function parameter ... it's unaesthetic

Yes.  It makes the lines longer and harder to untangle in the debugger.

> I dislike function calls that are written out twice, whilst only one parameter is different

Yes, a bit ugly.

> Are you worried about a ternary operator here?

That's a good compromise.
The ternary operator gives us a simple either/or.
The function call has known parameters.

mschilli87:
Sorry for the noise guys but Germany was just kicked out of the FIFA world cup by South Korea (for all US Americans: [That is a big thing!](https://youtu.be/DlJEt2KU33I)), so I feel  it's OK to over-react emotionally. :wink:

I just wanted to point out that it's exactly these kinds of discussions that make me happy to have switched from Mutt to NeoMutt. Code readability means code maintainablity and that matters so much!

Thank you so much for caring.

PS: [Korea!!!](https://youtu.be/bCcHvoFIzQs)

PPS: [I! know! there! is! no! need! for! exclamation! marks! (!)!, but!, I! could! not! resist!](https://github.com/neomutt/neomutt/issues/1275)  :wink:

BertramScharpf:
About two years ago I had an interesting conversation with a maker of string instruments and with a stonemason. While talking, I first sketched the idea that the future of software development is not to write faster code or -- mainly -- more extensive code. I rather think there is a need to layer code with aesthetics. As one cannot control every single fibre in a log of wood or the cells of a molecular lattice, sooner or later we are not able to keep control over sources larger more than 100k lines. As the stonemason does not only concern what he forms but has a strong attitude how to hold the chisel as well, and as the violin maker is confident that the F-hole does not only look better but has the better acoustics as well, so I am deeply convinced that beautiful code is more maintainable, more robust, and more secure. Aesthetics can do much more than any cycle of testing and fixing.

Coming generations of software developers will have to keep aesthetics in mind, and it's not too soon to start with that.

Of course, aesthetics will always be a matter of taste.

flatcap:
I've merged some commits (thanks) and rebased this PR on top.

hcache/hcache.c
toogley: btw: as `group` is boolean, can we rename it  to e.g. `hasGroup` ?

flatcap: @toogley 
> `hasGroup`

no, no, NO!  That's terribly un-c!  `has_group` is much better :-)

I'll raise an issue (tomorrow) to discuss our options.

BertramScharpf: Sorry, I cannot resist: How about `bHasGroup`?

mschilli87: Not sure if you guys *tried* to point out my comment as bikeshedding or if that's just the natural progression when discussing these kinds of things... :laughing:

flatcap:
FYI, here's a cocci script to create the same result:

```spatch --in-place --sp-file return.cocci file.c```

```diff
@@
expression E;
@@

(
- return (E);
+ return E;
)
```

Some other [cocci scripts](https://github.com/neomutt/coccinelle).

Issue #1284, opened 2018-07-05
change-vfolder no longer works


Austin-Ray:
## Expected Behaviour

Using `change-vfolder` should bring a prompt to change virtual folders. 

## Actual Behaviour

`change-vfolder` does nothing.

If bound to a key with `?`, i.e. `<change-vfolder>?`, `change-vfolder` will be ignored and the help page will display.

## Steps to Reproduce

Minimal `.neomuttrc` included in Extra Info section

* Use Neomutt's notmuch integration and `virtual_spoolfile`
* Execute `<change-vfolder>` (I use `macro pager,index c "<change-vfolder>"  "Change folder"`)
* Nothing happens

## How often does this happen?

- Always

## When did it start to happen?

`git bisect` revealed it started with commit e85c7353.

## NeoMutt Version

```
NeoMutt 20180622-60-9e927a
Copyright (C) 1996-2016 Michael R. Elkins and others.
NeoMutt comes with ABSOLUTELY NO WARRANTY; for details type 'neomutt -vv'.
NeoMutt is free software, and you are welcome to redistribute it
under certain conditions; type 'neomutt -vv' for details.

System: Linux 4.17.4-1-ck (x86_64)
ncurses: ncurses 6.1.20180127 (compiled with 6.1.20180127)
libidn: 1.34 (compiled with 1.34)
hcache backends: gdbm

Compiler:
cc (GCC) 8.1.1 20180531
Copyright (C) 2018 Free Software Foundation, Inc.
This is free software; see the source for copying conditions.  There is NO
warranty; not even for MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.

Configure options: --prefix=/usr --sysconfdir=/etc --libexecdir=/usr/lib --gpgme --enable-lua --notmuch --gss --ssl --sasl --with-ui=ncurses --with-idn=/usr --gdbm

Compilation CFLAGS: -march=native -mtune=native -O2 -pipe -fstack-protector-strong -fno-plt -std=c99 -fno-delete-null-pointer-checks -D_ALL_SOURCE=1 -D_GNU_SOURCE=1 -D__EXTENSIONS__ -I/usr/include  -DNCURSES_WIDECHAR

Default options:
  +attach_headers_color +compose_to_sender +compress +cond_date +debug 
  +encrypt_to_self +forgotten_attachments +forwref +ifdef +imap +index_color 
[...]

PR #1285, opened 2018-07-05
Do not interrupt switch fallthrough

Austin-Ray:
### What does this PR do?

Commit e85c735 introduced the `check-stats` function, designated by
`OP_CHECK_STATS`.

`OP_CHECK_STATS` had been inserted below `OP_MAIN_CHANGE_VFOLDER`, which is
`change-vfolder`'s designation; however, `OP_MAIN_CHANGE_VFOLDER` relies
on C's switch fallthrough logic to execute. As a result,
`change-vfolder` executed `check-stats`'s logic instead of the correct
logic.

This PR moves `OP_CHECK_STATS` to above the `#ifdef` that contains
`OP_MAIN_CHANGE_VFOLDER`, allowing the original `change-vfolder` logic to 
execute instead of `check-stats`'s logic.

#### What are the relevant issue numbers?

Issue #1284

### DIFF
--- a/curs_main.c
+++ b/curs_main.c
@@ -1905,6 +1905,10 @@ int mutt_index_menu(void)
         break;
       }
 
+      case OP_CHECK_STATS:
+        mutt_check_stats();
+        break;
+
 #ifdef USE_NOTMUCH
       case OP_MAIN_VFOLDER_FROM_QUERY:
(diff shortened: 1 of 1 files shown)


flatcap:
Sorry about that.
An upstream commit got in the way.

PR #1289, opened 2018-07-11
 add keep_ctx config option 

sassmann:
* **What does this PR do?**

Mailbox contexts are kept in memory, instead of being discarded on
mailbox switch. This comes with the trade-off of requiring more memory.
However, switching mailboxes is ultra-fast once they've been loaded.

This is a first attempt to implement this, not asking for a merge yet.
Feedback is welcome though!

### DIFF
--- a/buffy.h
+++ b/buffy.h
@@ -27,6 +27,7 @@
 #include <stdbool.h>
 #include <sys/types.h>
 #include <time.h>
+#include "context.h"
 #include "where.h"
 
 struct Buffer;
@@ -53,6 +54,7 @@ struct Buffy
   char *desc;
   off_t size;
   struct Buffy *next;
+  struct Context *ctx;       /**< store context for re-use */
   bool new; /**< mailbox has new mail */
 
   /* These next three are only set when MailCheckStats is set */
--- a/curs_main.c
+++ b/curs_main.c
@@ -453,6 +453,8 @@ void update_index(struct Menu *menu, struct Context *ctx, int check, int oldcoun
 static int main_change_folder(struct Menu *menu, int op, char *buf,
                               size_t buflen, int *oldcount, int *index_hint)
 {
+  struct Context *ctx_exist = NULL;
+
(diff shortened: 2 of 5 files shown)


sassmann:
tested for 2 weeks, request merge

flatcap:
> what about making it default, removing the option?

Before we could do that, I'd want to know how much memory each of the Contexts is taking up.
Some users have hundreds of mailboxes, with 100,000s of emails and keep NeoMutt open for long periods.
Do we put an upper limit on it?
How would the user clear the cache?

Also, the memory-management in NeoMutt is dreadful.
If this code wasn't perfect and didn't cover **every** route through the mailbox code,
then it would affect all users.

flatcap:
On a side note, I had an idea for a testing ground.
(I don't want to add code/options that I later have to remove)

```
# Enable features under development
./configure --development --keep_ctx
```

And prefix the option names: `$devel_keep_ctx`

That way users can easily test new features, but they know that the aren't fully supported yet.

If the feature proves safe, then it gets promoted out of 'devel'.

sassmann:
@toogley I consciously made the decision to not use the feature by default as it comes at a cost, namely increased memory footprint. neomutt might be used on system with little memory so the user should actively enable the feature.
@flatcap I didn't want to overengineer this yet. If we really need to implement some memory limits we can add an option for it later. Easy way to drop memory is to restart neomutt.

flatcap:
> I didn't want to overengineer

Absolutely.  Keep it simple.

sassmann:
@flatcap 
For comparison memory footprint of neomutt
after start (no mailbox opened)
pidof neomutt | xargs ps -o rss
  RSS
15392
after reading several maildirs with 100k mails total
pidof neomutt | xargs ps -o rss
  RSS
248320
That's about 2,3kB per mail used.

How about I rename the option to keep_ctx_devel? That doesn't screw up sorting.
The code itself is guarded by the config option and it's off by default. I think that's as much precaution as we need for now.

PR #1307, opened 2018-07-28
MUTT_FORMAT_PLAIN FormatFlag to toggle DISP_TO

x3nb63:
adds a FormatFlag to toggle the "To " created by %F expando, see https://github.com/neomutt/neomutt.github.io/issues/69

* this flag is not controllable by the user so far. Perhaps its better to make a new expando because there may be users wanting that "To "?

* given the current flag name of MUTT_FORMAT_PLAIN, it would be logical to apply it to DISP_CC, DISP_BCC, and DISP_FROM as well, which would further change behaviour of %F. (thats in `make_from()` )

Please comment. I am new to this project and to writing C, so I expect to do more iterations on the whole thing.


flatcap:
OK, that's a good start.
I've rebased your changes on top of master.
This should get you sync'd:
- `git fetch`
- `git reset --hard origin/issue/69`

I suggest adding a new expando, perhaps `%Fp` (From plain).
You'll need to alter `index_format_str()`
Have a look at the `%zc`, `%zs`, `%zt` expandos for how to tell `F` from `Fp`.

x3nb63:
ok, so aa9f143 has the `%Fp` expando working.

The `DISP_PLAIN` header type is for clarity, as I think its odd to use `DISP_FROM` when formatting a To: field (`env->to`). All it takes is a empty string, that is _no formatting_.

Rebase onto master failed, too many changes. Please tell me how to go from here.

flatcap:
> Rebase onto master failed, too many changes.

Don't worry.  When I've been making huge changes, I'm happy to do the rebasing.
Currently, `master` isn't completely stable, so I might leave it in place, for now.

> ok, so aa9f143 has the %Fp expando working.

Great.  I'll have a close look later.

> Please tell me how to go from here.

If the code works and looks tidy, then you're done :-)

hdrline.c
```diff
@@ -861,12 +865,15 @@ static const char *index_format_str(char *buf, size_t buflen, size_t col, int co
       if (!optional)
       {
```
flatcap: I think I'd factor out the `src[0]` check, maybe:
`const bool plain = (src[0] == 'p');`

hdrline.c
```diff
@@ -861,12 +865,15 @@ static const char *index_format_str(char *buf, size_t buflen, size_t col, int co
...
         colorlen = add_index_color(buf, buflen, flags, MT_COLOR_INDEX_AUTHOR);
-        make_from(hdr->env, tmp, sizeof(tmp), false);
+        make_from(hdr->env, tmp, sizeof(tmp), false, (src[0]=='p' ? MUTT_FORMAT_PLAIN : 0));
         mutt_format_s(buf + colorlen, buflen - colorlen, prec, tmp);
         add_index_color(buf + colorlen, buflen - colorlen, flags, MT_COLOR_INDEX);
+
+        if (src[0]=='p')  src++;
```
flatcap: split into two lines

hdrline.c
```diff
@@ -861,12 +865,15 @@ static const char *index_format_str(char *buf, size_t buflen, size_t col, int co
...
       }
-      else if (mutt_addr_is_user(hdr->env->from))
+      else if (mutt_addr_is_user(hdr->env->from)) {
```
flatcap: move `{` to next line

x3nb63:
thanks. Will do. I am just offline for a week, but then.

Also thanks for all the work you put in here.

hdrline.c
x3nb63: called it 'is_plain`

x3nb63:
if it spares your time I can try to tweak it on top of master

flatcap:
> if it spares your time I can try to tweak it on top of master

OK, thanks.  Give it go!

PR #1308, opened 2018-07-29
Attachmentfix

0x747972:
This pull requests "fixes" the "bug" I filed last week.
It does simply so by waiting for like 0.25 seconds to give programs returning directly a bigger time window to read the soon-to-be unlinked file. (I know this is pretty myopic.)

The other method we had briefly discussed, namely keeping a list of files to be deleted on exit, would probably require the introduction of at least one global variable to be introduced. (So neomutt has always access to that list, anytime it quits. But I don't really know neomutts inner workings.) This would not be compatible with your style guide, i guess.

### DIFF
--- a/mutt_attach.c
+++ b/mutt_attach.c
@@ -633,6 +633,9 @@ int mutt_view_attachment(FILE *fp, struct Body *a, int flag, struct Header *hdr,
 
 return_error:
 
+  //lolpatch for program that return promptly
+  sleep(1);
+
   if (entry)
     rfc1524_free_entry(&entry);
   if (fp && tempfile[0])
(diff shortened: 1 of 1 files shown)


gahr:
https://gitlab.com/muttmua/mutt/wikis/MuttFaq/Attachment

Please look for “How to view HTML attachments in a new tab in firefox”.

Basically, it suggests to have the sleep in the corresponding mailcap entry, instead of hardcoded in the code.

0x747972:
Yeah, I fully agree that the solution is kinda hackish, but seems to be a simple way to archieve this end without rewriting lots of code, i suppose.

flatcap:
Hmm...  I'm very reluctant to refuse code, but...
Yes.  Sorry, it's too hacky.

I think that 0.25s is too short for many people.
Make it longer and you'd adversly affect other users.

There's already a mailcap workaround.
I'd prefer to suggest that people use _that_, until we create a better solution.

Issue #589, opened 2017-05-22
No UTF8 support for subject line?


andreas-bulling:
Hi,

UTF8 support works fine for neomutt in my terminal (iTerm2 on MacOSX) except for the subject line: When I start a new email and try to enter a German umlaut in the subject I only end up getting garbage.

I am using NeoMutt 20170428 (1.8.2)

flatcap:
I've just had a quick look and everything seems fine (for me a a couple of others on IRC).
I've tested extended ascii and a range of Unicode characters (both sending and receiving).

What's your locale set to in the terminal?
What's `charset` set to in mutt -- `mutt -D | grep charset`

andreas-bulling:
Thanks for your quick reply. As I said, message body is not a problem, only subject line (after pressing "m") causes problems.

Output of `mutt -D | grep charset`:
```
assumed_charset=""
attach_charset=""
charset="utf-8"
config_charset=""
send_charset="us-ascii:utf-8"
```

andreas-bulling:
Just tried something else: If I press button "[", with the German "input source" enabled in MacOSX, I get a "ü" in the terminal but the subject line shows "C<". Is this some modifier key issue? But, again, if I then edit the message in nvim everything is fine, i.e. the right characters are sent to/by the terminal.

riesebie:
* andreas-bulling <notifications@github.com> [2017-05-23 09:23 -0700]:

> Thanks for your quick reply. As I said, message body is not a problem, only subject line (after pressing "m") causes problems.
> 
> Output of `mutt -D | grep charset`:
> ```
> assumed_charset=""
> attach_charset=""
> charset="utf-8"
> config_charset=""
> send_charset="us-ascii:utf-8"

What happens if you set
send_charset="utf-8"
?
-- 
  The path to source is always uphill!
                                -unknown-

andreas-bulling:
no change :/

gahr:
@andreas-bulling do you happen to have `mime_subject` unset?

andreas-bulling:
I just typed some German umlaute in another terminal and tried to copy them into the subject line - doesn't work, I directly receive garbage

gahr:
That is surprising. Does that happen with any input forms (say, `To:`) or just with `Subject:`?

andreas-bulling:
Yes, this also happens for the `To:` input form and I also can't type German umlaute there either as I just noticed (never tried before)

gahr:
[...]

PR #1310, opened 2018-08-09
[WIP] Do not let libintl.h override setlocale

dpgao:
This PR aims to partly address #589--only partly--because I believe there are two separate problems reported there, and this fixes only one of them. Both @rlue and I encountered the problem where if `locale` is set to the following:
```
LANG=
LC_COLLATE="C"
LC_CTYPE="UTF-8"
LC_MESSAGES="C"
LC_MONETARY="C"
LC_NUMERIC="C"
LC_TIME="C"
LC_ALL=
```
then `charset` would be incorrectly set to `us-ascii`, causing problems when displaying UTF-8 characters. After some debugging I found out that https://github.com/neomutt/neomutt/blob/eb8b461a79c0c32c8c8eae70a0b762b57585f354/main.c#L323 is actually not using the `setlocale` function provided in `locale.h`, but rather overridden by a macro defined in `libintl.h`, which behaves incorrectly in macOS.

This PR includes a crude fix that addresses the immediate issue by disabling the macro. This is not an optimal solution, so I would love to hear more suggestions and work on this more.

I can think of some possible long-term solutions:
* Refactor the code so that `libintl.h` is isolated from other files as much as possible.
* Change the build configuration to automatically disable the `libintl.h` macro on certain platforms.

**Update**: I have submitted a [bug report](https://savannah.gnu.org/bugs/index.php?54479) to the `gettext` project.

### DIFF
--- a/main.c
+++ b/main.c
@@ -29,6 +29,7 @@
  */
 
 #define MAIN_C 1
+#define GNULIB_defined_setlocale
 
 #include "config.h"
 #include <errno.h>


gahr:
How about simply defining this macro in auto.def?

```
diff --git a/auto.def b/auto.def
index a6c7ff2b6..6e9d53b75 100644
--- a/auto.def
+++ b/auto.def
@@ -459,6 +459,7 @@ if {[get-define want-nls]} {
    user-error "Unable to find gettext tools (msgfmt, msgmerge, xgettext).\
                Consider --disable-nls"
  }
+ define-append CFLAGS -DGNULIB_defined_setlocale
 }
```

gahr:
So is it only the call with `LC_ALL` as a first argument that's not working properly on osx? You mean all other calls to `setlocale` should go to libintl's implementation? I'm not sure this sounds robust to me...

dpgao:
> So is it only the call with LC_ALL as a first argument that's not working properly on osx?

Yes, because the `libintl` version of `setlocale` specifically overrides its behaviour when passed with `LC_ALL`.

> You mean all other calls to setlocale should go to libintl's implementation? I'm not sure this sounds robust to me...

`mutt` avoided (or didn't run into) this problem because they did not include `libintl` in the file where `setlocale(LC_ALL, "")` is called. We can either do what they do, or explicitly disable the override as you suggested above.

gahr:
I understand - I don't have enough intl-foo to completely understand the consequence of either, mostly because I don't know how libintl_setlocale differs from setlocale and why it's broken. Having said that, I think your proposed change is fine, as in, works for me :)

dpgao:
Great! We can probably merge this for now, and I'll watch closely when `libintl` fixes the behaviour. By then we can revert this change.

flatcap:
> I would love to hear more suggestions and work on this more
> Refactor the code so that libintl.h is isolated from other files as much as possible.

I've just had a quick look through our uses of `setlocale()`.
There are ten places:
- 4 places: `setlocale` to C and call `strftime()`
- 4 places: `setlocale` to C and call `mutt_make_string()` which has date/time components
- `main.c` has `setlocale(LC_ALL, "");`
- Some strange code in `ncrypt/crypt_gpgme.c`

I'd happily accept a PR to create `locale.c` which wraps up most of these cases.

dpgao:
I'm willing to help on this. Should I simply wrap `setlocale` into a function called `mutt_set_locale`, or should I differentiate between different use cases and create functions for each? Also, would it be good to wrap all of `gettext` into `mutt_` functions so that `libintl.h` is included in only one file?

dpgao:
> In upstream Mutt:
> main.c includes "mutt.h" includes "lib.h" includes <libintl.h>

It turns out that the default Homebrew installation of `mutt` disables NLS support, so `libintl.h` is not included at all.

> but the correct thing to do (I think) would be to restore the previous locale settings:

Yes, but we actually need _three_ `setlocale` calls to accomplish this: one to back up the current locale, one to set the new locale, and one to restore the old locale. Apart from this we also need to copy the backup buffer and free it at the end. This [page](https://ftp.gnu.org/old-gnu/Manuals/glibc-2.2.3/html_chapter/libc_7.html#SEC112) provides a way to do this in a rather functional way, but we cannot do this. Maybe a macro instead?

gahr:
How about we don't do anything, add EXTRA_CFLAGS=-DGNULIB_defined_setlocale to the macos formulas, and wait for libintl to fix the bug at their side? :)

flatcap:
Yeah, that's probably best.
I'm just impatient when it comes to waiting for 'upstream' to do things :-)

dpgao:
I found a [POSIX](http://pubs.opengroup.org/onlinepubs/9699919799/functions/strftime.html) function `strftime_l` that takes an explicit locale parameter instead of using the global locale. This way we won't need to backup and restore the locale ourselves. The only problem is probability. How do I check if the function exists on all supported platforms?

flatcap:
Thanks for investigating, I hadn't heard of that function.
Unfortunately, we chose to restrict NeoMutt to POSIX:2001 for compatibility with Mutt (and lots of old distros).

While we're still re-architecting NeoMutt, we'll stick with the old standards.

I'll merge your original solution later today.
Thanks.

PR #1313, opened 2018-08-14
test/address.c: remove duplicated include

z2v:
* **What does this PR do?**
Remove duplicated include from test/address.c


flatcap:
Oops!  Thanks

Issue #1214, opened 2018-05-17
Can't add '&type=<threads|messages>' to vfolder query


jpoimboe:
## Expected Behaviour

With the vfolder-from-query command, I added &type=messages to the query in order to override my nm_query_type default of threads, like:

`<vfolder-from-query>neomutt&type=messages`

I expected to see a vfolder of search results.

## Actual Behaviour

The vfolder is empty.  The shown URL is `///home/jpoimboe/Mail?type=threads&query=neomutt%26type=messages`.

Notice the '&' was encoded, which I think was introduced with #603.

## Steps to Reproduce

`<vfolder-from-query>neomutt&type=messages`

## How often does this happen?

- Always

## When did it start to happen?

- When I migrated to neomutt from mutt-kz master

## NeoMutt Version

```
NeoMutt 20180323
Copyright (C) 1996-2016 Michael R. Elkins and others.
NeoMutt comes with ABSOLUTELY NO WARRANTY; for details type 'neomutt -vv'.
NeoMutt is free software, and you are welcome to redistribute it
under certain conditions; type 'neomutt -vv' for details.

System: Linux 4.16.7-300.fc28.x86_64 (x86_64)
ncurses: ncurses 6.1.20180224 (compiled with 6.1)
libidn: 1.34 (compiled with 1.33)
hcache backends: tokyocabinet, lmdb

Compiler:
Using built-in specs.
COLLECT_GCC=cc
COLLECT_LTO_WRAPPER=/usr/libexec/gcc/x86_64-redhat-linux/8/lto-wrapper
OFFLOAD_TARGET_NAMES=nvptx-none
OFFLOAD_TARGET_DEFAULT=1
Target: x86_64-redhat-linux
[...]

PR #1319, opened 2018-08-18
Allow `type=` statements in `vfolder-from-query`

Austin-Ray:
**What does this PR do?**

This addresses an issue with how Neomutt encodes notmuch queries. If one uses `vfolder-from-query` and includes a `type=` statement, Neomutt will encode the statement as part of the query, which prevents the query type from being changed. Including `type=` statements is documented as supported behavior[0].

This PR achieves the desired result by introducing a case-insensitive implementation of `strstr(...)`, a function for removing all instances of a case-insensitive substring, and combining the two functions to find if a `type=` statement exists and to set the appropriate `query_type` in `NmCtxData`if one exists.

[0] https://neomutt.org/feature/notmuch#3-2-%C2%A0items

#1214

### DIFF
--- a/mutt/string.c
+++ b/mutt/string.c
@@ -1016,3 +1016,47 @@ bool mutt_str_inline_replace(char *buf, size_t buflen, size_t xlen, const char *
 
   return true;
 }
+
+/**
+ * mutt_str_strcasestr - Find a substring within a string without worrying about case
+ * @param haystack String that may or may not contain the substring
+ * @param needle   Substring we're looking for
+ * @retval ptr  Beginning of substring
+ * @retval NULL Substring is not in substring
+ *
+ * This performs a byte-to-byte check so it will return unspecified
+ * results for multibyte locales.
+ */
+const char *mutt_str_strcasestr(const char *haystack, const char *needle)
+{
+  if (!needle)
+    return NULL;
+
+  size_t haystack_len = mutt_str_strlen(haystack);
+  size_t needle_len = mutt_str_strlen(needle);
+
+  // Empty string exists at the front of a string. Check strstr if you don't believe me.
+  if (needle_len == 0)
+    return haystack;
+
+  // Check size conditions. No point wasting CPU cycles.
+  if ((haystack_len == 0) || (haystack_len < needle_len))
+    return NULL;
+
+  // Only check space that needle could fit in.
+  // Conditional has + 1 to handle when the haystack and needle are the same length.
+  for (size_t i = 0; i < (haystack_len - needle_len) + 1; i++)
+  {
+    for (size_t j = 0; j < needle_len; j++)
+    {
+      if (tolower((unsigned char) haystack[i + j]) != tolower((unsigned char) needle[j]))
+        break;
+
+      // If this statement is true, the needle has been found.
+      if (j == (needle_len - 1))
+        return haystack + i;
+    }
+  }
+
+  return NULL;
+}
--- a/mutt/string.c
+++ b/mutt/string.c
@@ -1018,6 +1018,28 @@ bool mutt_str_inline_replace(char *buf, size_t buflen, size_t xlen, const char *
 }
 
 /**
+ * mutt_str_remall_strcasestr - Remove all occurrences of substring, ignoring case
+ * @param str     String containing the substring
+ * @param target  Target substring for removal
+ * @retval 0 String contained substring and substring was removed successfully
+ * @retval 1 String did not contain substring
+ */
+int mutt_str_remall_strcasestr(char *str, const char *target)
+{
+  int retval = 1;
+
+  // Look through an ensure all instances of the substring are gone.
+  while ((str = (char *) mutt_str_strcasestr(str, target)))
+  {
+    size_t target_len = mutt_str_strlen(target);
+    memmove(str, str + target_len, 1 + strlen(str + target_len));
+    retval = 0; // If we got here, then a substring existed and has been removed.
+  }
+
+  return retval;
+}
+
+/**
  * mutt_str_strcasestr - Find a substring within a string without worrying about case
  * @param haystack String that may or may not contain the substring
  * @param needle   Substring we're looking for
--- a/notmuch/mutt_notmuch.c
+++ b/notmuch/mutt_notmuch.c
@@ -1856,6 +1856,36 @@ done:
 }
 
 /**
+ * Parse a query type out of a query.
+ * @param data Mailbox, used for the query_type
+ * @param buf  Buffer for URI
+ *
+ * If a user writes a query for a vfolder and includes a type= statement, that
+ * type= will be encoded, which Notmuch will treat as part of the query=
+ * statement. This method will remove the type= and set it within the Mailbox
+ * struct.
+ */
+void nm_parse_type_from_query(struct NmMboxData *data, char *buf)
+{
+  // The six variations of how type= could appear.
+  const char *variants[6] = { "&type=threads", "&type=messages",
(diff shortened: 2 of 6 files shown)


mutt/string.c
```diff
@@ -959,3 +959,71 @@ const char *mutt_str_getenv(const char *name)
...
+ * @param str     String containing the substring
+ * @param target  Target substring for removal
+ * @retval 0      String contained substring and substring was removed successfully.
+ * @retval != 0   String did not contain substring.
+ */
+int mutt_str_remall_strcasestr(char *str, const char *target)
+{
+  // There's no point in removing something that is not there.
+  if (mutt_str_strcasestr(str, target) == NULL)
+    return 1;
```
flatcap: Do we want to do this (expensive) check unnecessarily?
We might as well let the `while` do it.

mutt/string.c
```diff
@@ -959,3 +959,71 @@ const char *mutt_str_getenv(const char *name)
...
+
+  // Look through an ensure all instances of the substring are gone.
+  while ((str = (char *) mutt_str_strcasestr(str, target)))
+  {
+    size_t target_len = mutt_str_strlen(target);
+    memmove(str, str + target_len, 1 + strlen(str + target_len));
```
flatcap: ... and set a flag here, to determine the return value

notmuch/mutt_notmuch.c
```diff
@@ -1883,6 +1883,34 @@ int nm_read_entire_thread(struct Context *ctx, struct Header *h)
...
+ * @param ctx   Mailbox, used for the query_type
+ * @param buf   Buffer for URI
+ */
+void parse_nm_query_type_from_query(struct NmCtxData *data, char *buf) {
+  // The six variations of how type= could appear.
+  const char *variants[6] = {"&type=threads", "&type=messages",
+    "type=threads&", "type=messages&",
+    "type=threads", "type=messages"};
+
+  for (int i = 0; i < 6; i++)
```
flatcap: Replace `6` with `mutt_array_size(variants)`

notmuch/mutt_notmuch.c
```diff
@@ -1883,6 +1883,34 @@ int nm_read_entire_thread(struct Context *ctx, struct Header *h)
...
+ * remove the type= and set it within the Mailbox struct.
+ *
+ * @param ctx   Mailbox, used for the query_type
+ * @param buf   Buffer for URI
+ */
+void parse_nm_query_type_from_query(struct NmCtxData *data, char *buf) {
+  // The six variations of how type= could appear.
+  const char *variants[6] = {"&type=threads", "&type=messages",
+    "type=threads&", "type=messages&",
+    "type=threads", "type=messages"};
```
flatcap: This seems a bit awkward...
The first two are only correct if at the end of a query.
The second two, only at the start.
The final two only if they are the entire query.

Otherwise, respectively...

- `notmuch://?query=hello&type=threadsJUNK`
- `notmuch://?JUNKtype=threads&query=hello`
- `notmuch://?JUNKtype=threadsJUNK`

mutt/string.c
Austin-Ray: I am unsure how to achieve this due to my lack of familiarity with C. Could you provide an example or link to an example?

Austin-Ray:
> However a couple of the tests don't work!

Which tests do not work? All test cases passed in Travis-CI.
Is it how I wrote the tests?

notmuch/mutt_notmuch.c
flatcap: ok, thanks

flatcap:
```
test/neomutt-test -v test_string_strcasestr

Test test_string_strcasestr:
  test/string.c:129: Check Expected: (null), Actual (null)... ok
  test/string.c:130: Check Expected: (null), Actual (null)... ok
  test/string.c:131: Check Expected: (null), Actual (null)... ok
  test/string.c:139: Check Expected: , Actual ... ok
  test/string.c:140: Check Expected: (null), Actual (null)... ok
  test/string.c:141: Check Expected: hello, Actual hello... ok
  test/string.c:149: Check Expected: (null), Actual (null)... ok
  test/string.c:150: Check Expected: (null), Actual (null)... ok
  test/string.c:151: Check Expected: (null), Actual (null)... ok
  test/string.c:157: Check Expected: hello, Actual: hello... ok
  test/string.c:163: Check Expected: hello, world!, Actual: hello, world!... ok
  test/string.c:170: Check Expected: hello, world, Actual: hello, world... failed
  test/string.c:177: Check Expected: hello, Actual: hello... failed
  test/string.c:183: Check Expected: (null), Actual: (null)... ok
  2 conditions have FAILED.

Summary:
  Count of all unit tests:       14
  Count of run unit tests:        1
  Count of failed unit tests:     1
  Count of skipped unit tests:   13
  FAILED: 1 of 1 unit tests have failed.
```

I can investigate later.

mutt/string.c
Austin-Ray: Okay, I think I might have figured out how to achieve this.

Would it be something like:

```C
int retval = 0;

while ((retval = (str = (char *) mutt_str_strcasestr(str, target))))
{
   // loop body
}

return retval;
```

That seems mess-y and hard to read, but compiles and works; however, I believe it is abuse of the standard (and some poor, future individual's sanity.)

flatcap:
```sh
# gcc fails
gcc --version
gcc (GCC) 8.1.1 20180712 (Red Hat 8.1.1-5)

# clang works
clang --version
clang version 6.0.1 (tags/RELEASE_601/final)
```

Hmm... this is very interesting.  It's to do with optimisations.
The test at line 170 compares two pointers, rather than doing `strcmp()`.

```c
if (retval == expected)
```

In my gcc, you have two strings in the code:
- `char *haystack = "test! hello, world";`
- `const char *expected = "hello, world";`

Your compiler has noted that despite being non-const, that you don't actually modify the haystack.
It also notes that the `expected` string is a substring of `haystack`.

Effectively, it's compiling:
- `const char *haystack = "test! hello, world";`
- `const char *expected = haystack + 6;`

When you search for the `needle`, you get a pointer into the `haystack`, which coincidentally is where `expected` is.  So, `(retval == expected)` !

If the compiler doesn't make these heavy optimisations, then `(retval == expected)` will **never** be true because they're entirely separate strings.

Austin-Ray:
> If the compiler doesn't make these heavy optimisations, then (retval == expected) will never be true because they're entirely separate strings.

That makes complete sense. I would need to use `strcmp(...)` to actually check the string content. I have yet to grok the C way of handling strings. This error was introduced by muscle memory since I use other languages where such a comparison is valid.

Since I wrote `mutt_str_strcasestr(...)` to mimic `strstr(...)`'s behavior, would it be a good idea to use `strstr(...)` to check the output (e.g. since I know the input, i can write the `strstr(...)` with the right case then compare pointers.)?

flatcap:
> I use other languages where such a comparison is valid.

Yeah, we're not using `std::string`... yet.

> would it be a good idea to use strstr(...) to check the output
> ... then compare pointers

Yes, that would be good.  Check exactly what we're expecting to have returned.

Austin-Ray:
I addressed the test case issue, removed a redundant `mutt_str_strcasestr(...)` call, and removed the magic number from `parse_nm_query_type_from_query(...)`.

Austin-Ray:
@flatcap Are you still interested in this PR? I can rebase this PR if so.

flatcap:
> Are you still interested in this PR?
> I can rebase this PR if so.

Yes and yes please.

Austin-Ray:
@flatcap Branch has been rebased and passed in CI.

PR #1338, opened 2018-09-13
More translations

winderickxeli:
I've added more dutch translations.


flatcap:
Thanks.  I've grafted your changes onto the `translate` branch.
Using that branch insulates you from all the crazy changes I'm making in `master`.

flatcap:
Unless you need a second opinion, I suggest using the `translate` branch directly.
```sh
git clone -b translate https://github.com/neomutt/neomutt
cd neomutt
# Edit po/nl.po
# Commit changes
git push
```

Later, if you remember to `git pull` before making any new changes, then your new commits won't need merging or rebasing.

winderickxeli:
Hi @flatcap, thanks! I'll do that for future pushes

Issue #1345, opened 2018-09-22
support notmuch thread reconstruction even when not in vfolder


zacchiro:
I'm the author of [notmuch-mutt](https://notmuchmail.org/notmuch-mutt/), but I've been looking into neomutt's notmuch support as its integration potential is *much* better. Thanks a lot for that. There is a specific use case in which notmuch-mutt user experience is still better, though, and I think it can be easily improved into neomutt.

Specifically, it is useful to be able to reconstruct the full thread of a given message (similar to what `<entire-thread>` does) even when the user is *not* in a vfolder. Currently, trying to use `<entire-thread>` in that situation results in an error: "No virtual folder, aborting.".

Here is what I think should be done instead:

- run a notmuch query on-the-fly filtering on `id:MESSAGE_ID`, where `MESSAGE_ID` is the message identifier of the currently selected message (in the index or pager)
- jump to the vfolder corresponding to the result of that query, similarly to what `<vfolder-from-query>` does
- expand the thread as `<entire-thread>` currently does

I think that doing this would provide a much smoother UX. What do you think?

flatcap:
Hi @zacchiro.  Welcome to NeoMutt!  Sorry for not reaching out to you.

> neomutt's notmuch support as its integration potential is much better.
> Thanks a lot for that.

Hats off to @karelzak

> it is useful to be able to reconstruct the full thread of a given message

> run a notmuch query on-the-fly filtering on id:MESSAGE_ID,
> jump to the vfolder corresponding to the result of that query,
> expand the thread as <entire-thread> currently does

> I think that doing this would provide a much smoother UX. What do you think?

Sounds reasonable (but I'm not a Notmuch expert).

@neomutt/notmuch-users What do you think?

PR #1346, opened 2018-09-23
notmuch: reconstruct thread on-the-fly upon <entire-thread> w/o vfolder

zacchiro:
* **What does this PR do?**

make `<entire-thread>` work when not in a virtual folder

#1345

### DIFF
--- a/curs_main.c
+++ b/curs_main.c
@@ -1898,8 +1898,25 @@ int mutt_index_menu(void)
       {
         if (!Context || (Context->mailbox->magic != MUTT_NOTMUCH))
         {
-          mutt_message(_("No virtual folder, aborting"));
-          break;
+          if (!CURHDR || !CURHDR->env || !CURHDR->env->message_id)
+          {
+            mutt_message(_("No virtual folder and no Message-Id, aborting"));
+            break;
+          } // no virtual folder, but we have message-id, reconstruct thread on-the-fly
+          strncpy(buf, "id:", sizeof(buf));
+          int msg_id_offset = 0;
+          if ((CURHDR->env->message_id)[0] == '<')
+            msg_id_offset = 1;
+          mutt_str_strcat(buf, sizeof(buf), (CURHDR->env->message_id) + msg_id_offset);
+          if (buf[strlen(buf) - 1] == '>')
+            buf[strlen(buf) - 1] = '\0';
+          if (!nm_uri_from_query(Context->mailbox, buf, sizeof(buf)))
+          {
+            mutt_message(_("Failed to create query, aborting"));
+            break;
+          }
+          else
+            main_change_folder(menu, op, buf, sizeof(buf), &oldcount, &index_hint);
         }
         CHECK_MSGCOUNT;
         CHECK_VISIBLE;


flatcap:
Thanks @zacchiro 

@neomutt/notmuch-users 
Please can you give it a test.

flatcap:
**Note to all**:
`master` is currently a little unstable, since I'm making some big changes.
So if NeoMutt crashes, it's my fault :-)

zacchiro:
This should probably not be merged yet, as it still causes a segfault in the following scenario:

- open a regular folder with at least one message
- use `<entire-thread>` as per this patch to jump to the vfolder showing its thread
- open again the initial folder (e.g., with `Ctrl+O` in the sidebar)
- → segfault

I've no idea if it's my fault or not and I'm completely new to the (neo)mutt codebase. The reason might be obvious to more experienced developers… Comments and help is very welcome!

zacchiro:
here's the backtrace I'm seeing for the aforementioned segfault:
```
                                                  Program received signal SIGSEGV, Segmentation fault.
                                                                                                      0x0000555555603a13 in crypt_query (m=0x6cf000006ce) at ncrypt/crypt.c:654
654	  if (!m)
(gdb) bt
#0  0x0000555555603a13 in crypt_query (m=0x6cf000006ce) at ncrypt/crypt.c:654
#1  0x00005555555bd33f in mx_update_context (ctx=ctx@entry=0x55555624c750, new_messages=<optimized out>) at mx.c:1197
#2  0x00005555555fecd4 in maildir_add_to_context (ctx=ctx@entry=0x55555624c750, md=0x0) at maildir/mh.c:851
#3  0x0000555555600c29 in maildir_move_to_context (md=0x7fffffff9910, ctx=0x55555624c750) at maildir/mh.c:1262
#4  0x0000555555600c29 in mh_read_dir (ctx=ctx@entry=0x55555624c750, subdir=subdir@entry=0x55555564d5a6 "cur") at maildir/mh.c:1262
#5  0x0000555555600e34 in maildir_read_dir (ctx=0x55555624c750) at maildir/mh.c:1281
#6  0x0000555555600e34 in maildir_mbox_open (ctx=0x55555624c750) at maildir/mh.c:2278
#7  0x00005555555bd9cf in mx_mbox_open (path=path@entry=0x7fffffff9ec0 "/home/zack/Maildir/*censored*", flags=0) at mx.c:313
#8  0x00005555555bdbef in mx_mbox_open (path=path@entry=0x7fffffff9ec0 "/home/zack/Maildir/*censored*", flags=<optimized out>) at mx.c:243
#9  0x0000555555589dc4 in main_change_folder (menu=0x5555558b17a0, op=op@entry=223, buf=buf@entry=0x7fffffff9ec0 "/home/zack/Maildir/*censored*", oldcount=oldcount@entry=0x7fffffff9d30, index_hint=index_hint@entry=0x7fffffff9d34, buflen=1024) at curs_main.c:617
#10 0x00005555555911a9 in mutt_index_menu () at curs_main.c:2246
#11 0x0000555555572a70 in main (argc=<optimized out>, argv=0x7fffffffddf8, envp=<optimized out>) at main.c:1214
(gdb) 
```

flatcap:
Hmm... I tried your branch and it crashed in `malloc` -- something very strange is going on.
Good news, though.  It's not your fault.

I've rebased your commit onto `neomutt-20180716` (the latest release) and it works correctly.
See branch [devel/notmuch-thread-wo-vfolder](https://github.com/neomutt/neomutt/tree/devel/notmuch-thread-wo-vfolder)

I'll have a closer look, soon.

zacchiro:
I can confirm that the segfault reported above is not due to this PR, but affects more generally current HEAD. To reproduce:

- open a mailbox
- do a notmuch search that will find some messages in that mailbox
- (as a result you will jump to a new vfolder)
- open the same mailbox again with `Ctrl+O` in the sidebar
- → segfault

Austin-Ray:
I played with this PR today to see if it broke anything. Below is the `.neomuttrc.min` I used (since I normally use `vfolders`.)
```bash
# enable sidebar
set sidebar_visible

# setup maildir
set spoolfile = /home/aray/.mail/austin@austinray.io/INBOX
set folder    = /home/aray/.mail
mailboxes     =austin@austinray.io/INBOX

# configure necessary shortcuts.
bind index,pager \Co sidebar-open
bind index,pager \Ce entire-thread
bind index,pager \Cx vfolder-from-query

# vim: ft=neomuttrc
```

This PR seems to be in good shape. I did not find anything wrong with it. The broken `sidebar_open` is not your fault, and `change-folder` works without issue. It'll be up to @flatcap to determine if we should fix `sidebar_open` in this PR or open another for it.

The only thing I can nitpick is the naming of `mid_offset` may mislead some readers. I thought it was middle offset at first instead of message-id offset. Perhaps use `msg_id_offset` instead?

zacchiro:
On Sat, Oct 06, 2018 at 10:08:04PM -0700, Austin Ray wrote:
> This PR seems to be in good shape. I did not find anything wrong with it.

Thanks for your time looking into this.

> The only thing I can nitpick is the naming of `mid_offset` may mislead
> some readers. I thought it was middle offset at first instead of
> message-id offset. Perhaps use `msg_id_offset` instead?

Good point, I didn't think of "mid[dle]", but it was indeed confusing.
I've updated the PR to follow your variable naming suggestion.

flatcap:
Thanks.

After some tests with `git bisect` and the Address Sanitizer, I found and fixed the re-open problem -- hopefully :-)

zacchiro:
On Mon, Oct 08, 2018 at 05:32:39AM -0700, Richard Russon wrote:
> After some tests with `git bisect` and the Address Sanitizer, I found
> and fixed the re-open problem -- hopefully :-)

I confirm that the re-open problem is fixed for me too.

Thanks (for the fix and for the PR merge).

Cheers

Issue #721, opened 2017-08-17
sidebar and pager don't display the same number of messages when using  nm_query_type set to threads.


d-k-c:
* Expected behavior

`%m` in `pager_format` and `%S` in `sidebar_format` display the same value

* Actual behavior

They display different values. %S gets its value with "notmuch count", while %m gets its value from the number of results that "notmuch show" returns. These two values are different if nm_query_type is set to threads.

* Steps to reproduce

- Configure pager to display %m and sidebar to display %S
- set nm_query_type to threads
- find search terms such that `notmuch count <search-terms>` and `notmuch show --entire-threads <search-terms> | grep "message{" | wc -l` don't return the same thing
- create a virtual-mailbox based on this query.

d-k-c:
poking @guyzmo as commit 771612750 is the cause of this bug

Austin-Ray:
fixed in #1355

Austin-Ray:
Fix was reverted in b71ac2b due to poor performance

PR #1355, opened 2018-10-11
notmuch: count messages in queried threads

Austin-Ray:
This makes `sidebar_format`'s `%S` display the same value as
`pager_format`'s `%m`.

`count_query` did not handle counting all messages in `type=threads`
queries. It returned the number of messages matching the `query=`
qualifier. Refactored `count_query` into two methods handling
`type=threads` and `type=messages`, respectively.

**Screenshots (if relevant)**
Compare the right-most number of the sidebar for `inbox threads` to the pager. 
Before:
before
After:
after

#721

### DIFF
--- a/notmuch/mutt_notmuch.c
+++ b/notmuch/mutt_notmuch.c
@@ -1731,12 +1731,66 @@ done:
 }
 
 /**
+ * count_query_messages - Count the number of messages in all queried threads
+ * @param query Executed query
+ * @retval num Number of messages
+ */
+static unsigned int count_query_thread_messages(notmuch_query_t *q)
+{
+  unsigned int count = 0;
+  notmuch_threads_t *threads;
+  notmuch_thread_t *thread;
+
+#if LIBNOTMUCH_CHECK_VERSION(5, 0, 0)
+  notmuch_query_search_threads(q, &threads);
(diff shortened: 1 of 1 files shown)


notmuch/mutt_notmuch.c
```diff
@@ -1730,13 +1730,67 @@ static int rename_filename(struct NmMboxData *mdata, const char *old,
   return rc;
 }
 
+/**
+ * count_query_messages - Count the number of messages in all queried threads
```
flatcap: s/count_query_messages/count_query_thread_messages/

notmuch/mutt_notmuch.c
```diff
@@ -1730,13 +1730,67 @@ static int rename_filename(struct NmMboxData *mdata, const char *old,
...
+ * @param query Executed query
+ * @retval num Number of messages
+ */
+static unsigned int count_query_thread_messages(notmuch_query_t *q)
+{
+  unsigned int count = 0;
+  notmuch_threads_t *threads;
+  notmuch_thread_t *thread;
```
flatcap: Please initialise all pointers to `NULL`.

Austin-Ray:
The PR has been tidied.

PR #1365, opened 2018-10-17
Try adding status indicator R for my address in Reply-To

nobrowser:
* **What does this PR do?**

It adds a new "R" flag for the 3rd position of the index %Z field, for when user's own address is in the Reply-To header unless none of the other flags for that field is applicable.

* **Motivation**

Some mailing lists now do the following transformation of the headers: rewrite the From header to the list address, and add a Reply-To header with the poster address.  This feature helps highlight my own posts in folders dedicated to such lists.  Of course if you are _subscribed_ to the list (in the Mutt sense) you'll get the "L" flag instead, but in that case you don't care about hightlighting your posts anyway.

     for that)
Not yet.  I'll add that once I see it compiles with travis.

Hopefully

Yes, I took care of that

Yes

None, I'm just throwing this out there :p

### DIFF
--- a/doc/manual.xml.head
+++ b/doc/manual.xml.head
@@ -1803,6 +1803,10 @@ color sidebar_divider   color8  default     <emphasis role="comment"># Dark grey
                 <entry>L</entry>
                 <entry>message is sent to a subscribed mailing list</entry>
               </row>
+              <row>
+                <entry>R</entry>
+                <entry>message has your address in the Reply-To field</entry>
+              </row>
             </tbody>
           </tgroup>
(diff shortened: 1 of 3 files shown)


nobrowser:
And I have updated the manual now, so all check-in requirements are satisfied.

flatcap:
Oh!  That's a pleasant surprise, after you'd asked for advice on the **Mutt** mailing list.
(I _did_ notice your 'User-Agent' :-)

The PR looks good to me.
We'll just wait a couple of days to see if anyone has any comments.

Issue #1364, opened 2018-10-15
Segfault in SSL_write


l29ah:
20180716-513-573638

```
#0  0x00007fc0e07c960a in SSL_write () from /usr/lib64/libssl.so.1.0.0
#1  0x000056243da46b7d in ssl_socket_write (conn=<optimized out>, buf=<optimized out>, count=<optimized out>)
    at conn/ssl.c:1364
#2  0x000056243da44cbe in mutt_socket_write_d (conn=0x56243ecd4150, buf=0x56243eca59a0 "a0000 LOGOUT\r\n", len=14, 
    dbg=dbg@entry=2) at conn/socket.c:183
#3  0x000056243da378b4 in cmd_start (adata=0x56243ecd0540, cmdstr=cmdstr@entry=0x56243da7c84a "LOGOUT", 
    flags=flags@entry=0) at imap/command.c:206
#4  0x000056243da37911 in imap_cmd_start (adata=<optimized out>, cmdstr=cmdstr@entry=0x56243da7c84a "LOGOUT")
    at imap/command.c:1075
#5  0x000056243da39862 in imap_logout (adata=adata@entry=0x56243ecd43d8) at imap/imap.c:1137
#6  0x000056243da39938 in imap_logout_all () at imap/imap.c:759
#7  0x000056243d9d038d in main (argc=1, argv=<optimized out>, envp=<optimized out>) at main.c:1218
```

l29ah:
Another, similar one on 20180716-516-6047ee:
```
#0  0x00007ff13f0c75b5 in SSL_write () from /usr/lib64/libssl.so.1.0.0
#1  0x0000562eac13dbe3 in ssl_socket_write (conn=<optimized out>, buf=<optimized out>, count=<optimized out>) at conn/ssl.c:1364
#2  0x0000562eac13bd24 in mutt_socket_write_d (conn=0x562eac4db910, buf=0x562eac5379d0 "a0000 IDLE\r\n", len=12, dbg=dbg@entry=2) at conn/socket.c:183
#3  0x0000562eac12e91a in cmd_start (adata=adata@entry=0x562eac4d7d00, cmdstr=cmdstr@entry=0x562eac1730d3 "IDLE", flags=flags@entry=8) at imap/command.c:206
#4  0x0000562eac12e994 in imap_cmd_idle (adata=adata@entry=0x562eac4d7d00) at imap/command.c:1364
[...]

PR #1370, opened 2018-10-26
Do not use disconnected socket.

stoeckmann:
If an IMAP connection was never set up, do not try to write a LOGOUT
command through that connection or to close that socket.

Both results in segmentation faults.

How to reproduce:

- Connect to a server with an invalid or self-signed TLS certificate
- Press "q" to cancel the connection
- Press "q" again to leave neomutt
- neomutt crashes with a segmentation fault

This should at least partly fix https://github.com/neomutt/neomutt/issues/1364


flatcap:
Welcome to NeoMutt, @stoeckmann.
Thanks for the PR, it looks good.  I'll give it a test later.

flatcap:
Lovely.  Thanks a lot :-)

PR #1381, opened 2018-10-31
notmuch: count flagged messages

Austin-Ray:
`nm_nonctx_get_count(...)` did not count flagged messages. This means
they could not display flagged message counts in the sidebar _unless_ it
was the active mailbox (handled by a different function.)

This commit introduces NmFlaggedTag, mirrored from NmUnreadTag, and will
query the database for flagged messages.

### DIFF
--- a/browser.c
+++ b/browser.c
@@ -943,7 +943,7 @@ static int examine_vfolders(struct Menu *menu, struct BrowserState *state)
   {
     if (nm_path_probe(np->m->path, NULL) == MUTT_NOTMUCH)
     {
-      nm_nonctx_get_count(np->m->path, &np->m->msg_count, &np->m->msg_unread);
+      nm_nonctx_get_count(np->m);
       add_folder(menu, state, np->m->path, np->m->desc, NULL, np->m, NULL);
       continue;
     }
--- a/doc/manual.xml.head
+++ b/doc/manual.xml.head
@@ -14556,6 +14556,12 @@ virtual-mailboxes "My INBOX" "notmuch://?query=tag:inbox"
             limit is nm_db_limit.
           </para>
           <para>
+            Due to a limitation with <literal>libnotmuch</literal>, unread and
+            flagged message count may be inaccurate with limit statements.
+            <literal>libnotmuch</literal> cannot return a specific tag count
+            within the first X messages of a query.
+          </para>
+          <para>
             <emphasis role="bold">type=&lt;threads|messages&gt;</emphasis>
           </para>
           <para>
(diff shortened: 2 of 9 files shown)


flatcap:
That looks good.
I have some questions/suggestions.

`nm_nonctx_get_count()` has three callers:

- All 3 callers ask for all 3 counts.
  It is worth simplifying the function to pass `Mailbox *m` and populate all three?

- 2 callers use `m->path`, 1 uses `m->realpath`
  Is there a distinction here?
  If not, can we just pass `Mailbox *m`?

In `nm_nonctx_get_count()` we apply a limit to ALL messages.
Would we ever limit the NEW, or FLAGGED messages?
Or are they counts of emails that we aren't going to retrieve?

In the doxygen header for `count_query`, please can you fix my typo: `numner` :-)

Austin-Ray:
> nm_nonctx_get_count() has three callers:

Yes, we could replace it with `Mailbox`. 

I ultimately want to remove the calls in `sidebar.c` and `browser.c` and use `mailbox.c` as a source of number counting. `sidebar.c`'s call affects performance if you have a large number of `virtual-mailboxes` since counting gets expensive with 50+ mailboxes.

> Would we ever limit the NEW, or FLAGGED messages?
> Or are they counts of emails that we aren't going to retrieve?

Applying the limit is a safe idea to make sure we never exceed the total message count. Our numbers could be inaccurate.

Since `libnotmuch` does not contain a method to count the number of unread/flagged messages in the first X messages, our numbers can be off. We work around this by counting when a mailbox is opened to ensure accurate numbers for the _opened_ mailbox (I would like to remove calls to this method, it's for updating `Context` and violates single point of truth design philosophy.)

e.g. Consider a notmuch query that returns 1,000,000 emails with 500,000 unread and 500,000 flagged. If we apply the limit for 20,000 emails, `nm_nonctx_get_count(...)` still reports 500,000 unread and 500,000 flagged, but we recount the 20,000 emails in to ensure correct unread/flagged numbers for the _opened mailbox_. Applying the limit will make sure `nm_nonctx_get_count(...)` will return 20,000 unread and 20,000 flagged, but the mailbox may have fewer.

This is a limitation of `libnotmuch` and our subsystem (constrained by performance.) A lot of our code functions on the assumption of no limit.

Austin-Ray:
@flatcap: I pushed a new commit to the branch. This commit may be slightly out of scope for this PR, but it builds on first commit. The second commit removes the special case for notmuch. After the architectural changes, it is no longer necessary.

Austin-Ray:
@flatcap I removed my second commit (it's covered by my other PR) and documented the limitation with limit statements regarding potentially inaccurate unread and flagged counts.

PR #1384, opened 2018-11-02
notmuch: remove sidebar and browser special cases

Austin-Ray:
**What does this PR do?**

Removes the separation between mailboxes and virtual-mailboxes in the sidebar and browser. virtual-mailboxes are the ugly stepchild in NeoMutt, and there is arbitrary separation between them. This branch is working towards making virtual-mailboxes no different than any other.

Altered:
 - Behavior of `%f`. Use path instead of description. 
   `%i` display descriptions.

Removed:
 - `sidebar-toggle-virtual` is created an unnecessary separation between mailboxes and virtual-mailboxes. It's still included as a command (so people don't get config warnings), but does nothing. It might be a good idea to remove it so people don't execute a useless keybinding.

### DIFF
--- a/browser.c
+++ b/browser.c
@@ -63,9 +63,6 @@
 #ifdef USE_NNTP
 #include "nntp/nntp.h"
 #endif
-#ifdef USE_NOTMUCH
-#include "notmuch/mutt_notmuch.h"
-#endif
 #ifdef USE_POP
 #include "pop/pop.h"
 #endif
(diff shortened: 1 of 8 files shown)


Austin-Ray:
I rebase-edited `sidebar: unify virtual-mailboxes` to remove `notmuch.h` from `curs_lib.c` since it was no longer needed.

PR #1390, opened 2018-11-05
Delete temporary file upon signal reception

0x747972:
See commit message.


flatcap:
@gahr and I had a long discussion yesterday on IRC about temporary files.
We decided that the following were useful:
- pausing before deleting files
- having an option to **not** delete the file

This would simplify the case for new users and allow power users to control the lifespan of the temp files.

This PR is probably still needed, but I'll need to merge @gahr's changes first.

Issue #1383, opened 2018-11-01
Storing Paths in Mailboxes


flatcap:
This issue discusses how we use (or could better use) paths in NeoMutt.
It raises some questions.
Your thoughts are welcome.

## Source

Paths are user-specified:
- 'c' `<change-folder>`
- `$spoolfile`
- `mailboxes` command

They may be abbreviated:
- `!!` - Previous mailbox
- `-` - Previous mailbox
- `+` - Relative to `$folder`
- `=` - Relative to `$folder` 
- `!` - `$spoolfile`
- `<` - `$record`
- `>` - `$mbox`
- `^` - Current mailbox
- `~` - User's home directory
- `~person` - `person`s home directory
- `@alias` - Default save mailbox for user `alias`

They may contain backend 'decorations', e.g.
- `imap://host:port/folder`
- `pop://user@host/folder`

## Storage

The Mailbox has stores two paths:
- `path`
- `realpath`

I suggest that we rename `realpath` to `canon` and use them consistently:

- `path`
  The mailbox location, relative to the host.
  This has no decoration.

- `canon`
  The canonical reference for this mailbox.
  Uniquely decorated with any non-standard attributes (excluding any passwords)

**Notes**:
- Mailbox comparison is done using the `canon` path and is delegated to the backend.
- The `canon` path may also be stored as separate components

## Display

NeoMutt may show the user different versions of the Mailbox path:

- `path` - '/all-mail/folder'
- `$folder`-abbreviated path - '+folder'
- `canon` - 'imaps://USER@HOST:PORT/all-mail/folder'
- `desc` - A user-supplied description (using `named-mailboxes`)

### maildir, mbox

- `path` - Full filesystem path
  Path has been tidied of duplicate `/`s or `/./` combinations, etc

- `canon` - Full filesystem path
  This path has also had any symlinks resolved.

### compressed

- `path` - Full filesystem path of compressed file
  Path has been tidied of duplicate `/`s or `/./` combinations, etc
  This is what's displayed to the user.

- `canon` - Full filesystem path of the temporary (uncompressed) file
  This path has also had any symlinks resolved.

### imap

- `path` - Relative path, e.g. '/folder'
  Path has been tidied of duplicate `/`s or `/./` combinations, etc

- `canon` - Fully decorated mailbox path, e.g. `imaps://USER@HOST:PORT/folder`
  Path has been tidied of duplicate `/`s or `/./` combinations, etc

**Notes**:
[...]

PR #1393, opened 2018-11-06
mxapi: mx_path_canon should not double / uri

sileht:
At least for imap if "folder=imap://example.com" and mailboxes contains "=Foo"

mx_path_canon will construct imap://example.com//Foo

This change ensures we do not add the extra /.

### DIFF
--- a/mx.c
+++ b/mx.c
@@ -1462,8 +1462,16 @@ int mx_path_canon(char *buf, size_t buflen, const char *folder, enum MailboxType
     }
     else if ((buf[0] == '+') || (buf[0] == '='))
     {
-      buf[0] = '/';
-      mutt_str_inline_replace(buf, buflen, 0, Folder);
+      size_t folder_len = mutt_str_strlen(Folder);
+      if ((folder_len > 0) && (Folder[folder_len - 1] != '/'))
+      {
+        buf[0] = '/';
+        mutt_str_inline_replace(buf, buflen, 0, Folder);
+      }
+      else
+      {
+        mutt_str_inline_replace(buf, buflen, 1, Folder);
+      }
     }
     else if ((buf[1] == '/') || (buf[1] == '\0'))
     {


mx.c
```diff
@@ -1462,8 +1462,16 @@ int mx_path_canon(char *buf, size_t buflen, const char *folder, enum MailboxType
     }
     else if ((buf[0] == '+') || (buf[0] == '='))
     {
-      buf[0] = '/';
-      mutt_str_inline_replace(buf, buflen, 0, Folder);
+      size_t Folderlen = mutt_str_strlen(Folder);
```
gahr:
```suggestion
      size_t folder_len = mutt_str_strlen(Folder);
```

mx.c
```diff
@@ -1462,8 +1462,16 @@ int mx_path_canon(char *buf, size_t buflen, const char *folder, enum MailboxType
...
+      if ((Folderlen > 0) && (Folder[Folderlen - 1] != '/'))
```
gahr:
```suggestion
      if ((folder_len > 0) && (Folder[folder_len - 1] != '/'))
```

gahr:
Notice that if you do change-folder, =Foo, and press tab, it'll correctly resolve to $folder/Foo without double slashes. It might be worth investigating how *that* is done.

sileht:
I just found this code disabled https://github.com/neomutt/neomutt/blob/master/imap/imap.c#L3013

Maybe related, I will check that later.

flatcap:
Thanks @sileht.
All this code is still a little wobbly.

> I just found this code disabled [/imap/imap.c@`master`#L3013](https://github.com/neomutt/neomutt/blob/master/imap/imap.c#L3013)

Yes, I decided recently that `$folder` was a front-end thing (notes in #1383).

sileht:
I wonder why `mx_path_canon()` uses `Folder` to pre-canonicalise `buf` and then we call `ops->path_canon()` with `folder` ?

sileht:
(Noob question) Do we have case where `folder `!= `Folder` ?

flatcap:
> I wonder why `mx_path_canon()` uses `Folder` to pre-canonicalise `buf` and then we call `ops->path_canon()` with `folder` ?

Ah, that shows my first attempt to remove `$folder`.
That parameter can probably go, now (and `path_pretty()` probably needs a rethink).

> Do we have case where `folder `!= `Folder` ?

I don't think so.

PR #1396, opened 2018-11-08
imap_conn_find removal

sileht:
All code using imap_conn_find is currently broken.

ImapPassive is set by default making `imap_conn_find()` always returning NULL.

So only `imap_mbox_open()` is currently working as this is the only function that use the new `imap_conn_find2()`.

This change aims to completely remove `imap_conn_find()`

`imap_conn_find2()` is renamed `imap_login()` as it prepare the imap connection and does the reverse as `imap_logout()`.

Some new private function have been introduced:

* `imap_prepare_mailbox()` to share to parsing of the `Mailbox` and the setup of `ImapMbox` and `struct ImapAccountData`.
* `imap_ac_data_find()` to find a valid and populated `ImapAccountData` for an `ImapMbox`.


sileht:
Pull-request updated, HEAD is now 5c3e01d3b68c9ee4735f59a59c108b342313c1ca

PR #1400, opened 2018-11-09
 Add mutt_str_startswith API

gahr:
* **What does this PR do?**

Add a `mutt_str_startswith` API to be used instead of the common pattern `mutt_str_strncmp(string, prefix, strlen(prefix))`. Further commits for changing these will come.


Austin-Ray:
It seems unusual to return a `size_t` when the function's name and description suggests a `bool`.

gahr:
True. However, you *can* use this function’s return value as it is was a boolean, and the added benefit is that we don’t throw away useful information. This is used in some of my changes to avoid 1) computing strlen on the prefix, which doubles the runtime complexity, or 2) hardcoding the prefix length.

Austin-Ray:
I can understand that use-case.

At first glance, it doesn't look like `mutt_str_startswith(...)` handles the case where the `prefix` is longer than `str`, and it took me a minute to figure it out. Could add a comment above the return statement explaining how it handles it?

Perhaps modifying it to be along the lines of:
```C
// If prefix is valid at this point, it is longer than str so return 0.
// Otherwise, calculate the length of the prefix.
return *prefix ? 0 : (prefix - saved_prefix);
```

I only ask because pointer arithmetic is a bit arcane to me (it becomes less so everyday!)

flatcap:
I have:
- rebased onto master
- made sure all of the commits compile
- clang-formatted everything
- verified the changes

I've found five changes where I think the logic has become inverted,
and one in `browser.c` where I think we still need the prefix check.
See: f0529d4bf

The final commit, dc34e9466, is some subjective tidying (open to debate).
I'm not very fond of `if (!*NONULL(str)`.

browser.c
```diff
@@ -866,7 +866,7 @@ static int examine_directory(struct Menu *menu, struct BrowserState *state,
       if (mutt_str_strcmp(de->d_name, ".") == 0)
         continue; /* we don't need . */
 
-      if (mutt_str_startswith(de->d_name, prefix, CASE_MATCH))
+      if (prefix && !mutt_str_startswith(de->d_name, prefix, CASE_MATCH))
```
gahr: You are right, but I think we need to restore the full `prefix && *prefix`

PR #1405, opened 2018-11-10
imap: make imap_mailbox_check aware of Mailbox

sileht:
imap: make imap_mailbox_check aware of Mailbox

With multiple imap account not loaded at startup time, when we switch to
another account, stats are almost not updated. If check_stats if true at
this time by change, mutt crash. Because we unexpectly update the
msg_unread of Context->mailbox.

Imap is the only backend that check stats is done for all backends at
once.

This change makes it more like other backends, and it make all imap
mailboxes stats to be loaded on next check_stats=true without crash.


sileht:
This have been merged part of : https://github.com/neomutt/neomutt/pull/1415

PR #1406, opened 2018-11-10
Always do mailbox_check() the first we see a mailbox

sileht:
Always do mailbox_check() the first we see a mailbox

When multiple accounts are used, stats are all 0 until the next
MailCheckStatsInterval is reached.

This change does an initial check when we a mailbox for the first time.


mailbox.h
```diff
@@ -107,6 +107,7 @@ struct Mailbox
   bool changed : 1;   /**< mailbox has been modified */
   bool readonly : 1;  /**< don't allow changes to the mailbox */
   bool quiet : 1;     /**< inhibit status messages? */
+  bool first_check_stats_done : 1; /**< true when the check have been done at least on time */
```
flatcap: Do we need a new bool?
Could we just check the `M->stats_last_checked` timestamp?

flatcap: OK, fair enough.

mailbox.c
```diff
@@ -774,7 +774,9 @@ int mutt_mailbox_check(int force)
   struct MailboxNode *np = NULL;
   STAILQ_FOREACH(np, &AllMailboxes, entries)
   {
-    mailbox_check(np->m, &contex_sb, check_stats);
+    mailbox_check(np->m, &contex_sb, check_stats || (!np->m->first_check_stats_done && MailCheckStats));
+    if (!np->m->first_check_stats_done)
+      np->m->first_check_stats_done = true;
```
gahr: As a matter of preference, I would just always assign no matter what the previous value is..

PR #1407, opened 2018-11-10
mxapi: introduce check_stats

Austin-Ray:
Extend MxApi to have a common method for checking a mailbox's statistics. This is a proof-of-concept commit as of opening the PR, and only the notmuch backend has been modified. I hope to gather feedback on the way to unify all backend check stats calls.


sileht:
We are connected I proposed https://github.com/neomutt/neomutt/pull/1405 2 hour ago to be able to do the same for imap :)

Austin-Ray:
> We are connected

Great minds think alike :)

I'm porting the other backends currently, but hadn't touched IMAP yet. The local backends have some extra parameters that need to be passed (and aren't included in 65da02c). I'll push the local backend changes in a minute and see if the proposed call works for the IMAP backend.

Austin-Ray:
@sileht I cherry-picked your commit from #1405 and added `check_stats(...)`. I noticed that there are multiple stat checking methods within `imap/imap.c`. 

Do you have to know if they should be merged? I saw `imap_check_mailbox()` (which is why i added `force` in my initial commit)

flatcap:
This is a great idea (both of you).

The old code just had a list of Mailboxes -- it had no real clue how they were related.
For efficiency, `mutt_mailbox_check()` called `imap_mailbox_check()` so it could group together some of the network traffic.  Now we have Accounts, I think we can be smarter still.

I suggest two mxapi check functions:

- `mbox_check_stats(M, flags)` - Check the stats of a single Mailbox

- `ac_check_stats(A, flags)` - Check the stats of all Mailboxes on an Account
  I think Imap and Maildir could perform some optimisations.
  Other backends could simply call `mbox_check_stats()` for each Mailbox.

flatcap:
`mx.c` shouldn't need to know which backends are local.  The `struct stat` for the local filesystems was a bit of a cheat -- it didn't really get us much.

Soon, all the checks will be performed on Mailboxes, so we'll already know the type and have an Account.  

I haven't _really_ delved into POP and NNTP.
POP shouldn't be too hard.  It only has one Mailbox per Account.
I'm not sure how/when NNTP checks for updates.

Austin-Ray:
Okay so we'll need to refactor MBOX and MMDF to not use `struct stat`. It should be possible since MAILDIR and MH do not use one.

Is this  a complete list of backends?

- notmuch
- mbox
- mmdf
- maildir
- mh
- imap
- pop
- nntp

flatcap:
Compressed relies on some tricky behaviour to work.
I'm not sure how to make it play nicely with the mxapi.

Austin-Ray:
How do we want `mx_ac_check_stats(...)` to work? 
Should it loop through calling `mx_mbox_check_stats(...)`?
How should errors be propagated? Return on first error?

flatcap:
> How do we want `mx_ac_check_stats(...)` to work?

My original thought was that `mx.c` would use `.ac_check_stats()` where possible, if not it'd use `.mbox_check_stats()`, but we'd need to mark which Mailboxes we'd already visited.

I think we should insist that all backends implement both methods,
so `mx_ac_check_stats()` reduces to just `FOREACH(AllAccounts)` call `.ac_check_stats()`.

> How should errors be propagated? Return on first error?

Good question.  I suppose there are several levels of broken.
- Account-level, e.g. bad credentials
- Mailbox-level, e.g. non-existent Mailbox
- General warnings

Errors will stop us recursing, but we should carry on with other mailboxes if possible.

My Config code passes a `Buffer` to every function for error messages, we might want to consider something similar.
For now, `mutt_{error,warning,message}()` should be fine.

> Also, what flags do we want?

Don't know.  It just felt like we might need some flags there :-)
The current code has a `force` parameter, but I don't think that's appropriate for the backends.
We can always add them later.

Austin-Ray:
I haven't been able to get around to working on this PR due to work (It's a new job.) I'll work a little on this over the weekend.

Austin-Ray:
Rebased onto master

Austin-Ray:
Rebased onto master (and made sure to avoid clobbering @flatcap's changes)

Austin-Ray:
@flatcap I do not know how to approach removing the `struct stat` for the `mbox` and `mmdf` backends. Any suggestions?

flatcap:
> how to approach removing the `struct stat`

Simply drop it.  There's no need for the MXAPI to do it.

In the past, we would have a list of paths to work with.
The MXAPI ran `stat` on each one to help determine the mailbox type.
For efficiency we passed the stat info to the backend.
This avoided each backend having to run `stat` only to find that it wasn't their file.

Now, we have a `Mailbox`, so we know the type and can pass control straight to the backend.
The backends can perform `stat` if they need it.

Austin-Ray:
> Simply drop it. There's no need for the MXAPI to do it.

Will do.

> The backends can perform stat if they need it

The `mbox` and `mmdf` backends compare against a `stat` in `Mailbox`. It seems quite unusual and tightly coupled.

On November 26, 2018 5:19:49 AM EST, Richard Russon <notifications@github.com> wrote:
>> how to approach removing the `struct stat`
>
>Simply drop it.  There's no need for the MXAPI to do it.
>
>In the past, we would have a list of paths to work with.
>The MXAPI ran `stat` on each one to help determine the mailbox type.
>For efficiency we passed the stat info to the backend.
>This avoided each backend having to run `stat` only to find that it
>wasn't their file.
>
>Now, we have a `Mailbox`, so we know the type and can pass control
>straight to the backend.
>The backends can perform `stat` if they need it.
>
>-- 
>You are receiving this because you were mentioned.
>https://github.com/neomutt/neomutt/pull/1407#issuecomment-441588129

-- 

flatcap:
I've rebased over my Maildir changes.

Ah.  `M->stats_last_checked` doesn't belong there any more.
I suggest creating an `MboxMboxData` -- private data for a `Mailbox`
(I might rejig the naming at some point :-)

Compare:
- https://github.com/neomutt/gfx/blob/master/account/mbox.svg
- https://github.com/neomutt/gfx/blob/master/account/maildir.svg

Austin-Ray:
@flatcap I have (hopefully) fixed the history. I have a copy of the branch you pushed located at https://github.com/Austin-Ray/neomutt/tree/dev/mxapi/check-count-flatcap for comparison.

I implemented a proper `mbox_check_stats()` function. Could you take a look at 97143bb and tell me if my solution for `mbox` and `mmdf` using a `struct stat` works? 

I based it off of `mailbox_check` gets its `struct stat`. I figure it should work, but I do not use `mbox` or `mmdf` so I could not give it proper testing.

flatcap:
I haven't made any function changes.
The first commit just moves a function and drops a couple of unnecessary prototypes.
The second renames the mxapi functions to match the others.

flatcap:
The code looks good now, but the commits are a bit of a jumble:
- functions are introduced, then **later** renamed or moved
- a831ff33 adds mbox support **and** removes the old mxapi

I'd _prefer_ to see the commits refactored to reduce the churn (no code changes are required).

I'd normally do this by:
1. squash all the commits into one big commit
2. create a new branch
3. repeat:
    1. cherry-pick the changes using `git checkout -p old-branch`
    2. test build
    3. commit changes

Austin-Ray:
@flatcap I definitely went into this without have a complete plan so the history is completely jumbled. I had intended to clean it up this weekend, but I was roped into some family events. I aim to have the history cleaned up before Wednesday.

flatcap:
No worries; thanks; no rush (respectively :-)

Austin-Ray:
@flatcap I corrected the history (pretty trivial to do with `git rebase`; thought it would be harder.) I only modified the history to make `mbox_check_stats(...)` appear in one commit. I've check to minor changes just to make sure I didn't lose any of them.

There are three back-ends that are not using this API: `nntp`, `pop`, and `compressed` (though it delegates stat checking to `mbox` (i think that's the correct backend)).

flatcap:
>  I corrected the history ...

It's better, but not quite there.
The first commit introduces `imap_mailbox_check()` and the fourth renames it to `imap_mbox_check_stats()`.

> There are three back-ends that are not using this API: `nntp`, `pop` and `compressed`

Yes, that makes sense.

Austin-Ray:
> The first commit introduces imap_mailbox_check() and the fourth renames it to imap_mbox_check_stats().

I thought this PR would be squashed into one commit then pushed onto master. Would you like something different?

flatcap:
Here's what I had in mind:
- Each commit is an independent piece of work
- All the commits compile

If we changed our mind about one of the backends, we could revert a single commit without affecting the others.

PR #1408, opened 2018-11-10
Also delete account if needed during unmailboxes

sileht:
Also delete account if needed during unmailboxes

When we do:

    mailboxes imaps://foo
    unmailboxes *
    mailboxes imaps://foo

* account->mailboxes is not cleaned and have pointer to freed mailboxes
* account is still there even if not used

This changes cleanup account->mailboxes and account if needed.

### DIFF
--- a/account.c
+++ b/account.c
@@ -48,7 +48,48 @@ void account_free(struct Account **a)
     return;
 
   if ((*a)->adata)
-    (*a)->free_adata((void **) a);
+    (*a)->free_adata(&(*a)->adata);
 
   FREE(a);
+}
+
+/**
+ * account_remove - Remove an Account from AllAccounts
+ * @param a Account to remove
+ */
+void account_remove(struct Account *a)
+{
+    struct Account *np = NULL;
+    TAILQ_FOREACH(np, &AllAccounts, entries)
+    {
+      if (np == a)
+      {
+        TAILQ_REMOVE(&AllAccounts, np, entries);
+        break;
+      }
+    }
+}
+
+/**
+ * account_remove_mailbox - Remove a Mailbox from an Account
+ * @param a Account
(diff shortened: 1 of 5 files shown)


flatcap:
That's good, but I think the code would be better placed in `account.c`, leaving `mx_ac_remove()` as a wrapper.

account.c
```diff
@@ -48,7 +48,45 @@ void account_free(struct Account **a)
...
+
+/**
+ * account_remove - Remove a Account from AllAccounts
+ * @param a Account to remove
+ */
+void account_remove(struct Account *a)
+{
+    struct Account *np = NULL;
+    struct Account *nptmp = NULL;
+    TAILQ_FOREACH_SAFE(np, &AllAccounts, entries, nptmp)
```
gahr: No need to use the safe version, as you're breaking out of the loop after having invalidated the iterator.

account.c
```diff
@@ -48,7 +48,45 @@ void account_free(struct Account **a)
...
+/**
+ * account_remove_mailbox - Remove mailbox from an account
+ * @param a Account
+ * @param m MailBox the mailbox to remove
+ */
+void account_remove_mailbox(struct Account *a, struct Mailbox *m)
+{
+  struct MailboxNode *np = NULL;
+  struct MailboxNode *nptmp = NULL;
+  STAILQ_FOREACH_SAFE(np, &a->mailboxes, entries, nptmp)
```
gahr: Same as above

gahr:
I agree with @flatcap, 

``` diff
- mx_ac_remove(np->m);
+ if (np->m->account)
+   account_remove_mailbox(np->m->account, np->m)
```

or similar, and let `account_remove_mailbox` free itself if the mailbox is the last one.

PR #1411, opened 2018-11-12
Pull request for better-adata-handling

sileht:
imap: simplify getter around ImapAccountData

Since we have accounts and only one connection for imap, we can simply the
code. This does:

* Remove now useless attributes of imap_prepare_mailbox()
* Rename imap_ac_data_find() to imap_adata_find() and move it in util.c aside
  other adata related function.
* Parse path within ImapMbox in imap_adata_find() to remove some common code.
* Create ImapAccountData resource during imap_ac_add()

### DIFF
--- a/imap/browse.c
+++ b/imap/browse.c
@@ -225,13 +225,9 @@ int imap_browse(char *path, struct BrowserState *state)
     // Pick first mailbox connected on the same server
     if (imap_account_match(&mx.account, &mx_tmp.account))
     {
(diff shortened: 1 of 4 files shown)


flatcap:
This is great stuff.  More please.

gahr:
@sileht would it be convenient for you to create these branches in the official neomutt repo instead of your own fork? It would make is simpler for us to rebase / collaborate :)

PR #1412, opened 2018-11-13
Devel/list find const

nobrowser:
While working on the message-id stuff, I found this little buglet in the library routines.  Funny that nobody else got there first :-P

`mutt_list_find` in mutt/list.c doesn't change the list, so the list argument should be const.  Otherwise gcc warns, and if you have -Werror on (as you should), errs.

Oh and btw: I am not at all sure that the optimization in `mutt_list_find` which compares the pointers first is valid.  Can you really compare two arbitrary pointers without getting UB?


gahr:
Nice catch! Re comparing pointers, I think it’s a very reasonable optimization, it’s totally defined behaviour until you try to actually dereference them.

PR #1415, opened 2018-11-13
Pull request for devel/imapmbox-removal

sileht:
This change reduce the usage of ImapMbox and imap_path_parse()

This change introduces `struct ImapMailboxData` to mainly store
the mbox name in all its form.

Next change will use this names instead of parsing again and again the
path.

At this end we should be able to drop ImapMbox and imap_parse_path().

### DIFF
--- a/imap/browse.c
+++ b/imap/browse.c
@@ -385,7 +385,7 @@ int imap_mailbox_create(const char *path)
   char name[LONG_STRING];
   short n;
 
-  adata = imap_adata_find(path, name, sizeof(name));
+  adata = imap_adata_find(path, name, sizeof(name), false);
   if (!adata)
   {
     mutt_debug(1, "Couldn't find open connection to %s\n", path);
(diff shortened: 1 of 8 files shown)


flatcap:
@sileht 
I've rebased this over all the recent changes.
It compiles and it seems to work correctly, but I haven't reviewed the code yet.

sileht:
I have seen only one tiny issue, this mailboxes browser does not always have pretty names.

sileht:
I have fixed the mailboxes browser issue in 6bec235 and I think this also fixes the "new mail notification" that @gahr have seen.

gahr:
Copying from IRC for the record:

I'm seeing a delay when switching to mailboxes view. If I type ":exec change-folder<enter>" there's a delay before the prompt "Open mailbox ('?' for list):" 

This is due to `imap_mailbox_check` now returning 0 on success, instead of the number of mailboxes with new mail. In the caller you do https://github.com/neomutt/neomutt/pull/1415/files#diff-4344ca6dcc0377a492e110b75c01d2d7R375 but `imap_mailbox_check` returning 0 does not imply that the mailbox has new messages. Look at `mutt_mailbox` in `mailbox.c:893`, it uses the return from `mutt_mailbox_check` to figure whether any mailbox has new mail. With your change, `mutt_mailbox_check` will return the total number of mailboxes, even if none of them has new mail

PR #1416, opened 2018-11-14
Devel/alternative ids

nobrowser:
This implements the ~I patterns as previously discussed.  I now agree that the variant where neomutt knows about the external program and executes it to get the message-ids is superior, so that is what I did.

No docs yet, but debugged.  I had some valgrind fun, ARGH!  Does anyone else find it surprising that `mutt_file_read_line` frees its buffer when it hits EOF, even when it didn't allocate it?


nobrowser:
WTH IS GOING ON ???

Why do my changes keep getting mixed up with those of @sileht ??

PR #1418, opened 2018-11-14
Do not create an additional ctx for syncing a mailbox

gahr:
When the additional ctx is closed, the mailbox is free'd too, which causes a crash back in the index.


sileht:
Make sense to me, I will test it.

Issue #1386, opened 2018-11-04
Pattern type for external indexers


nobrowser:
## What would you like to be able to do?

Use indexers other than notmuch, cleanly and efficiently.

## What needs to change in NeoMutt?

Add a new pattern type (for example ~I).  It would take a command line as its argument.  Running the command line should produce a bunch of message-ids on stdout, separated by white space; the meaning of the pattern would be to match mails whose message-id is one of the bunch.

Example:

`  <limit> ~I snarf_mairix_msgids <enter>`

where `snarf_mairix_msgids` is a wrapper script that prompts for a query string, passes it to mairix, reads the "raw" output from mairix i.e. a list of file names, and gets the message ids from the files.

## Have you tried any alternative methods?

Yes of course - I am aware that this is practically a FAQ.  But all the available solutions are horrible.  Please see the latest thread on neomutt-users and see the solution proposed there.  Shudder.

And do not get me started on notmuch and "virtual folders" :-(

nobrowser:
Yes and yes.  Have you seen the thread in neomutt-users I mentioned above?  Ok Ok here is a link:

[mailman archive](http://mailman.neomutt.org/pipermail/neomutt-users-neomutt.org/2018-November/000434.html)

If you do not find the solution quoted there (and that is essentially the _only_ one existing) to be disgusting, your taste is .. different.  In fact its originator (Pasky of kernel/git fame) admits as much on his blog where he posted it.

Austin-Ray:
> Have you seen the thread in neomutt-users I mentioned above

I had not, sorry, but I just read it. I'm not sure I completely understand it (might be due to it being late on my end.)

This would essentially be applying a filter to a mailbox (maildir/mbox/mh/etc.) based on the results of an external indexer. The benefit would be using an actual mailbox instead of creating and opening a new mailbox with a query's results?

What other benefits are there for this method? Are there any drawbacks?

> If you do not find the solution quoted there ... to be disgusting,

It is quite disgusting. Fortunately, neomutt has `vfolder-from-query` for notmuch :)

nobrowser:
I think you got it.

The benefit is being able to use other indexers in a reasonable way.  For me, mairix at least is much faster than notmuch.  And I like to see the results in the normal mutt way, by flagging or limiting but still staying in the physical folder, to see the matched mails in their context.

The drawback is of course feeping creaturitis :-)  That, and using another letter of the alphabet for the new pattern, one of the last few letters available.

[...]

Issue #1180, opened 2018-04-30
Encapsulate the Progress Bar


flatcap:
Looking for something easier?  ([try these](https://github.com/neomutt/neomutt/issues?q=label:"good%20first%20issue"+is:open))

## Description

Wrap up the code for the Progress Bar, reducing the effort for the callers and
unify the logging of progress events.

### Part 1: Review Logging Levels

There are currently 5 debug logging levels, which aren't used consistently.
The new guideline is:
- Level 1-3: General debug info (at the programmer's discretion)
- Level 4: Progress bar, trace info, e.g. "Entering function()"
- Level 5: Data read/written/transferred, or other private info

There are about 65 calls to `mutt_debug()` which need reviewing.
This shouldn't take too long.
For each message decide whether it should be:
- Level 3: General
- Level 4: Progress
- Level 5: Data

### Part 2: Create a LL\_PROGRESS Log Level

Create a new debugging level `LL_PROGRESS` and change the progress bar to use it.

You'll need to alter the logging code, too.
Progress messages should **always** be displayed on screen, but only be written
to file if `$debug_level >= 4`.

Now run `neomutt -d3` with some large folders, etc, and find any other progess
messages that should have their level altered.

Try some large searches, too.

### Part 3: Encapsulate the Progress Bar

Change the code to dynamically `create()` and `destroy()` the Progress object.
This will allow the progress bar to signal when it's done.

Currently, the code is littered with calls to `mutt_window_clearline()` many of
which only exist to clear away the last progress message.

### Future direction

(This isn't required for this issue)

The update frequency for the progress bar is controlled by:
- `net_inc`
- `read_inc`
- `time_inc`
- `write_inc`

It would make sense for this to be handled by the progess bar itself.

flatcap:
> `mutt_expando_format()` takes a `char*` not a `struct Buffer`
> (Is anyone considering to change that?)

Hmm... 500 lines of fragile code, used in over 20 places?
No.  Nobody's had the courage to tackle that.
(I've only standardised the function/variable naming across all callers)

> guess a "good enough" size ... to know if the output was truncated

I don't think truncation is likely; the callers have ample space.
Those formatting a line for the screen are given 1KiB, and those formatting commands are given 8KiB.

> Maybe this should be an Issue?

Hmm... it could certainly do with refactoring, but I think that might require a lot of analysis first.
There are a lot of hard-coded rules and quirks in there.

> I can imagine myself working on it.

[...]

PR #1419, opened 2018-11-15
Pattern type for external indexers

gahr:
#1386 

According to @nobrowser , this is ready for review and merge.


doc/manual.xml.head
```diff
@@ -7304,6 +7304,14 @@ set index_format="%4C %Z %{%b %d} %-15.15L (%?l?%4l&amp;%4c?)%*  %s"
                   <quote>Message-ID</quote> field
                 </entry>
               </row>
+              <row>
+                <entry>~I <emphasis>QUERY</emphasis></entry>
+                <entry>
+                  messages whose <quote>Message-ID</quote> field is included in
+                  the results returned from en external search program, when the program
+                  is run with <emphasis>QUERY</emphasis> as its argument. *****)
```
gahr: Is `*****)` intended?

doc/manual.xml.head
```diff
@@ -7612,6 +7620,11 @@ set index_format="%4C %Z %{%b %d} %-15.15L (%?l?%4l&amp;%4c?)%*  %s"
           pattern with <quote>^</quote>. This example matches all mails which
           only has recipients from Germany.
         </para>
+        <para>
+          *****)
```
gahr: Is `*****)` intended?

doc/manual.xml.head
nobrowser: Yes, that's what I intended.  It refers to a footnote later on.  But maybe I should just inline the text I put in the footnote, especially because it consits of a further cross-reference.  What do you think?

gahr: Oh, I'm fine either way - just wasn't aware of the notation.

doc/manual.xml.head
nobrowser: I thought the footnote was useless, so I folded it back to where it was referenced from.

Really ready now.

nobrowser:
You need to install this little pearl as mairix_filter somewhere in your path:

```
#! /bin/sh -Cefu

mairix -r "$@" | sed -e 's@^@formail -x message-id: <@' | sh

```
Then add this to `.muttrc`:

```
set external_search_command=mairix_filter
```

That should be all, you can now use ~I patterns for limitting and tagging.

flatcap:
OK, it works, but it's hardly pretty.
- Run a shell
- Run `mairix` -\> list of files (which aren't limited to the current folder)
- Run `formail` -\> list of message ids
- Parse lots of text

On the mailing list you described Notmuch as _Rube Goldberg_!
NeoMutt talks to Notmuch, efficiently, via a shared library.

I also noticed that `mairix` can handle mbox files, but `formail` can't.

nobrowser:
But notmuch searches (efficient as they are) are not the same as mutt searches, and drop you into a strange land of a "virtual folder".  That is the core of my distaste for notmuch; the slowness of notmuch indexing only adds to that.

The shell and formail could be avoided if mairix had a mode to output the message-id of the results, but it has no such thing, to my knowledge.  Furthermore, the very point of this interface is that it is generic; it can be used with any indexer or indeed any program that outputs message-ids, and not just with a finite list of "supported" programs.  From this POV the shell is an advantage :)

Austin-Ray:
> OK, it works, but it's hardly pretty.

I agree with @flatcap 

---
Feel free to ignore this defense of `notmuch` (it's my duty as the notmuch subsystem maintainer :P)

> notmuch searches ... not the same as mutt searches, and drop you into a strange land of a "virtual folder".

It's similar to calling a remote server, receiving a list of emails, and displaying them. There's not much magic. I like to think of notmuch as a logical disk volume as opposed to physical volumes like maildir, mbox, etc.

pattern.c
```diff
@@ -237,6 +238,80 @@ static bool eat_regex(struct Pattern *pat, struct Buffer *s, struct Buffer *err)
...
+  mutt_list_clear(&pat->p.multi_cases);
+  pid = mutt_create_filter(cmd_buf.data, NULL, &fp, NULL);
+  if (pid < 0)
+  {
+    mutt_buffer_printf(err, "unable to fork command: %s\n", cmd_buf.data);
+    FREE(&cmd_buf.data);
+    return false;
+  }
+  
+  while ((newline = mutt_file_read_line(oldline, &lsz, fp, NULL, 0)) != NULL) {
```
Austin-Ray: For sake of readability, i think it would be best to move the assignment out of the conditional. It'd result in duplication, but will require less cognitive load to interpret the program's behavior.

pattern.c
```diff
@@ -237,6 +238,80 @@ static bool eat_regex(struct Pattern *pat, struct Buffer *s, struct Buffer *err)
...
+{
+  struct Buffer cmd_buf;
+  struct Buffer tok_buf;
+  int result;
+  FILE *fp = NULL;
+  pid_t pid;
+  char* oldline = NULL;
+  char* newline;
+  char* dupline;
+  char* nows;
```
Austin-Ray: Initialise these pointers to `NULL`

flatcap:
> this interface ... is generic
> the shell is an advantage :)

Yes, that's certainly what makes it useful.
I'm wondering if we can make it a bit smarter and more flexible, while we're at it.

I think that adding parameter substitution would be useful:
- Expand `%s` to the search term
- Expand `%f` to the current folder

pattern.c
nobrowser: Let me just say that I see the tradeoff and I choose the other evil LOL

Maybe if doing it your way would result in other improvements (like saving a variable or 2), I might be swayed.  So I will think about it.

pattern.c
nobrowser: It won't hurt, though it won't help either.  Consider it done.

nobrowser:
@flatcap , that is a very interesting and constructive suggestion, thanks.  I'll think about it.

pattern.c
Austin-Ray: I'm only looking out for myself here :P I __know__ i will skim over the assignment-in-conditional and be confused at some point in the future :)

nobrowser:
btw, [here](https://gist.github.com/nobrowser/3e697e70ae4fcaa6df6fdf4c58438a49) is a perl single process version of the filter.  Presumably a few microseconds faster :-)

flatcap:
> I think that adding parameter substitution would be useful:
>     * Expand `%s` to the search term
>     * Expand `%f` to the current folder

See [`expand_command_str()`](https://github.com/neomutt/neomutt/blob/master/compress.c#L344-L351) and [`compress_format_str()`](https://github.com/neomutt/neomutt/blob/147d518b3faa3b9896b1ade506f8f964393a0e1f/compress.c#L303-L325) for cut'n'paste goodness.

pattern.c
nobrowser: Hi, I pushed an update that addresses the style comments, with the help of the new file iterators.  Please ACK; after that I will go on to the more substantial changes requested by @flatcap .

flatcap:
I've made a few minor changes:
- Rebased onto `master`
- clang-format on `pattern.c` -- a few stray `{`s
- Initialise all pointers to `NULL`

The code looks good.

> @Austin-Ray seems to be on a break

Yes, he's been a bit busy with real life.

> the last few lines of `eat_query()` implement reading all lines from a pipe and storing them in a list; that is a fairly generic task, so maybe it should be another library function?

I see what you mean.
I'd say leave it for now, unless you can find another couple of examples to convert.

flatcap:
Do you have anything more to add?
Or are you ready to merge?

nobrowser:
@flatcap , I still have to work on your expansion syntax suggestion, so do not merge yet.

Thanks for the review and the cleanups.

nobrowser:
Hmm, so I started off adding the interpolations.  I took a look at `mutt_expando_format` and OMG!  Now I have second and third thoughts about this part.
`%s` could be useful if the query string wasn't the last command line argument for some indexer.  But I have not seen an indexer like that, and anyway a wrapper script will be required in most cases and the wrapper can put the query string last.  So I think %s in fact adds very little.
`%f` was meant for efficiency I guess, since the indexer will come back with messages from all folders and neomutt will waste time iterating over the ones from non-current folders.  That is a valid concern, but I don't think it warrants weighing this code down with the expando mess.
So I'm asking to merge this in its current form.  I'll continue thinking about another way to weed down the results to the current folder.

flatcap:
> ... wrapper script will be required in most cases ...
> ... I think %s in fact adds very little

OK

> `%f` was meant for efficiency

Yep

> I don't think it warrants weighing this code down with the expando mess.

Hmm... the compress code needed about 20 lines to support two expandos.

https://github.com/neomutt/neomutt/blob/ad048c4598d3c24437f8e031f46e30bfeba83b19/compress.c#L303-L325

>  I'll continue thinking about another way to weed down the results to the current folder

The only alternative is to add another argument to the command -- this implies a fixed order for the arguments.
The wrapper script could ignore it, if it wished.

> I'm asking to merge this in its current form

OK, will do.

nobrowser:
> Hmm... the compress code needed about 20 lines to support two expandos.

I had gotten as far as writing a version of that callback for our situation.  But then I realized `mutt_expando_format` takes a `char*` as destination, not a `struct Buffer `.  (Is anyone considering to change that?)  So I'd have to guess a "good enough" size, or play a very complex game with a "canary" string of some sort to know if the output was truncated.

Hmm.  Maybe this should be an Issue?  I can imagine myself working on it.

Thanks for getting the feature in.

flatcap:
> `mutt_expando_format()` takes a `char*` not a `struct Buffer`
> (Is anyone considering to change that?)

Hmm... 500 lines of fragile code, used in over 20 places?
No.  Nobody's had the courage to tackle that.
(I've only standardised the function/variable naming across all callers)

> guess a "good enough" size ... to know if the output was truncated

I don't think truncation is likely; the callers have ample space.
Those formatting a line for the screen are given 1KiB, and those formatting commands are given 8KiB.

> Maybe this should be an Issue?

Hmm... it could certainly do with refactoring, but I think that might require a lot of analysis first.
There are a lot of hard-coded rules and quirks in there.

> I can imagine myself working on it.

Can I instead interest you in a different refactoring job: #1180 - Encapsulate the Progress Bar

Several backends use it, all slightly differently, and they all have to decide whether to log or not.
I'd like all the globals (`$net_inc`, `$read_inc`, `$time_inc`, `$write_inc`) wrapped up and all the decisions made in `progress.c`.

Thanks.

nobrowser:
I'll look at #1180 soon.

Also, I have a new PR with the folder filtering change (as you suggested above), and a side fix.

PR #1421, opened 2018-11-16
imap: add missing conn_account free

sileht:
imap: add missing conn_account free


sileht:
Duplicate of https://github.com/neomutt/neomutt/pull/1422

PR #1426, opened 2018-11-17
 imap: move some ImapAccountData attributes to ImapMboxData

sileht:
This moves all mailboxes related data to ImapMboxData.


flatcap:
I didn't know if it was practical to move all the "selected mailbox" data.
Very nice.

PR #1428, opened 2018-11-18
imap: remove imap status

sileht:
ImapStatus doesn't make sense since we have ImapMboxData.

Some attributes are duplicated a synced between strucuture.

This change moves all STATUS informations in ImapMboxData
and remove ImapStatus structure.

mboxcache is also removed because ImapMboxData stay in memory during the
whole Mailbox life, no more need to cache that.


imap/command.c
```diff
@@ -845,91 +853,79 @@ static void cmd_parse_status(struct ImapAccountData *adata, char *s)
...
+#if 0
+  // NOTE(sileht): I don't get this part, I haven't found where cmddata is reused
+  // after that.
   /* caller is prepared to handle the result herself */
   if (adata->cmddata && adata->cmdtype == IMAP_CT_STATUS)
   {
     memcpy(adata->cmddata, status, sizeof(struct ImapStatus));
     return;
   }
+#endif
```
flatcap: I agree, this block can go (and the `IMAP_CT_STATUS` enum value).
It isn't used and I can't find any evidence it was _ever_ used for 'STATUS'.

This means we don't need `enum ImapCommandType` any more
We might want to rename `cmddata` as it's only used by 'LIST'.

PR #1429, opened 2018-11-18
imap: move flags in ImapMboxData

sileht:
Flags are a per mailbox thing.
So moves in ImapMboxData.


imap/imap.c
```diff
@@ -165,7 +165,7 @@ static void set_flag(struct ImapAccountData *adata, int aclbit, int flag,
                      const char *str, char *flags, size_t flsize)
 {
   if (mutt_bit_isset(adata->mailbox->rights, aclbit))
-    if (flag && imap_has_flag(&adata->flags, str))
+    if (flag && imap_has_flag(&((struct ImapMboxData *)adata->mailbox->mdata)->flags, str))
```
Austin-Ray: This lengthy casting looks a bit nasty. Using `imap_mdata_get(adata->mailbox)` would improve readability.

gahr:
This is good, but `imap_mdata_get` is not guaranteed to return a non-NULL pointer. We should either check its return or make sure it's never NULL.

sileht:
In "theory" we will always have mdata set there as we manipulate messages of the opened mailbox. If we doesn't have it, it's bug.
So my question is how do we prefer to track the potential bug if it's unset.
* don't catch NULL and crash, so people can report a backtrace ? (annoying for user, but cool for us)
* catch NULL, log an error and continue ? (not cool for us, and user may have their neomutt in an unexpected)

flatcap:
For now, I think leave it unchecked.
The `Mailbox` is an object that **we**'ve populated.

Generally, I like to see library functions checking all their parameters.
In those cases there are many callers and a single function.

In the mxapi we have the reverse -- a single function that calls many others.
The mxapi function seems a logical point to do all the testing -- create a contract saying that when we call api functions, the `Mailbox` will be guaranteed to have a `magic` (that matches the backend), a `path`, a `realpath` and an `Account`.

PR #1434, opened 2018-11-19
imap: reduce complexity of imap_exec

sileht:
imap_exec flag IMAP_CMD_FAIL_OK is the only one that configure the
function return instead of the function behavior.

Also returned value of the function depends on the flag make hard to
understand what happen after imap_exec() is run.

This change removes this flags, instead we always returns the detailed
returns code, the caller have to deal with it.

Return code are always IMAP_EXEC_SUCCESS/IMAP_EXEC_ERROR/IMAP_EXEC_FATAL and
depends on if the command succeed, failure or if the imap connection
fail.

### DIFF
--- a/imap/command.c
+++ b/imap/command.c
@@ -134,10 +134,10 @@ static int cmd_queue(struct ImapAccountData *adata, const char *cmdstr, int flag
   {
     mutt_debug(3, "Draining IMAP command pipeline\n");
 
-    const int rc = imap_exec(adata, NULL, IMAP_CMD_FAIL_OK | (flags & IMAP_CMD_POLL));
+    const int rc = imap_exec(adata, NULL, flags & IMAP_CMD_POLL);
 
-    if (rc < 0 && rc != -2)
-      return rc;
+    if (rc == IMAP_EXEC_ERROR)
+      return IMAP_CMD_BAD;
   }
 
   struct ImapCommand *cmd = cmd_new(adata);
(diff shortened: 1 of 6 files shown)


imap/imap_private.h
```diff
@@ -73,10 +73,14 @@ struct Progress;
...
-#define IMAP_CMD_QUEUE   (1 << 2)
-#define IMAP_CMD_POLL    (1 << 3)
+#define IMAP_CMD_PASS    (1 << 0)  /**< Run the imap command and all previous commands queued */
+#define IMAP_CMD_QUEUE   (1 << 1)  /**< Queue a command */
+#define IMAP_CMD_POLL    (1 << 2)  /**< Poll the tcp connection before running the imap command */
+
+/* imap_exec flags (see imap_exec) */
+#define IMAP_EXEC_SUCCESS (1 << 0) /**< Imap command executed or queued succeed */
+#define IMAP_EXEC_ERROR   (1 << 1) /**< Imap command failure */
+#define IMAP_EXEC_FATAL   (1 << 2) /**< Imap connection failure */
```
flatcap: Unlike the flags, above, these return codes are mutually-exclusive.
They would probably be better as an `enum`.

imap/command.c
```diff
@@ -1240,8 +1240,6 @@ const char *imap_cmd_trailer(struct ImapAccountData *adata)
  * Also, handle untagged responses.
```
flatcap: The `@retval`s above need changing to match the new return codes.

sileht: done

imap/imap_private.h
sileht: done

flatcap:
I changed a few Doxygen comment blocks to make them auto-link.
- `#` before a constant name: `#IMAP_CMD_PASS`
- `()` after a function name: `imap_cmd_step()`

PR #1437, opened 2018-11-19
Implement IMAP reconnection support.

EdwardBetts:
If the IMAP connect drops neomutt will reconnect to the IMAP server. Closes: #1248 

I've tested this on my own machine with a mock IMAP server written in Python and with dovecot. It works in both cases.

### DIFF
--- a/imap/command.c
+++ b/imap/command.c
@@ -162,10 +162,10 @@ static void cmd_handle_fatal(struct ImapAccountData *adata)
     return;
 
   struct ImapMboxData *mdata = adata->mailbox->mdata;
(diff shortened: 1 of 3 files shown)


sileht:
I haven't tested it yet, but the quick look looks great. I wanted to split mx_mbox_open for other stuff, thx !

sileht:
I have tried, but it does not work well for me, I see all my emails twice after the reconnection and next any mailbox operation crash.

EdwardBetts:
Thanks for trying this. I'm going to close for now and reopen when I've got something that works better.

PR #1446, opened 2018-11-21
postpone: Don't reopen Postponed context

sileht:
If the selected Mailbox is already the Postponed one, no need to open a
new context.

This change reuse the current context.

### DIFF
--- a/postpone.c
+++ b/postpone.c
@@ -287,7 +287,12 @@ int mutt_get_postponed(struct Context *ctx, struct Email *hdr,
   if (!Postponed)
     return -1;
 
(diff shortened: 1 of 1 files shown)


postpone.c
```diff
@@ -283,11 +283,21 @@ int mutt_get_postponed(struct Context *ctx, struct Email *hdr,
   int code = SEND_POSTPONED;
   const char *p = NULL;
   int opt_delete;
+  bool postponed_opened = false;
```
gahr: you could perhaps do without this variable and use `PostContext == ctx` ?

sileht:
I have rebased it, it doesn't crash anymore.

PR #1450, opened 2018-11-22
Add more convenient functions to map file lines

nobrowser:
Adds a couple more convenient functions to read lines from a file.

This should help with addressing the review comments in #devel/extenal-search.

### DIFF
--- a/mutt/file.c
+++ b/mutt/file.c
@@ -689,6 +689,55 @@ char *mutt_file_read_line(char *s, size_t *size, FILE *fp, int *line, int flags)
 }
 
 /**
+ * mutt_file_iter_line - iterate over the lines from an open file pointer
+ * @param iter  State of iteration including ptr to line
+ * @param fp    File pointer to read from
+ * @param flags Same as mutt_file_read_line()
+ * @retval      true iff data read, false on eof
+ *
+ * This is a slightly cleaner interface for mutt_file_read_line() which avoids
+ * the eternal C loop initialization ugliness.  Use like this:
+ *
+ * ```
+ * struct MuttFileIter iter = { 0 };
+ * while (mutt_file_iter_line(&iter, fp, flags))
+ * {
+ *   do_stuff(iter.line, iter.line_num);
+ * }
+ * ```
+ */
+bool mutt_file_iter_line(struct MuttFileIter *iter, FILE *fp, int flags)
+{
+  char *p = mutt_file_read_line(iter->line, &iter->size, fp, &iter->line_num, flags);
+  if (!p)
+    return false;
+  iter->line = p;
+  return true;
+}
+
+/**
+ * mutt_file_map_lines - Process lines of text read from a file pointer
+ * @param func      Callback function to call for each line, see mutt_file_map_t
+ * @param user_data Arbitrary data passed to ``func''
+ * @param fp        File pointer to read from
+ * @param flags     Same as mutt_file_read_line()
+ * @retval          true iff all data mapped, false if ``func'' returns false
+ */
+bool mutt_file_map_lines(mutt_file_map_t func, void *user_data, FILE *fp, int flags)
+{
+  struct MuttFileIter iter = { 0 };
+  while (mutt_file_iter_line(&iter, fp, flags))
+  {
+    if (!(*func)(iter.line, iter.line_num, user_data))
+    {
+      FREE(&iter.line);
+      return false;
+    }
+  }
+  return true;
+}
+
+/**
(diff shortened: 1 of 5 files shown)


mutt/file.h
```diff
@@ -35,6 +35,18 @@ extern char *Tmpdir;
 #define MUTT_CONT (1 << 0) /**< \-continuation */
 #define MUTT_EOL  (1 << 1) /**< don't strip `\n` / `\r\n` */
 
+/* State record for mutt_file_iter_line() */
+struct mutt_file_iter {
+  char* s;                      /**< the line data */
```
Austin-Ray: Wouldn't it be more understandable if `s` was named `line` and `line` was named `line_num`?

It'd make `mutt/file.c:726` easier to understand.

mutt/file.c
```diff
@@ -688,6 +688,50 @@ char *mutt_file_read_line(char *s, size_t *size, FILE *fp, int *line, int flags)
...
+ * @param d              arbitrary data passed to ``f''
+ * @param fp             file pointer to read from
+ * @param flags          same as mutt_file_read_line
+ * @retval               true iff all data mapped, false if ``f'' returns false
+ */
+bool mutt_file_map_lines(mutt_file_map_function f, void *d, FILE *fp, int flags)
+{
+  struct mutt_file_iter i = { 0 };
+  while (mutt_file_iter_line(&i, fp, flags)) {
+    if (!(*f)(i.s, i.line, d)) {
```
Austin-Ray: This line is arcane to read. 

The variable names are too abstract and do not offer insight into what their purpose is. I can read the above documentation to understand what they do, but I'm forgetful and will need to reference the docs a few times :(

I'd recommend:
 - rename `f` -> `func`
 - rename `i` -> `iter`
 - rename `d` -> `data`

Austin-Ray:
I apologise for the pedantic review (i'm sure you've realised i am a bit anal for self-descriptive code.)

The code looks logically correct. Could you provide a few test cases to demonstrate its logical integrity?

mutt/file.h
nobrowser: These names were copied literally from the existing function mutt_file_read_line().

If I change mine, they need to be changed there too.  To me consistency in an API is even more important than easy readability of names.

mutt/file.c
nobrowser: Yes, good point.  Will do.

nobrowser:
I will be new to the test framework so it will take a while, but yes I'll see what I can do about writing a supporting test for these functions.

Austin-Ray:
I agree. Let's unify the naming. 

@flatcap: any objections or concerns with changing the naming?

nobrowser:
* Updated naming (for my functions only, as requested by @flatcap )
* Added test cases

mutt/file.c
nobrowser: See my latest push

Austin-Ray:
Looks good to me. I approve.

@flatcap: Anything to add?

flatcap:
Thanks @nobrowser.
It's a good idea.  It's only made a bit complicated because of the function that it's wrapping.

I've made a few minor changes, which I've put in separate commits for clarity (I'll squash them before merging).

- ccc630a5e clang-format - project code style
- eb7e722c2 rename struct/typedef
  structs are `MixedCase` and typedefs are suffixed `_t`
- 39e6b54e9 tidy - minor stuff
  Reduce the scope of some variables, e.g. moving `int` into the `for` loop
  Testing pointers with `!` rather than against `NULL`
  Filling out more Doxygen function comments (you might want to check that my descriptions make sense)

I've fully documented the typedef for the function callback -- I've done this in quite a few places in NeoMutt.
That way, we can document every callback as simply "Implements typedef_name_t".
e.g. [docs](https://github.com/neomutt/neomutt/blob/master/format_flags.h#L45) and [implementation](https://github.com/neomutt/neomutt/blob/master/compress.c#L296)

> Updated naming (for my functions only, as requested by @flatcap)

Sorry if I wasn't clear.
I'm happy for you to rename variables, etc, in existing functions.
As long as they're in a separate commit from the new code.

nobrowser:
Hmm, the style guide must have changed since I last discussed it :-(  Thanks for doing all this work for me, I can certainly appreciate it.

For Emacs users among the project members - which of the predefined cc-mode styles does this fit most closely?  Looks kinda like "gnu" to me but I want opinions.

I will amend the PR renaming the existing functions today.

nobrowser:
Ok, renamed old func in separate commit as requested.  Should really be good to go in now.

PR #1451, opened 2018-11-23
Change dependency chain regarding git_ver

mjsir911:
* **What does this PR do?**

Replaces `git_ver.h` with `git_ver.{c,o}` and removes dependency from all files on `git_ver.h`, having a dependency on only the final `$(NEOMUTT)` target

Every time I changed my git's state, the entire repository was rebuilt because `git_ver.h` changed: https://pastebin.com/v629aTPr

On top of that, it seems on some machines this dependency does not even get processed and the `git_ver.h` file is not updates: https://pastebin.com/xKXhD78R

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -33,7 +33,7 @@ doc/neomuttrc.5
 # Generated source
 config.h
 conststrings.c
-git_ver.h
+git_ver.c
 hcache/hcversion.h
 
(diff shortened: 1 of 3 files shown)


flatcap:
Rebased; Added `git_ver.c` to `GENERATED`.

Issue #1353, opened 2018-10-10
Viewing Attachments immediately deletes temp-file in /tmp/


whnr:
Workflow:
1) Open attachment with mailcap (`xdg-open %s`)
2) File viewer opens and mutt creates tempfile
3) Tempfile is _immediately_ deleted by mutt
4) Fileviewer complains that there is nothing to view

I guess this is even intended behaviour. If yes: Is there a config-switch for this? `set wait-key` does not seem to be it?

whnr:
Hi @flatcap 

Thanks for the quick reply!

> What you need is a sleep:
> This gives the browser a chance to open the file before it's deleted.

This works if the receiving app is either using it's own temp system (Browser) or keeps stuff in RAM. However, there are a few that watch for changes in the files. This will lead images disappearing exactly after "sleep" seconds.

But any change would be a complete rewrite of the way temporary files are handled. It might be best to leave it in the users hands. If one can use mutt she might be able to hack a usable shell script.

I like the workaround with copying away files to another temporary directory, that is removed, when mutt is closed. Credit: https://unix.stackexchange.com/a/109738

whnr:
OK, this is getting a bit weird. Before opening a bug. I want a second opinion. Now I'm copying files away when opening them. When I shutdown neomutt I iissue the following command to clear my temp-files:

    shutdown-hook `rm -f /tmp/mutt/*`

However, the command seems to be issued, when I start neomutt the next time! Is there any known mixup?

flatcap:
Ha!  Brilliant.
It took me a while to understand what was going on and why.

The `startup-hook` and `shutdown-hook` take a **NeoMutt** command, not a shell command.

NeoMutt reads the config file at startup, sees the backticks and executes them in order to get the result.  `rm -f /tmp/mutt/*` doesn't output anything, so NeoMutt sees `shutdown-hook` with no parameters, which gets ignored.

It's a bit of a pain, but you can run external commands like this:
```
startup-hook  'echo `action.sh ARGS`'
shutdown-hook 'echo `action.sh ARGS`'
```

The single quotes prevent the backticks being expanded.
The `echo` command prevents and empty command error (and isn't noticeable).

whnr:
Perfect, that was it.

Using an “! comand.sh” in this context does not work either. That was my first try.

I guess I’d like that information in the documentation. I’ll check the website on how to contribute.

flatcap:
> I’ll check the website on how to contribute.

[...]

PR #1452, opened 2018-11-24
How to run shell commands in global-hooks

whnr:
Addition is based on issue https://github.com/neomutt/neomutt/issues/1353

* Added a single paragraph to the documentation of global hooks
* Should make it simpler for anyone with a similar issue.
* The basic task (disappearing temp files in attachments) has been solved already.

This is my first pull request here. I hope the contribution helps…

### DIFF
--- a/doc/manual.xml.head
+++ b/doc/manual.xml.head
@@ -12501,6 +12501,18 @@ set abort_noattach_regex = "\\&lt;attach(|ed|ments?)\\&gt;"
             </para>
           </listitem>
         </itemizedlist>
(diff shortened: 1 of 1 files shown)


flatcap:
Welcome to NeoMutt, @whnr!

Thanks for the docs.  I've put `<screen>` tags around the example to make it stand out.
I've [updated the website](https://neomutt.org/feature/global-hooks).

Note: We're currently in a phase of heavy development, so it may be a little while before the next release.

Thanks.

Issue #1326, opened 2018-08-29
Architectural Changes


flatcap:
I have a lot of **big** ideas for NeoMutt, but most of them require changes to the architecture of the code.

Currently the code has a very large overlap between data and display elements.
Separating these two parts will simplify the code.

Everyone's invited to discuss my ideas and ask questions about my plans.

flatcap:
IRC Discussion 2018-07-16

![arch](https://flatcap.org/mutt/neo-arch1.svg)

`<flatcap>` that shows who calls whom, and includes various external connections
`<flatcap>` the browser api doesn't exist (beyond detailed notes)
`<flatcap>` I have some deep questions about the header cache, I think it might be in the wrong place
`<flatcap>` and the header cache is deeply entwined in each mailbox
`<flatcap>` there are a few places where a mailbox gets the cached version, then tweaks it, then gives it to the index
`<flatcap>` I'm wondering if we can eliminate these tweaks, or at least externalise them in some way
`<flatcap>` if we could...  we could move the cache up a level in the diagram.  the mailbox api would talk to it (and the mailboxes would be unaware)
`<gahr>` are you saying that it should be mailbox -> cache -> conn, so the cache should be auto-feeding itself from either local storage or servers?
`<edwinbalani>` not an inline cache then, so to speak :-)
`<flatcap>` the mailbox api checks the cache, otherwise talks to the mailbox
`<flatcap>` ooh, I hadn't thought that far, edwinbalani 
`<flatcap>` as there are various 'level' of retrieval, there'd need to be flags, etc
`<flatcap>` imap has: list of emails, headers, bodystructure, full download
`<flatcap>` I'd be interested to know how much memory the Context actually uses (for a normal mailbox, with 1000 emails)
`<flatcap>` the other hcache thing I noticed was that several structs are being serialised 'whole' meaning which is wasting a few hundred bytes per record

flatcap:
![key](https://flatcap.org/mutt/arch-key.svg)
![arch](https://flatcap.org/mutt/arch.svg)
This one shows the account/mailbox/email data, and three views: sidebar, index, pager
(the faded objects are references to objects in the data tree (left)

flatcap:
### Reorganising Context/Mailbox

The Context (the current mailbox) has duplicate copies of 6 values which are kept in the Mailbox.
These commits de-dupe the Context.

See branch [devel/context](https://github.com/neomutt/neomutt/tree/devel/context)

[...]

PR #1457, opened 2018-11-27
Factor out mx_update_context() from mbox_open()

flatcap:
Following on from [this diagram of Context use](https://github.com/neomutt/neomutt/issues/1326#issuecomment-438085645), I've refactored the backends to move `mx_update_context()` out of `*_mbox_open()`.

Here are a couple of diagrams that may illustrate what I've done.
The represent the function call tree leading to `mx_update_context()`.

**Key:**
- **Solid Red** - The problem function, `mx_update_context()`, which has been eliminated
- **Dot** - New locations of `mx_update_context()` calls
- **Purple** - MXAPI functions
- **Green** - Other functions called by NeoMutt (from outside the email backend)
- **Red** - Private functions

### Mbox backend

mbox

### Nntp backend

nntp

---

This is an improvement for `mbox_open()`, but it just shifts the problem down to `mbox_sync()` and `mbox_check()`.

### DIFF
--- a/compress.c
+++ b/compress.c
@@ -495,12 +495,10 @@ int comp_ac_add(struct Account *a, struct Mailbox *m)
  * Then determine the type of the mailbox so we can delegate the handling of
  * messages.
  */
-static int comp_mbox_open(struct Context *ctx)
+static int comp_mbox_open(struct Mailbox *m, struct Context *ctx)
 {
   if (!ctx || !ctx->mailbox || (ctx->mailbox->magic != MUTT_COMPRESSED))
     return -1;
-
-  struct Mailbox *m = ctx->mailbox;
 
   struct CompressInfo *ci = set_compress_info(m);
   if (!ci)
@@ -541,7 +539,7 @@ static int comp_mbox_open(struct Context *ctx)
   }
 
   m->account->magic = m->magic;
-  return ci->child_ops->mbox_open(ctx);
+  return ci->child_ops->mbox_open(m, ctx);
 
 cmo_fail:
   /* remove the partial uncompressed file */
--- a/imap/imap.c
+++ b/imap/imap.c
@@ -1974,13 +1974,9 @@ int imap_login(struct ImapAccountData *adata)
 /**
  * imap_mbox_open - Implements MxOps::mbox_open()
  */
-static int imap_mbox_open(struct Context *ctx)
+static int imap_mbox_open(struct Mailbox *m, struct Context *ctx)
 {
-  if (!ctx || !ctx->mailbox)
-    return -1;
-
-  struct Mailbox *m = ctx->mailbox;
-  if (!m->account)
+  if (!m || !m->account)
     return -1;
 
   char buf[PATH_MAX];
--- a/maildir/maildir_private.h
+++ b/maildir/maildir_private.h
@@ -96,7 +96,7 @@ int                     maildir_check_dir      (struct Mailbox *m, const char *d
 void                    maildir_delayed_parsing(struct Mailbox *m, struct Maildir **md, struct Progress *progress);
 struct MaildirMboxData *maildir_mdata_get      (struct Mailbox *m);
 int                     maildir_mh_open_message(struct Mailbox *m, struct Message *msg, int msgno, bool is_maildir);
-int                     maildir_move_to_context(struct Context *ctx, struct Maildir **md);
+int                     maildir_move_to_context(struct Mailbox *m, struct Context *ctx, struct Maildir **md);
 int                     maildir_parse_dir      (struct Mailbox *m, struct Maildir ***last, const char *subdir, int *count, struct Progress *progress);
 void                    maildir_parse_flags    (struct Email *e, const char *path);
 struct Email *          maildir_parse_message  (enum MailboxType magic, const char *fname, bool is_old, struct Email *e);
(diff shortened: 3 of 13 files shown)


notmuch/mutt_notmuch.c
```diff
@@ -2173,7 +2173,6 @@ static int nm_mbox_open(struct Mailbox *m, struct Context *ctx)
   m->mtime.tv_sec = time(NULL);
   m->mtime.tv_nsec = 0;
 
-  mx_update_context(ctx, m->msg_count);
```
Austin-Ray: Removing this `mx_update_context(...)` call breaks the progress bar `notmuch`. This appears to be only back-end whose progress bar is broken by removing `mx_update_context(...)` so it must be an implementation issue.

maildir/maildir_private.h
```diff
@@ -96,7 +96,7 @@ int                     maildir_check_dir      (struct Mailbox *m, const char *d
 void                    maildir_delayed_parsing(struct Mailbox *m, struct Maildir **md, struct Progress *progress);
 struct MaildirMboxData *maildir_mdata_get      (struct Mailbox *m);
 int                     maildir_mh_open_message(struct Mailbox *m, struct Message *msg, int msgno, bool is_maildir);
-int                     maildir_move_to_context(struct Context *ctx, struct Maildir **md);
+int                     maildir_move_to_context(struct Mailbox *m, struct Maildir **md);
 int                     maildir_parse_dir      (struct Mailbox *m, struct Maildir ***last, const char *subdir, int *count, struct Progress *progress);
 void                    maildir_parse_flags    (struct Email *e, const char *path);
```
Austin-Ray: Not an issue, but this one out of alignment parenthesis really bothers me

compress.c
```diff
@@ -495,13 +495,11 @@ int comp_ac_add(struct Account *a, struct Mailbox *m)
  * Then determine the type of the mailbox so we can delegate the handling of
  * messages.
  */
-static int comp_mbox_open(struct Context *ctx)
+static int comp_mbox_open(struct Mailbox *m, struct Context *ctx)
 {
   if (!ctx || !ctx->mailbox || (ctx->mailbox->magic != MUTT_COMPRESSED))
```
Austin-Ray: Since `mx_mbox_open(...)` includes `NULL` checking and initialises our the `Context`, is this guard statement still necessary?

imap/imap.c
```diff
@@ -1974,13 +1974,9 @@ int imap_login(struct ImapAccountData *adata)
...
  */
-static int imap_mbox_open(struct Context *ctx)
+static int imap_mbox_open(struct Mailbox *m, struct Context *ctx)
 {
-  if (!ctx || !ctx->mailbox)
-    return -1;
-
-  struct Mailbox *m = ctx->mailbox;
-  if (!m->account)
+  if (!m || !m->account)
```
Austin-Ray: Is this guard statement necessary given `mx_mbox_open(...)`'s `NULL` checking and initialisation strategy?

nntp/nntp.c
```diff
@@ -2440,13 +2443,9 @@ int nntp_ac_add(struct Account *a, struct Mailbox *m)
...
  */
-static int nntp_mbox_open(struct Context *ctx)
+static int nntp_mbox_open(struct Mailbox *m, struct Context *ctx)
 {
-  if (!ctx || !ctx->mailbox)
-    return -1;
-
-  struct Mailbox *m = ctx->mailbox;
-  if (!m->account)
+  if (!m || !m->account)
```
Austin-Ray: Guard clause necessary?

pop/pop.c
```diff
@@ -825,13 +822,9 @@ int pop_ac_add(struct Account *a, struct Mailbox *m)
...
  */
-static int pop_mbox_open(struct Context *ctx)
+static int pop_mbox_open(struct Mailbox *m, struct Context *ctx)
 {
-  if (!ctx || !ctx->mailbox)
-    return -1;
-
-  struct Mailbox *m = ctx->mailbox;
-  if (!m->account)
+  if (!m || !m->account)
```
Austin-Ray: Guard clause necessary?

Austin-Ray:
My review may premature and a bit pedantic (regarding guard clauses; don't know if we should still use them), but hopefully, I've caught an issue or two early :)

gahr:
You must have an off-by-one error somewhere, as the address sanitizer kicks in here:
```
(gdb) bt
#0  SuppressErrorReport () at /wrkdirs/usr/ports/devel/llvm60/work/compiler-rt-6.0.1.src/lib/asan/asan_report.cc:371
#1  ReportGenericError () at /wrkdirs/usr/ports/devel/llvm60/work/compiler-rt-6.0.1.src/lib/asan/asan_report.cc:384
#2  0x00000000004780ce in __asan_report_store4 () at /wrkdirs/usr/ports/devel/llvm60/work/compiler-rt-6.0.1.src/lib/asan/asan_rtl.cc:118
#3  0x0000000000638d03 in mx_update_context (ctx=0x607000001670, new_messages=9921) at mx.c:1268
#4  0x0000000000636601 in mx_mbox_open (m=0x61d000006480, path=0x7fffffffc2c0 "imaps://ptrcrt.ch/INBOX", flags=0) at mx.c:377
#5  0x00000000005e524a in main (argc=1, argv=0x7fffffffea20, envp=0x7fffffffea30) at main.c:1209

(gdb) list
1263          e->security = crypt_query(e->content);
1264        }
1265
1266        if (!ctx->pattern)
1267        {
1268          m->v2r[m->vcount] = msgno;
1269          e->virtual = m->vcount++;
1270        }
1271        else
1272          e->virtual = -1;

(gdb) p m->vcount
$5 = 9946
(gdb) p m->hdrmax 
$6 = 9946
(gdb) 
```

I think `m->hdrmax` holds the number of entries in `m->v2r`, so trying to store at index `m->vcount` when `m->vcount == m->hdrmax` is wrong.

flatcap:
Updates, after discussion on IRC.

@Austin-Ray's missing progress bar has been resolved -- I'd reset the message count when I shouldn't have.

I'd like to remove the guards, but I'm going to keep them for now.
I still like the idea of the mxapi having a contract -- certain objects are guaranteed -- but some functions will require more objects than others.  Still thinking...

@gahr's crash has been resolved.  I hadn't removed the `mx_update_context()` from the imap code, meaning it got called twice.

maildir/maildir_private.h
flatcap: I like your attention to detail, but I can't see any mis-alignment.

Austin-Ray: I must have been delirious because I could have sworn a parenthesis was one column to the left of the others

Austin-Ray:
@flatcap Resolved out all of my review comments.

PR #1459, opened 2018-11-28
postpone: fix Postcount when we are in current selected mailbox

sileht:


gahr:
go to drafts, compose, save draft -> you got a New message in drafts. Compose -> "Recall postponed..?" -> yes -> error "No postponed messages"

sileht:
@gahr if you apply #1446 on top of this, you should not have the issue.

PR #1461, opened 2018-11-28
imap: handle case where adata->mailbox is NULL

### LINKED ISSUE #1460: imap: unsubscribe from selected mailbox causes a crash
1. Open NeoMutt on an IMAP account
2. Switch to Mailboxes view
3. Unsubscribe from Inbox (the selected mailbox)
4. Crash

sileht:
When selected mailbox in unsubscribed, adata->mailbox may be NULL.

make imap_cmd_finish() handles it.

Closes #1460

### DIFF
--- a/imap/command.c
+++ b/imap/command.c
@@ -1305,7 +1305,7 @@ void imap_cmd_finish(struct ImapAccountData *adata)
 
   adata->closing = false;
 
-  struct ImapMboxData *mdata = adata->mailbox->mdata;
+  struct ImapMboxData *mdata = imap_mdata_get(adata->mailbox);
 
   if (mdata && mdata->reopen & IMAP_REOPEN_ALLOW)
   {


gahr:
Have you tested that it solves #1460? I don’t think it does: in my stack trace, all pointers are valid (as in `!= NULL`), but the pointed-to memory’s already been free’d.

sileht:
No, I have just guessing because I can't reproduce #1460.

gahr:
You might try to run under an address sanitizer: https://clang.llvm.org/docs/AddressSanitizer.html

sileht:
Even with address sanitizer it doesn't crash for me.

gahr:
I would say let's merge this as it's a safe net, although it doesn't solve #1460.  See #1463 for an alternative approach at solving it.

PR #1464, opened 2018-11-29
Do not use the account's delimiter if it wasn't figured yet

### LINKED ISSUE #1438: IMAP: 'neomutt -f =foo/bar' opens the foo mailbox
I expect that running `neomutt -f =foo/bar` will open the 'foo/bar' mailbox, instead neomutt opens the 'foo' mailbox, or complains if 'foo' isn't a mailbox.

This problem is only being triggered for me when using IMAP. Reading local Maildir mailboxes with neomutt works fine.

I ran git bisect and it says d6e6635 is when this change happened.

Here's some example settings from muttrc
```
set folder=imap://edward@localhost/
mailboxes =foo/bar
```

gahr:
Closes #1438

* **What does this PR do?**

The account's delimiter character is used to normalize IMAP paths. Between the account is created and the IMAP session is logged-in, the delimiter remains unset (defaults to `\0`). This commit changes `imap_fix_path` not to use such a default delimiter.

This allows to specify a mailbox on startup. It has the (mild) downside that if the mailbox is specified with the wrong delimiter, say `/` instead of `.`, we now exit with an error instead of truncating the mailbox to the parent path.

So, if my IMAP's delimiter is `.`.

| Command | Before this change | After this change |
|---------------|------------------------|----------------------|
| `neomutt -f Lists.Neomutt-devel` | Opens `Lists` | Opens `Lists.Neomutt-devel` |
| `neomutt -f Lists/Neomutt-devel` | Opens `Lists` | Fails with `Invalid mailbox name` error  |

### DIFF
--- a/imap/util.c
+++ b/imap/util.c
@@ -788,8 +788,9 @@ char *imap_fix_path(struct ImapAccountData *adata, const char *mailbox, char *pa
 {
   int i = 0;
   char delim = '\0';
+  bool use_account_delim = (adata && adata->delim != '\0');
 
-  if (adata)
+  if (use_account_delim)
     delim = adata->delim;
 
   while (mailbox && *mailbox && i < plen - 1)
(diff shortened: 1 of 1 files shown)


gahr:
The fact that we fail to turn the path into something IMAP recognizes is a regression wrt to upstream, which works fine with both `-f foo/bar` and `-f foo.bar` in my case.

gahr:
We'll go with #1467 instead.

PR #1467, opened 2018-11-29
imap: login on startup

### LINKED ISSUE #1438: IMAP: 'neomutt -f =foo/bar' opens the foo mailbox

sileht:
This ensure we always have a valid adata.

Closes #1438

### DIFF
--- a/imap/imap.c
+++ b/imap/imap.c
@@ -343,8 +343,6 @@ int imap_prepare_mailbox(struct Mailbox *m)
   struct ImapMboxData *mdata;
   struct ImapAccountData *adata = m->account->adata;
 
-  mutt_account_hook(m->realpath);
-
   if (!m->mdata)
   {
     struct Url *url = url_parse(m->path);
@@ -353,9 +351,6 @@ int imap_prepare_mailbox(struct Mailbox *m)
     m->free_mdata = imap_mdata_free;
     url_free(&url);
   }
-
-  if (imap_login(adata) < 0)
-    return -1;
 
   return 0;
 }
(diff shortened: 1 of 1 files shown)


flatcap:
Moving the login _this_ early means we have the folder delimiter from the server.

This doesn't really affect the startup for normal users.
If `$spoolfile` is set, then `main()` calls `mx_mbox_open()`:

```
#0 imap_login (adata=0x680290) at imap/imap.c:1925
#1 imap_prepare_mailbox (m=0x67e1b0) at imap/imap.c:357
#2 imap_mbox_open (m=0x67e1b0, ctx=0x63bf10) at imap/imap.c:1987
#3 mx_mbox_open (m=0x67e1b0, path=0x7fffffffbef0 "imap://10.8.0.1/INBOX", flags=0) at mx.c:370
#4 main (argc=1, argv=0x7fffffffd0c8, envp=0x7fffffffd0e8) at main.c:1209
```

With this PR, `imap_login()` is simply called slightly earlier in the same sequence:

```
#0 imap_login (adata=0x680290) at imap/imap.c:1925
#1 imap_ac_add (a=0x5dd5b0, m=0x67e1b0) at imap/imap.c:1899
#2 mx_ac_add (a=0x5dd5b0, m=0x67e1b0) at mx.c:1711
#3 mx_mbox_open (m=0x67e1b0, path=0x7fffffffbef0 "imap://10.8.0.1/INBOX", flags=0) at mx.c:285
#4 main (argc=1, argv=0x7fffffffd0c8, envp=0x7fffffffd0e8) at main.c:1209
```

flatcap:
Hmm... but if `$spoolfile` isn't set, `imap_check_subscribed` doesn't seem to work.

Issue #1465, opened 2018-11-29
we need a "save-attach-dir" variable, or a "chdir" command


libBletchley:
## What would you like to be able to do?

I would like the default directory for attachments to be controlled by the user configuration.

## What needs to change in NeoMutt?

The current default directory for attachment saves is the directory neomutt was launched from, which is almost never the desired directory.  The default should be controllable by the user.  The default likely corresponds with a mailbox, so I would envision something like this:
```
folder-hook +in/class_math* set save-attach-dir = ~/class/math/
```
or this would perhaps have the same effect
```
folder-hook +in/class_math* 'chdir ~/class/math/'
```

## Have you tried any alternative methods?
The macro approach looks like this:
```
macro attach W <save-entry><kill-line>/home/gawron/attachments/ 
```
The problem is that there would have to be a separate macro for every folder, which would create lots of macro pollution.  The user would quickly run out of keystrokes and forget them.

## Additional Info
There has been a past attempt to solve this problem via feature request https://github.com/neomutt/neomutt/issues/196 and it was found to be *difficult* in terms of effort.  Presumably that's because creating a new kind of hook is non-trivial.  But simply creating a variable or `chdir` command may be easier.

PR #1468, opened 2018-11-30
Add the "attach_save_dir" config variable

gahr:
This specifies the location where attachments are saved. The default is the current working directory.

Issue #1465


matt-snider:
@gahr I was looking for just this feature today while saving an attachment - nice! :+1: 

Just wondering if it should also be added to the docs or neomuttrc file

Issue #1469, opened 2018-12-02
Make encryption flag chars configurable like $flag_chars


binaryplease:
## What would you like to be able to do?

I would like to use different characters for the encrypted/signed/etc. flag in the index. 
The config variable $flag_chars (https://neomutt.org/guide/reference.html#3-80-%C2%A0flag_chars) already allows to do this for some of the other flags, but I can't change the remaining ones that way.

## What needs to change in NeoMutt?
The used characters are currently hardcoded. We would need a new variable we can set in the configuraion.

flatcap:
Thanks, @binaryplease.  That sounds like a reasonable idea.
(and thanks for your offer, on IRC, to implement it)
This would mean that **all** the status characters would be configurable.

As you've noticed, `$flag_chars` covers most of the options.
Each case is a simple binary 'on' or 'off'.  They are:

1. `*` Mail is tagged
2. `!` Mail is flagged as important
3. `D` Mail is marked for deletion
4. `d` Mail has attachments marked for deletion
5. `r` Mail has been replied to
6. `O` Mail is Old (Unread but seen)
7. `N` Mail is New (Unread but not seen)
8. `o` Thread is Old (Unread but seen)
9. `n` Thread is New (Unread but not seen)

The crypto status of the message could have several different possibilities.
We _could_ extend `$flag_chars`, but that might make it difficult to add new options later  (backwards **in**compatible!)

I suggest a config variable `$crypto_chars`:
1. `S` Message is signed, and the signature is successfully verified
2. `P` Message is PGP encrypted
3. `s` Message is signed
4. `K` Message contains a PGP public key
5. ` ` Message has no crypt info (perhaps)

@somini @sassmann 
You've worked on this code, what do you think?

flatcap:
Fortunately, most of the code is already written.
1. Add a new config item (copy the `flag_chars` variable)
   - Definition in `init.h`
   - Add global variable in `hdrline.[ch]`
2. Create enum for flags in `hdrline.c`
3. Edit `index_format_str()` to use new chars

binaryplease:
Will get started on this this evening. 

One question regarding the backwards incompatibility: Why would it be incompatible if we just extend $flag_chars at the end? 

I just tried adding more chars to $flag_chars, any excess is ignored. So this would mean, if you set more characters, they are could be used. If not (shorter string) use the default.

flatcap:
[...]

PR #1473, opened 2018-12-03
Implements #1469

binaryplease:
* **What does this PR do?**
Implements #1469


flatcap:
Rebased; squashed; fixed a few typos.

gahr:
Maybe it was better to use `crypt_chars` instead of `crypto_chars`, for consistency with all other `crypt_*` variables.

flatcap:
Hmm... I agree.
Please can you rename it (and the backing variables) and push it, before people start to use it.

PR #1487, opened 2018-12-10
imap: use Mailbox instead ImapAccountData on some functions

sileht:
Some functions manipulate  the current select mailbox. We currently pass to them the `ImapAccountData`.

This function then uses `adata->mailbox`. Never check is `adata->mailbox` is `NULL`. Sometimes it manipulates don't use other attributes of `adata`. It manipulates Email without ensuring the Email is belong the 'adata->mailbox'.

This change passes the 'Mailbox' as parameters of these functions. And if the function need that the Mailbox is selected, it explicitly checks it's the same as `adata->mailbox`.

Later we can imagine that all functions that uses the connection will run a function called `ensure_mailbox_selected` to select the `Mailbox` if needed.


imap/util.c
```diff
@@ -784,20 +784,16 @@ void imap_error(const char *where, const char *msg)
...
-  if (adata)
-    delim = adata->delim;
 
   while (mailbox && *mailbox && i < plen - 1)
   {
     if ((ImapDelimChars && strchr(ImapDelimChars, *mailbox)) || (delim && *mailbox == delim))
     {
       /* use connection delimiter if known. Otherwise use user delimiter */
-      if (!adata)
+      if (!delim)
```
Austin-Ray: Is this logically equivalent? My (limited) understanding of the `\0` character comparisons leans me towards no. Does that work or is a comparison to `\0` needed?

I apologise for the newb question.

sileht: It works because the char '\0' is 0 as integer. But I will make it explicit.

PR #1490, opened 2018-12-11
imap: fix mbox_open when the name is '\0'

sileht:
When user doesn't explicitly set the inbox to INBOX, imap_mbox_open
crashes, because the Mailbox is not attached to an account.

This change fixes m->path, m->realpath to ensure the mailbox is found.
And also add check to not crashes if something similar occurs.

This fix the imap issue, but something else looks wrong.
For sure, imap_mbox_open have been called with an Mailbox with
'account' set but not 'mdata'. This means imap_ac_add have not been
called for this mailbox and somewhere else is setting account without
calling the backend code.


sileht:
https://github.com/neomutt/neomutt/pull/1491 is prefered.

PR #1498, opened 2018-12-14
imap: don't keep hcache pointer

sileht:
The pointer set in ImapMboxData->hcache is never used. Each time
the hcache is used, it's reopened, and sometimes closed.

Keeping hcache reference in ImapMboxData is not necessary, hcache is
used in only 3 places.

imap_sync_mailbox
imap_expunge_mailbox
imap_read_headers


flatcap:
Closed in favour of #1501.

PR #1499, opened 2018-12-16
Devel/ext search with folder

nobrowser:
It passes the extra folder path argument to the external search filter, as we discussed.

Also, a small side fix in compress.c : escape_path() was double-evaluated.

I have left a number of separate commits because I think they reflect separate changes, but I can squash them if you prefer that.

### DIFF
--- a/compress.c
+++ b/compress.c
@@ -268,18 +268,20 @@ static const char *compress_format_str(char *buf, size_t buflen, size_t col, int
     return src;
 
   struct Mailbox *m = (struct Mailbox *) data;
+  const char *stuffing = NULL;
 
   switch (op)
   {
     case 'f':
       /* Compressed file */
-      snprintf(buf, buflen, "%s", NONULL(mutt_path_escape(m->realpath)));
+      stuffing = mutt_path_escape(m->realpath);
       break;
     case 't':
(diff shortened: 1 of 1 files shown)


flatcap:
Thanks.  I've done a little light tidying and merged the search changes.
The only notable changes are:
- rename `mutt_escape_path()` to `mutt_path_escape()` to match the other library functions
- add `mairix_filter.pl` to install script

I'm just not sure about the remaining commit - 8a3448aec Fix double evaluation of mutt_escape_path() inside a macro.
For the `f` and `t` cases, it's not really saving any cpu cycles.

Perhaps (to ensure that `buf` is defined) we just want:
```c
  default:
    buf[0] = '\0';
```

nobrowser:
> Fix double evaluation of mutt_escape_path() inside a macro.
> For the f and t cases, it's not really saving any cpu cycles.

Really?  `NONULL(x)` is defined as `((x) ? (x) : "")` .  So `NONULL(mutt_path_escape(s))` expands to `((mutt_path_escape(s)) ? (mutt_path_escape(s)) : "")` . SWIM ?

flatcap:
Ah, of course.  Well spotted.

PR #1506, opened 2018-12-23
Command prompt should close when deleting the ':' as it does in vim

### LINKED ISSUE #1505: Command prompt should close when deleting the ':' as it does in vim

When you open the command prompt, then change your mind, pressing delete doesn't close the command prompt. This is the behaviour in vim and should work in neomutt the same way.

binaryplease:
Fixes #1505

### DIFF
--- a/enter.c
+++ b/enter.c
@@ -303,7 +303,13 @@ int mutt_enter_string_full(char *buf, size_t buflen, int col, int flags, bool mu
 
         case OP_EDITOR_BACKSPACE:
           if (state->curpos == 0)
+          {
+            // Pressing backspace when no text is in the command prompt shold exit the prompt
+            if (state->lastchar == 0)
+              goto bye;
+            // Pressing backspace with text in the command prompt should just beep
             BEEP();
+          }
           else
           {
             i = state->curpos;


enter.c
```diff
@@ -303,7 +303,14 @@ int mutt_enter_string_full(char *buf, size_t buflen, int col, int flags, bool mu
...
         case OP_EDITOR_BACKSPACE:
           if (state->curpos == 0)
-            BEEP();
+          {
+            // Pressing backspace when no text is in the command prompt shold exit the prompt
+            if (state->lastchar == 0)
+              goto bye;
+            // Pressing backspace with text in the command prompt should just beep
+            else
+              BEEP();
```
flatcap: I'd drop the `else` here and unindent the `BEEP`.
In this case it barely matters, but it's good to reduce the indent by returning early.
e.g.
BAD
```c
void function(void)
{
  if (cond)
  {
    // 100s of lines
  }
}
```

GOOD
```c
void function(void)
{
  if (!cond)
    return;

  // 100s of lines
}
```

binaryplease: Ok, removed the else as requested.

binaryplease:
I have updated the commit message, should be clearer now.
Let me know if something else is missing. Git setting show my full name over here.

PR #1507, opened 2018-12-27
Improvements to command_t parse errors

nocriw:
* **What does this PR do?**
Group command_t error codes into an enumeration and improve errors quality by differentiating between errors and warnings.


nocriw:
@flatcap could you please check both `mutt_lua_parse()` & `mutt_lua_source_file()` in `mutt_lua.c`? They both return 2 if successful, should we create a flag for this or change it to 0?

mutt_commands.h
```diff
@@ -29,6 +29,7 @@ struct Buffer;
  * enum CommandError - Error codes for command_t parse functions
  */
 enum CommandError {
+  COMMAND_SUC =  0,
   COMMAND_ERR = -1,
   COMMAND_WAR = -2,
   COMMAND_FIN       ///< Distinguish "finish" command from successful return.
```
flatcap: Hmm...   What do you suppose the value of `COMMAND_FIN` is?!

nocriw: Oh sorry, the value was 1! Since it was only used in the finish command I didn't associate a value.

```suggestion
  COMMAND_FIN =  1  ///< Distinguish "finish" command from successful return.
```

flatcap:
> both `mutt_lua_parse()` & `mutt_lua_source_file()` return 2 if successful

How odd.  Nothing else calls those functions, so please change them to the new enum.

flatcap:
I've made a few small changes:
- Renamed the enumerations, e.g. `_WAR` to `_WARNING`
- Tidied the Doxygen comments
- Run clang-format

It's a lovely job.  Many thanks.

PR #1519, opened 2019-01-03
notmuch: make sure nm_db_longrun_done() release DB

karelzak:
Since commit 1cdfc0da08e09b1756a7b4d422996e907c4390d0 the function
nm_db_release() does not release DB if long-run state is enabled.
This change make code more readable and shorter.

Unfortunately, nm_db_release() is called from nm_db_longrun_done(), so
after the change the function nm_db_longrun_done() is no-op and long
transaction never finish. It means:

 - changes are not committed the database
 - after mutt restart old state (tags) is read from DB
 - mismatch between notmuch tags and maildir filenames

This commit disables long-run state in nm_db_longrun_done() before
nm_db_release() is called.

### DIFF
--- a/notmuch/nm_db.c
+++ b/notmuch/nm_db.c
@@ -302,8 +302,14 @@ void nm_db_longrun_done(struct Mailbox *m)
 {
   struct NmAccountData *adata = nm_adata_get(m);
 
-  if (adata && (nm_db_release(m) == 0))
-    mutt_debug(2, "nm: long run deinitialized\n");
+  if (adata)
+  {
+    adata->longrun = false; /* to force nm_db_release() released DB */
+    if (nm_db_release(m) == 0)
+      mutt_debug(2, "nm: long run deinitialized\n");
+    else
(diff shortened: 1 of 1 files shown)


notmuch/nm_db.c
```diff
@@ -302,8 +302,13 @@ void nm_db_longrun_done(struct Mailbox *m)
 {
   struct NmAccountData *adata = nm_adata_get(m);
 
-  if (adata && (nm_db_release(m) == 0))
-    mutt_debug(2, "nm: long run deinitialized\n");
+  if (adata) {
```
Austin-Ray: Please move the brace to a new line to match the coding style :)

flatcap: I've just done that.

Austin-Ray:
Hi @karelzak thanks for the PR (and many thanks for writing the `notmuch` integration.)

Everything looks good besides the one minor code style issue. I appreciate the analysis you've provided :)

Thanks for catching my mistake!

Issue #1224, opened 2018-05-25
Convert Pattern to use STAILQ


flatcap:
## Description

Many NeoMutt structures implement their own linked-list type.
The Patterns are simply stored as a long linked list.

`STAILQ` is a singly-linked list with pointers to both the head and the tail.
It's defined in [mutt/queue.h](https://github.com/neomutt/neomutt/blob/master/mutt/queue.h),

**Note**: This isn't too hard, but there are lots of uses of Pattern.

### Part 2: Change all users to match

There are plenty of examples in NeoMutt's code of how to iterate through a list.
For example: [tags.c](https://github.com/neomutt/neomutt/blob/master/tags.c), [tags.h](https://github.com/neomutt/neomutt/blob/master/tags.h)

**Note**: When deleting items, use `STAILQ_FOREACH_SAFE()`

yuzibo:
Hello,
Do we hope move it into "mutt/*" dir at last?

flatcap:
Sort of.
The `mutt` library dir contains **only** functions that are:
- Independent of NeoMutt code
- Tidy and Doxygen documented

There are some dependencies between library files, [as seen in the docs](https://neomutt.org/code/mutt.html), but the library can be compiled and tested separately.

The `test` dir in the source has a growing set of "[acutest](https://github.com/mity/acutest)" test cases.

Each file that we add to the libary makes it easier to understand (and improve) the core of NeoMutt.

nnathan:
I've done a rudimentary implementation. I chose to use the SLIST macros instead of STAILQ since:

1. The features of STAILQ are unnecessary. The linked list is built by appending to the tail (->next) in the original code. There is no need for list concatenat
2. Doesn't justify the slowdown.

I'm frankly unimpressed with switching to SLIST. All the changes made were simple renaming of `pat->next` with `SLIST_NEXT(pat, entries)`. It is difficult to use the SLIST macros properly with SLIST_HEAD(), SLIST_FOREACH, etc. since most of the code passes around `struct Pattern *` as the head of the list instead of an SLIST_HEAD() object. Changing the code to use SLIST_HEAD() would require considerable amounts of refactoring (which I'm open to).

What are your thoughts?

As an aside:
I used this issue as an opportunity to familiarise myself with the BSD queue.h macros. In case anyone ends up finding this issue and wondering how they work I put up a rudimentary example: https://gist.github.com/nnathan/ce79b21797513fc41b9e71a350a1ae0b .

PR #1522, opened 2019-01-06
test: add pattern compilation tests

melbit-naveennathan:
* **What does this PR do?**

Adds a couple of tests for pattern compilation.

This will ease transition to using SLIST/STAILQ to replace Pattern linked-list structures per #1224 .

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -12,6 +12,7 @@ neomutt
 pgpewrap
 test/config-test
 test/neomutt-test
+test/pattern-test
 
 # Build products
 *.o
--- a/Makefile.autosetup
+++ b/Makefile.autosetup
@@ -61,31 +61,33 @@ ALL_FILES!=	(cd $(SRCDIR) && git ls-files 2>/dev/null) || true
 ###############################################################################
 # neomutt
 NEOMUTT=	neomutt$(EXEEXT)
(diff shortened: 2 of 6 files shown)


gahr:
We really want the test to just depend on the librarized part of neomutt. I don't think we want to build a test binary that's all neomutt except main. Can you please try and put the part you want to test in a library such as `mutt/`?

melbit-naveennathan:
Per earlier discussion I've migrated the pattern testing into its own executable to reduce the pollution of unnecessary dependencies on existing testsuites.

flatcap:
I've just pushed an updated version to branch [`devel/pattern_test`](https://github.com/neomutt/neomutt/tree/devel/pattern_test)
(I can't seem to update this PR -- there's a tickbox when you create it)

The new commit:
- Reverts the build changes requiring all of NeoMutt's objects
- Adds `extract.c` - a complete copy of `mutt_extract_token()` from `init.c`
- Adds `dummy.c` - empty functions to fulfill dependencies
- Builds against `pattern.c`

The dummy functions mean that many of the features of `mutt_extract_token()` won't work, but Pattern's dependencies are now much clearer.

nnathan:
I've incorporated the changes and migrated tests for SLIST_* macros over. It is now living in https://github.com/neomutt/neomutt/tree/test/pattern . This won't get merged in until we can refactor so that it only uses the libraries per discussion with @gahr.

PR #1528, opened 2019-01-08
Don't override ndata with temporary context

sileht:
postponed/sendlib code can reopen an already opened mailbox.

This overrides m->ndata with a temporary context but never revert it back to
the previous one.

At first glance, when postponed/sendlib code open the mailbox they don't care
to have their context updated with the notification system. It should be
the Context used for the UI that should be updated instead.

So until we don't open a mailbox twice anymore, just set ndata when it
looks accurate to workaround the issue.


gahr:
I reverted c128c131185fb076476991f477565ba115352faa as it's the cause of #1529.

PR #1543, opened 2019-01-16
tags: Use STAILQ instead of strsep

sileht:
This change introduces a new string function mutt_str_split(src, sep) to
return a STAILQ of the split string.

This new funtion is used first in tags code.

### DIFF
--- a/mutt/string.c
+++ b/mutt/string.c
@@ -38,6 +38,7 @@
 #include "logging.h"
 #include "memory.h"
 #include "string2.h"
+#include "list.h"
 #ifdef HAVE_SYSEXITS_H
 #include <sysexits.h>
 #endif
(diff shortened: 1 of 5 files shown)


flatcap:
Code **and** test cases.  Fabulous!

flatcap:
There are five **config variables** which are lists:

- `:` assumed_charset
- `:` attach_charset
- `,` hidden_tags
- `:` mailcap_path
- `:` send_charset

Ideas for **flags**:

- `TRIM_WHITESPACE`
  Humans often type a space in comma-separated lists
- `ALLOW_DUPES`
  I don't think any of the current config makes sense to have repeated entries
- `ALLOW_EMPTY`
  e.g. `apple:banan::cherry`
  I don't think we need this, yet, either
- `CASE_SENSITIVE`
  Is case important when looking for dupes?

**Questions**:

- Should we be checking for escaped separators?
  `apple,banana\,cherry`
- Should we store the flags in the list itself?
- Should we store the item count in the list?

**Functions**

- `add_items()`, `remove_items()`
  We don't yet support `+=` or `-=` in the config
- `is_member()`
- `count_members()`
- `list_to_string()`
  Reverse the parsing process

mutt/string.c
```diff
@@ -1147,3 +1148,32 @@ const char *mutt_str_strcasestr(const char *haystack, const char *needle)
 
   return NULL;
 }
+
+/**
+ * mutt_str_split - return list of the words of the strings
+ * @param src String to split
+ * @param sep Word separator.
+ */
+struct ListHead mutt_str_split(const char *src, int sep)
```
gahr: Why `int sep` instead of `char sep`?

PR #1545, opened 2019-01-23
mxapi: introduce mx_msg_flag_modified()

### LINKED ISSUE #1485: Removing notmuch unread tag when opening a message
Hi there,

I'm new to neomutt but I've spent a lot of time the last several days configuring it. I've got a setup that includes several mail accounts synchronized with offlineimap and tagged with notmuch. That part seems to be working properly. On the neomutt side, I have several virtual mailboxes to display mail from all accounts based on tags. 

What doesn't seem to work, is removing the `unread` notmuch tag when displaying the message (the \S mail flag is updated). 

Here are some settings that may be relevant:
```
set nm_exclude_tags = ""
set nm_unread_tag = unread
message-hook ~A 'push "<modify-labels>-unread<enter>"'
```

I expected things to work out of the box mostly once `nm_unread_tag` was set, but figured as a last resort `message-hook` would work, but it just hangs and I need to force kill it. Maybe there is an error there.

I assume this is a fairly common configuration, so maybe I'm totally off the mark here.

Austin-Ray:
Introduce the `mx_msg_flag_modified()` function, which is used to alert a
backend if the flags are changed. This allows the backend to perform
backend-specific tasks. The notmuch backend needs this to proper update
the database when a message is opened or toggle-new is called.

Fixes #1485

### DIFF
--- a/compress.c
+++ b/compress.c
@@ -981,6 +981,7 @@ struct MxOps MxCompOps = {
   .msg_open_new     = comp_msg_open_new,
   .msg_commit       = comp_msg_commit,
   .msg_close        = comp_msg_close,
+  .msg_flag_modified = NULL,
   .msg_padding_size = comp_msg_padding_size,
   .msg_save_hcache  = comp_msg_save_hcache,
   .tags_edit        = comp_tags_edit,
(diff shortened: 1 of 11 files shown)


gahr:
I would rather define the ops entry as NULL and not call it, instead of having a dummy function that always return 0.

Austin-Ray:
@gahr Thank you for taking the time to look at this PR. 

The dummy function is also redundant since `mx_msg_flag_modified()` will return `NULL` if we set the `ops` entry to null.

I will make the correction.

flatcap:
Thanks for creating this PR, but I think my notification system will be more versatile (and fulfil your needs).

PR #1548, opened 2019-01-31
Implemented summary :bind/:macro commands

nocriw:
* **What does this PR do?**

Implement the :bind and :macro command to the set of summary commands.

Issue #162

### DIFF
--- a/icommands.c
+++ b/icommands.c
@@ -5,6 +5,7 @@
  * @authors
  * Copyright (C) 2016 Christopher John Czettel <chris@meicloud.at>
  * Copyright (C) 2018 Richard Russon <rich@flatcap.org>
+ * Copyright (C) 2019 Victor Fernandes <criw@pm.me>
  *
  * @copyright
  * This program is free software: you can redistribute it and/or modify it under
@@ -29,13 +30,16 @@
 #include "mutt.h"
 #include "icommands.h"
 #include "globals.h"
+#include "keymap.h"
 #include "muttlib.h"
+#include "opcodes.h"
 #include "pager.h"
 #include "version.h"
 
 // clang-format off
-static enum CommandResult icmd_set(struct      Buffer *, struct Buffer *, unsigned long, struct Buffer *);
-static enum CommandResult icmd_version(struct  Buffer *, struct Buffer *, unsigned long, struct Buffer *);
+static enum CommandResult icmd_bind   (struct Buffer *buf, struct Buffer *s, unsigned long data, struct Buffer *err);
+static enum CommandResult icmd_set    (struct Buffer *buf, struct Buffer *s, unsigned long data, struct Buffer *err);
+static enum CommandResult icmd_version(struct Buffer *buf, struct Buffer *s, unsigned long data, struct Buffer *err);
 
 /**
  * ICommandList - All available informational commands
@@ -43,6 +47,8 @@ static enum CommandResult icmd_version(struct  Buffer *, struct Buffer *, unsign
  * @note These commands take precedence over conventional mutt rc-lines
  */
 const struct ICommand ICommandList[] = {
+  { "bind",     icmd_bind,     0 },
+  { "macro",    icmd_bind,     1 },
   { "set",      icmd_set,      0 },
   { "version",  icmd_version,  0 },
   { NULL,       NULL,          0 },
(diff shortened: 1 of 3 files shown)


icommands.c
```diff
@@ -98,6 +103,137 @@ enum CommandResult mutt_parse_icommand(/* const */ char *line, struct Buffer *er
...
+    mutt_buffer_printf(err, _("Could not create temporary file %s"), tempfile);
+    return MUTT_CMD_ERROR;
+  }
+
+  for (int i = 0; i < MENU_MAX; i++)
+  {
+    const char *menu_name = mutt_map_get_name(i, Menus);
+
+    if (mutt_str_strcasecmp(buf->data, "all") == 0 ||
+        mutt_str_strcasecmp(buf->data, menu_name) == 0)
```
flatcap: We could make this a bit more efficient by testing for `all` outside of the for loop.

icommands.c
```diff
@@ -98,6 +103,137 @@ enum CommandResult mutt_parse_icommand(/* const */ char *line, struct Buffer *er
...
+  mutt_mktemp(tempfile, sizeof(tempfile));
+  fpout = mutt_file_fopen(tempfile, "w");
+  if (!fpout)
+  {
+    mutt_buffer_printf(err, _("Could not create temporary file %s"), tempfile);
+    return MUTT_CMD_ERROR;
+  }
+
+  const bool all_menus = mutt_str_strcasecmp(buf->data, "all") == 0;
+  for (int i = 0; i < MENU_MAX; i++)
```
Austin-Ray: I think we should find a way to break out this block (and its children blocks) statement into functions to improve readability and make unit testing possible. The control flow is not conducive to quickly understanding what's happening.

icommands.c
```diff
@@ -98,6 +103,137 @@ enum CommandResult mutt_parse_icommand(/* const */ char *line, struct Buffer *er
...
+      fflush(fpout);
+      const long init_size = mutt_file_get_size(tempfile);
+
+      for (map = Keymaps[i]; map; map = next)
+      {
+        char binding[MAX_SEQ];
+        next = map->next;
+        km_expand_key(binding, MAX_SEQ, map);
+
+        if (data == 1 && map->op == OP_MACRO)
```
Austin-Ray: We should replace `data == 1` with a boolean variable to reduce cognitive load when reading the code. The same for line 170.

nocriw:
@Austin-Ray thanks for the feedback :) I did some changes and it would be great if you could take a look!

Austin-Ray:
@nocriw You're welcome :) I apologise for not getting back for several days. The code looks a lot easier to understand!

icommands.c
```diff
@@ -98,6 +103,173 @@ enum CommandResult mutt_parse_icommand(/* const */ char *line, struct Buffer *er
...
+  fpout = mutt_file_fopen(tempfile, "w");
+  if (!fpout)
+  {
+    mutt_buffer_printf(err, _("Could not create temporary file %s"), tempfile);
+    return MUTT_CMD_ERROR;
+  }
+
+  const bool bind = data == 0;
+  const bool dump_all = mutt_str_strcasecmp(buf->data, "all") == 0;
+  for (int i = 0; i < MENU_MAX; i++)
```
Austin-Ray: I feel like this `for` loop exists for the sake of `dump_all`. If we're not dumping everything, we can use `mutt_map_get_value()` instead of looking for an appropriate menu.

icommands.c
```diff
@@ -98,6 +103,173 @@ enum CommandResult mutt_parse_icommand(/* const */ char *line, struct Buffer *er
...
+  }
+
+  mutt_file_fclose(&fpout);
+
+  if (!valid_menu)
+  {
+    mutt_buffer_printf(err, _("%s: no such menu"), buf->data);
+    return MUTT_CMD_ERROR;
+  }
+  else if (mutt_file_check_empty(tempfile) == 1)
```
Austin-Ray: Shouldn't this comparison be `== 0`?

`mutt_file_check_empty()` has three returns: 1 if the file has some in it, 0 if empty, or -1 if error. 

If there are no bindings/macros, the file would be empty, correct?

icommands.c
Austin-Ray: If we didn't worry about `dump_all` support, would the code be simplified to the following?

```C
const int menu_index = mutt_mapping_get_value(...);

// handle no menu error

dump_menu(...);
```

If so, I think we should write the code as:

```C
if (dump_all)
    dump_all_menus(...); // contains our for loops
else
{
    const int menu_index = mutt_mapping_get_value(...);

    // handle no menu error

    dump_menu(...);
}
```

Then the code would have less mixture between low-level and high-level policy, no special special cases in the `for` loop, and we can unit test the `dump_all_menus(...)` method!

nocriw: That's a great improvement to the code quality! I applied these suggestions already

icommands.c
nocriw: Yep, I talked with @flatcap about this the other day on the IRC channel, turns out that `mutt_file_check_empty()` comment is incorrect.

If you take a look at the `mutt_file_check_empty()`'s return condition (`return st.st_size == 0;`) it should return 1 if empty

nocriw:
Also, should we make `dump_bind`and `dump_macro` return something if we want to test them?

icommands.c
Austin-Ray: >  it should return 1 if empty

Ah, I should know better than to assume documentation is 100% accurate.

Austin-Ray:
>  should we make `dump_bind` return something if we want to test it?

That's a good point! I hadn't thought about that.

I suppose good software design would be to have each function do as little as possible. What do you think about the following modifications:

*  `dump_bind` and `dump_macro` return a string instead of writing directly to the file
* `dump_menu` collects the results of `dump_bind`/`dump_macro`. either as a big string or in a collection.
* `dump_all_menus` collect all `dump_menu` results and returns them.
* `icmd_bind` takes the result of `dump_menu` or `dump_all_menus` and handles the file writes.

This has some benefits:

* File writing is performed in a single location
* Less parameters passed to `dump_*` methods
* `dump_*` methods can be unit tested without needing file I/O to check file content

@flatcap Any thoughts?

Austin-Ray:
Whoops! I meant to dismiss my review and don't know how to undo the approval. Just ignore it please.

nocriw:
I like the idea, I suppose this should be as easy as create a Buffer and pass it as parameter to  `dump_menu` & `dump_all_menus` then write it back to the file. Let's see if flatcap has something to say and I'll work on it :)

flatcap:
> I meant to dismiss my review and don't know how to undo the approval

I think if you click on '**Show all reviewers**' (below), then '**Dismiss review**', that'll do it.

flatcap:
We may be overthinking this :-) but...

Yes, I like the idea of passing a `struct Buffer`.
Save a bit of effort by alloc'ing fairly big to start with.

icommands.c
```diff
@@ -98,6 +103,188 @@ enum CommandResult mutt_parse_icommand(/* const */ char *line, struct Buffer *er
...
+ */
+static enum CommandResult icmd_bind(struct Buffer *buf, struct Buffer *s,
+                                    unsigned long data, struct Buffer *err)
+{
+  FILE *fpout = NULL;
+  char tempfile[PATH_MAX];
+  const bool bind = (data == 0);
+
+  if (!MoreArgs(s))
+    mutt_buffer_strcpy(buf, "all");
```
flatcap: Could we just set a `bool all = true` here?

icommands.c
```diff
@@ -98,6 +103,188 @@ enum CommandResult mutt_parse_icommand(/* const */ char *line, struct Buffer *er
...
+
+  mutt_mktemp(tempfile, sizeof(tempfile));
+  fpout = mutt_file_fopen(tempfile, "w");
+  if (!fpout)
+  {
+    mutt_buffer_printf(err, _("Could not create temporary file %s"), tempfile);
+    return MUTT_CMD_ERROR;
+  }
+
+  if (mutt_str_strcasecmp(buf->data, "all") == 0)
```
flatcap: ... and test `bool all` to save a `strcmp()`

nocriw: But then we'll miss the case where the user types `bind all`, in which `all` may end being interpreted as a menu name instead of 'all menus'

flatcap: Sorry, I wasn't clear.
I meant to keep the `strcmp()`, but test the bool first.

Austin-Ray:
> We may be overthinking this :-)

Aye, but the code will be very clear, testable, and maintainable. I'd rather spend time overthinking now than think about it when something breaks.

nocriw:
Changed stuff:

* Joined `dump_macro` & `dump_bind` into `dump_map` since both of them do the same task and now share arguments. Readability in my opinion stills ok, let me know. :)
* Write to file is not done by icmd_bind instead of `dump_map` & `dump_all_menus`.
* Less parameters in dump_*!
* The last review by @flatcap applied.

@flatcap How much do you think we should allocate for the buffer at start?

icommands.c
```diff
@@ -98,6 +104,191 @@ enum CommandResult mutt_parse_icommand(/* const */ char *line, struct Buffer *er
...
+    struct Mapping menu = { buf->data, menu_index };
+    dump_menu(filebuf, &menu, bind);
+  }
+
+  fputs(filebuf->data, fpout);
+  fflush(fpout);
+  mutt_file_fclose(&fpout);
+  mutt_buffer_free(&filebuf);
+
+  if (mutt_file_check_empty(tempfile) == 1)
```
flatcap: We can do this check more efficiently by using `buffer_is_empty()` on `filebuf`
(and moving the check before the file write).

icommands.c
```diff
@@ -98,6 +104,191 @@ enum CommandResult mutt_parse_icommand(/* const */ char *line, struct Buffer *er
...
+  {
+    mutt_buffer_printf(err, _("%s: no %s for this menu"), buf->data,
+                       bind ? "bindings" : "macros");
+    return MUTT_CMD_ERROR;
```
flatcap: This is a very unlikely code path, but if it were to occur, we need to delete the temporary file.

icommands.c
```diff
@@ -98,6 +104,190 @@ enum CommandResult mutt_parse_icommand(/* const */ char *line, struct Buffer *er
...
+
+    if (map->desc)
+    {
+      mutt_buffer_add_printf(buf, "macro %s %s \"%s\" \"%s\"\n", menu->name,
+                             key_binding, tmp.data, map->desc);
+    }
+    else
+    {
+      mutt_buffer_add_printf(buf, "macro %s %s \"%s\"\n", menu->name, key_binding, tmp.data);
+    }
```
flatcap: We need to free `tmp`'s data here.

icommands.c
```diff
@@ -98,6 +104,190 @@ enum CommandResult mutt_parse_icommand(/* const */ char *line, struct Buffer *er
...
+                       dump_all ? "all" : buf->data, bind ? "binds" : "macros");
+    mutt_buffer_free(&filebuf);
+    return MUTT_CMD_ERROR;
+  }
+
+  mutt_mktemp(tempfile, sizeof(tempfile));
+  fpout = mutt_file_fopen(tempfile, "w");
+  if (!fpout)
+  {
+    mutt_buffer_printf(err, _("Could not create temporary file %s"), tempfile);
```
flatcap: We need to free `filebuf` here.

Austin-Ray:
> * Joined `dump_macro` & `dump_bind` into `dump_map` since both of them do the same task and now share arguments. Readability in my opinion stills ok, let me know.

I'm of the opinion that these shouldn't have been combined. `dump_macro` and `dump_bind` have very little overlap so `dump_map` essentially inlined these functions to save a few lines. 

If I'm debugging this portion of the code, my cognitive load is increased because I need to determine if the control flow in one block affects the control flow in another.

Can we get some unit tests?

icommands.c
```diff
@@ -98,6 +104,193 @@ enum CommandResult mutt_parse_icommand(/* const */ char *line, struct Buffer *er
...
+      dump_map(buf, menu, map, true);
+    }
+    else if (!bind && map->op == OP_MACRO)
+    {
+      empty = false;
+      dump_map(buf, menu, map, false);
+    }
+  }
+
+  return empty;
```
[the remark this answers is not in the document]
nocriw: I like this, but trying out I found out that there is a problem since the initial size of the buffer is 4KiB.

So while the buffer isn't reallocated for the first time, the return always translates to `return (4096 - 4096) == 0`

nocriw:
@Austin-Ray I don't mind `dump_map` being combined, however I understand your point and it's gonna be a little easier when debugging, so I separated them again.

I'll try to write some tests. :)

icommands.c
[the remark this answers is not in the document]
Austin-Ray: Ah, that makes sense.

PR #1554, opened 2019-02-07
Separate neomutt objects into separate lib

psyomn:
Just wanted to show a draft before I did anything more. I've recently been given the 
pleasure to work on a refactor which will simplify some regex routines. That part should be ready-ish.

What I wanted to do, was test some neomutt specific code, and I was not able to, due to the way things are bing built right now. I was wondering if an approach in separating the neomutt stuff the way I did would be acceptable? 

Anyway let me know what you think, and I hope this is not too terrible to look at :stuck_out_tongue: 

BTW, if this is ok, and let's say we decide to proceed with this, I'll follow the below guidelines!

(This is incomplete right now, but I'll put more work on it soon, if again, it is OK to do something like this)

* * * 

This should make things more testable

* **What does this PR do?**
Separate neomutt specific objects into separate `libneomutt.a` lib.

     for that)

### DIFF
--- a/Makefile.autosetup
+++ b/Makefile.autosetup
@@ -90,6 +90,37 @@ CLEANFILES+=	$(NEOMUTT) $(NEOMUTTOBJS)
 ALLOBJS+=	$(NEOMUTTOBJS)
 
 ###############################################################################
+# libneomutt
+LIBNEOMUTTOBJS=	account.o addrbook.o alias.o bcache.o browser.o color.o commands.o \
+		complete.o compose.o compress.o conststrings.o context.o copy.o \
+		curs_lib.o edit.o editmsg.o enriched.o enter.o \
+		filter.o flags.o git_ver.o handler.o hdrline.o help.o hook.o \
+		index.o init.o keymap.o mailbox.o main.o menu.o muttlib.o \
+		mutt_account.o mutt_attach.o mutt_body.o mutt_header.o \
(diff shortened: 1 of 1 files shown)


gahr:
What you are proposing [is already there](https://github.com/neomutt/neomutt/blob/40af0561629e106ee6918b8a487082f25373c3ac/Makefile.autosetup#L248-L258). Whenever there's a chance to disentangle a particular .c/.h pair and move it to `mutt/`, we try to do that. This is used exactly for the purpose you mentioned, i.e., testing. The requirement for moving something to `mutt/` is that it doesn't depend from anything outside `mutt/`.

More disentangling and refactoring is welcome :)

flatcap:
Hi @psyomn 

> refactor which will simplify some regex routines.
> That part should be ready-ish.

Great

>  test some ... code

If you just want to play about with your new regex code, you could use the [test-library](https://github.com/neomutt/test-library.git) repo.
The `Makefile` looks for a fully-built NeoMutt in `../neo`

When the code's working, we'd love to see some tests added to NeoMutt's `test/` dir.
These 'acutest' tests are run automatically after every commit (on Travis).

> test some **neomutt** specific code

To test the **NeoMutt** use of regexes, I'm afraid your best bet is to debug NeoMutt.
We're still working hard to make more areas of the code independent.

As for this PR... Sorry, no thanks :-)

psyomn:
Thanks for the info, I'll take a look soon!

PR #1560, opened 2019-02-15
Start to refactor mutt_index_menu()

flatcap:
I'd like to split up the ~2600 line function `mutt_index_menu()`.
This is a bit of preliminary tidying.

- b5ee70728 Refactor index's check macros
  Replace over 200 macros with a function to test the pre-requisites

- 11b31bdf2 reduce scope of loop variables
  All the `for` loops now have their own variable

- c7d530590 reduce scope of rc
  Each `case` now has their own `rc` variable

The only shared variable left is `buf` -- a temporary buffer.
The next step is to move all the large `case` statements into functions.

**Note**: This tidying also needs to happen to `mutt_pager()` too.

### DIFF
--- a/index.c
+++ b/index.c
@@ -1018,8 +1018,6 @@ int mutt_index_menu(void)
   bool tag = false; /* has the tag-prefix command been pressed? */
   int newcount = -1;
   int oldcount = -1;
-  int rc = -1;
-  char *cp = NULL; /* temporary variable. */
   int index_hint;  /* used to restore cursor position */
   bool do_mailbox_notify = true;
   int close = 0; /* did we OP_QUIT or OP_EXIT out of this menu? */
@@ -1397,18 +1395,16 @@ int mutt_index_menu(void)
           }
           else
           {
-            int rc2;
-
             mutt_message(_("Fetching %s from server..."), buf);
-            rc2 = nntp_check_msgid(Context, buf);
-            if (rc2 == 0)
+            int rc = nntp_check_msgid(Context, buf);
+            if (rc == 0)
             {
               e = Context->mailbox->emails[Context->mailbox->msg_count - 1];
               mutt_sort_headers(Context, false);
               menu->current = e->virtual;
               menu->redraw = REDRAW_FULL;
             }
-            else if (rc2 > 0)
+            else if (rc > 0)
               mutt_error(_("Article %s not found on the server"), buf);
           }
         }
@@ -1425,7 +1421,7 @@ int mutt_index_menu(void)
         {
           int oldmsgcount = Context->mailbox->msg_count;
           int oldindex = CUR_EMAIL->index;
-          int rc2 = 0;
+          int rc = 0;
 
           if (!CUR_EMAIL->env->message_id)
           {
@@ -1446,8 +1442,8 @@ int mutt_index_menu(void)
             {
               if (!mutt_hash_find(Context->mailbox->id_hash, ref->data))
               {
-                rc2 = nntp_check_msgid(Context, ref->data);
-                if (rc2 < 0)
+                rc = nntp_check_msgid(Context, ref->data);
+                if (rc < 0)
                   break;
               }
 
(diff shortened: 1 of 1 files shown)


index.c
```diff
@@ -3091,15 +3090,17 @@ int mutt_index_menu(void)
...
+        if (!prereq(Context, menu, CHECK_IN_MAILBOX | CHECK_MSGCOUNT | CHECK_VISIBLE | CHECK_ATTACH))
+          break;
         bool edit;
         if (op == OP_EDIT_RAW_MESSAGE)
         {
-          CHECK_READONLY;
-          /* L10N: CHECK_ACL */
-          CHECK_ACL(MUTT_ACL_INSERT, _("Cannot edit message"));
+          if (!prereq(Context, menu, CHECK_READONLY) ||
+              /* L10N: CHECK_ACL */
```
Austin-Ray: I think we should move the comment above the conditional, it makes it easier to see the entire conditional.

index.c
```diff
@@ -2183,6 +2194,7 @@ int mutt_index_menu(void)
       {
         struct Mailbox *m = NULL;
         OptNews = false;
+        char *cp = NULL;
```
Austin-Ray: This initialisation needs to be moved below. It is within and `#ifdef` block, but `cp` is used outside the `#ifdef`. If this is to scope a NNTP specific `cp` variable, we should add a comment.

index.c
```diff
@@ -1774,24 +1789,23 @@ int mutt_index_menu(void)
...
-        CHECK_VISIBLE;
-        CHECK_READONLY;
-        /* L10N: CHECK_ACL */
-        /* L10N: Due to the implementation details we do not know whether we
-           undelete zero, 1, 12, ... messages. So in English we use
-           "messages". Your language might have other means to express this.
-         */
-        CHECK_ACL(MUTT_ACL_DELETE, _("Cannot undelete messages"));
+        if (!prereq(Context, menu, CHECK_IN_MAILBOX | CHECK_MSGCOUNT | CHECK_VISIBLE | CHECK_READONLY) ||
+            /* L10N: CHECK_ACL */
```
Austin-Ray: I think we should move the comment above the conditional, it makes it easier to see the entire conditional.

index.c
```diff
@@ -1543,32 +1563,32 @@ int mutt_index_menu(void)
...
         break;
+      }
 
         /* --------------------------------------------------------------------
          * 'index' specific commands
          */
 
       case OP_MAIN_DELETE_PATTERN:
+        if (!prereq(Context, menu, CHECK_IN_MAILBOX | CHECK_MSGCOUNT | CHECK_VISIBLE | CHECK_READONLY | CHECK_ATTACH) ||
+            /* L10N: CHECK_ACL */
```
Austin-Ray: I think we should move the comment above the conditional, it makes it easier to see the entire conditional.

flatcap:
Thanks for the review.

> I think we should move the comment above the conditional, it makes it easier to see the entire conditional.

This was a bit awkward.  The `L10N` comment needed to be next to the `_("...")` line, otherwise it wouldn't appear in the `.po` files (after `make update-po`).  I couldn't reverse the order of the `if (X || Y)` because the checks needed to happen in that order.  I settled for splitting the `if` in two.

> This initialisation needs to be moved below. It is within and `#ifdef` block, but `cp` is used outside the `#ifdef`.

Argh!  What a mess!  Worse still I'd added a `{` in a conditional block.
I've rearranged the code to work.

> I think this function's control-flow should be modified to be independent of jump statements.

I'm not sure what to do here; I want to keep the original behaviour:
- The checks need to occur in the order given
- Only one error message should be shown
- If an error occurs, we need to `mutt_flushinp()`

Suggestions welcome.

Austin-Ray:
> I'm not sure what to do here; I want to keep the original behaviour

The first though that comes to mind is adding a prefix to the conditional so we have `result && <rest of conditional>` to keep the original behaviour. It makes the conditional longer, but will remove the control-flow jumps.

Issue #1579, opened 2019-02-28
Neomutt segfaults when piping message


jindraj:
## Expected Behaviour

not to segfault

## Actual Behaviour

segfaults

## Steps to Reproduce

according to git bisect the commit which introduced this behaviour is 2dc7b33f0e30b300086ebe53382a045016374dbd
and run urlview or anything else which pipes message

## How often does this happen?

- Always when message is piped (when urlview is called or other macros with pipe-message) - tested on urlview.

## When did it start to happen?

- When I upgraded from HEAD bb8cda3b8cb687c7e9f2d6e9154dafd0102e363f to latest commit today
now running last good commit… ff590a729e12f5feee15dbb74505292fe73fa342
## NeoMutt Version

```
NeoMutt 20180716-1118-ff590a
Copyright (C) 1996-2016 Michael R. Elkins and others.
NeoMutt comes with ABSOLUTELY NO WARRANTY; for details type 'neomutt -vv'.
NeoMutt is free software, and you are welcome to redistribute it
under certain conditions; type 'neomutt -vv' for details.

System: Darwin 18.2.0 (x86_64)
ncurses: ncurses 5.7.20081102 (compiled with 5.7.20081102)
libiconv: 1.11
libidn: 1.35 (compiled with 1.35)
GPGme: 1.12.0
hcache backends: tokyocabinet, lmdb

Compiler:
Apple LLVM version 10.0.0 (clang-1000.10.44.4)
Target: x86_64-apple-darwin18.2.0
Thread model: posix
InstalledDir: /Library/Developer/CommandLineTools/usr/bin

Configure options: --prefix=/usr/local/Cellar/neomutt/HEAD-ff590a7 --enable-gpgme --with-gpgme=/usr/local/opt/gpgme --gss --lmdb --notmuch --sasl --tokyocabinet --with-ssl=/usr/local/opt/openssl --with-ui=ncurses

Compilation CFLAGS: -g -O2 -std=c99 -D_ALL_SOURCE=1 -D_GNU_SOURCE=1 -D__EXTENSIONS__ -I/usr/local/opt/gpgme/include -I/usr/local/Cellar/neomutt/HEAD-ff590a7/include -DNCURSES_WIDECHAR -I/usr/local/opt/openssl/include

Default options:
  +attach_headers_color +compose_to_sender +compress +cond_date +debug
  +encrypt_to_self +forgotten_attachments +forwref +ifdef +imap +index_color
[...]

PR #1580, opened 2019-02-28
Fix mutt_pipe_message() and mutt_print_message()

flatcap:
A discrepancy in the prototype and the function lead to these functions crashing.

Fixes: #1579


flatcap:
> Can confirm you changed all instances of `mutt_pipe_message()` and `mutt_print_message()` to have correct number of parameters.

Yes, two calls of each.
Now that the prototype's correct, the compiler would fail if there were others.

PR #1585, opened 2019-03-04
Ifdef envvars

jindraj:
* **What does this PR do?**
Adds ability for ifdef to work with environment variables.

chaotic asciinema…
[asciicast](https://asciinema.org/a/pQAXhEzRl4cPxzSbaI5TeIo46)

updated.

I'm not sure. I build it if it passes tests we will see.

Didn't seem to be necessary.

   - Code follows the [style guide]
I hope so. I tried to keep style I've seen around. Maybe braces are not necessary:
https://neomutt.org/dev/coding-style#braces-placement but I would let it until some code style review.

No existing issue. But "request" on IRC. Actually fekir was asking same or similar question few days ago.
```
-- Mon, 04 Mar 2019 --
11:20:37            fekir | Is it possible to check in the config file is neomutt is being used interactively or not?
11:21:09            fekir | I would like not to load some configs (like colors), if I#M writing and sending an email directly from bash
11:35:20          jindraj | fekir: maybe using ifdef but I'm still unable prove it 😃
11:38:42            fekir | jindraj: thanks, did not know the ifdef, though I'm unable to see if there is something like "interactive"
11:44:37             gahr | fekir: maybe you could add a "set my_noninteractive=1" when running from a script
11:48:09          jindraj | fekir: I was thinking about `ifdef $PS1 "source config_files_you_want_to_have_in_interactive_mode"`. But it seems like ifdef works with neomutt variables only.
11:56:23            fekir | gahr: how should I accomplish that? Suppose my current command is " echo 'Hello' | neomutt -s 'mail' abc@mail.com -a file.txt". AFAIK I cannot add a config
                          | variable from the cli
```

### DIFF
--- a/init.c
+++ b/init.c
@@ -1184,6 +1184,27 @@ warn:
 }
 
 /**
+ * is_function - Is the argument a neomutt function?
+ * @param name  Command name to be searched for
+ * @retval true  Function found
+ * @retval false Function not found
+*/
+static bool is_function(const char *name)
+{
+  for (int i = 0; i < MENU_MAX; i++)
+  {
+    const struct Binding *b = km_get_table(Menus[i].value);
+    if (!b)
+      continue;
+
+    for (int j = 0; b[j].name; j++)
+      if (mutt_str_strcmp(name, b[j].name) == 0)
+        return true;
+  }
+  return false;
+}
+
+/**
  * parse_ifdef - Parse the 'ifdef' and 'ifndef' commands - Implements ::command_t
  *
  * The 'ifdef' command allows conditional elements in the config file.
@@ -1203,54 +1224,13 @@ static enum CommandResult parse_ifdef(struct Buffer *buf, struct Buffer *s,
 
   mutt_extract_token(buf, s, 0);
 
-  /* is the item defined as a variable? */
-  struct HashElem *he = cs_get_elem(Config, buf->data);
-  bool res = (he != NULL);
-
-  /* is the item a compiled-in feature? */
-  if (!res)
-  {
-    res = feature_enabled(buf->data);
-  }
-
-  /* or a function? */
-  if (!res)
-  {
-    for (int i = 0; !res && (i < MENU_MAX); i++)
-    {
-      const struct Binding *b = km_get_table(Menus[i].value);
-      if (!b)
-        continue;
-
-      for (int j = 0; b[j].name; j++)
-      {
-        if (mutt_str_strcmp(buf->data, b[j].name) == 0)
-        {
-          res = true;
-          break;
-        }
-      }
-    }
-  }
-
-  /* or a command? */
-  if (!res)
-  {
-    for (int i = 0; Commands[i].name; i++)
-    {
-      if (mutt_str_strcmp(buf->data, Commands[i].name) == 0)
-      {
-        res = true;
-        break;
-      }
-    }
-  }
-
-  /* or a my_ var? */
-  if (!res)
-  {
-    res = !!myvar_get(buf->data);
-  }
+  // is the item defined as:
+  bool res = cs_get_elem(Config, buf->data) // a variable?
+             || feature_enabled(buf->data)  // a compiled-in feature?
+             || is_function(buf->data)      // a function?
+             || mutt_command_get(buf->data) // a command?
+             || myvar_get(buf->data)        // a my_ variable?
+             || mutt_str_getenv(buf->data); // an environment variable?
 
   if (!MoreArgs(s))
   {
@@ -2592,7 +2572,6 @@ static enum CommandResult parse_unsubscribe_from(struct Buffer *buf, struct Buff
 }
 #endif
 
-#ifdef USE_LUA
 /**
  * mutt_command_get - Get a Command by its name
  * @param s Command string to lookup
@@ -2606,7 +2585,6 @@ const struct Command *mutt_command_get(const char *s)
       return &Commands[i];
   return NULL;
 }
-#endif
 
 #ifdef USE_LUA
 /**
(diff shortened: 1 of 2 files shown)


init.c
```diff
@@ -1253,6 +1253,15 @@ static enum CommandResult parse_ifdef(struct Buffer *buf, struct Buffer *s,
     res = !!myvar_get(buf->data);
   }
 
+  /* or an environment variable? */
+  if (!res)
+  {
+    if (mutt_str_getenv(buf->data) != NULL)
+    {
+      res = true;
+    }
```
flatcap: Actually, we can reduce this to your original suggestion:
```c
res = !!mutt_str_getenv(buf->data);
```

`mutt_str_getenv()` returns a pointer which we `!!` **not** twice (a common C trick to boolify a pointer).

jindraj: I've seen that syntax around but didn't know what it does. It's up to you and I'm all for it. It's shorter and it's consistent with my_ vars condition above.

init.c
```diff
@@ -1253,6 +1253,12 @@ static enum CommandResult parse_ifdef(struct Buffer *buf, struct Buffer *s,
     res = !!myvar_get(buf->data);
   }
 
+  /* or an environment variable? */
+  if (!res)
+  {
+    res = !!mutt_str_getenv(buf->data);
```
Austin-Ray: We can combine this `if` statement with the previous one so that it is:

```C
/* or a my_ var or environment variable? */
  if (!res)
  {
    res = !!myvar_get(buf->data) || !!mutt_str_getenv(buf->data);
  }
```

Saves some duplication

jindraj: done

gahr:
I can't help seeing the potential to refactor the whole `res` assignment to a single:

```c
const bool res =
     is_config_variable(buf->data)
  || is_enabled_feature(buf->data)
  || is_function(buf->data)
  || is_command(buf->data)
  || is_myvar(buf->data)
  || is_envvar(buf->data);
```

flatcap:
Yes, that would be pretty tidy.

flatcap:
This pull request **introduces 3 alerts** when merging bb86a2f60b2e7e6279ea4f5a913c5008922c56e5 into 89f5688ccad4bccb12f06ec3138a141122aa7b77 - [view on LGTM.com](https://lgtm.com/projects/g/neomutt/neomutt/rev/pr-6a5457f154a8574445850713845d5ef096b592c5)

**new alerts:**

* 3 for Constant return type

---

*Comment posted by [LGTM.com](https://lgtm.com)*

init.c
```diff
@@ -1184,6 +1184,67 @@ static enum CommandResult parse_group(struct Buffer *buf, struct Buffer *s,
...
+static bool is_function(const char * name)
+{
+  bool res = false;
+  for (int i = 0; i < MENU_MAX; i++)
+  {
+    const struct Binding *b = km_get_table(Menus[i].value);
+
+    for (int j = 0; b && !res && b[j].name; j++)
+      if (mutt_str_strcmp(name, b[j].name) == 0)
+        res = true;
```
gahr: return true;

init.c
```diff
@@ -1184,6 +1184,67 @@ static enum CommandResult parse_group(struct Buffer *buf, struct Buffer *s,
...
+  }
+  return res;
```
gahr: return false;

init.c
```diff
@@ -1184,6 +1184,67 @@ static enum CommandResult parse_group(struct Buffer *buf, struct Buffer *s,
...
+{
+  return mutt_str_getenv(name);
+}
+
+/**
+ * is_enabled_feature - return bool if argument is an enabled feature
+*/
+static bool is_enabled_feature(const char * name)
+{
+  return feature_enabled(name);
```
gahr: you can use `feature_enabled(name)` directly in `parse_ifdef` without the need to wrap it

init.c
```diff
@@ -1184,6 +1184,67 @@ static enum CommandResult parse_group(struct Buffer *buf, struct Buffer *s,
...
+
+/**
+ * is_function - return bool if argument is a neomutt command
+*/
+static bool is_command(const char * name)
+{
+  bool res = false;
+  for (int i = 0; !res && Commands[i].name; i++)
+    if (mutt_str_strcmp(name, Commands[i].name) == 0)
+      res = true;
```
gahr: return true;

init.c
```diff
@@ -1184,6 +1184,67 @@ static enum CommandResult parse_group(struct Buffer *buf, struct Buffer *s,
...
+  return res;
```
gahr: return false;

init.c
jindraj: when I change `res = true;` to `return true;` I still need res for the for condition.

jindraj: Oh, got it. I can return right in the loop. So it's not necessary for the loop to check the condition again, escape loop and after that return.

jindraj:
After leaving `myvar_get` and `mutt_str_getenv` without `!!` seems to work well. Tested on both functions with ifdef/ifndef in config file.

gahr:
Great, thanks! I've tested with the following config file, and the expected messages (the ones without `foo`) are printed as expected:

```
set my_bar = moo

ifdef allow_ansi       "echo allow_ansi"
ifdef allow_ansi_foo   "echo allow_ansi_foo"
ifdef quasi_delete     "echo quasi_delete"
ifdef quasi_delete_foo "echo quasi_delete_foo"
ifdef break-thread     "echo break-thread"
ifdef break-thread_foo "echo break-thread_foo"
ifdef unignore         "echo unignore"
ifdef unignore_foo     "echo unignore_foo"
ifdef my_bar           "echo my_bar"
ifdef my_bar_foo       "echo my_bar_foo"
ifdef LOGNAME          "echo LOGNAME"
ifdef LOGNAME_FOO      "echo LOGNAME_FOO"
```

PR #1595, opened 2019-03-08
Change hcache version when Email struct change

sileht:
On imap when some mails are expunged, many mail disapear from current
mailbox. It's just a rendering issue, when the mailbox is reopened all
expected mails are there.

After a git bisect session, I got the culprit:
b60b10f

This change rename Header to Email struct in the list of structs 
that increases hcache version.

### DIFF
--- a/hcache/hcachever.sh
+++ b/hcache/hcachever.sh
@@ -46,7 +46,7 @@ getstruct () {
   done
 
   case $STRUCT in
-    Address|Body|Buffer|Envelope|Header|ListNode|Parameter)
+    Address|Body|Buffer|Email|Envelope|ListNode|Parameter)
       BODY=`cleanbody "$BODY"`
       echo "$STRUCT: $BODY"
     ;;


flatcap:
Thanks.  It's a bit odd, though.
The `typedef`'d type is 16 bits, so it ought to be big enough.
I wonder what I'm missing.

flatcap:
I can't seem to repeat the problem.
How repeatable is it?
What steps are you taking?

Can you try disabling the header cache (I'm wondering if my changing the size of `Email` is the cause).

Last night, I went through all 280 uses of `Email.security`, but they all seem to just set/clear/test the `SEC_*` or `APPLICATION_*` flags.  I can't see anything strange.

sileht:
I disable header/message cache and the issue is gone.
Then I backup and removed my cache dir and re-enable header/message cache and it seems to works too. 
And I finish my restoring my cache dir and break again.

So I guess it's related to a cache created before the SecurityFlags size change.

flatcap:
> So I guess it's related to a cache created before the SecurityFlags size change.

Great.  Thanks for the info.  I'll have a think.

I might just commit your fix until I understand what's going wrong with the hcache calculations, or maybe reorder the fields to make sure the cache is invalidated.

sileht:
I take a look to hcache implementation and it assume Email structure never change:
https://github.com/neomutt/neomutt/blob/master/hcache/serialize.c#L656-L657

And there is nothing that validate that what was loaded from the binary blob is valid.

gahr:
> And there is nothing that validates that what was loaded from the binary blob is valid.
There is something, but it's weak. The [hcachever.sh](https://github.com/neomutt/neomutt/blob/master/hcache/hcachever.sh) script is used to compute a hash of the structure, but it only takes into consideration a few things, like field names and orderd. Admittedly, at a first glance it looks like a change in the bitfield would be noticed.

gahr:
Oh, hcachever.sh simply doesn't take `struct Email` into consideration...

gahr:
I propose the following to fix this issue:

```diff
diff --git a/hcache/hcachever.sh b/hcache/hcachever.sh
index 495864e4a..48df0a3d7 100755
--- a/hcache/hcachever.sh
+++ b/hcache/hcachever.sh
@@ -46,7 +46,7 @@ getstruct () {
   done

   case $STRUCT in
-    Address|Body|Buffer|Envelope|Header|ListNode|Parameter)
+    Address|Body|Buffer|Email|Envelope|Header|ListNode|Parameter)
       BODY=`cleanbody "$BODY"`
       echo "$STRUCT: $BODY"
     ;;
```

sileht:
Thx @gahr that's work well.

flatcap:
Ah! Well spotted!
**That**'s the problem :-)
Now it makes sense.

But the exact change we need is `s/Header/Email/`
The struct was renamed, not missing.

Issue #1591, opened 2019-03-06
Implement unbind command


jindraj:
## What would you like to be able to do?
PR: #1599
Create unbind command similar to original [unbind patch](https://web.archive.org/web/20160313160541if_/http://home.uchicago.edu:80/~dgc/mutt/#unbind
) by David Champion.

Could be useful to:
1. **new users** won't be afraid too much of destructing their messages.
2. **advanced users** could use it for multiple rcfiles each with different bindings.

## What needs to change in NeoMutt?
  * create new command, parse function and some logic about bindings
  * parse_keymap (f.e.: reflect unbind command name in error messages)

## Have you tried any alternative methods?
N/A
## Additional Info
- Shoud `unbind` affect macros as well or is it better to implement `unmacro` command?

> 14:37:05          flatcap | As bind and macro are separate, I think unbind and unmacro should be too.

### binding which should be kept:
 - `:` command prompt
 - `q` quitting neomutt 
 … but neomutt won't be usable. missing `m` compose new message,`y` send message  and many other core keys. unbinding all sound cool with complete set of different bindings.

### Syntax
`unbind <menu-name[,…]|*> [<key_sequence>]`

### Examples
- unbind `@` from one, multiple or all menu-names
`unbind index @`
`unbind index,pager @`
`unbind * @`
- unbind all bindings from one, multiple or all menu-names
`unbind index`
`unbind index,pager`
`unbind *`

jindraj:
first steps here: b0c2a7ee30c35b4e7ec1f3c6ebb257300131a021

## what's new
- new command unbind
- `bool bind` to `char *cmd` in `parse_keymap` function  to reflect command name in error messages
- keeps `q` and `:` bound
Keep them bound only when `unbind <menu-name>` (unbinding all key_sequences)
But unbind them when `unbind <menu-name> q` or `unbind <menu-name> :`

## next
- over all keys from keymap
- iterate over all menus when `*` is specified as `menu-name`

jindraj:
It seems that since the syntax of both unbind and unmacro commands is the same It should be sufficient to have only one parse function.

flatcap:
OK, now we've got the syntax, here's a rough plan of how the code should look:

- extract first param (either "index,pager" or "*")
- is it "*"?
  set all_menus = true
- else parse menus (this might fail)
- extract next param (this is optional)
  - is it "*"
    set all_keys = true
  - else key = string
  - is all_menus set AND (key = "q" or ":")
    error
[...]

PR #1599, opened 2019-03-12
Devel/unbind

jindraj:
* **What does this PR do?**
implements `unbind` and `unmacro` commands
```
unbind <menu-name[,…]|*> [<key_sequence>]
unmacro <menu-name[,…]|*> [<key_sequence>]
```
N/A

doxygen documentation included
neomutt documment

     for that)
in progress, I'm getting familiar with the xml format.

not yet (travis failing with parenthesis around operand MUTT_UNBIND in `data & MUTT_UNBIND`

yes

tried to follow

#1591

### DIFF
--- a/keymap.c
+++ b/keymap.c
@@ -573,7 +573,7 @@ int km_dokey(int menu)
   int pos = 0;
   int n = 0;
 
-  if (!map)
+  if (!map && (menu != MENU_EDITOR))
     return retry_generic(menu, NULL, 0, 0);
 
   while (true)
@@ -666,6 +666,9 @@ int km_dokey(int menu)
       if (func)
         continue;
     }
+
+    if (!map)
+      return tmp.op;
 
     /* Nope. Business as usual */
     while (LastKey > map->keys[pos])
@@ -1296,6 +1299,152 @@ enum CommandResult mutt_parse_bind(struct Buffer *buf, struct Buffer *s,
   }
   FREE(&key);
   return rc;
+}
+
+/**
+ * parse_menu - Parse menu-names into an array
+ * @param menu     Array for results
+ * @param s        String containing menu-names
+ * @param err      Buffer for error messages
+ *
+ * Expects to see: <menu-string>[,<menu-string>]
+ */
+static void *parse_menu(int *menu, char *s, struct Buffer *err)
+{
+  char *menu_names_dup = mutt_str_strdup(s);
+  char *menu_name = NULL;
+
+  while ((menu_name = strsep(&menu_names_dup, ",")))
+  {
+    int value = mutt_map_get_value(menu_name, Menus);
+    if (value == -1)
+    {
+      mutt_buffer_printf(err, _("%s: no such menu"), menu_name);
+      FREE(&menu_names_dup);
+      return NULL;
+    }
+    else
+      menu[value] = true;
+  }
+
+  FREE(&menu_names_dup);
+  return NULL;
+}
+
+/**
+ * km_unbind_all - Free all the keys in the supplied Keymap
+ * @param map  Keymap mapping
+ * @param mode Undo bind or macro, e.g. #MUTT_UNBIND, #MUTT_UNMACRO
+ *
+ * Iterate through Keymap and free keys defined either by "macro" or "bind".
+ */
+static void km_unbind_all(struct Keymap **map, unsigned long mode)
+{
+  struct Keymap *next = NULL;
+  struct Keymap *first = NULL;
+  struct Keymap *last = NULL;
+  struct Keymap *cur = *map;
+
+  while (cur)
+  {
+    next = cur->next;
+    if (((mode & MUTT_UNBIND) && !cur->macro) || ((mode & MUTT_UNMACRO) && cur->macro))
+    {
+      FREE(&cur->macro);
+      FREE(&cur->keys);
+      FREE(&cur->desc);
+      FREE(&cur);
+    }
+    else if (!first)
+    {
+      first = last = cur;
+    }
+    else if (last)
+    {
+      last->next = cur;
+      last = cur;
+    }
+    else
+    {
+      last = cur;
+    }
+    cur = next;
+  }
+
+  if (last)
+    last->next = NULL;
+
+  *map = first;
+}
+
+/**
+ * mutt_parse_unbind - Parse the 'unbind' command - Implements ::command_t
+ *
+ * Command unbinds:
+ * - one binding in one menu-name
+ * - one binding in all menu-names
+ * - all bindings in all menu-names
+ *
+ * unbind `<menu-name[,...]|*>` [`<key_sequence>`]
+ */
+enum CommandResult mutt_parse_unbind(struct Buffer *buf, struct Buffer *s,
+                                     unsigned long data, struct Buffer *err)
+{
+  int menu[MENU_MAX] = { 0 };
+  bool all_keys = false;
+  char *key = NULL;
+
+  mutt_extract_token(buf, s, 0);
+  if (mutt_str_strcmp(buf->data, "*") == 0)
+  {
+    for (int i = 0; i < MENU_MAX; i++)
+      menu[i] = 1;
+  }
+  else
+    parse_menu(menu, buf->data, err);
+
+  if (MoreArgs(s))
+  {
+    mutt_extract_token(buf, s, 0);
+    key = buf->data;
+  }
+  else
+    all_keys = true;
+
+  if (MoreArgs(s))
+  {
+    const char *cmd = (data & MUTT_UNMACRO) ? "unmacro" : "unbind";
+
+    mutt_buffer_printf(err, _("%s: too many arguments"), cmd);
+    return MUTT_CMD_ERROR;
+  }
+
+  for (int i = 0; i < MENU_MAX; i++)
+  {
+    if (menu[i] != 1)
+      continue;
+    if (all_keys)
+    {
+      km_unbind_all(&Keymaps[i], data);
+      km_bindkey("<enter>", MENU_GENERIC, OP_GENERIC_SELECT_ENTRY);
+      km_bindkey("<return>", MENU_GENERIC, OP_GENERIC_SELECT_ENTRY);
+      km_bindkey("<enter>", MENU_MAIN, OP_DISPLAY_MESSAGE);
+      km_bindkey("<return>", MENU_MAIN, OP_DISPLAY_MESSAGE);
+      km_bindkey("<backspace>", MENU_EDITOR, OP_EDITOR_BACKSPACE);
+      km_bindkey("\177", MENU_EDITOR, OP_EDITOR_BACKSPACE);
+      km_bindkey(":", MENU_GENERIC, OP_ENTER_COMMAND);
+      km_bindkey(":", MENU_PAGER, OP_ENTER_COMMAND);
+      if (i != MENU_EDITOR)
+      {
+        km_bindkey("?", i, OP_HELP);
+        km_bindkey("q", i, OP_EXIT);
+      }
+    }
+    else
+      km_bindkey(key, i, OP_NULL);
+  }
+
+  return MUTT_CMD_SUCCESS;
 }
 
 /**
--- a/doc/manual.xml.head
+++ b/doc/manual.xml.head
@@ -3865,6 +3865,23 @@ set alias_file=~/.mail_aliases
         <arg choice="plain">
           <replaceable class="parameter">function</replaceable>
         </arg>
+        <command>unbind</command>
+        <group choice="req">
+          <arg choice="plain">
+            <replaceable class="parameter">*</replaceable>
+          </arg>
+          <arg choice="plain">
+            <replaceable class="parameter">map</replaceable>
+          </arg>
+          <arg choice="opt" rep="repeat">
+            <replaceable class="parameter">,map</replaceable>
+          </arg>
+        </group>
+        <group choice="opt">
+          <arg choice="plain">
+            <replaceable class="parameter">key</replaceable>
+          </arg>
+        </group>
       </cmdsynopsis>
       <para>
         This command allows you to change the default key bindings (operation
@@ -3875,6 +3892,12 @@ set alias_file=~/.mail_aliases
         Multiple maps may be specified by separating them with commas (no
         additional whitespace is allowed). The currently defined maps are:
       </para>
+      <note>
+        <para>
+          Missing key sequence in unmacro command means unmacro all macros in menus
+          given in <emphasis>menu</emphasis>.
+        </para>
+      </note>
       <anchor id="maps" />
       <variablelist>
         <varlistentry>
@@ -4174,6 +4197,15 @@ bind index g  noop
 bind index gg first-entry
 </screen>
 
+        <para>
+          The same is also possible using <literal>unbind</literal>.
+        </para>
+
+<screen>
+unbind index g
+bind index gg first-entry
+</screen>
+
       </sect2>
     </sect1>
 
@@ -4314,6 +4346,24 @@ folder-hook work "set sort=threads"
         <arg choice="opt">
           <replaceable class="parameter">description</replaceable>
         </arg>
+        <command>unmacro</command>
+        <group choice="req">
+          <arg choice="plain">
+            <replaceable class="parameter">*</replaceable>
+          </arg>
+          <arg choice="plain">
+            <replaceable class="parameter">map</replaceable>
+          </arg>
+          <arg choice="opt" rep="repeat">
+            <replaceable class="parameter">,map</replaceable>
+          </arg>
+        </group>
+        <group choice="opt">
+          <arg choice="plain">
+            <replaceable class="parameter">key</replaceable>
+          </arg>
+        </group>
+
       </cmdsynopsis>
       <para>
         Macros are useful when you would like a single key to perform a series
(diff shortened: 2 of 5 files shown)


jindraj:
I would appreciate any comments to make code better and easier. Also help to choose good doxygen description parameters.

keymap.c
```diff
@@ -1298,6 +1298,186 @@ enum CommandResult mutt_parse_bind(struct Buffer *buf, struct Buffer *s,
...
+  struct Keymap *cur, *next, *first, *last;
+
+  first = last = NULL;
+  cur = *map;
+
+  while (cur != NULL)
+  {
+    next = cur->next;
+    if ((mode & MUTT_UNBIND && cur->macro == NULL) ||
+        (mode & MUTT_UNMACRO && cur->macro != NULL))
```
Austin-Ray: Wrap the individual conditions in `()` so we end up with `(mode & MUTT_UNBIND) && (cur->macro == NULL)...`

keymap.c
```diff
@@ -1298,6 +1298,186 @@ enum CommandResult mutt_parse_bind(struct Buffer *buf, struct Buffer *s,
...
+        {
+          km_bindkey("<return>", MENU_GENERIC, OP_GENERIC_SELECT_ENTRY);
+          km_bindkey("<enter>", MENU_GENERIC, OP_GENERIC_SELECT_ENTRY);
+          km_bindkey(":", MENU_GENERIC, OP_ENTER_COMMAND);
+          km_bindkey("q", MENU_GENERIC, OP_QUIT);
+          km_bindkey("?", MENU_GENERIC, OP_HELP);
+        }
+      }
+    }
+    else if (all_menus == false && all_keys == true)
```
Austin-Ray: Unless I'm mistaken, this and the above block can be merged into:

```C
if (all_keys)
{
  // Change the name of this one because I'm too lazy to think of a good one right now :)
  int upper_bound = all_menus ? MENU_MAX : nummenus;

  for (int i = 0; i < upper_bound; ++i)
  {
     km_unbind_all(&Keymaps[i],data);
     if ((data & MUTT_UNBIND) == MUTT_UNBIND && i == MENU_GENERIC)
     {
       km_bindkey("<return>", MENU_GENERIC, OP_GENERIC_SELECT_ENTRY);
       km_bindkey("<enter>", MENU_GENERIC, OP_GENERIC_SELECT_ENTRY);
       km_bindkey(":", MENU_GENERIC, OP_ENTER_COMMAND);
       km_bindkey("q", MENU_GENERIC, OP_QUIT);
       km_bindkey("?", MENU_GENERIC, OP_HELP);
     }
   }
}
```

keymap.c
```diff
@@ -1298,6 +1298,186 @@ enum CommandResult mutt_parse_bind(struct Buffer *buf, struct Buffer *s,
...
+          km_bindkey("?", MENU_GENERIC, OP_HELP);
+        }
+      }
+    }
+    else if (all_menus == true && all_keys == false)
+    {
+      for (int i = 0; i < MENU_MAX; ++i)
+        km_bindkey(key, i, OP_NULL);
+    }
+    else // unbind key from specified menus
```
Austin-Ray: Similarly, this can be merged into:

```C
else {
    int upper_bound = all_menus ? MENU_MAX : nummenus;
    for (int i = 0; i < upper_bound; ++i) {
        target_menu = all_menus ? menu[i] : i;
        km_bindkey(key, target_menu, OP_NULL);
    }
}
```

jindraj: this segfaults in km_bindkey(). I'll take a look at it tomorrow.

Austin-Ray: @jindraj that is my fault. i made a mistake w/ `target_menu =...`. correct line is `target_menu = all_menus ? i : menu[i];`

I'll triple check my messages now to save you any headache  :)

keymap.c
jindraj: I thought I replied before to this too, but apparently not.
After merging those blocks, it behaves really weird. unbind and unmacro removes all keys (it's good to se with unmacro and macro command to show what has left. The keys in km_bindkeys() are not bound , hits the condition.
And then for example after `unbind index` trying to get into command prompt `:`  fills the whole command prompt from left to right with colons.

This could be probably the reason. 
`        km_unbind_all(&Keymaps[i],data);`
`        km_unbind_all(&Keymaps[menu[i]],data);`

jindraj: should be solved in the latest commit…

keymap.c
```diff
@@ -1298,6 +1298,173 @@ enum CommandResult mutt_parse_bind(struct Buffer *buf, struct Buffer *s,
...
+          km_bindkey("q", MENU_GENERIC, OP_EXIT);
+          km_bindkey("q", MENU_PAGER, OP_EXIT);
+          km_bindkey("?", MENU_GENERIC, OP_HELP);
+        }
+      }
+    }
+    else
+    {
+      int menus = all_menus ? MENU_MAX : nummenus;
+      for (int i = 0; i < menus; ++i) {
```
Austin-Ray: After making the changes, both parts of the `if` have the same basic format. We can change extract the `for` loop so the conditional is within in and have something like:

```C
int menus = all_menus ? MENU_MAX : nummenus;

for (int i = 0; i < menus; ++i) 
{
    int map = all_menus ? i : menu[i];

    if (all_keys) 
    {
        km_unbind_all(&Keymaps[map],data);
        if ((data & MUTT_UNBIND) == MUTT_UNBIND && i == MENU_GENERIC)
        {
          km_bindkey("<return>", MENU_GENERIC, OP_GENERIC_SELECT_ENTRY);
          km_bindkey("<enter>", MENU_GENERIC, OP_GENERIC_SELECT_ENTRY);
          km_bindkey(":", MENU_GENERIC, OP_ENTER_COMMAND);
          km_bindkey("q", MENU_GENERIC, OP_EXIT);
          km_bindkey("q", MENU_PAGER, OP_EXIT);
          km_bindkey("?", MENU_GENERIC, OP_HELP);
        }
    }
    else
    {
        km_bindkey(key, map, OP_NULL);
    }
}
```

Also, since `if (MoreArgs(s))` has a `return` in it, we can remove the `else` and have the `for` loop directly after. Reduces some indentation and number of braces.

keymap.c
```diff
@@ -1298,6 +1298,173 @@ enum CommandResult mutt_parse_bind(struct Buffer *buf, struct Buffer *s,
...
+        int target_menu = all_menus ? i : menu[i];
+        km_bindkey(key, target_menu, OP_NULL);
+      }
+    }
+  }
+
+  return r;
```
Austin-Ray: Since this the only place `r` is used, we can remove the variable and just return `MUTT_CMD_SUCCESS`

keymap.c
```diff
@@ -1298,6 +1298,166 @@ enum CommandResult mutt_parse_bind(struct Buffer *buf, struct Buffer *s,
...
+    menu[i] = mutt_map_get_value(s, Menus);
+    if (menu[i] == -1)
+      mutt_buffer_printf(err, _("%s: no such menu"), s);
+    i++;
+    if (q)
+      s = q + 1;
+  }
+  *nummenus = i;
+  return NULL;
+}
```
Austin-Ray: Stealing from #1543 (sorry, we have to use this old way until that's merged), we can remove the manual string splitting and `strsep()` instead.

```C
int i = 0;

char *menu_names_dup = mutt_str_strdup(s);
char *menu_name = strsep(&menu_names_dup, ",");

while (menu_name)
{
   menu[i] = mutt_map_get_value(menu_name, Menus);
   if (menu[i] == -1)
      mutt_buffer_printf(err, _("%s: no such menu"), menu_name);

   i++;
   menu_name = strsep(&menu_names_dup, ","); // This is keep assignment out of the conditional.
}

FREE(&menu_names_dup);

*nummenus = i;
return NULL;
```

This makes it a bit easier to understand what's happening (in my opinion of course, if you disagree please voice that.) However, this keeps two parts of the code I do not like:

First,
```C
menu[i] = mutt_map_get_value(menu_name, Menus);
if (menu[i] == -1)
   mutt_buffer_printf(err, _("%s: no such menu"), menu_name);
```
With this program flow, it's possible for `menu` to be `[-1, -1, -1, -1]` if there are four non-existent menu names. A static array does not seem like a good data structure choice here. We need something that grows so we can have code such as:

```C
int value = mutt_map_get_value(menu_name, Menus);

if (value != -1)
{
   // Add to data structure.
}
```

Second, what happens if we specify more menus than room in the `menu`?

We will run out of space in crash since we exceeded our allocated memory. 

The current way it's code has the opposite problem, if we specify more menus than `maxmenus`, it just stops parsing them. Try it: `unbind test,test,test,test,test,test,test,test,test,test,test,test,test,test,test,test,test,test,test,index q` doesn't work, but `unbind index q` does! A fixed array is not the best choice of a data structure in this case. We should use a list. I haven't had to use Neomutt's list implementation so this will be a learning experience for both of us :D 

@flatcap are you available to tutor us?

After we swap to a dynamic data structure, `i`, `nummenus`, and `maxmenus` are unnecessary since we can query the data structure for its size.

flatcap: >  if there are four non-existent menu names.

We seem to have lost a `return` statement after the error message.

> if we specify more menus than `maxmenus`, it just stops parsing them

An alternative way to do the parsing might be:
- create array of size `MENU_MAX`
- initialise to zero
- `x = mutt_map_get_value(...)`
- `array[x] = true;`

So rather than an array:
- `3, 5, 2, -1`
we have:
- `0, 1, 1, 0, 1, 0, 0, 0, 0`

Austin-Ray: > An alternative way to do the parsing might be:

That's a clever approach and it should simplify things overall. It handles arbitrarily large inputs with a fixed amount of space.

So in `parse_menu()` becomes:

```C
char *menu_names_dup = mutt_str_strdup(s);
char *menu_name = strsep(&menu_names_dup, ",");

while (menu_name)
{
   int value = mutt_map_get_value(menu_name, Menus);
   if (value == -1)
   {
      mutt_buffer_printf(err, _("%s: no such menu"), menu_name);
      FREE(&menu_names_dup);
      return NULL; // this is just a placeholder.
   }
   else
   {
       menu[value] = true;
   }

   menu_name = strsep(&menu_names_dup, ","); 
}

FREE(&menu_names_dup);

return NULL;
```

If we drop the early return, we can handle cases where a correct menu follows an incorrect one e.g. `unbind test,pager q`. We would want to remove the `FREE()` as well.

Additionally, your approach simplifies `mutt_parse_unbind` as well. Instead of tracking a boolean for `all_menus`, we can set `menus` can be assigned to an array of `MAX_MENU` size with all values of `1`. It's `for` loop would become this:

```C
  for (int i = 0; i < MENU_MAX; ++i)
  {
    if (menu[i] != 1)
      continue; // Look what you made me type, flatcap. The horror you've wrought upon me

    int map = menu[i]; // This can be inlined or left as variable.
    if (all_keys)
    {
      km_unbind_all(&Keymaps[map],data);
      if ((data & MUTT_UNBIND) == MUTT_UNBIND && i == MENU_GENERIC)
      {
        km_bindkey("<return>", MENU_GENERIC, OP_GENERIC_SELECT_ENTRY);
        km_bindkey("<enter>", MENU_GENERIC, OP_GENERIC_SELECT_ENTRY);
        km_bindkey(":", MENU_GENERIC, OP_ENTER_COMMAND);
        km_bindkey("q", MENU_GENERIC, OP_EXIT);
        km_bindkey("q", MENU_PAGER, OP_EXIT);
        km_bindkey("?", MENU_GENERIC, OP_HELP);
      }
    }
    else
    {
      km_bindkey(key, map, OP_NULL);
    }
  }
```
which has a more uniform control flow.

jindraj: I kept return and free in parse_menu() until we agree what's the best. The same situation as with list of km_bindkeys in mutt_parse_unbind().

Austin-Ray:
I have found a bug. When I execute `unbind *` then press `:`, the prompt is filled with `:` and neomutt locks up. This does not affect `unmacro *`.

jindraj:
> I have found a bug. When I execute `unbind *` then press `:`, the prompt is filled with `:` and neomutt locks up. This does not affect `unmacro *`.

I got this from lldb. I don't know why it happens yet. 
```
* thread #1, queue = 'com.apple.main-thread', stop reason = signal SIGSTOP
  * frame #0: 0x00007fff63901f2b libsystem_c.dylib`wcwidth + 129
    frame #1: 0x000000010014af43 neomutt`mutt_mb_wcwidth(wc=L':') at mbyte.c:180
    frame #2: 0x000000010014b088 neomutt`mutt_mb_width_ceiling(s=L":::::::::::::::::::::::::::::::::::::::::::::::::::::::::::::::::::::::::::::::::::::::::::::::::::::::::::::::::::::::::::::::::::::::::::::::::::::::::::::::::::::::::::::::::::::::::::::::::::::::::::::::::::::::::::::::::::::::::::::::::::::::::::::::::::::::::::::::::::::::::::::::::::::::::::::::::::::::::::::::::::::::::::::::::::::::::::::::::::::::::::::::::::::::::::::::::::::::::::::::::::::::::::::::::::::::::::::::::::::::::::::::::::::::::::::::::::::::::::::::::::::::::::::::::::::::::::::::::::::::::::::::::::::::::::::::::::::::::::::::::::::::::::::::::::::::::::::::::::::::::::::::::::::::::::::::::::::::::::::::::::::::::::::::::::::::::::::::::::::::::::::::::::::::::::::::::::::::::::::::::::::::::::::::::::::::::::::::::::::::::::::::::::::::::::::::::::::::::::::::::::::::::::::::::::::::::::::::::::::::::::::::::::::::::::::::::::::::::::::::::::::::::::::::::::::::::::::::::::::::::::::::::::::::::::::::::::::::::::::::::::::::::::::::::::::::::::::::::::::::::::::::::::::::::::::::", n=1303, w1=15072) at mbyte.c:219
    frame #3: 0x0000000100028737 neomutt`mutt_enter_string_full(buf="", buflen=1024, col=1, flags=64, multiple=false, files=0x0000000000000000, numfiles=0x0000000000000000, state=0x0000000100e1e9a0) at enter.c:230
    frame #4: 0x00000001000219ef neomutt`mutt_get_field_full(field=":", buf="", buflen=1024, complete=64, multiple=false, files=0x0000000000000000, numfiles=0x0000000000000000) at curs_lib.c:274
    frame #5: 0x0000000100011068 neomutt`mutt_enter_command at commands.c:820
    frame #6: 0x000000010004406e neomutt`mutt_index_menu at index.c:3078
    frame #7: 0x000000010005a02a neomutt`main(argc=1, argv=0x00007ffeefbfee20, envp=0x00007ffeefbfee30) at main.c:1217
    frame #8: 0x00007fff6387ded9 libdyld.dylib`start + 1
```

gahr:
Looks like `km_dokey` has a failure path if `Keymaps[menu] == NULL`. Have a look at `retry_generic` and the result at `enter.c` line 739.

flatcap:
hmm... not quite.  after `unbind *` then `:` the <kbd>q</kbd> key doesn't work!

flatcap:
That's good.
Things that shouldn't work, don't, I can't lock it up, and `:bind` is very empty :-)

Please can you squash down your commits and rebase them onto master.

jindraj:
> That's good.
> Things that shouldn't work, don't, I can't lock it up, and `:bind` is very empty :-)
> 
> Please can you squash down your commits and rebase them onto master.
Sure. But there are still some things I would like to solve before:

- Does doc/manual.xml.head sufficiently describe the behaviour? Is the XML syntax correct? I wasn't able to make docs on my setup.
- similar with doxygen
- what to do with unnecessary changes in `parse_keymap()` - where I changed last argument from `bool bind` to `char *cmd`? Keep it? revert?

flatcap:
> what to do with unnecessary changes in `parse_keymap()`
> where I changed last argument from `bool bind` to `char *cmd`? Keep it? revert?

As we're back to two options, I think we should revert that.

> Is the XML syntax correct?

Yes, the XML validates.
I might adjust the wording slightly (later).

> similar with doxygen

Doxygen is fairly simple to test, if you have it installed: `doxygen doxygen/doxygen.conf`

It reports a few errors for the change that you're going to revert.
I'll check it again after.

keymap.c
```diff
@@ -1298,6 +1301,152 @@ enum CommandResult mutt_parse_bind(struct Buffer *buf, struct Buffer *s,
...
+      continue;
+    if (all_keys)
+    {
+      km_unbind_all(&Keymaps[i], data);
+      km_bindkey("<enter>", MENU_GENERIC, OP_GENERIC_SELECT_ENTRY);
+      km_bindkey("<return>", MENU_GENERIC, OP_GENERIC_SELECT_ENTRY);
+      km_bindkey("<enter>", MENU_MAIN, OP_DISPLAY_MESSAGE);
+      km_bindkey("<return>", MENU_MAIN, OP_DISPLAY_MESSAGE);
+      km_bindkey("<backspace>", MENU_EDITOR, OP_EDITOR_BACKSPACE);
+      km_bindkey("\177", MENU_EDITOR, OP_EDITOR_BACKSPACE);
```
Austin-Ray: What key does `\177` correspond to? Converting from octal to decimal to ASCII says it's Delete, but I don't see anything bound to delete after executing `unbind *`

jindraj: Yes, it's DEL character. It's for MENU_EDITOR otherwise it's not possible to delete characters in command prompt. At least in my case on MBP.

flatcap: After `:unbind *`, `:bind editor`, I see:
```
bind editor ^? backspace
bind editor <BackSpace> backspace
```

`^?` or <kbd>Ctrl-?</kbd> corresponds to ASCII `<Del>`.

Issue #1600, opened 2019-03-12
Speed up virtual mailboxes (caching?)


weilbith:
I switched to use primary `virtual-mailboxes`, cause they support custom tags to show at the index window and I can use more virtual folder/mailboxes without the need to create actual IMAP folders.
Unfortunately opening a mailbox (especially on the initial inbox) needs pretty long to load. Same on switching forth and back again. Header caching solved this problem for "normal" Mdirs. To be honest I'm not aware about the features in comparison of all the caching backends. Is it possible to use one to cache the virtual mailboxes as well? Not sure if this works from design.

Austin-Ray:
At the moment, there is no caching for `virtual-mailboxes` within the `notmuch` subsystem. I'll have to investigate further to determine the feasibility of adding it, but thank you for bringing up the idea! It would certainly help with large mailboxes (my archive takes some time to open.)

weilbith:
Hmm okay. So it is a possible future feature. Do we close this issue then? (not sure how your workflow is for such such)

flatcap:
> So it is a possible future feature

Yes.  It's certainly desirable, but I don't know how we can implement it.  Yet.
Ideally, the Notmuch backend should piggy-back on top of the Maildir backend.
That sort of sharing will require a bit of thought.

> Do we close this issue then?

I've added the idea to the [Big List](https://github.com/neomutt/flatcap-brain-dump).
I'll leave the issue open for now, while we're thinking.
If we can't think of a short-term solution it'll get put in the [Hard drawer](https://github.com/neomutt/neomutt/issues?q=label%3Adifficulty%3Ahard+is%3Aclosed).

weilbith:
Okay, so how does the flow looks like?
I image it will call notmuch with the configured query and use its output. 
First point I don't know: do this output is enough already by using the correct constallation of [these parameter](https://www.mankier.com/1/notmuch-search) or do we have to go through each entry in the list and manually load the information?
In case it is the second point, it would be interesting to know where the information are now taken from. If it uses the `--output file` to read each header manually it is slow for sure. But maybe references to the header cache of the Maildir could be used. Not sure if this is in that case faster to filter this cache by the notmuch output.

flatcap:
> how does the flow looks like?

1. user enters query
2. neomutt asks notmuch for matches
   - returns a list of paths
3. neomutt reads the headers of each file to create the index

Step three would be sped up by reading a pre-parsed header from the cache.

[...]

PR #1602, opened 2019-03-14
Make NotMuch hcache-aware

gahr:
#1600

### DIFF
--- a/imap/imap.c
+++ b/imap/imap.c
@@ -59,9 +59,7 @@
 #include "pattern.h"
 #include "progress.h"
 #include "sort.h"
-#ifdef USE_HCACHE
 #include "hcache/hcache.h"
-#endif
 #ifdef ENABLE_NLS
 #include <libintl.h>
 #endif
--- a/notmuch/mutt_notmuch.c
+++ b/notmuch/mutt_notmuch.c
@@ -971,7 +971,7 @@ static void append_message(header_cache_t *h, struct Mailbox *m,
 
   if (from_cache)
   {
-    mutt_hcache_free(h, from_cache);
+    mutt_hcache_free(h, &from_cache);
   }
   else
   {
--- a/notmuch/mutt_notmuch.c
+++ b/notmuch/mutt_notmuch.c
@@ -85,6 +85,11 @@ char *C_NmUnreadTag;  ///< Config: (notmuch) Tag to use for unread messages
 char *C_NmFlaggedTag; ///< Config: (notmuch) Tag to use for flagged messages
 char *C_NmRepliedTag; ///< Config: (notmuch) Tag to use for replied messages
 
+/**
+ * nm_hcache_open - Open a header cache
+ * @param m Mailbox
+ * @retval ptr Header cache handle
+ */
 static header_cache_t *nm_hcache_open(struct Mailbox *m)
 {
 #ifdef USE_HCACHE
@@ -94,6 +99,10 @@ static header_cache_t *nm_hcache_open(struct Mailbox *m)
 #endif
 }
 
+/**
+ * nm_hcache_close - Close the header cache
+ * @param h Header cache handle
+ */
 void nm_hcache_close(header_cache_t *h)
 {
 #ifdef USE_HCACHE
--- a/notmuch/mutt_notmuch.c
+++ b/notmuch/mutt_notmuch.c
@@ -55,6 +55,7 @@
 #include "mutt/mutt.h"
 #include "config/lib.h"
 #include "email/lib.h"
+#include "hcache/hcache.h"
 #include "mutt.h"
 #include "mutt_notmuch.h"
 #include "account.h"
@@ -873,8 +874,8 @@ static struct Email *get_mutt_email(struct Mailbox *m, notmuch_message_t *msg)
  * @param msg   Notmuch message
  * @param dedup De-duplicate results
  */
-static void append_message(struct Mailbox *m, notmuch_query_t *q,
-                           notmuch_message_t *msg, bool dedup)
+static void append_message(header_cache_t *h, struct Mailbox *m,
+                           notmuch_query_t *q, notmuch_message_t *msg, bool dedup)
 {
   char *newpath = NULL;
   struct Email *e = NULL;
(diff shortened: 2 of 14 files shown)


jindraj:
the first commit worked well, a850d4cb57ab56890144ec357cc377c81bc900c6
Starting neomutt for the first time with empty cache dir works. But opening virtual-mailbox for the second time crashes neomutt. Here's backtrace from lldb.
```
Reading messages... 0neomutt(71706,0x10022d5c0) malloc: *** error for object 0x5c8a200f: pointer being freed was not allocated
neomutt(71706,0x10022d5c0) malloc: *** set a breakpoint in malloc_error_break to debug
Process 71706 stopped
* thread #1, queue = 'com.apple.main-thread', stop reason = breakpoint 1.1
    frame #0: 0x00007fff63a420de libsystem_malloc.dylib`malloc_error_break
libsystem_malloc.dylib`malloc_error_break:
->  0x7fff63a420de <+0>: pushq  %rbp
    0x7fff63a420df <+1>: movq   %rsp, %rbp
    0x7fff63a420e2 <+4>: nop
    0x7fff63a420e3 <+5>: nopl   (%rax)
Target 0: (neomutt) stopped.
(lldb) bt
* thread #1, queue = 'com.apple.main-thread', stop reason = breakpoint 1.1
  * frame #0: 0x00007fff63a420de libsystem_malloc.dylib`malloc_error_break
    frame #1: 0x00007fff63a35676 libsystem_malloc.dylib`malloc_vreport + 437
    frame #2: 0x00007fff63a354a3 libsystem_malloc.dylib`malloc_report + 152
    frame #3: 0x00000001000e496b neomutt`mutt_mem_free(ptr=0x0000000102842e00) at memory.c:84 [opt]
    frame #4: 0x000000010008fbcf neomutt`append_message(h=0x00000001019224f0, m=0x0000000102023000, q=0x0000000100e06560, msg=<unavailable>, dedup=<unavailable>) at mutt_notmuch.c:974 [opt]
    frame #5: 0x000000010008c60d neomutt`read_threads_query at mutt_notmuch.c:1055 [opt]
    frame #6: 0x000000010008c5c1 neomutt`read_threads_query(m=<unavailable>, q=0x0000000100e06560, dedup=<unavailable>, limit=<unavailable>) at mutt_notmuch.c:1189 [opt]
    frame #7: 0x000000010008e38b neomutt`nm_mbox_open(m=0x0000000102023000) at mutt_notmuch.c:2207 [opt]
    frame #8: 0x000000010004c6df neomutt`mx_mbox_open(m=0x0000000102023000, flags='\0') at mx.c:331 [opt]
    frame #9: 0x000000010003d700 neomutt`main(argc=<unavailable>, argv=0x00007ffeefbfedf0, envp=<unavailable>) at main.c:1207 [opt]
    frame #10: 0x00007fff6387ded9 libdyld.dylib`start + 1
```

gahr:
@weilbith, @Austin-Ray , can you please give this branch a go? Thanks!

weilbith:
@gahr
Wow, awesome that you implemented a solution so fast (not sure if my issue was the trigger, but...).
Unfortunately I have installed _NeoMutt_ only by distro packages so far. I just checked what I need to have in mind to build from source. So how I need to configure my build, to have this working? Haven't seen an `all` feature kind option.

Austin-Ray:
@gahr I've tested your patch and at first it didn't seem like there was any improvement, but my attention is fickle and I am prone to error so I made a small empirical comparison.

Here are five runs w/ header cache disabled.
```zsh
λ athena neomutt → λ git devel/issue-1600* → time ./neomutt -f "notmuch://?type=threads&query=tag:archive AND NOT tag:spam&limit=40000" -e "push q"
Mailbox is unchanged
./neomutt -f   5.45s user 1.03s system 107% cpu 5.997 total
λ athena neomutt → λ git devel/issue-1600* → time ./neomutt -f "notmuch://?type=threads&query=tag:archive AND NOT tag:spam&limit=40000" -e "push q"
Mailbox is unchanged
./neomutt -f   5.50s user 0.96s system 107% cpu 6.003 total
λ athena neomutt → λ git devel/issue-1600* → time ./neomutt -f "notmuch://?type=threads&query=tag:archive AND NOT tag:spam&limit=40000" -e "push q"
Mailbox is unchanged
./neomutt -f   5.38s user 1.14s system 107% cpu 6.046 total
λ athena neomutt → λ git devel/issue-1600* → time ./neomutt -f "notmuch://?type=threads&query=tag:archive AND NOT tag:spam&limit=40000" -e "push q"
Mailbox is unchanged
./neomutt -f   5.45s user 1.04s system 107% cpu 6.029 total
λ athena neomutt → λ git devel/issue-1600* → time ./neomutt -f "notmuch://?type=threads&query=tag:archive AND NOT tag:spam&limit=40000" -e "push q"
Mailbox is unchanged
./neomutt -f   5.36s user 1.06s system 107% cpu 5.974 total
```

For an average of `6.0098` seconds

Five runs w/ header cache enabled:
```zsh
λ athena neomutt → λ git devel/issue-1600* → time ./neomutt -f "notmuch://?type=threads&query=tag:archive AND NOT tag:spam&limit=40000" -e "push q"
Mailbox is unchanged
./neomutt -f   3.87s user 0.82s system 107% cpu 4.356 total
λ athena neomutt → λ git devel/issue-1600* → time ./neomutt -f "notmuch://?type=threads&query=tag:archive AND NOT tag:spam&limit=40000" -e "push q"
Mailbox is unchanged
./neomutt -f   3.89s user 0.79s system 107% cpu 4.358 total
λ athena neomutt → λ git devel/issue-1600* → time ./neomutt -f "notmuch://?type=threads&query=tag:archive AND NOT tag:spam&limit=40000" -e "push q"
Mailbox is unchanged
./neomutt -f   3.89s user 0.77s system 107% cpu 4.337 total
λ athena neomutt → λ git devel/issue-1600* → time ./neomutt -f "notmuch://?type=threads&query=tag:archive AND NOT tag:spam&limit=40000" -e "push q"
Mailbox is unchanged
./neomutt -f   3.93s user 0.72s system 106% cpu 4.384 total
λ athena neomutt → λ git devel/issue-1600* → time ./neomutt -f "notmuch://?type=threads&query=tag:archive AND NOT tag:spam&limit=40000" -e "push q"
Mailbox is unchanged
./neomutt -f   4.13s user 0.59s system 107% cpu 4.384 total
```
For an average of `4.3638` seconds.

That's `37.7%` faster! Well done!

gahr:
@weilbith [this guide](https://neomutt.org/dev/build) should be fairly up-to-date. `./configure --help` should be pretty self-explanatory. Feel free to ask specific questions :)

@Austin-Ray thanks!

imap/imap.c
```diff
@@ -59,9 +59,7 @@
 #include "pattern.h"
 #include "progress.h"
 #include "sort.h"
-#ifdef USE_HCACHE
```
Austin-Ray: With the removal of a bunch of `#ifdef USE_HCACHE`, won't we encounter issues if someone compiles without `bdb`,  `gdbm`, `lmdb`, `kyotocabinet`, and `qdbm`?

notmuch/mutt_notmuch.c
```diff
@@ -84,6 +85,31 @@ char *C_NmUnreadTag;  ///< Config: (notmuch) Tag to use for unread messages
...
 char *C_NmRepliedTag; ///< Config: (notmuch) Tag to use for replied messages
 
+/**
+ * nm_hcache_open - Open a header cache
+ * @param m Mailbox
+ * @retval ptr Header cache handle
+ */
+static header_cache_t *nm_hcache_open(struct Mailbox *m)
+{
+#ifdef USE_HCACHE
```
Austin-Ray: [1] I'm not a fan of defining an hcache-specific function then having it's behavior differ whether its been compile with or without a cache. It'll lead to some confusion...

notmuch/mutt_notmuch.c
```diff
@@ -1060,18 +1110,23 @@ static bool read_mesgs_query(struct Mailbox *m, notmuch_query_t *q, bool dedup)
   if (!msgs)
     return false;
 
+  header_cache_t *h = nm_hcache_open(m);
```
Austin-Ray: [1] ... particularly in places like here.

Imagine if I compiled Neomutt w/o hcache and decided to debug its behavior. If I see a function call for hcache, it will lead to some confusion. I explicitly compiled without a header cache so there should be no traces of it.

Does this seem like an actual issue or am I being too pedantic?

imap/imap.c
flatcap: It shouldn't be a problem.  `hcache/hcache.h` only contains a definitions and has no dependencies (no database stuff).

We only need the typedef for `header_cache_t` if the header cache is disabled.

notmuch/mutt_notmuch.c
flatcap: > I explicitly compiled without a header cache so there should be no traces of it.

Agreed.

We need some standardisation between the backends on _how_ they use the header cache.

> Does this seem like an actual issue or am I being too pedantic?

It's a mess, but I think we should commit the changes...

and start thinking about how caching should work in a perfect situation.

gahr:
> I explicitly compiled without a header cache so there should be no traces of it.

Well, technically, you don’t compile in any backend implementations, you don’t explicitely exclude header cache. I think the top level should stay, and just return no-op because no backend is chosen

notmuch/mutt_notmuch.c
gahr: I agree - I think we should always compile hcache.c in, and let it return NULL on open and fetch and be a no-op on store. The rest of the code should be transparent. I think I can work on it on a separate issue.

PR #1606, opened 2019-03-21
notmuch: close thread queries' hcache on SigInt

Austin-Ray:
While looking for ways to improve `notmuch` performance, I noticed `read_threads_query()` does not close hcache on SigInt. It's companion function, `read_mesgs_query()` closes hcache on SigInt.

This is my mistake as I failed to notice it in #1602's review.

Sadly, this removes Neomutt's ability to run at >100% efficiency when re-opening a thread query after SigInt

110-percent


gahr:
Good catch, thanks!

PR #1609, opened 2019-03-26
Add new option to display only non-empty mailboxes in sidebar

darnir:
* **What does this PR do?**
Sometimes, there are multiple mailboxes available for an account, but a lot of them are empty. I want to display only the mailboxes that contain emails in them. This PR adds that option via a new configuration: `$sidebar_non_empty_mailbox_only`.

     for that)

### DIFF
--- a/sidebar.c
+++ b/sidebar.c
@@ -391,13 +391,13 @@ static void update_entries_visibility(void)
 
     if (non_empty_only && (i != OpnIndex) && (sbe->mailbox->msg_count == 0))
     {
-        sbe->is_hidden = true;
+      sbe->is_hidden = true;
     }
 
     if (new_only && (i != OpnIndex) && (sbe->mailbox->msg_unread == 0) &&
         (sbe->mailbox->msg_flagged == 0) && !sbe->mailbox->has_new)
     {
-        sbe->is_hidden = true;
+      sbe->is_hidden = true;
     }
   }
 }
--- a/sidebar.c
+++ b/sidebar.c
@@ -373,15 +373,9 @@ static void update_entries_visibility(void)
 
     sbe->is_hidden = false;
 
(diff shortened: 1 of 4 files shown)


sidebar.c
```diff
@@ -355,25 +356,28 @@ static int cb_qsort_sbe(const void *a, const void *b)
 static void update_entries_visibility(void)
 {
   short new_only = C_SidebarNewMailOnly;
+  short non_empty_only = C_SidebarNonEmptyOnly;
```
gahr: I know you're just doing as the previous line, but I really can't think of a reason for copying the boolean flag into a temporary short? Can we get rid of both?

darnir: Agreed! I was about to mention that soon, once this is done. I'll fix it in a commit.

gahr: My comment was more in favour of removing `new_only` and `non_empty_only` altogether. Why have them at all when they are non-modified copies of the `C_*` flags?

flatcap:
This isn't a suggestion to alter the code, but...
It's almost like we need a pattern to control what to display:
`sidebar_display = "~N | ~O | ~F | (~J > 0)"`
(`~J` = message count -- not implemented)

sidebar.c
darnir: Aah. To me, they're there because it makes the code a little easier to parse.

gahr: Can we at least make them const, so one knows they're really just aliases and not going to change because of some logic some lines below?

darnir: Agreed. Force pushed an updated commit

darnir:
Anything blocking this merge?

flatcap:
> Anything blocking this merge?

Not really.
I'm cautious about adding new config options, but this one seems reasonable.
We can have a think about the practicality of pattern matcher.

PR #1614, opened 2019-03-29
 Fix bogus mailbox size accounting

gahr:
* **What does this PR do?**

Mailbox size update upon addition of messages was carried out in various places, using various methods. Update upon deletion was handled in a central place. When the methods used to account addition and subtracted don't match, you end up having bogus accounting:

```
---NeoMutt: +Trash [Msgs:0 17592186044412M]
```

This PR adds a centralized way to account mailbox sizes. While at it, I noticed NotMuch mailboxes weren't accounted for when deleting emails. I can't confirm that this fixes the particular issue, as I'm not a NotMuch user.

### DIFF
--- a/mailbox.c
+++ b/mailbox.c
@@ -533,3 +533,37 @@ void mutt_mailbox_changed(struct Mailbox *m, enum MailboxNotification action)
 
   m->notify(m, action);
 }
+
+/**
+ * email_size - Helper function to make sure we always use the same metric to
+ * compute the size of an email
+ * @param e Email
+ * @retval num Size of the email, in bytes
+ */
+static size_t email_size(const struct Email *e)
+{
+  if (!e || !e->content)
+    return 0;
+  return e->content->length + e->content->offset - e->content->hdr_offset;
+}
+
+/**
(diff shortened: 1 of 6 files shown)


gahr:
@Austin-Ray hi - could you please double check whether NotMuch exhibits the same misbehaviour and whether it's fixed by this PR? Thanks!

Austin-Ray:
Hi @gahr, I can confirm that this fix works with the `notmuch` backend.

mailbox.c
```diff
@@ -533,3 +533,35 @@ void mutt_mailbox_changed(struct Mailbox *m, enum MailboxNotification action)
...
+
+/**
+ * email_size - Helper function to make sure we always use the same metric to
+ * compute the size of an email
+ * @param e Email
+ * @retval num Size of the email, in bytes
+ */
+static size_t email_size(const struct Email *e)
+{
+  return e->content->length + e->content->offset - e->content->hdr_offset;
```
Austin-Ray: I think we should check that `e` and `e->content` are valid pointers just to be safe.

PR #1630, opened 2019-04-02
Allow reopening of Maiboxes (workaround)

flatcap:
These four commits workaround crashes when opening a Mailbox for the second time.

- cdef6e945 fix ev_message()
- 43fe27433 fix save_message()
- d6c0ba6f4 fix pop_fetch_mail()
- 7c6a6160f fix trash_append()

They're not ideal, but they work.

They've led me to consider an 'OpenMailbox' object -- an **instance** of an open Mailbox.  This would allow different callers to open a Mailbox with different options, but still share the same data.

```c
struct OpenMailbox
{
  struct Mailbox *mailbox;
  bool readonly;
  bool append_mode;
};
```


commands.c
```diff
@@ -1068,12 +1068,14 @@ int mutt_save_message(struct Mailbox *m, struct EmailList *el, bool delete,
 #endif
 
   struct Mailbox *m_save = mx_path_resolve(buf);
-  struct Context *ctx_save = mx_mbox_open(m_save, MUTT_APPEND);
+  struct Context *ctx_save = mx_mbox_open(m_save, 0);
```
Austin-Ray: [0] Since we've gone through the trouble of `typedef`ing `OpenMailboxFlags`, shouldn't this be `MUTT_OPEN_NO_FLAGS`?

editmsg.c
```diff
@@ -173,7 +174,7 @@ static int ev_message(enum EvMessage action, struct Mailbox *m, struct Email *e)
     goto bail;
   }
 
-  struct Context *ctx_app = mx_mbox_open(m, MUTT_APPEND);
+  struct Context *ctx_app = mx_mbox_open(m, 0);
```
Austin-Ray: Same as [0]

mx.c
```diff
@@ -490,22 +490,27 @@ static int trash_append(struct Mailbox *m)
 #endif
 
   struct Mailbox *m_trash = mx_path_resolve(C_Trash);
-  struct Context *ctx_trash = mx_mbox_open(m_trash, MUTT_APPEND);
+  struct Context *ctx_trash = mx_mbox_open(m_trash, 0);
```
Austin-Ray: Same as [0]

pop/pop.c
```diff
@@ -648,13 +648,16 @@ void pop_fetch_mail(void)
   }
 
   struct Mailbox *m_spool = mx_path_resolve(C_Spoolfile);
-  struct Context *ctx = mx_mbox_open(m_spool, MUTT_APPEND);
+  struct Context *ctx = mx_mbox_open(m_spool, 0);
```
Austin-Ray: Same as [0]

commands.c
flatcap: Yeah :-)
These "new" commits have actually been sitting around for a while
(they originally predated the `OpenMailboxFlags` changes).

flatcap:
> How exactly do these changes prevent the crash from occurring?

It's all about how `mx_mbox_open()` handles flags.
These four examples are where we have a Mailbox open in 'normal' mode and try to open it again in 'append' mode.

'Append' mode only makes sense for 'mbox' and 'compress', but the flag affects `mx_mbox_open()`.  It causes the function to create a duplicate Mailbox, which is then used and discarded.  The original Mailbox is now out of sync.

The workaround, simply hides the flag for `mx_mbox_open()`, but temporarily sets the internal bool `append`.  This means that we work with **one** Mailbox.

This is what lead me to wonder about a Mailbox wrapper, above.

Austin-Ray:
> > How exactly do these changes prevent the crash from occurring?
> 
> It's all about how `mx_mbox_open()` handles flags.
> These four examples are where we have a Mailbox open in 'normal' mode and try to open it again in 'append' mode.
> 
> 'Append' mode only makes sense for 'mbox' and 'compress', but the flag affects `mx_mbox_open()`. It causes the function to create a duplicate Mailbox, which is then used and discarded. The original Mailbox is now out of sync.
> 
> The workaround, simply hides the flag for `mx_mbox_open()`, but temporarily sets the internal bool `append`. This means that we work with **one** Mailbox.
> 
> This is what lead me to wonder about a Mailbox wrapper, above.

This makes sense to me. Can you put that into the git commit so the rationale is in-band with the changes?

sileht:
I'm testing your branch, personally I was getting ton of crashs by setting: `folder-hook . 'set record = "^"'`

But now, at first glance, it works well. If I see something in next days, you will be informed :)

flatcap:
@Austin-Ray done
@sileht thanks

flatcap:
@gahr 
> `m->append` and `MUTT_APPEND` are about a different thing?

They're the same -- `MUTT_APPEND` sets `Mailbox->append`.
I'm just avoiding the side-effects of `mx_mbox_open()` when opened for the second time.

PR #1634, opened 2019-04-04
curses: clear screen on exit

flatcap:
On IRC, compi123 reported problems, on FreeBSD 11.2, with text being left behind on the 'alternate' screen after NeoMutt closes.

This PR forcibly clears the screen on exit.

### DIFF
--- a/main.c
+++ b/main.c
@@ -200,6 +200,8 @@ static void reset_tilde(struct ConfigSet *cs)
  */
 void mutt_exit(int code)
 {
+  clear();
+  refresh();
   mutt_endwin();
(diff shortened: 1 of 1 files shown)


flatcap:
To test whether anything has been left behind by NeoMutt, use the following escape sequences:
```sh
echo -en "\e[?1049h" # Switch to 'alternate' screen
echo -en "\e[?1049l" # Switch to 'normal' screen
```

flatcap:
**Note**: This was reverted a year later in 0830efa0e.
See #2146

PR #1642, opened 2019-04-08
Tidying: non-functional code improvements

flatcap:
- c2b2e100b test: sync mutt_extract_token()
- 253e5dc7e boolify mutt_sig_allow_interrupt()
- f4abf3eb4 unify variable names
- 31ba61d73 use MUTT_TOKEN_NO_FLAGS
- 7eccb18dd reduce scope of variables
- 760358834 rename 'hdr' variables
- 85c0cc368 Replace 0 flags with their NO_FLAGS equivalent

### DIFF
--- a/browser.c
+++ b/browser.c
@@ -580,9 +580,11 @@ static const char *folder_format_str(char *buf, size_t buflen, size_t col, int c
   }
 
   if (optional)
-    mutt_expando_format(buf, buflen, col, cols, if_str, folder_format_str, data, 0);
+    mutt_expando_format(buf, buflen, col, cols, if_str, folder_format_str, data,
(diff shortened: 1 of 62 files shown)


flatcap:
> but I see some function calls missing `NO_FLAGS`

Thanks.  I've fixed those two.  I'll try to make another pass for them.
The trouble is, it's hard to automate, both backtracking to find all the callers and making the `0` to `NO_FLAGS` edit.

flatcap:
For reference:
- 23 Flag Types
- 221 Symbols
- 122 Functions using a Flag type as parameter
- 706 Callers of those functions

I think I've got them all now.

Issue #1647, opened 2019-04-12
neomutt no longer marks replied messages as such


l29ah:
1e0091a8b923db07e4af59f0c98c762786233f32 is the first bad commit

flatcap:
Thanks for the bisect, @l29ah.

2eddd7c fixes the reply bug, but... there seems to be another problem.

After the flag is set, it's not being committed.
If you immediately quit NeoMutt, the flag isn't saved (you have to `<sync-mailbox>`).

Austin-Ray:
On 04/12, Richard Russon wrote:
> After the flag is set, it's not being committed.  If you immediately
> quit NeoMutt, the flag isn't saved (you have to `<sync-mailbox>`).

Reply flag is committed w/ notmuch and maildir backends.

flatcap:
> Reply flag is committed w/ notmuch and maildir backends.

Maybe in combination, but I don't see it with a plain maildir mailbox.
I see `m->changed == false` and `m->append == true` so `mx_mbox_close()` isn't doing anything.

Austin-Ray:
On 04/14, Richard Russon wrote:
> > Reply flag is committed w/ notmuch and maildir backends.
> 
> Maybe in combination, but I don't see it with a plain maildir mailbox.

Standalone maildir commits the reply flag on my machine.

flatcap:
> Standalone maildir commits the reply flag on my machine.

Ah, I know why our tests differ.
I have `set record = '^' # Save outgoing to current mailbox` causing `mutt_write_fcc()` is setting `m->append`.

PR #1648, opened 2019-04-14
fix setting the 'replied' flag

flatcap:
When replying to an email, mark the original as 'replied'.

Fixes: #1647

**Notes**: `cur` used to refer to the `Email` to change.  It it was `NULL`, then the function would look for tagged `Email`s.  Now, the function is passed an `EmailList` to work on.

### DIFF
--- a/maildir/maildir.c
+++ b/maildir/maildir.c
@@ -507,7 +507,10 @@ int maildir_mbox_check(struct Mailbox *m, int *index_hint)
   /* Incorporate new messages */
   num_new = maildir_move_to_mailbox(m, &md);
   if (num_new > 0)
+  {
     mutt_mailbox_changed(m, MBN_INVALID);
+    m->changed = true;
+  }
 
   mutt_buffer_pool_release(&buf);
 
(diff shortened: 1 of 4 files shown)


maildir/maildir.c
```diff
@@ -507,7 +507,10 @@ int maildir_mbox_check(struct Mailbox *m, int *index_hint)
   /* Incorporate new messages */
   num_new = maildir_move_to_mailbox(m, &md);
   if (num_new > 0)
+  {
     mutt_mailbox_changed(m, MBN_INVALID);
+    m->changed = true;
```
Austin-Ray: I feel like it makes more to put `m->changed = true;` inside `mutt_mailbox_changed(...)`. Otherwise, `mutt_mailbox_changed(...)` seems like an inaccurate name.

flatcap: The name `mutt_mailbox_changed()` isn't a great one -- `mutt_mailbox_notify()` might be better.
I didn't want to move `m->changed = true` because I didn't want to change the behaviour of the other callers -- which may, or may not, be correct :-)

Ultimately, `mutt_mailbox_changed()` is a temporary fix and will die soon
("soon" == when I stop making small changes and finish the big ones).

PR #1649, opened 2019-04-14
Tidy Buffer usage

flatcap:
Simplify functions by allocating Buffers on the heap, rather than using ones on the stack.

### DIFF
--- a/hook.c
+++ b/hook.c
@@ -437,16 +437,13 @@ char *mutt_find_hook(HookFlags type, const char *pat)
  */
 void mutt_message_hook(struct Mailbox *m, struct Email *e, HookFlags type)
 {
-  struct Buffer err, token;
   struct Hook *hook = NULL;
   struct PatternCache cache = { 0 };
+  struct Buffer *err = mutt_buffer_pool_get();
+  struct Buffer *token = mutt_buffer_pool_get();
 
   current_hook_type = type;
 
-  mutt_buffer_init(&err);
-  err.dsize = 256;
-  err.data = mutt_mem_malloc(err.dsize);
-  mutt_buffer_init(&token);
   TAILQ_FOREACH(hook, &Hooks, entries)
   {
     if (!hook->command)
@@ -457,12 +454,12 @@ void mutt_message_hook(struct Mailbox *m, struct Email *e, HookFlags type)
       if ((mutt_pattern_exec(SLIST_FIRST(hook->pattern), 0, m, e, &cache) > 0) ^
           hook->regex.not)
       {
-        if (mutt_parse_rc_line(hook->command, &token, &err) == MUTT_CMD_ERROR)
+        if (mutt_parse_rc_line(hook->command, token, err) == MUTT_CMD_ERROR)
         {
-          FREE(&token.data);
-          mutt_error("%s", err.data);
+          mutt_buffer_pool_release(&token);
+          mutt_error("%s", mutt_b2s(err));
           current_hook_type = 0;
-          FREE(&err.data);
+          mutt_buffer_pool_release(&err);
 
           return;
         }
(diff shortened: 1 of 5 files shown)


hook.c
```diff
@@ -460,24 +454,20 @@ void mutt_message_hook(struct Mailbox *m, struct Email *e, HookFlags type)
...
+        if (mutt_parse_rc_line(hook->command, token, err) == MUTT_CMD_ERROR)
         {
-          FREE(&token.data);
-          mutt_error("%s", err.data);
-          current_hook_type = 0;
-          FREE(&err.data);
-
-          return;
+          mutt_error("%s", mutt_b2s(err));
+          break;
```
Austin-Ray: [0] Correct me if I'm wrong: Changing the `return` to the `break`changes the behavior of the method by continuing on error instead of bailing.

If so, we should extract the `return` to `break` change to its own small commit. It's out of scope of a tidying commit, which shouldn't change behavior.

hook.c
```diff
@@ -644,22 +629,19 @@ void mutt_account_hook(const char *url)
...
+      if (mutt_parse_rc_line(hook->command, token, err) == MUTT_CMD_ERROR)
       {
-        FREE(&token.data);
-        mutt_error("%s", err.data);
-        FREE(&err.data);
-
+        mutt_error("%s", mutt_b2s(err));
         inhook = false;
-        return;
+        break;
```
Austin-Ray: Same as [0]

init.c
```diff
@@ -281,28 +281,24 @@ static int complete_all_nm_tags(const char *pt)
...
+    if (mutt_parse_rc_line(np->data, token, err) == MUTT_CMD_ERROR)
     {
-      mutt_error(_("Error in command line: %s"), err.data);
-      FREE(&token.data);
-      FREE(&err.data);
-
-      return -1;
+      mutt_error(_("Error in command line: %s"), mutt_b2s(err));
+      rc = 1;
+      break;
```
Austin-Ray: [1] Everything from [0], but this changes the error return code from this function as well.

init.c
```diff
@@ -2631,21 +2622,18 @@ int mutt_dump_variables(bool hide_sensitive)
...
+    if (mutt_parse_rc_line(cmd, token, err) == MUTT_CMD_ERROR)
     {
-      mutt_message("%s", err.data);
-      FREE(&token.data);
-      FREE(&err.data);
-
-      return 1; // TEST17: can't test
+      mutt_message("%s", mutt_b2s(err));
+      rc = 1; // TEST17: can't test
+      break;
```
Austin-Ray: Same as [0]

gahr:
The downside is that now you have two calls to malloc for each buffer. Would you be in favour of a mixed approach to at least allocate on the stack the fixed portion?

```c
struct Buffer buf;
mutt_buffer_init_size(&buf, 256);
```

flatcap:
@gahr 

> The downside is that now you have two calls to malloc for each buffer

Yeah, I know :-(

Currently, the number of `Buffer` users is ballooning as Upstream converts everything to use them.

> Would you be in favour of a mixed approach to at least allocate on the stack the fixed portion?

Hmm... it's a mixed blessing.
Fewer `malloc`s, but extra functions -- a new `init()` and a new `free()`.

To reduce the impact, Upstream have added the `BufferPool` -- a telescoping set of `Buffer`s of 1024 bytes.  Initially there are 5 in the pool.

Using the `BufferPool` we can reduce our number of `malloc`s.

hook.c
flatcap: Yes and no.
It's cheeky of me to mix this in, but the function's behaviour is unchanged
(with the possible exception below where I've messed up the return code).

I'll refactor this commit into the `Buffer` changes and the control-flow changes.

gahr:
+1 to use the pool

flatcap:
@Austin-Ray 
Naughty code removed.

flatcap:
Changed to use the Buffer pool

PR #1650, opened 2019-04-14
Refactor conditionals with increments

flatcap:
Consider the conditional:
```c
if (is_valid && (++num > 10))
{
  action();
}
```

Here, `num` is only incremented if the `is_valid` is true.

These commits refactor the code to be equivalent to:

```c
if (is_valid)
{
  num++;
  if (num > 10)
  {
    action();
  }
}
```

This makes it clear when the number is being incremented.

### DIFF
--- a/mutt/file.c
+++ b/mutt/file.c
@@ -1157,12 +1157,16 @@ int mutt_file_lock(int fd, bool excl, bool timeout)
       prev_sb = sb;
 
     /* only unlock file if it is unchanged */
-    if ((prev_sb.st_size == sb.st_size) && (++count >= (timeout ? MAX_LOCK_ATTEMPTS : 0)))
+    if (prev_sb.st_size == sb.st_size)
     {
-      if (timeout)
-        mutt_error(_("Timeout exceeded while attempting flock lock"));
-      rc = -1;
-      break;
+      count++;
+      if (count >= (timeout ? MAX_LOCK_ATTEMPTS : 0))
+      {
+        if (timeout)
+          mutt_error(_("Timeout exceeded while attempting flock lock"));
+        rc = -1;
+        break;
+      }
     }
 
     prev_sb = sb;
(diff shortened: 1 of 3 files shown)


Austin-Ray:
On 04/14, Richard Russon wrote:
> This makes it clear when the number is being incremented.

While this clarifies _when_ the number is incremented, it doesn't
clarify _why_ the numbers are incremented. The conditionals themselves
are more confusing than the increment logic.

gahr:
I think this is going one step too far. The original conditionals look certainly clearer to me than the rearranged ones, as they introduce only one conditional block instead of two and they make it clear that the one conditional block is only entered when two conditions are tied by some logic (AND, OR).

flatcap:
@Austin-Ray 

> it doesn't clarify _why_ the numbers are incremented

True.  I haven't concerned myself with that.
If you'd like to figure that out and comment the code, please be my guest.

flatcap:
@gahr

> The original conditionals look certainly clearer to me

The complexity of the conditional isn't what's worrying me.
It's the side-effect of the pre-incremented variable `++num` (above).
(The four commits "fix" the four occurrences in the code)

**Pre**-increments happen before the clause in question, but does that mean:
1. `num` is incremented before the `if`
2. `num` is incremented **only** when the first `if` clause is true

(2. is the correct solution)

Nearly thirty years of using C and _I_ wasn't certain -- I had to write a test program to check.
Reading the C specification would also have solved the problem -- but would probably lead to many hours of dry reading.

The original code is more compact, but its behaviour isn't clear to the untrained eye.

Austin-Ray:
On 04/15, Richard Russon wrote:
> @Austin-Ray 
> 
> > it doesn't clarify _why_ the numbers are incremented
> 
> True.  I haven't concerned myself with that.
> If you'd like to figure that out and comment the code, please be my guest.

I'll look into that.

gahr:
> Pre-increments happen before the clause in question

This is wrong. Pre-increments - like any other expression - happen *within* the clause that contain them; they just happen *before* the value is fetched. This has very little to do with how clauses linked by logical conditions behave.

My position is that the confusion (and your famous dislike for pre-increments ;)) don't justify the code change - which becomes in my opinion harder to reason about. Can we kill this PR?

flatcap:
OK.  Killed.

PR #1651, opened 2019-04-14
promote two functions to libmutt

flatcap:
Upstream introduced two new functions.  Move them to libmutt.
- `mutt_buffer_concat_path()`
- `mutt_path_getcwd()`

### DIFF
--- a/mutt/path.c
+++ b/mutt/path.c
@@ -36,6 +36,7 @@
 #include <stdlib.h>
 #include <string.h>
 #include <unistd.h>
+#include "buffer.h"
 #include "logging.h"
 #include "memory.h"
 #include "message.h"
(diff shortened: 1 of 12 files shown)


mutt/path.c
```diff
@@ -567,3 +568,22 @@ char *mutt_path_escape(const char *src)
...
+
+/**
+ * mutt_path_getcwd - Get the current working directory
+ * @param cwd Buffer for the result
+ */
+void mutt_path_getcwd(struct Buffer *cwd)
+{
+  mutt_buffer_increase_size(cwd, PATH_MAX);
+  char *retval = getcwd(cwd->data, cwd->dsize);
+  while (!retval && (errno == ERANGE))
```
gahr: I know it's not your code, but can this ever happen?

flatcap: Possibly...

The `ERANGE` condition is where our buffer isn't big enough.

Upstream's code has `_POSIX_PATH_MAX`, so failure is a distinct possibility.
As we're using `PATH_MAX` it's less likely, but...

[POSIX:2001](http://pubs.opengroup.org/onlinepubs/009695399/functions/getcwd.html) isn't clear on the matter
[POSIX:2008](http://pubs.opengroup.org/onlinepubs/9699919799/functions/getcwd.html) notes:
> If the pathname obtained from getcwd() is longer than {PATH_MAX} bytes...

gahr: I think all over the code we make the assumption that PATH_MAX is enough to store a path. I'd make the same assumption here... but yeah, the check doesn't hurt.

PR #1655, opened 2019-04-16
Start testing in earnest

flatcap:
There are currently [46 string functions](https://github.com/neomutt/test-library#string-mutt) in a library of [373 functions](https://github.com/neomutt/test-library#library-373-functions).

This PR adds a test stub for each of the string functions, see `test/string/*.c`
So far, each stub contains a few degenerate tests that pass `NULL` for all the pointer parameters, plus any existing tests from `test/string.c`

To run the tests:
```sh
./configure
make test
```

or, to run a specific test:

```sh
test/string-test -v test_mutt_str_strcasecmp
```

Additionally, there's coverage testing, using `lcov`:
```sh
./configure --lcov
make coverage
# Open 'lcov/index.html' in your browser
```

Running `neomutt`, or the tests will accumulate coverage data in the files `*.gc??`


Makefile.autosetup
```diff
@@ -424,6 +424,17 @@ distclean: clean
 	$(RM) $(DEPFILES) .clang_complete autosetup/jimsh0 config.h config.log \
 		conststrings.c contrib/Makefile doc/Makefile doc/neomutt.1 \
 		html Makefile po/Makefile test/Makefile
+	$(RM) *.gc?? */*.gc?? test/*/*.gc??
+	$(RM) lcov.info lcov
+
+# coverage testing
+force:
```
gahr: ?

flatcap: The `lcov` target creates an `lcov` directory.
The `force` target means the `lcov` target will always be rebuilt when asked for.

I suppose I could change the target to `coverage` and drop `force:`

gahr: Oh, I think .PHONY would be clearer in this case, but yeah, let's call that `coverage`

flatcap:
Now: `make coverage`

PR #1658, opened 2019-04-17
Support group-chat-reply in the index

### LINKED ISSUE #1654: group-chat-reply does not work
I have just moved to  neomutt and wonder why didnt find this wonderful mail client earlier. 
However, It makes me a little uncomfortable when "Reply All" a mail. The guys in CC list is not included in the composing mail. 

Scanning through the reference, there is a function "group-chat-reply" with description: "reply to all recipients preserving To/Cc" should be exactly what I need. However, trying to use this function, either using  :exec or binding it to a key, both does not work. 
I scanned through the latest source code, and it looks like the function quite new, and is not implemented fully. 

Am I correct ?

gahr:
Fixes #1654

### DIFF
--- a/index.c
+++ b/index.c
@@ -3142,11 +3142,11 @@ int mutt_index_menu(void)
       case OP_GROUP_REPLY:
       case OP_GROUP_CHAT_REPLY:
       {
-        SendFlags replyFlags = SEND_REPLY;
+        SendFlags replyflags = SEND_REPLY;
         if (op == OP_GROUP_REPLY)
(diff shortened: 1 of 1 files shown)


index.c
```diff
@@ -3140,7 +3140,13 @@ int mutt_index_menu(void)
         break;
 
       case OP_GROUP_REPLY:
+      case OP_GROUP_CHAT_REPLY:
       {
+        SendFlags replyFlags = SEND_REPLY;
```
flatcap: Would you mind changing `replyFlags` to say, `reply_flags`.

PR #1662, opened 2019-04-20
More testing/coverage

flatcap:
The main changes are:
- Add basic tests for the path code
- Get coverage for the config code up to 100%
- Add more (pointer) checks to existing library code


config/set.c
```diff
@@ -261,7 +261,7 @@ bool cs_register_type(struct ConfigSet *cs, unsigned int type, const struct Conf
 bool cs_register_variables(const struct ConfigSet *cs, struct ConfigDef vars[], int flags)
 {
   if (!cs || !vars)
-    return CSR_ERR_CODE; /* LCOV_EXCL_LINE */
+    return false;
```
Austin-Ray: We should document this return value.

test/config/dump.c
```diff
@@ -0,0 +1,368 @@
...
+  { "Lemon",      DT_REGEX,                0, &VarLemon,      0,                           NULL },
+  { "Mango",      DT_SORT,                 0, &VarMango,      1,                           NULL },
+  { "Nectarine",  DT_STRING,     F_SENSITIVE, &VarNectarine,  IP "nectarine",              NULL },
+  { NULL },
+};
+// clang-format on
+
+void mutt_pretty_mailbox(char *buf, size_t buflen)
+{
+}
```
Austin-Ray: Let's discard this since it's empty.

flatcap: That's used by `dump_config()`
The fact that it doesn't do anything doesn't matter for the purposes of testing.

config/set.c
flatcap: `cs` and `vars` are mandatory and if either is missing the result is a failure.
The retval is type `bool` and I declare:
```
// @retval bool True, if all variables were registered successfully
```

I can change it, if you really think it adds something:
```
// @retval true  All variables were registered successfully
// @retval false Not all variables were registered successfully
```

flatcap:
In case you haven't tried the coverage testing, it looks like this:
- https://flatcap.org/mutt/lcov/index.html

PR #1664, opened 2019-04-21
imap: fix oauth

sileht:
Currently when oauth success, we got this and neomutt UI got stuck:

[2019-04-21 15:35:38]<5> mutt_socket_readln_d() 5< a0001 OK xxxxxxxxxxxxx@xxxxxxxxx authenticated (Success)
[2019-04-21 15:35:38]<2> mutt_socket_write_d() 5> [2019-04-21 15:35:38]<2> imap_login() Communication encrypted at 256 bits
[2019-04-21 15:35:38]<3> imap_mdata_new() hcache uidvalidity 1, uidnext 228, modseq 0
[2019-04-21 15:35:38]<M> mx_mbox_open() Reading imaps://imap.gmail.com/INBOX...
[2019-04-21 15:35:38]<M> imap_mbox_open() Selecting INBOX...
[2019-04-21 15:35:38]<2> mutt_socket_write_d() 5> a0002 SELECT "INBOX"
[2019-04-21 15:35:38]<5> mutt_socket_readln_d() 5< * BAD [CLIENTBUG] Invalid tag

In fact neomutt send ^A to the server, but this should occurs only if
the authentication fails.

This changes fixes that.

### DIFF
--- a/imap/auth_oauth.c
+++ b/imap/auth_oauth.c
@@ -79,7 +79,7 @@ enum ImapAuthRes imap_auth_oauth(struct ImapAccountData *adata, const char *meth
   FREE(&oauthbearer);
   FREE(&ibuf);
 
-  if (rc == IMAP_EXEC_SUCCESS)
+  if (rc != IMAP_EXEC_SUCCESS)
   {
     /* The error response was in SASL continuation, so continue the SASL
      * to cause a failure and exit SASL input.  See RFC 7628 3.2.3 */
@@ -87,7 +87,7 @@ enum ImapAuthRes imap_auth_oauth(struct ImapAccountData *adata, const char *meth
     rc = imap_exec(adata, ibuf, IMAP_CMD_NO_FLAGS);
   }
 
-  if (rc != IMAP_EXEC_SUCCESS)
+  if (rc == IMAP_EXEC_SUCCESS)
   {
     mutt_clear_error();
     return IMAP_AUTH_SUCCESS;


imap/auth_oauth.c
```diff
@@ -79,15 +79,15 @@ enum ImapAuthRes imap_auth_oauth(struct ImapAccountData *adata, const char *meth
...
+  if (rc != IMAP_EXEC_SUCCESS)
   {
     /* The error response was in SASL continuation, so continue the SASL
      * to cause a failure and exit SASL input.  See RFC 7628 3.2.3 */
     mutt_socket_send(adata->conn, "\001");
     rc = imap_exec(adata, ibuf, IMAP_CMD_NO_FLAGS);
   }
 
-  if (rc != IMAP_EXEC_SUCCESS)
+  if (rc == IMAP_EXEC_SUCCESS)
```
flatcap: is it worth making this an `else` ?

sileht: rc can be the one from L77 or L87, so we can't use else.

sileht:
For the story I have introduced this bug here: 04892a48ea76b37f6aa9b369b5bc4f5cc362878c

imap/auth_oauth.c
flatcap: Sorry, don't mind me -- I'm a liability :-)

sileht:
I think some other have this kind of issue, I will fix them.

flatcap:
> I have introduced this bug here: 04892a4

When I saw the bug, I immediately ran `git blame` -- it looked like a classic **flatcap**-refactoring-error :-)

PR #1672, opened 2019-04-23
Improve hierarchy information accuracy in IMAP browser

flatcap:
@nnathan 
This is your patch (via upstream).  Please can you check it still works.

---

Currently the IMAP browser relies on LIST and LSUB (for listing subscribed
folders) which may not provide the required hierarchy information.

RFC3348 section 3 goes as far as stating that a client mustn't rely on LSUB
for hierarchy information.

This patch implements the LIST command extensions specified in RFC5258
requiring that a server must respond with hierarchy information for
listed folders (whether or not filtering on subscribed folders).

### DIFF
--- a/imap/browse.c
+++ b/imap/browse.c
@@ -190,7 +190,8 @@ int imap_browse(const char *path, struct BrowserState *state)
   char buf[PATH_MAX];
   char mbox[PATH_MAX];
   char munged_mbox[PATH_MAX];
-  char list_cmd[5];
+  char list_cmd[18];
+  int len;
   int n;
   char ctmp;
   bool showparents = false;
@@ -204,7 +205,6 @@ int imap_browse(const char *path, struct BrowserState *state)
 
   save_lsub = C_ImapCheckSubscribed;
   C_ImapCheckSubscribed = false;
-  mutt_str_strfcpy(list_cmd, C_ImapListSubscribed ? "LSUB" : "LIST", sizeof(list_cmd));
 
   // Pick first mailbox connected to the same server
   struct MailboxNode *np = NULL;
@@ -221,6 +221,21 @@ int imap_browse(const char *path, struct BrowserState *state)
   }
   if (!adata)
     goto fail;
+
+  if (C_ImapListSubscribed)
+  {
+    const char *lsub_cmd = "LSUB";
+
+    /* RFC3348 section 3 states LSUB is unreliable for hierarchy information.
+     * The newer LIST extensions are designed for this.  */
+    if (adata->capabilities & IMAP_CAP_LIST_EXTENDED)
+      lsub_cmd = "LIST (SUBSCRIBED)";
+    mutt_str_strfcpy(list_cmd, lsub_cmd, sizeof(list_cmd));
+  }
+  else
+  {
+    mutt_str_strfcpy(list_cmd, "LIST", sizeof(list_cmd));
+  }
 
   mutt_message(_("Getting folder list..."));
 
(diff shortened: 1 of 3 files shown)


nnathan:
Just tested it, it's working identical to the mutt-dev version, which (I believe) is correct.

Thanks for expediting this one.

imap/browse.c
```diff
@@ -220,6 +220,21 @@ int imap_browse(const char *path, struct BrowserState *state)
...
+    /* RFC3348 section 3 states LSUB is unreliable for hierarchy information.
+     * The newer LIST extensions are designed for this.  */
+    if (adata->capabilities & IMAP_CAP_LIST_EXTENDED)
+      lsub_cmd = "LIST (SUBSCRIBED)";
+    mutt_str_strfcpy (list_cmd, lsub_cmd, sizeof (list_cmd));
+  }
+  else
+  {
+    mutt_str_strfcpy (list_cmd, "LIST", sizeof (list_cmd));
+  }
```
gahr: Cool! Would the code be more readable if we either use `mutt_str_strfcpy` directly in all branches or set the cmd in all branches and call `mutt_str_strfcpy` only once?

version 1:

```c
if (C_ImapListSubscribed)
{
  /* RFC3348 section 3 states LSUB is unreliable for hierarchy information.
   * The newer LIST extensions are designed for this.  */
  if (adata->capabilities & IMAP_CAP_LIST_EXTENDED)
    mutt_str_strfcpy (list_cmd, "LIST (SUBSCRIBED)", sizeof (list_cmd));
  else
    mutt_str_strfcpy (list_cmd, "LSUB", sizeof (list_cmd));
}
else
{
  mutt_str_strfcpy (list_cmd, "LIST", sizeof (list_cmd));
}
```

version 2:
```c
const char *cmd = NULL;
if (C_ImapListSubscribed)
{
  /* RFC3348 section 3 states LSUB is unreliable for hierarchy information.
   * The newer LIST extensions are designed for this.  */
  if (adata->capabilities & IMAP_CAP_LIST_EXTENDED)
    cmd = "LIST (SUBSCRIBED)";
  else
    cmd = "LSUB";
}
else
{
  cmd = "LIST";
}
mutt_str_strfcpy (list_cmd, cmd, sizeof (list_cmd));
```

version 3
```c
mutt_str_strfcpy (
  list_cmd, 
  !C_ImapListSubscribed ? "LIST" : (adata->capabilities & IMAP_CAP_LIST_EXTENDED) ? "LIST (SUBSCRIBED)" : "LSUB"
  sizeof (list_cmd));
```

Hint: one of the versions is a joke.

flatcap: Hmm... 2) _does_ look nicer.

flatcap:
Someone upstream has reported that `<toggle-subscribed>` is broken by this commit, so I'll just wait a bit before merging.

gahr:
@nnathan, also, where does IMAP say that the hierarchy separator reported by LSUB might be incorrect? I understand that the flags might be incorrect, but the hierarchy separator?

nnathan:
> @nnathan, also, where does IMAP say that the hierarchy separator reported by LSUB might be incorrect? I understand that the flags might be incorrect, but the hierarchy separator?

Hm you mean the usual separator such as dot or slash? I don't recall it ever coming up. When the RFCs talk about the 'hierarchy information' they're talking about the \HasChildren and \HasNoChildren flags that are returned in LIST/LSUB responses.

gahr:
Oh - understood. Yeah, I assumed "hierarchy information" was referring to "hierarchy separator", which is the name for the usual dot or slash.

nnathan:
Looks like [kevin/imap-sub](https://gitlab.com/muttmua/mutt/commit/39e72bf750761316e41c7227ec6644dbcce6c74b) from the mutt tree fixes the upstream issue. Also looks like it implements @gahr "version 2." So probably worth waiting to import that patch once it's committed to master.

flatcap:
> import that patch once it's committed to master

Thanks.  Will do.

flatcap:
- Upstream fix added
- Rebased onto `master`

nnathan:
Just tested the upstream fix. The toggle-subscription view  is working identically to mutt-dev.

PR #1673, opened 2019-04-24
browser: show "Subscribed" in IMAP browser

nnathan:
Mutt (since 1999) has displayed "Subscribed"/"Directory" in the IMAP browser when toggling filtering mode.

Not sure why but this is absent in neomutt. This PR restores this functionality.

Question: should I also import the language translations?

### DIFF
--- a/po/bg.po
+++ b/po/bg.po
@@ -177,6 +177,10 @@ msgstr ""
 msgid "Mailboxes [%d]"
 msgstr "Пощенски кутии [%d]"
 
+#, c-format
+msgid "Subscribed [%s], File mask: %s"
+msgstr "Абониран [%s], Файлова маска: %s"
+
 #: browser.c:994
 #, c-format
 msgid "Directory [%s], File mask: %s"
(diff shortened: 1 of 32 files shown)


flatcap:
Hmm... that's odd.  I wonder where it went.
Ah, the 'Sensible Browser' patch e0cb84ca1.

> Question: should I also import the language translations?

Yes please.

If you cut'n'paste the lines we need, but...
please **don't** run `make update-po` -- I'll do that when I next merge the translate branch.

flatcap:
Three of the translations look garbled: `bg.po`, `el.po`, `ko.po`
The files you've copied from aren't in utf-8 format, but various other character sets.

The charset is listed in the header of the `po` file.
You can convert them using iconv:
`iconv -f cp1251 -t utf-8 bg.po > bg2.po`

po/bg.po
```diff
@@ -177,6 +177,10 @@ msgstr ""
 msgid "Mailboxes [%d]"
 msgstr "Пощенски кутии [%d]"
 
+#, c-format
+msgid "Subscribed [%s], File mask: %s"
+msgstr "Àáîíèðàí [%s], Ôàéëîâà ìàñêà: %s"
```
flatcap: Just this one left; all the others look good.

PR #1682, opened 2019-04-29
Testing: Basic Library Coverage

flatcap:
Add test templates and degenerate tests for:

- libmutt (base64, buffer, charset, date, envlist, file, hash, history, list, logging, mapping, mbyte, md5, memory, regex, sha1, signal)
- libemail (attach, body, email, envelope, from, parameter, parse, rfc2047, rfc2231, tags, thread, url)

This code all lives in `test/`.
**Please check** that `make test` succeeds for all tests.
**Note**: Some tests print error messages, but they don't affect the outcome.

---

Fix code for degenerate test cases for:

- libmutt (base64, buffer, charset, date, envlist, file, hash, history, list, logging, mbyte, md5, memory, regex, sha1)
- libemail (attach, body, envelope, parameter, parse, rfc2047, rfc2231, thread, url)

These are changes to the library code.
They are almost entirely of the form `if (!ptr) return NULL`

The changes shouldn't affect the behaviour of the code, since I've only added checks where the code crashed from `NULL` pointers.

---

Current test coverage for the four libraries looks like [this](https://flatcap.org/mutt/lcov/).


flatcap:
> Not sure we want to keep adding test binaries

Yes, fair point.
I'm thinking four binaries, one for each library.

We now have a LOT of very similar Makefile rules that can probably be combined.
Also, I've got to merge the existing tests into their library-specific versions.

Issue #1680, opened 2019-04-27
Translation problem: "no binds/macros for this menu" cannot be properly translated


mgedmin:
I was translating

```po
#: icommands.c:268
#, c-format
msgid "%s: no %s for this menu"
```

and I checked the source to see what `%s` will expand to:

```c
    mutt_buffer_printf(err, _("%s: no %s for this menu"),
                       dump_all ? "all" : buf->data, bind ? "binds" : "macros");
```

and I see:

- the word "binds" or "macros" cannot be translated
- some languages may need to use different declensions for "binds"/"macros" in this context rather than in other contexts, so it would be better to translate entire sentences rather than construct them piecemeal.

flatcap:
@reisradomil Would you mind looking at this?
Thanks.

reisradomil:
Will do.

reisradomil:
I've reworded the line in question, see pull-request #1686 for more details.

@mgedmin thanks for reporting this.

flatcap:
I'll re-sync the translations shortly (I've got some large upstream changes to apply, first).

PR #1686, opened 2019-04-30
Fix piecemeal constructed translation string (issue #1680)

reisradomil:
Give the translator a complete sentence to translate instead of translating parts and combining them together.

In addition, this pull request contains
* Added hints (aka comments) for the translators of the strings
* Add filename to the error strings. (This is a unification of slightly different worded sentences)

Note:
The po files were updated but only their contents. The line numbers and other changes (form other source files) are not updated. This allows for a better review.

This fixes issue #1680.


flatcap:
Brilliant!  Thanks @reisradomil.

PR #1692, opened 2019-05-02
Do not link in a new Account after a failed login 

### LINKED ISSUE #1690: Crash when opening an IMAP mailbox after a login failure
As reported on the neomutt-devel mailling list, NeoMutt crashes when trying to open any folder after a failed IMAP login.

The proposed patch is

```diff
diff --git a/imap/imap.c b/imap/imap.c                                                                                                                      
index 60c3b3e94..d1d89fb84 100644                                                                                                                           
--- a/imap/imap.c                                                                                                                                           
+++ b/imap/imap.c                                                                                                                                           
@@ -1820,7 +1820,7 @@ int imap_sync_mailbox(struct Mailbox *m, bool expunge, bool close)                                                                    
  */                                                                                                                                                        
 struct Account *imap_ac_find(struct Account *a, const char *path)                                                                                          
 {                                                                                                                                                          
-  if (!a || (a->magic != MUTT_IMAP) || !path)                                                                                                              
+  if (!a || (a->magic != MUTT_IMAP) || !path || !a->adata)                                                                                                 
     return NULL;                                                                                                                                           
```

This kinds of fix the crash, but it looks wrong to me that we ever get an Account with attached adata.

gahr:
Fixes #1690

### DIFF
--- a/init.c
+++ b/init.c
@@ -1344,7 +1344,6 @@ static enum CommandResult parse_mailboxes(struct Buffer *buf, struct Buffer *s,
     {
       a = account_new();
       a->magic = m->magic;
-      TAILQ_INSERT_TAIL(&AllAccounts, a, entries);
       new_account = true;
     }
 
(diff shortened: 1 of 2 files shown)


init.c
```diff
@@ -1370,6 +1370,11 @@ static enum CommandResult parse_mailboxes(struct Buffer *buf, struct Buffer *s,
     {
       //error
       mailbox_free(&m);
+      if (new_account)
+      {
+        TAILQ_REMOVE(&AllAccounts, TAILQ_LAST(&AllAccounts, AccountList), entries);
```
flatcap: Am I missing something, or:
```suggestion
        TAILQ_REMOVE(&AllAccounts, a, entries);
```

gahr: heh yes of course

PR #1694, opened 2019-05-02
Unit Test updates

flatcap:
This might have been better as three PRs :-)

1. quieten successful tests
   Reduce the visual clutter with running `make test`

2. **MEGATEST**
   All the tests are now part of `neomutt-test`

3. Disable testing by default
   `./configure --testing` for Unit Testing
   `./configure --coverage` for Coverage Testing (enables Unit Testing)
   Without these options, nothing is generated in `/test`

I'm happy to invert the logic of the 3<sup>rd</sup> commit.
I just don't want to build 400 test files, _every_ time.

---

Please check that `make test` still works for you (the Makefile is now **much** smaller).


auto.def
```diff
@@ -451,11 +452,21 @@ if {[get-define want-sasl]} {
...
+  define USE_COVERAGE_TESTS
   define-append CFLAGS -fprofile-arcs -ftest-coverage
   define-append LDFLAGS -fprofile-arcs -ftest-coverage
 }
 
+###############################################################################
+# Unit Testing
+if {[get-define want-testing]} {
+  define USE_UNIT_TESTS
+  set subdirs "$subdirs test"
```
gahr: lappend subdirs test

Issue #1689, opened 2019-05-02
IMAP lgoin hangs due to "OK ENABLE completed"


zklhp:
## Expected Behaviour

I am trying to login a private IMAP server running Coremail system. NeoMutt should login correctly using IMAP.

## Actual Behaviour

NeoMutt hangs with "Logging in..." on screen.

If NeoMutt crashed, did your OS create a 'coredump' file?

## Steps to Reproduce

With config

```
set folder      = "imap://***@***:143/"
set imap_user   = "***@***"
set imap_pass   = "***"
set smtp_pass   = "***"
set spoolfile   = +INBOX
mailboxes       = +INBOX
set imap_authenticators="login"
set imap_pipeline_depth = 0
```

The log with debug level 5 look like

```
[2019-05-02 15:45:57] NeoMutt-20180716 debugging at level 5
[2019-05-02 15:45:57]<M> log_file_open() Debugging at level 5 to file '/home/kel/.neomuttdebug0'
[2019-05-02 15:45:57]<2> mutt_window_reflow() entering
[2019-05-02 15:45:57]<2> mutt_window_reflow() entering
[2019-05-02 15:45:57]<2> mutt_window_reflow() entering
[2019-05-02 15:45:57]<2> mutt_window_reflow() entering
[2019-05-02 15:45:57]<2> mutt_window_reflow() entering
[2019-05-02 15:45:57]<2> mutt_window_reflow() entering
[2019-05-02 15:45:57]<2> source_rc() Reading configuration file '/etc/neomuttrc'.
[2019-05-02 15:45:57]<5> parse_attach_list() added */.* [9]
[2019-05-02 15:45:57]<5> parse_attach_list() added text/x-vcard [7]
[2019-05-02 15:45:57]<5> parse_attach_list() added application/pgp.* [2]
[2019-05-02 15:45:57]<5> parse_attach_list() added application/x-pkcs7-.* [2]
[2019-05-02 15:45:57]<5> parse_attach_list() added text/plain [7]
[2019-05-02 15:45:57]<5> parse_attach_list() added message/external-body [4]
[2019-05-02 15:45:57]<5> parse_attach_list() added message/external-body [4]
[2019-05-02 15:45:57]<2> source_rc() Reading configuration file '/home/kel/.mutt/muttrc'.
[2019-05-02 15:45:57]<3> imap_parse_path() Using default IMAP port 143
[2019-05-02 15:45:57]<3> imap_parse_path() Using default IMAPS port 993
[...]

PR #1698, opened 2019-05-03
Allow to disable IMAP's ENABLE extension to work-around broken servers

gahr:
Issue #1689


gahr:
This is not a bug, just a work around to allow users to connect to broken servers.

sileht:
This looks like a server bug, but why not.

PR #1701, opened 2019-05-05
Upstream changes

flatcap:
These are mostly `char*` -\> `struct Buffer` conversions, however the first two are tricky.

- 4a02618b4 Change `Mailbox->realpath` from `char[PATH_MAX]` to `char *`
- e4133e5cd Change `Mailbox->path` from `char[PATH_MAX]` to `struct Buffer`

As you can imagine, these caused a lot of churn to the code.
I **think** that all my merges are correct and nothing's been missed.

Please test with care.


flatcap:
This brings us up-to-date with upstream.

gahr:
Running the branch now - looks good so far

PR #1707, opened 2019-05-09
Fix bugs raised by automated testing

flatcap:
- Replace `atoi()` uses with `mutt_str_atoi()`
  which has range checking and returns errors

- Fix memory leaks during header parsing
  These were triggered by some fairly degenerate headers

- Check the value of the 'Content-Length' header more carefully
  Large numbers could cause the `long` to overflow

Thanks to LGTM and oss-fuzz.

### DIFF
--- a/address/address.c
+++ b/address/address.c
@@ -566,10 +566,10 @@ struct Address *mutt_addr_parse_list(struct Address *top, const char *s)
       /* add group terminator */
       cur = mutt_addr_new();
       if (last)
-      {
         last->next = cur;
-        last = cur;
-      }
+      else
+        top = cur;
+      last = cur;
 
       phraselen = 0;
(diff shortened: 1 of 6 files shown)


flatcap:
Done.
These bugs were picked up by fuzzing.
It's not probable that an email with these flaws would end up in your inbox.

address/address.c
```diff
@@ -566,10 +566,10 @@ struct Address *mutt_addr_parse_list(struct Address *top, const char *s)
       /* add group terminator */
       cur = mutt_addr_new();
       if (last)
-      {
         last->next = cur;
-        last = cur;
-      }
+      else
+        top = cur;
+      last = cur;
```
gahr: One of these days Address needs to be converted to a STAILQ

PR #1708, opened 2019-05-10
feat: use mailbox description for spoolfile

Austin-Ray:
'spoolfile' is limited to mailboxes paths so if one wants to use a
'named-mailboxes' or 'virtual-mailboxes' as the spoolfile, its path must
be duplicated in the 'spoolfile' assignment. 'virtual-mailboxes' has a
helper variable, 'virtual_spoolfile', for selecting the first
'virtual-mailboxes' entry as the spoolfile.

This commit allows one to use a mailbox description for the 'spoolfile,
which removes the need to duplicate a 'named-mailboxes' or
'virtual-mailboxes' path and makes 'virtual_spoolfile' defunct. The
configuration syntax is now more consistent.

It is important to note that mailbox descriptions are preferred to
paths.

---

Snippet for testing

```rc
named-mailboxes "named-mailbox" "=INBOX"
set spoolfile="named-mailbox"
```

### DIFF
--- a/doc/manual.xml.head
+++ b/doc/manual.xml.head
@@ -15272,12 +15272,13 @@ virtual-mailboxes "My INBOX" "notmuch://?query=tag:inbox"
 
         <table id="table-notmuch-variables">
           <title>Notmuch Variables</title>
-          <tgroup cols="3">
+          <tgroup cols="4">
             <thead>
               <row>
                 <entry>Name</entry>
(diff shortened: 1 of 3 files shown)


gahr:
Could you please also update the docs?

Austin-Ray:
Documentation added. Please let me know if there's anything else that needs to be done.

PR #1713, opened 2019-05-17
Convert Address to use TAILQ

gahr:
## What I have done

1. Converted `struct Address` to use TAILQ instead of a hand-made raw list using a `->next` pointer

```diff
 /**
  * struct Address - An email address
  */
 struct Address
 {
   char *personal; /**< real name of address */
   char *mailbox;  /**< mailbox and host address */
   bool group : 1; /**< group mailbox? */
   bool is_intl : 1;
   bool intl_checked : 1;
-  struct Address *next;
+  TAILQ_ENTRY(Address) entries;
 };
+TAILQ_HEAD(AddressList, Address);
```
2. Fixed the API in `address.h` so that function that handle a list of addresses (`struct AddressList`) are named `mutt_addrlist_*` and functions that handle a single address (`struct Address`) are named `mutt_addr_...

3. Made everything compile and run fine with 1. and 2. (which is most of the churn introduced)

## What I still need to do

1. Enhance the test suite for address - I really want to be sure everything works fine

2. More field tests - I am running this branch but I want to stress more use cases

## What can you do

1. Code review - there's plenty to read

2. Field tests - please run this branch with your setup and report any crashes / misbehaviours


flatcap:
That's an amazing job!  Thanks.

I've pushed some changes, though they're not as drastic as the diff suggests.

- Run clang-format over the code
- Move the 'Implement mutt_addresslist_new' before the code that uses it
- Add `void` to `()` (pesky C++ programmers)
- Add/fix Doxygen header blocks
- Initialise all pointers to `NULL`
- Update function prototypes to exactly match functions
- Merge a couple of minor fix commits

These changes have been back-dated to keep the history tidy.
For comparison, the original code is in [[devel/address-stailq-orig](https://github.com/neomutt/neomutt/tree/devel/address-stailq-orig)]

I'll read the code in depth tomorrow.

gahr:
Wow - thanks a lot for the cleanup!

flatcap:
For testing (and another PR on another day), how do you feel about:

- All library functions must return something (simplifies tests, slightly)
  (even if it's just `true` you managed to supply all the mandatory params)

- No static functions in libraries (so all functions are testable)
  Or perhaps a conditional `#define LIB_STATIC static` (disabled when testing is enabled)

gahr:
Thanks for the review and for fixing stuff for me :)

> In config/address.c you note that the config system can only store single addresses.
Can you think of any features that might benefit from config addresss lists? (for future reference)

Uhm not really.. I thought about that but couldn't come up with a use case... not yet at least

> Do we need both mutt_addrlist_free() and mutt_addrlist_free_all()?

Probably not: if I change `result_to_addr` to take an `AddressList` to fill in, there will be no need for `mutt_addrlist_free` at all. Also, `mutt_addrlist_free_all` sucks a bit as a name.. how about `mutt_addrlist_clear`? We'd kill `mutt_addrlist_new` too: all AddressList structs would be on the stack.

>  All library functions must return something (simplifies tests, slightly)

I thought about that - that'd make testing easier but I don't like returning a dummy value just to be able to wrap the function call in `TEST_CHECK`. Another possibility would be `TEST_CHECK((null_function(), 1))` but that sucks too... how about just calling the function in the test function? There's no need for `TEST_CHECK` after all.

> No static functions in libraries (so all functions are testable)

I don't think I like this. We want to test the API and make sure *callable* functions behave as advertised. Whatever a module does internally, such as refactoring common code into static functions, should be the module's business only. I haven't given this much thought, so this is not a strong opinion.

flatcap:
> how about `mutt_addrlist_clear`?
> We'd kill `mutt_addrlist_new` too: all AddressList structs would be on the stack.

Sounds good.

> > All library functions must return something

> ... but that sucks too...
> how about just calling the function in the test function?
> There's no need for `TEST_CHECK` after all

Yeah OK, as long as we've got `TEST_MSG`s so we can follow what's happening.

> > No static functions in libraries

> I don't think I like this.
> We want to test the API and make sure _callable_ functions behave as advertised.

Yes, OK.  Black boxes.

Without a specification for the API, we'll have to pick our test cases carefully and use coverage as a metric.

gahr:
> Yeah OK, as long as we've got TEST_MSGs so we can follow what's happening.

I think I'm missing something. The purpose of this test is to make sure the call to `mutt_addr_cat` does not crash. What's the purpose of `TEST_CHECK`? I would simplify tests such as this one to just call the function.

https://github.com/neomutt/neomutt/blob/8117d4703130c27e6b9e9319a29a21ea1d5f3daa/test/address/mutt_addr_cat.c#L38-L39

flatcap:
>  would simplify tests such as this one to just call the function.

The `TEST_CHECK` gives you feedback when you run `neomutt-test -v`

If you run `mutt_addr_cat()` lots of times and one crashes, the `TEST_CHECK` would tell you which one.

gahr:
Mh, kind of.. it'd give you an idea because you'd see the message printed by the last test to succeed. Not beautiful, but probably the best we can do.. unless we enhance acutest with a `TEST_NOCRASH(expr)`

flatcap:
> unless we enhance acutest with a `TEST_NOCRASH(expr)`

oh yeah.  it didn't even occur to me to change acutest :-)

Issue #1714, opened 2019-05-20
entire-thread jumps to first message in thread


anarazel:
## Expected Behaviour

It used to that showing an entire-thread (with notmuch configured), would fetch all the messages for the thread, keeping focus on the message for which the thread was fetched.

## Actual Behaviour

Now the thread is fetched, and the first message in the thread is focussed. Which makes it much harder to read responses etc, which is what I'm usually trying to do when using entire-thread. I have to resort to afterwards filtering with ~i, and then show all messages again. Pretty cumbersome.

## Steps to Reproduce

configure notmuch + mutt
bind 'entire-thread' do key (bind index,pager . entire-thread)
focus message that's in a thread, but *not* the first message in that thread
use entire-thread 

the thread will be fetched, and the first message in the thread will have the focus.

## How often does this happen?

Always

## When did it start to happen?

Bisected down to 918885f32db41a452c23379c62dd2a9a752930d1 by @gahr. Haven't dug deeper into the issue, but appears to have been an unintended side-effect.

gahr:
Thanks for the detailed report and the bisect, @anarazel . Yep - that was totally unintended. I am not a notmuch user myself, so I must have missed some use case. I'll try and look into it.

anarazel:
Quick note: I've debugged this, writing up explanation right now.

gahr:
Cool - patches are welcome :)

anarazel:
After a bit of debugging, I think the bug currently - the code has changed substantially since the commit pointed out above - resides in the OP_MAIN_ENTIRE_THREAD case of mutt_index_menu():

```C
        int oc = Context->mailbox->msg_count;
        if (nm_read_entire_thread(Context->mailbox, CUR_EMAIL) < 0)
        {
          mutt_message(_("Failed to read thread, aborting"));
          break;
        }
        if (oc < Context->mailbox->msg_count)
        {
          struct Email *oldcur = CUR_EMAIL;

          if ((C_Sort & SORT_MASK) == SORT_THREADS)
            mutt_sort_headers(Context, false);
          menu->current = oldcur->virtual;
```
One can see that `oldcur` is determined *after* the nm_read_entire_thread(). But the latter does a `mutt_mailbox_changed(m, MBN_INVALID);`, which ends up triggering a `mutt_sort_headers()` which in turn does `mutt_sort_threads()`. Thus, by the time `oldcur` is determined, it's already pointing to the start of the thread.

[...]

PR #1715, opened 2019-05-20
entire-thread: determine current position before resorting.

anarazel:
Previously oldcur was determined after nm_read_entire_thread(),
leading to entire-thread unintentionally jumping to the first message
in the thread.

Before 918885f32db41 that used to work, because mutt_sort_headers()
wasn't called from within nm_read_entire_thread().

Fix by moving oldcount computation a few lines up. Rename oc variable,
to avoid confusion.

Fixes: #1714

### DIFF
--- a/index.c
+++ b/index.c
@@ -1983,16 +1983,15 @@ int mutt_index_menu(void)
         }
         if (!prereq(Context, menu, CHECK_IN_MAILBOX | CHECK_MSGCOUNT | CHECK_VISIBLE))
           break;
-        int oc = Context->mailbox->msg_count;
+        oldcount = Context->mailbox->msg_count;
+        struct Email *oldcur = CUR_EMAIL;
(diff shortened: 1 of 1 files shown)


gahr:
The proposed change makes sense to me - requesting review from notmuch-aware people.

anarazel:
@Austin-Ray 
> Changing oc to oldcount shadows the parameter oldcount. We should pick a different name.

Ugh, yes. I wonder if is there a reason to not just set that oldcount, rather than declaring a separate variable?

Austin-Ray:
>  wonder if is there a reason to not just set that oldcount

I do not see a reason. `oldcount` is declared within `mutt_index_menu` and isn't after the switch statement. It seems safe.

@flatcap: Thoughts?

flatcap:
OK.  Reuse the `oldcount` for now and we'll merge the fix.
If anyone wants to tidy/refactor/etc in another PR, that'd be welcomed.

anarazel:
Ok, done.

PR #1718, opened 2019-05-21
Avoid redundant sort in entire-thread.

anarazel:
918885f added, indirectly, a mutt_sort_headers() call to
nm_read_entire_thread() (via mutt_mailbox_changed() ->
ctx_mailbox_changed() -> ctx_update()), but didn't remove the sort
after nm_read_entire_thread() in mutt_index_menu().

Remove the now superfluous sort.


gahr:
LGTM - requesting a review from our NotMuch guy :)

PR #1722, opened 2019-05-25
doc: add example to find all maildir mailboxes

toogley:
* **What does this PR do?**

When you have many Maildir mailboxes (or if you change it regularly), it can be cumbersome to maintain the `mailboxes` list manually. This PR adds an example to automate that.

**NOTE**: I don't think the section ["Monitoring new Mail"](https://neomutt.org/guide/configuration.html#mailboxes) is correct for this example - but i haven't found a better place, since this section is the target-link for the `mailboxes` command in the [reference](https://neomutt.org/guide/reference#2-%C2%A0configuration-commands)

### DIFF
--- a/doc/manual.xml.head
+++ b/doc/manual.xml.head
@@ -9306,6 +9306,14 @@ subjectrx '\[[^]]*\]:? *' '%L%R'
             don't completely rely on this feature.
           </para>
         </note>
+        <note>
+          <para>
+            When using Maildir, you don't have to manually specify all your mailboxes. You can use this command instead:
+            <screen>
+              mailboxes `find ~/.mail/ -type d -name cur | sed 's:/cur/*$::' | tr '\n' ' '`
+            </screen>
+          </para>
+        </note>
       </sect2>
 
       <sect2 id="calc-mailbox-counts">


flatcap:
Thanks, @toogley 

> `find ~/.mail/ -not -name 'cur' -not -name 'tmp' -not -name 'new' -mindepth 1 -type d -exec basename {} \; | awk '{ print "="$0 }' | tr '\n' ' '`

I think we can simplify this example.

The `basename` and `awk` turn `/home/user/.mail/apple` into `=apple`, but NeoMutt doesn't care.
If `$folder` is set, they'll be abbreviated, if not, they'll still work.

`find ~/.mail/ -not -name cur -not -name 'tmp' -not -name 'new' -mindepth 1 -type d | tr '\n' ' '`

This might find some non-email dirs, but inverting the search might help.
(I'm not sure how to merge `tr` into `sed` in a portable way).

`find ~/.mail/ -type d -name cur | sed 's:/cur$::' | tr '\n' ' '`

Also, there's a GNU find way that's even shorter:

`find ~/.mail/ -type d -name cur -printf "%h "`

toogley:
> find ~/.mail/ -type d -name cur | sed 's:/cur$::' | tr '\n' ' '

Since Neomutt is portable i think choosing a non gnu-find solution makes sense. However, i don't understand the your sed command - why the colons? the manpages say this is a label, but i don't understand the need here.

I'd suggest `find ~/.mail/ -type d -name cur | sed 's/\/cur$//' | tr '\n' ' '`. That removes also the last path deliminter. Or does your solution has advantages?

flatcap:
> ... portable

Certainly, but there's no reason we can't have both in the contrib section

> sed, colon

`sed`'s very versatile, but the default delimiter is `/`.
When you want to deal with paths this can be a pain.

The delimiter can be one of many punctuation marks, colon included, so these are equivalent:
- `s/\/cur//`
- `s:/cur::`
- `s!/cur!!`

but in the last two, we don't need to escape the `/`

> removes also the last path deliminter

Hmm... Yeah, that might vary by `find` version.
How about:

`find ~/.mail/ -type d -name cur | sed 's:/cur/*$::' | tr '\n' ' '`

toogley:
I dislike the contrib directory. I don't think users expect such information there.

> The delimiter can be one of many punctuation marks, colon included, so these are equivalent: 

Ah, nice. Thanks. I think `find ~/.mail/ -type d -name cur | sed 's:/cur/*$::' | tr '\n' ' '` is best.

PR #1723, opened 2019-05-26
make OP_HALF_{UP,DOWN} behave symmetric.

sstark:
- the calculation of number of lines to scroll up/down should not
  depend on whether the number of rows in the pager/index/terminal is
  odd or even.
  - this patch will make the behaviour symmetric such that in both cases
  (even/odd number of rows) scrolling up and down by half a page (or the
  other way round) will get you back to the exact same line as before.

### DIFF
--- a/pager.c
+++ b/pager.c
@@ -2573,8 +2573,8 @@ int mutt_pager(const char *banner, const char *fname, PagerFlags flags, struct P
       case OP_HALF_UP:
         if (rd.topline)
         {
-          rd.topline = up_n_lines(rd.pager_window->rows / 2, rd.line_info,
-                                  rd.topline, rd.hide_quoted);
+          rd.topline = up_n_lines(rd.pager_window->rows / 2 + (rd.pager_window->rows % 2),
+                                  rd.line_info, rd.topline, rd.hide_quoted);
         }
         else
           mutt_error(_("Top of message is shown"));


flatcap:
Welcome to NeoMutt, @sstark.
Thanks for the PR, very nice attention to detail :-)

For anyone wanting to see the problem / test the fix:

Config: `pager.rc`
```
bind pager [ half-up
bind pager ] half-down

# Down / Up 100 times
macro pager <F5> [[[[[[[[[[[[[[[[[[[[[[[[[[[[[[[[[[[[[[[[[[[[[[[[[[[[[[[[[[[[[[[[[[[[[[[[[[[[[[[[[[[[[[[[[[[[[[[[[[[[
macro pager <F6> ]]]]]]]]]]]]]]]]]]]]]]]]]]]]]]]]]]]]]]]]]]]]]]]]]]]]]]]]]]]]]]]]]]]]]]]]]]]]]]]]]]]]]]]]]]]]]]]]]]]]

macro pager <F7> ':set invhelp<enter>'
```

Use this [long.mbox](https://github.com/neomutt/neomutt/files/3220200/long.mbox.txt)

`neomutt -n -F pager.rc -f long.mbox`

Scroll down a long way, then use <kbd>F5</kbd> / <kbd>F6</kbd> to jump up and down.
You should return to exactly the same place.

Then, toggle the help line with <kbd>F7</kbd> (this ensures the fix works for both odd and even numbers of lines).

sstark:
I was hesitant calling this a bug as it's probably just hurting some symmetry fetishists feelings :)

This bug is also present in mutt. Is it important for neomutt to have upstream receive fixes like this? Is mutt still considered "upstream"?

flatcap:
> I was hesitant calling this a bug as it's probably just hurting some symmetry fetishists feelings :)

There are a lot of ancient Mutt users who very scared by change :-)

> This bug is also present in mutt.
> Is it important for neomutt to have upstream receive fixes like this?

It's not important to me -- I don't see eye-to-eye with Kevin (the Mutt maintainer).
However, they're more open to contributions now.

I'm sure they'd be happy to merge your improvement (probably best not to mention NeoMutt, though ;-)

https://gitlab.com/muttmua/mutt

> Is mutt still considered "upstream"?

Just about (NeoMutt's code has diverged a LONG way), but...

Every Mutt commit has been applied to NeoMutt (as of a few days ago).
It's getting quite time-consuming, but it's still worth the effort.

PR #1729, opened 2019-05-29
Improve mailcap handling

anarazel:
This PR is primarily to get feedback about whether this is a reasonable approach. And not yet whether it can be committed as is.

The goal of the first patch is to increase the chances of finding a mailcap match. In my experience often "description" will have actual content, even though "filename" doesn't. Therefore it makes sense to match for description if filename doesn't find a match.

The second one is to allow to format attachments correctly. I use a syntax highlighter for patch attachments, but without this path that breaks when composing emails, as the attachment will be formatted with that syntax highlighter as well. Mailcap has 'compose' entries for that, but mutt didn't use them.

### DIFF
--- a/handler.c
+++ b/handler.c
@@ -540,7 +540,7 @@ static int autoview_handler(struct Body *a, struct State *s)
   int rc = 0;
 
   snprintf(type, sizeof(type), "%s/%s", TYPE(a), a->subtype);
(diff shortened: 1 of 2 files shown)


flatcap:
- Rebased
- Tidied commit messages
- Expanded evil `<tab>`

anarazel:
Thanks!
-- 

PR #1732, opened 2019-06-05
Update to latest autosetup, use @include feature

gahr:
By using autosetup's new @include feature, we are left with a single Makefile at build time. This should make it easier to spot dependency errors and faster to process.


flatcap:
The out-of-tree build doesn't work any more:

```
mkdir p; cd p
../configure
```

- Error: Makefile.autosetup:471: Include file po/Makefile.autosetup is missing

flatcap:
Commit/title: `s/lastest/latest/`

PR #1733, opened 2019-06-05
Save attachments without prompting.

### LINKED ISSUE #1603: Enhance user experience when saving attachments
From IRC:

> Anyone got any suggestions on improving the user-experience for saving lots of attachments? At the moment, I view attachments, tag (lots), then `;s`, choose a directory, then I have to confirm every save. Is there a better way just to take all the ones I've tagged and save to directory in one go?

How about we change the logic as follows:

- ask for a directory to save the attachment, default to `attach_save_dir`
- for each attachment
  - ask for a filename only if the attachment doesn't have one
  - ask for a confirmation only if saving the attachment would overwrite an existing file
    - Y -> overwrite this one
    - N -> do not overwrite this one, but keep looping on the other attachments
    - A -> overwrite this one, remember the choice for the other attachments in the loop
    - C -> cancel, quit looping

dharrigan:
If the 'save_without_prompt' variable is set (to true), then when saving
attachments the user will not be prompted each time for the name of the file.

Handling conflicting filenames is also handled.

closes #1603 

-=david=-

### DIFF
--- a/recvattach.c
+++ b/recvattach.c
@@ -71,6 +71,7 @@
 
 /* These Config Variables are only used in recvattach.c */
 char *C_AttachSaveDir; ///< Config: Default directory where attachments are saved
+char *C_AttachSaveWithoutPrompting; ///< Config: If true, then don't prompt to save
 char *C_AttachSep; ///< Config: Separator to add between saved/printed/piped attachments
 bool C_AttachSplit;    ///< Config: Save/print/pipe tagged messages individually
 bool C_DigestCollapse; ///< Config: Hide the subparts of a multipart/digest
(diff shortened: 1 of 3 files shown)


recvattach.c
```diff
@@ -581,6 +582,57 @@ static int query_save_attachment(FILE *fp, struct Body *body, struct Email *e, c
...
+                     mutt_is_message_type(body->type, body->subtype));
+
+  if (is_message)
+  {
+    mutt_str_strfcpy(tfile, buf, sizeof(tfile));
+  }
+  else
+  {
+    rc = mutt_check_overwrite(body->filename, buf, tfile, sizeof(tfile), &opt, NULL);
+    if (rc == -1 || rc == 3) // abort or cancel
```
flatcap: `mutt_check_overwrite()` doesn't return `3`

dharrigan: Fixed.

PR #1738, opened 2019-06-07
Optionally use pkg-config for 3rd party dependencies

gahr:
This commit adds a new --pkgconf switch to configure. When specified, the compiler and linker flags for most 3rd party dependencies will be figured out using pkg-config.

### DIFF
--- a/auto.def
+++ b/auto.def
@@ -9,7 +9,7 @@ if {![catch {exec uname} out] && $out eq {Linux}} {
 autosetup-require-version 0.6.8
 
 # Standard autosetup modules and our own modules
-use system cc cc-lib mutt-gettext mutt-iconv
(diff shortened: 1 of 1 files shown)


flatcap:
I gave it a try with: `./configure --pkgconf --everything` but it failed with:
```
Checking for gpgme >= 1.4.0...not found
auto.def:206: Error: bad class "true": must be alnum, alpha, ascii, boolean, control, digit, double, graph, integer, lower, print, punct, space, upper, or xdigit
```

(I have gpgme-1.12.0)

gahr:
Ouch - I forgot to test with the bundled Tcl implementation

gahr:
I think we have a reasonable subset to start testing

PR #1739, opened 2019-06-07
Add 'account' command

flatcap:
## New Config Command: `account`

This is the beginning of Account-specific config.

- 5858405d6 add `account` command
  Add a new command and a stack for nested `account` commands

- b30988e1c update config tests
  Fix tests for named `Account`s

This is a quick hack for demonstration purpuses:

- 3f89b0b0f use `Account`-specific `$index_format`

The last two commits are just for testing purposes.
They generate graphviz diagrams and dump Account info.

- 4bc14be9d add dot_dump()
- a1e5b4e65 add debug logging

---

## Setup

- Build this PR, branch [devel/account](https://github.com/neomutt/neomutt/tree/devel/account)
- Clone the [sample-mail](https://github.com/neomutt/sample-mail) repo
- Set an environment variable to the `sample-mail` repo
  `export DEMO_DIR=/home/user/sample-mail`

There are two sample config files: `demo.rc` and `base.rc` (which sources `child.rc`)
They are **equivalent**.

`demo.rc` begins:

```conf
account cars
  set folder = "$DEMO_DIR/cars"
  set index_format = "CARS %4C %Z %{%b %d} %-15.15L (%?l?%4l&%4c?) %s"
  set sort = "from"
  set sort_aux = "from"

  mailboxes "+audi"
  mailboxes "+bentley"
  mailboxes "+chrysler"
  mailboxes "+daimler"
```

## Notes

Leading whitespace is ignored -- it's included for clarity.

The `account` command's scope ends at:
- A new `account NAME` command
- An empty `account` command
- End of file

To use the Account-specific config, the code needs changing to lookup the value, rather than use the global.

In this PR:
- Only four config items have inheritance
- Only `$folder` and `$index_format` are actually used

## Testing

The config files are stripped to an absolute minimum.
They show the sidebar and enable:

- <kbd>Ctrl-P</kbd> - Sidebar previous
- <kbd>Ctrl-N</kbd> - Sidebar next
- <kbd>Ctrl-O</kbd> - Sidebar open

The code can be tested in two ways:

### Normal GUI mode

Use either of `demo.rc` or `base.rc`

`./neomutt -n -d5 -F $DEMO_DIR/demo.rc`

Opening mailboxes in different accounts will show up in the custom `$index_format`
Pressing <kbd>M</kbd> (`<show-log-messages>`) will generate a graphviz diagram of the objects.

### Batch mode

Enabling 'batch mode' with `neomutt -B`, will parse the config, then dump the results.
It will also generate a graphviz diagram of the objects (because I like diagrams :-)

Use either of `demo.rc` or `base.rc`

`./neomutt -n -d5 -F $DEMO_DIR/demo.rc -B`

## What's Next?

- Documentation

- More inheritance
  Each inherited config item will require careful code changes

- Sidebar updates
  The Sidebar needs to be taught about named Accounts

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -39,6 +39,7 @@ git:
 
 install:
   - git clone --depth 1 https://github.com/neomutt/travis-build.git ~/config
+  - git clone --depth 1 https://github.com/neomutt/neomutt-test-files.git ~/test-files
 
 before_script:
   - ccache --zero-stats
@@ -47,6 +48,8 @@ before_script:
   - export -f travis_nanoseconds
   - export -f travis_time_finish
   - export -f travis_time_start
+  - export NEOMUTT_TEST_DIR="$HOME/test-files"
+  - (cd ~/test-files && ./setup.sh)
 
 script:
   - ~/config/build
--- a/Makefile.autosetup
+++ b/Makefile.autosetup
@@ -71,7 +71,7 @@ NEOMUTTOBJS=	addrbook.o alias.o bcache.o browser.o commands.o \
 		muttlib.o mx.o mx_path.o myvar.o pager.o pattern.o postpone.o progress.o query.o \
 		recvattach.o recvcmd.o resize.o rfc3676.o score.o send.o sendlib.o \
 		sidebar.o smtp.o sort.o state.o status.o system.o version.o \
-		mutt_commands.o mutt_config.o command_parse.o
+		mutt_commands.o mutt_config.o command_parse.o tracker.o
 
 @if !HAVE_WCSCASECMP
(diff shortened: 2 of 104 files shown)


flatcap:
This has been absorbed into #2104 (refactor path handling).

toogley:
@flatcap 

to give at least some feedback on the idea to avoid decrease of motivation: i really like the idea of this PR, but i haven't yet carefully evaluated my use case of multiple accounts to see if this PR works for me in detail, or what would be missing. Therefore i haven't said something yet, but to be honest i don't know when i would have time to do so. 

so sorry for being inactive :(

But in general, i really like the progress you and the neomutt team have managed!

PR #1740, opened 2019-06-08
agentcrypt: password encryption with ssh-agent

ndilieto:
This commit adds a new 'agentcrypt' feature to NeoMutt, which allows
to securely store encrypted passwords in .neomuttrc and decrypt them
without typing anything, as long as the ssh key used to encrypt the
password is loaded or forwarded by the ssh-agent.

For more information see the modified imap_pass, pop_pass and smtp_pass
documentation in neomuttrc.5

The new feature is disabled by default as it depends on libagentcrypt
which is available at https://github.com/ndilieto/libagentcrypt

To enable it, first install libagentcrypt and then add --agentcrypt to
NeoMutt's ./configure invocation before rebuilding.

### DIFF
--- a/auto.def
+++ b/auto.def
@@ -59,6 +59,9 @@ options {
   # SASL (IMAP and POP auth)
   sasl=0                    => "Use the SASL network security library"
   with-sasl:path            => "Location of the SASL network security library"
+  # agentcrypt (use ssh agent to encrypt/decrypt passwords)
+  agentcrypt=0              => "Use the agentcrypt library (https://github.com/ndilieto/libagentcrypt)"
+  with-agentcrypt:path      => "Location of the agentcrypt library"
 # Lua
   lua=0                     => "Enable Lua scripting support"
   with-lua:path             => "Location of Lua"
@@ -107,9 +110,9 @@ options {
 if {1} {
   # Keep sorted, please.
   foreach opt {
-    bdb backtrace coverage doc everything fmemopen full-doc gdbm gnutls gpgme
-    gss homespool idn idn2 inotify kyotocabinet lmdb locales-fix lua mixmaster
-    nls notmuch pgp qdbm sasl smime ssl testing tokyocabinet
+    agentcrypt bdb backtrace coverage doc everything fmemopen full-doc gdbm
+    gnutls gpgme gss homespool idn idn2 inotify kyotocabinet lmdb locales-fix
+    lua mixmaster nls notmuch pgp qdbm sasl smime ssl testing tokyocabinet
   } {
     define want-$opt [opt-bool $opt]
   }
@@ -118,8 +121,8 @@ if {1} {
   # relative --enable-opt to true. This allows "--with-opt=/usr" to be used as
   # a shortcut for "--opt --with-opt=/usr".
   foreach opt {
-    bdb gdbm gnutls gpgme gss homespool idn idn2 kyotocabinet lmdb lua mixmaster
-    ncurses nls notmuch qdbm sasl slang ssl tokyocabinet
+    agentcrypt bdb gdbm gnutls gpgme gss homespool idn idn2 kyotocabinet lmdb
+    lua mixmaster ncurses nls notmuch qdbm sasl slang ssl tokyocabinet
   } {
     if {[opt-val with-$opt] ne {}} {
       define want-$opt 1
@@ -450,6 +453,17 @@ if {[get-define want-sasl]} {
   }
   if {![get-define USE_SASL]} {
     user-error "Unable to find SASL"
+  }
+}
+
+# AGENTCRYPT
+if {[get-define want-agentcrypt]} {
+  if {[check-inc-and-lib agentcrypt [opt-val with-agentcrypt $prefix] \
+                         libagentcrypt.h agc_encrypt agentcrypt]} {
+     define USE_AGENTCRYPT
+  }
+  if {![get-define USE_AGENTCRYPT]} {
+    user-error "Unable to find AGENTCRYPT"
   }
 }
 
--- a/init.h
+++ b/init.h
@@ -1660,9 +1660,18 @@ struct ConfigDef MuttVars[] = {
   ** prompt you for your password when you invoke the \fC<imap-fetch-mail>\fP function
   ** or try to open an IMAP folder.
   ** .pp
+  ** If NeoMutt was compiled with agentcrypt support and the string begins with the
+  ** \fBACT:\fP or \fBact:\fP prefix, an attempt to decrypt it is made.
+  ** The encrypted password can be obtained by launching the command
+  ** \fBagentcrypt -t\fP and typing the password followed by a newline.
+  ** If decryption fails for any reason, NeoMutt will prompt you for the password
+  ** instead.
+  ** .pp
   ** \fBWarning\fP: you should only use this option when you are on a
   ** fairly secure machine, because the superuser can read your neomuttrc even
-  ** if you are the only one who can read the file.
+  ** if you are the only one who can read the file. If you use agentcrypt, note
+  ** that the superuser can also decrypt the password, but only when you are
+  ** logged in and have your key loaded or forwarded by ssh-agent.
   */
   { "imap_passive",             DT_BOOL, R_NONE, &C_ImapPassive, true },
   /*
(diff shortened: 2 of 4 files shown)


flatcap:
Welcome to NeoMutt, @ndilieto.

Thanks for the PR, it sounds pretty useful.

It's going to require a bit of discussion whether we accept your changes.
If `agentcrypt` were an established library, with lots of users (and packaged support), the discussion would be much simpler.
However:

- You joined GitHub two months ago
- You created the `agentcrypt` five **hours** ago
- `agentcrypt` uses a license I can't quite identify

I understand, everyone has to start somewhere.

Thanks.

ndilieto:
> 
>     * You joined GitHub two months ago

If you need references, I am the author of https://github.com/ndilieto/uacme which is at the moment the only Let's Encrypt client written in C that supports the standardized protocol (ACMEv2, RFC8555). It's listed on their third party client options: https://letsencrypt.org/docs/client-options/ and is currently being packaged for OpenWRT: https://github.com/openwrt/packages/pull/9150

> 
>     * You created the `agentcrypt` repo five **hours** ago

Is that relevant? I've been using it for months on my own server and put it on github to have it pulled by you.

> 
>     * `agentcrypt` uses a license I can't quite identify
> 

It's the ISC license. You can basically do whatever you want with it as long as you don't strip the copyright. It's also compatible with the GPL. https://en.wikipedia.org/wiki/ISC_license

flatcap:
> >  You created the `agentcrypt` repo five **hours** ago

> Is that relevant?

I little, yes.
It gives the _impression_ that this software was thrown together in an afternoon.

You're asking me to merge some code into my project, so I'm looking for some reassurance that the code is safe, reliable, robust.

> I've been using it for months on my own server

OK, fair enough.

> It's the ISC license.

Ah, lovely.

flatcap:
I'm a coder, not a project manager, so let me be open about everything I've been thinking...

What is `agentcrypt`?
- Uses ssh-agent as a secure password store
- third-party library
- new, unpackaged

Can't we do this, already?
- Using backticks (``` `` ```)
  - Do we just need a **quiet** way to run external programs?
- Using Lua (maybe)

Who's the expected audience?
- Are there lots of potential users
- What are the alternatives?
- Would **I** use it?

There are **lots** of password managers out there.
Is this the right solution?
- Yes.  Small dedicated library
- No.  Too specific, want a Password API

How can users, use it?
- Need to clone, build and install their own library
- Bundle `agentcrypt` with NeoMutt?

Reliability?
- Is is secure?
  It deals with passwords, after all
- Is is robust?
  If a library goes wrong, it'll crash NeoMutt
- Who's going to support the code?
  (and for how long?)

What else do we need?
- Documentation (once we've sorted out the above)

An article I read on Open Source Project Management, suggested that you should always merge anything you're offered.
(You can always sort out any problems later :-)

ndilieto:
>     * Uses ssh-agent as a secure password store

To be precise, it doesn't use ssh-agent to store, but rather to encrypt passwords. The encrypted password is stored in the $imap_pass variable

> 
> Who's the expected audience?

My use case: I use NeoMutt on my virtual private server to read email. I don't want to type passwords but I also don't want to store them in the clear - a break-in to the server would expose them. 
 
>     * Are there lots of potential users
Anyone who uses NeoMutt on a network machine.

>     * What are the alternatives?
One could encrypt the whole file where $imap_pass and similar are stored, then decrypt upon running NeoMutt. But this would be ugly and less secure (the decrypted file would need to be around while running). With my solution the decrypted password only lives in RAM while NeoMutt runs. 

We could actually also improve security further by making NeoMutt use the agc_malloc function to allocate the memory for the copy of the decrypted password. That is because on systems that support locking memory, agc_malloc locks it (so it cannot be swapped out to disk). The reason I haven't done this in the PR is that I wanted to keep it as small as possible, and also disabled by default.

>     * Would **I** use it?
I have been using it for months (even though I only decided to publish it recently. I hope you see the amount of effort that has already gone into it)

> 
> There are **lots** of password managers out there.
> Is this the right solution?
> 
>     * Yes.  Small dedicated library
> 
>     * No.  Too specific, want a Password API
Not quite sure what you mean with this one.  Why would that be bad? 

> How can users, use it?
> 
>     * Need to clone, build and install their own library
If you pushed the commit (which by default leaves it disabled) and users found it useful, at some point some distribution will enable it and will then need to package libagentcrypt too. I will also try to get it packaged independently of what happens with NeoMutt.
> 
>     * Bundle `agentcrypt` with NeoMutt?
Actually my first implementation was exactly that. But then I realized how useful it could be so I unbundled it as a library.
> 
> 
> Reliability?
> 
>     * Is is secure?
>       It deals with passwords, after all
The code is available for your inspection, the algorithms are documented and are based on libsodium, one of the most secure crypto libraries around. The author of libsodium liked it, apparently: https://github.com/jedisct1/libsodium-doc/pull/84

> 
>     * Is is robust?
>       If a library goes wrong, it'll crash NeoMutt
Look at the source code and judge for yourself. It's written conservatively and checks the output of every function call that can fail. You cannot know this but I've also checked it for memory leaks with valgrind. Of course there could still be bugs but the obvious ones are all gone by now.
> 
>     * Who's going to support the code?
>       (and for how long?)
Me. For as long as I am able to read a screen and type on a keyboard. Pray they never get rid of keyboards in favour of touch screens... 
> 
> 
> What else do we need?
> 
>     * Documentation (once we've sorted out the above)
I have patched the documentation of $imap_pass &co in the code. As a result the neomuttrc.5 man page will already be up to date with the changes. The documentation of the utility to encrypt passwords before storing them in $imap_pass is at https://ndilieto.github.io/libagentcrypt/agentcrypt.1.html and that of the library itself at https://ndilieto.github.io/libagentcrypt/
> 
> 
> An article I read on Open Source Project Management, suggested that you should always merge anything you're offered.
> (You can always sort out any problems later :-)

I am not quite sure about unconditional acceptance. I would first look at the code and check it's a) useful b) well written c) documented properly.

ndilieto:
Hello, 

is there anything I can or should do in order to move this forward?

sstark:
> One could encrypt the whole file where $imap_pass and similar are stored, then decrypt upon running NeoMutt. But this would be ugly and less secure (the decrypted file would need to be around while running). With my solution the decrypted password only lives in RAM while NeoMutt runs.

Maybe I do not understand what you mean. But if you use "source program_that_sets_password|", wouldn't it be the same? And this could be something that uses gpg, your library, or anything else.

It is not clear to me why your method is any better then sourcing the password from an external command like gpg or use backticks. With gpg you can also use agent-forwarding.

gahr:
The basic idea is to decrypt the password on use, not on startup, and to never store the decrypted password in memory. We have discussed this a bit yesterday on IRC, and my (unauthoritative) opinion is that we might want to go with something more generic to express the concept of lazy expansion of variable values.

What if we had something like `set imap_password="get_pass.sh||"` (note the double `|`) that would invoke `get_pass.sh` **each time** `imap_password` is accessed? I'm a lot more in favour of an approach like this, which doesn't hard-code or impose the use of any particular password storage / querying tool.

ndilieto:
> Maybe I do not understand what you mean. But if you use "source program_that_sets_password|", wouldn't it be the same? And this could be something that uses gpg, your library, or anything else.
> 
Possibly but for example how would the external program deal with the situation that arises when no key is available to decrypt the password? If you look at my patch, this gracefully falls back to asking the user for a password. Not sure how it would work with an external program.

> It is not clear to me why your method is any better then sourcing the password from an external command like gpg or use backticks. With gpg you can also use agent-forwarding.

ssh agent forwarding and gpg agent forwarding are two different things. ssh agent forwarding works with ssh keys and it is widely used to do passwordless login into remote boxes. gpg agent forwarding is a relatively new feature that isn't supported in all current distributions and even when supported requires extra setup: https://wiki.gnupg.org/AgentForwarding

ssh agent forwarding is very easily enabled by just adding '-A' to the ssh command line.

gahr:
> The basic idea is to decrypt the password on use, not on startup, and to never store the decrypted password in memory. 

I realize this is not what the PR does. If I read the code correctly, it decrypts on first use, then stores the decrypted password in the variable as usual.

In this case, I see even less value in having it. I guess you could easily wrap the call to ssh-agent and the fallback to asking for a password with e.g. pinentry in a shell script and use the usual `source`. Am I missing anything?

ndilieto:
> I realize this is not what the PR does. If I read the code correctly, it decrypts on first use, then stores the decrypted password in the variable as usual.
>
The caching of the password in memory is neomutt's original code and I left it untouched. I thought of preventing it by not setting the MUTT_ACCT_PASS flag in account->flags but I am not sure about the implications.
 
> In this case, I see even less value in having it. I guess you could easily wrap the call to ssh-agent and the fallback to asking for a password with e.g. pinentry in a shell script and use the usual `source`. Am I missing anything?

pinentry does not work with ssh agent. Again, ssh-agent is NOT gpg-agent.

gahr:
> pinentry does not work with ssh agent. Again, ssh-agent is NOT gpg-agent.

I do understand that. What I say is, you could pretty easily script or produce an executable that does what you propose to add to NeoMutt. A command-line application that requests the key-list from ssh-agent and tries to decrypt the payload passed as an argument with those keys, and falls-back to asking for a password through pinentry if no key was found to decode the payload.

I think your workflow is a valid one, just the implementation should be a separate tool, not hard-coded into NeoMutt.

If you had a binary, let's call it `agc_decrypt`, you could easily
```
set imap_pass="`agc_decrypt encrypted-password`"
```

PR #1745, opened 2019-06-15
config: split out sidebar

flatcap:
This is a small (working) step to splitting up the config.

It moves all the sidebar config defines from `init.h` to `sidebar.c`.
I've added a new function,`init_components()`, that gives sidebar a chance to register its own config.

**Pros**:
- Most of the sidebar's global variables don't need to be `extern`d any more.
- Start to break up the dependency nightmare that is `init.h`

**Cons**:
- As the code stands, the config items aren't sorted in the docs (guide, manual, neomuttrc)

---

**Next Steps**

- Split up more components: imap, notmuch, nntp, etc
- Strip the documentation comments from the code


flatcap:
Not yet.

PR #1757, opened 2019-06-27
build: check which idn2 headers to include

flatcap:
I'm not sure how this managed to get past all the tests I ran, but...
Running `./configure --pkgconf --disable-idn --idn2` means that `HAVE_IDN2_H` isn't defined,
so `address/idna.c` doesn't compile.

### DIFF
--- a/auto.def
+++ b/auto.def
@@ -807,20 +807,20 @@ if {[get-define want-idn] && [get-define want-idn2]} {
   user-error "Cannot specify both --idn and --idn2"
 }
 if {[get-define want-idn]} {
-  if {[get-define want-pkgconf]} {
-    pkgconf true libidn
+  proc find-idn1-includes {} {
     # These are used to figure which header to include
     if {!([cc-check-includes stringprep.h] || [cc-check-includes idn/stringprep.h]) ||
         !([cc-check-includes idna.h] || [cc-check-includes idn/idna.h])} {
       user-error "Unable to find GNU libidn"
     }
+  }
+  if {[get-define want-pkgconf]} {
+    pkgconf true libidn
+    find-idn1-includes
   } else {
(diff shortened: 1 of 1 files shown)


gahr:
Thanks, good catch! I went one step further and refactored the common headers checks.

auto.def
```diff
@@ -807,20 +807,20 @@ if {[get-define want-idn] && [get-define want-idn2]} {
...
+    pkgconf true libidn
+    find-idn-includes
   } else {
     set idn_prefix [opt-val with-idn $prefix]
     cc-with [list -cflags -I$idn_prefix/include -libs -L$idn_prefix/lib] {
-      if {!([cc-check-includes stringprep.h] || [cc-check-includes idn/stringprep.h]) ||
-          !([cc-check-includes idna.h] || [cc-check-includes idn/idna.h])} {
-        user-error "Unable to find GNU libidn"
-      }
+      find-idn-includes
```
flatcap: Is this right?
Here we check for `stringprep.h` and `idna.h` in two places...

auto.def
```diff
@@ -833,14 +833,19 @@ if {[get-define want-idn]} {
...
   if {[get-define want-pkgconf]} {
     pkgconf true libidn2
+    find-idn-includes
   } else {
     set idn_prefix [opt-val with-idn2 $prefix]
     cc-with [list -cflags -I$idn_prefix/include -libs -L$idn_prefix/lib] {
-      if {!([cc-check-includes idn2.h] || [cc-check-includes idn/idn2.h])} {
-        user-error "Unable to find GNU libidn2"
-      }
+      find-idn-includes
```
flatcap: ... but here we're checking for `idn2.h`

PR #1759, opened 2019-06-29
Avoid UB when clearing stdin

fekir:
Flushing `stdin` is undefined behaviour.

References
 http://www.open-std.org/jtc1/sc22/wg14/www/docs/n1570.pdf, chapter 7.21.5.2 (The fflush function)
 http://en.cppreference.com/w/c/io/fflush

* **What does this PR do?**

Replace a call to `fflush` on stdin

No documentation to update

If the function is appropriate, I'll gladly document it

It should

None

### DIFF
--- a/curs_lib.c
+++ b/curs_lib.c
@@ -524,6 +524,19 @@ void mutt_perror_debug(const char *s)
 }
 
 /**
+ * mutt_flush_stdin - remove characters from stdin until '\n' or EOF is encountered
+ */
+void mutt_flush_stdin(void)
+{
+  int c;
+  do 
+  {
+    c = fgetc(stdin);
+  } while ((c != '\n') && (c != EOF));
+}
+
+
+/**
  * mutt_any_key_to_continue - Prompt the user to 'press any key' and wait
  * @param s Message prompt
  * @retval num Key pressed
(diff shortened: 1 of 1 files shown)


curs_lib.c
```diff
@@ -792,6 +792,15 @@ void mutt_flushinp(void)
...
 }
 
+void mutt_flush_stdin(void)
+{
+  int c = fgetc(stdin);
+  while (c != '\n' && c != EOF)
+  {
+    c = fgetc(stdin);
+  }
+}
```
flatcap: How about using a `do ... while()` loop to reduce the duplication:
```c
int c;
do    
{ 
  c = fgetc(stdin);
} while ((c != '\n') && (c != EOF));
```

curs_lib.c
```diff
@@ -792,6 +792,15 @@ void mutt_flushinp(void)
   flushinp();
 }
 
+void mutt_flush_stdin(void)
```
flatcap: Please can you add a Doxygen comment block (something simple).

mutt_curses.h
```diff
@@ -111,6 +111,7 @@ struct KeyEvent
   int op; /**< function op */
 };
 
+void mutt_flush_stdin(void);
```
flatcap: As there's only one use, I think we should drop this and move the function before `mutt_any_key_to_continue()` so that we don't need to forward-declare it.

curs_lib.c
fekir: I do not like that form (actually mine neither), but there is a clear advantage, will change it.

mutt_curses.h
fekir: OK

PR #1760, opened 2019-06-29
Avoid using c++ reserved identifiers

fekir:
`try` is a keyword in c++
While this should not affect C code, it affects unfortunately the
tooling.

This specific instance triggers a "syntax error" in Cppcheck, so it
could hide other problems.

(Cppcheck version  1.86, code parsed with `--language=c --std=c99`)

Of course it is a bug in Cppcheck, but renaming the variable is the most
pragmatic approach, as the bug in Cppcheck does not manifest itself in
a simple program.

* **What does this PR do?**

Remove a reserved keyword from the c++ standard in order to use cppcheck.

Notice:

It might be beneficial for neomutt if a c++ compiler would be able to parse it, as it has better static analysis compared to a c compiler.
Unfortunately there are other reserved keywords, so it would be probably a long-term goal, if there is some interest in it (not the scope of this request or feature-branch).
This is also important for "public" library header files (in case neomutt wants to split some logic in a separate library), as otherwise possible c++ clients wouldn't be able to bind the library.

### DIFF
--- a/address/group.c
+++ b/address/group.c
@@ -180,15 +180,15 @@ static void group_add_addrlist(struct Group *g, const struct AddressList *al)
   if (!g || !al)
     return;
 
-  struct AddressList new = TAILQ_HEAD_INITIALIZER(new);
-  mutt_addrlist_copy(&new, al, false);
-  mutt_addrlist_remove_xrefs(&g->al, &new);
+  struct AddressList al_new = TAILQ_HEAD_INITIALIZER(al_new);
+  mutt_addrlist_copy(&al_new, al, false);
+  mutt_addrlist_remove_xrefs(&g->al, &al_new);
   struct Address *a = NULL, *tmp = NULL;
-  TAILQ_FOREACH_SAFE(a, &new, entries, tmp)
+  TAILQ_FOREACH_SAFE(a, &al_new, entries, tmp)
   {
     mutt_addrlist_append(&g->al, a);
   }
-  assert(TAILQ_EMPTY(&new));
+  assert(TAILQ_EMPTY(&al_new));
 }
 
 /**
--- a/alias.c
+++ b/alias.c
@@ -362,7 +362,6 @@ struct AddressList *mutt_get_address(struct Envelope *env, const char **pfxp)
 void mutt_alias_create(struct Envelope *cur, struct AddressList *al)
 {
   struct Address *addr = NULL;
-  struct Alias *new = NULL;
   char buf[1024], tmp[1024] = { 0 }, prompt[128];
   char *pc = NULL;
   char *err = NULL;
@@ -413,8 +412,8 @@ retry_name:
     }
   }
 
-  new = mutt_alias_new();
-  new->name = mutt_str_strdup(buf);
+  struct Alias *alias = mutt_alias_new();
+  alias->name = mutt_str_strdup(buf);
 
   mutt_addrlist_to_local(al);
 
@@ -429,20 +428,20 @@ retry_name:
   {
     if ((mutt_get_field(_("Address: "), buf, sizeof(buf), 0) != 0) || !buf[0])
     {
-      mutt_alias_free(&new);
+      mutt_alias_free(&alias);
       return;
     }
 
-    mutt_addrlist_parse(&new->addr, buf);
-    if (TAILQ_EMPTY(&new->addr))
+    mutt_addrlist_parse(&alias->addr, buf);
+    if (TAILQ_EMPTY(&alias->addr))
       BEEP();
-    if (mutt_addrlist_to_intl(&new->addr, &err))
+    if (mutt_addrlist_to_intl(&alias->addr, &err))
     {
       mutt_error(_("Bad IDN: '%s'"), err);
       FREE(&err);
       continue;
     }
-  } while (TAILQ_EMPTY(&new->addr));
+  } while (TAILQ_EMPTY(&alias->addr));
 
   if (addr && addr->personal && !mutt_is_mail_list(addr))
     mutt_str_strfcpy(buf, addr->personal, sizeof(buf));
@@ -451,22 +450,22 @@ retry_name:
 
   if (mutt_get_field(_("Personal name: "), buf, sizeof(buf), 0) != 0)
   {
-    mutt_alias_free(&new);
+    mutt_alias_free(&alias);
     return;
   }
-  mutt_str_replace(&TAILQ_FIRST(&new->addr)->personal, buf);
+  mutt_str_replace(&TAILQ_FIRST(&alias->addr)->personal, buf);
 
   buf[0] = '\0';
-  mutt_addrlist_write(buf, sizeof(buf), &new->addr, true);
-  snprintf(prompt, sizeof(prompt), _("[%s = %s] Accept?"), new->name, buf);
+  mutt_addrlist_write(buf, sizeof(buf), &alias->addr, true);
+  snprintf(prompt, sizeof(prompt), _("[%s = %s] Accept?"), alias->name, buf);
   if (mutt_yesorno(prompt, MUTT_YES) != MUTT_YES)
   {
-    mutt_alias_free(&new);
+    mutt_alias_free(&alias);
     return;
   }
 
-  mutt_alias_add_reverse(new);
-  TAILQ_INSERT_TAIL(&Aliases, new, entries);
+  mutt_alias_add_reverse(alias);
+  TAILQ_INSERT_TAIL(&Aliases, alias, entries);
 
   mutt_str_strfcpy(buf, C_AliasFile, sizeof(buf));
   if (mutt_get_field(_("Save to file: "), buf, sizeof(buf), MUTT_FILE) != 0)
@@ -498,14 +497,14 @@ retry_name:
       fputc('\n', fp_alias);
   }
 
-  if (check_alias_name(new->name, NULL, 0))
-    mutt_file_quote_filename(new->name, buf, sizeof(buf));
+  if (check_alias_name(alias->name, NULL, 0))
+    mutt_file_quote_filename(alias->name, buf, sizeof(buf));
   else
-    mutt_str_strfcpy(buf, new->name, sizeof(buf));
+    mutt_str_strfcpy(buf, alias->name, sizeof(buf));
   recode_buf(buf, sizeof(buf));
   fprintf(fp_alias, "alias %s ", buf);
   buf[0] = '\0';
-  mutt_addrlist_write(buf, sizeof(buf), &new->addr, false);
+  mutt_addrlist_write(buf, sizeof(buf), &alias->addr, false);
   recode_buf(buf, sizeof(buf));
   write_safe_address(fp_alias, buf);
   fputc('\n', fp_alias);
--- a/browser.c
+++ b/browser.c
@@ -539,7 +539,7 @@ static const char *folder_format_str(char *buf, size_t buflen, size_t col, int c
 
     case 'N':
       snprintf(fmt, sizeof(fmt), "%%%sc", prec);
-      snprintf(buf, buflen, fmt, folder->ff->new ? 'N' : ' ');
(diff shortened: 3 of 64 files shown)


flatcap:
hehe, that's been bugging me for ages.

I'm not entirely happy with `try_`, but it'll do for now.
(Perhaps we can think of an alternative word)

Do you feel like tackling the others? (There are quite a few of them)

- `new`, `this`, `virtual`, `template`, `catch`, `not`, `public`, `protected`, `private`

fekir:
>Do you feel like tackling the others?

Yeah, I can give a look around.
Did not know if there was interest.

How should I rename those?
Adding a pre/postfix is the easiest way, unless there might be a more meaningful name.

flatcap:
> How should I rename those?

Stick with the postfix `_` for now.
Once you've marked them, it'll be easy to find and rename them later (if we decide to).

fekir:
Ugh, I've just found a member variable `not` (see mutt/regex3.h:61:8:), this is also bad in C (alternate tokens).

I'll stick with the postfix then

flatcap:
btw, ctags is pretty useful for finding these, e.g. find local variables:
`ctags -x --c-kinds=l -R .`

fekir:
> btw, ctags is pretty useful for finding these, e.g. find local variables:
> `ctags -x --c-kinds=l -R .`

oh, did not see your comment.

So, I removed all reserved keywords I could find, and with

````
make CC="g++ -fpermissive"
````

I'm nearly able to compile neomutt.

Considerations:
  * I took the liberty to rename some variables without the prefix, those are in a separate commit
  * `not` and `or` are problematic in `C` too, as the alternate tokens are macros
  * Even if the project does not want to switch to the c++ language, it would not hurt to look at the compiler warnings, for example it diagnostic that there are two different signatures for `comp_path_probe`, a possible issue with the `NONULL` macro, signed/unsigned conversions.


Since I've touched a lot of lines, just check with care and cherry-pick, rebase, adapt whatsoever what's useful, we can eventually rebase the feature branch on the latest master and see whats still missing from my changes.

flatcap:
Wow!  That's a great job.  Thanks.
There are a few missing conversions in the testing code (that's why Travis has failed).
Add `--testing` to `./configure` to enable them.

fekir:
Added `--testing` in my configure and new findings thanks to `ctags`

flatcap:
I've had a quick skim through the changes, and they look good; I'll do a detailed review in a bit.

There's one keyword left (perhaps it's a test :-) in `pattern.c` -- `struct PatternFlags.class`
Also, some of the Doxygen comments don't match the params any more.
Run: `doxygen doxygen/doxygen.conf` to see the errors.

flatcap:
It looks good.
Just one thing: in `hcache/*`, I think I'd prefer if all the function names matched the API, so `hcache_*_delete_header()`

gahr:
> It might be beneficial for neomutt if a c++ compiler would be able to parse it, as it has better static analysis compared to a c compiler.

uhm.. are you talking about a particular C compiler having a better static analysis than another particular C++ compiler, or this is generally true for all implementations? It is my understanding that, e.g., LLVM, does a first pass into IR before doing any analysis, so which front-end language you choose makes little difference.

Also, C++ is not a proper superset of C, so unless we *switch* to C++ I don't see how you would make the NeoMutt source code compilable in --lang=c++ mode.

gahr:
As much as I dislike `try` and `delete` as variable names, I don't necessarily like `try_` and `delete_` any better. I admire the goal, but can we please make a full step instead of a half one and find *proper* names for those variables? Thanks!

fekir:
> Also, some of the Doxygen comments don't match the params any more.
> Run: `doxygen doxygen/doxygen.conf` to see the errors.

I did not pay particular attention as the new names should still get decided.
Otherwise I should update them all a second time.
Good to know there is a way to check those names automatically.

>  Just one thing: in hcache/*, I think I'd prefer if all the function names matched the API, so hcache_*_delete_header()

Ok, is the new name fine?

Hi @gahr, 
> uhm.. are you talking about a particular C compiler having a better static analysis than another particular C++ compiler, or this is generally true for all implementations? It is my understanding that, e.g., LLVM, does a first pass into IR before doing any analysis, so which front-end language you choose makes little difference.
>
> Also, C++ is not a proper superset of C, so unless we switch to C++ I don't see how you would make the NeoMutt source code compilable in --lang=c++ mode.

No, I'm not talking about any compiler in particular.
Maybe my statement was not that clear.
As you said, C++ is not strictly superset of C, it disallows some type of conversion that are generally speaking unsafe. Thus compiling C code as C++, if possible, can help to catch some bugs at compile time.
I do not think if it is standard practice, but some projects did it

 * https://lwn.net/Articles/542457/
 * https://lkml.org/lkml/2018/4/1/116

And I'm sure I read at least another article on this topic, but can't remember the project...

I actually believe that some of the diagnostic with `comp_path_probe`, `NONULL`  `fpermissive` and signed/unsigned  conversions could have been triggered with gcc too (with appropriate flags of course), but maybe not all of them since gcc doesn't want to trigger to many false positives/too many warnings for valid code.

> As much as I dislike try and delete as variable names, I don't necessarily like try_ and delete_ any better. I admire the goal, but can we please make a full step instead of a half one and find proper names for those variables? Thanks!

Well, the new names are worse than the old, a prefixed `_` is hard to read and does not add clarity.
I begun with `try_` just as a one-shot for solving the issue with `cppcheck`, but all names are open to bikeshedding.
The first "important" step was to enlist all variables, did I understand correctly @flatcap?
Now that all places are listed, lets

> [...] check with care and cherry-pick, rebase, adapt whatsoever what's useful [...]

I already replaced some names with more proper ones, but those where just suggestions too.

flatcap:
@fekir

Your renames look good:

- 202b2274 Remove `catch` tokens
  **`catch`** -\> **`restore`**

- 0c59f777 Remove `new` tokens
  **`new`** -\> **`has_new_mail`**

- 8b4cf199 Remove `delete` tokens
  **`delete`** -\> **`delete_original`**

- e63f684b Remove `public` tokens
  **`public`** -\> **`only_public_key`**

- d9063ce0 Remove `new` tokens
  **`old`** -\> **`old_filename`** shorten to **`_file`**?
  **`new`** -\> **`new_filename`** shorten to **`_file`**?

---

Here are some suggestions for the others

- a82c280f Avoid using `try` as variable name
  What the user's typed
  **`str`** or **`user`**

- dff30cef Remove `template` tokens
  Strings like `%L%R`
  **`tmpl`** or **`format`**

- 52ec5b2c Remove `virtual` tokens
  **`virtual`** -\> **`vnum`**

- 472dbc8f Remove `this` tokens
  **`this`** -\> **`ctx_curr`** (current Context)
  **`this`** -\> **`key`** (struct SmimeKey)

- a3749899 Remove `class` tokens
  **`class`** -\> **`qc`** (struct QClass)

- d9063ce0 Remove `new` tokens
  In `rotate_logs()`, you've renamed: **`new`** -\> **`new_file`**
  I suggest renaming **`old`** -\> **`old_file`** for consistency

---

These are less convincing:

- 30b5d58b Remove `not` tokens
- 1acaac71 Remove `or` tokens
  These two are tough.  Maybe **`logic_not`**, **`logic_or`**
  If we change the names, `DT_REGEX_ALLOW_NOT` becomes confusing

- dfdaead9 Remove `new` tokens
  - `struct AddressList new_` -\> **`al`** (standard abbreviation)
  - `struct Alias *new` -\> **`alias`**
  - `struct NntpMboxData.new` -\> **`has_new_mail`**
  - `struct AttachPtr *new` -\> **`ap`**
    There are several of these
  - `struct Parameter *new` -\> **`pnew`**
    There are several of these
  - `struct Body *new_` -\> **`bnew`**
    There are several of these
  - In `insert_message()`
    - **`new`** -\> **`add`**
    - **`newparent`** -\> **`parent`**
  - In `cmd_parse_status()`
    **`short new`** -\> **`bool new_mail`**
  - In `imap_tags_edit()`
    **``new``** -\> **`new_tag`**
  - In `mutt_label_message()`
    **`new`** -\> **`new_label`**
  - `struct MuttThread *new` -\> **`tnew`**
    There are several of these
  - In `mutt_traverse_thread()`
    - **`new`** -\> **`new_email`**
    - **`old`** -\> **`old_email`**
  - In `nntp_active_fetch()`
    **`new`** -\> **`mark_new`**
  - In `remove_filename()`
    - **`new`** -\> **`new_file`**
    - **`old`** -\> **`old_file`**

- 9464b0ad Remove `delete` tokens
  To match the functions to the API prototype:
  - **`hcache_bdb_delete`** -> **`hcache_bdb_delete_header`**
  - **`hcache_gdbm_delete`** -> **`hcache_gdbm_delete_header`**
  - **`hcache_kyotocabinet_delete`** -> **`hcache_kyotocabinet_delete_header`**
  - **`hcache_lmdb_delete`** -> **`hcache_lmdb_delete_header`**
  - **`hcache_qdbm_delete`** -> **`hcache_qdbm_delete_header`**
  - **`hcache_tokyocabinet_delete`** -> **`hcache_tokyocabinet_delete_header`**

flatcap:
> I actually believe that some of the diagnostic with `comp_path_probe`,
> `NONULL`  `fpermissive` and signed/unsigned  conversions could have been
> triggered with gcc too

Currently Travis has gcc v5.4.0 or clang v7.
It builds with `-Werror` so that any warnings cause the build to fail.
I'm happy to add more warning options if we can.

Also, I'd be interested in adding `printf` format string checking for the logging macros, etc.

> The first "important" step was to enlist all variables

Yes.  Now they're marked it's trivial to rename them (once we decide).

fekir:
Thank you for the review.

I'm unsure what's the best way to proceed, since a rebase would destroy all references...

flatcap:
> I'm unsure what's the best way to proceed, since a rebase would destroy all references...

If you just add them add separate commits we can squash when we're happy.

fekir:
Hello, new status:

All commits except those with "Remove xxx" in the title, could be merged (I can eventually make a separate merge request).

The remaining commits are a "class" instance, and "not" and "or", as I did not understood if the proposed tokens (logic_not, logic_or) play well with DT_REGEX_ALLOW_NOT. (and since there where many instances, I wanted to avoid unnecessary work)


The commit "Fix functions declarations/definitions" where "issues" found with `make CC="g++ -fpermissive"`.

I'm not 100% sure if those are real issues in C (if it's UB or not), but surely having declarations/definitions consistent should be the way to go.

flatcap:
@fekir 

I've checked all the 'Replace' commits and they're all good.
Very thorough.  Thanks.

> All commits ... could be merged

I'd prefer them separate.

> (I can eventually make a separate merge request)

No need.

>  a "class" instance

I've renamed that to 'flags' in a local commit which I'll push in the morning.
(I've learnt that pushing to GitHub at 3am is a bad idea :-)

> "Fix functions declarations/definitions" where "issues" found

They're good too.

> but surely having declarations/definitions consistent should be the way to go

Yes, definitely.

> There are other couple of places where `g++` believes there is an error,
> I'll analyse them in separate bug reports/merge requests.

Great.

flatcap:
> fekir wrote:
> and "not" and "or", as I did not understood if the proposed tokens (logic_not, logic_or) play well with DT_REGEX_ALLOW_NOT. (and since there where many instances, I wanted to avoid unnecessary work)

@gahr 
Please can you have a look at 97de10042 and f7c30540b to see if you can think of a reasonable naming strategy.  Thanks.

gahr:
How about using names from logics / set theory?

or -> disjunction
and -> conjunction
not -> negation

gahr:
Or terms related to the context:

97de100 : or -> patt_or, or_pattern, ..
f7c3054 : not -> rex_not, regex_invert, ...

flatcap:
I've rebased, clang-formatted and filled in the `class`, `not` and `or` cases.

fekir:
Thank you!

PR #1766, opened 2019-07-06
Rename typedef to PDR_FLAGS

fekir:
As there is another definition of PatternFlags

* **What does this PR do?**

Rename `typedef uint16_t PatternFlags`, as there is another definition of `PatternFlags`.

Multiple definitions with the same name can be problematic, in this case both types are defined in the same header file.

I must say, I'm not 100% sure this is also UB in the used C standard, but for forward compatibility with newer standard it is beneficial:
https://stackoverflow.com/questions/34986335/does-c-have-one-definition-rule-like-c

I choose to name to mimic `MUTT_PDR_NO_FLAGS`, suggestion are welcome.

### DIFF
--- a/pattern.c
+++ b/pattern.c
@@ -98,7 +98,7 @@ bool C_ThoroughSearch; ///< Config: Decode headers and messages before searching
 
 #define MUTT_MAXRANGE -1
 
-typedef uint16_t PatternFlags;     ///< Flags for parse_date_range(), e.g. #MUTT_PDR_MINUS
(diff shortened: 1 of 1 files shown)


flatcap:
Ah!  This is a good one.  Pure carelessness on my part :-)

I tend to standardise one code feature at a time, across the entire codebase.
I check my work, but I focus on **just** the change I've made.
Plus, if the compiler's happy, I'm happy.
Conclusion: I need a better compiler :-)

Even though the change is local to one file, I've changed the name to **MixedCase** to match the other typedefs.

PR #1768, opened 2019-07-07
Avoid using reserved identifiers

fekir:
Found with `clang -Wreserved-id-macro`


flatcap:
Nice.
I'm slightly conflicted on `queue.h`, though -- it's a 3<sup>rd</sup> party source file.
@gahr and I debated whether we should format/tidy it to match our code.  We decided against it so that any upstream change would be easier to spot.

I suppose the benefit of this _non-code_ change is another compiler flag we can enable.

PR #1769, opened 2019-07-07
add braces to all case blocks

flatcap:
I've used `uncrustify` to add `{}`s to all `case` blocks.
This isn't required by `gcc`/`clang`, but `g++` gives us a lots of warnings.

@fekir 
If this is this helpful, I'll keep it.
(If not, it wasn't much effort to do)

### DIFF
--- a/addrbook.c
+++ b/addrbook.c
@@ -81,26 +81,36 @@ static const char *alias_format_str(char *buf, size_t buflen, size_t col, int co
   switch (op)
   {
     case 'a':
+    {
       mutt_format_s(buf, buflen, prec, alias->name);
       break;
+    }
     case 'f':
+    {
       snprintf(fmt, sizeof(fmt), "%%%ss", prec);
       snprintf(buf, buflen, fmt, alias->del ? "D" : " ");
       break;
+    }
     case 'n':
+    {
       snprintf(fmt, sizeof(fmt), "%%%sd", prec);
       snprintf(buf, buflen, fmt, alias->num + 1);
       break;
+    }
     case 'r':
+    {
       addr[0] = '\0';
       mutt_addrlist_write(addr, sizeof(addr), &alias->addr, true);
       snprintf(fmt, sizeof(fmt), "%%%ss", prec);
       snprintf(buf, buflen, fmt, addr);
       break;
+    }
     case 't':
+    {
       buf[0] = alias->tagged ? '*' : ' ';
       buf[1] = '\0';
       break;
+    }
   }
 
   return src;
@@ -267,6 +277,7 @@ new_aliases:
     {
       case OP_DELETE:
       case OP_UNDELETE:
+      {
         if (menu->tagprefix)
         {
           for (i = 0; i < menu->max; i++)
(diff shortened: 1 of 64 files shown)


fekir:
Hello,

I had to recompile with `g++` to see the warning you ment.

In `C++` you are not required to put braces in the `case` statements, as you are not in `C`.

The issue is that in those statements, some variables are in scope only conditionally:

````
pager.c:3535:12: warning: jump to case label [-fpermissive]                                                                                                                                                          
       case OP_SIDEBAR_TOGGLE_VISIBLE:                                                                                                                                                                         
            ^~~~~~~~~~~~~~~~~~~~~~~~~                                                                                                                                                                                
pager.c:2727:13: note:   crosses initialization of ‘int err’                                                                                                                                                         
         int err = REG_COMP(&rd.search_re, searchbuf, REG_NEWLINE | rflags);                                                                                                                                         
             ^~~                                                                                                                                                                                                     
pager.c:2726:13: note:   crosses initialization of ‘int rflags’                                                                                                                                                      
         int rflags = mutt_mb_is_lower(searchbuf) ? REG_ICASE : 0;                                                                                                                                                   
             ^~~~~~                                                                                                                                                                                                  
````

While apparently in `C` this is not an issue, in `C++` it is.

First, after the label you can read the variable, but if the correct code path was not taken, it will be uninitialized (since there are no warning about it, I suppose this is not really the case)
In `C++` there are destructors. I suppose thus that this code is not permitted fot not overcomplicating the language.

In `C++` you are thus a little more limited with `goto` (and a switch is a `goto`), but those limitations do not exist with other control fow structures like `if`, `for`, ....


To prove my point, here three samples:


Example1) compiles both in C and C++, notice that `int a` exists in scope only if `argc != 0`

````
int main(int argc, char**){
  (void) argv;
  if(argc==0){
    return 0;
  }
  int a = 1;
  return a;
}
````


Example2) is only valid C, even if it should behave exactly as the first example:
Note that after `label`, we could query `a`.
If we did not have `return a;`, we would read from an uninitialized variable, even if declaration and initialisation are togethere.
Which is, IMHO, a flaw.

````
int main(int argc, char** argv){
  (void) argv;
  if(argc==0){
    goto label;
  }
  int a = 1;
  return a;
  
  label:
  return 0;
}
````


Example3) is valid C and C++, and shows that uninitialized variable are not necessarily an issue.

````
int main() {
  int a;
  return 0;
}
````



I do not think that adding braces unconditionally to all braces is necessarily helpful, as (I guess), in most cases they do not add any value.
For those cases where g++ warns, I think it would be a very good idea.
I also tried to compile Example2) with `clang -Weverything`, but no luck, it did not generate any warning.
I had more luck with `gcc -Wjump-misses-init`, so maybe it makes sense to add this flag for gcc and fix only the offending cases?

It would also greatly reduce the scope of the change, `make clean && make CC="gcc -Wjump-misses-init" 2>&1 | grep Wjump-misses-init | wc -l` tells me there are "only" 922 instances of this warning.
And many warnings are surely duplicates, since a `switch` has normally multiple `case`s.

flatcap:
Thanks for the analysis.

> I do not think that adding braces unconditionally to all braces is necessarily helpful

hehe, I thought it'd was too good to be true :-)

> `gcc -Wjump-misses-init`

Ooh, good find.
I'll give that a try and see if I can refine the PR.

flatcap:
>  I'll give that a try and see if I can refine the PR.

Nope.  Never mind.

PR #1772, opened 2019-07-10
Devel/const correctness

fekir:
* **What does this PR do?**

Improve const-correctness.

I've analysed most through `gcc` `-Wcast-qual`, `-Wdiscarded-qualifiers`, `-Wdiscarded-array-qualifiers`, and `g++ -fpermissive`.

All those changes should not change any code behaviour, just reduce the places where to look for UB because of a const casted away.

There are other places where the cast is technically fine, for example functions like `mutt_str_skip_whitespace` should work both with `char*` and with `const char*`. But because of the C type system, and absence of something like templates/generics/overloads, it is not possible 

 * a function that work for both types without a cast/that does not generate a `-Wcast-qual` warning
 * an overload for those types (the most similar things to overload is macro based, and available since  c11: the [generic selection](https://en.cppreference.com/w/c/language/generic))
 * no code duplication (unless replacing the function with a macro)

thus I have no idea how to improve the code in those cases.

Besides, the signature of `iconv` on my system (ubuntu) is `extern size_t iconv (iconv_t __cd, char **__restrict __inbuf, size_t *__restrict __inbytesleft, char **__restrict __outbuf, size_t *__restrict __outbytesleft);`, thus `ICONV_CONST` is defined to nothing.
As this correctly generates `-Wcast-qual` warnings, which I guess are benign, just my function definition not "correct", maybe it makes sense to wrap this function call and to the cast only once.
Or is `iconv` expected in this case to change the content of `inbuf`?

### DIFF
--- a/addrbook.c
+++ b/addrbook.c
@@ -140,8 +140,8 @@ static int alias_tag(struct Menu *menu, int sel, int act)
  */
 static int alias_sort_alias(const void *a, const void *b)
 {
-  struct Alias *pa = *(struct Alias **) a;
-  struct Alias *pb = *(struct Alias **) b;
+  const struct Alias *pa = *(struct Alias const *const *) a;
+  const struct Alias *pb = *(struct Alias const *const *) b;
   int r = mutt_str_strcasecmp(pa->name, pb->name);
 
   return RSORT(r);
@@ -157,8 +157,8 @@ static int alias_sort_alias(const void *a, const void *b)
  */
 static int alias_sort_address(const void *a, const void *b)
 {
-  struct AddressList *pal = &(*(struct Alias **) a)->addr;
-  struct AddressList *pbl = &(*(struct Alias **) b)->addr;
+  const struct AddressList *pal = &(*(struct Alias const *const *) a)->addr;
+  const struct AddressList *pbl = &(*(struct Alias const *const *) b)->addr;
   int r;
 
   if (pal == pbl)
@@ -169,8 +169,8 @@ static int alias_sort_address(const void *a, const void *b)
     r = 1;
   else
   {
-    struct Address *pa = TAILQ_FIRST(pal);
-    struct Address *pb = TAILQ_FIRST(pbl);
+    const struct Address *pa = TAILQ_FIRST(pal);
+    const struct Address *pb = TAILQ_FIRST(pbl);
     if (pa->personal)
     {
       if (pb->personal)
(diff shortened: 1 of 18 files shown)


flatcap:
> Improve const-correctness.

Wow, that's a **lot** of `const`s!

You changes look _odd_, but correct.
(I'm not sure I've actually seen `const *const *` outside of an example in a book :-)

> absence of something like templates/generics/overloads, it is not possible

generics?  That's a new one on me.

> ... I have no idea how to improve the code in those cases.

Hmm...  I see what you mean.

> ... `iconv` ... `ICONV_CONST` is defined to nothing.
> Or is `iconv` expected in this case to change the content of `inbuf`?

It's just conflicting historic definitions.  `iconv` doesn't touch the `inbuf`.

A while ago I wondered about refactoring all the `iconv` calls into one file, hiding away the details, but I couldn't find anyone brave enough to attempt it.

---
Looking through all the warnings, I think I spotted a couple more we can fix:

**curs_lib.c**
```diff
-      addnstr((char *) s, k);
+      addnstr(s, k);
```

**keymap.c**:
```diff
-        char *s = tigetstr((char *) keyname);
+        char *s = tigetstr(keyname);
```

---

Are you still thinking/working on this?

fekir:
> You changes look odd, but correct.
> (I'm not sure I've actually seen const *const * outside of an example in a book :-)

There a reason I try to avoid pointers-to-pointers: I find them very hard to read, especially with some qualifiers!

I think that `Alias const *const *` is easier to read than `const Alias *const *`/`const Alias * const*`/ or other space combinations, as I do not need to read a little from the left and right.
Of course I have no problem chaning them to another convention if it is more readable/appropriate to the common neomutt programmer :-)

> generics? That's a new one on me.

I've meant here the Java generics (https://en.wikipedia.org/wiki/Generics_in_Java), which are completely different from templates in C++, but can be used in similar manner.

> >   ... iconv ... ICONV_CONST is defined to nothing.
> >   Or is iconv expected in this case to change the content of inbuf?
> It's just conflicting historic definitions. iconv doesn't touch the inbuf.

Ok, whondering why on my system I have (believe) old definition.

> Looking through all the warnings, I think I spotted a couple more we can fix:

AFAIK those functions take a char*, so the cast is correct.

flatcap:
> I think that `Alias const *const *` is easier to read

Yes.

Anything more to add to this PR?

fekir:
I do not think so

Issue #1765, opened 2019-07-06
Implicit conversion between bool and QuadOption


fekir:
Hello, I have a question about the correctness/understandability about this (init.c:3343:39, function `enum QuadOption query_quadoption(enum QuadOption opt, const char *prompt)`) line of code:

````
opt = mutt_yesorno(prompt, (opt == MUTT_ASKYES));
````

I assumed `mutt_yesorno` would take as second parameter a `bool`, but it actually takes an `enum QuadOption`.

In the `mutt_yesorno` function, the parameter is evaluated here:

````
mutt_str_asprintf(&answer_string, " ([%s]/%s): ", (def == MUTT_YES) ? yes : no,
                    (def == MUTT_YES) ? no : yes);
````
Thus given

````
enum QuadOption
{
  MUTT_ABORT = -1, ///< User aborted the question (with Ctrl-G)
  MUTT_NO,         ///< User answered 'No', or assume 'No'
  MUTT_YES,        ///< User answered 'Yes', or assume 'Yes'
  MUTT_ASKNO,      ///< Ask the user, defaulting to 'No'
  MUTT_ASKYES,     ///< Ask the user, defaulting to 'Yes'
};
````

if `opt == MUTT_ASKYES` -> converts to `1`, converts to `MUTT_YES` -> prints `yes`
if `opt != MUTT_ASKYES` -> converts to `0`, converts to `MUTT_NO` -> prints `no`

`MUTT_YES` and `MUTT_NO` are handled separately, thus `opt` could be `MUTT_ABORT` too.
I'm unable to see if this scenario is possible,but would it be the correct behaviour?

If not, then a switch case is missing.

Otherwise I would propose to change the code to 

````
opt = mutt_yesorno(prompt, (opt == MUTT_ASKYES) ? MUTT_YES : MUTT_NO);
````

In order not to rely on the specific `enum` values.

The same question holds for `mutt_yesorno`, when called with something different than `MUTT_YES` and `MUTT_NO`.

The documentation states that `* @param def Default answer, see #QuadOption`, but in reality this is true if `def` is only `MUTT_YES` or `MUTT_NO`.
Is it thus a programming error calling the function with something else or the documentation that does not reflect the intended functionality?

flatcap:
Thanks for the detailed analysis.
At times like these, it's good to check the upstream Mutt source.
Yep, there's a bug.  They rely on the fact that a boolean `true` is equivalent to `MUTT_YES`.

[...]

PR #1776, opened 2019-07-14
Avoid implicit conversion between bool and quadopt

fekir:
* **What does this PR do?**

Make the conversion between `bool and `QuadOption` explicit, and avoid
depending on the underlying `enum` value.

Also state that `mutt_yesorno` works correctly for `MUTT_YES` and `MUTT_NO`

https://github.com/neomutt/neomutt/issues/1765

* **Further consideration**

I've added an `assert` to ensure code-correctness/self-documentation, but I've noticed that they are not used so much in the code base.
Are there some guidelines? Are they accepted/avoided?
On https://neomutt.org/dev/coding-style they are not mentioned.

### DIFF
--- a/curs_lib.c
+++ b/curs_lib.c
@@ -326,7 +326,7 @@ void mutt_edit_file(const char *editor, const char *file)
 /**
  * mutt_yesorno - Ask the user a Yes/No question
  * @param msg Prompt
- * @param def Default answer, see #QuadOption
+ * @param def Default answer, #MUTT_YES or #MUTT_NO (see #QuadOption)
  * @retval num Selection made, see #QuadOption
  */
 enum QuadOption mutt_yesorno(const char *msg, enum QuadOption def)
--- a/init.c
+++ b/init.c
@@ -3327,7 +3327,7 @@ enum QuadOption query_quadoption(enum QuadOption opt, const char *prompt)
       return opt;
 
     default:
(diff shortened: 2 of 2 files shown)


flatcap:
> I've added an `assert`
> On 'coding-style' they are not mentioned.

I tried to keep that doc as short as I could.
The shorter it is, the more likely people are to read it :-)

It's probably time to update that doc.

> Are they accepted/avoided?

I tend favour logging over assert, since assert will stop the program.
(Unless there's some critical reason why we can't go on)

In this code, I don't think we _really_ need to log or assert.

fekir:
> In this code, I don't think we really need to log or assert.

IMHO:

I do not think either that it should be logged or checked at runtime, since it would be a programming error and not a runtime error.

What's really the expected behaviour of the function if the input value is invalid/if the precondition is not held?
I think the `assert` is legitimate in his case, as in production it will work as it did all the time, but when developing it would help to catch those logical errors.

Stopping the program ensures that there is no code-path with the invalid value (if properly tested), and can help people less accustomed with the code spotting dead branches or invalid statements (as you do not want the assertion to fail), while simply logging does not decrease the complexity of the program.

Given that, I do not have any problems removing it, that's why I asked.

fekir:
Removed the `assert`

flatcap:
> Removed the `assert`

Thanks

> (I can still put it back if I've convinced you)

Keep pushing me; I'll come around :-)

PR #1780, opened 2019-07-19
Refactor regex with custom wrapper

psyomn:
* **What does this PR do?**

Minor refactor of the use of `regexec` with our own very `regex_is_match`/`regex_is_match_full`.

I'm still in the process of verifying a lot of this, but I'm creating a in spirit WIP PR so I kick myself into finishing this (sorry for the long wait). 

With prior discussions of testing, I've decided to take an unorthodox venue in testing as you may see, which will test old uses of `regexec` against its corresponding `regexec_is_match`, etc. 

Also creating this PR in the meantime so I can know if I'm doing something super wrong, or not in the project's spirit (so I hope that's OK).

N/A

   - [x] Documentation created/updated (you have to edit
     for that)

   - [x] All builds and tests are passing

   - [x] Added [doxygen code documentation](https://neomutt.org/dev/doxygen)

   - [x] Code follows the [style guide](https://neomutt.org/dev/coding-style)

None that I know of.

### DIFF
--- a/browser.c
+++ b/browser.c
@@ -707,8 +707,7 @@ static int examine_directory(struct Menu *menu, struct BrowserState *state,
         continue;
       if (prefix && *prefix && !mutt_str_startswith(mdata->group, prefix, CASE_MATCH))
         continue;
-      if (C_Mask && C_Mask->regex &&
-          !((regexec(C_Mask->regex, mdata->group, 0, NULL, 0) == 0) ^ C_Mask->pat_not))
+      if (!mutt_regex_match(C_Mask, mdata->group))
       {
         continue;
       }
@@ -767,8 +766,7 @@ static int examine_directory(struct Menu *menu, struct BrowserState *state,
       {
         continue;
       }
-      if (C_Mask && C_Mask->regex &&
-          !((regexec(C_Mask->regex, de->d_name, 0, NULL, 0) == 0) ^ C_Mask->pat_not))
+      if (!mutt_regex_match(C_Mask, de->d_name))
       {
         continue;
       }
--- a/test/Makefile.autosetup
+++ b/test/Makefile.autosetup
@@ -323,6 +323,7 @@ PATTERN_OBJS	= test/pattern/comp.o \
 
 REGEX_OBJS	= test/regex/mutt_regex_compile.o \
 		  test/regex/mutt_regex_free.o \
+		  test/regex/mutt_regex_match.o \
 		  test/regex/mutt_regexlist_add.o \
 		  test/regex/mutt_regexlist_free.o \
 		  test/regex/mutt_regexlist_match.o \
--- a/commands.c
+++ b/commands.c
@@ -150,7 +150,7 @@ static void update_protected_headers(struct Email *e)
 
     mutt_str_replace(&e->env->subject, prot_headers->subject);
     FREE(&e->env->disp_subj);
-    if (regexec(C_ReplyRegex->regex, e->env->subject, 1, pmatch, 0) == 0)
+    if (mutt_regex_capture(C_ReplyRegex, e->env->subject, 1, pmatch))
       e->env->real_subj = e->env->subject + pmatch[0].rm_eo;
     else
       e->env->real_subj = e->env->subject;
--- a/email/parse.c
+++ b/email/parse.c
@@ -1135,8 +1135,7 @@ struct Envelope *mutt_rfc822_read_header(FILE *fp, struct Email *e, bool user_hd
     {
       regmatch_t pmatch[1];
 
-      if (C_ReplyRegex && C_ReplyRegex->regex &&
-          (regexec(C_ReplyRegex->regex, env->subject, 1, pmatch, 0) == 0))
+      if (mutt_regex_capture(C_ReplyRegex, env->subject, 1, pmatch))
       {
         env->real_subj = env->subject + pmatch[0].rm_eo;
       }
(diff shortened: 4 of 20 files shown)


flatcap:
Wow!  That looks **much** tidier.
I'll have a close look through it tomorrow.
Thanks.

browser.c
```diff
@@ -702,8 +702,7 @@ static int examine_directory(struct Menu *menu, struct BrowserState *state,
         continue;
       if (prefix && *prefix && !mutt_str_startswith(mdata->group, prefix, CASE_MATCH))
         continue;
-      if (C_Mask && C_Mask->regex &&
-          !((regexec(C_Mask->regex, mdata->group, 0, NULL, 0) == 0) ^ C_Mask->pat_not))
+      if (mutt_regex_is_match(C_Mask, mdata->group))
```
psyomn: I've double and triple checked this a few times, but I would really appreciate another pair of eyes here (I've also got a test to make sure that this is indeed the right behavior, but still paranoid).

test/Makefile.autosetup
```diff
@@ -467,6 +469,7 @@ TEST_OBJS	= test/main.o \
 		  $(MBYTE_OBJS) \
 		  $(MD5_OBJS) \
 		  $(MEMORY_OBJS) \
+		  $(MISC_OBJS) \
```
psyomn: I am merely suggesting this, but I can get rid of it if you don't think this is necessary.

psyomn:
Thanks a lot for the feedback! I hope my corrections are OK!

browser.c
gahr: In the old code:

* `(regexec(...) == 0)` --> a match
* `((regexec(...) == 0) ^ C_Mask->pat_not)` --> either a match or a negation flag (but not both)
* `!((regexec(...) == 0) ^ C_Mask->pat_not)` --> not the above

So I think you need an additional negation around `mutt_regex_is_match`

browser.c
```diff
@@ -762,8 +761,7 @@ static int examine_directory(struct Menu *menu, struct BrowserState *state,
       {
         continue;
       }
-      if (C_Mask && C_Mask->regex &&
-          !((regexec(C_Mask->regex, de->d_name, 0, NULL, 0) == 0) ^ C_Mask->pat_not))
+      if (mutt_regex_is_match(C_Mask, de->d_name))
```
gahr: Same as above

browser.c
psyomn: Thanks for checking my commit! I knew I was doing something wrong :frowning_face: 

I've added a test to check exactly this, now.

psyomn:
I applied the required feedback! Hope this is good!

gahr:
I think I would perhaps drop the `_is` from the function names,  for consistency with the existing `mutt_regexlist_match` and `mutt_replacelist_match`.

psyomn:
I can do that tonight, will push a commit later.

flatcap:
rebased and squashed

gahr:
Oh, I take that back. I think the `flags` parameter to `mutt_regex_match_full` must go away. Those are passed to `regexec`, as opposed to the DT_REGEX flags which are used in `regcomp`. We currently have no use for `regexec` flags, so I would simply remove the parameter.

psyomn:
Thanks, nice catch. I'll push another commit tonight. 

Is it cool if I force push my changes this time around?

flatcap:
> Is it cool if I force push my changes this time around?

Yes, that's fine.

psyomn:
So, sorry if I have misunderstood, but I think I missed a few regexec uses: 

https://github.com/neomutt/neomutt/blob/master/pager.c#L1131-L1133

https://github.com/neomutt/neomutt/blob/master/pager.c#L1209-L1210

https://github.com/neomutt/neomutt/blob/master/pager.c#L1758-L1759

These actually do in fact use some of the flags passed in through the `mutt_regex_match_full` method at the moment. However these are the only few occurences. 

I believe a good course of action is to: 
- clean up the rest of `regexec` uses I missed with the new wrapper
- add another function with the possibility to pass the flags would be advisable

Does this sound sane? Or did I not understand something?

flatcap:
>  think I missed a few regexec uses

Oops.  I didn't think to look :-)

> clean up the rest of `regexec` uses I missed with the new wrapper

Great.  Please can you put the new changes in a separate commit so we can review them more easily.

> add another function with the possibility to pass the flags would be advisable

Hmm...  For now, leave the 'custom' uses of `regexec` in `pager.c`.

As long as we've got all our 'struct Regex` variables wrapped up, that'll be good.

Thanks.

psyomn:
Alright, so I did the above changes, and took your feedback and applied it too! I force pushed, and will not force push from this point on until we speak again, if there are issues. Thanks for your patience, and was pretty fun working on this! Let me know if other things are amiss. 

Another point: I've noticed not all parts use the wrapped up regex. In the future, is this a venue we'd like to take? If it is so, perhaps I could continue working on such a refactor in the future (if that's cool)!

gahr:
Thanks @psyomn , I agree with @flatcap that we should keep the few uses of regexec w/ flags alone for the time being. Didn't we decide on `mutt_regex_capture` for the instead of `mutt_regex_full`?

flatcap:
Thanks, the latest changes all look good.
The number of `regexec()` calls has dropped from 45 to 24.

> Thanks for your patience

You're welcome.  We're happy to help people help us :-)

psyomn:
> Didn't we decide on mutt_regex_capture instead of mutt_regex_full?

Yep, had limited time yesterday so I'll probably do this tonight in a separate commit. (We can squash again all commits to one commit once done). Thanks!

flatcap:
Thanks @psyomn.

I've just created #1796 for the remaining regex uses (if you're interested).

psyomn:
Very interested! Thanks so much for the guidance. I'll start looking at that tonight!

PR #1787, opened 2019-07-24
Fix struct Buffer used as %s

l2dy:
* **What does this PR do?**
`cmd` is a `Buffer` struct, it does not correspond to `%s` in the format string.

     for that)


gahr:
I fear that's not the only place where Buffer is assumed to start with a char* :(

flatcap:
> I fear that's not the only place ... :(

I've found an fixed another 14 similar cases in be4f2675e

Because of some of the macro magic in the logging code, I can't get `__attribute__ ((format (printf, 2, 3)))` to work.
As a workaround, I temporarily created dummy logging functions that had the proper prototype.

PR #1797, opened 2019-08-03
fix(notmuch): ensure message_id is correct

### LINKED ISSUE #1793: `notmuch new` triggers "Mailbox was externally modified" warning in neomutt, even when no new mail
I've just migrated from mutt to neomutt + notmuch, and there's one problem that I can't solve.

When I run `notmuch new`, neomutt *always* issues the following warning

```
Mailbox was externally modified.  Flags may be wrong.
```

even when `notmuch new` doesn't actually (seem to) modify the mailbox:

```bash
$ notmuch new
No new mail.
```

Notably, the same warning does *not* occur in regular mutt, so I guess it's a warning about the *virtual* mailbox that I'm viewing in neomutt.

It's particularly annoying because it also issues an audible beep + visual bell, which I use to notify me of new mail.

Is this a bug, or have I configured something wrong? It may be related to #170, but I can't tell.

(FWIW, I'm only using notmuch to create a single virtual mailbox/folder that combines two regular mailboxes/folders. Nothing too fancy.)

Austin-Ray:
The notmuch-backend initializes a provided email struct with content
from notmuch. In rare instances, a provided email struct has a
message_id that does not match the one from notmuch. No validation to
ensure message_ids match is performed.

An issue occurs when running nm_mbox_check(), which stores all messages
in a hash table. The incorrect message_id is the key so our email is
considered missing, reloaded into the mailbox, and a duplicate appears
in the index. This triggers a warning from neomutt regarding external
modification.

Instead, validate the message_ids to ensure a match so our hash table
keys are correct. As a result, a reload is no longer performed and
neomutt's warning need not occur.

Fixes #1793

### DIFF
--- a/notmuch/mutt_notmuch.c
+++ b/notmuch/mutt_notmuch.c
@@ -778,8 +778,20 @@ static int init_email(struct Email *e, const char *path, notmuch_message_t *msg)
 
   mutt_debug(LL_DEBUG2, "nm: [e=%p, edata=%p] (%s)\n", (void *) e, (void *) e->edata, id);
 
+  char *nm_msg_id = nm2mutt_message_id(id);
   if (!e->env->message_id)
-    e->env->message_id = nm2mutt_message_id(id);
+  {
+    e->env->message_id = nm_msg_id;
+  }
+  else if (mutt_str_strcmp(e->env->message_id, nm_msg_id) != 0)
+  {
+    FREE(&e->env->message_id);
+    e->env->message_id = nm_msg_id;
+  }
+  else
+  {
+    FREE(&nm_msg_id);
+  }
 
   if (update_message_path(e, path) != 0)
     return -1;


notmuch/mutt_notmuch.c
```diff
@@ -778,8 +778,10 @@ static int init_email(struct Email *e, const char *path, notmuch_message_t *msg)
 
   mutt_debug(LL_DEBUG2, "nm: [e=%p, edata=%p] (%s)\n", (void *) e, (void *) e->edata, id);
 
-  if (!e->env->message_id)
-    e->env->message_id = nm2mutt_message_id(id);
+  char *nm_msg_id = nm2mutt_message_id(id);
+  if (!e->env->message_id || (mutt_str_strcmp(e->env->message_id, nm_msg_id) != 0))
+    e->env->message_id = nm_msg_id;
+
```
flatcap: We need to free the `message_id` in the case where it exists, but differs from Notmuch's.

```c
  char *nm_msg_id = nm2mutt_message_id(id);
  if (mutt_str_strcmp(e->env->message_id, nm_msg_id) != 0)
  { 
    FREE(&e->env->message_id);
    e->env->message_id = nm_msg_id;
  }
```

but now I'm worried about `Mailbox->id_hash` (the Mailbox's hash of Message-Ids).
If we pass the `Mailbox` into `init_email()`, we can:

```c
  mutt_hash_delete(m->id_hash, old_message_id, e);  
  mutt_hash_insert(m->id_hash, new_message_id, e);  
```

giving us:

```c
  char *nm_msg_id = nm2mutt_message_id(id);
  if (!e->env->message_id)
  { 
    e->env->message_id = nm_msg_id;
    mutt_hash_insert(m->id_hash, e->env->message_id, e);
  }  
  else if (mutt_str_strcmp(e->env->message_id, nm_msg_id) != 0)
  {
    mutt_hash_delete(m->id_hash, e->env->message_id, e);
    FREE(&e->env->message_id);
    e->env->message_id = nm_msg_id;
    mutt_hash_insert(m->id_hash, e->env->message_id, e);
  }
```

Austin-Ray: I'm not sure the insertions/deletions are necessary.

The hash table is first initialized in `get_mutt_email()`, which is called from `nm_mbox_check()`. `nm_mbox_check()` occurs after the `init_email()` calls. At that point, the hash table is accurate.

If we "delete" an  email by untagging it through `notmuch`'s cli, we receive the "Mailbox was externally modified. Flags may be wrong." message, which is reasonable because it was externally modified. The user should reload their mailbox for accurate information.

If we "add" an email by tagging it through `notmuch`'s cli, we don't receive any message. New messages just appear, which is incorrect behavior. We should warn the user that they should reload.

If we do proceed with inserting/deleting, we will need to initialize the hash table in `init_email()`.

flatcap: OK, fair enough.
Then we just need the `FREE()` for the replacement case.

Austin-Ray: I've added a `FREE()` for the replacement case and another for when we don't use `nm_msg_id` since `nm2mutt_message_id()` has a malloc call.

PR #1806, opened 2019-08-18
Increase length of attachment pipe command

alyssais:
This matches the length of the pipe command for messages, and is much less likely to be hit accidentally.


flatcap:
Good catch, @alyssais.
In fact, I've made it bigger still.
Wherever we use paths or commands, we use buffers of `PATH_MAX` bytes.

PR #1814, opened 2019-08-24
fix: alias expansion in compose

flatcap:
This fixes two bugs when expanding aliases, of the form "alias1, alias1" in compose.

- only the first alias was expanded
- alias address was leaked

This bug was introduced when we converted the `Address` to use `TAILQ`.

---

### Testing

The problem can be seen, and the fix tested, using the config file below.
Run:
- `./neomutt -n -f /dev/null -F alias.rc`

Then use <kbd>F1</kbd> .. <kbd>F10</kbd> to compose emails.
Each should result in fully-qualified addresses to `@example.com`, with zero duplicates.

**alias.rc**
```sh
set abort_unmodified = no
set editor           = cat
set postpone         = no
set record           = ""

alias person \
	Adam Apple <apple@example.com>

alias people \
	Barry Banana <banana@example.com>, \
	Cat Cherry <cherry@exmaple.com>, \
	Doug Damson <damson@example.com>

alias nested \
	Eric Endive <endive@example.com>, \
	person, \
	Freda Fig <fig@example.com>

alias deeper \
	nested

macro index <F1> "mGreta Grape <grape@example.com><enter>test<enter>"
macro index <F2> "mperson<enter>test<enter>"
macro index <F3> "mpeople<enter>test<enter>"
macro index <F4> "mnested<enter>test<enter>"
macro index <F5> "mdeeper<enter>test<enter>"

macro index <F6> "mGreta Grape <grape@example.com>, Greta Grape <grape@example.com><enter>test<enter>"
macro index <F7> "mperson, person<enter>test<enter>"
macro index <F8> "mpeople, people<enter>test<enter>"
macro index <F9> "mnested, nested<enter>test<enter>"
macro index <F10> "mdeeper, nested, nested, deeper<enter>test<enter>"

# vim: syn=neomuttrc
```

### DIFF
--- a/alias.c
+++ b/alias.c
@@ -60,9 +60,6 @@
  */
 static void expand_aliases_r(struct AddressList *al, struct ListHead *expn)
 {
-  bool i;
-  const char *fqdn = NULL;
-
   struct Address *a = TAILQ_FIRST(al);
   while (a)
   {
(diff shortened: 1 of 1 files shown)


alias.c
```diff
@@ -83,20 +83,29 @@ static void expand_aliases_r(struct AddressList *al, struct ListHead *expn)
           }
         }
 
-        if (!i)
+        if (i)
```
Austin-Ray: Can we rename `i` to be more descriptive of it's purpose? I think `duplicate` makes the conditional easier to understand.
```C
if (duplicate)
{
            // We've already seen this alias, so drop it
          struct Address *next = TAILQ_NEXT(a, entries);
          TAILQ_REMOVE(al, a, entries);
          mutt_addr_free(&a);
          a = next;
          continue;
}
```
vs.
```C
if (i)
{
            // We've already seen this alias, so drop it
          struct Address *next = TAILQ_NEXT(a, entries);
          TAILQ_REMOVE(al, a, entries);
          mutt_addr_free(&a);
          a = next;
          continue;
}
```

alias.c
```diff
@@ -83,20 +83,29 @@ static void expand_aliases_r(struct AddressList *al, struct ListHead *expn)
...
+          TAILQ_REMOVE(al, a, entries);
+          mutt_addr_free(&a);
+          a = next;
+          continue;
+        }
+
+        mutt_list_insert_head(expn, mutt_str_strdup(a->mailbox));
+        struct AddressList copy = TAILQ_HEAD_INITIALIZER(copy);
+        mutt_addrlist_copy(&copy, alias, false);
+        expand_aliases_r(&copy, expn);
```
Austin-Ray: Can you add a comment about what's happening here?

From what I can put together, this is the recursive expansion of aliases to ensure nested aliases work. Is that correct?

flatcap:
> Let's add some unit tests to catch any future regressions :)

Unfortunately, that's a bit trickier.

The function relies on quite a few global and Config variables.
It also does expansion of local usernames.

Austin-Ray:
On 08/24, Richard Russon wrote:
> The function relies on quite a few global and Config variables.
> It also does expansion of local usernames.

That complicates matters. We should strive for testability to keep our
future sanity intact.

I see two approaches. Let me know if they are possible at this point:

 1. Refactor body of expand_aliases_r() into low-level method that takes
    in global/config values as parameters. Use expand_aliases_r() as
    high-level method to call new lower-level method. Something like:

```C
static void expand_aliases_r(...)
{
  _expand_aliases_r(..., C_UseDomain);
}

static void _expand_aliases_r(..., bool use_domain)
{
...
}
```

 2. Refactor parts of expand_aliases_r() into their own methods and unit
    test them.

flatcap:
Ultimately, we want a list of addresses containing `@`.
I can't see a reason why we can't do this in two stages (plus a de-dupe).

1. `expand_aliases()` -- testable
  Pass in the `Aliases` global

2. `expand_local_users()` -- not easily tested
  This relies on `/etc/passwd`, `C_UseDomain`, `C_Hostname` and `C_HiddenHost`

3. `addrlist_dedupe()` -- already tested

<sub>... but I don't have time to do it myself</sub>

gahr:
F1 through F5 work fine with current master. Are those examples of leaks, rather than wrong expansion?

flatcap:
> F1 through F5 work fine with current master.
> Are those examples of leaks, rather than wrong expansion?

Yes, sorry.  That's what drew me to the code in the first place.

PR #1817, opened 2019-08-27
Tidy some Buffer API, keep most buffers on the stack

gahr:
* **What does this PR do?**

Tidy (in my opinion) some buffer and mailbox API. Keep most buffers on the stack, thus halving the number of allocations per buffer usage (before: 1 allocation for the buffer struct, 1 for the char array).


flatcap:
Rebased, tidied and a couple of fixes.

flatcap:
rebased; squashed down to its essential commits.

PR #1823, opened 2019-08-30
Add mutt_date_epoch_ms API, convert gettimeofdays to use the new API

gahr:
Also, convert nanosleeps into usleep, which has a simpler API.

Edit: I haven't touched `progress.c`, which will come in #1822

### DIFF
--- a/autocrypt/autocrypt.c
+++ b/autocrypt/autocrypt.c
@@ -30,7 +30,6 @@
 #include <errno.h>
 #include <string.h>
 #include <sys/stat.h>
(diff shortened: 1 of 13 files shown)


flatcap:
> Also, convert nanosleeps into usleep, which has a simpler API.

Unfortunately, `usleep()` [is deprecated](http://pubs.opengroup.org/onlinepubs/009695399/functions/usleep.html).
See the 'Application Usage' section.

> Applications are recommended to use nanosleep()

Issue #1802, opened 2019-08-12
Ability to toggle number of levels to hide with toggle-quoted


darrenkenny:
## What would you like to be able to do?

Toggle quoted is very useful for long e-mails, but without the context of the lines above
when people response in-line, it becomes pretty useless.

## What needs to change in NeoMutt?

It would be great if it was possible to specify the level of quotes to be hidden, or a number
of lines of context to be maintained when calling `<toggle-quoted>`.

## Have you tried any alternative methods?

I have tried setting the quote_regex to skip the first level, so something like:

```
set quote_regex="^[|>:}#][ \t]?([ \t]*[|>:}#])+"
```

And that gets me some of the way there, but then I lose the colouring of the quote levels,
and anything else that uses the quote level.

flatcap:
Yes, that sounds pretty useful.

It could be a little tricky to implement -- the pager code is already quite complicated.

darrenkenny:
I just took a quick scan through pager.c, and it seems that, with the right code to get a config value, etc. to obtain a value for `toggle_quote_from_level` that it might be as simple
as the change:

```diff
diff --git a/pager.c b/pager.c
index 76535a9232d3..0d59e7b27d3e 100644
--- a/pager.c
+++ b/pager.c
@@ -1711,7 +1711,8 @@ static int display_line(FILE *fp, LOFF_T *last_pos, struct Line **line_info,
     }

     /* this also prevents searching through the hidden lines */
-    if ((flags & MUTT_HIDE) && ((*line_info)[n].type == MT_COLOR_QUOTED))
+    if ((flags & MUTT_HIDE) && ((*line_info)[n].type == MT_COLOR_QUOTED) &&
+       ((*line_info)[n].quoted->index >= toggle_quote_from_level )
       flags = 0; /* MUTT_NOSHOW */
   }

```

Does that look right? If so, if I can get the cycles, I'll take a stab at implementing it.

Any hints on how to create such a config variable, etc?

flatcap:
Oh!  I revise my estimate to: "not as tricky as I thought" :-)

> Does that look right?

Yes, that looks reasonable.

> Any hints on how to create such a config variable?

That's easy enough...
- Define a new Config Item in `init.h`, of type `DT_NUMBER`
- Add a new global variable to `globals.h`
  - type `short`
  - name `C_` + config name converted to MixedCase

> I'll take a stab at implementing it

Thank you very much.
Any questions, just ask.

[...]

PR #1825, opened 2019-09-02
Add ability to limit levels hidden with toggle-quoted command

darrenkenny:
By setting the configuration variable `toggle_quoted_show_levels`, when you
execute the `toggle-quoted` command, the specified number of levels
will continue to be visible.

Closes: #1802

### DIFF
--- a/pager.c
+++ b/pager.c
@@ -1712,8 +1712,11 @@ static int display_line(FILE *fp, LOFF_T *last_pos, struct Line **line_info,
     }
 
     /* this also prevents searching through the hidden lines */
-    if ((flags & MUTT_HIDE) && ((*line_info)[n].type == MT_COLOR_QUOTED))
+    if ((flags & MUTT_HIDE) && ((*line_info)[n].type == MT_COLOR_QUOTED) &&
+        (((*line_info)[n].quote == NULL) ||
+         ((*line_info)[n].quote->index >= C_ToggleQuotedShowLevels))) {
       flags = 0; /* MUTT_NOSHOW */
+    }
   }
 
   /* At this point, (*line_info[n]).quote may still be undefined. We
--- a/pager.c
+++ b/pager.c
@@ -1669,34 +1669,36 @@ static int display_line(FILE *fp, LOFF_T *last_pos, struct Line **line_info,
     }
   }
 
+  struct Line * const curr_line = &(*line_info)[n];
+
   if (flags & MUTT_PAGER_LOGS)
   {
     /* determine the line class */
-    if (fill_buffer(fp, last_pos, (*line_info)[n].offset, &buf, &fmt, &buflen, &buf_ready) < 0)
+    if (fill_buffer(fp, last_pos, curr_line->offset, &buf, &fmt, &buflen, &buf_ready) < 0)
     {
       if (change_last)
         (*last)--;
       goto out;
     }
 
-    (*line_info)[n].type = MT_COLOR_MESSAGE_LOG;
+    curr_line->type = MT_COLOR_MESSAGE_LOG;
     if (buf[11] == 'M')
-      (*line_info)[n].syntax[0].color = MT_COLOR_MESSAGE;
+      curr_line->syntax[0].color = MT_COLOR_MESSAGE;
     else if (buf[11] == 'W')
-      (*line_info)[n].syntax[0].color = MT_COLOR_WARNING;
+      curr_line->syntax[0].color = MT_COLOR_WARNING;
     else if (buf[11] == 'E')
-      (*line_info)[n].syntax[0].color = MT_COLOR_ERROR;
+      curr_line->syntax[0].color = MT_COLOR_ERROR;
     else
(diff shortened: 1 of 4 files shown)


flatcap:
Thanks for the PR, @darrenkenny 

May I ask a favour:  Please can you tweak the commit message to `Closes: #1802`
That allows GitHub to [perform some magic](https://help.github.com/en/articles/closing-issues-using-keywords), when this PR is closed.

Something like this should do it:
```sh
git commit --amend
git push --force origin issue-1802
```

darrenkenny:
Done

pager.c
```diff
@@ -1710,8 +1710,10 @@ static int display_line(FILE *fp, LOFF_T *last_pos, struct Line **line_info,
     }
 
     /* this also prevents searching through the hidden lines */
-    if ((flags & MUTT_HIDE) && ((*line_info)[n].type == MT_COLOR_QUOTED))
+    if ((flags & MUTT_HIDE) && ((*line_info)[n].type == MT_COLOR_QUOTED) &&
+        ((*line_info)[n].quote->index >= C_ToggleQuotedShowLevels)) {
```
gahr: `(*line_info)[n].quote` is `NULL` for non-quoted lines. That triggered a crash as soon as I run-tested this branch ;)

darrenkenny: Curious, I never saw any crashes, but OK, I can see it could happen depending on some variables set.

On the presumption that the value of quote and quote->index never mattered before, I'm
suggesting the following to behave as before if quote is NULL, and only evaluate 
the index if it is not NULL.

Does that sound OK? 

If so, how about this change:

```diff
diff --git a/pager.c b/pager.c
index 3881efd9326c..b1377725c317 100644
--- a/pager.c
+++ b/pager.c
@@ -1711,7 +1711,8 @@ static int display_line(FILE *fp, LOFF_T *last_pos, struct Line **line_info,

     /* this also prevents searching through the hidden lines */
     if ((flags & MUTT_HIDE) && ((*line_info)[n].type == MT_COLOR_QUOTED) &&
-        ((*line_info)[n].quote->index >= C_ToggleQuotedShowLevels)) {
+        (((*line_info)[n].quote == NULL) ||
+         ((*line_info)[n].quote->index >= C_ToggleQuotedShowLevels))) {
       flags = 0; /* MUTT_NOSHOW */
     }
   }
```

pager.c
```diff
@@ -1710,8 +1710,11 @@ static int display_line(FILE *fp, LOFF_T *last_pos, struct Line **line_info,
     }
 
     /* this also prevents searching through the hidden lines */
-    if ((flags & MUTT_HIDE) && ((*line_info)[n].type == MT_COLOR_QUOTED))
+    if ((flags & MUTT_HIDE) && ((*line_info)[n].type == MT_COLOR_QUOTED) &&
+        (((*line_info)[n].quote == NULL) ||
+         ((*line_info)[n].quote->index >= C_ToggleQuotedShowLevels))) {
```
Austin-Ray: I think this conditional is becoming too crowded and hampering readability. We should refactor it so it's easier to understand.

Something like this:
```C
struct Line cur_line_info = (*line_info)[n]; // This is probably not the right type, just using it as an example

if ((flags & MUTT_HIDE) && (cur_line_info.type == MT_COLOR_QUOTED) &&
        ((cur_line_info.quote == NULL) ||
         (cur_line_info.quote->index >= C_ToggleQuotedShowLevels)))
```
or
```C
bool should_hide(Line *line) // Probably need a different name
{
    return (line.quote == NULL) || (line.quote->index >= C_ToggleQuotedShowLevels);
}

static int display_line(...)
{
    ...
    struct Line cur_line_info = (*line_info)[n]; // This is probably not the right type, just using it as an example

    if ((flags & MUTT_HIDE) && (cur_line_info.type == MT_COLOR_QUOTED) && should_hide(cur_line_info))
    {
    ...
    }
    ...
}
```

Just to alleviate the mental burden when reading the conditional.

gahr:
> `struct Line cur_line_info = (*line_info)[n]; // This is probably not the right type, just using it as an example`

I was about to ask to factor out the verbose `(*line_info)[n]` into a local variable too, then I thought that might be best done in a separate clean up stage. Of course if @darrenkenny wants to go for it, cool :)

darrenkenny:
I think it would be best done as a separate patch, but if you think it is doable here, as an additional commit probably, here's my first attempt, attached.

[cur_line_info.patch.txt](https://github.com/neomutt/neomutt/files/3566596/cur_line_info.patch.txt)

darrenkenny:
So should I add that patch to the commit? Or should I refactor too as suggested by Austin-Ray

flatcap:
> So should I add that patch to the commit?

Yes please, as a second commit

> Or should I refactor too as suggested by Austin-Ray

If you're happy to, you could add this as a third commit.

That would make it easy for us to see which looks best.
When we're happy we can squash the work into one tidy commit.

flatcap:
@darrenkenny Do you have some time to make the extra changes, suggested above?

PR #1848, opened 2019-09-10
Add mutt_date_epoch() API, use it instead of time()

### LINKED ISSUE #1824: Continue tidying use of time
* change `mutt_date_epoch_ms()` to `mutt_date_now_ms()`
* add `mutt_date_now()` (same as above, but in seconds)
* convert all usages of `time(NULL)` to `mutt_date_now()`
* possibly kill all uses of `time_t`, use `size_t` everywhere (now sure how hard this will be)

gahr:
Fixed #1824

### DIFF
--- a/browser.c
+++ b/browser.c
@@ -39,7 +39,6 @@
 #include <stdlib.h>
 #include <string.h>
 #include <sys/stat.h>
-#include <time.h>
 #include "mutt/mutt.h"
 #include "config/lib.h"
 #include "email/lib.h"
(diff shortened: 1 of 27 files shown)


browser.c
```diff
@@ -412,8 +411,7 @@ static const char *folder_format_str(char *buf, size_t buflen, size_t col, int c
         }
         else
         {
-          time_t tnow = time(NULL);
-          t_fmt = ((tnow - folder->ff->mtime) < 31536000) ? "%b %d %H:%M" : "%b %d  %Y";
+          t_fmt = ((mutt_date_epoch() - folder->ff->mtime) < 31536000) ? "%b %d %H:%M" : "%b %d  %Y";
```
flatcap: While we're here, could we comment `31536000` as 'One Year' (or `#define` it).

ncrypt/smime.c
```diff
@@ -100,7 +100,7 @@ struct SmimeCommandContext
 };
 
 char SmimePass[256];
-time_t SmimeExptime = 0; /* when does the cached passphrase expire? */
+size_t SmimeExptime = 0; /* when does the cached passphrase expire? */
```
flatcap: Should this go back to `time_t`, now?

PR #1852, opened 2019-09-14
fix: mutt_any_key_to_continue()

flatcap:
Originally, `mutt_any_key_to_continue()` relied on `fflush(stdin)` to clear the input buffer, but that isn't portable.
This version changes the terminal settings, temporarily, so that it can read single characters without blocking, or line-buffering.

**To test**:
- Add an error to your `.neomuttrc`
- Start NeoMutt
- See error message and prompt: 'Press any key to continue...'

**Expected behaviour**:
- <kbd>q</kbd>  exits NeoMutt

**Actual behaviour**:
- <kbd>q</kbd><kbd>Enter</kbd> is required

### DIFF
--- a/curs_lib.c
+++ b/curs_lib.c
@@ -524,18 +524,6 @@ void mutt_perror_debug(const char *s)
 }
 
 /**
- * mutt_flush_stdin - remove characters from stdin until '\n' or EOF is encountered
- */
-void mutt_flush_stdin(void)
-{
-  int c;
-  do
-  {
-    c = fgetc(stdin);
-  } while ((c != '\n') && (c != EOF));
-}
-
-/**
  * mutt_any_key_to_continue - Prompt the user to 'press any key' and wait
  * @param s Message prompt
  * @retval num Key pressed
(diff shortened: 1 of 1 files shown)


curs_lib.c
```diff
@@ -543,28 +531,44 @@ void mutt_flush_stdin(void)
...
   else
     fputs(_("Press any key to continue..."), stdout);
   fflush(stdout);
-  int ch = fgetc(stdin);
-  mutt_flush_stdin();
-  tcsetattr(fd, TCSADRAIN, &old);
+
+  char ch = '\0';
+  // Wait for a character.  This might timeout, so loop.
+  while (read(fd, &ch, 1) == 0)
```
Austin-Ray: Is there a more readable way to construct this control flow? I missed the `;` the first time I read this code.

flatcap: A `;` on a blank line is common, idiomatic, C.
I count over 40 occurrences in the code.

Austin-Ray: I trust your judgment since you're much more experience with C than me.

fekir: @flatcap how did you count blank lines?
This would count too

````
while (read(fd, &ch, 1) == 0)
{
    ;
}
````


I missed the `;` on a first read too.
Personally I would prefer braces even for such simple statements (granted, with neomutt style guide policy about braces it doubles the line numbers), and it also makes code more consistent.
But I know some people prefer it this way, as stated in neomutt coding style guide..

Comment (flatcap) [reply curs_lib.c]: > how did you count blank lines?

Nothing fancy: `grep -B1 "^ \+;$" -R .`

> I would prefer braces even for such simple statements

An alternative would be a comment:

```c
while (condition)
  ; // Do nothing
```

I'll give it some consideration.

> neomutt coding style guide

I tried to keep it short, so people would read it.
At some point, I'd like to write down every 'rule' I've consciously applied.

flatcap:
> Is `tcsetattr` really supposed to work when opening opening `/dev/tty` as `O_RDONLY`?

Good question.  The calls succeed, so I'm confident the answer is 'yes' :-)

I think we're telling the kernel **how** we'd like to use the device, independent of actually reading or writing.

> could not see any statement in the documentation

No, nor I

> that the error message is misleading.
> pressing `q` closes the application

Ah, you got me there.
I snuck that in (56fafc676) when I was refactoring / testing all the paths through the startup code.

Altering the message would mean two new translations: the message and the 'q' (for consistency).

PR #1853, opened 2019-09-14
Upstream Buffer changes

flatcap:
This is less of a request for a review, more a case of here's some upstream merging for comparison.

- f9785ffa3 Convert mutt_display_message() to buffer pool
- 9d54533ba Convert mutt_save_message() to use buffer pool
- 46f9fce0c Convert recvattach save_attachment functions to use buffer pool.
- 819778dc2 Convert query_pipe_attachment to use buffer pool
- 51ac592d6 Convert print_attachment_list to use buffer pool

[This gist](https://gist.github.com/flatcap/1d12ea8380837184e84eb474dcce23ab) contains the original upstream patches and the tidied versions.
The tidied ones don't apply, but they make it simpler to cut'n'paste.

### DIFF
--- a/commands.c
+++ b/commands.c
@@ -1007,11 +1007,13 @@ int mutt_save_message(struct Mailbox *m, struct EmailList *el,
 
   bool need_passphrase = false;
   int app = 0;
-  char buf[PATH_MAX];
+  int rc = -1;
   const char *prompt = NULL;
   struct stat st;
   struct EmailNode *en = STAILQ_FIRST(el);
   bool single = !STAILQ_NEXT(en, entries);
+
+  struct Buffer *buf = mutt_buffer_pool_get();
 
   if (delete_original)
   {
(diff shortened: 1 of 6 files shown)


commands.c
```diff
@@ -215,7 +212,7 @@ int mutt_display_message(struct Mailbox *m, struct Email *e)
       if (e->security & APPLICATION_SMIME)
         crypt_smime_getkeys(e->env);
       if (!crypt_valid_passphrase(e->security))
-        return 0;
+        goto cleanup;
```
Austin-Ray: Will jumping to `cleanup` cause an issue since `cleanup` is defined as:

```C
cleanup:
  mutt_buffer_pool_release(&tempfile);
  return rc;
```

and `tempfile` does not exist when the `goto` is called?

commands.c
```diff
@@ -1034,67 +1039,70 @@ int mutt_save_message(struct Mailbox *m, struct EmailList *el,
...
-    return -1;
+    goto cleanup;
 
   /* Trim any trailing '/' */
-  if (buf[pathlen - 1] == '/')
-    buf[pathlen - 1] = '\0';
+  if (buf->data[pathlen - 1] == '/')
+    buf->data[pathlen - 1] = '\0';
 
   /* This is an undocumented feature of ELM pointed out to me by Felix von
```
Austin-Ray: `unimportant`: You wouldn't happen to know what undocumented feature is mentioned here, would you?

recvattach.c
```diff
@@ -456,36 +456,26 @@ int attach_tag(struct Menu *menu, int sel, int act)
...
-  bool addsep = (savedir[savedirlen - 1] != '/');
-  size_t newbuflen = savedirlen + buflen + addsep;
-
-  if (bufsize < newbuflen)
+  struct Buffer *tmp = mutt_buffer_pool_get();
+  if (C_AttachSaveDir)
   {
-    return;
+    mutt_buffer_addstr(tmp, C_AttachSaveDir);
+    if (tmp->dptr[-1] != '/')
```
Austin-Ray: I've never seen `-1` used as an index. What does it do?

commands.c
flatcap: No, I wasn't aware of it.
Reading the code, it seems that saving a message to `.` means the last place you saved to, rather than the current directory of NeoMutt.

This doesn't seem to be documented in neomutt's manual either.
I wonder if we could kill this unexpected behaviour.

commands.c
flatcap: Argh!  UB!  Good catch.

recvattach.c
flatcap: It's simply the previous character, the one before `tmp->dptr[0]`

In this context, it's a little bit hacky.
The `Buffer` holds a large array of `char`, and `dptr` keeps track of the `NUL` at the end.
A string appended to the `Buffer` will start at `dptr`.
`dptr[0] == '\0'`
`dptr[-1]` is the last character of the string in the `Buffer` (which we know isn't empty).

commands.c
flatcap: Fixed the commit, in-place

recvattach.c
Austin-Ray: > It's simply the previous character, the one before tmp->dptr[0]

Isn't that undefined behavior since we're accessing data outside the bounds of `dptr`? 

I understand that it's safe here because we know `dptr` points to the `Buffer`'s `char`, I'm just curious.

commands.c
Austin-Ray: > This doesn't seem to be documented in neomutt's manual either.
I wonder if we could kill this unexpected behaviour.

It might be safer to document the behavior. It's plausible that someone relies upon it.

recvattach.c
flatcap: No.  It would only be undefined behaviour if we went outside of the block of memory that the `Buffer` allocated.
Within that block, we're free to reference it any way we choose.

> I understand that it's safe here

Yeah and that's quite a few assumptions.
We _ought_ to check that:
- `data != NULL` (there's data)
- `dsize != 0` (ditto)
- `dptr < (data + dsize - 1)` (ptr is within range)

Issue #1730, opened 2019-05-31
color sidebar_new sets the color corresponding to O instead of N emails


doa379:
I have color sidebar_new <color0> <color1>

## Expected Behaviour

The sidebar should color only those mailboxes with N emails.

## Actual Behaviour

The sidebar colors all mailboxes with O emails instead of mailboxes with N emails.

## How often does this happen?

- Always

## NeoMutt Version

```
NeoMutt 20180716

System: Linux 5.0.0-19.1-liquorix-amd64 (x86_64)
ncurses: ncurses 6.1.20181013 (compiled with 6.1.20180714)
libidn: 1.33 (compiled with 1.33)
hcache backends: tokyocabinet

Compiler:
Using built-in specs.
COLLECT_GCC=cc
COLLECT_LTO_WRAPPER=/usr/lib/gcc/x86_64-linux-gnu/7/lto-wrapper
OFFLOAD_TARGET_NAMES=nvptx-none
OFFLOAD_TARGET_DEFAULT=1
Target: x86_64-linux-gnu
Configured with: ../src/configure -v --with-pkgversion='Debian 7.3.0-25' --with-bugurl=file:///usr/share/doc/gcc-7/README.Bugs --enable-languages=c,ada,c++,go,brig,d,fortran,objc,obj-c++ --prefix=/usr --with-gcc-major-version-only --program-suffix=-7 --program-prefix=x86_64-linux-gnu- --enable-shared --enable-linker-build-id --libexecdir=/usr/lib --without-included-gettext --enable-threads=posix --libdir=/usr/lib --enable-nls --with-sysroot=/ --enable-clocale=gnu --enable-libstdcxx-debug --enable-libstdcxx-time=yes --with-default-libstdcxx-abi=new --enable-gnu-unique-object --disable-vtable-verify --enable-libmpx --enable-plugin --enable-default-pie --with-system-zlib --with-target-system-zlib --enable-objc-gc=auto --enable-multiarch --disable-werror --with-arch-32=i686 --with-abi=m64 --with-multilib-list=m32,m64,mx32 --enable-multilib --with-tune=generic --enable-offload-targets=nvptx-none --without-cuda-driver --enable-checking=release --build=x86_64-linux-gnu --host=x86_64-linux-gnu --target=x86_64-linux-gnu
Thread model: posix
gcc version 7.3.0 (Debian 7.3.0-25) 

[...]

Issue #1629, opened 2019-04-01
Bogus "New mail" in every mailbox


hvkls:
## Expected Behaviour

New mail should be reported in mailboxes with new mail only.

## Actual Behaviour

On Mac OS Mojave (10.14.4) and neomutt [https://github.com/neomutt/neomutt/commit/92885b5d46e81007c0a6a699121e5693bd76165a](https://github.com/neomutt/neomutt/commit/92885b5d46e81007c0a6a699121e5693bd76165a), all of my 79 mailboxes (MBOX format) report new mail, while there is only one mailbox where this is actually true. 

When I open the respective mailboxes, the number of "new mail" boxes decreases. But when I quit and restart neomutt, again all of them have the bogus 'new mail'.

## When did it start to happen?

Sorry, all I can remember is I used to compile neomutt with --enable-buffy-size (was that the name?) and it worked fine back then. 

## NeoMutt Version

NeoMutt 20180716
Copyright (C) 1996-2016 Michael R. Elkins and others.
NeoMutt comes with ABSOLUTELY NO WARRANTY; for details type 'neomutt -vv'.
NeoMutt is free software, and you are welcome to redistribute it
under certain conditions; type 'neomutt -vv' for details.

System: Darwin 18.5.0 (x86_64)
ncurses: ncurses 6.1.20180127 (compiled with 6.1.20180127)
libiconv: 1.15
libidn: 1.35 (compiled with 1.35)
hcache backends: kyotocabinet

Compiler:
Apple LLVM version 8.0.0 (clang-800.0.42.1)
Target: x86_64-apple-darwin15.6.0
Thread model: posix
InstalledDir: /Applications/Xcode.app/Contents/Developer/Toolchains/XcodeDefault.xctoolchain/usr/bin

Configure options: --prefix=/opt --sysconfdir=/opt/etc --with-gpgme=/opt/local --notmuch --with-ui=ncurses --with-ncurses=/opt/local --enable-locales-fix --with-ssl=/opt/local --gpgme --gss --smime --gnutls --sasl --kyotocabinet --with-lua=/opt/local --enable-debug

Compilation CFLAGS: -g -O2 -std=c99 -D_ALL_SOURCE=1 -D_GNU_SOURCE=1 -D__EXTENSIONS__ -I/opt/local/include -I/opt/include -DNCURSES_WIDECHAR

Default options:
  +attach_headers_color +compose_to_sender +compress +cond_date +debug 
[...]

PR #1859, opened 2019-09-26
Fix mbox's understanding of 'new'

flatcap:
Recent refactoring introduced several bugs related to whether a mailbox was considered _new_.

- 252e4b757 fix: mbox_check_stats() retvals
  We weren't distinguising between 'success: no new mail' and 'success: some new mail', which led to unnecessary checks (which are expensive for mbox).

- cf599f50b fix mbox's handling of new mail
  Make sure that `Mailbox.has_new` behaves correctly, which means checking
  - The timestamp on the mbox file
  - Are there new emails in the Mailbox?
  - Has the user visited the Mailbox?

- a4d9169b8 fix <change-folder> when $change_folder_next is set
  Setting the config variable should pick the **next** Mailbox with new email, not the first

- 1c5ded977 fix sidebar colour for 'new' mailboxes
  Only set the **new** colour, if `Mailbox.has_new` is set

Fixes: #1629 
Fixes: #1730

### DIFF
--- a/maildir/maildir.c
+++ b/maildir/maildir.c
@@ -549,7 +549,7 @@ static int maildir_mbox_check_stats(struct Mailbox *m, int flags)
   if (check_new || check_stats)
     maildir_check_dir(m, "cur", check_new, check_stats);
 
-  return 0;
+  return (m->msg_new > 0);
 }
 
 /**
--- a/maildir/maildir.c
+++ b/maildir/maildir.c
@@ -549,7 +549,7 @@ static int maildir_mbox_check_stats(struct Mailbox *m, int flags)
   if (check_new || check_stats)
     maildir_check_dir(m, "cur", check_new, check_stats);
 
-  return (m->msg_new > 0);
+  return m->msg_new;
 }
 
 /**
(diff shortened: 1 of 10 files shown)


maildir/maildir.c
```diff
@@ -549,7 +549,7 @@ static int maildir_mbox_check_stats(struct Mailbox *m, int flags)
   if (check_new || check_stats)
     maildir_check_dir(m, "cur", check_new, check_stats);
 
-  return 0;
+  return (m->msg_new > 0);
```
Austin-Ray: `nitpick`: It'd be best to write this out to explicitly return an integer instead of a boolean. While I know a boolean value maps to an integer, the mixed return types can cause some confusion (at least in my case.) In addition, it'd be more consistent with other backends' implementations.

mbox/mbox.c
```diff
@@ -1781,7 +1781,10 @@ static int mbox_mbox_check_stats(struct Mailbox *m, int flags)
     }
   }
 
-  return 0;
+  if (m->msg_new == 0)
+    m->has_new = false;
+
+  return (m->msg_new > 0);
```
Austin-Ray: `nitpick`: Same point as my `maildir/maildir.c` comment.

maildir/maildir.c
flatcap: Hmm... yeah.  I'm not quite sure what I was thinking.
I've changed `check_stats()` and wrappers and implementations to return `int` - number of new emails.

PR #1870, opened 2019-10-11
refactor Email, Body, Envelope

flatcap:
**Note**: I'm not planning to merge this PR, it's just an RFC.

Move all the header-cached members to the top of the struct.

---

Below the cached members are a few 'questionable' members, which probably shouldn't be cached.
Next come a bunch of pointers,
Then finally all the members that are ignored by the header cache (mostly View data).

The header cache currently stores the entire `struct Email` which is 192 bytes.
By only caching the important data, we can reduce this to 56 bytes (possibly less).

Most of these non-cached members will eventually be moved to a new `struct EmailView`.

---

I'm still working on splitting Data and Views (see [Event-Driven NeoMutt](https://gist.github.com/flatcap/42390bac63c98612818e82ab62152010)).

### DIFF
--- a/email/body.h
+++ b/email/body.h
@@ -33,54 +33,62 @@
  */
 struct Body
 {
+  LOFF_T offset;                  ///< offset where the actual data begins
+  LOFF_T length;                  ///< length (in bytes) of attachment
+
+  bool badsig           : 1;      ///< Bad cryptographic signature (needed to check encrypted s/mime-signatures)
+  bool force_charset    : 1;      ///< Send mode: don't adjust the character set when in send-mode.
+  bool goodsig          : 1;      ///< Good cryptographic signature
+#ifdef USE_AUTOCRYPT
+  bool is_autocrypt     : 1;      ///< Flag autocrypt-decrypted messages for replying
+#endif
+  bool noconv           : 1;      ///< Don't do character set conversion
+  bool use_disp         : 1;      ///< Content-Disposition uses filename= ?
+  bool warnsig          : 1;      ///< Maybe good signature
+
+  unsigned int disposition : 2;   ///< content-disposition
+  unsigned int encoding    : 3;   ///< content-transfer-encoding
+  unsigned int type        : 4;   ///< content-type primary type
+
+  // ------------------------------------------------------------
+  // Questionable
+  bool attach_qualifies : 1;      ///< This attachment should be counted
+  bool collapsed        : 1;      ///< Used by recvattach
+  bool deleted          : 1;      ///< Attachment marked for deletion
+  bool tagged           : 1;      ///< This attachment is tagged
+  bool unlink           : 1;      ///< If true, `filename` should be unlink()ed before free()ing this structure
+  short attach_count;             ///< Number of attachments
+  long hdr_offset;                ///< Offset in stream where the headers begin.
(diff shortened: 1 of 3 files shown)


flatcap:
When combined with #1871, the amount of header cache data in `struct Email` is reduced from 192 to just **32 bytes**.
```c
struct Email
{
  SecurityFlags security;      ///< bit 0-10: flags, bit 11,12: application, bit 13: traditional pgp
  bool active           : 1;   ///< Message is not to be removed
  bool expired          : 1;   ///< Already expired?
  bool flagged          : 1;   ///< Marked important?
  bool mime             : 1;   ///< Has a MIME-Version header?
  bool old              : 1;   ///< Email is seen, but unread
  bool read             : 1;   ///< Email is read
  bool replied          : 1;   ///< Email has been replied to
  bool superseded       : 1;   ///< Got superseded?
  bool trash            : 1;   ///< Message is marked as trashed on disk (used by the maildir_trash option)
  unsigned int zhours   : 5;   ///< Sender TZ: Hours away from UTC
  unsigned int zminutes : 6;   ///< Sender TZ: Minutes away from UTC
  bool zoccident        : 1;   ///< Sender TZ: True, if west of UTC, False if east
  time_t date_sent;            ///< Time when the message was sent (UTC)
  time_t received;             ///< Time when the message was placed in the mailbox
  int lines;                   ///< How many lines in the body of this message?
};
```

flatcap:
The amount of header cache data in `struct Body` can be reduced from 176 to just **40 bytes**.
```c
struct Body2
{
  long hdr_offset;                ///< Offset in stream where the headers begin
  LOFF_T offset;                  ///< offset where the actual data begins
  LOFF_T length;                  ///< length (in bytes) of attachment
  time_t stamp;                   ///< Time stamp of last encoding update
  signed short attach_count;      ///< Number of attachments
  bool attach_qualifies    : 1;   ///< This attachment should be counted
  bool badsig              : 1;   ///< Bad cryptographic signature (needed to check encrypted s/mime-signatures)
  bool collapsed           : 1;   ///< Used by recvattach
  bool deleted             : 1;   ///< Attachment marked for deletion
  bool force_charset       : 1;   ///< Send mode: don't adjust the character set when in send-mode
  bool goodsig             : 1;   ///< Good cryptographic signature
  bool is_autocrypt        : 1;   ///< Flag autocrypt-decrypted messages for replying
  bool is_signed_data      : 1;   ///< A lot of MUAs don't indicate S/MIME signed-data correctly,
  bool noconv              : 1;   ///< Don't do character set conversion
  bool tagged              : 1;   ///< This attachment is tagged
  bool unlink              : 1;   ///< If true, `filename` should be unlink()ed before free()ing this structure
  bool use_disp            : 1;   ///< Content-Disposition uses filename= ?
  bool warnsig             : 1;   ///< Maybe good signature
  unsigned int disposition : 2;   ///< content-disposition
  unsigned int encoding    : 3;   ///< content-transfer-encoding
  unsigned int type        : 4;   ///< content-type primary type
};
```

flatcap:
This will return, when I've got more time for analysis.

flatcap:
Branch backed up to: https://github.com/flatcap/neomutt/tree/2020-02-23/email

PR #1871, opened 2019-10-12
pop: refactor refno

flatcap:
Move `Email->refno` to `PopEmailData->refno`

`Email->refno` was only being used in the POP backend.
The value **was** saved in the Header Cache, but the restored value was never used (see `pop_fetch_headers()`).


flatcap:
> include the "why" in the commit message

Done.  Thanks.

PR #1872, opened 2019-10-14
fix read-only mbox

flatcap:
If an mbox Mailbox doesn't have write privs, try opening it read-only.

---

Before the architectural changes, there could be multiple open Mailbox
objects representing a single folder.

By default a folder would be given a read-only Mailbox.  When the folder
needed to be written to, another Mailbox would be opened read-write.
This lead to synchronisation problems between the two Mailboxes.

### DIFF
--- a/mbox/mbox.c
+++ b/mbox/mbox.c
@@ -900,6 +900,36 @@ int mbox_ac_add(struct Account *a, struct Mailbox *m)
 }
 
 /**
+ * mbox_open_readwrite - Open an mbox read-write
+ * @param m Mailbox
+ * @retval ptr FILE handle
+ *
+ * This function ensures that the FILE and readonly flag are changed atomically.
+ */
+static FILE *mbox_open_readwrite(struct Mailbox *m)
+{
+  FILE *fp = fopen(mailbox_path(m), "r+");
+  if (fp)
+    m->readonly = false;
+  return fp;
+}
+
+/**
+ * mbox_open_readwrite - Open an mbox read-only
+ * @param m Mailbox
+ * @retval ptr FILE handle
+ *
+ * This function ensures that the FILE and readonly flag are changed atomically.
+ */
+static FILE *mbox_open_readonly(struct Mailbox *m)
+{
+  FILE *fp = fopen(mailbox_path(m), "r");
+  if (fp)
+    m->readonly = true;
+  return fp;
+}
+
+/**
  * mbox_mbox_open - Open a Mailbox - Implements MxOps::mbox_open()
  */
 static int mbox_mbox_open(struct Mailbox *m)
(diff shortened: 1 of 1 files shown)


mbox/mbox.c
```diff
@@ -913,6 +913,11 @@ static int mbox_mbox_open(struct Mailbox *m)
...
   adata->fp = fopen(mailbox_path(m), "r+");
   if (!adata->fp)
+  {
+    adata->fp = fopen(mailbox_path(m), "r");
+    m->readonly = true;
+  }
+  if (!adata->fp)
   {
     mutt_perror(mailbox_path(m));
     return -1;
```
Austin-Ray: Quick question: This may be handled elsewhere, but do we need to clean up the `m->readonly` in the second `if (!adata->fp)` statement? Perhaps set `m->readonly = false` after opening read-write?

Example situation:
 1. Open mailbox read-write
 2. Open fails
 3. Open mailbox read-only
 4. Open fails; read-only set to true
 5. Reopen mailbox read-write
 6. Open is successful, but `m->readonly = false` from previous attempt.

flatcap: Yes, that sounds reasonable.  Changed.
Ultimately, I think that the read-only-ness needs to be a separate layer.
Perhaps, even, in the MailboxView.

mbox/mbox.c
```diff
@@ -914,6 +914,12 @@ static int mbox_mbox_open(struct Mailbox *m)
   adata->fp = fopen(mailbox_path(m), "r+");
   if (!adata->fp)
   {
+    adata->fp = fopen(mailbox_path(m), "r");
+    m->readonly = true;
+  }
+  if (!adata->fp)
+  {
+    m->readonly = false;
     mutt_perror(mailbox_path(m));
```
gahr: How about

```c
if (!adata->fp)
{
  adata->fp = fopen(mailbox_path(m), "r");
}
if (adata->fp)
{
      m->readonly = true;
}
else
{
  mutt_perror(mailbox_path(m));
  ...
}
```

gahr: How about adding two functions:

```c
FILE *mbox_open_readwrite(struct Mailbox *m)
{
  FILE *fp = fopen(mailbox_path(m), "r+");
  if (fp)
    m->readonly = false;
  return fp;
}

FILE *mbox_open_readonly(struct Mailbox *m)
{
  FILE *fp = fopen(mailbox_path(m), "r");
  if (fp)
    m->readonly = true;
  return fp;
}
```

Then use them

```c
adata->fp = mbox_open_readwrite(m);
if (!adata->fp)
{
  adata->fp = mbox_open_readonly(m);
}
if (!adata->fp)
{
  mutt_perror(mailbox_path(m));
  return -1;
}

flatcap: Thanks.  Applied.

Issue #1883, opened 2019-10-26
wrapmargin=4 wraps at 4 columns in the pager


sthen:
## Expected Behaviour

wrapmargin=4 works like wrap=-4

## Actual Behaviour

wrapmargin=4 works like wrap=4, wrapping at 4 columns wide in the pager

## Steps to Reproduce

Update to 2019-10-25, use a config with wrapmargin=4

## How often does this happen?

Always

## When did it start to happen?

When I upgraded from 20180716 to 2019-10-25

## NeoMutt Version

```
NeoMutt 2019-10-25
Copyright (C) 1996-2016 Michael R. Elkins and others.
NeoMutt comes with ABSOLUTELY NO WARRANTY; for details type 'neomutt -vv'.
NeoMutt is free software, and you are welcome to redistribute it
under certain conditions; type 'neomutt -vv' for details.

System: OpenBSD 6.6 (amd64)
ncurses: ncurses 5.7.20081102 (compiled with 5.7.20081102)
libiconv: 1.16
libidn2: 2.0.0 (compiled with 2.0.0)
GPGme: 1.13.1
hcache backends: kyotocabinet

Compiler:
OpenBSD clang version 8.0.1 (tags/RELEASE_801/final) (based on LLVM 8.0.1)
Target: amd64-unknown-openbsd6.6
Thread model: posix
InstalledDir: /usr/bin

Configure options: --mandir=/usr/local/man --with-ui=ncurses --docdir=/usr/local/share/doc/neomutt --with-lock=flock --idn=0 --idn2 --fmemopen --ssl --gpgme --kyotocabinet

Compilation CFLAGS: -O2 -pipe -g  -I/usr/local/include -std=c99 -D_ALL_SOURCE=1 -D_GNU_SOURCE=1 -D__EXTENSIONS__ -DNCURSES_WIDECHAR

Default options:
  +attach_headers_color +compose_to_sender +compress +cond_date +debug 
  +encrypt_to_self +forgotten_attachments +forwref +ifdef +imap +index_color 
  +initials +limit_current_thread +multiple_fcc +nested_if +new_mail +nntp +pop 
  +progress +quasi_delete +regcomp +reply_with_xorig +sensible_browser +sidebar 
  +skip_quoted +smtp +status_color +timeout +tls_sni +trash 

Compile options:
  -autocrypt +bkgdset +color +curs_set -fcntl +flock +fmemopen +futimens 
[...]

PR #1891, opened 2019-10-28
config: drop wrapmargin

flatcap:
'$wrapmargin' is a long-deprecated alternative to '$wrap', except in reverse.
Setting `wrapmargin=4` was equivalent to `wrap=-4`.

The new config system eliminated all the custom handling of options.

Closes: #1883 

---

It was never part of NeoMutt, so this is the simple solution.
Alternatively, I can fix the behaviour until some future 'Deprecation Day'.

### DIFF
--- a/contrib/sample.neomuttrc-tlr
+++ b/contrib/sample.neomuttrc-tlr
@@ -285,7 +285,7 @@ set smileys="^$"
 set ispell=iaspell
 
 set markers=no                          # Don't mark wrapped lines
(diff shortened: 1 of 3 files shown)


flatcap:
> In the future, ... warn the user if their config contains deprecated values

I have plan, of course :-)
The deprecated config items are already flagged.
The first stage of deprecation would issue warnings (possibly with a end-of-life date).
The second stage would issue errors and the third would be removal.

PR #1892, opened 2019-10-28
fix: crash when using 'alias -group'

flatcap:
This fixes the crash when creating alias groups.
However, it seems to leak the addresses and the group.
(I think it may always have leaked).

I can't see where the address groups are meant to go, or who's meant to own them.


gahr:
Yep, mutt_grouplist_free should probably iterate and call group_remove, but since it's called only on exit, who cares?

PR #1893, opened 2019-10-28
NULL Mailbox pointer if it's deleted

flatcap:
This is a bit of a hack.
It fixes the immediate problem where a Mailbox pointer was being invalidated by some long chain of events.

Before we perform the `folder-hook`, we set up an observer on the Mailbox.
The observer's data is the Mailbox pointer.
**If** the Mailbox is closed, then the pointer is `NULL`'d.
If not, we remove the observer before its data goes out of scope.

### DIFF
--- a/index.c
+++ b/index.c
@@ -604,6 +604,28 @@ void update_index(struct Menu *menu, struct Context *ctx, int check, int oldcoun
     menu->current = ci_first_message();
 }
 
+
+/**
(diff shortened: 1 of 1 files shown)


flatcap:
> let's make sure the commit message explains that it's a hack

Yes

> `TODO: properly fix` comment in the code would be nice as well

<sub>/me whispers anti-pattern ;-)</sub>

In this case, I _will_ add one.
Even with my wonderful notifications system, I don't know how to cleanly refactor this, yet.

PR #1895, opened 2019-10-28
fix: crash on failed ssl negotiation

flatcap:
The code freed the ssl data using a local pointer.
This mean that Connection was left with a dangling pointer.

(and remove a couple of redundant lines)


gahr:
LGTM - I'd be tempted to remove the `ssldata` variable and use `conn->sockdata` directly everywhere, to avoid falling into this pattern again.

flatcap:
> I'd be tempted to remove the `ssldata` variable

Yeah, good idea.  Done.

Issue #1898, opened 2019-10-29
crash/coredump after errormessage "Too many levels of symbolic links"


klartext:
## Expected Behaviour
Saving mail into folder, even though the folder is a symbolic link to a real file.
And in any case: no crash.

Saving an email into a folder (which is a symbolic link to the real folder-file).
After failure using "q" to quit neomutt.

## Actual Behaviour
Errormessage
```
<path-to-symlink-folder>: Too many levels of symbolic links (errno = 40)
```

When quitting right after that message, I got an errormessage with coredump.

```
<shallscriptname-that-invokes neomutt>: line 3:  3263 Aborted                 (core dumped) neomutt -f <imap-address>
```

crash after quit.

If NeoMutt crashed, did your OS create a 'coredump' file?
Yes.

## Steps to Reproduce
save an email into a folder, which is a symlink to the real folder file.
After you got the errormessage "Too many levels of symbolic links" in the status bar, quit.

## How often does this happen?

- Always

## When did it start to happen?
With upgrade to release NeoMutt 2019-10-25

## NeoMutt Version

```
NeoMutt 2019-10-25
Copyright (C) 1996-2016 Michael R. Elkins and others.
NeoMutt comes with ABSOLUTELY NO WARRANTY; for details type 'neomutt -vv'.
NeoMutt is free software, and you are welcome to redistribute it
under certain conditions; type 'neomutt -vv' for details.

System: Linux 5.3.7-arch1-1-ARCH (x86_64)
ncurses: ncurses 6.1.20180127 (compiled with 6.1.20180127)
libidn2: 2.2.0 (compiled with 2.2.0)
GPGme: 1.13.1
hcache backends: kyotocabinet, gdbm, lmdb

Compiler:
Using built-in specs.
COLLECT_GCC=cc
COLLECT_LTO_WRAPPER=/usr/lib/gcc/x86_64-pc-linux-gnu/9.2.0/lto-wrapper
Target: x86_64-pc-linux-gnu
[...]

PR #1905, opened 2019-10-30
Resolve symbolic links when saving a message

gahr:
Issue #1898

I'm not totally positive this is the right place to do this - perhaps symlinks should be resolved just before opening a mailbox, so the mailbox path isn't magically changed to follow the symlink.


flatcap:
> perhaps symlinks should be resolved just before opening a mailbox

Yeah, that would probably be better (noted).
But this will do for now.

Issue #1877, opened 2019-10-25
ctrl+c 


fekir:
## Expected Behaviour

While fetching message headers, pressing `ctrl+c` prompts following question

`Abort download and close mailbox? ([yes]/no):`

When pressing `y` or simply enter, the expected behavior would be for `neomutt` to close itself

## Actual Behavior
neomutt opens an empty mailbox (at least in my case the sidebar was filled with the correct folder names) and shows a red message "Error opening mailbox"`

## Steps to Reproduce

Open neomutt
press ctrl+c while loading headers
press enter

## How often does this happen?

- Always

## When did it start to happen?

When I compiled  `2019-10-25` for cygwin.

It might be related to https://github.com/neomutt/neomutt/commit/56fafc676d4a2220010b9db84b4cd851088b4269, as mentioned in https://github.com/neomutt/neomutt/pull/1852 when I found another inconsistent behavior, but I did not test it.

## NeoMutt Version

```
NeoMutt 2019-10-25
Copyright (C) 1996-2016 Michael R. Elkins and others.
NeoMutt comes with ABSOLUTELY NO WARRANTY; for details type 'neomutt -vv'.
NeoMutt is free software, and you are welcome to redistribute it
under certain conditions; type 'neomutt -vv' for details.

System: CYGWIN_NT-10.0-17134-WOW64 3.0.7-338.i686 (i686)
ncurses: ncurses 6.0.20171125 (compiled with 6.1.20190727)
libiconv: 1.14
libidn: 1.33 (compiled with 1.33)
GPGme: 1.9.0
hcache backends: bdb

Compiler:
Using built-in specs.
COLLECT_GCC=gcc
COLLECT_LTO_WRAPPER=/usr/lib/gcc/i686-pc-cygwin/7.4.0/lto-wrapper.exe
Target: i686-pc-cygwin
[...]

PR #1909, opened 2019-10-30
Allow interruption to quit while fetching IMAP headers

gahr:
Issue #1877

If we want to do that, this might be a way...


gahr:
As mentioned in #1877 , it is possible to get the same behaviour with

3. <kbd>Ctrl-C</kbd>, <kbd>Ctrl-C</kbd>, <kbd>y</kbd> -- Exit NeoMutt immediately

Issue #1913, opened 2019-10-31
Segfault for OP_COPY_MESSAGE with dest=src in mbox


luchr:
Problem with Neomutt 2019-10-25; Line numbers below refer to commit 3cd616659.

## Steps to reproduce

Take an mbox-file `testmailbox` with an email inside. Start neomutt

```
neomutt -n -f testmailbox
```

Press 'C' to start OP_COPY_MESSAGE and choose as destination the same mbox-file and confirm "yes" for append. Then I get a segfault.

## Expected Behaviour

no segfault

## Actual Behaviour

segfault

## Analysis; Why?

I hope this time-based analysis helps because one can see where src=dest is open WRONLY and then the next `fgets` on src fails.

```
in mutt_index_menu  at index.c:2657  call to mutt_save_message
  in mutt_save_message at commands.c
    line 1109 call to mx_mbox_open
      in mx_mbox_open at mx.c:314 call to mx_open_mailbox_append
        in mx_open_mailbox_append at mx.c
          line 193 m->append = true
          line 241 call to m->mx_ops->mbox_open_append
            in mbox_mbox_open_append at mbox/mbox.c:991 call to mutt_file_fopen with mode='w'
              in mutt_file_fopen at mutt/file.c
                line 598: flags |= O_WRONLY
                line 601: call to mutt_file_open
                line 604: call to fdopen(fd, mode='w')   [*]
    line 1132 call to mutt_save_message_ctx
      in mutt_save_message_ctx at commands.c:985 call to mutt_append_message
        in mutt_append_message at copy.c:886 call to append_message
          in append_message at copy.c:848 call to fgets on fp_in which was opened O_WRONLY [*]
            results in error (errno=9: Bad file descriptor) and return -1 propagated up
    line 1132 call returns -1 
    line 1136 call to mx_mbox_close
  mutt_save_message returns -1  next in loop 
line 1149: check for new mail
line 1154: mx_mbox_check
  in mx_mbox_check at mx.c:1035 call to m->mx_ops->mbox_check
    in mbox_mbox_check at mbox/mbox.c
      line 1112: fgets return NULL => modified = true
      line 1122: call to reopen_mailbox 
        in reopen_mailbox at mbox/mbox.c:582: call to mailbox_changed(m, MBN_CLOSED)
         -> ctx_mailbox_observer at context.c:304 ctx_cleanup sets Context->mailbox=NULL
line 1155: Context->mailbox is now NULL 
[...]

PR #1916, opened 2019-11-01
fix crash saving to mbox

flatcap:
Change the calls to `fopen()` to always allow reading.

Saving to an email to the current mbox opened the file write-only.
This causes a crash when NeoMutt tries to reuse the FILE handle for reading.

Fixes: #1913

### DIFF
--- a/mbox/mbox.c
+++ b/mbox/mbox.c
@@ -1574,7 +1574,9 @@ static int mbox_msg_open(struct Mailbox *m, struct Message *msg, int msgno)
   if (!adata)
     return -1;
 
-  msg->fp = adata->fp;
+  msg->fp = mutt_file_fopen(mailbox_path(m), "r");
+  if (!msg->fp)
+    return -1;
 
   return 0;
 }
@@ -1623,7 +1625,10 @@ static int mbox_msg_close(struct Mailbox *m, struct Message *msg)
   if (!msg)
     return -1;
 
-  msg->fp = NULL;
+  if (msg->write)
+    msg->fp = NULL;
+  else
+    mutt_file_fclose(&msg->fp);
 
   return 0;
 }
--- a/mbox/mbox.c
+++ b/mbox/mbox.c
@@ -989,18 +989,22 @@ static int mbox_mbox_open_append(struct Mailbox *m, OpenMailboxFlags flags)
   if (!adata)
     return -1;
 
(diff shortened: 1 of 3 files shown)


luchr:
Please don't merge this to master. Reading and writing (in parallel) is now possible, but the result is not what one would expect.

Example: Let's start with the following mbox-file:
```
From testperson@nowhere.go Fri Nov  1 13:29:54 2019
From: testperson@nowhere.go
To: testperson@nowhere.go
Subject: Testsubject
MIME-Version: 1.0
Content-Disposition: inline
X-Mutt-Fcc: =Sent
Status: RO
Date: Fri, 1 Nov 2019 13:29:54 +0100
Content-Length: 13
Lines: 1

Only a test.

```

after copying (key-shortcut 'C' and choosing the same mailbox as target and confirming with "yes") to the same mailbox the file looks like this:

```
From testperson@nowhere.go Fri Nov  1 13:29:54 2019
From testperson@nowhere.go Fri Nov  1 13:29:54 2019
o
o
bject: Testsubject
bject: Testsubject
ontent-Type: text/plain; charset=us-ascii
ontent-Type: text/plain; charset=us-ascii
ent
ent
us: RO
us: RO
ri, 1 Nov 2019 13:29:54 +0100
ri, 1 Nov 2019 13:29:54 +0100
Status: RO
C
Status: RO
C
 13
Lines: 1

```

Yes, somehow the mail was duplicated, but ... in a very unpleasant way.

flatcap:
Argh!  Thanks.

luchr:
Commit 9d1bbefad739dc4c78531023e57eb340bc917bb6 solves my testcase in issue #1913.

Thank you very much.

flatcap:
Great!

It's nearly there.  There's a slight issue re-opening the mailbox, and I need to check I haven't broken anything else :-)

flatcap:
OK, this should do for mbox, now.
I've separated the parts for clarity (as they're all independent):

- dca13d909 fix missing pointer check
  misc bug

- 968b800f9 mbox: don't leak file pointers
  on re-opening a Mailbox for appending

- 65e4ec7e6 mbox: after sync reopen with full privs
  sync prevented further writes to the Mailbox

- dce186b9d mbox: create separate read ptr
  Nasty bug causing interleaving of emails at the byte level
  
- a14029666 mx: don't reset data when re-opening a mailbox
  caused a visual duplication of all the Mailbox entries

PR #1959, opened 2019-11-12
fix build/tests for 32-bit arches

flatcap:
- Force mutt_date_epoch_ms() to return a 64-bit value
- Fix typo in test_mutt_date_gmtime()
- Fix some string-to-number tests

The only question I have is about the 32/64 bit test.
What's the best approach?

I've used this from `<limits.h>` and `<stdint.h>`:

```c
#if (__WORDSIZE == 64)
```


flatcap:
@aradici Please can you give this a try.

PR #1962, opened 2019-11-13
Avoid double question marks in bounce confirmation msg

gahr:


gahr:
> So, may I suggest the opposite? Strip the '?' from line 463.

Totally! Plus, I have increased the buffers for the prompt.. my terminal is way wider than 128 chars..

PR #1988, opened 2019-11-24
tests: fix context for failing idna and date-related tests

Kdecherf:
* **What does this PR do?**

This PR fixes two kind of test failures:
   - `test_mutt_idna_intl_to_local` and `test_mutt_idna_local_to_intl` fail when tests are run in a sandbox or with `--exec=never` because some variables like `C_Charset` are set to some values by other jobs and are not cleaned. `C_Charset`, `C_IdnEncode` and `C_IdnDecode` are now cleaned in the tests changing it.
   - a test is checking that the day of the year is always >= 119 which would cause it to fail when executed between Jan, 1st and Apr, 28th.

N/A

N/A


flatcap:
Good find!  Thanks for that.
I'd actually already fixed one of the '119' bugs, but missed the other.
(I've rebased the commits and tweaked a couple of variables names)

PR #2001, opened 2019-11-26
Removes extra quoted level in bundled themes

nkakouros:
* **What does this PR do?**
It removes an invalid highlight group from two themes in the contrib folder.

When sourcing for instance zenburn, I get the following error:

```
Warning in /usr/share/doc/neomutt/colorschemes/zenburn.neomuttrc, line 68: Maximum quoting level is 8
```

I removed `quoted9` in the themes. I hope this is the way to go.

### DIFF
--- a/contrib/colorschemes/vombatidae.neomuttrc
+++ b/contrib/colorschemes/vombatidae.neomuttrc
@@ -46,7 +46,6 @@ color quoted5       color36      color234
 color quoted6       color114     color234
 color quoted7       color109     color234
 color quoted8       color41      color234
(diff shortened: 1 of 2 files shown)


flatcap:
Thanks, @nkakouros.
I've decided to fix the code instead of changing the themes.
It's bug I meant to fix, but forgot about.

Normally, I'd update your branch, but GitHub won't let me.
(there's a tickbox when you create the Pull-Request to allow reviewer update).
Not to worry.

nkakouros:
That's great!

PR #2007, opened 2019-11-30
show notmuch library version

flatcap:
A trivial change, here.
Display the notmuch library version in the NeoMutt command-line version output.

On RHEL7 (LTS Redhat Version), I'm forced to build NeoMutt against an old version of the Notmuch libraries, due to dependency problems (Gmime 3.0, GPGMe 1.8.0).

We already display 'ncurses', 'libidn' and 'GPGme' library versions.

---

**Question**:

How do people feel about dropping the 'Compiler Version' info?

I've never found any occasion to need it?
(and it takes up a lot of space)

### DIFF
--- a/version.c
+++ b/version.c
@@ -389,8 +389,13 @@ static char *rstrip_in_place(char *s)
 void print_version(FILE *fp)
 {
   struct utsname uts;
+  bool tty = stdout ? isatty(fileno(stdout)) : false;
+  const char *fmt = "%s\n";
 
-  fprintf(fp, "%s\n", mutt_make_version());
+  if (tty)
+    fmt = "\033[1;36m%s\033[0m\n"; // Escape, cyan
+
+  fprintf(fp, fmt, mutt_make_version());
   fprintf(fp, "%s\n", _(Notice));
 
   uname(&uts);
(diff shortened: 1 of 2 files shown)


flatcap:
I've made a couple more changes:
- Added highlighting to the NeoMutt Version string
- Dropped the Compiler info

version.c
```diff
@@ -419,18 +426,19 @@ void print_version(FILE *fp)
   fprintf(fp, "\nGPGme: %s", mutt_gpgme_print_version());
 #endif
 
+#ifdef HAVE_NOTMUCH
+  fprintf(fp, "\nlibnotmuch: %d.%d.%d", LIBNOTMUCH_MAJOR_VERSION,
```
gahr: I'd spell this as "notmuch", as that seems the spelling used on https://notmuchmail.org/

flatcap: I wasn't sure what to put as the two names have different version numbers:

- `notmuch -v` : 'notmuch 0.29.1'
- `ls /usr/lib...` : 'libnotmuch.so.5.2.0'

NeoMutt only sees the '5.2.0'

gahr: Oh, it's the shared library's version, not the pacakge version.. confusing.. I'd prefer notmuch to dump the package version, which is what users look at generally

gahr: .. but notmuch doesn't (provide a package version), so yeah, better than nothing

Issue #2008, opened 2019-11-30
Cannot set a DT_MAILBOX type correctly from within Lua


darnir:
## Expected Behaviour

Running `mutt.set("trash", "+mail/Trash")` sets the trashdir to be `+mail/Trash`.

## Actual Behaviour

Instead of expanding the `+` with the path to `$folder` (which is set _before_ the lua script is sourced), the value of `$trash` is simply `+mail//Trash`.

As a result, when deleting a mail, neomutt tries to create a directory called `+mail` in my PWD before correctly resolving the path to the trashdir.

Looks like this is caused by some recent changes wherein the expansion of the mailbox paths is done earlier and Lua's calls somehow go around that expansion

If NeoMutt crashed, did your OS create a 'coredump' file?

## Steps to Reproduce

In a muttrc:

```
set folder = "/home/me/maildir"
lua mutt.set("trash", "+mail/Trash")
```

Check the value of trash when running Neomutt

## How often does this happen?

- Always

## When did it start to happen?

- When I upgraded
  Which version did you use to use?

The one from 2018 (before the massive overhaul)
## NeoMutt Version

```
NeoMutt 20191129
Copyright (C) 1996-2016 Michael R. Elkins and others.
NeoMutt comes with ABSOLUTELY NO WARRANTY; for details type 'neomutt -vv'.
NeoMutt is free software, and you are welcome to redistribute it
under certain conditions; type 'neomutt -vv' for details.

System: Linux 5.3.12-300.fc31.x86_64 (x86_64)
ncurses: ncurses 6.1.20190803 (compiled with 6.1.20190803)
libidn2: 2.3.0 (compiled with 2.3.0)
GPGme: 1.13.1
hcache backends: kyotocabinet, lmdb

Compiler:
Using built-in specs.
COLLECT_GCC=cc
COLLECT_LTO_WRAPPER=/usr/libexec/gcc/x86_64-redhat-linux/9/lto-wrapper
OFFLOAD_TARGET_NAMES=nvptx-none
OFFLOAD_TARGET_DEFAULT=1
Target: x86_64-redhat-linux
[...]

PR #2011, opened 2019-12-01
Fix setting of DT_MAILBOX type variables from Lua

darnir:
* **What does this PR do?**

Checks if the given variable in Lua maps to a DT_PATH type and if so,
expands the path to replace the "+" with the value of $folder

Closes: #2008

     for that)

Not needed

Not sure. It works fine, locally for my own configuration

Yes

### DIFF
--- a/mutt_lua.c
+++ b/mutt_lua.c
@@ -27,6 +27,13 @@
  * Integrated Lua scripting
  */
 
+#ifndef LUA_COMPAT_ALL
+#define LUA_COMPAT_ALL
+#endif
+#ifndef LUA_COMPAT_5_1
+#define LUA_COMPAT_5_1
+#endif
+
 #include "config.h"
 #include <lauxlib.h>
 #include <limits.h>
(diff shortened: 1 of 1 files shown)


darnir:
According to https://www.lua.org/pil/24.2.2.html the API is `lua_strlen` and it works on my Fedora 31 system as well. I wonder what version of Lua is available on travis that it refuses this API

flatcap:
Travis is running Ubuntu 18.04.
Ah, we're installing `liblua5.2-dev`

flatcap:
Hmm... Your guide says:

> This first edition was written for Lua 5.0.
> While still largely relevant for later versions, there are some differences.

In the header file, `lua_strlen()` is in a `#if defined(LUA_COMPAT_5_1)` block.
It seems that we want `lua_len()` -- https://www.lua.org/manual/5.3/manual.html#lua_len

flatcap:
It turns out that my Travis' Ubuntu 18.04 needed the symbol `LUA_COMPAT_ALL` defining.
My Fedora box and later versions of Ubuntu (like LGTM), need the symbol `LUA_COMPAT_5_1` defining.

Issue #2014, opened 2019-12-01
Neomutt crashes with segmentation fault when opening notmuch mailbox


rexVictor:
I'm trying to set up neomutt with mbsync and notmuch.
At the moment I run mbsync manually and afterwards a bash script manually that tags the mail.

While mbsync syncs three accounts to three different folders under ~/Mail.
The bash script looks like this:
```bash
notmuch new
notmuch tag -new +sent -- \
  tag:new and \( \
  from:my_mail_address1 or \
  from:my_mail_address2 or \
  ...
\)
```
My muttrc sources:
```
virtual-mailboxes "New" "notmuch://?query=tag:new"
virtual-mailboxes "Sent" "notmuch://?query=tag:sent"
```
## Expected Behaviour

When switching to Sent via sidebar, all mails tagged "sent" should show in index.

## Actual Behaviour
Neomutt crashes. However it does not crash if the mail from my Posteo mail account is not yet tagged by notmuch.

If NeoMutt crashed, did your OS create a 'coredump' file?
```
Process 889549 (neomutt) of user 1000 dumped core.
Stack trace of thread 889549:
#0  0x000056422155d10a email_free (neomutt)
#1  0x00005642214da5d1 mx_mbox_open (neomutt)
#2  0x00005642214b67ca main_change_folder (neomutt)
#3  0x00005642214b7b35 mutt_index_menu (neomutt)
#4  0x000056422148e5da main (neomutt)
#5  0x00007f726e885153 __libc_start_main (libc.so.6)
#6  0x000056422148e8de _start (neomutt)
```
## Steps to Reproduce
1. run mbsync with my posteo account
2. run my bash script which tags
3. start neomutt
4. switch to Sent mailbox

## How often does this happen?

Always

## When did it start to happen?

I'm installing new.
## NeoMutt Version

```
NeoMutt 20191129-11-ec01ad                                                                                                                                                [8/40707]
Copyright (C) 1996-2016 Michael R. Elkins and others.
NeoMutt comes with ABSOLUTELY NO WARRANTY; for details type 'neomutt -vv'.
NeoMutt is free software, and you are welcome to redistribute it
under certain conditions; type 'neomutt -vv' for details.

System: Linux 5.3.13-arch1-1 (x86_64)
ncurses: ncurses 6.1.20180127 (compiled with 6.1.20180127)
libidn2: 2.3.0 (compiled with 2.3.0)
GPGme: 1.13.1
[...]

PR #2016, opened 2019-12-02
fix: crash resizing notmuch mailbox

flatcap:
Fixes: #2014


flatcap:
Notmuch knows how many emails there are ahead of time.
That, combined with a change I made broke the way we use `mx_alloc_memory()`.
`M->msg_count` was larger than `M->email_max`, so the newly allocated memory wasn't being `NULL`d.

PR #2018, opened 2019-12-02
compose: draw status bar with highlights

zsugabubus:
* **What does this PR do?**
In compose menu, status bar was drawn without highlights applied. Now, it can be colorized with `color status ...`, just like every other status bars.

Not sure why it worked like this before.


flatcap:
Welcome to NeoMutt, @zsugabubus!

Sometimes the best ideas leave you wondering, "Why didn't we think of that before!"

Thanks :-)

Issue #1038, opened 2018-01-26
IMAP timeout causes `Please select mailbox first` error


stefanv:
I am on a somewhat unreliable network, so IMAP often times out.  When I then try to sync my mailbox, I see the following error:

```
imap_copy_messages [a0055 BAD Please select a mailbox first]?
```
Sometimes the line above starts with `imap_fast_trash`.

I'm not sure how to "select a mailbox first" as suggested; opening the folder is not an option, because that would first require syncing the current flags.

<details>
NeoMutt 20171215
Copyright (C) 1996-2016 Michael R. Elkins and others.
NeoMutt comes with ABSOLUTELY NO WARRANTY; for details type `neomutt -vv'.
NeoMutt is free software, and you are welcome to redistribute it
under certain conditions; type `neomutt -vv' for details.

System: Linux 4.12.0-2-amd64 (x86_64)
ncurses: ncurses 6.0.20171125 (compiled with 6.0)
libidn: 1.33 (compiled with 1.33)
hcache backends: tokyocabinet

Compiler:
Using built-in specs.
COLLECT_GCC=cc
COLLECT_LTO_WRAPPER=/usr/lib/gcc/x86_64-linux-gnu/7/lto-wrapper
OFFLOAD_TARGET_NAMES=nvptx-none
OFFLOAD_TARGET_DEFAULT=1
Target: x86_64-linux-gnu
[...]

PR #2026, opened 2019-12-05
imap: select the mailbox on reconnect

l29ah:
a band-aid solution; it's likely that some other IMAP states should
be restored on reconnection as well

hopefully fixes https://github.com/neomutt/neomutt/issues/1038


flatcap:
I've just tweak the code slightly: clang-format and add a function comment block.

PR #2037, opened 2019-12-11
Fix uncollapse_jump

### LINKED ISSUE #2036: uncollapse_jump not correctly respected
## Expected Behaviour

neomutt should show directly the first unread email in the thread, as mutt does. So, in the example below it should show B.

## Actual Behaviour

neomutt displays the first message in the thread (A) and not the first unread email (B). However, when you do next-unread it goes to the second unread message (D), skipping the first one. Also, if you do 'j' (instead of next-unread) it goes to the message after the first unread one (C) and not to the second message (B).

## Steps to Reproduce

Use neomutt with the following config:
```
set sort = threads                         # like gmail
set sort_aux = reverse-last-date-received  # like gmail
set uncollapse_jump                        # don't collapse on an unread message
set sort_re                                # thread based on regex
```

Have a thread with A->B->C->D->E and mark B and D as new.
Collapse the thread, then press enter.

- On every thread

## NeoMutt Version

```
NeoMutt 20191207  

System: Linux 5.3.15-300.fc31.x86_64 (x86_64)
ncurses: ncurses 6.1.20190803 (compiled with 6.1.20190803)
GPGme: 1.13.1
libnotmuch: 5.2.0
hcache backends: tokyocabinet, lmdb

Configure options: --sysconfdir=/etc SENDMAIL=/usr/sbin/sendmail ISPELL=/usr/bin/hunspell --notmuch --tokyocabinet --lmdb --gnutls --sasl --gss --gpgme --disable-idn --idn2 --autocrypt --lua

Compilation CFLAGS: -g -O2 -std=c99 -fno-delete-null-pointer-checks -D_ALL_SOURCE=1 -D_GNU_SOURCE=1 -D__EXTENSIONS__ -I/usr/include -DNCURSES_WIDECHAR

Default options:

Compile options:
  -locales_hack +lua +meta -mixmaster +nls +notmuch -openssl +pgp +sasl +smime 
  +sqlite +start_color +sun_attachment +typeahead 
MAILPATH="/var/mail"
SENDMAIL="/usr/sbin/sendmail"
SYSCONFDIR="/etc"
```

## Extra Info
Fedora 31

No

tmux

gahr:
Fixes #2036

### DIFF
--- a/index.c
+++ b/index.c
@@ -2504,10 +2504,10 @@ int mutt_index_menu(void)
           mutt_check_traditional_pgp(&el, &menu->redraw);
           emaillist_clear(&el);
         }
-        struct Email *e = get_cur_email(Context, menu);
-        if (!e)
+        e_cur = get_cur_email(Context, menu);
+        if (!e_cur)
           break;
-        int hint = e->index;
+        int hint = e_cur->index;
 
         op = mutt_display_message(MuttIndexWindow, Context->mailbox, e_cur);
         if (op < 0)


gahr:
The regression was introduced in e7682da912a64d1fe54b8a126dc072ef1699b5a5 - I think it's safe to merge :)

PR #2045, opened 2019-12-19
Nest Windows

flatcap:
These changes start to encapsulate panels/menus as "Dialogs".
(Dialogs will eventually be self-contained and event-driven).

## Visible Changes

These commits change the appearance of the screen.
The two main differences are the **Status Bar** and the **Sidebar**.

The Status Bar is now slightly narrower because it doesn't run under the Sidebar.
That's because the Status Bar is tightly coupled with the Index Panel.

The Sidebar is now only visible when the Index/Pager dialog is active.
It should never have been visible when composing an email, or selecting an alias.

### No functional changes

These changes do some preliminary tidying in an attempt to make the later commits clearer.

- 4121f6d0f tidy query_menu()
  return early to reduce indent
- 7b9e14121 rename menu members
- 2d06e6a3a rename window members
- 8eabfd587 factor out MuttSidebarWindow
  Pass window as a parameter
- 50d5aa906 window: add states
  Insert wrapper around window state
- c83901898 window: extend MuttWindow
  Add new members
- 8bb6bdf96 window: add helper functions

### Convert to a set of nested Windows

These changes look like a complete mess.
The original code created a few Windows that were continually reused by all the menus in NeoMutt.

The new code has nested Windows, only exposing three of them:

- `MuttHelpWindow` - Help bar at the top of the screen
- `MuttMessageWindow` - Message window / Command entry at the bottom of the screen
- `MuttDialogWindow` - A container which holds a stack of dialogs

Each time NeoMutt wants to show a menu to the user, it constructs a "Dialog" of nested Windows.
The next step is to attach their currently global data to the Window.
This will mean that a second instance of a Dialog would be independent of the first, e.g.

- Index Dialog -\> Compose Dialog -\> Index Dialog

The new file `reflow.c` performs the re-shaping, of Windows to fit the screen.
It **only** relies on settings on the `MuttWindow` to perform its work. e.g.

- `visible = true`
- `MUTT_WIN_ORIENT_VERTICAL`
- `MUTT_WIN_SIZE_FIXED`
- `req_cols = 15`

Commits:

- 5406347bc window: create nested layout
- 1088f9276 window: reflow
- 1338b7726 pager: externalise the Windows
- a0b18282b dialogs
- 3d26edb0b sidebar: split out config

### Upgrade all dialogs to use new Windows

These changes are still a bit verbose, a bit repetitive.
This is because the event loop (`mutt_menu_loop()`) happens **in** the dialog. 
The plan is to move this outside, so that the dialogs simply react to events.

- 5401574cd dialog addrbook
- 9f56e9c70 dialog attach
- 8f015f630 dialog autocrypt
- 7355fa12a dialog browser
- 57c401806 dialog compose
- 8188087dc dialog gnutls
- ddd4480ee dialog history
- d90a6b289 dialog ncrypt 3
- 10aad7a77 dialog postpone
- 7443c5c41 dialog query
- 0ce44dfb8 dialog remailer
- d0ea2b58f dialog ssl

---

layout

- **R** - Root Window
- **D** - Dialogs Container
- **I** - Index Dialog
- **4**,**5**,**6** - Containers for layout

### DIFF
--- a/Makefile.autosetup
+++ b/Makefile.autosetup
@@ -70,7 +70,8 @@ NEOMUTTOBJS=	addrbook.o alias.o bcache.o browser.o color.o commands.o \
 		mutt_parse.o mutt_signal.o mutt_socket.o mutt_thread.o mutt_window.o \
 		muttlib.o mx.o myvar.o pager.o pattern.o postpone.o progress.o query.o \
 		recvattach.o recvcmd.o resize.o rfc3676.o score.o send.o sendlib.o \
-		sidebar.o smtp.o sort.o state.o status.o system.o terminal.o version.o
+		sidebar.o smtp.o sort.o state.o status.o system.o terminal.o version.o \
+		reflow.o
 @if HAVE_LIBUNWIND
 NEOMUTTOBJS+=	backtrace.o
 @endif
--- a/addrbook.c
+++ b/addrbook.c
@@ -186,6 +186,34 @@ static int alias_sort_address(const void *a, const void *b)
 }
 
 /**
+ * mutt_dlg_alias_observer - Listen for config changes affecting the Alias menu - Implements ::observer_t()
+ */
+int mutt_dlg_alias_observer(struct NotifyCallback *nc)
+{
+  if (!nc || !nc->event || !nc->data)
+    return -1;
+
+  struct EventConfig *ec = (struct EventConfig *) nc->event;
+  struct MuttWindow *dlg = (struct MuttWindow *) nc->data;
+
+  if (mutt_str_strcmp(ec->name, "status_on_top") != 0)
+    return 0;
+
+  struct MuttWindow *win_first = TAILQ_FIRST(&dlg->children);
+
+  if ((C_StatusOnTop && (win_first->type == WT_INDEX)) ||
+      (!C_StatusOnTop && (win_first->type != WT_INDEX)))
+  {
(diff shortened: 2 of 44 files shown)


flatcap:
For testing, I hammered the layout with these macros whilst in the index, pager, compose, etc.
```
macro index,pager,compose,alias,browser <f5> "<enter-command>toggle help<enter><enter-command>set help?<enter>"
macro index,pager,compose,alias,browser <f6> '<enter-command>set pager_index_lines=`bc <<< "16 - $pager_index_lines"`<enter>'
macro index,pager,compose,alias,browser <f7> "<enter-command>toggle sidebar_on_right<enter><enter-command>set sidebar_on_right?<enter>"
macro index,pager,compose,alias,browser <f8> "<enter-command>toggle status_on_top<enter><enter-command>set status_on_top?<enter>"
```

A tricky config item is `status_on_top`.
If you change the value, it will affect both new dialogs afterwards and existing dialogs underneath the current one.

- Index panel
- `set status_on_top = yes`
- Compose email - status bar is on top
- `toggle status_on_top` - status bar is on bottom
- Cancel compose - status bar in **Index** is on the bottom

flatcap:
I've fixed the strange drawing problems after:
- `set pager_index_lines = 10`
- Open messsage
- `set pager_index_lines = 0`
- Close message

PR #2049, opened 2019-12-21
Add commands for manipulating content-id of attachments

yourealwaysbe:
Adds

    add-id
    edit-id
    pipe-id

for adding, editing, and piping attachment content-ids. This can be used
with cid: URLs in the email especially if using markdown. This is far from
a complete solution for embedding images in emails, but content IDs are
needed.

Also added documentation for new content ID commands

> All builds and tests are passing

Two are failing, but they were failing before i changed anything.


gahr:
Hi, thanks for the PR. I suggest a simplification:

* enhance NeoMutt so it always produces a Content-Id for an attachment
* add an expando to print the Content-Id to `attach_format`

This way, you'd be able to query the Content-id of any attachment, and reference them via `cid:` URIs in your markdown body.

Would that suffice for your use case? In particular, I don't understand why you'd need to explicitly `edit-id` and what use case you have for  `pipe-id`.

index.c
```diff
@@ -2572,6 +2572,24 @@ int mutt_index_menu(void)
           menu->redraw = REDRAW_CURRENT;
         break;
 
+      case OP_ADD_ID:
+        if (!prereq(Context, menu, CHECK_IN_MAILBOX | CHECK_MSGCOUNT | CHECK_VISIBLE | CHECK_ATTACH))
+          break;
+        mutt_add_content_id(CUR_EMAIL, CUR_EMAIL->content, NULL);
```
flatcap: Ah!  Since you created your branch, `CUR_EMAIL` (an evil macro) has been eliminated.
You can replace it, with:

```c
struct Email *e_cur = get_cur_email(Context, menu);
```

yourealwaysbe:
>     * enhance NeoMutt so it always produces a Content-Id for an attachment

I'd be happy with this if other were.

>     * add an expando to print the Content-Id to `attach_format`

This one i'm not sure about: the content IDs are quite long (~40 characters) so would probably be truncated in the view. Unless there's a way of displaying a particular expando in the message/command bar?

> Would that suffice for your use case? In particular, I don't understand why you'd need to explicitly `edit-id` and what use case you have for `pipe-id`.

edit-id i don't have much use for. It could be dropped.
pipe-id i would use with clipboard commands (e.g. wl-copy) to save having to display the id and then copy it with the mouse or similar.

flatcap:
Archived to: https://github.com/neomutt/neomutt-old/tree/content-id

PR #2073, opened 2020-01-12
ssl: drop RAND_status() checks

flatcap:
`RAND_status()` was added to OpenSSL 20 years ago.

https://github.com/openssl/openssl/blame/54a0d4ceb28d53f5b00a27fc5ca8ff8f0ddf9036/include/openssl/rand.h#L62

### DIFF
--- a/conn/ssl.c
+++ b/conn/ssl.c
@@ -72,11 +72,6 @@
 
 const int dialog_row_len = 128;
 
-/* Just in case OpenSSL doesn't define DEV_RANDOM */
-#ifndef DEV_RANDOM
-#define DEV_RANDOM "/dev/urandom"
-#endif
-
 /* LibreSSL defines OPENSSL_VERSION_NUMBER but sets it to 0x20000000L.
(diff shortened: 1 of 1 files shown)


flatcap:
> `DEV_RANDOM`

It was used in the `else` clause of the macro.
https://github.com/neomutt/neomutt/pull/2073/files#diff-ce1d6e16ac66f7b7485531cde2340711L105

> If so, how did you discover this?

... though I _do_ occasionally `grep -c` on all the `#define`s

flatcap:
btw, the CODEOWNERS didn't trigger the reviewers.
I'm not sure if it's because **I** match the owner regex, or because I'm an admin.

Austin-Ray:
@flatcap: I wonder if we need to enable the require review and require code owners to auto add them. I think I may have misinterpreted the `CODEOWNERS` documentation

flatcap:
> enable the require review and require code owners to auto add them

OK, let's give that a go.  I've enabled the option.

PR #2076, opened 2020-01-14
magic index/pager bars

flatcap:
The new nested Windows allows us to start being creative.
This is one possible use.

---

When set to an empty string, disable the index/pager bar completely.

  set status_format = ""
  set pager_format = ""

When combined with these settings, you can have a big view mode:

  set help = no
  set sidebar_visible = no
  set pager_index_lines = 0

### DIFF
--- a/index.c
+++ b/index.c
@@ -1082,7 +1082,9 @@ static void index_custom_redraw(struct Menu *menu)
       menu_redraw_current(menu);
   }
 
-  if (menu->redraw & REDRAW_STATUS)
+  struct MuttWindow *win_index_bar =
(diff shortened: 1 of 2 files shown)


flatcap:
> I'm not a fan of setting the format to an empty string to disable the index/pager.

Ah, that was just a user request from a while back -- something easily implemented.

> you have to remember your format string to go back

That could all be done with a macro and some `my_` variables.

> index_visible

Yeah, index_**bar**_visible, etc would be fine.
Another 'easy' task for someone to fill in the blanks.

PR #2080, opened 2020-01-15
refactor: imap's cmd_parse_status()

flatcap:
Tidy the code handling IMAP's response to a 'STATUS' command.

- 76e4fea50 tidy: cmd_parse_status()
- 555ba4dba add Account to ImapAccountData
- 9823f67fe lookup a Mailbox internally

When we get an untagged response from the server, we need to lookup the Mailbox.
We know the Account and have the Mailbox name from the server.

The MXAPI lookup functions perform a lot of unnecessary work searching through
all the Accounts, turning the Mailbox path into a URL and canonicalising it.

### DIFF
--- a/imap/command.c
+++ b/imap/command.c
@@ -781,6 +781,28 @@ static void cmd_parse_search(struct ImapAccountData *adata, const char *s)
 }
 
 /**
+ * find_mailbox - Find a Mailbox by its name
+ * @param adata Imap Account data
+ * @param name  Mailbox to find
+ * @retval ptr Mailbox
+ */
+static struct Mailbox *find_mailbox(struct ImapAccountData *adata, const char *name)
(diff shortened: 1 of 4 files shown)


gahr:
I guess the mailbox name returned by `STATUS` could be encoded as a quoted string (https://tools.ietf.org/html/rfc3501#section-4.3), hence the need for `imap_get_literal_count`. I don't see why an implementation would use it for short strings like mailbox names, but I wouldn't bet on that :)

flatcap:
> could be encoded as a quoted string

Ah!  That's the bit I was missing.  Thanks.
I'll **un**factor that bit.

flatcap:
Updated to leave the `imap_get_literal_count()` call intact.

Issue #2083, opened 2020-01-16
update_index() never gets called in index.c


sassmann:
Found a condition that can never be true.

https://github.com/neomutt/neomutt/blob/12d1180dc98019c5a4e040e3f53071b86e5af7ab/index.c#L741-L749

`mx_mbox_close()` only returns -1 or 0, so `check` gets assigned -1 or 0.

This condition can never be met as `MUTT_NEW_MAIL` has a value of 1 and `MUTT_REOPENED` has a value of 3.
Are we missing a `check = mx_mbox_check()` perhaps?

Similar situations here:

https://github.com/neomutt/neomutt/blob/12d1180dc98019c5a4e040e3f53071b86e5af7ab/index.c#L1836-L1841

https://github.com/neomutt/neomutt/blob/12d1180dc98019c5a4e040e3f53071b86e5af7ab/index.c#L1985-L1989

https://github.com/neomutt/neomutt/blob/12d1180dc98019c5a4e040e3f53071b86e5af7ab/index.c#L2030-L2048

flatcap:
Good catch!  Thanks, @sassmann 

---

P.S.  I've edited your post a little to show the source directly.

- '**Shift-Click**' to highlight a range of lines
- Click on the **`...`** in the margin
- Select '**Copy Permalink**'
- **Paste** link into a comment

sassmann:
Thanks Rich, I wasn't aware of that.

PR #2084, opened 2020-01-17
update_index() never gets called

flatcap:
Here's the call graph:

`imap_check_mailbox()` can return `MUTT_NEW_MAIL` or `MUTT_REOPENED`.
Despite what the function comments say, these functions can also return the values above:
- `mx_mbox_close()`
- `mx_mbox_sync()`
- `imap_sync_mailbox()`

Unfortunately, a couple of problems prevent the values being propagated back to the index.

### `mx_mbox_close()`

A recent refactoring meant that return values weren't being propagated.
Commit: 941e39acc fixes that (in this PR).

### `imap_sync_mailbox()`

This function **used** to return the result of `imap_check_mailbox()`.
That was changed by commit: 60270c9fa8

> imap: fixing race when syncing mailbox
> imap_check_mailbox returns a detailed state not only 0/-1.
> imap_sync_mailbox assumes that the function was returning only 0/-1.
> This change fixes that.

```diff
diff --git a/imap/imap.c b/imap/imap.c
@@ -1652,7 +1652,7 @@ int imap_sync_mailbox(struct Mailbox *m, bool expunge, bool close)
   imap_allow_reopen(m);
 
   rc = imap_check_mailbox(m, false);
-  if (rc != 0)
+  if (rc < 0)
     return rc;
 
   /* if we are expunging anyway, we can do deleted messages very quickly... */
```

I need to find out what this was fixing before I can change it back.

Issue: #2083

### DIFF
--- a/imap/imap.c
+++ b/imap/imap.c
@@ -1642,8 +1642,12 @@ out:
  * @param m       Mailbox
  * @param expunge if true do expunge
  * @param close   if true we move imap state to CLOSE
+ * @retval #MUTT_REOPENED  mailbox has been externally modified
+ * @retval #MUTT_NEW_MAIL  new mail has arrived
  * @retval  0 Success
  * @retval -1 Error
+ *
+ * @note The flag retvals come from a call to imap_check_mailbox()
  */
 int imap_sync_mailbox(struct Mailbox *m, bool expunge, bool close)
 {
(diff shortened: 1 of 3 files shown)


flatcap:
In **`imap_sync_mailbox()`**:

The original code returned early if `imap_check_mailbox()` returned flags.
After the fix, the function continued losing the returned flags.
Now, I've continued, but preserved the flags.

PR #2108, opened 2020-02-03
refactor: tidy Url usage

flatcap:
`struct Url` contains lots of pointers and a `STAILQ` (for the query strings).
Many of the `Url` uses didn't fully initialise the struct, which could lead to problems.

- Initialise all Urls on the stack
- Add `url_new()`
- Encode '=' too, in `url_pct_encode()`
- Generate query strings in `url_tobuffer()`
- Rename vars to 'url'

### DIFF
--- a/bcache.c
+++ b/bcache.c
@@ -62,7 +62,7 @@ struct BodyCache
 static int bcache_path(struct ConnAccount *account, const char *mailbox, struct BodyCache *bcache)
 {
   char host[256];
-  struct Url url = { U_UNKNOWN };
+  struct Url url = { 0 };
 
   if (!account || !C_MessageCachedir || !bcache)
     return -1;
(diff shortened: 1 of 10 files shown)


gahr:
I love objects on the stack :'-(

I understand you're addressing an issue of improper initialization of the `query_strings` members. I don't think this is a problem in practice, as
* the first member of a stailq structure is initialized to NULL when the stailq is empty
* reading the second member is only valid when the stailq is non-empty
so you're already on the safe side with the current implementation, unless I'm missing something.

flatcap:
Currently the code has 14 uninitialised `struct Url url;`

The notmuch paths are causing me some trouble.
They have query strings, which I'd like to reconstruct when using `url_tobuffer()`.
As this function is used by lots of other code, the `query_strings` needs to be initialised.

flatcap:
OK, take two.
Leave the `Url`s on the stack, but make sure they're initialised.

PR #2109, opened 2020-02-04
Unify unmailboxes and unvirtual-mailboxes commands

jindraj:
* **What does this PR do?**
`unmailboxes *` didn't consider virtual mailboxes (at least those defined by named-mailboxes)

I tested both unvirtual-mailboxes and unmailboxes commands. I have mailboxes defined as `named-mailboxes name mailbox`, where mailbox is maildir, mbox and notmuch query.
```
unmailboxes *
unmailboxes name
unmailboxes /path/to/mailbox

unvirtual-mailboxes *
unvirtual-mailboxes name
unvirtual-mailboxes /path/to/mailbox
```

All tested cases behaves as expected.

### DIFF
--- a/command_parse.c
+++ b/command_parse.c
@@ -2059,36 +2059,28 @@ enum CommandResult parse_unmailboxes(struct Buffer *buf, struct Buffer *s,
     struct MailboxNode *nptmp = NULL;
     STAILQ_FOREACH_SAFE(np, &ml, entries, nptmp)
     {
-      /* Decide whether to delete all normal mailboxes or all virtual */
-      bool virt = ((np->mailbox->magic == MUTT_NOTMUCH) && (data & MUTT_VIRTUAL));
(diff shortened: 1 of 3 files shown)


jindraj:
> Interestingly, I don't have a difference in behavior between `unmailboxes` and `unvirtual-mailboxes` even before your changes. Both can remove `mailboxes`, `named-mailboxes`, and `virtual-mailboxes` declaration.
> 
> Do you have a scenario where one works and the other doesn't? Perhaps I'm not testing comprehensively.

Sure. Having mix of local mailboxes (mbox, maildir) and notmuch mailboxes. Then `unmailboxes *` removed only the mboxes and maildirs. Notmuch mailboxes stayed untouched.
But `unmailboxes name` or `unamilboxes notmuch://?query=tag:mytag` worked well before.

flatcap:
Rebased, squashed.
Also, I've removed the `MUTT_VIRTUAL` symbol which isn't used any more.

Issue #2103, opened 2020-01-29
emails are double counted if FCC to current mailbox (maildir)


sassmann:
This issue happens with maildir.
By using the following hooks
```
folder-hook . unmy_hdr Fcc:
folder-hook INBOX$ my_hdr Fcc: =Sent,=INBOX
```
a copy of every mail sent from INBOX is also saved in INBOX.
However after sending the mail and waiting less than a minute the mail is shown twice in the index.
Tracing this behaviour with gdb, it seems the file gets renamed.
```
(gdb) p Context->mailbox->emails[34856]->path
$6 = 0x2e899660 "cur/1580220096.R2805722694203057930.p50:2,S"
(gdb) p Context->mailbox->emails[34857]->path
$7 = 0x52fc0eb0 "cur/1580220096.R2805722694203057930.p50,U=37387:2,S"

```
Because of the rename the file gets recognized as a new mail.

Reproduces with
3406e2d77ce6 maildir: Don't error on new mail

sassmann:
This might be caused by mbsync changing the filename in the background.
Wonder if there's a way to guard against this?

sassmann:
The hashes created in maildir_mbox_check() use maildir_canon_filename() as input.
https://github.com/neomutt/neomutt/blob/deaf77d65672dc5daab521e491e41778930e844b/maildir/maildir.c#L442-L452
maildir_canon_filename() cuts the fname at the ":"
https://github.com/neomutt/neomutt/blob/deaf77d65672dc5daab521e491e41778930e844b/maildir/shared.c#L1132-L1148
For the example above this results in
```
1580220096.R2805722694203057930.p50
1580220096.R2805722694203057930.p50,U=37387
```
Due to the difference the hashes created for both fnames will not match and things go awry.

Maybe we could strip the fname again at the "," ?

PR #2111, opened 2020-02-05
maildir: sanitize filename before hashing

sassmann:
https://github.com/neomutt/neomutt/issues/2103

### DIFF
--- a/maildir/shared.c
+++ b/maildir/shared.c
@@ -1128,6 +1128,10 @@ int mh_rewrite_message(struct Mailbox *m, int msgno)
  * maildir_canon_filename - Generate the canonical filename for a Maildir folder
  * @param dest   Buffer for the result
  * @param src    Buffer containing source filename
+ *
+ * @note         maildir filename is defined as: \<base filename\>:2,\<flags\>
+ *               but \<base filename\> may contain additional comma separated
+ *               fields.
  */
 void maildir_canon_filename(struct Buffer *dest, const char *src)
 {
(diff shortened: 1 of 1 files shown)


flatcap:
Hmm...
A Maildir filename is generally: `[UNIQUE-PART],[FLAGS]`
But, mbsync is adding the `U=UID` before the flags
(and Dovecot adds `S=NUM,W=NUM` before the flags, not that we see _that_)

As the `[UNIQUE-PART]` is unique, stripping the extra fields should be OK,
but I think it should happen in the `canon()` function for consistency.

Let me get a second opinion before you make any changes.

sassmann:
That would be fine with me. I was a bit hesitant to add it to `maildir_canon_filename()` because it is used in more places.

maildir/shared.c
```diff
@@ -1139,7 +1139,14 @@ void maildir_canon_filename(struct Buffer *dest, const char *src)
...
-  char *u = strrchr(dest->data, ':');
+  char *u = strchr(dest->data, ',');
+  if (u)
+  {
+    *u = '\0';
+    dest->dptr = u;
+    return;
+  }
+  u = strrchr(dest->data, ':');
   if (u)
```
gahr: I guess if we found ',' and changed that into a '\0' we won't find ':' anymore. Can we add a comment saying how we expect the format to be and how we want to modify it?

sassmann: Correct. The "," is used at least by mbsync and dovecot (probably more) to add some additional attributes *before* the flags. The old code only strips the flags. Since we know that we can already stop when we find a ",".
I can add comment to the commit, no problem.

gahr: Thanks!

PR #2113, opened 2020-02-07
Restore mailbox append flag after mbox_close

sassmann:
Make calls to mx_mbox_close() and mx_mbox_close() retain the same
value for m->append. This is import for special handling in
mx_mbox_close() of the append case.

### DIFF
--- a/commands.c
+++ b/commands.c
@@ -1145,8 +1145,8 @@ int mutt_save_message(struct Mailbox *m, struct EmailList *el,
     if (mutt_save_message_ctx(en->email, delete_original, decode, decrypt,
                               ctx_save->mailbox) != 0)
     {
-      m_save->append = old_append;
       mx_mbox_close(&ctx_save);
+      m_save->append = old_append;
       goto cleanup;
     }
 #ifdef USE_COMPRESSED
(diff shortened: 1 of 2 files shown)


gahr:
I think I kind of understand what you're doing, but can you please give some more context and a concrete example? It's a very complicated area, so a few more sentences won't hurt $future_me.

sassmann:
I'm preparing to add persistent contexts to neomutt and this is a prerequisite to be able to handle the append case correctly in mx_mbox_close().
Take mutt_save_message() as an example.
https://github.com/neomutt/neomutt/blob/3ae0d55f72032a5d74a4c526a00fafcf45e3a689/commands.c#L1122-L1123
m_save is opened with flag MUTT_NEWFOLDER
https://github.com/neomutt/neomutt/blob/3ae0d55f72032a5d74a4c526a00fafcf45e3a689/commands.c#L1204-L1205
but the appropriate call to mx_mbox_close() no longer knows about the  MUTT_NEWFOLDER flag and therefore cannot act upon it.

It's not an actual problem for the current code as mx_mbox_close() doesn't care for append flag value, but it will very soon.

gahr:
Got it, thanks

PR #2118, opened 2020-02-10
notmuch: change uri to url

flatcap:
Change all the '**URI**' references to '**URL**'

For a description of the differences, see: https://www.java67.com/2013/01/difference-between-url-uri-and-urn.html

All the **URI**s in NeoMutt are qualified with a scheme, e.g. `notmuch://`
This makes them **URL**s, which is also a more well-known term.

This PR changes one config variable: `nm_default_uri` to `nm_default_url` (but adds a synonym for backwards compatibility)
There are no functional changes to the code.

This PR only affects the Notmuch backend, so @Austin-Ray you get the right of veto.
If you're not happy, I'll just drop the change.

### DIFF
--- a/ChangeLog.md
+++ b/ChangeLog.md
@@ -86,7 +86,7 @@ yyyy-mm-dd
   - fix crash in mutt_extract_token()
   - force a screen refresh
   - fix crash sending message from command line
-  - notmuch: use nm_default_uri if no mailbox data
+  - notmuch: use nm_default_url if no mailbox data
   - fix forward attachments
   - fix: vfprintf undefined behaviour in body_handler
   - Fix relative symlink resolution
@@ -758,7 +758,7 @@ yyyy-mm-dd
   - Fix nntp group selection
   - Fix status color
   - Tidy up S/MIME contrib
-  - Do not try to create Maildir if it is an NNTP URI
+  - Do not try to create Maildir if it is an NNTP URL
   - Fix missing NONULL for mutt.set() in Lua
 * Translations
   - Fix German PGP shortkeys
--- a/doc/manual.xml.head
+++ b/doc/manual.xml.head
@@ -15172,8 +15172,8 @@ color index_tags green default
       <sect2 id="notmuch-using">
         <title>Using Notmuch</title>
 
-        <sect3 id="notmuch-folder-uri">
-          <title>Folders URI</title>
+        <sect3 id="notmuch-folder-url">
+          <title>Folders URL</title>
           <para>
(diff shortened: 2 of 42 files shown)


gahr:
The scheme is not what turns an URI into an URL. It's the fact that URLs can be "located", like on the network. See https://tools.ietf.org/html/rfc3986#section-1.1.3 and https://tools.ietf.org/html/rfc3986#section-3.1. I would not say that notmuch URIs are URLs, because unlike http or mailto urls, they cannot be used to locate a resource by themselves. They need additional context, like a notmuch database (e.g., `notmuch://?query=tag:inbox`).

I don't see the benefit of this change.

flatcap:
Fair enough.

I'm just trying to reduce the complexity of language in NeoMutt.
Is the user going to notice the different terminology?
Does the user need to know the difference?

There are 322 URLs and 33 URIs (almost all in Notmuch).

gahr:
Let's just use URI for everything: https://tools.ietf.org/html/rfc3305

flatcap:
OK, I'll give that a try.

flatcap:
I've made a first pass a the URL to URI conversion: [devel/uri](https://github.com/neomutt/neomutt/compare/devel/uri)

Here are the stats:
Settling on:
- **URL**: 42 files changed, 110 insertions(+), 109 deletions(-)
- **URI**: 84 files changed, 674 insertions(+), 673 deletions(-)

Austin-Ray:
+1 for unifying terminology, but renaming `nm_default_uri` to `nm_default_url` (or `smtp_url` to `smtp_uri` if we go with URL->URI) will break users configurations and require user correction. 

Is it possible to keep the old configuration name alongside the new one so we have backwards compatibility? e.g. deprecate `smtp_url` in favor of `smtp_uri` but accept either (preferring `smtp_uri`'s value.)

flatcap:
> Is it possible to keep the old configuration name alongside the new one

Yep, by the [power of synonyms](https://github.com/neomutt/neomutt/pull/2118/files#diff-02658e777616b6c8ca66a3f973ffb596R4925).
NeoMutt will silently accept the old name, but display the new one when dumping config.

Austin-Ray:
> > Is it possible to keep the old configuration name alongside the new one
> 
> Yep, by the [power of synonyms](https://github.com/neomutt/neomutt/pull/2118/files#diff-02658e777616b6c8ca66a3f973ffb596R4925).
> NeoMutt will silently accept the old name, but display the new one when dumping config.

Did not know about that functionality!

PR #2119, opened 2020-02-10
Rename library header files

flatcap:
This is an utterly trivial change.

Rename the header files that are `#include`d as library entry points.
Rather than `mutt/mutt.h`, `autocrypt/autocrypt.h`, they are now all `*/lib.h`

This makes it easier to see the library entry points.


gahr:
Cool! We could also go the other way around and have a single `lib/` direcory with `lib/mutt.h`, `lib/hcache.h` etc. We could then enforce that `*.c` files only include from `lib/` or locally. Just dumping thoughts :)

flatcap:
> the other way around and have a single `lib/` directory with `lib/mutt.h`, `lib/hcache.h` etc.

Yeah, that's a very `/usr/include` type of idea :-)
I'll have a play around later, to see what that might look like.

flatcap:
Currently there are two types of library header:
1. Aggregator: Contains just `#include`s
2. Kitchen sink: Everything the user will need in one place

A lot of the backends fall into category 2).
Now they're in separate dirs, I'd like to try splitting them up a bit.

flatcap:
I've created [[devel/libs2](https://github.com/neomutt/neomutt/compare/devel/libs2)], where I've moved all the headers to `/lib`, but there's no way it's going to compile yet.
There's probably a lot of work to be done, rearranging things, for this to work.

flatcap:
Updated `.clang-format` to better sort the new header files.

PR #2120, opened 2020-02-10
Fix commands that don’t need to have a non-empty mailbox to be valid

immae:
Some commands act on the whole mailbox (tag-pattern, delete-pattern,
search), and even though they don’t do anything when the mailbox is
empty, there is no reason to fail when it happens. This commit removes
the check that the mailbox is non-empty before doing said actions.

- [ ] All builds and tests are passing

This PR partially solved the issue of failing macros. It solves all cases where an existant message is not required, but it doesn’t handle cases where a macro would take both a user input AND a current message.
The functions I identified from that last case are:
- compose-to-sender (may not have a solution since a pre-input contains element from the selected message)
- edit-type (same)
- set-flag/clear-flag

### DIFF
--- a/index.c
+++ b/index.c
@@ -1642,7 +1642,7 @@ int mutt_index_menu(struct MuttWindow *dlg)
       case OP_JUMP:
       {
         int msg_num = 0;
-        if (!prereq(Context, menu, CHECK_IN_MAILBOX | CHECK_MSGCOUNT | CHECK_VISIBLE))
+        if (!prereq(Context, menu, CHECK_IN_MAILBOX))
           break;
         if (isdigit(LastKey))
           mutt_unget_event(LastKey, 0);
@@ -1686,8 +1686,7 @@ int mutt_index_menu(struct MuttWindow *dlg)
          */
 
       case OP_MAIN_DELETE_PATTERN:
-        if (!prereq(Context, menu,
-                    CHECK_IN_MAILBOX | CHECK_MSGCOUNT | CHECK_VISIBLE | CHECK_READONLY | CHECK_ATTACH))
+        if (!prereq(Context, menu, CHECK_IN_MAILBOX | CHECK_READONLY | CHECK_ATTACH))
         {
           break;
         }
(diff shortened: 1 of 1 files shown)


index.c
```diff
@@ -1852,12 +1852,17 @@ int mutt_index_menu(struct MuttWindow *dlg)
...
 
-      case OP_SEARCH:
+      // Initiating a search can happen on an empty mailbox, but
+      // searching for next/previous/... needs to be on a message and
+      // thus a non-empty mailbox
       case OP_SEARCH_REVERSE:
       case OP_SEARCH_NEXT:
       case OP_SEARCH_OPPOSITE:
         if (!prereq(Context, menu, CHECK_IN_MAILBOX | CHECK_MSGCOUNT | CHECK_VISIBLE))
           break;
```
flatcap: Please add a `/* fallthrough */` comment here to show that the behaviour is expected.

immae: Added :+1:

flatcap:
The Index supports 148 functions, some interactive, some not.

The functions can offer feedback in several ways:

- **No feedback**.  One of:
  1. Command has an obvious, visual effect
  2. Command was cancelled
  3. Command failed with no effect

- **Message**
  Notify the user that the command succeeded.

- **Warning**
  The command failed and we can explain what the user should do to fix the problem
  
- **Error**
  The command failed and we don't know why, or what the user could do to fix the problem

Currently, there are a LOT of error messages in NeoMutt.
I'd like to eliminate some of them and reduce many others to warnings. 

@immae 
I'm going to start adding notes to this PR.
If you have the time and enthusiasm to work on these problems, that'd be brilliant.
If not, don't worry :-)

flatcap:
## Movement functions

Most of these non-interactive give an error message when there aren't any emails in the mailbox.

When you reach the first/last message in the mailbox, I think we still want a **message** saying so.
We don't have scrollbars, so this is the only feedback that the user is at the end of the list.

I think that all the other error messages should be dropped.


| Operation         | Function                |
| :---------------- | :---------------------- |
| OP_BOTTOM_PAGE    | `menu_bottom_page()`    |
| OP_CURRENT_BOTTOM | `menu_current_bottom()` |
| OP_CURRENT_MIDDLE | `menu_current_middle()` |
| OP_CURRENT_TOP    | `menu_current_top()`    |
| OP_FIRST_ENTRY    | `menu_first_entry()`    |
| OP_HALF_DOWN      | `menu_half_down()`      |
| OP_HALF_UP        | `menu_half_up()`        |
| OP_LAST_ENTRY     | `menu_last_entry()`     |
| OP_MIDDLE_PAGE    | `menu_middle_page()`    |
| OP_NEXT_LINE      | `menu_next_line()`      |
| OP_NEXT_PAGE      | `menu_next_page()`      |
| OP_PREV_LINE      | `menu_prev_line()`      |
| OP_PREV_PAGE      | `menu_prev_page()`      |
| OP_TOP_PAGE       | `menu_top_page()`       |

flatcap:
I've been working through some more functions -- `OP_MAIN_NEXT_NEW` and friends -- but I'm not so sure about my ideas any more.
When Vim fails, on say a movement command, it doesn't print a message, it just clears the input buffer.
I'll keep thinking...

immae:
> I've been working through some more functions -- `OP_MAIN_NEXT_NEW` and friends -- but I'm not so sure about my ideas any more.
> When Vim fails, on say a movement command, it doesn't print a message, it just clears the input buffer.
> I'll keep thinking...

I thought of something like that just this morning: in case of problem, eat the input buffer because it’s the best way to ensure that no bad action happens (if you do a typo like "change folder to Trach and delete all there" you probably don’t want to delete everything in your current folder just because you mistyped Trash).

PR #2121, opened 2020-02-10
use global ContextList to store Contexts

sassmann:
Create a list to be able to store contexts globally.
The functionality will be useful to keep contexts in memory in the
future, instead of creating a new context on every mailbox switch.

No functional changes, just implement the infrastructure.

### DIFF
--- a/context.c
+++ b/context.c
@@ -42,6 +42,8 @@
 #include "sort.h"
 #include "ncrypt/lib.h"
 
+struct ContextList ContextList = STAILQ_HEAD_INITIALIZER(ContextList); ///< List of Contexts
+
 /**
  * ctx_free - Free a Context
  * @param[out] ptr Context to free
@@ -62,6 +64,7 @@ void ctx_free(struct Context **ptr)
   mutt_hash_free(&ctx->thread_hash);
   notify_free(&ctx->notify);
 
+  STAILQ_REMOVE(&ContextList, ctx, Context, entries);
   FREE(ptr);
 }
 
--- a/context.h
+++ b/context.h
@@ -29,6 +29,7 @@
 struct Email;
 struct EmailList;
 struct NotifyCallback;
+extern struct ContextList ContextList;
 
 /**
  * struct Context - The "current" mailbox
@@ -49,7 +50,9 @@ struct Context
 
   struct Mailbox *mailbox;
   struct Notify *notify;             ///< Notifications handler
+  STAILQ_ENTRY(Context) entries;     ///< Linked list
 };
+STAILQ_HEAD(ContextList, Context);
 
 /**
  * struct EventContext - An Event that happened to an Context
--- a/mx.c
+++ b/mx.c
@@ -255,8 +255,21 @@ struct Context *mx_mbox_open(struct Mailbox *m, OpenMailboxFlags flags)
   if (!m)
     return NULL;
 
-  struct Context *ctx = ctx_new();
-  ctx->mailbox = m;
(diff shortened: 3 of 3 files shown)


context.c
```diff
@@ -62,7 +65,15 @@ void ctx_free(struct Context **ptr)
   mutt_hash_free(&ctx->thread_hash);
   notify_free(&ctx->notify);
 
-  FREE(ptr);
+  STAILQ_FOREACH_SAFE(np, &contexts, entries, tmp)
+  {
+    if (np == ctx)
+    {
+      STAILQ_REMOVE(&contexts, np, Context, entries);
```
gahr: `STAILQ_REMOVE` already does the looping for you.

context.c
```diff
@@ -52,6 +54,7 @@ void ctx_free(struct Context **ptr)
     return;
 
   struct Context *ctx = *ptr;
+  struct Context *np = NULL, *tmp = NULL;
```
gahr: I would move these near the point of use, i.e., just before the STAILQ macro call.

context.h
```diff
@@ -29,6 +29,7 @@
 struct Email;
 struct EmailList;
 struct NotifyCallback;
+extern struct ContextList contexts;
```
gahr: I think we usually name globals CameCase?

mx.c
```diff
@@ -255,8 +255,22 @@ struct Context *mx_mbox_open(struct Mailbox *m, OpenMailboxFlags flags)
   if (!m)
     return NULL;
 
-  struct Context *ctx = ctx_new();
-  ctx->mailbox = m;
+  struct Context *ctx = NULL;
+  struct Context *np = NULL;
+  STAILQ_FOREACH(np, &contexts, entries)
+  {
+    if (np->mailbox == m)
```
gahr: Is pointer equality sufficient here? Do we assume that if we already have a mailbox, we'll always find it by the same pointer? Maybe that is correct. Should we also return an existing mailbox if their path match? Maybe, too..

sassmann: Yes, I think comparing pointers should be enough here, we can always extend the usage if necessary. In case you're curious on how I intend to use this have a look at
https://github.com/sassmann/neomutt/commit/d35aa695098b2dd941fda783c000da89425fe608

context.h
sassmann: OK, changed in new commit  1c77574.

context.c
sassmann: removed the loop in new commit 1c77574.

context.c
sassmann: no longer required, so dropped this.

context.c
```diff
@@ -62,7 +64,7 @@ void ctx_free(struct Context **ptr)
   mutt_hash_free(&ctx->thread_hash);
   notify_free(&ctx->notify);
 
-  FREE(ptr);
+  STAILQ_REMOVE(&ContextList, ctx, Context, entries);
```
gahr: you still need to free the `ctx` yourself, `STAILQ_REMOVE` only deals with unlinking from the list

context.h
```diff
@@ -29,6 +29,7 @@
 struct Email;
 struct EmailList;
 struct NotifyCallback;
+extern struct ContextList ContextList;
```
gahr: Uhm.. why?

context.c
sassmann: whoops, nice catch.

flatcap:
Archived to: https://github.com/neomutt/neomutt-old/tree/context-list

PR #2129, opened 2020-02-13
Fix context reinitialization in gpgme decryption

ftilde:
* **What does this PR do?**
In `decrypt_part` the gpgme context was correctly deinitialized before a
retry, but was not reinitialized due to an incorrect label location. This PR 
corrects the label position.

I'm not aware of relevant issues, but this PR fixes displaying of `smime-type=signed-data` inside of an encrypted envelope (not sure about the technically correct terms here) for me. These types of mails appear to be generated by Evolution and some apple mail client.


flatcap:
Welcome to NeoMutt, @ftilde.
Thanks for the fix -- well spotted!

PR #2133, opened 2020-02-14
libmutt: fix segfault when alias lookup returns NULL

andir:
`mutt_alias_lookup` is allowed to return NULL and thus we should try to
handle that case.

I have a macro in my neomutt configuration that is being rendered and
the template file contains placeholders like `@notmuchmutt@`. That
placeholder is then replaced with the path to the notmuch-mutt script. I
had a typo in the file so some occurrences were not being replaced
properly. Thus neomutt tried to use the `mutt_alias_lookup` function on
the string (since it contained the at-sign). That function did return
NULL and in the neomutt code we are then dereferencing `al` by using
`TAILQ_EMPTY`.

By properly checking if the `mutt_alias_lookup` invocation actually
returned something non-NULL we can work around this.

### DIFF
--- a/muttlib.c
+++ b/muttlib.c
@@ -223,7 +223,7 @@ void mutt_buffer_expand_path_regex(struct Buffer *buf, bool regex)
       case '@':
       {
         struct AddressList *al = mutt_alias_lookup(s + 1);
-        if (!TAILQ_EMPTY(al))
+        if (al && !TAILQ_EMPTY(al))
         {
           struct Email *e = email_new();
           e->env = mutt_env_new();
--- a/mx.c
+++ b/mx.c
@@ -1406,7 +1406,7 @@ int mx_path_canon(char *buf, size_t buflen, const char *folder, enum MailboxType
     {
       /* elm compatibility, @ expands alias to user name */
       struct AddressList *al = mutt_alias_lookup(buf + 1);
-      if (TAILQ_EMPTY(al))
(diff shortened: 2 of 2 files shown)


andir:
I am inclined to patch `TAILQ_EMPTY` to do the `NULL` check as the same pattern occurs a few more times throughout the source tree. Opinions?

flatcap:
Thanks for the fix @andir.

> I am inclined to patch `TAILQ_EMPTY` to do the `NULL` check

Yes, that's tempting, but `mutt/queue.h` is 3<sup>rd</sup>-party code (used in BSDs, Linux, etc).
When we adopted it, we decided to **not** change it to our needs, so that it would be easier to merge upstream changes.

>  the same pattern occurs a few more times throughout the source tree

If you have the time to find/flag/fix any of these, that'd be greatly appreciated!

andir:
I just pushed the only other version of this that was obvious to me after a quick `rg -A4 mutt_alias_lookup` on the tree.

It would probably be a good idea to of some kind of Coccinelle based lookup for more of these cases but I can't guarantee that I've the time & energy to do that right now.

flatcap:
> ... Coccinelle ...

I've [used Coccinelle in the past](https://github.com/neomutt/coccinelle), but it was hard work to create the scripts.
They quickly become brain-melting.

> I can't guarantee that I've the time & energy to do that right now

No worries.  Thanks.

PR #2135, opened 2020-02-14
Fix ability to specify user/login in url for smtp

ftilde:
* **What does this PR do?**
Even though user was correctly extracted from the smtp url, the login
was always read from the config variable, as it was never set before.
Now, immediately after extracting info from the url we copy the smtp
user to the login field if the user was set, but login wasn't.

I'm not aware of relevant issues. However, on current master I cannot send mails via the builtin smtp because of the described issue. On the most recent release (20191207) this still works. I suspect this broke during a refactoring/introduction of the `smtp_user` configuration option.

### DIFF
--- a/smtp.c
+++ b/smtp.c
@@ -365,6 +365,15 @@ static int smtp_fill_account(struct ConnAccount *cac)
     return -1;
   }
 
+  // If we set the user from the url, we also have to set the login now,
+  // because later attempts will only try to read from C_SmtpUser via
+  // `smpt_get_field` which may not be set.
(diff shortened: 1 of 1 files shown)


flatcap:
Ah!  This may already be fixed by another route.
Just before you created this PR, I pushed d4b578bfd6 -- it fixes a problem that: if 'login' doesn't exist, it should ask for the 'user'.
Please can you give `master` a try.
Thanks.

ftilde:
Yup, that looks like a similar problem and `d4b578bfd60448b92952123726d07778f4b2940e` works fine for me!

PR #2137, opened 2020-02-15
config: add path type

flatcap:
Further sub-class some config items:

> alias_file, attach_save_dir, autocrypt_dir, certificate_file, debug_file, entropy_file, header_cache, history_file, message_cachedir, news_cache_dir, newsrc, signature, smime_ca_location, smime_certificates, smime_keys, ssl_ca_certificates_file, ssl_client_cert, tmpdir

Before, they had type `DT_STRING|DT_PATH`.
Even though they were labelled as Paths, they weren't treated any differently from strings.

Now, `DT_PATH` has been promoted to a full type, with sub-types: `DT_PATH_FILE` and `DT_PATH_DIR`.
When config values are stored, they will be tidied: `~` will be expanded, and junk like `//` or `/./` will be removed.

- 9212976282 config: add path type
- 44cc5f73b5 test: add config/path
- 81aef62c5c config: convert path variables
- 34c633d740 propagate is_dir
- 31127e2d49 drop reset_tilde
  A 'temporary' function that expanded `~` in the initial values for config items

### DIFF
--- a/Makefile.autosetup
+++ b/Makefile.autosetup
@@ -278,7 +278,7 @@ ALLOBJS+=	$(LIBCOREOBJS)
 # libconfig
 LIBCONFIG=	libconfig.a
 LIBCONFIGOBJS=	config/address.o config/bool.o config/dump.o config/enum.o \
-		config/long.o config/mbtable.o config/number.o config/quad.o \
+		config/long.o config/mbtable.o config/number.o config/path.o config/quad.o \
 		config/regex.o config/set.o config/slist.o config/sort.o \
 		config/string.o config/subset.o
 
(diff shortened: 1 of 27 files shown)


core/mailbox.h
```diff
@@ -44,6 +44,7 @@ enum MailboxType
   MUTT_MAILBOX_ANY = -2,   ///< Match any Mailbox type
   MUTT_MAILBOX_ERROR = -1, ///< Error occurred examining Mailbox
   MUTT_UNKNOWN = 0,        ///< Mailbox wasn't recognised
+  MUTT_LOCAL_PATH,         ///< Local file/directory
```
gahr: I don't see this being used. Is it here for future use?

flatcap: Oh... It's still not used in the future (temporal grammar allowing).
There are lots of files used in the code and I wondered if the same machinery could be used to work with them.  (still undecided).  I'll drop it for now.

flatcap:
> How about a path type that embeds its is_dir info

Hmm... nice!  That should make things much tidier.

NilsIrl:
Because of this, default values, e.g. `folder` is set to `~/Mail` do not get expanded

Issue #1511, opened 2018-12-29
Pressing ctrl+c should close the command prompt


binaryplease:
Pressing `ctrl+c` should close the command prompt, as it does in vim.

@flatcap over on IRC:
`Ctrl-C is trapped by the terminal and turned into SIGINT, which neomutt traps and sets the variable SigInt  it _might_ be as simple as checking SigInt (and resetting it)
`

Steef435:
I did some digging.

mutt_enter_command (commands.c) calls mutt_get_field (curs_lib.c), which calls mutt_enter_string_full (enter.c), which calls km_dokey (keymap.c), which finally calls mutt_getch (curs_lib.c) and this is the function that calls mutt_query_exit on SigInt (and currently no way to configure this behavior).

The only "nice" way I can think of to implement this is to make it configurable by function argument. Together, the above functions are used more than 50 times in the code base, so for all those calls the desired behavior would have to be evaluated.

So yeah, it's as simple as checking SigInt, but it's buried so deeply in the code hierarchy that I think it will take some work (I'm not familiar enough to judge which behavior is desired for each call).

flatcap:
Nice investigating!

Before you start altering code, I'd just like to get a second opinion from the other developers.
We already have <kbd>Ctrl-G</kbd> (a scarcely documented abort key).

---

Yes, a parameter on `mutt_getch()` would work to distinguish the interruptible case (from normal keypresses).
There _are_ quite a lot of calls to `mutt_getch()`, but we only want to interrupt the full text-entry and it seems that they all come from one place `mutt_get_field_full()`:

![get-field](https://flatcap.org/mutt/get-field.svg)

binaryplease:
Sorry, have been gone for a while. I'm still interested in this and would be willing to help.
Does that mean that only `mutt_get_field_full()` would have to be modified?

flatcap:
> Sorry, have been gone for a while

No worries.

> Does that mean that only `mutt_get_field_full()` would have to be modified?

I _think_ so.

> how is this related to #1510

It's the same sort of thing.

@neomutt/reviewers 
For text-entry fields, what do we want to (continue) support(ing)?
- <kbd>Ctrl-G</kbd>
- <kbd>Ctrl-C</kbd>
- <kbd>Escape</kbd>

gahr:
I'd love to have them all

Steef435:
[...]

Issue #1510, opened 2018-12-29
Escape should close command prompt (vim-like behaviour)


binaryplease:
Similar to #1506, when pressing the escape key while in the command prompt, it should close as it does in vim. This is helpful if you change your mind when entering commands and what to exit the prompt without any changes.

gahr:
I tend to agree that I would expect escape to close my prompt too. Anyway, ctrl-g does that already.

Steef435:
Per discussion in #1511, I made a hacky (experimental) implementation of this in steef435/neomutt@97e2389 (this one does use a hard coded solution in mutt_getch, and abuses existing infrastructure in km_dokey for letting mutt_getch pass opcodes). While this seems to work on my end, the delay is longer than expected, probably because of the inotify polling which I don't fully understand.

flatcap:
That's very nice @steef435 

I see what you mean about the delay; I couldn't see where that was coming from (but I only had a quick look).  If you `bind` some other key to `cancel`, the response is instantaneous.

I built without `libinotify` (`./configure --disable-inotify`) and that's not the cause.
`libinotify` is watching the filesystem for changes, e.g. new files.

Please could you look into the delay.
The logging only has second-resolution, so you'll need to be creative!

steef435:
Ah, thank you!

Apparently curses internally delays escape presses for escape sequences. For ncurses there's the ESCDELAY (see man ESCDELAY) environment variable that influences this timeout. Setting it to 0 makes the Esc detection code work.

However, I then noticed that the general keybindings broke somewhere along the way. Pressing Esc in the default view shows a change folder prompt filled with "ancel>", so most likely, the name of OP_EDITOR_CANCEL gets thrown in the key buffer somewhere.. I intend to continue working on this but it will probably take at least a week until I have time again. (so, if someone feels up to the task..)

flatcap:
> ... ESCDELAY ...

Aha!

> ... the general keybindings broke somewhere along the way
> `ancel>`

Ah, that makes sense.  We need to distinguish text-entry from just waiting for a key.
I've suggested to @binaryplease to make that change.

> probably take at least a week until I have time again

no worries.

JJK96:
Any progress on this, <kbd>Ctrl</kbd>+<kbd>g</kbd> is a lot harder to hit for me

flatcap:
> Any progress on this?

Unfortunately, not.
We were hoping this would be simple, but it wasn't.

[...]

PR #2139, opened 2020-02-15
Add abort_key config option

ftilde:
* **What does this PR do?**
This PR adds the "abort_key" config option which can be used to change the key
used to abort prompts (previously unchangeable, now per default Ctrl-G). The
key can specified in the same format as bind/macro commands.

In contrast to previous work (see #1510) this PR replaces the code that previously explicitly checks for the keycode of Ctrl-G with checks against the keycode of the key specified by the newly introduced config option `abort_key`.  I propose this solution because I am under the impression that this abort behaviour is so low-level that it cannot be handled via an <abort> command that could be assigned using `bind`.

It should probably also be noted that this PR does not include any changes to the ncurses ESCDELAY (changable via environment variable or `set_escdelay`) to make using `Esc` as an abort key nicer to use. This is because the delay issue is (at least at its core) not related and it can be worked around by the user by starting neomutt using `ESCDELAY=0 neomutt`.

#1510, in principle also #1511, but this requires may more work, because Ctrl-C raises a SIGINT instead of being handled by the "normal" input system, I think.

Due to me not really being familiar with the code base, for some of these changes I was not sure where to best put them. (See for example mutt_init_abort_key which must happen after config parsing, in contrast to other keymap related initialization which apparently has to be performed before ncurses initialization.) Let me know if you have any thoughts on this.

### DIFF
--- a/keymap.c
+++ b/keymap.c
@@ -143,7 +143,8 @@ static struct Mapping KeyNames[] = {
   { NULL, 0 },
 };
 
-int LastKey; ///< contains the last key the user pressed
+int LastKey;        ///< contains the last key the user pressed
+keycode_t AbortKey; ///< code of key to abort prompts, normally Ctrl-G
 
 struct Keymap *Keymaps[MENU_MAX];
 
@@ -814,6 +815,48 @@ static const char *km_keyname(int c)
   else
     snprintf(buf, sizeof(buf), "\\x%hx", (unsigned short) c);
   return buf;
+}
+
+/**
+ * mutt_init_abort_key - Parse the abort_key config string
+ *
+ * Parse the string into C_AbortKeyStr and put the keycode into AbortKey.
+ */
+void mutt_init_abort_key(void)
+{
+  keycode_t buf[2];
+  size_t len = parsekeys(C_AbortKeyStr, buf, mutt_array_size(buf));
+  if (len == 0)
+  {
+    mutt_error(_("Abort key is not set, defaulting to Ctrl-G"));
+    AbortKey = ctrl('G');
+    return;
+  }
+  if (len > 1)
+  {
+    mutt_warning(
+        _("Specified abort key sequence (%s) will be truncated to first key"), C_AbortKeyStr);
+  }
+  AbortKey = buf[0];
+}
+
+/**
+ * mutt_abort_key_config_observer - Listen for abort_key config changes - Implements ::observer_t()
+ */
+int mutt_abort_key_config_observer(struct NotifyCallback *nc)
+{
+  if (!nc->event_data)
+    return -1;
+  if (nc->event_type != NT_CONFIG)
+    return 0;
+
+  struct EventConfig *ec = nc->event_data;
+
+  if (mutt_str_strcmp(ec->name, "abort_key") != 0)
+    return 0;
+
+  mutt_init_abort_key();
+  return 0;
 }
 
 /**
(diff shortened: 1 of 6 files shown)


ftilde:
I would also like to point out: This of course does not solve all problems related to binding `Esc` as an abort key. Doing so makes alt-keybindings problematic. However, this problem is not directly related to the abort key binding itself. For example, when using the following binding, both `alt-a` and first pressing `Esc` then `a` scroll the index up even with `ESCDELAY=0` in the environment:
```
bind  index          <Esc>a next-line
```
For whatever reason ncurses decided to hand the escape sequence for alt-key events directly to the user, although it doesn't do that for `Insert`, for example.

I still feel like the proposed change is a step in the right direction though and it enables users who don't care about Alt-X keybindings to use <Esc> as the abort key.

ftilde:
I have thought about this some more and added another note to the documentation of the binding option (explaining that the `abort_key` cannot be used in other bindings). This is already the case at the moment with `Ctrl-G`: Something like the following will have no effect when added to the config file:
```
bind  index          \CG next-line
```
WIth this in bind it should be no (bigger) surprise that bindings like `<Esc>a` break when you specify `<Esc>` as the abort key. If you are otherwise happy with the approach of this PR we could/should add a check for binding-abort_key conflicts just like it is already done for binding-binding conflicts.

LarsHaalck:
Coming from Vim, I would appreciate this and really like this approach. Considering that almost everything else can be configured freely, I think it makes total sense to add such a feature.

reedts:
I second this! The possibility to use ESC to abort would make life in neomutt much easier for any vim users!

flatcap:
I haven't forgotten you, I've just been a bit busy (and there aren't many developers).

flatcap:
Rebased and tidied the code

keymap.c
```diff
@@ -815,6 +816,29 @@ static const char *km_keyname(int c)
...
 
+/**
+ * mutt_init_abort_key - Parse the abort_key config string
+ *
+ * Parse the string into C_AbortKeyStr and put the keycode into AbortKey.
+ */
+void mutt_init_abort_key(void)
+{
+  keycode_t buf[2];
+  size_t len = parsekeys(C_AbortKeyStr, buf, sizeof(buf) / sizeof(buf[0]));
```
flatcap: We can simplify this with the `mutt_array_size()` macro.

keymap.c
```diff
@@ -1579,15 +1603,15 @@ void mutt_what_key(void)
 {
   int ch;
 
-  mutt_window_mvprintw(MuttMessageWindow, 0, 0, _("Enter keys (^G to abort): "));
+  mutt_window_mvprintw(MuttMessageWindow, 0, 0, _("Enter keys (press abort key to resume): "));
```
flatcap: It'd be nice to tell the user _which_ key that is.
See: `km_find_func()`, `km_expand_key()` in `mutt_body_handler()`

keymap.c
ftilde: Should be addressed in 21d22cd81d3493a3464f9240471432369999dc04

keymap.c
ftilde: Good point, I've changed it inbe1b66b3f725cef6f31c5a9d1e23451781d4dfa0.

ftilde:
Oh man, sorry about that. I totally forgot that you can change configuration values later somehow. I have added an observer for the `abort_key` config value in 4416ca7cc3dcd4389d6efca59f045a5e4024e888.

mutt_config.c
```diff
@@ -87,6 +87,20 @@ bool C_IgnoreLinearWhiteSpace = false;
...
+  ** Specifies the key that can be used to abort prompts.  The format is the
+  ** same as used in "bind" commands.  The default is equivalient to "\G".
+  ** Note that the specified key should not be used in other bindings, as the
+  ** abort operation has higher precedence and the binding will not have the
+  ** desired effect.
+  ** .pp
+  ** Example:
+  ** .ts
+  ** set abort_key = "<Esc>"
+  ** .te
```
flatcap: Please can you add a short note about `ESCDELAY=0`.
What it does, why it's a good idea.

ftilde: Sure thing, see 92e39177c57464c1d51d3adbe8380e9ed35287dd.

PR #788, opened 2017-09-26
imap: retrieve bodystructure when mailbox is opened

sileht:
* **What does this PR do?**

Currently %X in index_format is unusable because it downloads
all messages.

This change implements BODYSTRUCTURE parsing from rfc3501.

The message content is set with this structure avoiding
mutt_parse_mime_message() to call mx_open_message for imap.

%X is no more need to download message to be calculated.

* **Are there points in the code the reviewer needs to double check?**

everything 

* **Does this PR meet the acceptance criteria?**

   - [] All builds are passing

Almost the change depends of some code move from the imap flags PR to compile.

   - [] Code follows the [style guide](https://www.neomutt.org/dev/coding-style)

Not yet checked, my clang-format is sick.

* **What are the relevant issue numbers?**

Main parts of #786

Issue #786, opened 2017-09-26
imap should use fetch BODYSTRUCTURE 


sileht:
* Why do you not like the current state?

When you open the pager all attachements are downloaded when usually you just need the first text/plain attachment.

%X in index_format in unusable with IMAP, because it downloads open all attachements of all messages just to get the number of attachements.

* What would you like to see to change?

I would like to use BODYSTRUCTURE and BODY.PEEK[<ID>]. The idea to fill the Body objects tree
without the body->content set. And only set body->content on demand.
"On demand" means: loading the part needed by the pager when you open it, or lading an attachment when you open it in the attachment browser.

NOTE: I already have written the BODYSTRUCTURE parser according the rfc3501 my working branch is here: https://github.com/sileht/neomutt/tree/devel/bodystruct.
It's far to be ready, most of the structure is not kept in memory for now, but the parser have successfully parsed all my mails. I have done that as initial step to ensure I will be able to propose a PR.

gahr:
I would love to see this done, thanks for working on this!

sileht:
I just got %X working this morning :)

scottkosty:
Related to #2004.

PR #2149, opened 2020-02-24
imap: retrieve bodystructure when mailbox is opened

flatcap:
Currently `%X` in `$index_format` is unusable because it downloads all messages.

This change implements BODYSTRUCTURE parsing from [RFC3501](https://tools.ietf.org/html/rfc3501).

The message content is set with this structure so `mutt_parse_mime_message()` doesn't need to call `mx_msg_open()` for imap.

`%X` no longer needs to download message to be calculated.

Implements: #786 

<sub>(This PR replaces #788, in order to move the branch into the neomutt repo.)</sub>


flatcap:
Archived to: https://github.com/neomutt/neomutt-old/tree/bodystruct

Issue #2148, opened 2020-02-24
Segfault in header cache compression


gahr:
Sorry I failed to see this when reviewing #1922 some months ago.

The compression functions return NULL if the input buffer is too short to be (effectively) compressed ([lz4](https://github.com/neomutt/neomutt/blob/b49866857ed56b536265da80fde69d4bd192e110/hcache/compr_lz4.c#L61-L62), [zlib](https://github.com/neomutt/neomutt/blob/b49866857ed56b536265da80fde69d4bd192e110/hcache/compr_zlib.c#L62-L63), [zstd](https://github.com/neomutt/neomutt/blob/b49866857ed56b536265da80fde69d4bd192e110/hcache/compr_zstd.c#L94-L95)).

Caller code [does not check](https://github.com/neomutt/neomutt/blob/b49866857ed56b536265da80fde69d4bd192e110/hcache/hcache.c#L557) the return value, and feeds it blindly into the header cache backend, causing a segfault.

This happens at every mailbox open, when storing the mailbox' UIDVALIDITY (a 4-bytes integer).

I remember testing the header cache compression mechanism while back and noticing a huge decrease in on-disk size. I wonder how we missed this.

@mcmilk what do you suggest?

Edit: fixed link to pull req.

mcmilk:
Maybe I had some more variables for this in my first pull requests ... but then we have optimized the things ;-)

I will create a small bug fix for it ...

PR #2152, opened 2020-02-25
Fix issue #2148 and check for valid compressed data from compression

mcmilk:
* **What does this PR do?**

These small changes should fix issue #2148

Please check, if my assumption is correct and maybe add this small fix.

### DIFF
--- a/hcache/compr_zstd.c
+++ b/hcache/compr_zstd.c
@@ -98,7 +98,7 @@ static void *compr_zstd_compress(void *cctx, const char *data, size_t dlen, size
 
   size_t ret;
   size_t len = ZSTD_compressBound(dlen);
-  mutt_mem_realloc(&ctx->buf, len);
+  mutt_mem_realloc_msg(&ctx->buf, len, _("Out of memory, please use smaller Levels for Zstandard or take LZ4."));
 
   if (ctx->cdict)
     ret = ZSTD_compress_usingCDict(ctx->cctx, ctx->buf, len, data, dlen, ctx->cdict);
@@ -120,13 +120,19 @@ static void *compr_zstd_decompress(void *cctx, const char *cbuf, size_t clen)
 {
   struct ComprZstdCtx *ctx = cctx;
 
-  if (!cctx || clen < 8)
+  if (!cctx)
     return NULL;
 
-  size_t ret;
-  size_t len = ZSTD_getFrameContentSize(cbuf, clen);
+  unsigned long long len = ZSTD_getFrameContentSize(cbuf, clen);
+  if (len == ZSTD_CONTENTSIZE_UNKNOWN)
+    return NULL;
+  else if (len == ZSTD_CONTENTSIZE_ERROR)
+    return NULL;
+  else if (len == 0)
+    return NULL;
   mutt_mem_realloc(&ctx->buf, len);
 
+  size_t ret;
   if (ctx->ddict)
     ret = ZSTD_decompress_usingDDict(ctx->dctx, ctx->buf, len, cbuf, clen, ctx->ddict);
   else
--- a/hcache/compr_zstd.c
+++ b/hcache/compr_zstd.c
@@ -91,7 +91,7 @@ static void *compr_zstd_open(void)
  */
 static void *compr_zstd_compress(void *cctx, const char *data, size_t dlen, size_t *clen)
 {
-  if (!cctx || (dlen < 10))
+  if (!cctx)
     return NULL;
 
   struct ComprZstdCtx *ctx = cctx;
@@ -106,7 +106,7 @@ static void *compr_zstd_compress(void *cctx, const char *data, size_t dlen, size
     ret = ZSTD_compressCCtx(ctx->cctx, ctx->buf, len, data, dlen, C_HeaderCacheCompressLevel);
 
   if (ZSTD_isError(ret))
-    mutt_error("ZSTD_compress() failed!");
+    return NULL;
 
   *clen = ret;
 
@@ -120,6 +120,9 @@ static void *compr_zstd_decompress(void *cctx, const char *cbuf, size_t clen)
 {
   struct ComprZstdCtx *ctx = cctx;
 
+  if (!cctx || clen < 8)
+    return NULL;
+
   size_t ret;
   size_t len = ZSTD_getFrameContentSize(cbuf, clen);
   mutt_mem_realloc(&ctx->buf, len);
@@ -130,7 +133,7 @@ static void *compr_zstd_decompress(void *cctx, const char *cbuf, size_t clen)
     ret = ZSTD_decompressDCtx(ctx->dctx, ctx->buf, len, cbuf, clen);
 
   if (ZSTD_isError(ret))
-    mutt_error("ZSTD_decompress() failed!");
+    return NULL;
 
   return ctx->buf;
 }
--- a/hcache/compr_lz4.c
+++ b/hcache/compr_lz4.c
@@ -58,7 +58,7 @@ static void *compr_lz4_open(void)
  */
 static void *compr_lz4_compress(void *cctx, const char *data, size_t dlen, size_t *clen)
 {
-  if (!cctx || (dlen < 10))
+  if (!cctx)
     return NULL;
 
   struct ComprLz4Ctx *ctx = cctx;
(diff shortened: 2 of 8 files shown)


flatcap:
If you squash your commits then `git push origin --force`, github will update itself.

mcmilk:
> If you squash your commits then `git push origin --force`, github will update itself.

Thanks for the hint. Do you think the fix is okay ?

flatcap:
> Do you think the fix is okay ?

I think so -- it's what I expected to see
(but it was @gahr that managed to trigger the bug)

flatcap:
Update: I can repeat the original crash and this fixes the problem for me.

gahr:
Thanks for the quick reaction, @mcmilk, however:

Thought: so we fall back to storing the uncompressed blob whenever the compression routine fails (for whatever reason).
Question: what happens on read, when we pass this blob we previously stored to the decompression routine in `mutt_hcache_fetch_raw`? Do we also assume the decompression routine handles both compressed and uncompressed blobs just fine?
Answer: we currently don't, and the result is a failure in decompression and a crash when trying to dereference the value returned from `mutt_hcache_fetch_raw`.

```
Zlib uncompress() failed!

(gdb) bt
#0  0x000000000075dd60 in imap_mdata_new (adata=0x60e0000009e0, name=0x60400000e6ef "INBOX") at imap/util.c:199
#1  0x000000000073f9b7 in imap_ac_add (a=0x607000001b40, m=0x612000001fc0) at imap/imap.c:1914
#2  0x00000000004e524b in mx_ac_add (a=0x607000001b40, m=0x612000001fc0) at mx.c:1673
#3  0x00000000004e347c in mx_mbox_open (m=0x612000001fc0, flags=0 '\000') at mx.c:285
#4  0x000000000048f10e in main (argc=1, argv=0x7fffffffe9c0, envp=0x7fffffffe9d0) at main.c:1149
(gdb) 
```

Question: how did you manage to make this work with any IMAP connection? I crash as soon as I try to fetch UIDVALIDITY.

gahr:
Also, because the new code will segfault trying to read existing cache databases, please bump our header cache [base version](https://github.com/neomutt/neomutt/blob/master/hcache/hcachever.sh#L3).

flatcap:
Dropping the `(dlen < 10)` checks will probably do the trick.

There are four places, outside of the hcache code, where `mutt_hcache_store_raw()` is called.
Each stores a short string, but there's only one value of each per account.
Removing the minimum size check will make these values a bit bigger, but all the other callers will have lots of data that _will_ compress.

mcmilk:
Hello again,

I didn't notice, that mutt_hcache_store_raw() is used in other places than header caching ... sorry for that :-)

The using is just as this:
- nntp/newsrc.c with fixed size of 5
- imap/util.c with fixed size of 10
- imap/message.c with fixed sizes of 7, 8 and 12
- and then of cause the hcache interface

Solution would be, compress only for things larges than e.g. 32 bytes ...

Check also the decompression stuff, whenever it fails to decompress things, the user may have an uncompressed hcache entry or also some uncompressable blob ... if so, then the decompression routines (zlib, lz4 and also zstd) are safe for invalid input.... and they will return some error ... this error path should always return in simple returning the uncompressed data blob ...

I will check this now, and will updated my pull request...

gahr:
I wouldn’t rely on a failure in the decompression routine to sort a compressed blob from an uncompressed one, especially when we don’t store only text but binary data as well. I would hate to see a raw blob being uncompressed into some weird number that causes a difficult to debug segfault later on. I suggest we always compress when asked to do so.

flatcap:
> I suggest we always compress when asked to do so

Same here.  Compress always and don't worry about the few cases that are inefficient.

flatcap:
I've added some verbosity:
- Log line: Header cache will use zstd compression
- Version info: hcache compression: lz4, zlib, zstd

mcmilk:
> I wouldn’t rely on a failure in the decompression routine to sort a compressed blob from an uncompressed one, especially when we don’t store only text but binary data as well. I would hate to see a raw blob being uncompressed into some weird number that causes a difficult to debug segfault later on. I suggest we always compress when asked to do so.

Compressed / uncompressed blobs are sorted by the other hashing in line 284, here: `mutt_buffer_printf(name, "%s|%s%s", hcache_get_ops()->name, folder, cm ? cm : "");`

You may decrease my 32 bytes to 16 bytes ... so this way the compression overhead is done only for these bytes ... but if you compress all values, then the compression will make the header cache databases even bigger then smaller... which is not what we want. I add 4 bytes for the size, and then some more bytes are added by each different compression method.

mcmilk:
> I wouldn’t rely on a failure in the decompression routine to sort a compressed blob from an uncompressed one, especially when we don’t store only text but binary data as well. I would hate to see a raw blob being uncompressed into some weird number that causes a difficult to debug segfault later on. I suggest we always compress when asked to do so.

When compression fails, then the current code just writes the uncompressed stuff ... so the decompression of that will also fail... which then, also returns the uncompressed again... which is what we want.

We could also to some failing to compress/uncompress messages...but I don't see the benefits of this for the user.

flatcap:
> When compression fails, then the current code just writes the uncompressed stuff

Agreed

> so the decompression of that will also fail

Probably, but that's not _guaranteed_.

1. The compression code doesn't add a header to the compressed data; it's just raw bytes.
2. Some of the short strings we store are binary

Imagine that we store some binary data: `7F 9E DB 7F`, perhaps the hash of some object.
Unfortunately in this case, these random bytes are **also** valid compressed data.

We stored it **un**compressed, but now the uncompress succeeds, giving us some other binary data: `2D A1 F7 43`

This **is** unlikely, but it **is** possible.
It's the sort of thing that would cause rare and near-impossible-to-find bugs.

---

99.9% (unverified :-) of the data we store in the header cache will be large headers which will get compressed, saving space.

0.1% of the data will be short data: index numbers, etc.  Forcing these to be compressed will waste a few bytes.

If every row in the database is compressed, then we don't have to worry about freak bugs.

---

In the future, perhaps we could add a separate function for storing small data.
For now, it's not worth it.  There's too little to worry about.

mcmilk:
> > When compression fails, then the current code just writes the uncompressed stuff
> 
> Agreed
> 
> > so the decompression of that will also fail
> 
> Probably, but that's not _guaranteed_.

Not _guaranteed_ is correct here :-/

Shall I create a small mini header for the decompression like this now ?

1. two bytes header compression magic: 0x20bb
2. one byte for the used compression method (0x00=none; 0x01=lz4; 0x02=zlib; 0x03=zstd)

This adds three bytes meta data, but we have some filter for valid values for magic + methods... all unknown bytes should result in some error fatal error message.

Is this okay ?

gahr:
This is being a bit overengineered in my opinion. Also, how do you make sure that uncompressed data doesn't begin with your magic header, by accident?

flatcap:
Yes, that would only work if the header was added to **all** data and `0x00` meant uncompressed.

Thanks for coming up with all these ideas, @mcmilk ; it's good to consider lots of possibilities.

However, as there are so few short fields stored, please can we go back to the earlier suggestions:
- Drop the length checks
- Compress everything
- Don't worry about a small inefficiency

If, in the future, we find that we're storing more small fields, we can always change our minds.

Thanks.

mcmilk:
> * Drop the length checks
> * Compress everything
> * Don't worry about a small inefficiency

All three suggestions are done ... I rebased the changes into one commit and force pushed this then. Is this the right way to do this ?

hcache/compr_zstd.c
```diff
@@ -120,6 +120,9 @@ static void *compr_zstd_decompress(void *cctx, const char *cbuf, size_t clen)
 {
   struct ComprZstdCtx *ctx = cctx;
 
+  if (!cctx || clen < 8)
```
gahr: Why do you check for `clen < 8`? How can you even get there?

gahr:
I see that the header cache is still falling back to storing the uncompressed data if compression fails / returning the raw contents if decompression fails. This will trigger crashes. Can we please simplify the logic so that when a compression method is selected, it's all or nothing? If you fail to compress, you return failure. If you fail to uncompress, you return failure.

Also, in `mutt_hcache_free`, you are passing the compressed data `on_disk` to the header cache backend free function, but you fail to free the uncompressed data returned by the various `decompress` functions. Are you leaking memory?

hcache/compr_zstd.c
mcmilk: Overseen ... also the lz4 <0 thing.

mcmilk:
> I see that the header cache is still falling back to storing the uncompressed data if compression fails / returning the raw contents if decompression fails. This will trigger crashes. Can we please simplify the logic so that when a compression method is selected, it's all or nothing? If you fail to compress, you return failure. If you fail to uncompress, you return failure.
> 
> Also, in `mutt_hcache_free`, you are passing the compressed data `on_disk` to the header cache backend free function, but you fail to free the uncompressed data returned by the various `decompress` functions. Are you leaking memory?

No, in my test I was not leaking memory. The thing uses realloc() for it's one buffer ... and in the end it's freed.

gahr:
> When for some reason the data can not be compressed ... it's written without compression. I am fine with this

I am not. As already mentioned, there is no way to figure a compressed from an uncompressed blob. This will lead to crashes. Please do as asked by flatcap in https://github.com/neomutt/neomutt/pull/2152#issuecomment-593892219 and do not overengineer the thing. It is a cache after all, best effort by design.

flatcap:
- rebase onto master
- edit commit messages to add blank line after subject
  (`git log` was getting upset)

flatcap:
Your changes were adapted, extended and merged into 4c6dbe83a3.

mcmilk:
Thank you :-)

PR #2153, opened 2020-02-26
add 'arrow_string' config variable

jindraj:
* **What does this PR do?**
Adds new config option `arrow_string` which changes default `->` arrow_cursor

     for that)

        Actually I haven't found anything even about arrow_curor in manual.xml.head.

no need for that

### DIFF
--- a/menu.c
+++ b/menu.c
@@ -337,7 +337,7 @@ static void menu_make_entry(char *buf, size_t buflen, struct Menu *menu, int i)
 static void menu_pad_string(struct Menu *menu, char *buf, size_t buflen)
 {
   char *scratch = mutt_str_strdup(buf);
-  int shift = C_ArrowCursor ? 3 : 0;
+  int shift = C_ArrowCursor ? mutt_strwidth(C_ArrowString) + 1 : 0;
   int cols = menu->win_index->state.cols - shift;
 
   mutt_simple_format(buf, buflen, cols, cols, JUSTIFY_LEFT, ' ', scratch,
@@ -433,7 +433,7 @@ void menu_redraw_index(struct Menu *menu)
         mutt_curses_set_color(MT_COLOR_INDICATOR);
         if (C_ArrowCursor)
         {
-          mutt_window_addstr("->");
+          mutt_window_addstr(C_ArrowString);
           mutt_curses_set_attr(attr);
           mutt_window_addch(' ');
         }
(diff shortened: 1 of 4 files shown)


gahr:
Why don't we make arrow_string empty-able so when it's empty is disabled and we deprecate arrow_cursor?

jindraj:
> Why don't we make arrow_string empty-able so when it's empty is disabled and we deprecate arrow_cursor?

Or use arrow_cursor for the arrow_string:

 * arrow_cursor is off
```
set arrow_cursor = no
set arrow_cursor = ""
unset arrow_cursor
```

 * arrow_cursor is on and set to default
```
set arrow_cursor
arrow_cursor = yes
```
 * arrow_cursor is on and set to custom string (example)
```
arrow_cursor = ">"
```

deprecating arrow_cursor completely would break mutt compatibility. I guess we want users coming with mutt configs to have working config file as much as possible.

But is it possible to mix string and yes/no config option?

gahr:
If we can, that would be best, but I'm not sure @flatcap 's config machinery can handle that.

menu.c
```diff
@@ -479,8 +482,10 @@ void menu_redraw_motion(struct Menu *menu)
 
   if (C_ArrowCursor)
   {
-    /* clear the pointer */
-    mutt_window_addstr("  ");
+    /* clear the arrow */
+    size_t b = mutt_strwidth(C_ArrowString);
+    for (int p = 0; p <= b; p++) //mutt_window_addstr() would be more pleasant here
```
jindraj: I can change `mutt_window_addch()` and entire for loop to:
```
mutt_window_printf("%*s", b+1, "");
```

jindraj: or even save spare variable which i didn't know how to name…
```
mutt_window_printf("%*s", mutt_strwidth(C_ArrowString) + 1, "");
```

flatcap:
> But is it possible to mix string and yes/no config option?

So the values would be: `yes`, `no`, **empty**, or **any other string**.

> I'm not sure @flatcap 's config machinery can handle that.

Wonderful as the config code is, no it can't handle that.
You _could_ create a validator function, but what would it check?  (if one of the possibilities is a free-form string)

menu.c
flatcap: That's nice use of `printf()` however not everyone will be familiar with the syntax.
I suggest a comment:
```
// Print space chars to match the screen width of C_ArrowString
```

jindraj:
> You _could_ create a validator function, but what would it check? (if one of the possibilities is a free-form string)
I'm not sure how would be possible to resolve situation like `set arrow_cursor`
which is the same as `set arrow_cursor=yes`.  That probably requires option to be boolean. maybe the validation function would be better for this than free-form string.

flatcap:
> I'm not sure how would be possible to resolve situation like `set arrow_cursor`

Yeah, it would be tricky.

So for now, two separate config variables.

PR #2162, opened 2020-03-03
add / tidy debug code

flatcap:
- `configure --debug-backtrace`
  Rename the configure option for consistency

- `configure --debug-parse-test`
  Add configure option
  Wrap all code in `#ifdef`s

- `configure --debug-graphviz`
  Add GraphViz dumps for Account/Mailbox/etc
  Code is triggered by:
  - <kbd>M</kbd> (`show-log-messages`)
  - Batch mode (`neomutt -B`)
  - Seg-fault

- `configure --debug-notify`
  Dump all notifications

- `configure --debug-window`
  Dump lots of info about the nested Windows

Other developers might benefit from this code (and I'm tired of rebasing it).
All the code is built conditionally.


auto.def
```diff
@@ -113,6 +110,13 @@ options {
   pkgconf=0                 => "Use pkg-config during configure"
 # Enable all options
   everything=0              => "Enable all options"
+# Debug options
+  debug-backtrace=0         => "DEBUG: Enable backtrace support with libunwind"
+  with-backtrace:path       => "Location of libunwind"
+  debug-graphviz=0          => "DEBUG: Enable Graphviz dump"
+  debug-notify=0            => "DEBUG: Enable Notifications dump"
+  debug-parse-test=0        => "DEBUG: Enable 'neomutt -T' for config testing"
+  debug-window=0            => "DEBUG: Enable Windows dump"
```
gahr: I would spell `windows` without the upper case `W`, to avoid triggering unwanted attentions ;)

auto.def
```diff
@@ -123,7 +127,7 @@ options {
 if {1} {
   # Keep sorted, please.
   foreach opt {
-    autocrypt backtrace bdb coverage doc everything fmemopen full-doc gdbm
+    autocrypt bdb coverage debug-backtrace debug-graphviz debug-notify debug-parse-test debug-window doc everything fmemopen full-doc gdbm
```
gahr: I'd love to keep the line lengths reasonable

debug/graphviz.c
```diff
@@ -0,0 +1,907 @@
...
+#include "notmuch/lib.h"
+
+// #define GV_HIDE_CONTEXT
+#define GV_HIDE_CONTEXT_CONTENTS
+// #define GV_HIDE_MBOX
+// #define GV_HIDE_NEOMUTT
+// #define GV_HIDE_CONFIG
+// #define GV_HIDE_MDATA
+
+void dot_type_bool(FILE *fp, const char *name, bool val)
```
gahr: Can we please make all internal functions static?

debug/lib.h
```diff
@@ -28,19 +28,52 @@
...
+const char *get_mailbox_event    (int id);
+const char *get_mailbox_type     (enum MailboxType type);
+void        notify_dump_account  (struct NotifyCallback *nc);
+void        notify_dump_color    (struct NotifyCallback *nc);
+void        notify_dump_command  (struct NotifyCallback *nc);
+void        notify_dump_config   (struct NotifyCallback *nc);
+void        notify_dump_context  (struct NotifyCallback *nc);
+void        notify_dump_email    (struct NotifyCallback *nc);
+void        notify_dump_global   (struct NotifyCallback *nc);
+void        notify_dump_mailbox  (struct NotifyCallback *nc);
```
gahr: Don't you want to split these into dedicated headers, e.g. debug/notify.h, and guard the inclusion?

gahr:
Just a couple of notes about the interaction with the rest of the system; I am not reviewing the code under `debug/*.c`.

auto.def
flatcap: Done

debug/graphviz.c
flatcap: Done

debug/lib.h
flatcap: I've stripped them out.  I'll put them in a separate file if they're needed individually.

auto.def
flatcap: Folded

flatcap:
> I am not reviewing the code under `debug/*.c`

Probably wise.  There be dragons! :-)

PR #2179, opened 2020-03-15
fix emptying trash

flatcap:
Use `mx_mbox_find()` to lookup `$trash` because it can compare Mailbox paths more carefully.

Recent changes have made it more likely that an IMAP Mailbox will include the `user@` prefix in its path.
This mean that some comparisons againts `$trash` were failing, so it was impossible to empty the trash.

### DIFF
--- a/mx.c
+++ b/mx.c
@@ -792,7 +792,8 @@ int mx_mbox_close(struct Context **ptr)
   }
 
   /* copy mails to the trash before expunging */
-  if (purge && (m->msg_deleted != 0) && (mutt_str_strcmp(mailbox_path(m), C_Trash) != 0))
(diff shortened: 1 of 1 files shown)


mx.c
```diff
@@ -792,7 +792,8 @@ int mx_mbox_close(struct Context **ptr)
   }
 
   /* copy mails to the trash before expunging */
-  if (purge && (m->msg_deleted != 0) && (mutt_str_strcmp(mailbox_path(m), C_Trash) != 0))
+  struct Mailbox *m_trash = mx_mbox_find(m->account, C_Trash);
```
gahr: perhaps `const`, but LGTM

flatcap: done

PR #2184, opened 2020-03-16
Demystify NeoMutt

flatcap:
### Rename `magic` to `type`

No functional changes.
The members don't store a magic number, they store an `enum MailboxType`

Renames struct members:
- `Account.magic`
- `Mailbox.magic`
- `MxOps.magic`
- `NmEmailData.magic`

plus all the associated function parameters and local variables.


gahr:
[this](https://github.com/neomutt/neomutt/blob/master/doc/manual.xml.head#L6659), too, perhaps?

PR #2195, opened 2020-03-18
fix: COLUMNS env var

flatcap:
When first opening NeoMutt, the pager Window might not have been sized.
This lead to `$COLUMNS` being set to `0`, causing mailcap problems.

---

To test:
- NeoMutt config: `auto_view text/html`
- Mailcap: `text/html; elinks -dump -dump-width ${COLUMNS:-80} %s; nametemplate=%s.html; copiousoutput`
- Start NeoMutt and immediately open an email with HTML part

Result:
- `elinks` fails because `$COLUMNS` is set, but is `0`

### DIFF
--- a/commands.c
+++ b/commands.c
@@ -211,7 +211,8 @@ int mutt_display_message(struct MuttWindow *win_index, struct MuttWindow *win_ib
   mutt_message_hook(m, e, MUTT_MESSAGE_HOOK);
 
   char columns[16];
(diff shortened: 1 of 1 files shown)


gahr:
Worst case the index window will be more narrow than the pager window (becase of the sidebar), so elinks will only populate part of the screen, right?

flatcap:
For now, the index and pager will always be the same width.
This'll need fixing if we create a widescreen mode (side-by-side).

PR #2196, opened 2020-03-18
use mutt_file_fopen() on config variables

flatcap:
Config strings can be NULL, so use mutt_file_fopen() for safety.


gahr:
LGTM, except I don't think `mutt_perror("fopen");` would be correct in this case.

flatcap:
Good point.  The other's don't log, so I've dropped the `perror()`.

Issue #2066, opened 2020-01-10
Upgrade charset config to DT_SLIST


flatcap:
The 6 'charset' config variables in NeoMutt are colon-separated strings.
They're currently `DT_STRING` (just a plain string); they need upgrading to `DT_SLIST`.

- `assumed_charset`
- `attach_charset`
- `send_charset`
- ~`charset`~
- ~`config_charset`~
- ~`newsgroups_charset`~

Converting them to `struct Slist` means that the parsing will be done for us and the charset code can be simplified to use `STAILQ_FOREACH`.

`smtp_authenticators` can be used as a template.

**Update**: Three of them can only be single values

PR #2197, opened 2020-03-19
use SLIST for charsets

andir:
* **What does this PR do?**

A modest attempt at #2066

### DIFF
--- a/email/rfc2047.c
+++ b/email/rfc2047.c
@@ -426,7 +426,7 @@ static char *decode_word(const char *s, size_t len, enum ContentEncoding enc)
  * @retval 0 Success
  */
 static int encode(const char *d, size_t dlen, int col, const char *fromcode,
-                  const char *charsets, char **e, size_t *elen, const char *specials)
+                  const struct Slist *charsets, char **e, size_t *elen, const char *specials)
 {
   int rc = 0;
   char *buf = NULL;
@@ -624,18 +624,20 @@ static int encode(const char *d, size_t dlen, int col, const char *fromcode,
  * @param[in]     col      Starting index in string
  * @param[in]     charsets List of charsets to choose from
  */
-void rfc2047_encode(char **pd, const char *specials, int col, const char *charsets)
+void rfc2047_encode(char **pd, const char *specials, int col, const struct Slist *charsets)
 {
   if (!C_Charset || !pd || !*pd)
     return;
 
+  struct Slist *fallback_charsets = NULL;
   if (!charsets)
-    charsets = "utf-8";
+    charsets = fallback_charsets = slist_parse("utf-8", SLIST_SEP_COLON);
 
   char *e = NULL;
   size_t elen = 0;
   encode(*pd, strlen(*pd), col, C_Charset, charsets, &e, &elen, specials);
 
+  slist_free(&fallback_charsets);
   FREE(pd);
   *pd = e;
 }
--- a/init.c
+++ b/init.c
@@ -40,7 +40,6 @@
 #include <unistd.h>
 #include "mutt/lib.h"
 #include "address/lib.h"
-#include "compress/lib.h"
 #include "config/lib.h"
 #include "email/lib.h"
 #include "core/lib.h"
@@ -63,6 +62,7 @@
 #include "options.h"
 #include "protos.h"
 #include "sort.h"
+#include "compress/lib.h"
 #include "history/lib.h"
 #include "store/lib.h"
 #ifdef USE_HCACHE
@@ -1571,9 +1571,9 @@ int charset_validator(const struct ConfigSet *cs, const struct ConfigDef *cdef,
   if (value == 0)
     return CSR_SUCCESS;
 
-  const char *str = (const char *) value;
+  const struct Slist *list = (const struct Slist *) value;
 
-  if ((strcmp(cdef->name, "charset") == 0) && strchr(str, ':'))
+  if ((strcmp(cdef->name, "charset") == 0) && (list->count > 1))
   {
     mutt_buffer_printf(
         err, _("'charset' must contain exactly one character set name"));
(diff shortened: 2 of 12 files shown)


flatcap:
That's a great start, and exactly what we're looking for.

There are quite a few `FREE()`s that need to become `slist_free()`,
and one line (sendlib.c:1072) is a bit complicated now.

```c
char *fchs = b->use_disp ? (C_AttachCharset ? (STAILQ_FIRST(&C_AttachCharset->head) ? STAILQ_FIRST(&C_AttachCharset->head)->data : C_Charset) : C_Charset) : C_Charset;
```

Perhaps it can be simplified, or perhaps it needs splitting up so humans can understand it.

andir:
> That's a great start, and exactly what we're looking for.
> 
> There are quite a few `FREE()`s that need to become `slist_free()`,

Those should be done now. There was also one case where I didn't actually use the newly constructed list.

> and one line (sendlib.c:1072) is a bit complicated now.
> 
> ```c
> char *fchs = b->use_disp ? (C_AttachCharset ? (STAILQ_FIRST(&C_AttachCharset->head) ? STAILQ_FIRST(&C_AttachCharset->head)->data : C_Charset) : C_Charset) : C_Charset;
> ```
> 
> Perhaps it can be simplified, or perhaps it needs splitting up so humans can understand it.

Yeah, I thought about that while writing the code. I think this version would probably be most(?) readable?
```c
    char *fchs = C_Charset;

    if (b->use_disp)
    {
      if (C_AttachCharset && (STAILQ_FIRST(&C_AttachCharset->head)))
      {
        fchs = STAILQ_FIRST(&C_AttachCharset->head)->data;
      }
    }
```
We could flatten the nested if-statements but my intuition is that this is more obvious and easier to grasp. Opinions?

flatcap:
Yeah, the `if`s are much easier to read.

flatcap:
I'm getting a warning:
```
sendlib.c: In function ‘mutt_get_content_info’:
sendlib.c:1081:51: error: pointer type mismatch in conditional expression [-Werror]
 1081 |         (convert_file_from_to(fp, fchs, chs ? chs : C_SendCharset, &fromcode,
      |                                                   ^
```

also, please can you rebase your branch onto master,
and (if you have it) run `clang-format` on your changes.

Thanks.

flatcap:
I've got some ideas for simplification...

- change the `$charset` validator to `single_charset_validator()`
  It's already flagged `DT_NOT_EMPTY`, so we don't need a special case for it

- change `charset_validator()` to use `single_charset_validator()`
  reduce the duplication of code

email/rfc2047.c
```diff
@@ -624,18 +624,21 @@ static int encode(const char *d, size_t dlen, int col, const char *fromcode,
...
 
   if (!charsets)
-    charsets = "utf-8";
+    charsets = fallback_charsets = slist_parse("utf-8", SLIST_SEP_COLON);
 
   char *e = NULL;
   size_t elen = 0;
   encode(*pd, strlen(*pd), col, C_Charset, charsets, &e, &elen, specials);
 
+  FREE(&fallback_charsets);
```
flatcap: `slist_free()`

init.c
```diff
@@ -1569,9 +1569,9 @@ int charset_validator(const struct ConfigSet *cs, const struct ConfigDef *cdef,
   if (value == 0)
     return CSR_SUCCESS;
 
-  const char *str = (const char *) value;
+  const struct Slist *list = (const struct Slist *) value;
 
-  if ((strcmp(cdef->name, "charset") == 0) && strchr(str, ':'))
+  if ((strcmp(cdef->name, "charset") == 0) && list->count > 1)
```
flatcap:
```suggestion
  if ((strcmp(cdef->name, "charset") == 0) && (list->count > 1))
```
Personal taste :-)

mutt/charset.c
```diff
@@ -303,25 +304,22 @@ int mutt_ch_convert_nonmime_string(char **ps)
...
     if (n == 0)
       return 0;
-    char *fromcode = mutt_mem_malloc(n + 1);
-    mutt_str_strfcpy(fromcode, c, n + 1);
     char *s = mutt_str_substr_dup(u, u + ulen);
-    int m = mutt_ch_convert_string(&s, fromcode, C_Charset, 0);
-    FREE(&fromcode);
+    int m = mutt_ch_convert_string(&s, c, C_Charset, 0);
     FREE(&s);
     if (m == 0)
```
flatcap: While you're altering this function, please can you change:
- `m` to `rc` (return code) for consistency
- `c` to `from` (maybe)
- `s` to `str` (maybe)

so that when the reader gets to `int m = mutt_ch_convert_string(&s, c, C_Charset, 0);`
they have an idea what's happening.

mutt/charset.c
```diff
@@ -1019,42 +1016,36 @@ void mutt_ch_set_charset(const char *charset)
...
 
-  for (const char *p = charsets; p; p = q ? q + 1 : 0)
+  STAILQ_FOREACH(np, &charsets->head, entries)
   {
-    q = strchr(p, ':');
-
-    size_t n = q ? q - p : strlen(p);
+    char *p = (char *) np->data;
+    size_t n = strlen(p);
     if (n == 0)
```
flatcap: again, not your fault, but can we have some more descriptive variable names, rather than `p` and `n`

andir:
Thanks for the feedback. It should all be addressed.

andir:
> I'm getting a warning:
> 
> ```
> sendlib.c: In function ‘mutt_get_content_info’:
> sendlib.c:1081:51: error: pointer type mismatch in conditional expression [-Werror]
>  1081 |         (convert_file_from_to(fp, fchs, chs ? chs : C_SendCharset, &fromcode,
>       |                                                   ^
> ```
> 
> also, please can you rebase your branch onto master,
> and (if you have it) run `clang-format` on your changes.
> 
> Thanks.

I missed this one. Will work on that now.

andir:
I added another commit (that can potentially be squashed with the first) that handles the sendfile.c conversion. Since that code isn't covered by any tests I haven't tested it just yet. I should probably write a few.

sendlib.c
```diff
@@ -806,20 +805,27 @@ static size_t convert_file_to(FILE *fp, const char *fromcode, int ncodes,
...
+  struct ContentState *states =
+      mutt_mem_calloc(tocodes->count, sizeof(struct ContentState));
+  struct Content *infos = mutt_mem_calloc(tocodes->count, sizeof(struct Content));
 
-  for (int i = 0; i < ncodes; i++)
   {
-    if (mutt_str_strcasecmp(tocodes[i], "utf-8") != 0)
-      cd[i] = mutt_ch_iconv_open(tocodes[i], "utf-8", 0);
-    else
+    int i = 0;
```
andir: I decided to add another "scope" here so the counter (`i`) would go out of scope after this block of code as it is a very frequently used identifier and kinda intuitive when you see it int he code. It doesn't have a use after that loop.

It is probably not the best solution but without a `STAILQ_FOREACH_WITH_COUNTER` (or similar) macro that seemed reasonable.

mutt/charset.c
```diff
@@ -303,27 +304,24 @@ int mutt_ch_convert_nonmime_string(char **ps)
...
     size_t ulen = mutt_str_strlen(*ps);
 
     if (!u || !*u)
       return 0;
 
-    c1 = strchr(c, ':');
-    size_t n = c1 ? c1 - c : mutt_str_strlen(c);
+    size_t n = mutt_str_strlen(from);
     if (n == 0)
       return 0;
```
flatcap: We can simplify this.
The `DT_SLIST` stores empty strings as `NULL`.
(Currently, none of the config variables has `SLIST_ALLOW_EMPTY`)

I think we can reduce this to:
```c
if (!from)
  return 0;
```

and move it to the beginning of the `FOREACH`

flatcap: Looking more closely at this function, I spotted an error that I introduced -- the result of the conversion was being discarded.
I've pushed a fix and tidied the function.

sendlib.c
```diff
@@ -959,24 +966,7 @@ static size_t convert_file_from_to(FILE *fp, const char *fromcodes, const char *
...
-  for (c = tocodes; c; c = c1 ? c1 + 1 : 0)
-  {
-    c1 = strchr(c, ':');
-    if (c1 == c)
-      continue;
-    ncodes++;
-  }
-
-  /* Copy them */
-  tcode = mutt_mem_malloc(ncodes * sizeof(char *));
```
andir: Mh, we should probably have not merged this yet. This did segfault when you try to write a mail. At least that is what I was looking into yesterday. At line 986 we will try to write to `tcode` but that is now always `NULL`.

gahr: I agree - that should be reverted.

flatcap: sorry, reverted

flatcap:
I've created branch for you to work on [[devel/slist]](https://github.com/neomutt/neomutt/compare/devel/slist) which contains all your slist changes.
GitHub won't allow me to re-open this PR, so please open a new one when you're ready.

andir:
On 04:02 27.03.20, Richard Russon wrote:
> I've created branch for you to work on [[devel/slist]](https://github.com/neomutt/neomutt/compare/devel/slist) which contains all your slist changes.
> GitHub won't allow me to re-open this PR, so please open a new one when you're ready.

:+1: will do

PR #2198, opened 2020-03-20
Refactor the Compose screen

flatcap:
- Compose dialog uses multiple Windows
- `To:`, `Cc:` and `Bcc:` fields now span up to 5 lines
- `To:`, `Cc:` and `Bcc:` fields are now wrapped at address boundaries
- `To:`, `Cc:` and `Bcc:` fields indicate overflow with `(+3 more)`
- Truly fluid layout -- expand Envelope section to fit contents

---

Sample data: https://gist.github.com/flatcap/dfe02d1c71f080ca1de48409f8be0bad

```
neomutt -n -d1 -F names.rc -F auto.rc -f /dev/null
```

Press <kbd>F1</kbd> to open the compose menu.

### DIFF
--- a/compose.c
+++ b/compose.c
@@ -790,22 +790,30 @@ static int draw_envelope_addr(int field, struct AddressList *al,
   try_again:
     more_len = snprintf(more, sizeof(more),
                         ngettext("(+%d more)", "(+%d more)", count), count);
+    mutt_debug(LL_DEBUG3, "text: '%s'  len: %d\n", more, more_len);
 
     int reserve = ((count > 0) && (lines_used == max_lines)) ? more_len : 0;
+    mutt_debug(LL_DEBUG3, "processing: %s (al:%ld, wl:%d, r:%d, lu:%ld)\n",
+               np->data, addr_len, width_left, reserve, lines_used);
     if (addr_len >= (width_left - reserve))
     {
+      mutt_debug(LL_DEBUG3, "not enough space\n");
       if (lines_used == max_lines)
       {
+        mutt_debug(LL_DEBUG3, "no more lines\n");
+        mutt_debug(LL_DEBUG3, "truncating: %s\n", np->data);
         mutt_paddstr(width_left, np->data);
         break;
       }
 
       if (width_left == (win->state.cols - MaxHeaderWidth))
       {
+        mutt_debug(LL_DEBUG3, "couldn't print: %s\n", np->data);
         mutt_paddstr(width_left, np->data);
         break;
       }
 
+      mutt_debug(LL_DEBUG3, "start a new line\n");
       mutt_window_clrtoeol(win);
       row++;
       lines_used++;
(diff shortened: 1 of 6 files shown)


compose.c
```diff
@@ -648,94 +870,80 @@ static void draw_envelope(struct ComposeRedrawData *rd)
 {
   struct Email *e = rd->email;
   const char *fcc = mutt_b2s(rd->fcc);
+  const int cols = rd->win_envelope->state.cols - MaxHeaderWidth;
+  int row = 0;
+
+  row += draw_envelope_addr(HDR_FROM, &e->env->from, rd->win_envelope, row, 1);
```
gahr, line 876: `int row = draw_envelope_addr(HDR_FROM, &e->env->from, rd->win_envelope, row, 1);`

compose.c
```diff
@@ -624,20 +753,113 @@ static int check_attachments(struct AttachCtx *actx)
...
+  char more[32];
+  int more_len = 0;
+
+  char *sep = NULL;
+  struct ListNode *next = NULL;
+  struct ListNode *np = NULL;
+  STAILQ_FOREACH(np, &list, entries)
+  {
+    next = STAILQ_NEXT(np, entries);
+    addr_len = mutt_strwidth(np->data);
```
gahr, line 783: I'd move `addr_len`' definition here

compose.c
```diff
@@ -515,28 +644,28 @@ static void update_crypt_info(struct ComposeRedrawData *rd)
   }
 #endif
 
-  redraw_crypt_lines(rd);
+  int row = calc_security(rd->email, &rd->sec_rows);
+  row += 2 + rd->to_rows + rd->cc_rows + rd->bcc_rows;
```
gahr, line 648: Is this the same `2` used to initialize ` int used = 2;` ? If so, can we give it a name?

compose.c
```diff
@@ -624,20 +753,113 @@ static int check_attachments(struct AttachCtx *actx)
...
+      sep = "";
+    }
+
+    count--;
+  try_again:
+    more_len = snprintf(more, sizeof(more),
+                        ngettext("(+%d more)", "(+%d more)", count), count);
+    mutt_debug(LL_DEBUG3, "text: '%s'  len: %d\n", more, more_len);
+
+    reserve = ((count > 0) && (lines_used == max_lines)) ? more_len : 0;
```
gahr, line 800: I'd move `reserve`'s definition here

compose.c
```diff
@@ -359,24 +372,151 @@ static void autocrypt_compose_menu(struct Email *e)
...
+ * @param field Field to display, e.g. #HDR_FROM
+ */
+static void draw_header(struct MuttWindow *win, int row, enum HeaderField field)
+{
   mutt_curses_set_color(MT_COLOR_COMPOSE_HEADER);
-  mutt_window_mvprintw(rd->win, HDR_CRYPT, 0, "%*s", HeaderPadding[HDR_CRYPT],
-                       _(Prompts[HDR_CRYPT]));
+  mutt_window_mvprintw(win, row, 0, "%*s", HeaderPadding[field], _(Prompts[field]));
   mutt_curses_set_color(MT_COLOR_NORMAL);
+}
```
gahr, line 399: We could possibly merge these two functions, with a size_t to indicate the width. We could use 0 for "no padding". `printf("%*s\n", 0, "Hello world");` is fine.

flatcap: Hmm... that would change eleven calls:
```diff
- draw_header(rd->win_envelope, row++, HDR_NEWSGROUPS);
+ draw_header(rd->win_envelope, row++, HeaderPadding[HDR_NEWSGROUPS], _(Prompts[HDR_NEWSGROUPS]));
```

compose.c
flatcap: It turns out that we don't need those lines any more.

flatcap:
Changes made and back-dated.
I've added comments to all the `x += 2;` magic.

PR #2200, opened 2020-03-22
Libcompress update

mcmilk:
* **What does this PR do?**
This PR will externalise the two config variables `C_HeaderCacheCompressLevel` and `C_HeaderCacheCompressDictionary` from libcompress. The library itself gets 2 new functions for this:

1. set_level()
2. set_dictionary()

The two are only called once before compress()
If the default compression level is okay, the calling of these functions is not needed.

### DIFF
--- a/compress/lib.h
+++ b/compress/lib.h
@@ -3,7 +3,7 @@
  * API for the header cache compression
  *
  * @authors
- * Copyright (C) 2019 Tino Reichardt <milky-neomutt@mcmilk.de>
+ * Copyright (C) 2019-2020 Tino Reichardt <milky-neomutt@mcmilk.de>
  *
  * @copyright
  * This program is free software: you can redistribute it and/or modify it under
@@ -32,6 +32,10 @@
  * | compress/lz4.c      | @subpage compress_lz4  |
  * | compress/zlib.c     | @subpage compress_zlib |
  * | compress/zstd.c     | @subpage compress_zstd |
+ *
+ * Usage with Compression Level set to X:
+ * - open(level X) -> N times compress() -> close()
+ * - open(level X) -> N times decompress() -> close()
  */
 
 #ifndef MUTT_COMPRESS_LIB_H
@@ -48,10 +52,11 @@ struct ComprOps
 
   /**
    * open - Open a compression context
+   * @param[in]  level The compression level
    * @retval ptr  Success, backend-specific context
    * @retval NULL Otherwise
    */
-  void *(*open)(void);
+  void *(*open)(short level);
 
   /**
    * compress - Compress header cache data
--- a/compress/lz4.c
+++ b/compress/lz4.c
@@ -3,7 +3,7 @@
  * LZ4 header cache compression
  *
  * @authors
- * Copyright (C) 2019 Tino Reichardt <milky-neomutt@mcmilk.de>
+ * Copyright (C) 2019-2020 Tino Reichardt <milky-neomutt@mcmilk.de>
  *
  * @copyright
  * This program is free software: you can redistribute it and/or modify it under
@@ -34,22 +34,35 @@
 #include "lib.h"
 #include "hcache/lib.h"
 
+#define MIN_COMP_LEVEL 1  ///< Minimum compression level for lz4
+#define MAX_COMP_LEVEL 12 ///< Maximum compression level for lz4
+
 /**
  * struct ComprLz4Ctx - Private Lz4 Compression Context
  */
 struct ComprLz4Ctx
 {
-  void *buf; ///< Temporary buffer
+  void *buf;   ///< Temporary buffer
+  short level; ///< Compression Level to be used
 };
 
 /**
  * compr_lz4_open - Implements ComprOps::open()
  */
-static void *compr_lz4_open(void)
+static void *compr_lz4_open(short level)
 {
   struct ComprLz4Ctx *ctx = mutt_mem_malloc(sizeof(struct ComprLz4Ctx));
 
   ctx->buf = mutt_mem_malloc(LZ4_compressBound(1024 * 32));
+
+  if ((level < MIN_COMP_LEVEL) || (level > MAX_COMP_LEVEL))
+  {
+    mutt_warning(_("The compression level for %s should be between %d and %d"),
+                 compr_lz4_ops.name, MIN_COMP_LEVEL, MAX_COMP_LEVEL);
+    level = MIN_COMP_LEVEL;
+  }
+
+  ctx->level = level;
 
   return ctx;
 }
@@ -63,12 +76,13 @@ static void *compr_lz4_compress(void *cctx, const char *data, size_t dlen, size_
     return NULL;
 
   struct ComprLz4Ctx *ctx = cctx;
+
   int datalen = dlen;
   int len = LZ4_compressBound(dlen);
   mutt_mem_realloc(&ctx->buf, len + 4);
   char *cbuf = ctx->buf;
 
-  len = LZ4_compress_fast(data, cbuf + 4, datalen, len, C_HeaderCacheCompressLevel);
+  len = LZ4_compress_fast(data, cbuf + 4, datalen, len, ctx->level);
   if (len == 0)
     return NULL;
   *clen = len + 4;
--- a/hcache/hcache.c
+++ b/hcache/hcache.c
@@ -7,7 +7,7 @@
  * Copyright (C) 2004 Tobias Werth <sitowert@stud.uni-erlangen.de>
  * Copyright (C) 2004 Brian Fundakowski Feldman <green@FreeBSD.org>
  * Copyright (C) 2016 Pietro Cerutti <gahr@gahr.ch>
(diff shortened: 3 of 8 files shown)


gahr:
Because (according to their docs) ingesting the dictionary is a costly operation, I suggest to set the dictionary and compression level once in `mutt_hcache_open`.

compress/compress_private.h
```diff
@@ -24,12 +25,25 @@
...
+    .name            = #_name,                         \
+    .open            = compr_##_name##_open,           \
+    .set_level       = compr_##_name##_set_level,      \
+    .set_dictionary  = NULL,                           \
+    .compress        = compr_##_name##_compress,       \
+    .decompress      = compr_##_name##_decompress,     \
+    .close           = compr_##_name##_close,          \
+  };
+
+#define COMPRESS_DICT(_name)                  \
```
gahr: I'd prefer to have only one macro, and explictely define the `_set_dictionary()` function as a no-op for backends that do not support it.

compress/lib.h
```diff
@@ -30,6 +30,18 @@
...
+ * open() -> compress() -> close()
+ * open() -> decompress() -> close()
+ *
+ * Usage with Compression Level set to N:
+ * open() -> set_level(N) -> compress() -> close()
+ * open() -> decompress() -> close()
+ *
+ * Usage with Zstd Dictionary Compression and Level:
+ * open() -> set_level(N) -> set_dictionary() -> compress() -> close()
+ * open() -> set_dictionary() -> decompress() -> close()
```
gahr: These comments are misleading, as they suggest a single `compress` for a single `open`/`close` pairs. In reality, at least for IMAP, you often do one `open`, followed by a number of `compress`, followed by one `close`. This is why I suggest to move `set_level` and `set_dictionary` inside `open` instead of inside `store`.

mcmilk: I asked @flatcap before, he wanted set_level() and set_dictionary() instead of open(level, dictionary).

compress/compress_private.h
mcmilk: How does this work? Could you show me your suggestion ?

mcmilk:
> Because (according to their docs) ingesting the dictionary is a costly operation, I suggest to set the dictionary and compression level once in `mutt_hcache_open`.

You are right, I was to fast with the PR ;-/

compress/compress_private.h
mcmilk: I thing I got it, see 6744fc74eb439daf1dd9d2e3e361204f3ba8ca41 ;)

mcmilk:
If you want, I can rebase the commits, when everything is fine now.

gahr:
Thanks! Additionally, it occurred to me that because level and dictionary are only used after an open, would it perhaps make sense to pass them as arguments to open directly?

mcmilk:
> Thanks! Additionally, it occurred to me that because level and dictionary are only used after an open, would it perhaps make sense to pass them as arguments to open directly?

@flatcap wanted the functions set_level() and set_dictionary() instead of open(level, dictionary).
Maybe he wants to use these functions in other places, without the need of parameter ?

flatcap:
> @flatcap wanted the functions set_level() and set_dictionary() instead of open(level, dictionary).

Well... yes and no.

My first thought **was** `open(level, dictionary)`, but I wasn't happy because the dictionary was specific to zstd.
(if another compression came along with a different variable, then what?)

Perhaps, `open(level)` as that is common to all compression methods, and separate methods for the specifics.

I agree that we only want to set the level/dictionary once, but we **do** control all the code, so we can ensure that's what happens.

compress/lz4.c
```diff
@@ -50,10 +51,34 @@ static void *compr_lz4_open(void)
...
+ */
+static void compr_lz4_set_level(void *cctx, short level)
+{
+  if (!cctx)
+    return;
+
+  struct ComprLz4Ctx *ctx = cctx;
+
+  if (level < 1 || level > 12)
+    mutt_error("Wrong Compression Level for LZ4!");
```
flatcap: A few points:
- If the compression level is out of range, should we clamp the value?

- Add constants for the minimum, default and maximum compression levels

- Please can we change the message to help the user, e.g.
  ```c
  mutt_warning(_("The compression level for %s should be between %d and %d"),
               ctx->name, MIN_COMP_LEVEL, MAX_COMP_LEVEL);
  ```

Message:
- Tell the user the limits
- Reduced to a warning, because we're telling the user what to do next time
- Removed the `!` so it's less SHOUTY
- Add the gettext translation macro `_("text")`
- Made the message generic so it can be shared by all three compression methods
  (reducing the work for translators)

hcache/hcache.c
```diff
@@ -435,8 +435,13 @@ header_cache_t *mutt_hcache_open(const char *path, const char *folder, hcache_na
       return NULL;
     }
 
+    cops->set_level(hc->cctx, C_HeaderCacheCompressLevel);
+
+    /* when dictionary is set, but the file does no exist, there is no problem */
```
flatcap:
```suggestion
    /* when dictionary is set, but the file does not exist, there is no problem */
```

compress/lib.h
```diff
@@ -30,6 +30,18 @@
...
+ * open() -> N times compress() -> close()
+ * open() -> N times decompress() -> close()
+ *
+ * Usage with Compression Level set to N:
+ * open() -> set_level(N) -> N times compress() -> close()
+ * open() -> N times decompress() -> close()
+ *
+ * Usage with Zstd Dictionary Compression and Level:
+ * open() -> set_level(N) -> set_dictionary() -> N times compress() -> close()
+ * open() -> set_dictionary() -> N times decompress() -> close()
```
flatcap:
```suggestion
 * Default usage with Compression Level set to 1:
 * - open() -> N times compress() -> close()
 * - open() -> N times decompress() -> close()
 *
 * Usage with Compression Level set to N:
 * - open() -> set_level(N) -> N times compress() -> close()
 * - open() -> N times decompress() -> close()
 *
 * Usage with Zstd Dictionary Compression and Level:
 * - open() -> set_level(N) -> set_dictionary() -> N times compress() -> close()
 * - open() -> set_dictionary() -> N times decompress() -> close()
```

This comment is in markdown, so I've added some bullet points to prevent it being wrapped into three long sentences.

flatcap:
Aside...

atm, we're using the API pointers, e.g. `cops->open()`.
Do we want to add a thin wrapper of functions, hiding the details with a single opaque handle?

flatcap:
@mcmilk 
Sorry, I'm going to cause trouble again...
How upset would you be if we dropped the Zstd dictionary completely?

gahr and I have been discussing this at length and we come to some conclusions:

- Complexity:
  - Compressing with a dictionary means you're tied to it for the life of the cache
  - The dictionary complicates the interface

- Effectiveness:
  - There's only one user of the compression, the hcache, and a lot its data is binary
  - Even with a great dictionary it won't improve the compression ratio by much
  - When #1870 returns, the hcache data blobs will be much smaller and less compressible

Without the dictionary, we can have a simple: `open(level)` function.
Sorry for messing you about.

compress/lz4.c
mcmilk: Will do so.

mcmilk:
> @mcmilk
> Sorry, I'm going to cause trouble again...
> How upset would you be if we dropped the Zstd dictionary completely?
> 
> gahr and I have been discussing this at length and we come to some conclusions:
> 
> * Complexity:
>   
>   * Compressing with a dictionary means you're tied to it for the life of the cache
>   * The dictionary complicates the interface
> * Effectiveness:
>   
>   * There's only one user of the compression, the hcache, and a lot its data is binary
>   * Even with a great dictionary it won't improve the compression ratio by much
>   * When #1870 returns, the hcache data blobs will be much smaller and less compressible
> 
> Without the dictionary, we can have a simple: `open(level)` function.
> Sorry for messing you about.

Okay, I will remove zstd dictionary :(
I have not really a problem with it.

But therefore I will add rocksdb and tdb hcache backends ;-) And then we have finished the header compression stuff for some time, I think.

compress/compress_private.h
```diff
@@ -24,12 +24,12 @@
...
-    .compress   = compr_##_name##_compress,         \
-    .decompress = compr_##_name##_decompress,       \
-    .close      = compr_##_name##_close,            \
+  const struct ComprOps compr_##_name##_ops = {        \
+    .name            = #_name,                         \
+    .open            = compr_##_name##_open,           \
+    .compress        = compr_##_name##_compress,       \
+    .decompress      = compr_##_name##_decompress,     \
+    .close           = compr_##_name##_close,          \
   };
```
gahr: I guess the changes to this file can now be undone

mcmilk: Yes, of cause ;)

flatcap:
I just made a few minor changes:
- clang-format
- added `()`s to some logic
- Added Doxygen comments `///<` after the min/max constants (so the docs work nicely)

When you make a commit, please can you make sure there's a blank line after the subject,
otherwise `git` thinks those lines are a continuation and `git log` gets messy.  Thanks.

PR #2211, opened 2020-03-26
test: add test files

flatcap:
The tests aren't very exciting, they simply cover functions from `mutt/file.c`

The tricky part is the test files required for these functions.
They live in the [neomutt-test-files](https://github.com/neomutt/neomutt-test-files) repo.

First, Travis/Cirrus clones the repo.
The location is passed to `neomutt-test` using an environment variable `$NEOMUTT_TEST_DIR`
(we can't pass parameters to the Acutest program)

Most of the functions being tested are expecting a fully-qualified path.
The test function expands a template, "%s/file/empty", using the environment variable, and tests a function with the result.

### DIFF
--- a/.cirrus.yml
+++ b/.cirrus.yml
@@ -5,8 +5,9 @@ task:
   only_if: $CIRRUS_BRANCH == "master" || $CIRRUS_PR != ""
   skip: $CIRRUS_PR_DRAFT == "true"
   env:
+    CIRRUS_CLONE_DEPTH: 3
     PKG_COMMON:
-      cyrus-sasl db5 docbook-xsl gdbm gettext-tools gpgme iconv jimtcl
+      cyrus-sasl db5 docbook-xsl gdbm gettext-tools git gpgme iconv jimtcl
       kyotocabinet libgpg-error libidn2 libxslt lmdb lua53 lynx notmuch pkgconf
       qdbm tdb tokyocabinet urlview
     CONFIGURE_COMMON:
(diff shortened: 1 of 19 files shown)


.cirrus.yml
```diff
@@ -23,7 +25,10 @@ task:
       install_script: pkg install -y ${PKG_COMMON} libressl
       configure_script: ./configure ${CONFIGURE_COMMON} --ssl
     - name: FreeBSD / GnuTLS
-      install_script: pkg install -y ${PKG_COMMON} gnutls
+      install_script:
+        - pkg install -y ${PKG_COMMON} gnutls
+        - git clone --depth 1 https://github.com/neomutt/neomutt-test-files.git ${NEOMUTT_TEST_DIR}
+        - (cd ${NEOMUTT_TEST_DIR} && ./setup.sh)
```
gahr: Isn't it doing the thing only for the GnuTLS branch?

test/file/common.c
```diff
@@ -68,3 +68,36 @@ void file_tear_down(FILE *fp, const char *funcname)
...
+void test_gen_dir(char *buf, size_t buflen, const char *fmt)
+{
+  static const char *dir = NULL;
+  if (!dir)
+  {
+    const char *path = get_test_path();
+    const char *slash = strrchr(path, '/');
+    dir = slash + 1;
+  }
+  if (!dir)
```
gahr: I guess `dir` will never be NULL here, at best NULL + 1.

gahr:
Nice!

.cirrus.yml
flatcap: Ah!
I deleted the other branches, whilst testing, to reduce the load on Cirrus.
When I restored them, I forgot to add the testing code to the others.

But... do we need it?
We don't test the OpenSSL/GnuTLS/LibreSSL nor the Curses/Slang code, so there probably isn't any point repeating the tests.

test/file/common.c
flatcap: True.  I'm missing an `if (slash)` before the `dir = `.
However, the function isn't used any more, so it goes.

.cirrus.yml
gahr: Very true - the question is then which branch is the default. But it doesn't matter

flatcap:
**Changes**:
- Added **Test** build
  Minimal build + `make test`
- Removed testing from other builds
- Added `neomutt -v` to all builds

PR #2218, opened 2020-03-27
Draft of pre-defined/pre-compiled regexes

gahr:
* **What does this PR do?**

Define a number of pre-compiled regexes, so client code doesn't need to manage their memory.


flatcap:
It certainly makes the callers much tidier.

flatcap:
My [imap generator](https://gist.github.com/flatcap/1725d238f8c35a7be53ebb445c389a8a) may be useful for testing your URL-parsing.
It creates 160 variants of IMAP path.

flatcap:
- Rebased onto master
- Lightly squashed
  (back-dated fixes and tidying)
- Doxygen: updated comments, fix typos
- Renamed a few params
  (try to reserve `m` for Mailboxes)

gahr:
Potential incompatibility: I am stricter than we used to be in which set of characters we accept in parts of the URL, especially in query strings. This means that
`mailto:foo@bar.baz?subject=now yo see me`
won't work anymore. Is that an issue?

gahr:
> Potential incompatibility: I am stricter than we used to be in which set of characters we accept in parts of the URL, especially in query strings. This means that
> `mailto:foo@bar.baz?subject=now yo see me`
> won't work anymore. Is that an issue?

We resolved to allow `@` characters in the user part of and ' ' (space) characters in the query parts of a URL.

PR #2220, opened 2020-03-27
Document the _seven_ database backends a bit more.

mcmilk:
* **What does this PR do?**

This PR fixes the count of the backends and adds the homepages of the dbms and also the three compression methods.


mcmilk:
@flatcap @gahr - It would be nice, if we can re-add the internal backend compression routines with some config value like "header_cache_compress_builtin" ... this would be the text for the README.md file:

---
Four backends have the ability to offer internal compression, which is in
nearly all cases superior to the generic compression provided by setting
"header_cache_compress_method". The internal backend compression can be
enabled by setting "header_cache_compress_builtin" to the desired mode:
- KyotoCabinet (internal: zlib)
- QDBM (internal: zlib)
- TokyoCabinet (internal: zlib)
- RocksDB (internal: lz4, snappy, zstd, zlib)
---

What do you think about it ?

flatcap:
Merged in f49faa7464.

flatcap:
I've taken your changes a little further and merged the `README.md` into the Doxygen headers.
Here are the results:
- https://neomutt.org/code/hcache.html
- https://neomutt.org/code/store.html
- https://neomutt.org/code/compress.html

flatcap:
>  re-add the internal backend compression routines

We discussed this a bit and decided that is was just adding complexity without adding much benefit.

> ... superior ...

Yeah, probably, but by how much?
We're only compressing small data, so the gains would only be small too.

> function, sth. like `open(path,features)`

and we're back to complexity.
We're making the code (and docs) harder to understand to save a few bytes.

> I think `version()` can be removed, it's not used

Well spotted.
Even if we don't show the info to the user, it's probably worth logging it.

PR #2224, opened 2020-03-31
Cache a mailbox's URL

gahr:
In a simple test case involving opening an IMAP account with 39 subscribed mailboxes, this reduces the number to url_parse from 954 to 173.


gahr:
I'll close this so we don't have to rebase it everytime.

PR #2235, opened 2020-04-04
perl scripts

flatcap:
Upstream have converted two scripts to perl in order to be more portable.

- [hcachever.pl](https://github.com/neomutt/neomutt/blob/devel/perl/hcachever.pl) - Header Cache checksum
  This is to get around problems of `md5sum` availability

- [makedoc.pl](https://github.com/neomutt/neomutt/blob/devel/perl/makedoc.pl) - `mutt_config.c` comment parser

Are we interested?

**Note**: These scripts haven't been adjusted to work with our build.

### DIFF
--- a/hcachever.pl
+++ b/hcachever.pl
@@ -1,0 +1,112 @@
+#!/usr/bin/perl -w
+#
+# Copyright (C) 2020 Kevin J. McCarthy <kevin@8t8.us>
+#
+#     This program is free software; you can redistribute it and/or modify
+#     it under the terms of the GNU General Public License as published by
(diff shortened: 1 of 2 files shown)


flatcap:
No interest.
Hopefully `makedoc.c` will become obsolete (with the new help backend).
`hcachever.sh` is complicated, but that's probably a sign that we should refactor other things.

jindraj:
I can take a look on hcachever.sh and try to make it more firendly.

flatcap:
> I can take a look

Thanks.
You may want to read it before offering help, though (it's a scary mess :-)

PR #2265, opened 2020-04-16
Remove dep files (*.Po) in distclean

gahr:


gahr:
The problem is not here, it's in the generation of the Tpo files in the same directory as the object files (gmake) vs in the source directory (bsd make). I'll look further.

PR #2271, opened 2020-04-17
refactor charset variables to use SList's 

andir:
Let's give this another shot at a review.

I did use this branch for a bit but I also do not have that many fancy charset configurations or mails.

### DIFF
--- a/config/charset.c
+++ b/config/charset.c
@@ -34,6 +34,27 @@
 #include "lib.h"
 
 /**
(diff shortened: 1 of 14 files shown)


flatcap:
Currently, `make test` fails.
Specfically, this one crashes:
```
test/neomutt-test test_mutt_ch_get_default_charset
```

I've temporarily marked the PR as 'draft' -- this disables the CirrusCI builds which can be slow.

flatcap:
Archived to https://github.com/neomutt/neomutt-old#neomutt-2021-02-05

PR #2284, opened 2020-04-22
Avoid crashing / handle UTF-8 URL paths

### LINKED ISSUE #2283: url_parse crashes on UTF-8 paths when using PCRE2
While investigating #2282, I noticed that NeoMutt crashes in `imap_ac_find` because `url_parse` returns NULL (which is not checked) for this path: `imaps://foobar@gmail.com@imap.gmail.com/Отправленные письма`. This only happens with PCRE2 enabled.

gahr:
`[:alnum:]` doesn't match characters outside of ASCII. Let's use `\p{L}\p{N}` (letters and numbers) instead. This only matches a subset of printable characters, but should be enough for most cases (UTF-8 IMAP mailbox names).

Fixes #2283

### DIFF
--- a/test/url/url_parse.c
+++ b/test/url/url_parse.c
@@ -23,6 +23,7 @@
 #define TEST_NO_MAIN
 #include "config.h"
 #include "acutest.h"
+#include <locale.h>
 #include "mutt/lib.h"
 #include "address/lib.h"
 #include "email/lib.h"
@@ -165,6 +166,19 @@ static struct UrlTest test[] = {
     },
     "subject|#|"
   },
+  {
+    /* UTF-8 mailbox name */
+    "imaps://foobar@gmail.com@imap.gmail.com/Отправленные письма",
+    true,
+    {
+      U_IMAPS,
+      "foobar@gmail.com",
+      NULL,
+      "imap.gmail.com",
+      0,
+      "Отправленные письма"
+    }
+  }
 };
 // clang-format on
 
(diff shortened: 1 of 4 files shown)


gahr:
I am puzzled: `url_parse` is fine with the mailbox when it appears as a result of `imap_ac_find`, but the test fails.

test/url/url_parse.c
```diff
@@ -218,6 +219,9 @@ void check_query_string(const char *exp, const struct UrlQueryList *act)
 
 void test_url_parse(void)
 {
+  // let's pick a utf-8 locale, since we're also parsing utf-8 text */
+  setlocale(LC_ALL, "en_US.UTF-8");
```
flatcap, line 223: As a couple of other tests require this:
- [`test_rfc2047_encode()`
](https://github.com/neomutt/neomutt/blob/master/test/rfc2047/rfc2047_encode.c#L35)
- [`test_rfc2047_decode()`](https://github.com/neomutt/neomutt/blob/master/test/rfc2047/rfc2047_decode.c#L36)

How about adding it to [`test_common()`](https://github.com/neomutt/neomutt/blob/master/test/common.c#L43)

They have a fallback of `C.UTF-8`, which also works for `test_url_parse()`.

flatcap:
I've just piggybacked 65b8bc7458 onto this PR to add to the `neomutt -v` output:
- `+regex` -- libc regex
- `+pcre2` -- pcre2 regex

We could also add the specific pcre2 version if necessary.

Issue #2272, opened 2020-04-17
sidebar not indented with multiple imap accounts


reox:
I have two IMAP accounts set up using folder hooks and two configuration files like this:

```
source ~/.mutt/account_a
folder-hook $folder 'source ~/.mutt/account_a
source ~/.mutt/account_b
folder-hook $folder 'source ~/.mutt/account_b
macro index,pager <f2> '<sync-mailbox><refresh><enter-command>source ~/.mutt/account_a<enter><change-folder>!<enter><check-stats>'
macro index,pager <f3> '<sync-mailbox><refresh><enter-command>source ~/.mutt/account_b<enter><change-folder>!<enter><check-stats>'
```

The account files look like this:
```
unmailboxes *

set folder    = 'imap://user@imapserver/'
set spoolfile = $folder
set record    = "+INBOX.Sent"
set postponed = "+INBOX.Drafts"
set trash     = "+INBOX.Trash"


account-hook $folder "set imap_user=user; set imap_pass=1234"

send-hook . 'set from=user@domain'

set signature="~/.mutt/signature_a"

color status black blue

set imap_check_subscribed=yes
mailboxes $folder
```

So far, all the switching works fine. However I have troubles getting the indentation of the mailbox work correctly. Furthermore, it is different between the two servers.
The first server has the folders sorted like seen in the file above.
The other server however uses a folder structure like this:
```
set spoolfile = $folder
set record    = "+Sent"
set postponed = "+Drafts"
set trash     = "+Trash"
```

While The indentation works for server A, it does not for server B. Here, I only see a flat hierarchy of all folder (including subfolders). If I add `mailbox $folder` to server B, I see INBOX twice, while the first one is marked as active. If I remove the line, I still see all folders (and only one INBOX) but there is no active mailbox marked in the sidebar.
I have the feeling that `imap_check_subscribed` might cause issues here and double checked that the mailboxes are indeed subscribed. Now, the real funny part is, if I go into the mailbox overview and press `s` on a mailbox to subscribe it, it will be added to the sidebar - correctly intended! But if I switch accounts and go back to server B, all is gone and I see the flat list again.
Btw: Can you see in neomutt which mailboxes are subscribed? I just checked that in roundcube right now...

I do not really want to add all imap mailboxes to the `mailboxes` manually, there are quite a few...
Is there anything that could work?

reox:
[...]

PR #2289, opened 2020-04-23
fix sidebar indentation

flatcap:
Overhaul the sidebar indentation rules.

Fixes: #2272 

The diff isn't meaningful, so here's a summary.

Sidebar uses five config variables for indentation:

| Variable                | Description                                        |
| :---------------------- | :------------------------------------------------- |
| sidebar_short_path      | Abbreviate the paths using $folder                 |
| sidebar_folder_indent   | Indent nested folders                              |
| sidebar_delim_chars     | Characters that separate nested folders            |
| sidebar_component_depth | Strip leading path components from sidebar folders |
| sidebar_indent_string   | Indent nested folders using this string            |

1. Start with a Mailbox.
   If it has a name, use that, goto step 4)

2. Is **`$sidebar_short_path`** set?
   If NOT, goto step 4)

3. Is **`$folder`** set?
   1. Yes, abbreviate the path using **`$folder`**
   2. No, abbreviate the path using heuristics
      Strip the host part (or database part for notmuch)

4. Is **`$sidebar_folder_indent`** set?
   If NOT, goto step 7)

5. Calculate the **depth** of the path, using **`$sidebar_delim_chars`**
   (and remember the last component of the path)

6. Is **`$sidebar_component_depth`** set?
   Yes, reduce the **depth** by this amount

7. Display the **`$sidebar_indent_string`** **depth** times

8. Display the saved short path

### DIFF
--- a/sidebar.c
+++ b/sidebar.c
@@ -877,12 +877,11 @@ static void fill_empty_space(struct MuttWindow *win, int first_row,
  * imap_is_prefix - Check if folder matches the beginning of mbox
  * @param folder Folder
  * @param mbox   Mailbox path
- * @param plen   Prefix length
- * @retval true If folder is the prefix of mbox
+ * @retval num Length of the prefix
  */
-static bool imap_is_prefix(const char *folder, const char *mbox, size_t *plen)
+static int imap_is_prefix(const char *folder, const char *mbox)
 {
-  bool rc = false;
+  int plen = 0;
 
   struct Url *url_m = url_parse(mbox);
   struct Url *url_f = url_parse(folder);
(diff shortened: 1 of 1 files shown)


sidebar.c
```diff
@@ -903,15 +902,111 @@ static bool imap_is_prefix(const char *folder, const char *mbox, size_t *plen)
...
+    if (query)
+      return query + 1;
+    return mbox;
+  }
+
+  // Skip 'imaps://', etc
+  const char *last = strchr(mbox + 10, '/');
+  if (last)
+    mbox = last + 1;
+  return mbox;
```
Austin-Ray, line 986: Couldn't this part be simplified to:

```C
  const char target_char = type == MUTT_NOTMUCH ? '?' : '/';

  // Skip 'imaps://', etc
  const char *last = strchr(mbox + 10, target_char);
  if (last)
    mbox = last + 1;
  return mbox;
```

Also, could you explain how the `mbox = last + 1;` works since doesn't that violate the `const` invariant?

sidebar.c
```diff
@@ -903,15 +902,111 @@ static bool imap_is_prefix(const char *folder, const char *mbox, size_t *plen)
...
+ * @param mbox Mailbox path to shorten
+ * @param type Mailbox type
+ *
+ * Use heuristics to shorten a non-local Mailbox path.
+ * Strip the host part (or database part for Notmuch).
+ */
+static const char *abbrev_net(const char *mbox, enum MailboxType type)
+{
+  size_t len = mutt_str_strlen(mbox);
+  if ((len < 10) || ((type != MUTT_NNTP) && (type != MUTT_IMAP) &&
```
Austin-Ray, line 968: Could we assign a meaningful name to the value of `10` so it isn't a magic number? In the context of `notmuch`, I know it's the length of the protocol, but for IMAP, it's unclear to me.

Austin-Ray:
I just noticed you fixed an unreported bug with using `sidebar_short_path` on notmuch `mailboxes` declarations (using just URI like a psychopath). On master, this config:

```
mailboxes "notmuch:///home/aray/mail?query=tag:test"
mailboxes "notmuch://?query=tag:test"
```

Results in a sidebar like this:

```
mail?query=tag:test
?query=tag:test
```

On this branch, it yields

```
query=tag:test
query=tag:test
```

Yay! consistency!

sidebar.c
```diff
@@ -903,15 +902,111 @@ static bool imap_is_prefix(const char *folder, const char *mbox, size_t *plen)
   if (mutt_str_strncmp(url_m->path, url_f->path, flen) != 0)
     goto done;
 
-  if (url_m->user && !url_f->user)
-    *plen += mutt_str_strlen(url_m->user) + 1;
-  rc = true;
+  plen = strlen(mbox) - mlen + flen;
```
gahr, line 905: Does this mean that `imaps://foo@server.com/Inbox` and `imaps://server.com/Inbox` will match? Is that what we want?

sidebar.c
```diff
@@ -903,15 +902,111 @@ static bool imap_is_prefix(const char *folder, const char *mbox, size_t *plen)
...
+
+/**
+ * abbrev_net - Abbreviate a network Mailbox path
+ * @param mbox Mailbox path to shorten
+ * @param type Mailbox type
+ *
+ * Use heuristics to shorten a non-local Mailbox path.
+ * Strip the host part (or database part for Notmuch).
+ */
+static const char *abbrev_net(const char *mbox, enum MailboxType type)
```
gahr, line 965: I also think that `abbrev_net` could use some clarification. Isn't it amounting to returning the path+query part of a URL?

gahr:
> I just noticed you fixed an unreported bug with using `sidebar_short_path` on notmuch `mailboxes` declarations (using just URI like a psychopath). On master, this config:
> 
> ```
> mailboxes "notmuch:///home/aray/mail?query=tag:test"
> mailboxes "notmuch://?query=tag:test"
> ```
> 
> Results in a sidebar like this:
> 
> ```
> mail?query=tag:test
> ?query=tag:test
> ```
> 
> On this branch, it yields
> 
> ```
> query=tag:test
> query=tag:test
> ```
> 
> Yay! consistency!

I guess this is correct only of `set folder = /home/aray/mail` - is this the case?

flatcap:
> Could you update the commit message explaining how the code changes resolve the indentation issue?

Done.

sidebar.c
flatcap:
```c
const char *mbox = NULL;
```
Here, the char is const, but not the pointer.
So `mbox[0] = 'a';` isn't possible, but `mbox++;` is.

```c
const char const *mbox = NULL;
```
Here, the pointer is const too, so that would forbid `mbox++`.

sidebar.c
flatcap: I made it a constant and documented it.

sidebar.c
flatcap: I've renamed it to `abbrev_url()` and added some examples.

flatcap:
> > `mailboxes "notmuch:///home/aray/mail?query=tag:test"`
> > `mailboxes "notmuch://?query=tag:test"`

> > Results in a sidebar like this:
> > `query=tag:test`
> > `query=tag:test`

> I guess this is correct only of `set folder = /home/aray/mail` - is this the case?

`$folder` has priority, but the heuristics will match most people's `$folder` anyway.

- Path: **`notmuch:///home/aray/mail?query=tag:test`**

- `set folder = 'notmuch:///home/aray/mail` (exact match)
  Short: **`query=tag:test`**

- `set folder = 'notmuch:///home/aray` (partial match)
  Short: **`mail?query=tag:test`**

- `set folder = '/different` (heuristics)
  Short: **`query=tag:test`**

gahr:
>     * Path: notmuch:///home/aray/mail?query=tag:test
>     * `set folder = '/different` (heuristics)
>       Short: **`query=tag:test`**

I am not sure I agree with this one. You'll end up having `query=tag:test` for two identical queries against two different mailboxes?

flatcap:
> I am not sure I agree with this one.
> You'll end up having `query=tag:test` for two identical queries against two different mailboxes?

In theory, yes.
In practice, no.  You can only have one Notmuch database (a _current_ limitation of how we manage accounts).

I'm not sure how useful/practical having two databases would be (but it will be possible in the future).

With multiple databases, this is the same as the multiple inboxes problem.
It'd be up to the user to name the mailboxes, or insert labels.

Issue #2236, opened 2020-04-04
Add += and -= to the config system


jindraj:
### From IRC #neomutt - flatcaps thoughts about the feature:
I'd like to add += and -= to the config system, but I'd need a volunteer
(I have an idea how it'd work, but I haven't investigated it fully)
most of it would be trivial :-)

the tricky bit would be altering parse_set()
that function's been overhauled since the mutt days, so it's not too scary

I just haven't looked at what would be involved
a simple version would add some functions to ConfigSetType: plus_equals(), minus_equals()

(initially) we'd only implement those functions for numeric types: config/number,long
then in parse_set(), there's a `if (*s->dptr == '=')` which is looking for 'set x = y'
we'd want extra bits in there testing for `+=` and `-=`

the job of parse_set() is to split up the string and decide which function to call
the job of setting, resetting, adding, etc is done by the dedicated config types

## What would you like to be able to do?
Increment and decrement numeric config values.

Executing
- `set variable += N` would take current value of the variable and increment it by **N**
- `set variable -= N` would take current value of the variable and decrement it by **N**

## What needs to change in NeoMutt?
Described in the header text.

## Have you tried any alternative methods?
Yes
- Macros
Would be probably possible but would be messy.
- Lua
Works nice:
```
lua mutt.set("variable",mutt.get("variable") + 1)
lua mutt.set("variable",mutt.get("variable") - 1)
```

hsanson:
This is duplicate of #243

Issue #243, opened 2016-11-15
set plus-equals, minus-equals


flatcap:
- **Task**: Change NeoMutt to allow += and -= to change variables

Some of Vim's variables are lists and they allow the user to add/delete items using the syntax:

```
set variable += apple
set variable -= banana
```

Change NeoMutt to be like Vim.

- **Task**: Add a new variable type of "letter-list"

Some of Vim's variables are collections of characters (a letter list).
Create a new variable type to support this sort of list,
then extend the `+=` and `-=` to work with them.

This issue will be very helpful in solving: https://github.com/neomutt/neomutt/pull/202

PR #2297, opened 2020-04-26
Add += and -= to a config system

jindraj:
* **What does this PR do?**
Implements **+=** and **-=** operators to the config system.

- WIP in command_parse.c
- need to separate the addition substraction logic to a separate functions.

Remaining tasks:
```
flatcap | jindraj: [devel/config-increments]  It works!  Yeah!  So the next step is to
          add plus_equals() minus_equals() to the ConfigSetType
flatcap | you'll need some glue functions in config/set.c and config/subset.c
flatcap | then you can change parse_set() to allow/disallow +=,-= based on the support
          (and get them to do the work)
```

**To do**:

- [x] copy code to long
- [x] prevent non-numeric types crashing
- [x] allow `set config+=10` with no spaces
- [x] documentation
- [x] tests
    - [x] positive numbers
    - [x] negative numbers
    - [x] overflow
    - [x] underflow
    - [x] junk (non-numbers)
    - [x] empty

#243 + #2236 (dup)

### DIFF
--- a/command_parse.c
+++ b/command_parse.c
@@ -1122,7 +1122,7 @@ enum CommandResult parse_set(struct Buffer *buf, struct Buffer *s,
     }
 
     /* get the variable name */
-    mutt_extract_token(buf, s, MUTT_TOKEN_EQUAL | MUTT_TOKEN_QUESTION);
+    mutt_extract_token(buf, s, MUTT_TOKEN_EQUAL | MUTT_TOKEN_QUESTION | MUTT_TOKEN_PLUS | MUTT_TOKEN_MINUS);
 
     bool bq = false;
     bool equals = false;
--- a/command_parse.c
+++ b/command_parse.c
@@ -1126,6 +1126,8 @@ enum CommandResult parse_set(struct Buffer *buf, struct Buffer *s,
 
     bool bq = false;
     bool equals = false;
+    bool increment = false;
+    bool decrement = false;
 
     struct HashElem *he = NULL;
     bool my = mutt_str_startswith(buf->data, "my_", CASE_MATCH);
@@ -1174,6 +1176,33 @@ enum CommandResult parse_set(struct Buffer *buf, struct Buffer *s,
 
       query = true;
       s->dptr++;
+    }
+    else if (*s->dptr == '+' || *s->dptr == '-')
+    {
+      if (prefix)
+      {
+        mutt_buffer_printf(err, "ERR04 can't use prefix when incrementing or "
+                                "decrementing a variable");
+        return MUTT_CMD_WARNING;
+      }
+
+      if (reset || unset || inv)
+      {
+        mutt_buffer_printf(err, "ERR05 can't set a variable with the '%s' command",
+                           set_commands[data]);
+        return MUTT_CMD_WARNING;
+      }
+      if (*s->dptr == '+')
+        increment = true;
+      else
+        decrement = true;
+
+      s->dptr++;
+      if (*s->dptr == '=')
+      {
+        equals = true;
+        s->dptr++;
+      }
     }
     else if (*s->dptr == '=')
     {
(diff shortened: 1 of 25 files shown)


jindraj:
As we discussed this feature on IRC with @flatcap we talked only about numeric config options. But after reading #243 I got an impression this should also support strings, better said the mentioned variable type "letter-list" for #202? That seems to me as a different issue, right?

Concating strings  is already possible:
```
set my_variable   = "${my_variable}CONCATED_STRING"
```
We can allowing += for concating. Seems okay to me. There's probably no use for -=.
```
set my_variable += "CONCATED_STRING":
```

gahr:
I wouldn't overload += for both string concatenation and integer addition. If we want to implement string concatenation, I'm sure we can come up with some other  sequence of symbols :)

gahr:
If I read the code correctly, `+=` and `-=` will just behave like `=` if the variable set is `my_*` or anything special like a path or a command. I think it would be better to fail and indicate that `+=` and `-=` are not supported, instead of ignoring them silently.

jindraj:
> I wouldn't overload += for both string concatenation and integer addition. If we want to implement string concatenation, I'm sure we can come up with some other sequence of symbols :)

 Let's stick with numeric options only for the time being. You're right, we can create i.e. `.=` for strings.

> If I read the code correctly, `+=` and `-=` will just behave like `=` if the variable set is `my_*` or anything special like a path or a command. I think it would be better to fail and indicate that `+=` and `-=` are not supported, instead of ignoring them silently.

I'll add the fails for other types + I would like my_vars to work with `+=` and `-=` too if it'll be possible.

jindraj:
Now after looking into code more deeply it seems harder than before.

config/set.h:
- add `int plus_equals()` and `int minus_equals()` into `struct ConfigSetType`

config/{long,numeric}.c:
- add {long,numeric}_{plus,minus}_equals() functions in long.c and number.c + register them by ConfigSetType.
- in other configtypes (config/{mtable,path,address,…}.c ) fill apropriate place with NULLs.

+ the glue functions? Am I at least heading right way?

flatcap:
> Now after looking into code more deeply it seems harder than before.

Don't worry, it isn't.
There are a couple of levels of magic in the middle to make things flexible.

> The Plan

The plan is to keep `parse_set()` as simple as possible.
- Define some types, e.g. string, number
- Define some operations, e.g. set, reset, plus_equals

The types are responsible for performing any validation and the actual actions.
On success, they send out notifications to let others know.

> ConfigSetType

Start by adding functions here.
For consistency, I suggest: `string_plus_equals()`, `string_minus_equals()`.

The first part refers to the type of data we're passing, `string` or `native`.
For now, ignore the native calls.

The original `string_set()` could have been called `string_equals()`.

> `config/number.c`, `config/long.c`

Next, create implementation for type number and long types.
Most of the code can be cut'n'pasted.

> `config/set.c`

This file ties together the `ConfigDef` definitions of variables with the type code.
It contains lots of wrappers in two flavours: `str` and `he`.
These refer to how we reference the config variable, by string name or by HashElem pointer.

The `string_set()` function has wrappers `cs_str_string_set()` and `cs_he_string_set()`.

I suggest adding:
- `cs_str_string_plus_equals`, `cs_he_string_plus_equals`
- `cs_str_string_minus_equals`, `cs_he_string_minus_equals`

Again, most can be cut'n'pasted.

> `config/subset.c`

One final layer.  The **set** deals with a single config variable, but it also supports inheritance.
The `NeoMutt`, `Account` and `Mailbox` objects each have their own `ConfigSubset`.

The `NeoMutt` subset is effectively the global config.
When we want to set config variables, we simply pick the 'scope' by using the `NeoMutt`, `Account` or `Mailbox` subset.

> `parse_set()`

When we parse a `set` command we have a config variable, `sidebar_width` and an operation `+=`.
- Lookup the variable to get its `ConfigDef` and hence its type
- Lookup the type to get its `ConfigSetType`
- Check if the type supports the operation

If so, then perform the operation using the subset functions on `NeoMutt.subset`.

flatcap:
I've added the glue functions and sample calls in `parse_set()` -- little more than cut'n'pasting.
What remains is to move the 'working' from `parse_set()` into the number/long functions.

flatcap:
The code looks good and works well.

**To do**:

- [ ] copy code to long
- [ ] prevent non-numeric types crashing
- [ ] allow `set config+=10` with no spaces
- [ ] tests
    - [ ] positive numbers
    - [ ] negative numbers
    - [ ] overflow
    - [ ] underflow
    - [ ] junk (non-numbers)
    - [ ] empty

command_parse.c
```diff
@@ -1175,6 +1177,41 @@ enum CommandResult parse_set(struct Buffer *buf, struct Buffer *s,
...
+                           set_commands[data]);
+        return MUTT_CMD_WARNING;
+      }
+
+      if (!(DTYPE(he->type) == DT_NUMBER || DTYPE(he->type) == DT_LONG))
+      {
+	mutt_buffer_printf(err, "Variable '%s' has an invalid type %d",
+			   buf->data, he->type);
+        return MUTT_CMD_WARNING;
+      }
```
flatcap, line 1201: This works, but it's not the _best_ place.
If we add support to another type, e.g. `DT_SLIST`, we'd have to change the code here too.

The next step down the call-tree is the **subset** code.
Again, this doesn't really know about types, only inheritance.
(Note: there are no `DT_*` constants in that code)

One step deeper, we reach the **set** code.
This understands the _types_ of data.

I think we should check here, just before we'd crash :-)
We can return:
- `CSR_ERR_INVALID` - operation failed
- A new `CSR_INV_` flag (or'd together) meaning: operation not supported
- An error message

PR #2300, opened 2020-04-27
sample.neomuttrc-starter: Do not echo promted password

SweetVishnya:
So entered password won't be saved in bash history.


gahr:
A bit bash-specific, isn't it?

SweetVishnya:
It works in `zsh` also but not `sh`. Maybe I should add it as second comment?

gahr:
Yeah, maybe leave both options + a comment saying that your version is available in some shells?

SweetVishnya:
OK, I'll do it now.

SweetVishnya:
Done.

PR #2304, opened 2020-04-29
fix(sidebar): use abbreviated path if possible

### LINKED ISSUE #2293: Regressions in sidebar rendering since 2020-04-24
Relevant piece of config:

```
set folder = "~/.mail/maximbaz"
set mbox_type = Maildir
set spoolfile = "+INBOX"
set record = "+Sent"
set postponed = "+Drafts"
set trash = "+Trash"

unmailboxes *
unvirtual-mailboxes *

mailboxes =INBOX =Archive/2019 =Drafts =Sent =Junk =Trash
```

## Expected Behaviour

In all previous versions including 2020-04-17:

image

## Actual Behaviour

In 2020-04-24:

image

## NeoMutt Version

```
$ neomutt -v
NeoMutt 20200424

System: Linux 5.6.6-arch1-1 (x86_64)
GPGME: 1.13.1
GnuTLS: 3.6.13
libnotmuch: 5.2.0
compression: lz4, zlib, zstd

Configure options: --prefix=/usr --sysconfdir=/etc --libexecdir=/usr/lib --gpgme --sqlite --autocrypt --lua --notmuch --gss --gnutls --sasl --with-ui=ncurses --with-idn2=/usr --disable-idn --idn2 --lmdb --kyotocabinet --gdbm --lz4 --zlib --zstd

Compilation CFLAGS: -march=x86-64 -mtune=generic -O2 -pipe -fno-plt -std=c99 -D_ALL_SOURCE=1 -D_GNU_SOURCE=1 -D__EXTENSIONS__ -I/usr/include -DNCURSES_WIDECHAR

Default options:

Compile options:
  -locales_hack +lua +lz4 +meta -mixmaster +nls +notmuch -openssl +pgp +regex 
  +sasl +smime +sqlite +start_color +sun_attachment +typeahead +zlib +zstd 
MAILPATH="/var/mail"
SENDMAIL="/usr/sbin/sendmail"
SYSCONFDIR="/etc"

```

## Extra Info

* Operating System and its version: Arch Linux
* Were you using 'screen' or 'tmux'? No
* Is your email local (maildir) or remote (IMAP)? Maildir

Austin-Ray:
## Please wait for confirmation in #2293 that this works before merging

After the refactor in commit d77c704ceb85f34061e1333ab9c8eb5a264a0685,
the sidebar only displays a mailbox's full path/URI except in the case
of 'sidebar_short_path'. The full path/URI is not preferable for end
users.

What broke during the refactor is how the variable 'display' is updated.
It's initialized to a mailbox's full path/URI and only updated if using
'sidebar_short_path'. However, after initialization, an abbreviation is
calculated to remove the 'folder' or protocol.

To resolve this issue, if there's an abbreviation, store it in the
'display' variable.

Fixes #2293

Using minimal config (`test/test` is just an empty file):
```
set folder="~/mail/austin@austinray.io"
set sidebar_visible=yes
unmailboxes *
mailboxes =INBOX =test/test
```

image

### DIFF
--- a/sidebar.c
+++ b/sidebar.c
@@ -1091,9 +1091,10 @@ static void draw_sidebar(struct MuttWindow *win, int num_rows, int num_cols, int
       m->msg_flagged = Context->mailbox->msg_flagged;
     }
 
+    const char *full_path = mailbox_path(m);
     display = m->name;
     if (!display)
-      display = mailbox_path(m);
+      display = full_path;
 
     const char *abbr = m->name;
     if (!abbr)
@@ -1113,7 +1114,9 @@ static void draw_sidebar(struct MuttWindow *win, int num_rows, int num_cols, int
 
     mutt_buffer_reset(&result);
 
-    if (C_SidebarFolderIndent)
+    // Don't indent if we were unable to create an abbreviation.
+    // Otherwise, the full path will be indent, and it looks unusual.
+    if (C_SidebarFolderIndent && mutt_str_strncmp(display, full_path, sizeof(display)))
     {
       if (C_SidebarComponentDepth > 0)
         depth -= C_SidebarComponentDepth;
--- a/sidebar.c
+++ b/sidebar.c
@@ -1101,6 +1101,10 @@ static void draw_sidebar(struct MuttWindow *win, int num_rows, int num_cols, int
     if (!abbr)
       abbr = abbrev_url(display, m->type);
 
+    // Use the abbreviation if we have one. The full path is not preferable.
+    if (abbr)
+      display = abbr;
+
     const char *last_part = abbr;
     int depth = calc_path_depth(abbr, C_SidebarDelimChars, &last_part);
 
--- a/sidebar.c
+++ b/sidebar.c
@@ -1109,14 +1109,19 @@ static void draw_sidebar(struct MuttWindow *win, int num_rows, int num_cols, int
     const char *last_part = abbr;
     int depth = calc_path_depth(abbr, C_SidebarDelimChars, &last_part);
 
+    // At this point, we don't have an abbreviation so let's keep track
+    // before using short path.
+    bool no_abbr = mutt_str_strncmp(display, full_path, sizeof(display));
     if (C_SidebarShortPath)
+    {
       display = last_part;
+    }
 
     mutt_buffer_reset(&result);
 
     // Don't indent if we were unable to create an abbreviation.
     // Otherwise, the full path will be indent, and it looks unusual.
-    if (C_SidebarFolderIndent && mutt_str_strncmp(display, full_path, sizeof(display)))
+    if (C_SidebarFolderIndent && no_abbr)
     {
       if (C_SidebarComponentDepth > 0)
         depth -= C_SidebarComponentDepth;


gahr:
Thanks, it does look fine on my tests - I don't usually use the sidebar so my use cases are constructed a bit ad-hoc for this testing. So at the end of the day I think we could summarize the behaviour of sidebar as follows. Please correct me if I'm wrong.

* Given mailboxes both local and on different servers, with different separators
```
imaps://server1/INBOX 
imaps://server1/Parent
imaps://server1/Parent.Child
imaps://server2/INBOX
imaps://server2/Uncle
imaps://server2/Uncle.Nephew
/home/foobar/Mail/gmail/INBOX
/home/foobar/Mail/gmail/[GMail]/All Mail
```

* Given the following fixes settings
sidebar_delim_chars = "./"
folder = /home/foobar/Mail/gmail/

* What is displayed

| sidebar_short_path | sidebar_folder_indent | result |
| ------------------- | ---------------------- | ------ |
| no  | no  | INBOX<br>[GMail]/All Mail<br>INBOX<br>Parent<br>Parent.Child<br>INBOX<br>Uncle<br>Uncle.Nephew |
| yes | no  | INBOX<br>All Mail<br>INBOX<br>Parent<br>Child<br>INBOX<br>Uncle<br>Nephew |
| no   | yes | INBOX<br>&nbsp;&nbsp;&nbsp;&nbsp;[GMail]/All Mail<br>INBOX<br>Parent<br>&nbsp;&nbsp;&nbsp;&nbsp;Parent.Child<br>INBOX<br>Uncle<br>&nbsp;&nbsp;&nbsp;&nbsp;Uncle.Nephew |
| yes | yes |    INBOX<br>&nbsp;&nbsp;&nbsp;&nbsp;All Mail<br>INBOX<br>Parent<br>&nbsp;&nbsp;&nbsp;&nbsp;Child<br>INBOX<br>Uncle<br>&nbsp;&nbsp;&nbsp;&nbsp;Nephew | 

There are a couple of remaining issues:

1. I'm not sure about is how `sidebar_component_depth` is supposed to interact. Setting it to anything but 0 makes the indentation disappear.

2. If we merge this fix, I think we should also review the documentation for the various `sidebar_` variables. They tend to reference $fol

3. Most importantly, `$folder` only affects local mailboxes and changing it has strange effects. If I set it to, say, `/var/mail`, the above turns out as follows:

| sidebar_short_path | sidebar_folder_indent | result |
| ------------------- | ---------------------- | ------ |
| no  | no  | /home/foobar/Mail/gmail/INBOX<br>/home/foobar/Mail/gmail/[GMail]/All Mail<br>INBOX<br>Parent<br>Parent.Child<br>INBOX<br>Uncle<br>Uncle.Nephew |
| yes | no  | INBOX<br>All Mail<br>INBOX<br>Parent<br>Child<br>INBOX<br>Uncle<br>Nephew |
| no   | yes | &nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;/home/foobar/Mail/gmail/INBOX<br>&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;/home/foobar/Mail/gmail/[GMail]/All Mail<br>INBOX<br>Parent<br>&nbsp;&nbsp;&nbsp;&nbsp;Parent.Child<br>INBOX<br>Uncle<br>&nbsp;&nbsp;&nbsp;&nbsp;Uncle.Nephew |
| yes | yes | &nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;INBOX<br>&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;All Mail<br>INBOX<br>Parent<br>&nbsp;&nbsp;&nbsp;&nbsp;Child<br>INBOX<br>Uncle<br>&nbsp;&nbsp;&nbsp;&nbsp;Nephew| 

So two things happen: i) the local mailbox is not shortened "as expected", and ii) it gets indented way too much

I am not sure how much of this you want to fix in this PR :)

jindraj:
> 1. I'm not sure about how `sidebar_component_depth` is supposed to interact. Setting it to anything but 0 makes the indentation disappear.

It flattens the indentation by specified number of levels (from the left).
sidebar_component_depth=0
```
|sample
|  cars
|    audi
|    bentley
|    chrysler
|    daimler
```
sidebar_component_depth=1
```|sample
|cars
|  audi
|  bentley
|  chrysler
|  daimler
```
sidebar_component_depth=2
```|sample
|cars
|audi
|bentley
|chrysler
|daimler
```

Austin-Ray:
> I am not sure how much of this you want to fix in this PR :)

I'll fix as many new sidebar bugs as possible!

> i) the local mailbox is not shortened "as expected", 

Comparing to the following releases: 2018-07-16, 2019-12-17, and 2020-03-20, if we set `folder` to `/var/mail` then we will see the full path of a mailbox outside of `folder`. It's safe to say this not new behavior from the refactoring. It's outside the scope of this PR. We should create a discussion issue to keep track of it and determine a solution.

> ii) it gets indented way too much

It looks like the indentation is blindly going off the number of delimiter characters without considering if it should be indented. Hopefully will be a simple fix.

gahr:
> if we set folder to /var/mail then we will see the full path of a mailbox outside of folder. 

Yes, except this used to be the case for all mailboxes (IMAP) included. Now $folder only affects local mailboxes. This is a change in behaviour that I think we should consider as such. If we do want it, then I think we should apply it consistently.

Austin-Ray:
> 
> > ii) it gets indented way too much
> 
> It looks like the indentation is blindly going off the number of delimiter characters without considering if it should be indented. Hopefully will be a simple fix.

Pushed a commit to fix this. The solution is hacky IMO.

>
> > if we set folder to /var/mail then we will see the full path of a mailbox outside of folder.
>
> Yes, except this used to be the case for all mailboxes (IMAP) included. Now $folder only affects local mailboxes. This is a change in behaviour that I think we should consider as such. If we do want it, then I think we should apply it consistently.

I'm not sure if I'm understanding this correctly. Here's what I'm interpreting:

In the refactoring, the introduction of `abbrev_url()` causes IMAP maildirs to be abbreviated regardless of the value of `folder`. So using this config:

```
set folder="imaps://somehost/"
mailboxes =INBOX imaps://someotherhost/INBOX
```

Results in a sidebar like:

```
INBOX
imaps://someotherhost/INBOX
```

But after the refactoring, it results in:

```
INBOX
INBOX
```

Is this a correct understanding?

gahr:
> Pushed a commit to fix this. The solution is hacky IMO.

That works for `short_path = no` - thanks - `short_path = yes` is still messy. 

> Is this a correct understanding?

Yes, that's the way I understand it too. So the fix which is currently in master fixed half of the problem (only for network mailboxes), so if earlier we had a consistent ugly behaviour, now we have an inconsistent behaviour with is nice for network mailboxes and still ugly for local ones.

I think I'm fine with leaving local mailboxes unfixed for the time being, but we should at least update the docs so the variables reflect their actual behaviour.

Austin-Ray:
> That works for short_path = no - thanks - short_path = yes is still messy.

But the bug is slightly smaller now! :)
I'll try to resolve this on my lunch break.

Austin-Ray:
Updated the PR and it looks to resolve the `sidebar_short_path=yes` case. Hopefully it's not messy from the speed coding during lunch!

gahr:
That works, at the cost of not being able to indent local mailboxes when they don't start with `$folder`. I guess we'll have to live with it for the time being.

Austin-Ray:
> That works, at the cost of not being able to indent local mailboxes when they don't start with $folder. I guess we'll have to live with it for the time being

Was that previous behavior? I tried testing that on an older version, but didn't see that. Do you have a minimum config to check that behavior?

On April 29, 2020 12:19:00 PM EDT, Pietro Cerutti <notifications@github.com> wrote:
>That works, at the cost of not being able to indent local mailboxes
>when they don't start with `$folder`. I guess we'll have to live with
>it for the time being.
>
>-- 
>You are receiving this because you authored the thread.
>https://github.com/neomutt/neomutt/pull/2304#issuecomment-621316007

-- 

gahr:
Not a regression, I guess I was just restating the slightly ambiguous / inconsistent behaviour of local vs. network folders, for no good reason in my opinion.

It think this is a great improvement over the current master!

Austin-Ray:
> Not a regression, I guess I was just restating the slightly ambiguous / inconsistent behaviour of local vs. network folders, for no good reason in my opinion.

I just pushed a commit that should resolve the issue mentioned. Please let me know if I've understood the problem!

gahr:
> > Not a regression, I guess I was just restating the slightly ambiguous / inconsistent behaviour of local vs. network folders, for no good reason in my opinion.
> 
> I just pushed a commit that should resolve the issue mentioned. Please let me know if I've understood the problem!

Yep that goes back where we were: `short_path = no` only shortens the mailboxes of which `folder` is a prefix. I don't think that's correct.

I think the direction to go is the opposite: what I would like to see - but I think it might well be out of scope - is `sidebar_short_path = yes` shortening all mailboxes (network and local) independently of folder. This might involve a linear search of all mailboxes when adding a mailbox, to figure what is the common subpath.

I think this PR is good without the last commit.

Austin-Ray:
> I think this PR is good without the last commit.

I think we should keep the last commit for now since that restores the old, expected behavior.  You're right about improving the logic so `folder` is irrelevant, but primary concern right now is make a hotfix release for the sidebar.

We can improve the logic once users aren't being affected.

On April 30, 2020 2:58:29 AM EDT, Pietro Cerutti <notifications@github.com> wrote:
>> > Not a regression, I guess I was just restating the slightly
>ambiguous / inconsistent behaviour of local vs. network folders, for no
>good reason in my opinion.
>> 
>> I just pushed a commit that should resolve the issue mentioned.
>Please let me know if I've understood the problem!
>
>Yep that goes back where we were: `short_path = no` only shortens the
>mailboxes of which `folder` is a prefix. I don't think that's correct.
>
>I think the direction to go is the opposite: what I would like to see -
>but I think it might well be out of scope - is `sidebar_short_path =
>yes` shortening all mailboxes (network and local) independently of
>folder. This might involve a linear search of all mailboxes when adding
>a mailbox, to figure what is the common subpath.
>
>I think this PR is good without the last commit.
>
>-- 
>You are receiving this because you authored the thread.
>https://github.com/neomutt/neomutt/pull/2304#issuecomment-621650628

-- 

gahr:
Isn't this the worst of both worlds?

We'd get back to not having #2272 fixed + we would introduce another difference which is in my opinion worse.

Can you please double check the two tables I put in https://github.com/neomutt/neomutt/pull/2304#issuecomment-621042143 ? They show how the sidebar looks like in the two cases where folder is and is not a prefix of the local mailbox. That's before your latest commit. I don't think your latest commit makes it better.

I will re-review the whole thing when I have some more time and try to make tables for how it is and how it should be.

Austin-Ray:
> We'd get back to not having #2272 fixed + we would introduce another difference which is in my opinion worse.

Ohhh that makes sense. I understand now. I've dropped the last commit so you're good to review!

Issue #2110, opened 2020-02-05
Neomutt unable to open IMAP folders other than +INBOX


steevmi1:
## Expected Behaviour

Able to change from the +INBOX folder to any other folder while connected to an IMAP server.

## Actual Behaviour

The folder does "open", but shows no messages. Either one or more single-digit numbers or non-ASCII characters print (the same number, on one or more lines). After this, the +INBOX is also similarly unaccessible.

Neomutt doesn't crash consistently due to this, but seems to crash more frequently. Core file is dropped on OpenBSD, not on OS X. 

## Steps to Reproduce

Connect to IMAP server
Switch to any folder

## How often does this happen?

- Always

When connecting to IMAP hosted at the ISP pair.com, using IMAP from google.com is so far unaffected.

## When did it start to happen?

- When I upgraded

With the 2019-11-29 release.

## NeoMutt Version

```
NeoMutt 20191207
Copyright (C) 1996-2016 Michael R. Elkins and others.
NeoMutt comes with ABSOLUTELY NO WARRANTY; for details type 'neomutt -vv'.
NeoMutt is free software, and you are welcome to redistribute it
under certain conditions; type 'neomutt -vv' for details.

System: OpenBSD 6.6 (amd64)
ncurses: ncurses 5.7.20081102 (compiled with 5.7.20081102)
libiconv: 1.16
libidn2: 2.3.0 (compiled with 2.3.0)
GPGme: 1.13.1
hcache backends: kyotocabinet

Configure options: --mandir=/usr/local/man --with-ui=ncurses --docdir=/usr/local/share/doc/neomutt --with-lock=flock --idn=0 --idn2 --fmemopen --ssl --autocrypt --gpgme --sqlite --sasl --kyotocabinet

Compilation CFLAGS: -O2 -pipe -g  -I/usr/local/include -std=c99 -D_ALL_SOURCE=1 -D_GNU_SOURCE=1 -D__EXTENSIONS__ -DNCURSES_WIDECHAR

Default options:
  +attach_headers_color +compose_to_sender +compress +cond_date +debug 
  +encrypt_to_self +forgotten_attachments +forwref +ifdef +imap +index_color 
  +initials +limit_current_thread +multiple_fcc +nested_if +new_mail +nntp +pop 
  +progress +quasi_delete +regcomp +reply_with_xorig +sensible_browser +sidebar 
  +skip_quoted +smtp +status_color +timeout +tls_sni +trash 

Compile options:
[...]

PR #2307, opened 2020-04-30
Make sure buffers are initialized on error

### LINKED ISSUE #2145: Messy index lines
## Expected Behaviour
Index displayed correctly.

## Actual Behaviour

Everything other looks okay but index:
```
│�L�D�U
│�L�D�U
│�L�D�U
│ |iE�U
│ |iE�U
│ |iE�U
│ |iE�U
│ |iE�U
│ |iE�U
│�L�D�U
│�L�D�U
│�L�D�U
│�L�D�U
│�L�D�U
│�L�D�U
│�L�D�U
│�L�D�U
```

(Middle lines changed because I moved selection. I can “reset” lines with <kbd>Ctrl</kbd>+<kbd>L</kbd>.)

## Steps to Reproduce

I don’t know anything sure about reproduction (yet).
I’m using `i3` and I sometimes face this after recalling `neomutt` from scratchpad (but maybe it has nothing to do with it).

- Sometimes

First saw _about_ a month ago.

## NeoMutt Version

```
NeoMutt 20191207

System: Linux 5.4.15-arch1-1 (x86_64)
ncurses: ncurses 6.1.20180127 (compiled with 6.1.20180127)
GPGme: 1.13.1
libnotmuch: 5.2.0
hcache backends: kyotocabinet, gdbm, lmdb

Configure options: --prefix=/usr --sysconfdir=/etc --libexecdir=/usr/lib --gpgme --lua --notmuch --gss --gnutls --sasl --with-ui=ncurses --with-idn2=/usr --disable-idn --idn2 --lmdb --kyotocabinet --gdbm

Compilation CFLAGS: -march=x86-64 -mtune=generic -O2 -pipe -fno-plt -std=c99 -fno-delete-null-pointer-checks -D_ALL_SOURCE=1 -D_GNU_SOURCE=1 -D__EXTENSIONS__ -I/usr/include -DNCURSES_WIDECHAR

Default options:

Compile options:
  -locales_hack +lua +meta -mixmaster +nls +notmuch -openssl +pgp +sasl +smime 
  -sqlite +start_color +sun_attachment +typeahead 
MAILPATH="/var/mail"
SENDMAIL="/usr/sbin/sendmail"
SYSCONFDIR="/etc"

```

## Extra Info

```
$ echo $TERM
rxvt-unicode-256color
```

Changing `index_format` didn’t help.

zsugabubus:
Because of lack of proper error propagation functions could return
without touching `buf`.

* **What does this PR do?**
Initializes such buffers.

Fixes #2145. (And absolutely not sure, but maybe #2110 (at least the garbage).)

### DIFF
--- a/index.c
+++ b/index.c
@@ -825,6 +825,8 @@ static int main_change_folder(struct Menu *menu, int op, struct Mailbox *m,
  */
 void index_make_entry(char *buf, size_t buflen, struct Menu *menu, int line)
 {
+  buf[0] = '\0';
+
   if (!Context || !Context->mailbox || !menu || (line < 0) ||
       (line >= Context->mailbox->email_max))
     return;
@@ -927,7 +929,7 @@ int index_color(int line)
  */
 void mutt_draw_statusline(int cols, const char *buf, size_t buflen)
 {
-  if (!buf)
+  if (!buf || !stdscr)
     return;
 
   size_t i = 0;
@@ -942,9 +944,6 @@ void mutt_draw_statusline(int cols, const char *buf, size_t buflen)
     int first;
     int last;
   } *syntax = NULL;
-
-  if (!buf || !stdscr)
-    return;
 
   do
   {
--- a/sidebar.c
+++ b/sidebar.c
@@ -294,10 +294,13 @@ static const char *sidebar_format_str(char *buf, size_t buflen, size_t col, int 
 static void make_sidebar_entry(char *buf, size_t buflen, int width,
                                const char *box, struct SbEntry *sbe)
 {
-  if (!buf || !box || !sbe)
+  if (!buf)
     return;
 
-  mutt_str_strfcpy(sbe->box, box, sizeof(sbe->box));
+  if (box && sbe)
+    mutt_str_strfcpy(sbe->box, box, sizeof(sbe->box));
+  else
+    buf[0] = '\0';
 
   mutt_expando_format(buf, buflen, 0, width, NONULL(C_SidebarFormat),
                       sidebar_format_str, (unsigned long) sbe, MUTT_FORMAT_NO_FLAGS);


index.c
```diff
@@ -927,9 +929,6 @@ int index_color(int line)
  */
 void mutt_draw_statusline(int cols, const char *buf, size_t buflen)
 {
-  if (!buf)
```
zsugabubus, line 930: This check is redundant. There is one more 10 lines below.

gahr: I’d rather fail earlier than later

gahr:
Thanks that looks good, I’ll have a closer look later

index.c
zsugabubus: Ah, just 10 lines and happens nothing in between.

Then what about moving both checks here?

gahr: Definitely!

sidebar.c
```diff
@@ -294,10 +294,13 @@ static const char *sidebar_format_str(char *buf, size_t buflen, size_t col, int
...
 {
-  if (!buf || !box || !sbe)
+  if (!buf)
     return;
 
-  mutt_str_strfcpy(sbe->box, box, sizeof(sbe->box));
+  if (box && sbe)
+    mutt_str_strfcpy(sbe->box, box, sizeof(sbe->box));
+  else
+    buf[0] = '\0';
```
gahr, line 303: Is it correct to keep going if either `box` or `sbe` are not set?

How about just setting `buf[0] = '\0'` at the beginning of the function no matter what, then keeping the old check and return if either `buf`, `box`, or `sbe` are NULL?

zsugabubus: What about padding?

index.c
zsugabubus: Okay.

(I just wondered–because I see at lots of places–that are these `buf != NULL` checks really necessary? The most places I think `buf` is some array on stack, but if it is `malloc()`ed, the `NULL` check should be done immediately after allocation, so `NULL`s should never be passed to any functions.)

sidebar.c
gahr: What about it? I don’t think you can do much if any of those params are null.

How about just making sure buf is null-terminated and return as early as possible?

flatcap: It turns out that the checks are overkill; none of them can be `NULL`.
There's only one caller and:
- `buf` is `char str[256];`
- `box` is `struct Buffer result = mutt_buffer_make(256);`
- `sbe` is dereferenced before the call: `if (entry->is_hidden)`

Issue #2309, opened 2020-05-02
Test failure with version 2020-05-01


peterhoeg:
```
neomutt> FAILED: 1 of 436 unit tests has failed.
neomutt> make: *** [Makefile:1333: test] Error 1
builder for '/nix/store/7n8d9anp86xwmas654rcnp98yqzb50qq-neomutt-20200501.drv' failed with exit code 2; last 10 log lines:
    Case imaps://foobar@gmail.com@imap.gmail.com/Отправленные письма:
      url_parse.c:245: Check !((!!url) ^ (!!test[i].valid))... failed
        Expected: not NULL
        Actual  : NULL
  Test test_url_pct_decode...                     [ OK ]
  Test test_url_pct_encode...                     [ OK ]
  Test test_url_tobuffer...                       [ OK ]
  Test test_url_tostring...                       [ OK ]
  FAILED: 1 of 436 unit tests has failed.
  make: *** [Makefile:1333: test] Error 1
```
This was not an issue with 2020-04-17.

```
▶ nix-info -m
 - system: `"x86_64-linux"`
 - host os: `Linux 5.6.7, NixOS, 20.09.git.79b77e60dc0M (Nightingale)`
 - multi-user?: `yes`
 - sandbox: `yes`
 - version: `nix-env (Nix) 2.3.4`
 - channels(root): `"home-manager, nixos-20.09pre215947.82b54d49066"`
 - channels(peter): `"home-manager"`
 - nixpkgs: `/home/peter/src/active/nixpkgs_unstable`
```

flatcap:
Hmm... that's peculiar!

Please can you try this change to `test/url/url_parse.c`
```diff
-  setlocale(LC_ALL, "en_US.UTF-8");
+  setlocale(LC_ALL, "C.UTF-8");
```

peterhoeg:
Yep, that did it.

PR #2314, opened 2020-05-04
acutest updates

flatcap:
- Update to [latest acutest.h](https://github.com/mity/acutest/commit/cde0c8e699c843b70c978a1b66a11b346c933957)
- Use the `test_init()` function for all tests:
  - Check `$NEOMUTT_TEST_DIR` is set
  - Set the locale to `en_US.UTF-8` or `C.UTF-8`

Fixes: #2309

### DIFF
--- a/test/common.c
+++ b/test/common.c
@@ -23,6 +23,7 @@
 #define TEST_NO_MAIN
 #include "config.h"
 #include "acutest.h"
+#include <locale.h>
 #include <sys/stat.h>
 #include <sys/types.h>
 #include <unistd.h>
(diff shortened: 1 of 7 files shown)


test/common.c
```diff
@@ -71,8 +72,18 @@ void test_common(void)
     TEST_MSG("Test dir '%s' isn't a directory\n", path);
     goto done;
   }
+
+  if (!TEST_CHECK((setlocale(LC_ALL, "en_US.UTF-8") != NULL) ||
+                  (setlocale(LC_ALL, "C.UTF-8") != NULL)))
```
gahr, line 77: Do we need `en_US` or can we just use `C`? Is there a use case there the former is available and the latter is not?

flatcap: > Do we need `en_US`

Hmm... probably not.
I've dropped it (in a separate commit in case we need to revert it).

Issue #2303, opened 2020-04-28
imap: non-existing unsubscribed folder causes issues


gahr:
As discovered in #2110 , sometimes being subscribed to non-existing IMAP folder causes issues. I suspect this is due to us not handling a `NO` response gracefully enough, depending on when that comes in.

Even though it is not clear yet whether all instances of the "garbled index" problem boil down to this cause, I'll use this issue to investigate handling the non-existing unsubscribed folder problem.

gahr:
Example:

```
a1 LSUB "" "*"
...               
* LSUB () "." Lists.mit                                                            
* LSUB () "." Lists.mit.plv                                                        
...
a2 STATUS "Lists.mit" (UIDNEXT UIDVALIDITY UNSEEN RECENT MESSAGES)               
a2 NO Mailbox doesn't exist: Lists.mit (0.009 + 0.000 + 0.008 secs).
```

This is what we don't handle.

gahr:
I can't reproduce this problem. It looks to me like NeoMutt now handles a non-existent mailbox just fine:
```
[13:09:12]<M> IMAP command failed: a0045 NO Mailbox doesn't exist: Foobar
```
Then everything proceeds smoothly.

PR #2317, opened 2020-05-07
Inform but don't fail on IMAP NO/BAD responses

gahr:
Issue #2303

This is a first approach a fixing the issue. Instead of returning an error that ultimately ends up being handled as a fatal error + mailbox close, let's just show a message to inform the user that one of the IMAP commands has failed.
```
IMAP command failed: a0306 NO Mailbox doesn't exist: FooBar (0.001 + 0.000 secs).
```

I cannot foresee all implications of not returning an error, but at least we'd get some feedback and data to work on.


zsugabubus:
I tried out NeoMutt from master and I was unable to access my mailbox. Doing a bisect leaded here. [Log.](https://github.com/neomutt/neomutt/issues/2331) Shall I open a bug report?

PR #2319, opened 2020-05-07
Window notifications

flatcap:
In ["Event-Driven Sidebar"](https://neomutt.org/dev/roadmap/event-sidebar), I map out the route to splitting up the Windows and encapsulating them.
Many of the required notifications, e.g. Config, Colour, are already in place.
This PR adds Window notifications.

- e0ecb92ada window: rename dlg_index
  unify naming

- c17e983be4 window: name and notify
  ensure all user-visible windows are named (debug only) and have notifications support

- 3ebcd28900 window: set_visible
  helper function to make a window (and its children) visible or hidden

- a459d75001 notify: windows
  send out notifications on window changes

- 0fa17ab368 debug: notify windows
  update debugging to dump all window events

---

The debug logging looks like this:
```
[pager] visible 
[pager bar] visible 
[index] taller [10->39] 
[index bar] moved (C20,R10)->(C20,R39) 
```

### DIFF
--- a/addrbook.c
+++ b/addrbook.c
@@ -242,17 +242,30 @@ void mutt_alias_menu(char *buf, size_t buflen, struct AliasList *aliases)
   struct MuttWindow *dlg =
       mutt_window_new(MUTT_WIN_ORIENT_VERTICAL, MUTT_WIN_SIZE_MAXIMISE,
                       MUTT_WIN_SIZE_UNLIMITED, MUTT_WIN_SIZE_UNLIMITED);
-#ifdef USE_DEBUG_WINDOW
-  dlg->name = "aliases";
-#endif
   dlg->type = WT_DIALOG;
+  dlg->notify = notify_new();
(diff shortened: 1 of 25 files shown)


gahr:
> c17e983 window: name and notify

Does it make sense to enhance `mutt_window_new` to take a name and possibly a notify to set as parent?

flatcap:
> Does it make sense to enhance `mutt_window_new` to take
> a name

Probably not, it's debug-only.  It's useful now, but it may not stick around.

> a notify to set as parent?

Sounds reasonable, but there are a couple of gotchas, atm.

Not all windows have a `notify`.
The 'container' windows don't have an active role in events, so some of them have been skipped (the child talks to the grandparent).

Some windows don't get a notify parent until they're inserted into a parent.
e.g. a dialog window when it's `dialog_push()`d.

But all this might change once I've converted a few more dialogs.

PR #2326, opened 2020-05-09
build: fix out-of-tree git builds

nabijaczleweli:
The generated Makefile used to assume that the git repository was in the working directory, if at all present; this broke, e.g., Debian package builds from a git repository, since the package is built with `cd obj-$TRIPLE && ../configure`

The `$(realpath)` invocation subtly fixes vendoring the neomutt-test-files repository as well, since it strips files that do not exist (like `file/missing_symlink`) that make errored on previously, since `git_ver.c` depended on them

Also the generated git version is present in the output now.

### DIFF
--- a/Makefile.autosetup
+++ b/Makefile.autosetup
@@ -56,7 +56,7 @@ UNINSTALL_TARGETS=	@UNINSTALL_TARGETS@
 
 VPATH=		@VPATH@
 
-ALL_FILES!=	(cd $(SRCDIR) && git ls-files 2>/dev/null) || true
+ALL_FILES!=	(cd $(SRCDIR) && git ls-files | xargs realpath) 2>/dev/null || true
 
 ###############################################################################
 # neomutt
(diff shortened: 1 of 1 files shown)


gahr:
Please don't use gmake-specific features.

gahr:
Does this work as well for you?

```diff
diff --git a/Makefile.autosetup b/Makefile.autosetup
index 34cc7160e..881e8f5f4 100644
--- a/Makefile.autosetup
+++ b/Makefile.autosetup
@@ -56,7 +56,7 @@ UNINSTALL_TARGETS=    @UNINSTALL_TARGETS@
 
 VPATH=         @VPATH@
 
-ALL_FILES!=    (cd $(SRCDIR) && git ls-files 2>/dev/null) || true
+ALL_FILES!=    (git -C $(SRCDIR) ls-files 2>/dev/null | sed 's|^|$(SRCDIR)/|') || true
 
 ###############################################################################
 # neomutt
@@ -568,7 +568,7 @@ $(PGPEWRAP): $(PGPEWRAPOBJS)
 
 # generated
 git_ver.c: $(ALL_FILES)
-       version=`git describe --dirty --abbrev=6 --match "20[0-9][0-9][0-9][0-9][0-9][0-9]" 2> /dev/null | \
+       version=`git -C $(SRCDIR) describe --dirty --abbrev=6 --match "20[0-9][0-9][0-9][0-9][0-9][0-9]" 2> /dev/null | \
                sed -e 's/^[0-9]\{8\}//; s/-g\([a-z0-9]\{6\}\)/-\1/'`; \
        echo 'const char *GitVer = "'$$version'";' > $@.tmp; \
        cmp -s $@.tmp $@ || mv $@.tmp $@; \
```

nabijaczleweli:
Took some tinkering, but apparently `realpath` needed `-e` to not strip symlinks-to-nonexistent (which was the only thing the `$(realpath)` call in my version had done).
Updated the patch.

gahr:
`realpath -e` doesn’t exist everywhere. If the only reason to have it is to avoid errors on non-existing symlinks because you vendor neomutt-test-files, then I think it belongs with downstream patches.

Do you actually have to generate git_ver.c at all?

nabijaczleweli:
You do make a good point, that's probably what I'll end up doing.

NB: `git ls-files` still returns the index and, therefore, make fails if a file is deleted without updating the index, for no good reason. A `find` solution would probably work better for what `ALL_FILES` is supposed to mean, but.

gahr:
Thanks - ALL_FILES can probably be reworked a bit, elsewhere :)

PR #2332, opened 2020-05-11
Fix crash inside empty query menu

zsugabubus:
`mutt_mem_calloc()` does not alloc and returns NULL on zero items. In this
case `query_table` must not be dereferenced.

* **What does this PR do?**
Fixes a crash.


flatcap:
Good catch, @zsugabubus, but I think there's one more.
In the `if (!tagged)` section at the end of `query_menu()`, there's another unchecked `query_table`.

zsugabubus:
Oh, you are right. My bad. Added.

flatcap:
Rebased and squashed

PR #2343, opened 2020-05-14
Address Book tweaks

flatcap:
Three simple commits to improve the Address Book.

- ace468d9f fix formatting of address
  Bug-fix: Names with accented characters had the wrong width

- a6184dd6c add comment field
  Add a free-form comment field, after `# `, at the end of an alias,
  e.g.  `alias jim James Smith <js@example.com> # Pointy-haired boss`
  Plus an expando `%c` to display it.

- f8e7a1a3e add comment to alias_format
  Adjust the default `$alias_format` to include the comment expando.
  Priority has been given to the existing fields.
  Note: On an 80-char screen, the comment won't be visible.

- 3cd5411a0 alias: add option to save comment
  When saving an alias, allow an optional comment field

### DIFF
--- a/alias.c
+++ b/alias.c
@@ -751,6 +751,7 @@ void mutt_alias_free(struct Alias **ptr)
 
   mutt_alias_delete_reverse(a);
   FREE(&a->name);
+  FREE(&a->comment);
   mutt_addrlist_clear(&(a->addr));
   FREE(ptr);
 }
--- a/alias.c
+++ b/alias.c
@@ -473,8 +473,20 @@ retry_name:
   mutt_str_replace(&TAILQ_FIRST(&alias->addr)->personal, buf);
 
   buf[0] = '\0';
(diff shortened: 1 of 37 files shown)


alias.c
```diff
@@ -472,9 +472,21 @@ void mutt_alias_create(struct Envelope *cur, struct AddressList *al)
...
+  if (mutt_get_field(_("Comment: "), buf, sizeof(buf), MUTT_COMP_NO_FLAGS) == 0)
+    mutt_str_replace(&alias->comment, buf);
+
   buf[0] = '\0';
   mutt_addrlist_write(&alias->addr, buf, sizeof(buf), true);
-  snprintf(prompt, sizeof(prompt), _("[%s = %s] Accept?"), alias->name, buf);
+  if (alias->comment)
+  {
+    snprintf(prompt, sizeof(prompt), "[%s = %s # %s] %s", alias->name, buf,
+             NONULL(alias->comment), _("Accept?"));
```
gahr, line 484: Can't be NULL

flatcap: good point.  refactoring artifact

PR #2351, opened 2020-05-19
Rename Quiet to Verbose

flatcap:
Invert the `Mailbox->quiet` logic to reduce the number of negative `if`s.

There are ~60 uses of the `quiet` flag, but the vast majority are of the form:

```c
if (!m->quiet)
  mutt_message();
```

This is replaced by:

```c
if (m->verbose)
  mutt_message();
```

---

Ultimately, this flag shouldn't be part of `Mailbox` but removing it will take some time.

### DIFF
--- a/query.c
+++ b/query.c
@@ -144,11 +144,11 @@ static void query_free(struct Query **query)
 
 /**
  * run_query - Run an external program to find Addresses
- * @param s     String to match
- * @param quiet If true, don't print progress messages
+ * @param s       String to match
+ * @param verbose If true, print progress messages
  * @retval ptr Query List of results
  */
-static struct Query *run_query(char *s, bool quiet)
+static struct Query *run_query(char *s, bool verbose)
 {
   FILE *fp = NULL;
   struct Query *first = NULL;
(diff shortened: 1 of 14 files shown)


query.c
```diff
@@ -144,11 +144,11 @@ static void query_free(struct Query **query)
...
 /**
  * run_query - Run an external program to find Addresses
- * @param s     String to match
- * @param quiet If true, don't print progress messages
+ * @param s       String to match
+ * @param verbose If true, don't print progress messages
  * @retval ptr Query List of results
  */
-static struct Query *run_query(char *s, int quiet)
+static struct Query *run_query(char *s, bool verbose)
```
gahr, line 151: How about the callers?

query.c
```diff
@@ -144,11 +144,11 @@ static void query_free(struct Query **query)
 
 /**
  * run_query - Run an external program to find Addresses
- * @param s     String to match
- * @param quiet If true, don't print progress messages
+ * @param s       String to match
+ * @param verbose If true, don't print progress messages
```
gahr, line 148: I think `If false,...`

query.c
flatcap: Argh!  I'm losing my marbles.  Pushed.

query.c
flatcap: `s/don't //`

flatcap:
rebased

PR #2356, opened 2020-05-23
improve french translation

Aluriak:
Hello !

Thanks for that project, i will finally switch to CLI thanks to you.
In order to engage a little with that project, and since i'm speaking french,
i made some translation. First time of my life i'm doing software translation.

Follow few details on my translations.

## Regarding global work and existing data

- i currently didn't add any comment. Should i add anything to the header comment ?
- i didn't translated following names: *Autocrypt* (a technology name, i guess), *Newsgroup*.
- i noted that sometimes autocrypt is in lowercase (e.g. «*Create an initial autocrypt account?*»), sometimes it is entitled. I reproduced the casing in french.
- some translations make use of an additionnal `%s`, such as «Subscribe» that is translated to «S'abonner à %s».
- i observed that «peer» is translated «machine distante», i.e. «remote machine», whereas the french «pair» would be a litteral translation.
- the french translation «Le serveur SMTP ne supporte pas l'authentification» was used for many different english sentences, all linked to SMTP or IMAP server error handling or folder tagging. I guess this was very confusing for french users.
- «tag» was translated by the french «objet», like in «subject line». I replaced it by «étiquette», which is the standard way to name tags.
- i only noticed the `, fuzzy` annotation after making modifications. Hence, i probably forgot to remove some of them. However, that motivated me to go through the whole file a second time to look at the fuzzy annotated translations, and fix some of them.
- added a trailing `%s` to the string «SHA256 Fingerprint: » (line 1631), to mimick the previous «SHA1 Fingerprint: %s».

## Regarding french language specificities

- the correct term in french for encryption is not the false friend «encrypter», but «chiffrer». [French source](https://chiffrer.info/).
- i therefore fixed erroneous french translations, such as «chiffrage».
- i applied (and fixed on existing translations) french grammar rules regarding punctuation : spaces before colons and interrogation marks.

## Regarding difficulties in translations themselves

- i find it quite hard to properly translate «Prefer encryption», since its litteral and direct translation sounds odd. A complete phrase would be better.
- i translated «initial account» by «first account», since it french «initial account» sounds really weird, especially when combined with autocrypt. «Créer un premier compte autocrypt ?» sounds way better than «Créer un compte autocrypt initial ?»
- i can't translate the sentence «group tagged attachments as 'multipart/multilingual'» whithout a proper part of speech decomposition.
- i didn't translated «news» by «nouvelles», but by the widely used «news».

## Regarding possible improvements

- i didn't touch translations about «notmuch», since i wasn't sure about it. Is that a technology name ?
- currently, mailbox is translated into the litteral and very french «boite aux lettres». However, i know no one using that name for internet mailboxes. The widely used term is «boîte mail».
- sometimes, «BAL» acronym is used in place of «boîte aux lettres». The only use i remember of such acronym is a bachelor course in my student past, almost 10 years ago.
- For reference for these two translations, i called my (french) dad today, and asked him (1) if he could identify the BAL acronym, and (2) how he would translate «mailbox» in french. To (1), he wasn't able to identify the meaning of the acronym, until i helped him with the clue «you could find it in Thunderbird» (he is using mails and Thunderbird daily, at home and in job). To (2), he answered «boîte mail», and wasn't convinced by «boîte aux lettres». After discussion with other members of my family, we agreed that (1) BAL is not widely used («the kind of expression used by old peoples to make them look like they are young»), and (2) in the very (very) rare occasion they have to name the thing in which their mail are, they use «boîte mail» and wouldn't think about «boîte aux lettres», which is reserved for the physical box before the house in which advertisement magazines pile up for weeks before being moved to trash.
- NB: i'm not saying my family is a good reference, but they confirmed impressions i got about those translations. The main limit of that search is that they are absolutely not the kind of people using (neo)mutt, thus not representative of french neomutt users. If further research is needed, i may ask in a hundred member online forum.

### DIFF
--- a/po/fr.po
+++ b/po/fr.po
@@ -155,37 +155,37 @@ msgstr ""
 #. q:Exit  c:Create  D:Delete  a:Tgl Active  p:Prf Encr  ?:Help
 #: autocrypt/acct_menu.c:76
 msgid "Prf Encr"
-msgstr ""
+msgstr "Chffr Fav"
 
 #. L10N: Autocrypt Account menu.
 #. flag that an account has prefer-encrypt set
 #: autocrypt/acct_menu.c:117
 msgid "prefer encrypt"
-msgstr ""
+msgstr "préférer le chiffrement"
 
 #. L10N: Autocrypt Account menu.
 #. flag that an account has prefer-encrypt unset;
 #. thus encryption will need to be manually enabled.
 #: autocrypt/acct_menu.c:124
 msgid "manual encrypt"
-msgstr ""
+msgstr "chiffrement manuel"
 
 #. L10N: Autocrypt Account menu.
 #. flag that an account is enabled/active
 #: autocrypt/acct_menu.c:132
 msgid "active"
-msgstr ""
+msgstr "actif"
 
 #. L10N: Autocrypt Account menu.
 #. flag that an account is disabled/inactive
 #: autocrypt/acct_menu.c:138
 msgid "inactive"
-msgstr ""
+msgstr "inactif"
 
 #. L10N: Autocrypt Account Management Menu title
 #: autocrypt/acct_menu.c:178
 msgid "Autocrypt Accounts"
-msgstr ""
+msgstr "Comptes Autocrypt"
 
 #. L10N: This error message is displayed if a database update of an
 #. account record fails for some odd reason.
@@ -222,28 +222,28 @@ msgstr "Impossible de créer %s : %s"
 #. with this message.
 #: autocrypt/autocrypt.c:159
 msgid "Create an initial autocrypt account?"
-msgstr ""
+msgstr "Créer un premier compte autocrypt ?"
 
 #. L10N: Autocrypt is asking for the email address to use for the
 #. autocrypt account.  This will generate a key and add a record
 #. to the database for use in autocrypt operations.
 #: autocrypt/autocrypt.c:181
 msgid "Autocrypt account address: "
-msgstr ""
+msgstr "Adresse du compte Autocrypt : "
 
 #. L10N: Autocrypt prompts for an account email address, and requires
 #. a single address.  This is shown if they entered something invalid,
 #. nothing, or more than one address for some reason.
 #: autocrypt/autocrypt.c:190
 msgid "Please enter a single email address"
-msgstr ""
+msgstr "Veuillez entrer une seule adresse email"
 
 #. L10N: When creating an autocrypt account, this message will be displayed
 #. if there is already an account in the database with the email address
 #. they just entered.
 #: autocrypt/autocrypt.c:205
 msgid "That email address already has an autocrypt account"
-msgstr ""
+msgstr "Cet adresse email possède déjà un compte autocrypt"
 
 #. L10N: Autocrypt has a setting "prefer-encrypt".
 #. When the recommendation algorithm returns "available" and BOTH sender and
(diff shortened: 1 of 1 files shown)


flatcap:
Hi @aluriak, Welcome to NeoMutt!

> Thanks for that project, i will finally switch to CLI thanks to you.

You're welcome.

> First time of my life i'm doing software translation.

Thanks.  It's a good way to get involved.

> Regarding global work and existing data

> Should i add anything to the header comment ?

They're general notes to anyone else who might read the file.
You can add/update/delete as you wish.

At the very least, add a credit line for yourself.

> i currently didn't add any comment.

You're welcome to add comments for any specific translations.

Hopefully, most translations are understandable.
If not, I can add `L10N:` comments to the code to make them clearer.
When I sync the code to the translations the `L10N` comments are copied,
so that all the translators will benefit.

If in doubt, please ask.

> i didn't translated following names: *Newsgroup*.

OK, that's fine.

> *Autocrypt* (a technology name, i guess)

Yes

> i noted that sometimes autocrypt is in lowercase, sometimes it is entitled.
> I reproduced the casing in french.

OK, that's probably a mistake.  I've made a note to investigate.
As a translator you have a bird's eye view of the messages.
If you spot any other inconsistencies, let me know.

> some translations make use of an additionnal `%s`, such as «Subscribe» that is translated to «S'abonner à %s».

OK, that's definitely an error.
Translations that are marked "fuzzy" have been guessed by machine.
The compiler finds similar looking translations in case it saves time.

The translated string **MUST** have the same expandos (%s, etc) as the original.
Note: If you need to reorder them, there's a different syntax to use.
Here's an example from your translation that I've corrected:

```po
msgid "Variable '%s' has an invalid type %d"
msgstr "Le type %2$d de la variable '%1$s' est invalide"
```

- The first expando `%d` has `2$` inserted (its position in the original string)
- The second expando `%s` has `1$` inserted (its position in the original string)

You can check your translations at any time by running:

```sh
msgfmt --statistics -c -o /dev/null po/fr.po
```

> i observed that «peer» is translated «machine distante», i.e. «remote
> machine», whereas the french «pair» would be a litteral translation.

OK.  I'm going to have to leave many questions to your judgement.
I speak a little French, but certainly not enough to help.

Judging by Wikipedia, «pair» is a good choice:
- https://en.wikipedia.org/wiki/Peer-to-peer
- https://fr.wikipedia.org/wiki/Pair_%C3%A0_pair

> the french translation «Le serveur SMTP ne supporte pas l'authentification» was used for many different english sentences

Ah, that might be the automatic 'help', again.

> «tag» was translated by the french «objet»

There's a bit of overlap here.
We try to use `tag` for the act of selecting emails, etc
and `label` for a descriptive marker attached to emails.

> Regarding french language specificities

Lovely :-)

> Regarding difficulties in translations themselves

> i find it quite hard to properly translate «Prefer encryption»
> A complete phrase would be better.

That's a Autocrypt-specific expression.
It's a hint to the recipient of the email that you'd like the reply to be encrypted.

> i can't translate the sentence «group tagged attachments as 'multipart/multilingual'»

Emails can have multiple versions of the text, which look like attachments.
A common example is an HTML part and a text part.

A variant of this is where the separate parts are different languages
(but represent the same text).

The user attaches the separate text files, then tags (selects) them and then uses the `<group-multilingual>` function.

> i didn't translated «news» by «nouvelles», but by the widely used «news».

OK

> Regarding possible improvements

> i didn't touch translations about «notmuch», since i wasn't sure about it.
> Is that a technology name ?

Yes.  Notmuch is an email indexer/searcher.
It can create virtual mailboxes based on search criteria.

> currently, mailbox ... The widely used term is «boîte mail».

OK, «boîte mail» is good.

---

I've moved your updates into the [[translate]](https://github.com/neomutt/neomutt/tree/translate) branch.
Working in a separate branch means it's less work to keep the code and translations in sync.
I merge the translation branch every few weeks (and just before every release).

As a NeoMutt member, you can commit **directly** to the [translate] branch.
After every commit, the [Translation Leaderboard](https://neomutt.org/translate) is _auto-magically_ updated!

French has leapt from 73% (13<sup>th</sup>) to 84% (9<sup>th</sup>).

Thanks again!

flatcap:
As I've moved your changes, I'll close this PR now.
Feel free to use it to ask more questions,
or if you use IRC, I'm in `#neomutt` on `irc.freenode.net`

Aluriak:
Hello flatcap !

Thank you for your answer. I will commit onto `[translate]` in the future. I plan to fix the `label` and `tag` translations to follow your definitions, to fix even more fuzzy translations, and add myself in the leading comment.

> French has leapt from 73% (13th) to 84% (9th).
That… is a very efficient way to motivate translators :p

Have a good day !

Issue #722, opened 2017-08-19
(neo)mutt converts CRLF to LF in attachments with mime text/*


carlytinnerello:
Hi

When mutt saves/pipes an attachment which has MIME type text/* it converts any CRLF into LF (at least on Linux).
This, however, renders some attachments useless.

* It changes the checksum of files which contain a CRLF (example mail attached below) thus disturbing cryptographic verifications (Here I do not mean S/MIME, PGP/MIME).
* It damages files which are (willingly or accidentally) marked as text/*. For example Thunderbird sends (at least under some circumstances) PDF documents as text/pdf. Which then cannot be read by any PDF viewer any more.

In the spirit of the Robustness Principle 'be conservative in what you do, be liberal in what you accept from others', I suggest adding a command <save-raw-entry>/<save-entry-as-binary> (name is up for discussion) which saves the attachment/part of the mail without modifying it, only stripping the transfer encoding.

Looking into the code base, the function which tampers with the line endings is mutt_decode_base64 in handler.c. If the istext parameter is true, then it converts CRLF to LF.
The function itself is called by mutt_decode_attachment which sets istext to true if the attachment has MIME type text/*, message/*, application-gpg/pgp-keys.
The function mutt_decode_attachment is used by mutt_pipe_attachment and mutt_save_attachment in attach.c and by pipe_attachment in recvattach.c (but there might also be others).
Other encodings (e.g. quoted printable) seem not to have this issue.

Temporary workaround for people which have this problem:
a) Save the mail (as a whole) to a temporary mailfolder (maildir or mbox). Then edit the file such that the ContentType of the attachment is application/pdf, application/octet-stream, ...
Now open the mailfolder and save the attachment.
b) Alternatively, save the mail to a file extract the base64 encoded stuff and decode it yourself (e.g. with 'base64 -d').

Best
carly

Here is an example message to reproduce the problem
```
From: test@example.org
To: test@example.com
Subject: Test text attachment
MIME-Version: 1.0
Content-Type: multipart/mixed; boundary="XsQoSWH+UP9D9v3l"
Content-Disposition: inline
X-Mutt-Fcc: =sent
Date: Sat, 19 Aug 2017 12:27:10 +0000


--XsQoSWH+UP9D9v3l
Content-Type: text/plain; charset=us-ascii
Content-Disposition: inline

The MD5 sum of the attachment send (which contained a CRLF) was

    12b81ead7c0b006dea3a7195934fb921

The attachment saved has

[...]

PR #2359, opened 2020-05-25
Restore CRLF->LF conversion in base64 decoding for text/inline parts 

### LINKED ISSUE #2358: neomutt does not show message content with DOS style line-endings in PGP signed message
Due to checking lines only for "buf[0] == '\n'" it does not detects empty lines with "\r\n" line terminators and does not show message content as it doesn't detect the end of the PGP signed marker.
Relevant line: https://github.com/neomutt/neomutt/blob/408c7ec61318a3239c01c468ee36cb8725b1bb9c/ncrypt/crypt_gpgme.c#L2980

Testmessage is here: https://ionic.at/share/testmsg.txt
For the issue to manifest itself it is necessary to set pgp_auto_decode=yes

gahr:
Fixes #2358

### DIFF
--- a/handler.c
+++ b/handler.c
@@ -1487,6 +1487,7 @@ void mutt_decode_base64(struct State *s, size_t len, bool istext, iconv_t cd)
 {
   char buf[5];
   int ch, i;
+  bool cr = false;
   char bufi[BUFI_SIZE];
   size_t l = 0;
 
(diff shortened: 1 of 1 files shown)


flatcap:
The issue dates back years, to: a2b19cb867ca2dcbf61fa82b76883561ad45009e
relating to #722 (neo)mutt converts CRLF to LF in attachments with mime text/*

PR #2361, opened 2020-05-28
test: add compression tests

flatcap:
Add tests for lz4, zlib and zstd compression.

### DIFF
--- a/Makefile.autosetup
+++ b/Makefile.autosetup
@@ -605,6 +605,7 @@ coverage: all test
 	$(RM) mutt/exit.gc??
 	lcov --test-name "test" --output-file lcov.info --capture \
 	  --directory address \
+	  --directory compress \
 	  --directory config \
 	  --directory core \
 	  --directory email \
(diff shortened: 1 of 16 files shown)


gahr:
Can't the `test_compress_*` functions be refactored into a common algorithm to test them all?

flatcap:
> Can't the `test_compress_*` functions be refactored into a common algorithm to test them all?

One complication is that they're individually build-conditional.

They _could_ be merged.
But, excluding the compression level limits, the tests aren't _quite_ the same.
They don't _quite_ behave the same way on errors (possibly something that should be fixed).

gahr:
> But, excluding the compression level limits, the tests aren't quite the same.
> They don't quite behave the same way on errors (possibly something that should be fixed).

I think compression levels should be exposed by `ComprOps`, and given `
const struct ComprOps *cops = compress_get_ops("foo");`, you should be able to code the test in exactly the same way no matter the actual value of "foo".

flatcap:
Added:
- min,max compression levels
- validator that uses them

PR #2363, opened 2020-05-31
Refactor unsafe TAILQ returns

flatcap:
These commits don't change the behaviour of the functions, just the manner in which they return a list.

Refactor: `neomutt_mailboxlist_get_all()`, `rfc2231_encode_string()` and `mutt_list_str_split()`.

They used to allocate a TAILQ on the stack, then return the object.
This leads to a bug when the List is returned empty.

The functions now add their results to a List passed in as a parameter.

---

When the STAILQ is initialised, `stqh_last` is set to the address of    
`stqh_first` which equates to the address of the STAILQ itself.    
The STAILQ is on the stack.    
    
If the function returned without adding anything to the STAILQ, the    
`stqh_last` pointer would now be invalid, pointing to the now-dead stack.

### DIFF
--- a/browser.c
+++ b/browser.c
@@ -775,7 +775,8 @@ static int examine_directory(struct Menu *menu, struct BrowserState *state,
 
     init_state(state, menu);
 
-    struct MailboxList ml = neomutt_mailboxlist_get_all(NeoMutt, MUTT_MAILBOX_ANY);
+    struct MailboxList ml = STAILQ_HEAD_INITIALIZER(ml);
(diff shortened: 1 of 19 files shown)


gahr:
Do we usually put output argument in first or last position?

flatcap:
> Do we usually put output argument in first or last position?

We (read: I) haven't been consitent.
Which means that some day we're heading for a mega-refactor.

flatcap:
OK, refactored to `function(OUT_PARAM, x, y, z)`

PR #2367, opened 2020-06-07
Enable in ./configure to not record cflags

### LINKED ISSUE #2366: Allow to disable writing the Compliation CFLAGS of the version output
Hi and thank you for maintaining this package. This is more of a bug / request, not sure how to classify it.

My distro's package manager (Nix on NixOS) calculates package's dependencies by inspecting all of their files, including binary files. Many of our packages have so called "split outputs" which means it's possible for a package such as Neomutt, to not reference these packages' headers for example, because the headers are in a different output.

Neomutt references such `dev` outputs that belong to library packages such as libgpgme. This is how it looks on NixOS:

```
$ nix why-depends nixpkgs.neomutt nixpkgs.gpgme.dev
/nix/store/4qgjjyld305jm1xlp0bk9sz6f3knm0yx-neomutt-20200501
╚═══bin/.neomutt-wrapped: …1 -D__EXTENSIONS__ -I/nix/store/apv74b0wjz53x20mq5mmih0m20l7yicw-gpgme-1.13.1-dev/include -I/nix…
    => /nix/store/apv74b0wjz53x20mq5mmih0m20l7yicw-gpgme-1.13.1-dev
```

And it's easy to "see" this reference in:

```
$ neomutt -v | grep Compilation
Compilation CFLAGS: -g -O2 -std=c99 -fno-delete-null-pointer-checks -D_ALL_SOURCE=1 -D_GNU_SOURCE=1 -D__EXTENSIONS__ -I/nix/store/apv74b0wjz53x20mq5mmih0m20l7yicw-gpgme-1.13.1-dev/include -I/nix/store/zwjiz44n07wsp24p2sgn63qpsphmswsg-libassuan-2.5.3-dev/include -I/nix/store/8vjpfjawf0ipbna2snjkaliy4cznqw64-libgpg-error-1.36-dev/include -I/nix/store/4qgjjyld305jm1xlp0bk9sz6f3knm0yx-neomutt-20200501/include -DNCURSES_WIDECHAR -I/nix/store/kqs1qq78z424yl3vs8v7ssf6nkq8v6v1-libkrb5-1.18-dev/include
```

According to `git blame` probably ever since the fork from Mutt.

## NeoMutt Version

```
NeoMutt 20200501

System: Linux 5.6.15 (x86_64)
libidn: 1.35 (compiled with 1.35)
GPGME: 1.13.1-unknown
OpenSSL: OpenSSL 1.1.1g  21 Apr 2020
libnotmuch: 5.2.0
storage: lmdb
compression: zlib

Configure options: --prefix=/nix/store/4qgjjyld305jm1xlp0bk9sz6f3knm0yx-neomutt-20200501 --enable-autocrypt --gpgme --gss --lmdb --notmuch --ssl --sasl --with-homespool=mailbox --with-mailpath= ac_cv_path_SENDMAIL=sendmail --zlib

Compilation CFLAGS: -g -O2 -std=c99 -fno-delete-null-pointer-checks -D_ALL_SOURCE=1 -D_GNU_SOURCE=1 -D__EXTENSIONS__ -I/nix/store/apv74b0wjz53x20mq5mmih0m20l7yicw-gpgme-1.13.1-dev/include -I/nix/store/zwjiz44n07wsp24p2sgn63qpsphmswsg-libassuan-2.5.3-dev/include -I/nix/store/8vjpfjawf0ipbna2snjkaliy4cznqw64-libgpg-error-1.36-dev/include -I/nix/store/4qgjjyld305jm1xlp0bk9sz6f3knm0yx-neomutt-20200501/include -DNCURSES_WIDECHAR -I/nix/store/kqs1qq78z424yl3vs8v7ssf6nkq8v6v1-libkrb5-1.18-dev/include

Default options:

Compile options:
  +getaddrinfo -gnutls +gpgme +gss +hcache +homespool +idn +inotify 
  -locales_hack -lua -lz4 +meta -mixmaster +nls +notmuch +openssl +pgp +regex 
  +sasl +smime +sqlite +start_color +sun_attachment +typeahead +zlib -zstd 
MAILPATH="mailbox"
PKGDATADIR="/nix/store/4qgjjyld305jm1xlp0bk9sz6f3knm0yx-neomutt-20200501/share/neomutt"
SENDMAIL="/usr/sbin/sendmail"
SYSCONFDIR="/nix/store/4qgjjyld305jm1xlp0bk9sz6f3knm0yx-neomutt-20200501/etc"

```

## Extra Info

* Operating System and its version: NixOS

doronbehar:
This flag makes package managers such as Guix and Nix able to make their
neomutt package not reference "split outputs" library packages.

* **What does this PR do?**

Fixes #2366.

### DIFF
--- a/auto.def
+++ b/auto.def
@@ -110,6 +110,8 @@ options {
   with-zstd:path            => "Location of Zstandard"
 # System
   with-sysroot:path         => "Target system root"
+# Misc
+  include-path-in-cflags=1  => "Remove include paths from CFLAGS in the output of neomutt -v"
 # Testing
   testing=0                 => "Enable Unit Testing"
   coverage=0                => "Enable Coverage Testing"
@@ -136,9 +138,9 @@ if {1} {
   foreach opt {
     autocrypt bdb coverage debug-backtrace debug-graphviz debug-notify
     debug-parse-test debug-window doc everything fmemopen full-doc gdbm gnutls
-    gpgme gss homespool idn idn2 inotify kyotocabinet lmdb locales-fix lua lz4
-    mixmaster nls notmuch pcre2 pgp pkgconf qdbm rocksdb sasl smime sqlite ssl
-    testing tdb tokyocabinet zlib zstd
+    gpgme gss homespool idn idn2 include-path-in-cflags inotify kyotocabinet
+    lmdb locales-fix lua lz4 mixmaster nls notmuch pcre2 pgp pkgconf qdbm
+    rocksdb sasl smime sqlite ssl testing tdb tokyocabinet zlib zstd 
   } {
     define want-$opt [opt-bool $opt]
   }
@@ -1211,7 +1213,13 @@ if {[get-define want-debug-window]} {
 ###############################################################################
 # Generate conststrings.c
 set conststrings "\
-  unsigned char cc_cflags\[\] = {[text2c [get-define CFLAGS]]};\n\
+  unsigned char cc_cflags\[\] = {[text2c [expr {
+    [get-define want-include-path-in-cflags]
+    ? [get-define CFLAGS]
+    : [lmap x [get-define CFLAGS] {
+        expr {[string equal -length 2 $x {-I}] ? [continue] : $x}
+      }]
+  }]]};\n\
   unsigned char configure_options\[\] = {[text2c $conf_options]};\n"
 if {[catch {set fd [open conststrings.c w]
             puts $fd $conststrings


auto.def
```diff
@@ -110,6 +110,8 @@ options {
   with-zstd:path            => "Location of Zstandard"
 # System
   with-sysroot:path         => "Target system root"
+# Misc
+  cflags-record=1           => "Record the cflags used to build neomutt"
```
gahr, line 114: Please use a negative, since the default is true: “Disable recording compiler flags in the output of neomutt -v” or something

doronbehar:
> I don’t know those distros, can you please elaborate a bit on why this is advantageous?

On NixOS and Guix, every package lives in "the store" which is `/nix/store/ ` - each package has it's own directory tree and all it's files are linked to the user / system environment via symlinks, upon request.

So for example:

```
$ tree --filelimit=4 /nix/store/4qgjjyld305jm1xlp0bk9sz6f3knm0yx-neomutt-20200501/
/nix/store/4qgjjyld305jm1xlp0bk9sz6f3knm0yx-neomutt-20200501/
├── bin
│   └── neomutt
├── etc
│   └── neomuttrc
├── libexec
│   └── neomutt
│       ├── pgpewrap
│       └── smime_keys
└── share
    ├── doc
    │   └── neomutt [27 entries exceeds filelimit, not opening dir]
    ├── locale [31 entries exceeds filelimit, not opening dir]
    └── man
        ├── man1
        │   ├── neomutt.1.gz
        │   ├── pgpewrap_neomutt.1.gz
        │   └── smime_keys_neomutt.1.gz
        └── man5
            ├── mbox_neomutt.5.gz
            ├── mmdf_neomutt.5.gz
            └── neomuttrc.5.gz

11 directories, 10 files
```

And I have `neomutt` available in my `$PATH` because all of these files are symlinked to the "global `/bin/`" directory.

Additionally, our build recipes (`PKGBUILD`s in the Arch world for example) are mere instructions that let a package build. But the true runtime dependencies are decided by Nix according to the **contents of the files** in the output tree. These runtime dependencies are called references. 

When a package is big, we often "split" it, and we put headers in e.g:

```
/nix/store/apv74b0wjz53x20mq5mmih0m20l7yicw-gpgme-1.13.1-dev/include/gpgme.h
```

And the library it self in a different directory in the store:

```
/nix/store/g6305yvafn6f6mgy8zjxnz3sb650j3xw-gpgme-1.13.1/lib/libgpgme.so
```

Note how there's no suffix for the directory in which `libgpgme.so` is found.

This allows us to reduce the amount of references a certain package has. For example, although `mutt` was built against `libgpgme` which is located at `/nix/store/g6305yvafn6f6mgy8zjxnz3sb650j3xw-gpgme-1.13.1/lib/libgpgme.so` and it used it's headers at `/nix/store/apv74b0wjz53x20mq5mmih0m20l7yicw-gpgme-1.13.1-dev/include/`, it doesn't _reference_ any of it's header files or directories themselves. This is because in all of `mutt`'s files, found in `/nix/store/@@@-mutt-<version>/` the string `/nix/store/apv74b0wjz53x20mq5mmih0m20l7yicw-gpgme-1.13.1-dev` wasn't found.

We say in this case that `mutt ` doesn't reference `gpgme`'s `dev` output or alternatively, that it doesn't reference `gpgme.dev`. This means that when a user wants to use `mutt`, they will download from our cache a precompiled directory tree of mutt while `gpgme.dev` won't be needed or downloaded. On the other hand, naturally, the library - where `libgpgme.so` is, _will_ be needed.

From some reason, `neomutt`s' compilation seems to be a bit different from mutt's and the value `cc_cflags` gets at `version.c` also includes the `-I/nix/store/@@@-gpgme-1.13.1-dev/include` flags which were added by us in order to make our compiler find gpgme.

With this patch, or alternatively without recording the cc_cflags at all, it will be possible for NixOS to make `neomutt` not reference `gpgme.dev` - this means it'll reduce the disk usage of neomutt users that use NixOS.

auto.def
doronbehar: :+1:.

gahr:
Do you have any alternative way to record a package dependencies other than inspecting the strings in the resulting binary? Having cflags available for later inspection is a useful feature when debugging users issues.

doronbehar:
> Do you have any alternative way to record a package dependencies other than inspecting the strings in the resulting binary?

Not really, this behavior is a core constraint of Nix. One thing we sometimes do, (see [e.g](https://github.com/NixOS/nixpkgs/pull/89324/files#diff-f45629e931a0642c39c211de47030154R32)) is to replace with `sed` the hardcoded strings in the files / binaries, but we do this when it's impossible to figure out why does it happen - https://github.com/NixOS/nixpkgs/issues/88953 .

> Having cflags available for later inspection is a useful feature when debugging users issues.

I agree. As an alternative, I'm thinking perhaps of `-lgpgme` type of flags? I'm not sure if these are called as LDFLAGS?

doronbehar:
Isn't this:

```
Default options:

Compile options:
  +getaddrinfo -gnutls +gpgme +gss +hcache +homespool +idn +inotify
  -locales_hack -lua -lz4 +meta -mixmaster +nls +notmuch +openssl +pgp +regex
  +sasl +smime +sqlite +start_color +sun_attachment +typeahead +zlib -zstd
```

enough?

gahr:
> Not really, this behavior is a core constraint of Nix.

Very interesting. I wonder how you guys handle packages that load dependencies dynamically and anything else more complicated than simply referencing a header file in a header file. But I guess I can go figure :)

How about rescoping the patch to just remove <-I....> strings from CFLAGS? In my system I have
```
-g -O2 -std=c99 -D_ALL_SOURCE=1 -D_GNU_SOURCE=1 -D__EXTENSIONS__ -I/usr/local/include -DNCURSES_WIDECHAR
```

and I guess the only piece that messes with your build system is -I/usr/local/include?

flatcap:
I don't have any objection to the PR.

(**Note**: I know **nothing** about NixOs, but... :-)
I think the NixOS build process is wrong.

Using NeoMutt and GPGME as an example:

```
/nix/store/5B3B-neomutt-20200501/bin/neomutt
/nix/store/E4C1-gpgme-1.12.2/usr/lib/libgpgme.so.11
```

NeoMutt works because it uses the standard library path of /usr/lib.
A new GPGME comes out, but it's symlinked to the same place so NeoMutt doesn't notice.

```
/nix/store/495D-gpgme-1.13.1/usr/lib/libgpgme.so.11
```

(so far, just repeating what you've explained)

Now to the build process.
On a conventional Unix box, NeoMutt looks for header files in:
```
/usr/include
/usr/local/include
```

However, on NixOS, the NeoMutt build has been instructed to look in 
```
-I/nix/store/apv74b0wjz53x20mq5mmih0m20l7yicw-gpgme-1.13.1-dev/include
```

I think that the build process and build packages should work transparently.
I think the header files should be symlinked into `/usr/include`

If I build NeoMutt, then upgrade GPGME and rebuild NeoMutt, then the build options shouldn't change.

```
/nix/store/178A-gpgme-1.12.2-DEVEL/usr/include/gpgme.h
```

Replaced by:

```
/nix/store/C751-gpgme-1.12.2-DEVEL/usr/include/gpgme.h
```

but NeoMutt just sees `/usr/include/gpgme.h`

With this change, NeoMutt's compile flags are drastically simplified
(probably down to just `/usr/include` or `/usr/local/include`)

mschilli87:
@flatcap: Not sure I got your concerns but if I understood correctly, what you consider *wrong* is considered a feature by some (Guix user here myself): The whole point is to get package management to be functional in the sense of same input, same output without depending on global state. Thus, neomutt with a newer version of gpme (or any other underlying lib, or even a different (version of the same) compiler used at build) is considered a *different* 'version' of neomutt.

flatcap:
@mschilli87 Ah, I see.  Thanks

flatcap:
The build info _might_ be useful to us (though I can't quite remember an occasion when it actually was :-)

If we can't store the info in the _binary_ (or any part of the package), where _can_ we get it?

If a NixOS user reports a problem with `4qgjjyld305jm1xlp0bk9sz6f3knm0yx-neomutt-20200501`, where are the exact build details recorded?

version.c
```diff
@@ -505,6 +505,7 @@ void print_version(FILE *fp)
   rstrip_in_place((char *) configure_options);
   fprintf(fp, "\n\nConfigure options: %s\n", (char *) configure_options);
 
+#ifndef DISABLE_WRITING_CFLAGS
   rstrip_in_place((char *) cc_cflags);
   fprintf(fp, "\nCompilation CFLAGS: %s\n", (char *) cc_cflags);
```
gahr, line 510: I'd terminate the ifndef block here. Do "default options" and "compile options" also cause problems to guix/nix?

doronbehar: You are correct, it's only "Compilation CFLAGS" that matters for us. I've fixed this.

gahr:
I think I'm fine with having this, as long as we keep the default as it is.

doronbehar:
@flatcap 

> If we can't store the info in the binary (or any part of the package), where can we get it?

I guess the parts of CFLAGS that include the `-I` flags are not interesting, but I guess you would be interesting in:

```
Compilation CFLAGS: -g -O2 -std=c99 -fno-delete-null-pointer-checks -D_ALL_SOURCE=1 -D_GNU_SOURCE=1 -D__EXTENSIONS__
```

To answer your question, I guess a NixOS user will have to work a bit harder to get these. Or, we can perhaps modify this:

https://github.com/neomutt/neomutt/blob/ad07b032e7522dcd97d90b91072072b37324e0c6/auto.def#L1219-L1228

So `auto.def` will filter out `-I` flags?

Or perhaps do something here:

https://github.com/neomutt/neomutt/blob/ad07b032e7522dcd97d90b91072072b37324e0c6/auto.def#L221-L233

gahr:
@doronbehar yeah I'll be even more happy with a patch to auto.def to filter out -I<foo> words. Do you need help with that?

doronbehar:
> Do you need help with that?

I guess it'd be safer to do it in:

https://github.com/neomutt/neomutt/blob/ad07b032e7522dcd97d90b91072072b37324e0c6/auto.def#L1219-L1228

But I don't know tcl or whatever that language is in order to do it quickly as you will probably do it.

gahr:
@doronbehar would you be happy with my changes?

doronbehar:
@gahr these changes are absolutely OK :rocket:.

auto.def
```diff
@@ -110,6 +110,8 @@ options {
   with-zstd:path            => "Location of Zstandard"
 # System
   with-sysroot:path         => "Target system root"
+# Misc
+  include-path-in-cflags=1  => "Remove include paths from CFLAGS in the output of neomutt -v"
```
flatcap, line 114:
```suggestion
  include-path-in-cflags=1  => "Show include paths in CFLAGS in the output of neomutt -v"
```

To match the wording of the configure option.

gahr: autoconf will show the contrary of the default in the help screen:

```
./configure --help | grep cflags
  --disable-include-path-in-cflags  Remove include paths from CFLAGS in the output of neomutt -v
```

flatcap: ah!  that's smart

doronbehar:
Thank you!

PR #2375, opened 2020-06-15
implement slist_string_plus_equals() and slist_string_minus_equals()

jindraj:
* **What does this PR do?**
Adds += and -= for DT_SLIST

     for that)

#243 + #2236 (dup)

----
- [ ] documentation
- [ ] tests
    - [ ] add new list value
    - [ ] add value that already exists in list
    - [ ] remove existing value from list
    - [ ] remove value from empty list
    - [ ] add NULL value
    - [ ] remove NULL value


config/slist.c
```diff
@@ -218,6 +218,89 @@ static intptr_t slist_native_get(const struct ConfigSet *cs, void *var,
...
+
+  /* Store empty strings as NULL */
+  if (value && (value[0] == '\0'))
+    return rc |= CSR_SUC_NO_CHANGE; // return no change
+
+  struct Slist *list = *(struct Slist **) var;
+
+  if (slist_is_member(list, value))
+    return rc |= CSR_SUC_NO_CHANGE; // return no change
+  else
```
flatcap, line 241: Don't need the `else` and can unindent the next block

config/slist.c
```diff
@@ -218,6 +218,89 @@ static intptr_t slist_native_get(const struct ConfigSet *cs, void *var,
...
+    else
+      slist_add_string(list, value); // add value to list
+
+  if (cdef->validator)
+  {
+    rc = cdef->validator(cs, cdef, (intptr_t) list, err);
+
+    if (CSR_RESULT(rc) != CSR_SUCCESS)
+    {
+      slist_free(&list);
```
flatcap, line 253: When validation fails, if the list has:
- One item, we should free it
- Two or more, we only want to remove the tail item

config/slist.c
```diff
@@ -218,6 +218,89 @@ static intptr_t slist_native_get(const struct ConfigSet *cs, void *var,
...
+  /* Store empty strings as NULL */
+  if (value && (value[0] == '\0'))
+    return rc |= CSR_SUC_NO_CHANGE; // return no change
+
+
+  struct Slist *list = *(struct Slist **) var;
+
+  if (!list)
+    return rc |= CSR_SUC_NO_CHANGE; // return no change
+  else
```
flatcap, line 284: same as above

config/slist.c
jindraj: slist_dup() at the beginning of the function. Returning modified list if it passes through validator, or the original if it doesn't.

flatcap: yes

config/slist.c
```diff
@@ -218,6 +218,99 @@ static intptr_t slist_native_get(const struct ConfigSet *cs, void *var,
...
+    return CSR_ERR_CODE; /* LCOV_EXCL_LINE */
+
+  int rc = CSR_SUCCESS;
+
+  /* Store empty strings as NULL */
+  if (value && (value[0] == '\0'))
+    return rc |= CSR_SUC_NO_CHANGE; // return no change
+
+  struct Slist *list = *(struct Slist **) var;
+  struct Slist *orig = slist_dup(list);
```
flatcap, line 238: The naming of these two seems a bit odd.
I think I'd go for `orig = *var` and `copy = dup(orig)`

config/slist.c
```diff
@@ -218,6 +218,99 @@ static intptr_t slist_native_get(const struct ConfigSet *cs, void *var,
...
+
+  if (cdef->validator)
+  {
+    rc = cdef->validator(cs, cdef, (intptr_t) list, err);
+
+    if (CSR_RESULT(rc) != CSR_SUCCESS)
+    {
+      slist_free(&list);
+      *(struct Slist **) var = orig;
+      return (rc | CSR_INV_VALIDATOR);
```
flatcap, line 260: With the new naming, you might want to change this failed validator block :-)

config/slist.c
```diff
@@ -218,6 +218,98 @@ static intptr_t slist_native_get(const struct ConfigSet *cs, void *var,
...
+
+  /* Store empty strings as NULL */
+  if (value && (value[0] == '\0'))
+    return rc |= CSR_SUC_NO_CHANGE; // return no change
+
+  struct Slist *orig = *(struct Slist **) var;
+  struct Slist *copy = slist_dup(orig);
+
+  if (slist_is_member(copy, value))
+    return rc |= CSR_SUC_NO_CHANGE; // return no change
```
flatcap, line 241: If we return here, we leak `copy`

config/slist.c
```diff
@@ -201,6 +201,103 @@ static intptr_t slist_native_get(const struct ConfigSet *cs, void *var,
...
+
+  if (slist_is_member(copy, value))
+  {
+    slist_free(&copy);
+    return rc |= CSR_SUC_NO_CHANGE; // return no change
+  }
+
+  if (!copy)
+  {
+    mutt_list_insert_tail(&copy->head, mutt_str_strdup(value));
```
gahr, line 231: I'm not sure this is correct

jindraj: Thanks for pointing out. Now I see.

I thought it's necessary when the list is empty to do it other way, because [slist_add_string()](https://neomutt.org/code/mutt_2slist_8c.html#a56a8eca5b8c63f928ae9944661a208bd) has following code at the beginning of the function:
``` 
if (!list)
     return NULL;
```

I tested it wrong, there are two different situations:

1. Make list empty by removing last remaining member:
```
set hidden_tags = test  # let's set it some value
set hidden_tags -= test # remove it, so the list is empty
set hidden_tags += test #set it back
```
2. set list empty at the beginning
```
set hidden_tags = "" # set it empty in the beginning
set hidden_tags += test # segfaults.
```
1. works (hits the else condition branch)
2. segfaults (hits the if condition branch)

slist_add_string() works fine when the slist variable is emptied after removing the last member but not when the variable is set as empty string.

jindraj: this seems to work to me. This way it doesn't take the variables flags from cdef so `set hidden_tags += xyz,abc` inserts one element xyz,abc in both cases 1) and 2)
```
  if (!copy)
    copy = slist_parse(value, 0);
  else
    slist_add_string(copy, value); // add value to list
```

What do you think, @gahr?

flatcap:
Here are all the possible combinations (that I can think of).

## set (=)

| Test  | Start     | `set =`   | Result    |
| :---- | :-------- | :-------- | :-------- |
| s01   | `[]`      | `[]`      | `[]`      |
| s02   | `[]`      | `[A]`     | `[A]`     |
| s03   | `[]`      | `[A,B,C]` | `[A,B,C]` |
|       |           |           |           |
| s04   | `[P]`     | `[]`      | `[]`      |
| s05   | `[P]`     | `[A]`     | `[A]`     |
| s06   | `[P]`     | `[A,B,C]` | `[A,B,C]` |
|       |           |           |           |
| s07   | `[P,Q,R]` | `[]`      | `[]`      |
| s08   | `[P,Q,R]` | `[A]`     | `[A]`     |
| s09   | `[P,Q,R]` | `[A,B,C]` | `[A,B,C]` |

## add (+=)

| Test | Start     | `set +=` | Result      |
| :--- | :-------- | :------- | :---------- |
| a01  | `[]`      | `[]`     | `[]`        |
| a02  | `[]`      | `[A]`    | `[A]`       |
|      |           |          |             |
| a03  | `[A]`     | `[]`     | `[A]`       |
| a04  | `[A]`     | `[A]`    | `[A]`       |
| a05  | `[A]`     | `[B]`    | `[A,B]`     |
| a06  | `[B]`     | `[A]`    | `[B,A]`     |
|      |           |          |             |
| a07  | `[A,B,C]` | `[]`     | `[A,B,C]`   |
| a08  | `[A,B,C]` | `[A]`    | `[A,B,C]`   |
| a09  | `[A,B,C]` | `[B]`    | `[A,B,C]`   |
| a10  | `[A,B,C]` | `[C]`    | `[A,B,C]`   |
| a11  | `[A,B,C]` | `[D]`    | `[A,B,C,D]` |

## remove (-=)

| Test | Start     | `set -=` | Result     |
| :--- | :-------- | :------- | :--------- |
| r01  | `[]`      | `[]`     | `[]`       |
| r02  | `[]`      | `[A]`    | `[]`       |
|      |           |          |            |
| r03  | `[A,B,C]` | `[A]`    | `[B,C]`    |
| r04  | `[A,B,C]` | `[B]`    | `[A,C]`    |
| r05  | `[A,B,C]` | `[C]`    | `[A,B]`    |
| r06  | `[A,B,C]` | `[Z]`    | `[A,B,C]`  |

PR #2389, opened 2020-06-25
Fix sidebar abbreviation / indentation for named-mailboxes

gahr:
* **What does this PR do?**

There are two issues with the current implementation of the sidebar
drawing algorithm in presence of named-mailboxes:

1. The `%B` expando shows an abbreviated mailbox name. This causes
   `named-mailboxes "My Inbox @ server.com" +Inbox` to be abbreviated as
   `com` (possibly indented).

2. The `%D` expando shows an unindented mailbox name. This causes
   `named-mailboxes "My subfolder" +Inbox.Lists.Neomutt-devel` to be
   displayed more to the left wrt its siblings.

This commit modifies the sidebar rendering algorithm as follows.

1. The `%B` expando always shows the mailbox path
2. The `%D` expando shows the mailbox name for named-mailboxes and the
   mailbox path for regular ones.
3. Indentation is always applied.
4. Abbreviation is only ever applied to paths.

### DIFF
--- a/sidebar.c
+++ b/sidebar.c
@@ -72,7 +72,8 @@ static short PreviousSort = SORT_ORDER; /* sidebar_sort_method */
  */
 struct SbEntry
 {
-  char box[256];           ///< Formatted Mailbox name
+  char box[256];           ///< Mailbox path (possibly abbreviated)
+  int depth;               ///< Indentation depth
   struct Mailbox *mailbox; ///< Mailbox this represents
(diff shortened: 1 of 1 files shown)


flatcap:
The 'rename' case doesn't work any more.
```
named-mailboxes "something" +inbox
```
has no effect.

gahr:
Are you using the %D or %B expando?

flatcap:
Sorry, my mistake.
I'd reset the branch, but not rebased it over master (over your rename changes).

PR #2406, opened 2020-06-30
move random functions to libmutt

flatcap:
These four functions have no dependencies and wide usage, so they belong in the library:

- `mutt_rand32()`
- `mutt_rand64()`
- `mutt_rand_base32()`
- `mutt_randbuf()`

Also change the preferred random method from a Linux syscall, to `getrandom(2)` (available on Linux and BSD).

### DIFF
--- a/mutt/random.c
+++ b/mutt/random.c
@@ -1,0 +1,135 @@
+/**
+ * @file
+ * Random number/string functions
+ *
+ * @authors
+ * Copyright (C) 2020 Richard Russon <rich@flatcap.org>
+ *
+ * @copyright
+ * This program is free software: you can redistribute it and/or modify it under
+ * the terms of the GNU General Public License as published by the Free Software
+ * Foundation, either version 2 of the License, or (at your option) any later
+ * version.
+ *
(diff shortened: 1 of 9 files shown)


gahr:
`mutt_rand32` isn't used at all, it seems.

mutt/random.c
```diff
@@ -0,0 +1,130 @@
...
+    mutt_error(_("mutt_randbuf buflen=%zu"), buflen);
+    return -1;
+  }
+
+#ifdef HAVE_GETRANDOM
+  long ret;
+  do
+  {
+    ret = getrandom(buf, buflen, 0);
+  } while ((ret == -1) && (errno == EINTR));
```
gahr, line 66: I think this should loop until `buflen` bytes have been read: `getrandom` might return a short count.

gahr:
Can the call to `mutt_randbuf` in main.c be removed, too?

mutt/random.c
flatcap: done

flatcap:
> Can the call to `mutt_randbuf` in main.c be removed, too?

That's effectively a check that we've got a working source of random.

PR #2408, opened 2020-07-02
fix(sidebar): use %D in default format

Austin-Ray:
The current default `sidebar_format` uses `%B`, which displays the
mailbox path. Replace `%B` with `%D` to display mailbox names if user
uses named-mailboxes or virtual-mailboxes declarations.

With changes in #2389 stopping the sidebar from always displaying names,
despite `%B`, for named and virtual mailboxes, users will no
longer see the names. The current default forces named and virtual
mailbox users to update the default to see the names. This is
undesirable as those feature appears broken with default settings.

---

With minimal config:

```
set folder="~/mail"
set sidebar_visible=yes
named-mailboxes "Mailbox" "notmuch://?query=tag:inbox"
```

Before:

image

After:
image

Additionally, we should aware of #2389 will require all existing notmuch users with custom sidebar formats to update as they probably don't use `%D` in it. ca381a9b38f29332244ba3356e0b30009f785b3b  added `%D` so sidebar had the `%B` exception for notmuch. Let's make sure to add that notice in the release notes.

### DIFF
--- a/docs/manual.xml.head
+++ b/docs/manual.xml.head
@@ -693,7 +693,7 @@ set sidebar_divider_char = '│'          <emphasis role="comment"># Pretty line
               <link linkend="formatstrings-conditionals">conditionals</link>.
             </para>
             <para>
-              The default value is: <literal>%B%* %n</literal>
+              The default value is: <literal>%D%* %n</literal>
             </para>
             <para>
               A more detailed value is:
@@ -16375,7 +16375,7 @@ set sort_browser="reverse-size"
               <row>
                 <entry><literal>sidebar_format</literal></entry>
                 <entry>string</entry>
-                <entry><literal>%B%* %n</literal></entry>
+                <entry><literal>%D%* %n</literal></entry>
               </row>
               <row>
                 <entry><literal>sidebar_indent_string</literal></entry>
--- a/mutt_config.c
+++ b/mutt_config.c
@@ -3735,7 +3735,7 @@ struct ConfigDef MuttVars[] = {
   ** .pp
   ** \fBSee also:\fP $$sidebar_short_path, $$sidebar_indent_string, $$sidebar_delim_chars.
   */
-  { "sidebar_format", DT_STRING|DT_NOT_EMPTY|R_SIDEBAR, &C_SidebarFormat, IP "%B%*  %n" },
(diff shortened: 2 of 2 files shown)


gahr:
Just thinking.. do we still have a use case for *both* expandos, other than backwards compatibility? I.e., why would one give a mailbox a name and not display it?

Austin-Ray:
Accidentally said `%B` instead of `%D` in the git subject line.

jindraj:
This have broken named-mailboxes with notmuch. Maybe I'm doing something wrong? But definitely worked well before with 20200501, since then I used one devel branch which was based on 20200501.

- related config
[mailboxes config](https://github.com/jindraj/dotfiles/blob/master/.neomutt/neomuttrc.mailboxes#L14)
- sidebar mailboxes names shortening and indentation config:
```
set sidebar_component_depth = 0
set sidebar_delim_chars = "/"
set sidebar_folder_indent = yes
set sidebar_format = "%D%?F? [%F]?%* %?N?%N/?%S"
set sidebar_indent_string = "  "
set sidebar_short_path = yes
```

result:
```
              sample                           20
    GMail                                       0
GMail/INBOX                                     7
GMail/INBOX/work                                4
GMail/INBOX/personal                            6
GMail/unread                                  1/1
GMail/flagged [8]                               8
GMail/sent                                   9639
GMail/spam                                     30
GMail/trash                                   692
GMail/work [1]                              36952
GMail/work/ticketing                         2257
GMail/personal [4]                           4475
GMail/newsletters [1]                        4195
GMail/receipts                                 16
GMail/linkedin                                555
lists [2]                                  1/9398
lists/nslcd                                   133
lists/neomutt                                 305
lists/neomutt/users                           265
lists/neomutt/developers                       40
lists/github.com [2]                       1/8789
lists/github.com/neomutt [1]               1/8664
    lists/github.com/neomutt/neomutt [1]     8543
  lists/github.com/neomutt/management           5
  lists/github.com/neomutt/homebrew            56
  lists/github.com/neomutt/translators         51
lists/github.com/sensu-plugins [1]             80
lists/github.com/other                         45
lists/github.com/about-me                   1/344
archives [8]                              1/56451
archives/today                                1/4
archives/yesterday                             16
archives/this week                            1/4
archives/last week                            233
archives/this month                         1/133
archives/last month [1]                       886
archives/this year [3]                     1/5832
archives/last year [3]                      10975
```

The most weird thing is the indantion. **sample**, **Gmail** shouldn't be indented at all. and the **lists/github.com/neomutt/\*** are indented weird way too.

PR #2410, opened 2020-07-03
Fix and improve `change-folder` behavior

### LINKED ISSUE #1982: change-vfolder does not suggest an unread mailbox
## Expected Behaviour

`change-vfolder` should suggest a mailbox with unread messages like this:

 `Open mailbox ('?' for list):  Dummy`

## Actual Behaviour

It does not auto-complete any mailbox:

`Open mailbox ('?' for list): `

- Always

I bisected it and the regression was introduced in commit 4e5d94bc.

## NeoMutt Version

```
NeoMutt 20191111
```

Austin-Ray:
This pull request includes four distinct changes (1 bug fix, 3 user experience improvements):
1. Removes restriction that limits suggestions to non-notmuch mailboxes.
    Fixes #1982
2. Prefers names over paths if available.
3. Removes restriction for suggestions appearing only if current mailbox had new mail.
4. Suggests mailboxes with unread mail instead of new mail. New mails aren't guaranteed to be unread so it'd be weird to suggest a mailbox without unread  messages.

### DIFF
--- a/mutt_mailbox.c
+++ b/mutt_mailbox.c
@@ -333,8 +333,6 @@ struct Mailbox *mutt_mailbox_next(struct Mailbox *m_cur, struct Buffer *s)
       struct MailboxNode *np = NULL;
       STAILQ_FOREACH(np, &ml, entries)
       {
-        if (np->mailbox->type == MUTT_NOTMUCH) /* only match real mailboxes */
-          continue;
         mutt_buffer_expand_path(&np->mailbox->pathbuf);
         if ((found || (pass > 0)) && np->mailbox->has_new)
         {
--- a/mutt_mailbox.c
+++ b/mutt_mailbox.c
@@ -323,7 +323,7 @@ struct Mailbox *mutt_mailbox_next(struct Mailbox *m_cur, struct Buffer *s)
 {
   mutt_buffer_expand_path(s);
 
-  if (mutt_mailbox_check(m_cur, 0) > 0)
+  if (m_cur)
   {
     bool found = false;
     for (int pass = 0; pass < 2; pass++)
--- a/mutt_mailbox.c
+++ b/mutt_mailbox.c
@@ -334,7 +334,7 @@ struct Mailbox *mutt_mailbox_next(struct Mailbox *m_cur, struct Buffer *s)
       STAILQ_FOREACH(np, &ml, entries)
       {
         mutt_buffer_expand_path(&np->mailbox->pathbuf);
-        if ((found || (pass > 0)) && np->mailbox->has_new)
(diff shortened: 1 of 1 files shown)


gahr:
Nice idea - `:exec <change-folder>` suggests the next mailbox with unread messages by its name. However, if I type in the start of the mailbox name and press tab, it doesn't autocomplete.

Austin-Ray:
> However, if I type in the start of the mailbox name and press tab, it doesn't autocomplete.

Good point, let me see what's involved to improve that functionality.

Austin-Ray:
> However, if I type in the start of the mailbox name and press tab, it doesn't autocomplete.

Did this ever work for non-local/virtual mailboxes? On `neomutt-20180716` I can't get auto-complete for virtual-mailboxes or imaps; however, i get auto-complete for local inboxes if I use something like `=Inbox`

gahr:
> > However, if I type in the start of the mailbox name and press tab, it doesn't autocomplete.
> 
> Did this ever work for non-local/virtual mailboxes? On `neomutt-20180716` I can't get auto-complete for virtual-mailboxes or imaps; however, i get auto-complete for local inboxes if I use something like `=Inbox`

I can get autocompletion for IMAP mailboxes:

`c+F<tab>` --> `+FreeBSD`

Austin-Ray:
I've implemented a rough tab complete function. I'd appreciate some feedback to clean it up since it seems rough in the following ways:

1) Order of precedence for complete (I just threw the `name` check first to time sake)
2) Should the `mx` API contain the completion code for `name`s? 
    I think so since there's no reason for `complete.c` to know about `Account` or `Mailbox`. The `mx` call abstracts away things nicer IMO.
3) Should completion be first match (user's config order matters) or nearest neighbor?

flatcap:
> 1. Order of precedence for complete

To start with, I'd suggest alphabetical order.
We don't need any config and the user will be able to predict the behaviour.

Later, we could follow Vim and add a `$wildmode` config variable.
From Vim's docs:

```
'wildmode'
        ""              Complete only the first match.
        "full"          Complete the next full match.
                        After the last match, the original string is used and then the first match again.
        "longest"       Complete till longest common string.
                        If this doesn't result in a longer string, use the next part.
        When there is only a single match, it is fully completed in all cases.
```

flatcap:
There's lots of autocompletion in NeoMutt already and plenty more places where it would be useful.

> Should the `mx` API contain the completion code for `name`s?

No, I think it should be a separate API.
The backends could implement the CompleteAPI as well as the MxAPI.

> there's no reason for `complete.c` to know about `Account` or `Mailbox`

Absolutely.  Completion should know nothing.

PR #2413, opened 2020-07-03
Optimize checking for the current email in the index

gahr:
This introduces a SelectedEmail struct to keep track of both the receive
time and the message id of the currently selected email. This way, when
the index is resorted, we first check for timestamps and only use string
comparison if the timestamps match. That should make the checking
faster.


gahr:
> Code LGTM. Do you have any metrics to show the improvement in performance?

No, I guess it's barely noticeable, if anything. Which makes me wonder if the added complexity is actually worth it.

PR #2417, opened 2020-07-04
Convert libsend to use Config functions

flatcap:
Use the Config functions to get values from a ConfigSubset, rather than using the global variables.
This conversion is a step towards supporting Account- and Mailbox-specific config.

- 0626b7e1bd smtp: pass SmtpAccountData
- 08bc84627a smtp: move mutt_fqdn() to mutt_smtp_send()

  First, fix a couple of problems in the SMTP code.

  When NeoMutt creates an SMTP connection, it shares some data with other functions.
  Pass that data in `struct SmtpAccountData`, rather than using global variables.

  To send email, NeoMutt needs to use the hostname.
  Calculating the hostname requires some business logic (a fallback to the `Hostname` global variable).
  Move that logic to the top-level function `mutt_smtp_send()`.

- 4226a4e88a send: add ConfigSubset param to functions

  This large commit simply adds a `struct ConfigSet *sub` parameter to a LOT of functions.
  The actual value for this parameter is defaulted to `Neomutt->sub` outside of libsend.

- 2bf70925e6 config: add Config helpers

  Create functions to assist in getting Config values from a ConfigSubset.

- fd3e1c538b send: get Config from a ConfigSubset

  Eliminate all Config variables from libsend, using the Config helper functions.

## Code Conventions

Each conversion follows the same pattern:
Global Config variable `C_SendCharset` becomes local `c_send_charset`.

```diff
- function(C_SendCharset);
+ const char *c_send_charset = cs_subset_string(sub, "send_charset");
+ function(c_send_charset);
```

The local variables are declared const as they should not be altered.

### Globals

The global variables still have a few footholds in the send code.

### Safety

The helper functions use the Config variable's name to get the value.
As long as this name is correct and isn't disabled, then NeoMutt should work.

The helper functions perform a lot of checking, for now, to make sure they return the right value.

**Note**: There's no error checking on the caller side.
  If there's an error, the helper will return a "sensible" default value.

### Setting values

In >5000 lines of send code, there's only one place where a config value gets set: `invoke_mta()`.
That case is just down to poor design: set value; do something; restore value.

### Testing

Using a simple test program, it's now possible to send an email using smtp.

### DIFF
--- a/Makefile.autosetup
+++ b/Makefile.autosetup
@@ -188,7 +188,7 @@ $(PWD)/compress:
 ###############################################################################
 # libconfig
 LIBCONFIG=	libconfig.a
-LIBCONFIGOBJS=	config/address.o config/bool.o config/dump.o config/enum.o \
+LIBCONFIGOBJS=	config/address.o config/bool.o config/dump.o config/enum.o config/helpers.o \
 		config/long.o config/mbtable.o config/number.o config/path.o config/quad.o \
 		config/regex.o config/set.o config/slist.o config/sort.o \
 		config/string.o config/subset.o
--- a/alias/alias.c
+++ b/alias/alias.c
@@ -168,7 +168,7 @@ static void expand_aliases_r(struct AddressList *al, struct ListHead *expn)
   }
 
   const char *fqdn = NULL;
-  if (C_UseDomain && (fqdn = mutt_fqdn(true)))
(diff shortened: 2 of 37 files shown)


flatcap:
- Rebased
- Add missing DT_PATH helper
- Use `assert()`

PR #2418, opened 2020-07-05
feat: open neomutt with mailbox name

### LINKED ISSUE #2101: Open virtual mailbox on mutt startup with -f "mailbox" does nothing (exit code 0 without any additional logs)
#  Expected Behaviour

I have the following in my .muttrc : 

```
virtual-mailboxes "personal" 'notmuch://?query=path:personal/**'
virtual-mailboxes "ninja-sent" 'notmuch://?query=path:jsninja/** and from:HIDDEN'
```

By default, when I open mutt, the opened virtual mailbox is "personal".

I would like to open the mailbox "ninja-sent" from my terminal, without having to do keystrokes within mutt.

I'm running : 

```
neomutt -f "ninja-sent"
```

I would expect that neomutt opens the mailbox (virtual-mailbox) "ninja-sent"

## Actual Behaviour

However, the command exits with status code 0 without any additional information.

## Steps to Reproduce

See above with the .muttrc

- Always

- I think it always did, I never managed to configure it.

## NeoMutt Version

```
NeoMutt 20191207

System: Linux 4.15.0-74-generic (x86_64)
ncurses: ncurses 6.1.20180127 (compiled with 6.1.20180127)
libidn: 1.33 (compiled with 1.33)
libnotmuch: 5.0.0

Configure options: --enable-notmuch --enable-debug

Compilation CFLAGS: -g -O2 -std=c99 -D_ALL_SOURCE=1 -D_GNU_SOURCE=1 -D__EXTENSIONS__ -I/usr/include -DNCURSES_WIDECHAR

Default options:

Compile options:
  +getaddrinfo -gnutls -gpgme -gss -hcache -homespool +idn +inotify 
  -locales_hack -lua +meta -mixmaster +nls +notmuch -openssl +pgp -sasl +smime 
  -sqlite +start_color +sun_attachment +typeahead 
MAILPATH="/var/mail"
SENDMAIL="/usr/sbin/sendmail"
SYSCONFDIR="/etc"

```

## Extra Info

* Operating System and its version :Ubuntu 18.04.2 LTS
* Were you using 'screen' or 'tmux'? Yes
* Is your email local (maildir) or remote (IMAP)? local with notmuch

Austin-Ray:
Allow opening neomutt with a mailbox name with `neomutt -f <name>` if
name exists in the user's configuration.

To support this behavior, introduce 4 new functions to provide a new
high-level abstraction for finding mailboxes and lower-level
name-specific functionality.

 - mx_resolve() - Find a mailbox by checking name then path.
 - mx_name_resolve() - Resolve a mailbox by name.
 - mx_mbox_find_by_name() - Search all accounts for Mailbox with name.
 - mx_mbox_find_by_name_acc() - Search a specific account for Mailbox.

Since 'mx_resolve()' is a new high-level abstraction for finding a
mailbox, replace call 'mx_path_resolve()' with 'mx_resolve()' in
'main.c'. The initial mailbox open only needs a Mailbox struct.

Closes #2101

### DIFF
--- a/mx.c
+++ b/mx.c
@@ -1686,6 +1686,73 @@ struct Mailbox *mx_path_resolve(const char *path)
 }
 
 /**
+ * mx_mbox_find_by_name - Find a Mailbox with given name under an Account
+ * @param a    Account to search
+ * @param name Name to find
+ * @retval ptr Mailbox
+ */
+static struct Mailbox *mx_mbox_find_by_name_ac(struct Account *a, const char *name)
+{
+  if (!a || !name)
+    return NULL;
+
+  struct MailboxNode *np = NULL;
+
+  STAILQ_FOREACH(np, &a->mailboxes, entries)
+  {
+    if (mutt_str_equal(np->mailbox->name, name))
+      return np->mailbox;
+  }
+
+  return NULL;
+}
+
+/**
+ * mx_mbox_find_by_name - Find a Mailbox with given name
+ * @param name Name to search
+ * @retval ptr Mailbox
+ */
+static struct Mailbox *mx_mbox_find_by_name(const char *name)
+{
+  if (!name)
+    return NULL;
+
+  struct Account *np = NULL;
+  TAILQ_FOREACH(np, &NeoMutt->accounts, entries)
+  {
+    struct Mailbox *m = mx_mbox_find_by_name_ac(np, name);
+    if (m)
+      return m;
+  }
+
+  return NULL;
+}
+
+/**
+ * mx_resolve - Get a Mailbox from either a path or name
+ * @param path_or_name Mailbox path or name
+ * @retval ptr         Mailbox
+ *
+ * Order of resolving:
+ *  1. Name
+ *  2. Path
+ */
+struct Mailbox *mx_resolve(const char *path_or_name)
+{
+  if (!path_or_name)
+    return NULL;
+
+  // Order is name first because you can create a Mailbox from
+  // a path, but can't from a name. So fallback behavior creates
+  // a new Mailbox for us.
+  struct Mailbox *m = mx_mbox_find_by_name(path_or_name);
+  if (!m)
+    m = mx_path_resolve(path_or_name);
+
+  return m;
+}
+
+/**
  * mx_ac_add - Add a Mailbox to an Account - Wrapper for MxOps::ac_add()
  */
 int mx_ac_add(struct Account *a, struct Mailbox *m)
--- a/main.c
+++ b/main.c
@@ -1196,7 +1196,7 @@ int main(int argc, char *argv[], char *envp[])
     notify_send(NeoMutt->notify, NT_GLOBAL, NT_GLOBAL_STARTUP, NULL);
 
     repeat_error = true;
-    struct Mailbox *m = mx_path_resolve(mutt_b2s(&folder));
+    struct Mailbox *m = mx_resolve(mutt_b2s(&folder));
     Context = mx_mbox_open(m, ((flags & MUTT_CLI_RO) || C_ReadOnly) ? MUTT_READONLY : MUTT_OPEN_NO_FLAGS);
     if (!Context)
     {
--- a/mx.h
+++ b/mx.h
@@ -309,6 +309,7 @@ int             mx_path_parent     (char *buf, size_t buflen);
 int             mx_path_pretty     (char *buf, size_t buflen, const char *folder);
 enum MailboxType mx_path_probe     (const char *path);
 struct Mailbox *mx_path_resolve    (const char *path);
+struct Mailbox *mx_resolve         (const char *path_or_name);
 int             mx_tags_commit     (struct Mailbox *m, struct Email *e, char *tags);
 int             mx_tags_edit       (struct Mailbox *m, const char *tags, char *buf, size_t buflen);
 


gahr:
Nice feature. The API is getting a bit clumsy, though (see `resolve` vs. `find` functions in mx.h). I find it pretty counter-intuitive that `mx_path_resolve` is basically `mx_mbox_find2` + creation if it does not exist.

Your changes would introduce a `mx_name_resolve` that does *not* create the mailbox (because - as you say - you can't create a mailbox from a name).

Can we try to improve the API a bit while we introduce this feature? I suggest a clear distinction between functions that search (with scoping the search to a single account) and functions that create a mailbox. Example:

- `struct Mailbox *mx_find_name(const char *name);`
- `struct Mailbox *mx_find_name_ac(const char *name, const struct Account *ac);`
- `struct Mailbox *mx_find_path(const char *path);`
- `struct Mailbox *mx_find_path_ac(const char *path, const struct Account *ac);`
- `struct Mailbox *mx_create_path(const char *path);`

And the existing `mx_path_resolve` could just be

- `struct Mailbox *mx_find_or_create_path(const char *path); // mx_find_path + mx_create`

It's a bit longer, but clearer.

Then you *could* have an API that implementes the logic you want in `main.c`:

```c
struct Mailbox *mx_foobar(const char *name_or_path)
{
  struct Mailbox *m = mx_find_name(name_or_path);
  return m ? m : mx_find_or_create_path(name_or_path);
}
```

but I'm not sure the single-place usage warrants a dedicated function, and more importantly, I'm not sure how I would name it in a consistent and clear way.

Austin-Ray:
> I'm not sure the single-place usage warrants a dedicated function

The thought process behind the new function is eliminating the disparity between `named-mailboxes` and `mailboxes`. The `mx` API lacked parity for operations by `name` and most certainly lacked a function for "I don't know/care if this is a path or name, just give me a Mailbox!".  Since I don't like mixing independent changes, I didn't update the current usages of `mx_path_resolve()` as that wasn't absolutely necessary to add this feature.

> The API is getting a bit clumsy
> Can we try to improve the API a bit while we introduce this feature?

I agree the API needs cleaning, but it needs more deliberate planning and refactoring than suitable for tacking onto a feature pull request. As this PR's functions are implemented with the same naming convention as the existing API, I don't see the refactor as a dependency.

I'll gladly undertake the refactor immediately following this PR since there are rough edges and ambiguity to clarify  (I don't see why `resolve` must create a mailbox; we can debate later :)). In addition, we can add more functionality to cover smooth any rough edges elsewhere in the code.

Austin-Ray:
Amended commit to:
1. add a doc string to `mx_name_resolve`
2. rename `mx_mbox_find_by_name_acc` to `mx_mbox_find_by_name_ac` to match existing names.

gahr:
Fair enough - I’ll just ask you to make `static` all functions you’re not exposing in the API. Thanks!

Austin-Ray:
Made all functions except `mx_resolve()` static as that's the only new function with external usage. During refactor the other functions may change to non-static depending on use-case, but functionality exists.

mx.c
```diff
@@ -1685,6 +1685,92 @@ struct Mailbox *mx_path_resolve(const char *path)
...
+
+  return NULL;
+}
+
+/**
+ * mx_path_resolve - Find a Mailbox with a given name
+ * @param name  Mailbox name
+ * @retval ptr  Mailbox
+ */
+static struct Mailbox *mx_name_resolve(const char *name)
```
gahr, line 1741: well now this is a simple wrapper around `mx_mbox_find_by_name`. Do you still want to keep it?

mx.c
```diff
@@ -1685,6 +1685,92 @@ struct Mailbox *mx_path_resolve(const char *path)
...
+static struct Mailbox *mx_mbox_find_by_name_ac(struct Account *a, const char *name)
+{
+  if (!a || !name)
+    return NULL;
+
+  struct MailboxNode *np = NULL;
+
+  STAILQ_FOREACH(np, &a->mailboxes, entries)
+  {
+    if (np->mailbox->name && mutt_str_equal(np->mailbox->name, name))
```
gahr, line 1703: `mutt_str_equal` handles a NULL just fine

mx.c
```diff
@@ -1685,6 +1685,92 @@ struct Mailbox *mx_path_resolve(const char *path)
...
+      return np->mailbox;
+  }
+
+  if (!np)
+    return NULL;
```
gahr, line 1708: I'm not sure you can have `np != NULL` here

mx.c
Austin-Ray: Good point, thanks

mx.c
Austin-Ray: Good point, thanks

mx.c
Austin-Ray: Good point, thanks

mx.c
```diff
@@ -1685,6 +1685,77 @@ struct Mailbox *mx_path_resolve(const char *path)
...
+ * @param name Name to search
+ * @retval ptr Mailbox
+ */
+static struct Mailbox *mx_mbox_find_by_name(const char *name)
+{
+  if (!name)
+    return NULL;
+
+  char buf[PATH_MAX];
+  mutt_str_copy(buf, name, sizeof(buf));
```
gahr, line 1721: Why the additional buf instead of just passing name to mx_mbox_find_by_name_ac?

Austin-Ray: No reason. That was just an artifact leftover from basing the function on `mx_mbox_find2()`

PR #2419, opened 2020-07-06
Make more APIs Context-free

gahr:


flatcap:
Looks like we can de-Context, these too:
- `ci_next_undeleted()`
- `ci_previous_undeleted()`
- `ci_first_message()`

PR #2420, opened 2020-07-07
fix(sidebar): respect sidebar_folder_indent

Austin-Ray:
This commit changes sidebar indentation to respect user settings. Commit
6f01b47af53a15089ceb05e5651914ecaebb379a changed sidebar to always
indent, which disregarded 'sidebar_folder_indent' value. As such, the
behavior is forced on the user even if a user does not want indentation.
We should respect the user's decision.

### DIFF
--- a/sidebar.c
+++ b/sidebar.c
@@ -161,8 +161,10 @@ static const char *sidebar_format_str(char *buf, size_t buflen, size_t col, int 
     case 'D':
     {
       char indented[256];
-      size_t off = add_indent(indented, sizeof(indented), sbe);
-      snprintf(indented + off, sizeof(indented) - off, "%s",
+
+      size_t offset = add_indent(indented, sizeof(indented), sbe);
+
+      snprintf(indented + offset, sizeof(indented) - offset, "%s",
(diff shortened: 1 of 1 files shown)


gahr:
Thanks! I think a better way to fix this would be to adjust the logic where depth is set, around line 1132.

Austin-Ray:
Good call, done.

Austin-Ray:
I discovered `calc_path_depth()` has a side effect of setting the value of `last_part`. If we skip that operation, `sidebar_short_path()` breaks. Any ideas on producing a cleaner API?

gahr:
What I'd propose as a simple fix is
```diff
@@ -1128,6 +1128,8 @@ static void draw_sidebar(struct MuttWindow *win, int num_rows, int num_cols, int
       if (C_SidebarComponentDepth > 0)
         entry->depth -= C_SidebarComponentDepth;
     }
+    else if (!C_SidebarFolderIndent)
+      entry->depth = 0;
```

Austin-Ray:
Done

Issue #2342, opened 2020-05-13
[Feature Request] Thunderbird-like sorted threads


BlackLotus:
Threaded view with sorting like thunderbird.
The default threading in thunderbird (View->Sort by->Threaded, Date, Ascending) is perfect for me.
Newest messages are at the top of the inbox.
Threads start with the oldest message and the bottom message of the thread is the newest message.
Right now the default neomutt behaviour is that single answers are orderd from newest to oldest.
Here are a few neomutt settings that come close, but are not quite right.
```
set sort=threads
set sort_aux = date
````
## What needs to change in NeoMutt?
Probably just another option for sort_aux.

## Have you tried any alternative methods?
I tested every combinaton of thread and reverse-thread with these options
```
#set sort_aux = reverse-date
#set sort_aux = reverse-last-date
#set sort_aux = reverse-last-date-received
#set sort_aux = reverse-date-received
#set sort_aux = date
#set sort_aux = last-date
#set sort_aux = last-date-received
#set sort_aux = date-received
```
## Additional Info
Here an example how thunderbird sorts a thread
![thunderbird-threads](https://user-images.githubusercontent.com/20885/81870095-7f704d80-9575-11ea-9abb-0aaa078be205.png)

gahr:
Do `sort = reverse_threads` together with `sort_aux = last_date` kind of give you what you're looking for?

Austin-Ray:
@gahr : no, that makes threads appear upside-down (new message at top, thread root at bottom).

From a logical view, sorting seems broken.

With `sort=reverse-threads; sort_aux=date`, I'd expect threads to be newest to oldest and messages _under_ the thread root to be sorted by date; however, the `reverse` option just seems to flip the index entries so the thread root is at the bottom.

BlackLotus:
I found some comments that suggest that it's known to be broken (like the wrongly reversed order of the maildir), but won't be fixed to don't break working setups. That's why I suggest adding a new sorting method that acts "correctly" like thunderbird.

cwfoo:
Similar question from the Mutt mailing list: [sort/sort_aux different for sorting threads and sorting within threads](https://marc.info/?l=mutt-dev&m=115758586608122) (2006-09-06).

ebblake:
(salvaging the most relevant parts of #2970):
## What needs to change in NeoMutt?
[...]

PR #2433, opened 2020-07-17
support sort_th for sub thread

fengidri:
* **What does this PR do?**
When sort = threads, then top threads and sub threads will sort by
sort_aux.

But, i want top threads sort by reverse-date, and the sub threads
under one thread sort by date. So I add sort_th to do this.

     for that)

#2342

### DIFF
--- a/config/sort2.h
+++ b/config/sort2.h
@@ -47,6 +47,7 @@ extern const struct Mapping SortSidebarMethods[];
  */
 enum SortType
 {
+  SORT_DUMMY	=  0, ///< Don't use any sort mechanism
   SORT_DATE     =  1, ///< Sort by the date the email was sent
   SORT_SIZE     =  2, ///< Sort by the size of the email
   SORT_ALPHA    =  3, ///< Required by makedoc.c
--- a/mutt_config.c
+++ b/mutt_config.c
@@ -578,6 +578,9 @@ struct ConfigDef MainVars[] = {
   { "sort_re", DT_BOOL|R_INDEX|R_RESORT|R_RESORT_INIT, &C_SortRe, true, 0, pager_validator,
     "Sort method for the sidebar"
   },
+  { "sort_th", DT_SORT|DT_SORT_AUX|R_INDEX|R_RESORT|R_RESORT_SUB, &C_SortTh, SORT_DATE, 0, NULL,
+    "Sort for subthread"
+  },
   { "spam_separator", DT_STRING, &C_SpamSeparator, IP ",", 0, NULL,
     "Separator for multiple spam headers"
   },
--- a/mutt_config.c
+++ b/mutt_config.c
@@ -578,7 +578,7 @@ struct ConfigDef MainVars[] = {
   { "sort_re", DT_BOOL|R_INDEX|R_RESORT|R_RESORT_INIT, &C_SortRe, true, 0, pager_validator,
     "Sort method for the sidebar"
   },
-  { "sort_th", DT_SORT|DT_SORT_AUX|R_INDEX|R_RESORT|R_RESORT_SUB, &C_SortTh, SORT_DATE, 0, NULL,
+  { "sort_th", DT_SORT|DT_SORT_AUX|R_INDEX|R_RESORT|R_RESORT_SUB, &C_SortTh, SORT_DUMMY, 0, NULL,
     "Sort for subthread"
   },
   { "spam_separator", DT_STRING, &C_SpamSeparator, IP ",", 0, NULL,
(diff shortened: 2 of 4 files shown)


flatcap:
Thanks for the PR, @fengidri :heart: 

Over the years, a lot of ideas have been proposed to "fix the sorting".
One idea that I liked was a list of sort parameters, "thread,date,...".
The trouble is that nobody actually sat down and listed all the possible combinations.
(if you have time, that would be brilliant :-)

Your PR works and gives a little more control to the user. :tada:
 
But... it adds another config item.
Once we release NeoMutt with a new config item, it's hard to change our minds.

Let's see if anyone has thoughts.

Thanks again.

flatcap:
**Note**: I've rebased your branch over the recent changes to master.

We've had a chance to play around, now, and we've found a bug.

When `sort=threads sort_aux=last-date-received`, the default setting for `sort_th` gives a different email order to unpatched-NeoMutt.

It can be clearly seen if you <kbd>Alt-V</kbd> to collapse all threads, then switch between `sort_th=date` and `sort_th=reverse-date`.
The top-level emails move around.

BlackLotus:
Ok don't know if this is the correct way (to help the pull request and to solve the problem), but here I go.

> When sort=threads sort_aux=last-date-received, the default setting for sort_th gives a different email order to unpatched-NeoMutt.

If you want the default neomutt behavior to not to change at all the only way I see is to not use sort_th without being explicitly being set. 

I created a pull request against fengidri's repository with this branch https://github.com/BlackLotus/neomutt/tree/sort_th-fix
What it does is it sets the default value for sort_th to SORT_DUMMY and only executes the new method if it was changed. I probably implemented it wrong and/or misunderstood the actualy problem, but there it is.

flatcap:
> Ok don't know if this is the correct way (to help the pull request and to solve the problem)

No it isn't, but don't worry about it :-)
It's just taken me 10 minutes of fiddling, but the correct solution would have been to open a Pull Request against @fengidri's repo.

Starting in your repo, https://github.com/BlackLotus/neomutt
- Select your branch (sort_th-fix): https://github.com/BlackLotus/neomutt/tree/sort_th-fix
- Hit the "Pull Request" button: https://github.com/neomutt/neomutt/compare/master...BlackLotus:sort_th-fix?expand=1
- You might need to click "compare across forks"
- Select the base repo (fengidri/neomutt) : https://github.com/fengidri/neomutt/compare/master...BlackLotus:sort_th-fix?expand=1
- Then fill in the details and hit "Create Pull Request"

It should look like this:
pr

That would create a Pull-Request stacked on top of @fengidri's, who could choose to merge it into their branch.

I'll have a look at your changes later.  If they're not crazy, I'll merge them into the PR, manually.

BlackLotus:
I already did that (although I somehow selected the wrong repo _twice_ how is this even possible?).
Posted my comment here just to discuss the proposed solution.

fengidri:
Sorry for making you wait so long.

I have an idea, we can make some modifications in the sort configuration, such as:

`set sort=reverse-date-sent.date
`
How do you feel about this method, so you don't need to add a new configuration item in config. @flatcap

flatcap:
Sorry this is moving slowly.

Before I can merge this PR, I really need to understand the whole problem.
I don't want to find out, after a release, that I've broken a common use-case.

I've tried to think through all the possible combinations of `sort`, `sort_aux`,
but I simply don't have the time to give it my full attention.

What I'd like is a diagram of **every** possible sort combination.

This isn't as crazy as it sounds.

We'd need some sample Emails.
There'd be two sets of cases: flat, threaded.

The flat cases, we could have (say) up to three levels of sorting.
If the first level matches, try the second, then the third.
Documenting these should be pretty simple.

For the threaded cases, we'd want the same three levels, plus the keyword 'threads'.
These cases are more complicated because of the `last-` prefix.
The arrival of a new email can alter the parent's sort position.

```
01	Subject
02	├─>
03	│ └─>
04	└─>
05	  └─>
06	    └─>
07	      └─>
```

Starting with a sorted tree of emails, for each level of email,
we need to consider how a reply can change the sort order.

For the purposes of documentation, we don't need to know that _actual_ sort method.
So, sorting by subject, label, etc, can be considered as one sort method.
We just need to know that they're different:

```
set sort=meth1,meth2,meth3
```

If we can document all the possible cases, we can write **tests** for them.
That'd make me a very happy man :-)

This is quite a big job, but it's an important step to making NeoMutt reliable and predictable.

flatcap:
Archived to https://github.com/neomutt/neomutt-old#neomutt-2021-02-05

PR #2434, opened 2020-07-18
Refactor libsend's config

flatcap:
The first two commits are part of a large plan to de-centralise the config system.
This will allow Account- and Mailbox-specific config.

The third commit adds a one-line description to each of libsend's config.
This isn't used, but in future would allow context-sensitive help.
Also, it could be added as a comment when dumping config (e.g. `neomutt -D -H`).

@neomutt/translators
The final commit adds gettext (translation) support to the docs.
Translating the online help would be great, but it puts a large burden on the translators.
(There are currently 454 config variables)

---

- f250f4a26 config: no variable
  Allow config items to have **NO** global variable backing them

- 4afc525a3 send: no variable
  Convert libsend to have no global config variables

- 43d75041c config: add docs
  Add one-liner description to all of libsend's config items

- b36b3c46d config: translate one-liner docs
  Add gettext translation support for one-liner docs

---

The config definitions now look like:

```c
{ "attach_charset", DT_STRING, NULL, 0, 0, charset_validator,
  N_("When attaching files, use one of these character sets")
},
{ "forward_decrypt", DT_BOOL, NULL, true, 0, NULL,
  N_("Decrypt the message when forwarding it")
},
```

When all the config global variables are eliminated, the `NULL` (at position 3) can be eliminated, too.

### DIFF
--- a/config/set.c
+++ b/config/set.c
@@ -78,7 +78,12 @@ static void destroy(int type, void *obj, intptr_t data)
 
     cst = cs_get_type_def(cs, type);
     if (cst && cst->destroy)
-      cst->destroy(cs, cdef->var, cdef);
+    {
+      if (cdef->type & DT_NO_VARIABLE)
+        cst->destroy(cs, &cdef->var, cdef);
+      else if (cdef->var)
+        cst->destroy(cs, cdef->var, cdef);
+    }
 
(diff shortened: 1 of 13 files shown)


flatcap:
Send code merged; still thinking about the docs.

PR #2538, opened 2020-07-22
Adjust the edata use in Maildir and Notmuch

flatcap:
This PR is a little sleight-of-hand.

It has no visible change to the user (and barely any for the developer).

**Note**: This change alters the header caching.
  `maildir_flags` used to be cached (unnecessarily, as they're rebuilt from the filename)

---

[`struct Email`](https://github.com/neomutt/neomutt/blob/master/email/email.h#L37) has an [`edata`](https://github.com/neomutt/neomutt/blob/master/email/email.h#L107) member where backends can store private Email data.

This `edata` is used by Notmuch which means that Maildir can't use it (as the two backends share Emails).

This PR swaps the priorities around.
Maildir, as the base, gets to use `edata`, whilst Notmuch uses a private member in Email.

---

- 4e1b4a68d notmuch: hide details of edata
  Use a helper, `nm_edata_get()` rather than `edata` directly

- 5e40ea750 notmuch: move edata to Email->nm_edata
  Use a private data member, `Email.nm_edata`, rather than `edata`

- 5139c1f62 maildir: add MaildirEmailData
  New wrapper

- bb2a1a2de maildir: use MaildirEmailData
  Move Maildir private data into wrapper

---

In the near future, Maildir will get refactored.
It doesn't look like the other backends; it doesn't work like the other backends.
This PR will help in that transformation.

In the more distant future, I'd like to refactor Notmuch.
Currently, it's a hybrid.  Some work can be delegated to Maildir, some it does itself.
It has to understand quite a lot about Maildir, to work.
The ideal situation is that Notmuch becomes a layer on top of multiple Maildir mailboxes.
It won't need to know about Maildir config, or the header cache, that will be Maildir's job.
(A plan is starting to form :-)

### DIFF
--- a/debug/graphviz.c
+++ b/debug/graphviz.c
@@ -54,7 +54,7 @@
 // #define GV_HIDE_CONFIG
 // #define GV_HIDE_MDATA
 
-static void dot_email(FILE *fp, const struct Email *e, struct ListHead *links);
+static void dot_email(FILE *fp, struct Email *e, struct ListHead *links);
 static void dot_envelope(FILE *fp, const struct Envelope *env, struct ListHead *links);
 
 const char *get_content_type(enum ContentType type)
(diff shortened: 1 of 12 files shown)


Austin-Ray:
I'll take a look and test this tonight after work

flatcap:
Thanks.

> decouple the two backends

absolutely

> Notmuch that's leaky and causes many of the current issues

hmm... yes

> I'm still grokking the design choices

OK.  Remember that kzak had to maintain notmuch out-of-tree, which limited his design.

I'll try to draw a diagram (of course :-) of how I think things might fit together.

flatcap:
See also: #2541 - Some ideas on how to redesign the Notmuch backend

Issue #2539, opened 2020-07-22
UTF-8 attachment names from Outlook aren't decoded


nabijaczleweli:
## Expected Behaviour

UTF-8-encoded attachments to be decoded to `zęta.png` instead of `=?utf-8?Q?z=C4=99ta.png?=`

## Actual Behaviour

Aren't.

## Steps to Reproduce

Download this mailbox: [zęta.mbox.gz](https://github.com/neomutt/neomutt/files/4960828/zeta.mbox.gz), ungzip it, open in neomutt.

Mail with subject "zęta" was sent with Outlook, the MIME delimiter is
```
------=_NextPart_000_0008_01D6604B.707A8660
Content-Type: image/png;
	name="=?utf-8?Q?z=C4=99ta.png?="
Content-Transfer-Encoding: base64
Content-Disposition: attachment;
	filename="=?utf-8?Q?z=C4=99ta.png?="
```
and neomutt shows it as that string verbatim.

Mail with subject "zęta 2" was sent with NeoMutt, the MIME delimiter is
```
--ojrk22ledcbbpalg
Content-Type: image/png
Content-Disposition: attachment; filename*=utf-8''z%C4%99ta%2Epng
Content-Transfer-Encoding: base64
```
which shows up correctly as "zęta.png"

## How often does this happen?

- Always

## When did it start to happen?

dunno

## NeoMutt Version

```
NeoMutt 20200626
Copyright (C) 1996-2016 Michael R. Elkins and others.
NeoMutt comes with ABSOLUTELY NO WARRANTY; for details type 'neomutt -vv'.
NeoMutt is free software, and you are welcome to redistribute it
under certain conditions; type 'neomutt -vv' for details.

System: Linux 4.19.0-9-amd64 (x86_64)
ncurses: ncurses 6.1.20181013 (compiled with 6.1.20181013)
libidn: 1.33 (compiled with 1.33)
GPGME: 1.12.0
GnuTLS: 3.6.7
libnotmuch: 5.2.0
storage: tokyocabinet
compression: zlib, zstd

[...]

PR #2540, opened 2020-07-23
Add RFC2047 =?utf-8 example and the fact that Outlook makes it

nabijaczleweli:
* **What does this PR do?**
Update RFC2047-related documentation with the things that I'd looked for but didn't find. May be useful for others communicating with Outlook users in a.d. 2020

#2539

### DIFF
--- a/docs/config.c
+++ b/docs/config.c
@@ -3785,6 +3785,7 @@
 ** to save attachments to files named like:
 ** .ts
 ** =?iso-8859-1?Q?file=5F=E4=5F991116=2Ezip?=
+** =?utf-8?Q?z=C4=99ta.png?=
 ** .te
 ** .pp
 ** When this variable is \fIset\fP interactively, the change won't be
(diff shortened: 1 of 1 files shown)


docs/config.c
```diff
@@ -3785,14 +3785,15 @@
...
+** =?utf-8?Q?z=C4=99ta.png?=
 ** .te
 ** .pp
 ** When this variable is \fIset\fP interactively, the change won't be
 ** active until you change folders.
 ** .pp
 ** Note that this use of RFC2047's encoding is explicitly
 ** prohibited by the standard, but nevertheless encountered in the
-** wild.
+** wild and produced by, i.a., Outlook.
```
flatcap, line 3796:
```suggestion
** wild and produced by, e.g., Outlook.
```

"i.a." inter alia isn't common in English, but "e.g." exempli gratia is.

nabijaczleweli: Fair enough, updated.

PR #2542, opened 2020-07-26
Refactor: Help Bar generation

flatcap:
This PR changes how the Help Bar is generated (one-liner at the top of the screen).

- 29ffcd151 window: define a Window focus
- 3c6c7936c move help data to window
- 61a2eec6d create libhelpbar
- 16da23a1d helpbar: delegate window creation
- 698211edb helpbar: manage own config
- 164da02cd helpbar: add window data
- b600f7275 helpbar: recalc/repaint
- d9e1b07e6 helpbar: add observers
- 5bf245dc4 editor: add help

--------------------------------------------------------------------------------

## Changes

Visually, there are a few small changes:
- Display is tidier when a function isn't bound to a key
- Help Bar works in the Command Line Editor
- Changing a key binding will update the Help Bar

Internally, the code has been restructured to be a passive observer of events.

Before, many parts of the code would be responsible for checking if `$help`
were enabled, generating the help string, padding it and writing it to the
global Window, `HelpBarWindow`.

Now, that is managed by the passive Help Bar Window.
A Window simply adds the help strings to the Window and the Help Bar will do the rest.

In order to know **which** Window is active, we introduce the idea of _focus_.
From the `RootWindow` down, each Window can define one of its children as focussed.

--------------------------------------------------------------------------------

## Anatomy of a Passive Window

The Help Bar Window, only exposes two functions:
- `config_init_helpbar()` - Define the config variable
- `helpbar_create()` - Create the Window

It also defines two Window functions:
- `helpbar_recalc()` - Recalculate the display string
- `helpbar_repaint()` - Display the string

Everything else is controlled by receiving Events.
It observes:

- `NT_BINDING`
  - `NT_BINDING_NEW`
  - `NT_BINDING_DELETED`
  - `NT_BINDING_DELETE_ALL`

Changes to the key bindings cause the Help Bar to set `WA_RECALC`.

- `NT_COLOR`
  - `MT_COLOR_STATUS`

Change to `color status` causes the Help Bar to set `WA_REPAINT`.

- `NT_CONFIG`
  - `$help`

Change to `$help` causes the Help Bar to set `WA_REFLOW` on its parent.

- `NT_WINDOW`
  - `NT_WINDOW_FOCUS`
  - `NT_WINDOW_DELETE`

A change of focus causes the Help Bar to set `WA_RECALC`.
When the Window is deleted, the Help Bar unregisters its observer and cleans up.
(This works as a destuctor)

At the beginning of the next display loop, `window_redraw()` is called.
As necessary, it will:
- Reflow Windows
- Call `recalc()` on Windows
- Call `repaint()` on Windows

### DIFF
--- a/alias/dlgalias.c
+++ b/alias/dlgalias.c
@@ -44,7 +44,6 @@
 #include "alias.h"
 #include "format_flags.h"
 #include "gui.h"
-#include "helpbar.h"
 #include "keymap.h"
 #include "mutt_globals.h"
 #include "mutt_menu.h"
@@ -191,15 +190,15 @@ static void alias_menu(char *buf, size_t buflen, struct AliasMenuData *mdata)
 
   int t = -1;
   bool done = false;
-  char helpstr[1024];
 
   struct Menu *menu = mutt_menu_new(MENU_ALIAS);
   struct MuttWindow *dlg = dialog_create_simple_index(menu, WT_DLG_ALIAS);
+  dlg->help_data = AliasHelp;
+  dlg->help_menu = MENU_ALIAS;
 
   menu->make_entry = alias_make_entry;
   menu->tag = alias_tag;
   menu->title = _("Aliases");
-  menu->help = mutt_compile_help(helpstr, sizeof(helpstr), MENU_ALIAS, AliasHelp);
   menu->max = mdata->num_views;
   menu->mdata = mdata;
 
--- a/helpbar/helpbar.c
+++ b/helpbar/helpbar.c
@@ -28,6 +28,7 @@
 
 #include "config.h"
 #include <stddef.h>
+#include <stdbool.h>
 #include <stdio.h>
 #include "private.h"
 #include "mutt/lib.h"
@@ -43,10 +44,11 @@
  * @param txt    Text part, e.g. "delete"
  * @param menu   Current Menu, e.g. #MENU_PAGER
  * @param op     Operation, e.g. OP_DELETE
+ * @retval true If the keybinding exists
  *
  * This will return something like: "d:delete"
  */
-void make_help(char *buf, size_t buflen, const char *txt, enum MenuType menu, int op)
+static bool make_help(char *buf, size_t buflen, const char *txt, enum MenuType menu, int op)
 {
   char tmp[128];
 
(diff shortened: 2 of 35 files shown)


helpbar/private.h
```diff
@@ -0,0 +1,46 @@
...
+ * This is used to cache the data to generate the Help Bar text.
+ */
+struct HelpbarWindowData
+{
+  int                   help_menu; ///< Menu for key bindings, e.g. #MENU_PAGER
+  const struct Mapping *help_data; ///< Data for the Help Bar
+  char *                help_str;  ///< Formatted Help Bar string
+};
+
+struct MuttWindow *helpbar_create (void);
```
gahr, line 40: This is in `helpbar/lib.h` already

alias/dlgalias.c
```diff
@@ -191,15 +190,15 @@ static void alias_menu(char *buf, size_t buflen, struct AliasMenuData *mdata)
...
 
   struct Menu *menu = mutt_menu_new(MENU_ALIAS);
   struct MuttWindow *dlg = dialog_create_simple_index(menu, WT_DLG_ALIAS);
+  dlg->help_data = AliasHelp;
+  dlg->help_menu = MENU_ALIAS;
 
   menu->make_entry = alias_make_entry;
   menu->tag = alias_tag;
   menu->title = _("Aliases");
-  menu->help = mutt_compile_help(helpstr, sizeof(helpstr), MENU_ALIAS, AliasHelp);
```
gahr, line 202: I think you can get rid of the `help` member of `struct Menu` altogether :)

helpbar/helpbar.c
```diff
@@ -0,0 +1,291 @@
...
+static int helpbar_repaint(struct MuttWindow *win)
+{
+  if (!mutt_window_is_visible(win))
+    return 0;
+
+  struct HelpbarWindowData *wdata = helpbar_wdata_get(win);
+  if (!wdata)
+    return 0;
+
+  mutt_debug(LL_DEBUG1, "repaint\n");
```
gahr, line 152: LL_NOTIFY? Or perhaps you'd like to get rid of all LL_DEBUG1 additions?

helpbar/private.h
flatcap: done

alias/dlgalias.c
flatcap: D'oh! How'd I forget that?
done

helpbar/helpbar.c
flatcap: Ah, my debugging.
Demoted to LL_NOTIFY.

flatcap:
Comments addressed and rebased

PR #2544, opened 2020-07-28
add docs to config items

flatcap:
This trivial PR adds one-liner help to all the config items.

The docs can be viewed from the command line:
- `neomutt -D -O` (or `-DO`) - view all config
- `neomutt -O -Q name` - for a single config item

```sh
# Editor to use when '~v' is given in the built-in editor
set visual = "/usr/bin/nvim"

# Prompt to press a key after running external commands
set wait_key = no

# Filter headers when displaying/forwarding/printing/replying
set weed = yes
```

### DIFF
--- a/autocrypt/config.c
+++ b/autocrypt/config.c
@@ -43,15 +43,23 @@ char *AutocryptSignAs;     ///< Autocrypt Key id to sign as
 char *AutocryptDefaultKey; ///< Autocrypt default key id (used for postponing messages)
 // clang-format on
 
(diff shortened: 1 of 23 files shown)


gahr:
I like the feature, but not the fact that it uses a double `-D`. How about `-D -O` for one-liner?

flatcap:
> How about `-D -O` for one-liner?

Done.

At some point (soon), we're going to have to use `getopt_long(3)` and split up the help into themed sections.

PR #2553, opened 2020-07-30
Sort th fix

BlackLotus:
* **What does this PR do?**
Adds a dummy sort variable so that sort_th doesn't do anything by default. 

#2342


BlackLotus:
Did the pull request against the wrong repo

PR #2554, opened 2020-07-30
Sort th fix

BlackLotus:
* **What does this PR do?**
Adds a dummy sort variable so that sort_th doesn't do anything by default. 

#2342

### DIFF
--- a/config/sort2.h
+++ b/config/sort2.h
@@ -47,6 +47,7 @@ extern const struct Mapping SortSidebarMethods[];
  */
 enum SortType
 {
+  SORT_DUMMY	=  0, ///< Don't use any sort mechanism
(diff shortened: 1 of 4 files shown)


BlackLotus:
Holy shit how could I make the same mistake twice?

flatcap:
Hmm... I thought I recognised the name...
NeoMutt already has a contributor called "[Thomas Schneider](https://github.com/qsuscs)"

BlackLotus:
Yeah there are too many of me. At work there is another one on the same floor and a few other that I don't know personally.
Note: I'm not the football (soccer) player (neither https://en.wikipedia.org/wiki/Thomas_Schneider_(footballer) nor https://en.wikipedia.org/wiki/2010_AFL_draft#2011_rookie_draft ;))

PR #2559, opened 2020-08-01
notify: split types into bitfield

flatcap:
Split the notification type enum into a bitfield.
Then, change the observers to register their interest an specific events.

This adds a little complexity to the notification system, but
reduces the number of unwanted notifications received by the observers.

---

The alternative would be to have the Observers register **multiple** callback functions (one for each notification type)
See #2560

### DIFF
--- a/alias/dlgalias.c
+++ b/alias/dlgalias.c
@@ -202,7 +202,7 @@ static void alias_menu(char *buf, size_t buflen, struct AliasMenuData *mdata)
   menu->max = mdata->num_views;
   menu->mdata = mdata;
 
-  notify_observer_add(NeoMutt->notify, alias_data_observer, menu);
+  notify_observer_add(NeoMutt->notify, NT_ALIAS, alias_data_observer, menu);
   mutt_menu_push_current(menu);
 
   if ((C_SortAlias & SORT_MASK) != SORT_ORDER)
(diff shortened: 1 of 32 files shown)


daedric:
While I get the split, what's the reason to use the bitfield instead of the enum ? 
It is to avoid libraries to have to modify the central enum when they add notification type for instance?

flatcap:
> While I get the split, what's the reason to use the bitfield instead of the enum ?

The bitfield means we can `|` (OR) multiple event types together.
e.g.  Helpbar observes `NT_BINDING|NT_COLOR|NT_CONFIG|NT_WINDOW`

While this works, it means the caller needs a `switch` statement to decide what to do.
(So, I actually prefer #2560)

flatcap:
>  It is to avoid libraries to have to modify the central enum when they add notification type for instance?

No, I haven't thought about that problem yet :-)

daedric:
Yeah it is better to subscribe to an actual type :) 
I'm used to using enum for the actual different flag values that is why I was misled :)

flatcap:
> I'm used to using enum for the actual different flag values

Yeah, I think that's considered an abuse of C :-)

flatcap:
Closed in favour of #2560

Issue #1179, opened 2018-04-30
Convert Keymaps to use STAILQ


flatcap:
This is an "easy" task ([find more](https://github.com/neomutt/neomutt/issues?q=label:"good%20first%20issue"+is:open)) and it's open to everyone.  [_read more..._](https://neomutt.org/dev/easy-tasks)

## Description

Many NeoMutt structures implement their own linked-list type.  
The Keymaps are simply stored as a long linked list.

`STAILQ` is a singly-linked list with pointers to both the head and the tail.  
It's defined in [mutt/queue.h](https://github.com/neomutt/neomutt/blob/master/mutt/queue.h),

### Part 2: Change all users to match

There are plenty of examples in NeoMutt's code of how to iterate through a list.  
For example: [tags.c](https://github.com/neomutt/neomutt/blob/master/tags.c), [tags.h](https://github.com/neomutt/neomutt/blob/master/tags.h)

**Note**: When deleting items, use `STAILQ_FOREACH_SAFE()`

z2v:
I'll take the issue

yuzibo:
Hello,The issue is same to my taken issue:
https://github.com/neomutt/neomutt/pull/1187
I am not hope make noisy for you,but i want to talk about STAILQ usage.
In my case, i encounter segfault :(.
```c
static int parse_alternates(struct Buffer *buf, struct Buffer *s,
                            unsigned long data, struct Buffer *err)
{
  struct GroupContextHead gc = STAILQ_HEAD_INITIALIZER(gc);
  alternates_clean();
  do
  {
  ...
    if (parse_group_context(&gc, buf, s, data, err) == -1)
      goto bail;
   ...
    if (mutt_group_context_add_regex(&gc, buf->data, REG_ICASE, err) != 0)
      goto bail;
...
}
```
Is this right?
Anyway, i will close eye to your the issue.
Thank you

z2v:
@yuzibo Have you changed definintions for `parse_group_context` and `mutt_group_context_add_regex`? What's the type of first their argument in your branch?

yuzibo:
This is my detail code:
in group,h:
```c
struct Group
{
  struct Address *as;
  struct RegexList *rs;
  char *name;
};

/**
 * struct GroupContext - A set of Groups
 */
struct GroupContext
{
  struct Group *g;
  STAILQ_ENTRY(GroupContext) entries;
};
STAILQ_HEAD(GroupContextHead, GroupContext);
...
[...]

PR #2563, opened 2020-08-02
Convert Keymaps to use STAILQ

yousefakbar:
* **What does this PR do?**
Converts Keymaps and its functions to use STAILQ structures instead of being a simple array of linked lists.

Issue #1179

### DIFF
--- a/keymap.c
+++ b/keymap.c
@@ -146,7 +146,7 @@ static struct Mapping KeyNames[] = {
 int LastKey;        ///< contains the last key the user pressed
 keycode_t AbortKey; ///< code of key to abort prompts, normally Ctrl-G
 
-struct Keymap *Keymaps[MENU_MAX];
+struct KeymapList Keymaps[MENU_MAX];
 
 #ifdef NCURSES_VERSION
 /**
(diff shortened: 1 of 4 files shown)


keymap.c
```diff
@@ -195,6 +195,38 @@ static const struct Extkey ExtKeys[] = {
...
+{
+  struct Keymap *np = NULL, *tmp = NULL;
+
+  STAILQ_FOREACH_SAFE(np, km_list, entries, tmp)
+  {
+    STAILQ_REMOVE(km_list, np, Keymap, entries);
+    mutt_keymap_free(&np);
+  }
+
+  STAILQ_INIT(km_list);
```
gahr, line 227: So I think you're mixing two approaches here:

1. iterate, remove and free
```c
struct Keymap *np = NULL, *tmp = NULL;
STAILQ_FOREACH_SAFE(np, km_list, entries, tmp)
{
  STAILQ_REMOVE(km_list, np, Keymap, entries)
  mutt_keymap_free(&np);
}
```

and

2. iterate, free, adjust pointers at the end
```c
struct Keymap *np = STAILQ_FIRST(km_list), *next = NULL;
while (np)
{
  next = STAILQ_NEXT(np, entries);
  mutt_keymap_free(&np);
}
STAILQ_INIT(km_list);
```

By looking for `STAILQ_INIT`  in the code, it seems you're not the first one :)

yousefakbar:
Misclicked something in review on my phone. Is there something I need to change in my commits? My view is limited until I'm back on keyboard.

gahr:
I think the code looks ok - please have a look at my comment from yesterday regarding alternative ways to clean up a STAILQ. Please pick one - your code is using a mix of the two.

keymap.c
```diff
@@ -347,82 +401,73 @@ static enum CommandResult km_bind_err(const char *s, enum MenuType menu, int op,
...
+        char old_binding[MAX_SEQ];
+        char new_binding[MAX_SEQ];
+        km_expand_key(old_binding, MAX_SEQ, map);
+        km_expand_key(new_binding, MAX_SEQ, np);
+        if (err)
+        {
+          /* err was passed, put the string there */
+          snprintf(
+              err->data, err->dsize,
+              _("Binding '%s' will alias '%s'  Before, try: 'bind %s %s "
```
gahr, line 437: There are two calls to gettext with identical parameters. Can you please save the result in a variable and use it in both places?

keymap.c
yousefakbar: Fixed. Went with option 1.

keymap.c
yousefakbar: Done. Stored result in a string called 'err_msg'.

flatcap:
I've made a few minor changes:
- rebased
- fixed Doxygen comments
- run clang-format
- tidied commit message

PR #2567, opened 2020-08-05
Devel/imap msn

gahr:
Refactor and encapsulate IMAP's MSN handling


gahr:
Will be part of #2574

PR #2585, opened 2020-08-14
Indent sidebar according to the mailbox name, if any

nabijaczleweli:
Sorry I missed the DEVEL cycle, it's been a busy week.

* **What does this PR do?**
Fixes a regression against 2020-06-26.

The test case is 
```
:named-mailboxes feeds ~/mail/feeds feeds/podcasts ~/mail/feeds/podcasts feeds/webcomics ~/mail/feeds/webcomics
```

2020-08-14:
2020-08-14 screenshot
2020-06-26/with patch:
with patch screenshot

None.


Austin-Ray:
This pull request undoes the fix in #2389 that made `named-mailboxes`/`virtual-mailboxes` respect the `%B` flag. In its current incarnation, I don't think we want to do this.

There's a commit in the `devel/sidebar-operates-on-what-user-sees` branch I had created a while ago but got busy and didn't finish polishing it. Can you see if that would resolve your problem?

nabijaczleweli:
Cherry-picking `devel/sidebar-operates-on-what-user-sees` on top of `master` and switching to %D in `sidebar_format` did fix it, yeah. Updated the PR with it + some touch-ups.

Austin-Ray:
> We really need to sketch out a table with all combinations of mailbox types, relevant config, and expected results. I find it difficult to sort all that out mentally.

@gahr That commit doesn't change the rules of shortening, abbreviation, and indentation, only applies the operations to names in addition to path. Then if a user opts to display the name with `%D`, it's the shortened/abbreviated/indented form. Additionally, it prevents folder path indentation from being applied to displayed names.

gahr:
This breaks things differently and does not fix unnamed mailboxes being indented too much.

`named-mailboxes "Inbox @ example.com" +Inbox` now appears as <kbd>space</kbd>`com`.

If you have both imap and local mailboxes (unnamed) and set folder to the imap inbox, the local mailboxes will still be over-indented. This is a bug we decided we would leave for later when we fixed most of the sidebar display issues.

That's why I am saying we need to properly take into account all use cases, all configurations, and see what we want to display in each case.

gahr:
Also, in addition to breaking existing stuff, I think the approach here is wrong. I can't think of a good reason why a mailbox name would be "shortened" following the rules we shorten URLs.

PR #2603, opened 2020-08-21
Don't indent named mailboxes with %D

nabijaczleweli:
* **What does this PR do?**

Supersede #2585

https://github.com/neomutt/neomutt/pull/2585#issuecomment-675326474

### DIFF
--- a/sidebar/sidebar.c
+++ b/sidebar/sidebar.c
@@ -113,12 +113,16 @@ static const char *sidebar_format_str(char *buf, size_t buflen, size_t col, int 
     case 'B':
     case 'D':
     {
-      char indented[256];
-      size_t offset = add_indent(indented, sizeof(indented), sbe);
-      snprintf(indented + offset, sizeof(indented) - offset, "%s",
-               ((op == 'D') && sbe->mailbox->name) ? sbe->mailbox->name : sbe->box);
+      if((op == 'D') && sbe->mailbox->name)
+        mutt_format_s(buf, buflen, prec, sbe->mailbox->name);
+      else
+      {
+        char indented[256];
+        size_t offset = add_indent(indented, sizeof(indented), sbe);
+        snprintf(indented + offset, sizeof(indented) - offset, "%s", sbe->box);
 
-      mutt_format_s(buf, buflen, prec, indented);
+        mutt_format_s(buf, buflen, prec, indented);
+      }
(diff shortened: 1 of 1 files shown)


gahr:
I think it's wrong to disable indentation for named mailboxes. Here's why:

Using

```
set sidebar_format="%?N?%5N&     ?│%5S│%D%* %?F? !%F ?%?t? *%t?"
```

These mailboxes

```
+FreeBSD
+FreeBSD.@arch
```

are displayed in the sidebar like this

```
     │34858│FreeBSD
  153│  494│··@arch
```

Using

```
named-mailboxes "Arch stuff" +FreeBSD.@arch
```

I get

```
     │34858│FreeBSD
  153│  494│Arch stuff
```

while I would like to get

```
     │34858│FreeBSD
  153│  494│··Arch stuff
```

Can you please explain what is **your** rationale for wanting this change?

nabijaczleweli:
it fixes this:
screenshot

I would've preferred #2585 which'd make this consistent, and you [said](https://github.com/neomutt/neomutt/pull/2585#issuecomment-675326474) you didn't see a reason for the approach that would let you do exactly what you described above; to get my preferred behaviour, consistent with previous versions, I called my mailboxes `"feeds"`, `"  podcasts"`, `"  webcomics"` instead of `"feeds"`, `"feeds/podcasts"`, `"feeds/webcomics"`, which looks right:
screenshot

Even if it makes the window title useless

gahr:
Well yeah the fact that indentation is sometimes wrong (it is when your named mailboxes aren’t under `folder`) doesn’t imply it is always wrong and we should kill it altogether.

Regarding shortening: that is done based either on URL or folder. Shortening any named mailbox based on occurrences of slashes is not what it was designed to do. It would be a nice feature, but it needs to be implemented.

nabijaczleweli:
I mean, it *was* a nice feature until it was removed, but sure.

gahr:
Sorry, that is *not* what the other PR achieved. It might have for your particular use case, but it broke other stuff. I am sorry if you feel I turned down your PR based on mere personal taste. I have not.

nabijaczleweli:
Oh, no, I was merely referring to the fact that this *had* worked until it didn't; calling this a would-be nice feature is hence misleading at best. I'll see about another implementation.

Issue #2589, opened 2020-08-15
Address Book - Live Sorting


flatcap:
The Address Book shows a list of email aliases.
You can open the Address Book by composing an email: <kbd>m</kbd> (`<mail>`) and <kbd>\<Tab\></kbd> (`<complete>`).

<img width="400" src="https://raw.githubusercontent.com/neomutt/gfx/master/screenshots/screenshot/dlg-alias-100.svg">

The sort order is controlled by `$sort_alias`.

Curently, if you change the sort order, whilst in the Address Book, nothing happens.

## Task 1

Add an 'observer' function that listens for any config changes.
We already have one observer, that listens for changes to the alias data:

https://github.com/neomutt/neomutt/blob/5abcbeda3759228887ebf0c83c98333676d197c5/alias/dlgalias.c#L205

Add one that listens for `DT_CONFIG`.
If it receives a notification, check that it's for `$sort_alias` and then sort the data.

## Task 2

Add two functions to make sorting easier.

The Index View (email list) has two functions `<sort-mailbox>`, `<sort-reverse>`
They show a menu: `Rev-Sort (d)ate,(f)rm,(r)ecv,(s)ubj,t(o),(t)hread,(u)nsort,si(z)e,s(c)ore,s(p)am,(l)abel?`

Create two new alias functions: `<sort-alias>` and `<sort-alias-reverse>` that set `$sort_alias`.

RomeuG:
Hello!

About this line in the issue:

`Create two new alias functions: <sort-alias> and <sort-alias-reverse> that set $sort_alias.`

Shouldn't both alias functions be '<sort-alias-address>` and `<sort-alias>` ?

I am saying this because `$sort_alias` has no reverse in the documentation:

```
       sort_alias
              Type: sort order
              Default: alias

              Specifies how the entries in the "alias" menu are sorted.  The following are legal values:
              ‐ address (sort alphabetically by email address)
              ‐ alias (sort alphabetically by alias name)
              ‐ unsorted (leave in order specified in .neomuttrc)
```

Thanks!

flatcap:
Hi @RomeuG,

> Shouldn't both alias functions be `sort-alias-address` and `sort-alias` ?

They _could_ be, but I had something different in mind.
My `<sort-alias>` would bring up a menu to choose from.

> I am saying this because `$sort_alias` has no reverse in the documentation:

Not yet :-)

There are a lot of screens in NeoMutt and they all work _slightly_ differently.
[...]

PR #2607, opened 2020-08-22
Live Sorting in Address Book

RomeuG:
* **What does this PR do?**
This PR adds the capability of the address book screen to update  on `$sort_alias` configuration changes.

     for that)

I don't think it does meet the acceptance criteria. What does need to be documented here?

The relevant issue number is #2589 .

### DIFF
--- a/alias/dlgalias.c
+++ b/alias/dlgalias.c
@@ -246,11 +246,8 @@ static void dlg_select_alias(char *buf, size_t buflen, struct AliasMenuData *mda
 
   mutt_menu_push_current(menu);
 
-  if ((C_SortAlias & SORT_MASK) != SORT_ORDER)
-  {
-    qsort(mdata->av, mdata->num_views, sizeof(struct AliasView *),
-          ((C_SortAlias & SORT_MASK) == SORT_ADDRESS) ? alias_sort_address : alias_sort_name);
-  }
+  qsort(mdata->av, mdata->num_views, sizeof(struct AliasView *),
+        alias_get_sort_function(C_SortAlias));
 
   for (int i = 0; i < menu->max; i++)
     mdata->av[i]->num = i;
--- a/alias/dlgalias.c
+++ b/alias/dlgalias.c
@@ -145,9 +145,9 @@ static int alias_tag(struct Menu *menu, int sel, int act)
 }
 
 /**
- * alias_data_observer - Listen for data changes affecting the Alias menu - Implements ::observer_t
+ * alias_alias_observer - Listen for data changes affecting the Alias menu - Implements ::observer_t
  */
-static int alias_data_observer(struct NotifyCallback *nc)
+static int alias_alias_observer(struct NotifyCallback *nc)
 {
   if (!nc->event_data || !nc->global_data)
     return -1;
@@ -171,6 +171,29 @@ static int alias_data_observer(struct NotifyCallback *nc)
 
   menu->max = mdata->num_views;
   menu->redraw = REDRAW_FULL;
+  return 0;
+}
+
+/**
+ * alias_config_observer - Listen for `sort_alias` configuration changes and reorders alias list - Implements ::observer_t
+ */
+static int alias_config_observer(struct NotifyCallback *nc)
+{
+  if (!nc->event_data)
+    return -1;
+  if (nc->event_type != NT_CONFIG)
+    return 0;
+
+  struct EventConfig *ec = nc->event_data;
+
+  if (!mutt_str_equal(ec->name, "sort_alias"))
+    return 0;
+
+  struct AliasMenuData *mdata = nc->global_data;
+
+  qsort(mdata->av, mdata->num_views, sizeof(struct AliasView *),
+        ((C_SortAlias & SORT_MASK) == SORT_ADDRESS) ? alias_sort_address : alias_sort_name);
+
   return 0;
 }
 
@@ -202,7 +225,9 @@ static void dlg_select_alias(char *buf, size_t buflen, struct AliasMenuData *mda
   menu->max = mdata->num_views;
   menu->mdata = mdata;
 
-  notify_observer_add(NeoMutt->notify, NT_ALIAS, alias_data_observer, menu);
+  notify_observer_add(NeoMutt->notify, NT_ALIAS, alias_alias_observer, menu);
+  notify_observer_add(NeoMutt->notify, NT_CONFIG, alias_config_observer, mdata);
+
   mutt_menu_push_current(menu);
 
   if ((C_SortAlias & SORT_MASK) != SORT_ORDER)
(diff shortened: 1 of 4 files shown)


RomeuG:
Hello!

Well, I think I had success implementing everything you asked for in your latest message.
I took some liberties, I hope you don't mind.

Regardless, if you see anything that should be changed, please say so. :-)

Thanks!

RomeuG:
Ok, fixed merge conflicts, hope the checks pass.

RomeuG:
I understand where the missing return statement warning comes from:

```
int (*alias_get_sort_function())(const void *a, const void *b)
{
  switch ((C_SortAlias & SORT_MASK))
  {
    case SORT_ALIAS:
      return alias_sort_name;
    case SORT_ADDRESS:
      return alias_sort_address;
    case SORT_UNSORT:
      return alias_sort_unsort;
  }
}
```

I created that function to pick the sorting function.
There needs to be a default. Which one could it be?

alias/dlgalias.c
```diff
@@ -174,6 +176,45 @@ static int alias_data_observer(struct NotifyCallback *nc)
...
+/**
+ * alias_color_observer - Listen for color configuration changes and refreshes the menu
+ */
+static int alias_color_observer(struct NotifyCallback *nc)
+{
+  if ((nc->event_type != NT_COLOR) || !nc->event_data || !nc->global_data)
+    return -1;
+
+  if (nc->event_subtype != MT_COLOR_STATUS)
+    return 0;
```
flatcap, line 210: Although the Address Book doesn't have any custom colours, there are three that affect us:
- `MT_COLOR_INDICATOR`
- `MT_COLOR_NORMAL`
- `MT_COLOR_STATUS`

We _could_ check for each of them, but I think we should just drop the `if`.
If there are multiple `NT_COLOR` events, all we're doing is setting a flag.

alias/gui.c
```diff
@@ -85,20 +85,30 @@ int alias_sort_address(const void *a, const void *b)
...
     else if (addr_a && addr_b)
-      r = mutt_istr_cmp(addr_a->mailbox, addr_b->mailbox);
+      r = mutt_str_coll(addr_a->mailbox, addr_b->mailbox);
     else
       r = 0;
   }
   return RSORT(r);
 }
 
+int alias_sort_unsort(const void *a, const void *b)
```
flatcap, line 102: Please can you add a Doxygen comment block for this function.

alias/gui.h
```diff
@@ -27,12 +27,15 @@
 
 struct Alias;
 
+// clang-format off
```
flatcap, line 30: We don't need these in header files.
I routinely format the `.c` files, but the header files are tidied by hand (and vim-fu).

config/sort2.h
```diff
@@ -64,6 +66,7 @@ enum SortType
   SORT_PATH     = 18, ///< Sort by the folder's path
   SORT_LABEL    = 19, ///< Sort by the emails label
   SORT_DESC     = 20, ///< Sort by the folder's description
+  SORT_UNSORT   = 21, ///< Sort by the original configuration order
```
flatcap, line 69: Ah, sorry.  I should have been clearer.
We don't need a `SORT_UNSORT`.
Traditionally `SORT_ORDER` has meant -- 'in config file order'.
I've made a note to review the naming, to try to make it more consistent.

mutt_config.c
```diff
@@ -130,7 +130,8 @@ const struct Mapping SortAliasMethods[] = {
   // clang-format off
   { "address",  SORT_ADDRESS },
   { "alias",    SORT_ALIAS },
-  { "unsorted", SORT_ORDER },
+  { "unsorted", SORT_UNSORT },
+  { "order",    SORT_ORDER },
```
flatcap, line 134: We don't need this change, either.

alias/dlgalias.c
```diff
@@ -239,6 +283,54 @@ static void dlg_select_alias(char *buf, size_t buflen, struct AliasMenuData *mda
...
+        int sort = C_SortAlias;
+        bool resort = true;
+        bool reverse = (op == OP_SORT_REVERSE);
+
+        switch (mutt_multi_choice(
+            reverse ?
+                /* L10N: The highlighted letters must match the "Sort" options */
+                _("Rev-Sort (a)lias, a(d)dress, (u)nsorted or d(o)n't sort?") :
+                /* L10N: The highlighted letters must match the "Rev-Sort" options */
+                _("Sort (a)lias, a(d)dress, (u)nsorted or d(o)n't sort?"),
```
flatcap, line 298: We don't want the ` or d(o)n't sort` any more.
The user can hit <kbd>a</kbd>, <kbd>d</kbd>, <kbd>u</kbd> or <kbd>Ctrl-G</kbd> to cancel.

alias/dlgalias.c
```diff
@@ -239,6 +283,54 @@ static void dlg_select_alias(char *buf, size_t buflen, struct AliasMenuData *mda
...
+            break;
+
+          case 3:
+            sort = SORT_UNSORT;
+            break;
+
+          case 4: /* d(o)n’t sort */
+            sort = SORT_ORDER;
+            resort = false;
+            break;
```
flatcap, line 321: This can go too

alias/gui.c
```diff
@@ -126,6 +136,22 @@ static struct AliasView *alias_view_new(void)
...
+int (*alias_get_sort_function())(const void *a, const void *b)
+{
+  switch ((C_SortAlias & SORT_MASK))
+  {
+    case SORT_ALIAS:
+      return alias_sort_name;
+    case SORT_ADDRESS:
+      return alias_sort_address;
+    case SORT_UNSORT:
+      return alias_sort_unsort;
```
flatcap, line 151: It shouldn't be possible for any other values, but...
The default for `$sort_alias` is "alias", so I'd move `SORT_ALIAS` to the end and add a `default:`

flatcap (changes requested):
A few fixes, from where I lead you astray.  Sorry :-)
Thanks.

alias/dlgalias.c
RomeuG: Fixed!

alias/gui.c
RomeuG: Added.

alias/gui.h
RomeuG: I use Emacs, and I configured it to clang-format on save.
I guess I have to define a workaround :-)

config/sort2.h
RomeuG: It's ok, I could've been more perceptive :-)
Fixed.

mutt_config.c
RomeuG: Fixed with the fix above.

alias/dlgalias.c
RomeuG: Done.

alias/gui.c
RomeuG: I simply added a default, I think it is fine now? Didn't understand what you meant by moving `SORT_ALIAS` to the end.

flatcap: That's fine.
I simply meant moving the `case SORT_ALIAS` to the end of the `switch`, so that it could be combined with the `default`.

flatcap (approved):

flatcap:
I've done a bit of tiding, and I back-ported your fixes to the original commits.
Functionally, there are no changes.
Thanks.

RomeuG:
Sure. thank you very much!

Just one detail: shouldn't `alias_get_sort_function` use the `sort` parameter instead of `C_SortAlias` ?
If we have access to `C_SortAlias`, is there any reason to pass that parameter?

Thanks!

flatcap:
> shouldn't `alias_get_sort_function` use the `sort` parameter instead of `C_SortAlias`

Oh ****!

RomeuG:
:joy:

PR #2615, opened 2020-08-27
implement string_string_plus_equals()

jindraj:
* **What does this PR do?**
implements string_string_plus_equals()

     for that)

#243, #202 ?

### DIFF
--- a/config/string.c
+++ b/config/string.c
@@ -187,6 +187,44 @@ static intptr_t string_native_get(const struct ConfigSet *cs, void *var,
 }
 
 /**
+ * string_string_plus_equals - Concat String to a string - Implements ConfigSetType::string_plus_equals
+ */
+static int string_string_plus_equals(const struct ConfigSet *cs, void *var,
+                                     const struct ConfigDef *cdef,
+                                     const char *value, struct Buffer *err)
+{
+  /* Skip if the value is missing or empty string*/
+  if (!value || (value && (value[0] == '\0')))
+    return CSR_SUCCESS | CSR_SUC_NO_CHANGE;
+
+  int rc = CSR_SUCCESS;
+
+  char *str = NULL;
+  const char **var_str = (const char **) var;
+
+  if (*var_str)
+    mutt_str_asprintf(&str, "%s%s", *var_str, value);
+  else
+    str = mutt_str_dup(value);
+
+  if (cdef->validator)
+  {
+    rc = cdef->validator(cs, cdef, (intptr_t) str, err);
+
+    if (CSR_RESULT(rc) != CSR_SUCCESS)
+    {
+      FREE(&str);
+      return rc | CSR_INV_VALIDATOR;
+    }
+  }
+
+  string_destroy(cs, var, cdef);
+  *var_str = str;
+
+  return rc;
+}
+
+/**
  * string_reset - Reset a String to its initial value - Implements ConfigSetType::reset()
  */
 static int string_reset(const struct ConfigSet *cs, void *var,
@@ -228,7 +266,7 @@ const struct ConfigSetType cst_string = {
   string_string_get,
   string_native_set,
   string_native_get,
-  NULL, // string_plus_equals
+  string_string_plus_equals,
   NULL, // string_minus_equals
   string_reset,
   string_destroy,
--- a/alias/commands.c
+++ b/alias/commands.c
@@ -90,7 +90,6 @@ enum CommandResult parse_alias(struct Buffer *buf, struct Buffer *s,
     goto bail;
   }
 
-
   /* check to see if an alias with this name already exists */
   TAILQ_FOREACH(tmp, &Aliases, entries)
   {
--- a/docs/manual.xml.head
+++ b/docs/manual.xml.head
@@ -6827,13 +6827,14 @@ set spam_separator=", "
           consist of any number of printable characters and must be enclosed in
           quotes if they contain spaces or tabs. You may also use the escape
           sequences <quote>\n</quote> and <quote>\t</quote> for newline and
-          tab, respectively. <emphasis>quadoption</emphasis> variables are
-          used to control whether or not to be prompted for certain actions, or
-          to specify a default action. A value of <emphasis>yes</emphasis>
-          will cause the action to be carried out automatically as if you had
-          answered yes to the question. Similarly, a value of
-          <emphasis>no</emphasis> will cause the action to be carried out as if
-          you had answered <quote>no.</quote> A value of
+          tab, respectively. Content of a <emphasis>string</emphasis> variable
+          can be extended using <emphasis>+=</emphasis>.
+          <emphasis>quadoption</emphasis> variables are used to control whether
+          or not to be prompted for certain actions, or to specify a default action.
+          A value of <emphasis>yes</emphasis> will cause the action to be carried
+          out automatically as if you had answered yes to the question. Similarly,
+          a value of <emphasis>no</emphasis> will cause the action to be carried
+          out as if you had answered <quote>no.</quote> A value of
           <emphasis>ask-yes</emphasis> will cause a prompt with a default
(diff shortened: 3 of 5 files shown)


config/string.c
```diff
@@ -186,6 +186,40 @@ static intptr_t string_native_get(const struct ConfigSet *cs, void *var,
...
+static int string_string_plus_equals(const struct ConfigSet *cs, void *var, const struct ConfigDef *cdef,
+                             const char *value, struct Buffer *err)
+{
+  /* Skip if the value is missing or empty string*/
+  if (!value || (value && (value[0] == '\0')))
+    return CSR_SUCCESS | CSR_SUC_NO_CHANGE;
+
+  int rc = CSR_SUCCESS;
+
+  if (var)
```
flatcap, line 201: We don't need this check as `var` is guaranteed in the contract for `string_plus_equals()`:
https://github.com/neomutt/neomutt/blob/758d453f0326ee1e8f6b69931277820117e62c88/config/set.h#L159-L173

config/string.c
```diff
@@ -186,6 +186,40 @@ static intptr_t string_native_get(const struct ConfigSet *cs, void *var,
...
+  {
+    char *str = mutt_buffer_strdup(var);
+    mutt_str_cat(str, mutt_str_len(str) + 1 + mutt_str_len(value), value);
```
flatcap, line 204: `str` is only `strlen(var)` long, but you're telling `mutt_str_cat()` that it's much longer.

config/string.c
```diff
@@ -186,6 +186,40 @@ static intptr_t string_native_get(const struct ConfigSet *cs, void *var,
...
+
+    if (cdef->validator)
+    {
+      rc = cdef->validator(cs, cdef, (intptr_t) str, err);
+
+      if (CSR_RESULT(rc) != CSR_SUCCESS)
+        return rc | CSR_INV_VALIDATOR;
```
flatcap, line 211: If the validator fails, we leak `str`

flatcap (changes requested):

flatcap:
Suggested tests:

| String Before | String Added | Result        |
| :------------ | :----------- | :------------ |
| [empty]       | [empty]      | [empty]       |
| [empty]       | "apple"      | "apple"       |
| "apple"       | [empty]      | "apple"       |                                       
| "apple"       | "banana"     | "applebanana" |

config/string.c
jindraj: According to the documentation I understood it the same way you're telling me now. 

I tried this
```c
mutt_str_cat(str, mutt_str_len(str), value);
```

But it was shortening the original string `str` by 1 character. After adding the `1`, it stayed untouched + adding `mutt_str_len(value)` resolved it. I believe there's something wrong.

flatcap: `mutt_str_cat()` doesn't realloc the string, it expects a buffer with enough room for both strings.
`strdup(var)` is a block of memory `strlen(var)+1` long, however you're telling `mutt_str_cat()` that it's longer which will lead to a buffer overflow.
You need to measure the two strings then alloc some memory big enough for both, then copy the text.

flatcap: `mutt_str_asprintf()`s what you need (@flatcap finally remembers)
It allocs the space for you.

```c
char *new_value = NULL;
mutt_str_asprintf(&new_value, "%s%s", orig, add);
```

jindraj: this works fine. But when I `unset` the variable or set it to empty string `""`, then the original value is `"(null)"`, I have no idea where this origins.

flatcap: Hmm... I don't see that with the code, here.
Where do you see the `"(null)"` string?

config/string.c
```diff
@@ -198,24 +198,26 @@ static int string_string_plus_equals(const struct ConfigSet *cs, void *var, cons
 
   int rc = CSR_SUCCESS;
 
-  if (var)
+  char *str;
+  mutt_str_asprintf(&str, "%s%s", mutt_buffer_strdup(var), value);
```
flatcap, line 202: Ah, I see what's going on.
`var` is a pointer to the variable.
In this case, the variable is `char *C_Name` and `var` is `char **`
`var` (the pointer) will never be `NULL`, but the variable `*var` _might_ be (variable is empty).

That gives us two cases:
- `*var == NULL` -- just copy the value
- `*var != NULL` -- join the two strings

We've already filtered out the case where `value` is empty.

We don't want `mutt_buffer_strdup()`, just `*var`

flatcap (changes requested):

config/string.c
jindraj: Good, thanks. I think I sorted it out.

jindraj:
> Suggested tests:
> 
> String Before	String Added	Result
> [empty]	[empty]	[empty]
> [empty]	"apple"	"apple"
> "apple"	[empty]	"apple"
> "apple"	"banana"	"applebanana"

From my testing this is exactly how it behaves. Plus I tested validator using `header_cache_backend` config option.

config/string.c
```diff
@@ -186,6 +186,43 @@ static intptr_t string_native_get(const struct ConfigSet *cs, void *var,
...
+{
+  /* Skip if the value is missing or empty string*/
+  if (!value || (value && (value[0] == '\0')))
+    return CSR_SUCCESS | CSR_SUC_NO_CHANGE;
+
+  int rc = CSR_SUCCESS;
+
+  char *str = NULL;
+
+  if (*(char **) var == NULL)
```
flatcap, line 203: We could make this a variable to simplify the later uses:
```c
const char **global = (const char **) var;
```

config/string.c
```diff
@@ -186,6 +186,43 @@ static intptr_t string_native_get(const struct ConfigSet *cs, void *var,
...
+    mutt_str_asprintf(&str, "%s", value);
```
flatcap, line 204: Commonly known as `mutt_str_dup()` :-)

flatcap (changes requested):
Nearly there.
Then, we need some tests adding to `test/config/string.c`
(you can cut'n'paste some code from `test/config/slist.c` - `test_plus_equals()`.

If you have `lcov` installed, you can check your work by adding `./configure --testing --coverage` then `make coverage`

The output (`lcov/index.html`) will look like:
- https://flatcap.org/mutt/lcov/index.html
- https://flatcap.org/mutt/lcov/config/string.c.gcov.frameset.html

config/string.c
```diff
@@ -186,6 +186,44 @@ static intptr_t string_native_get(const struct ConfigSet *cs, void *var,
...
+
+    if (CSR_RESULT(rc) != CSR_SUCCESS)
+    {
+      FREE(&str);
+      return rc | CSR_INV_VALIDATOR;
+    }
+  }
+
+  string_destroy(cs, var, cdef);
+  *(const char **) var = str;
```
flatcap, line 221:
```suggestion
  *var_str = str;
```

config/string.c
```diff
@@ -186,6 +186,44 @@ static intptr_t string_native_get(const struct ConfigSet *cs, void *var,
...
+
+  int rc = CSR_SUCCESS;
+
+  char *str = NULL;
+  const char **var_str = (const char **) var;
+
+  if (mutt_str_equal(*var_str, ""))
+    str = mutt_str_dup(value);
+  else
+    mutt_str_asprintf(&str, "%s%s", *var_str, value);
```
flatcap, line 207:
```suggestion
  if (*var_str)
    mutt_str_asprintf(&str, "%s%s", *var_str, value);
  else
    str = mutt_str_dup(value);
```

There are two changes here:
- Replace the test with `(!*var_str)` -- empty strings are stored as `NULL`
- Swap the `if` around so that the `true` case is first

flatcap (changes requested):
Looks good now.
Just the tests to go.

flatcap (approved):

PR #2617, opened 2020-08-27
Query Live Sorting

RomeuG:
* **What does this PR do?**

Just like the Address Book Live Sorting PR, this one adds Sort and Reverse Sort functions to the Query window, where the user is able to sort the information by alias and address. It is also possible to "unsort".

A lot of the code from the PR mentioned above was reused.

I also took the liberty of putting the `alias_sort_observer` and the `alias_color_observer` functions into `alias/gui.c`, since they are both used by `dlgalias.c` and `dlgquery.c`.

There is no relevant issue number. (Is it needed?)

### DIFF
--- a/alias/dlgalias.c
+++ b/alias/dlgalias.c
@@ -177,42 +177,6 @@ static int alias_alias_observer(struct NotifyCallback *nc)
 }
 
 /**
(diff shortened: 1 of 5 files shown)


gahr (approved):
Works fine!

RomeuG:
Hey, thats awesome! :-)

Issue #2583, opened 2020-08-12
Add validators for {imap,pop,smtp,nntp}_authenticators


flatcap:
The config system can already validate several types of data.
For cases when it can't, we can use a validator function.

For example:

https://github.com/neomutt/neomutt/blob/797b99985644a23045b71bca8aef552a743bdb55/send/config.c#L269-L271

When setting `$wrap_headers`, the config system checks the variables type:
- Number, `DT_NUMBER`
- May not be negative, `DT_NOT_NEGATIVE`

If these tests are passed, the config system then calls the validator function which adds extra constraints
(the number must be between 78 and 998).

https://github.com/neomutt/neomutt/blob/797b99985644a23045b71bca8aef552a743bdb55/send/config.c#L40-L53

---

There are four authenticator config variables that aren't validated in any way.
- [`$imap_authenticators`](https://github.com/neomutt/neomutt/blob/master/imap/config.c#L74-L76), which is limited to [`imap_authenticators[]`](https://github.com/neomutt/neomutt/blob/master/imap/auth.c#L57-L72)
- [`$pop_authenticators`](https://github.com/neomutt/neomutt/blob/master/pop/config.c#L53-L55), which is limited to [`pop_authenticators[]`](https://github.com/neomutt/neomutt/blob/master/pop/auth.c#L380-L386)
- [`$smtp_authenticators`](https://github.com/neomutt/neomutt/blob/master/send/config.c#L241-L243), values are mentioned in [`smtp_auth()`](https://github.com/neomutt/neomutt/blob/master/send/smtp.c#L606)
- [`$nntp_authenticators`](https://github.com/neomutt/neomutt/blob/master/nntp/config.c#L81-L83), values are [mentioned here](https://github.com/neomutt/neomutt/blob/master/nntp/nntp.c#L602)

Challenge:
- Create an array of valid values for smtp and nntp
- Add a validator function for each config variable

Each config variable is a colon-separated list of strings.
There are functions in [`mutt/slist.c`](https://github.com/neomutt/neomutt/blob/master/mutt/slist.c) and [`config/slist.c`](https://github.com/neomutt/neomutt/blob/master/config/slist.c) that will help parse the strings.

yousefakbar:
> Create an array of valid values for smtp and nntp

@flatcap  For nntp, what kind of array is preferred? Below is the array I made for smtp_authenticators:

[...]

PR #2624, opened 2020-08-28
Add validators for {imap,smtp,pop,nntp}_authenticators

yousefakbar:
* **What does this PR do?**

Adds validators to {imap,pop,smtp,nntp}_authenticators config variables. This also includes some refactoring of code in some files.

#2583 .

* **Things to Review**

Obviously the whole commit should be looked at for errors etc. but here are the key areas I would like input/confirmation on:

1. **Placement of function declarations**: `bool {imap,pop,smtp,nntp}_auth_is_valid()` functions are placed in .h files to make them visible to the validator functions. I just want to make sure they're not wandering in the wrong places :).

2. **Changes in `smtp/`**: Namely, I added a SmtpAuth multiplexor (similar to imap and pop's) and slightly modified `smtp_auth()` and `smtp_auth_sasl()`. I tested them and they work fine, but I want to emphasize the change and make sure it gets an okay.

3. **Changes in `nntp/`**: Most notably, I changed the nntp_authenticator config variable's data type to SLIST (was just string prev.) to mirror the other authenticators. This might have effects on the `nntp_auth()` backend (which needs an overhaul, anyway) that I was not able to test.

### DIFF
--- a/conn/sasl.c
+++ b/conn/sasl.c
@@ -96,6 +96,15 @@ struct SaslSockData
   int (*close)(struct Connection *conn);
 };
 
+/**
+ * sasl_authenticators - Authenticaion methods supported by Cyrus SASL
+ */
+static const char *const sasl_authenticators[] = {
+  "ANONYMOUS", "CRAM-MD5",   "DIGEST-MD5",  "EXTERNAL", "G2",
+  "GSSAPI",    "GSS-SPNEGO", "KERBEROS_V4", "LOGIN",    "NTLM",
+  "OTP",       "PASSDSS",    "PLAIN",       "SCRAM",    "SRP"
+};
+
 /* arbitrary. SASL will probably use a smaller buffer anyway. OTOH it's
  * been a while since I've had access to an SASL server which negotiated
  * a protection buffer. */
@@ -106,6 +115,23 @@ struct SaslSockData
 static sasl_callback_t MuttSaslCallbacks[5];
 
 static sasl_secret_t *secret_ptr = NULL;
+
+/**
+ * sasl_auth_validator - Validate an auth method against Cyrus SASL methods
+ * @param authenticator Name of the authenticator to validate
+ * @retval bool True if argument matches an accepted auth method
+ */
+bool sasl_auth_validator(const char *authenticator)
+{
+  for (size_t i = 0; i < mutt_array_size(sasl_authenticators); i++)
+  {
+    const char *auth = sasl_authenticators[i];
+    if (mutt_istr_equal(auth, authenticator))
+      return true;
+  }
+
+  return false;
+}
 
 /**
  * getnameinfo_err - Convert a getaddrinfo() error code into an SASL error code
--- a/conn/sasl.h
+++ b/conn/sasl.h
@@ -26,8 +26,11 @@
 #define MUTT_CONN_SASL_H
 
 #include <sasl/sasl.h>
+#include <stdbool.h>
 
 struct Connection;
+
+bool sasl_auth_validator(const char *authenticator);
 
 int  mutt_sasl_client_new(struct Connection *conn, sasl_conn_t **saslconn);
 void mutt_sasl_done      (void);
--- a/imap/auth.c
+++ b/imap/auth.c
@@ -55,7 +55,9 @@ struct ImapAuth
  * imap_authenticators - Accepted authentication methods
  */
 static const struct ImapAuth imap_authenticators[] = {
-  { imap_auth_oauth, "oauthbearer" }, { imap_auth_plain, "plain" },
+  // clang-format off
+  { imap_auth_oauth, "oauthbearer" },
+  { imap_auth_plain, "plain" },
 #ifdef USE_SASL
   { imap_auth_sasl, NULL },
 #else
@@ -69,7 +71,28 @@ static const struct ImapAuth imap_authenticators[] = {
   { imap_auth_cram_md5, "cram-md5" },
 #endif
   { imap_auth_login, "login" },
+  // clang-format on
 };
+
+/**
+ * imap_auth_is_valid - Check if string is a valid imap authentication method
+ * @param authenticator Authenticator string to check
+ * @retval bool True if argument is a valid auth method
+ *
+ * Validate whether an input string is an accepted imap authentication method as
+ * defined by #imap_authenticators.
+ */
+bool imap_auth_is_valid(const char *authenticator)
+{
+  for (size_t i = 0; i < mutt_array_size(imap_authenticators); i++)
+  {
+    const struct ImapAuth *auth = &imap_authenticators[i];
+    if (auth->method && mutt_istr_equal(auth->method, authenticator))
+      return true;
+  }
+
+  return false;
+}
 
 /**
  * imap_authenticate - Authenticate to an IMAP server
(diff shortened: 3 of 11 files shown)


flatcap:
I've had a look through the code and it looks good -- it may take me a little while to review, though :-)

I noticed one bug, but not in your code, in mine!

If I `:set imap_authenticators="plain:plain"` then NeoMutt doesn't de-dupe the value.
If you have some time, please could you work on a fix for that.
(The `+=` code de-dupes, so you can probably cut'n'paste a little)

Thanks.

yousefakbar:
> it may take me a little while to review, though :-)

No rush at all :D.

> If you have some time, please could you work on a fix for that.

Sounds good! I'll work on that next.

flatcap:
I've merged the slist de-dupe and rebased over master.
Thanks.

flatcap:
I've had a think about SASL.
Please can you create an `sasl_auth_validator()` function in `conn/sasl.c`
It should check for all the possible Cyrus SASL authentication methods:
```
ANONYMOUS
CRAM-MD5
DIGEST-MD5
EXTERNAL
G2
GSSAPI
GSS-SPNEGO
KERBEROS_V4
LOGIN
NTLM
OTP
PASSDSS
PLAIN
SCRAM
SRP
```

We can then call this from the pop,imap,smtp,nntp validator functions.

Thanks.

conn/sasl.c
```diff
@@ -96,6 +96,13 @@ struct SaslSockData
   int (*close)(struct Connection *conn);
 };
 
+static char *sasl_authenticators[] =
+{
+  "ANONYMOUS", "CRAM-MD5", "DIGEST-MD5", "EXTERNAL", "G2", "GSSAPI",
+  "GSS-SPNEGO", "KERBEROS_V4", "LOGIN", "NTLM", "OTP", "PASSDSS", "PLAIN",
+  "SCRAM", "SRP"
+};
```
gahr, line 104: I'd like the ARRAY APIs to be used instead of raw arrays whenever possible. See c24d17a99d16e81872f08f721a1793a57bbdaa92 for an example.

gahr: nevermind, disregard this :)

flatcap:
I've slightly refactored your validator functions, for clarity.
I've dropped the **specific** SASL error message and inverted the `if`s.

**Before**:
```c
STAILQ_FOREACH(np, &imap_auth_methods->head, entries)
{
  if (!imap_auth_is_valid(np->data))
  {
#ifdef USE_SASL
    if (!sasl_auth_validator(np->data))
    {
      mutt_buffer_printf(err, _("Option %s: %s invalid sasl authenticator"),
                         cdef->name, np->data);
      return CSR_ERR_INVALID;
    }
#else
    mutt_buffer_printf(err, _("Option %s: %s is not a valid authenticator"),
                       cdef->name, np->data);
    return CSR_ERR_INVALID;
#endif
  }
}
```

**After**:
```c
STAILQ_FOREACH(np, &imap_auth_methods->head, entries)
{
  if (imap_auth_is_valid(np->data))
    continue;
#ifdef USE_SASL
  if (sasl_auth_validator(np->data))
    continue;
#endif
  mutt_buffer_printf(err, _("Option %s: %s is not a valid authenticator"),
                     cdef->name, np->data);
  return CSR_ERR_INVALID;
}
```

flatcap:
I've split the changes into 5 commits: sasl, imap, pop, smtp and nntp.

yousefakbar:
Looks good to me 👍🏼 .

conn/sasl.c
```diff
@@ -107,6 +116,23 @@ static sasl_callback_t MuttSaslCallbacks[5];
...
+ * sasl_auth_validator - Validate an auth method against Cyrus SASL methods
+ * @param authenticator Name of the authenticator to validate
+ * @retval bool True if argument matches an accepted auth method
+ */
+bool sasl_auth_validator(const char *authenticator)
+{
+  for (size_t i = 0; i < mutt_array_size(sasl_authenticators); i++)
+  {
+    const char *auth = sasl_authenticators[i];
+    if (auth && mutt_istr_equal(auth, authenticator))
```
gahr, line 129: Can `auth` ever be null?

conn/sasl.c
```diff
@@ -96,6 +96,15 @@ struct SaslSockData
   int (*close)(struct Connection *conn);
 };
 
+/**
+ * sasl_authenticators - Authenticaion methods supported by Cyrus SASL
+ */
+static char *sasl_authenticators[] = {
```
gahr, line 102: static const char * const?

conn/sasl.h
```diff
@@ -29,6 +29,8 @@
 
 struct Connection;
 
+bool sasl_auth_validator(const char* authenticator);
```
gahr, line 32: `*` goes with the name, not the type

nntp/nntp.c
```diff
@@ -109,6 +109,31 @@ struct ChildCtx
   anum_t *child;
 };
 
+/**
+ * nntp_authenticators - Accepted authentication methods
+ */
+static char *nntp_authenticators[] = { "user", "anonymous", "cram-md5", "gssapi", "sasl" };
```
gahr, line 115: `static const char * const`

nntp/nntp.c
```diff
@@ -109,6 +109,31 @@ struct ChildCtx
...
+ *
+ * Validate whether an input string is an accepted nntp authentication method as
+ * defined by nntp_authenticators.
+ */
+bool nntp_auth_is_valid(const char *authenticator)
+{
+  for (size_t i = 0; i < mutt_array_size(nntp_authenticators); i++)
+  {
+    const char *auth = nntp_authenticators[i];
+    if (auth && mutt_istr_equal(auth, authenticator))
```
gahr, line 130: Can `auth` ever be null?

flatcap (changes requested):
I've found a problem with the SMTP changes; it's a little complicated to explain.

POP and IMAP follow the same pattern.

If `$auth_methods` is set, then:
- Loop through the user's `$auth_methods`
- Call the function `authenticate(data, name)`
  The SASL function **uses** the `name` parameter (most others ignore it)
- If one succeeds then we're finished

If `$auth_methods` isn't set, then:
- Loop through all the `auth_methods[]`
- Call the function `authenticate(data, NULL)`
  This time the SASL function uses a list of valid authentication methods which were obtained **from the server**
- If one succeeds then we're finished

---

In SMTP, the auth functions need an extra param (`name`).
- `smtp_auth_oauth()`, `smtp_auth_plain()`, `smtp_auth_sasl()`

In the first code block, we want to pass the name, telling SASL to use a **specific** method.
In the second code block, we'll pass `NULL` to let SASL negotiate a method (using `adata->auth_mechs`).

flatcap (approved):
I've removed the nntp changes.
They require a bit more research and enthusiasm than I can spare right now.

However the rest of the changes look great.
Thanks.

Neustradamus:
Thanks to @yousefakbar, @flatcap, @grawity about SCRAM:
- https://github.com/neomutt/neomutt/pull/2624 + https://github.com/neomutt/neomutt/commit/b7bcb9aef103943251305c05840cbe96c09a6508
- https://github.com/neomutt/neomutt/pull/2688 + https://github.com/neomutt/neomutt/commit/85437597de9e692fa8c0aeab7eaaf27e43c563c6

Linked to:
- https://github.com/neomutt/neomutt/issues/1846
- https://gitlab.com/muttmua/mutt/-/issues/174

Issue #2622, opened 2020-08-28
Compose Enhancements


flatcap:
**To do**:

- [ ] Add config item to toggle the visibility of custom headers
- [ ] libcompose - allowing us to split up compose.c -- See 9b348bef8
      Create `compose/config.c` for the config
- [ ] Add notification for `my_hdr` changes
- [ ] Add observer for `my_hdr` changes
      This will allow the user to add/remove headers at will
- [ ] Move the function data to become Window data
- [ ] Eliminate config global variables

matthewhughes934:
Looks like fun, thanks for splitting it nicely into a list!

flatcap:
I'm having some second thoughts about how my `my_hdr` command changes would work.

I liked the idea of composing an email, then adding a custom header at the press of a button.
Tying that into the `my_hdr` command gets a bit confusing though.

`my_hdr` and `unmy_hdr` have a permanence about them -- they affect all future emails.

Perhaps the `my_hdr` observers are still useful, but we also need a `<compose-add-one-off-header>` function.
We'd need a remove function too, which would need a menu to select _which_ header you mean.
Hmm...

flatcap:
For the next couple, you'll need to **create** a new notification type.

If you look at the Event type, subtype and payload in [enum NotifyType](https://github.com/neomutt/neomutt/blob/master/mutt/notify_type.h#L31-L45), it should be clear.
Find where `my_hdr` and `unmy_hdr` are processed and send a global Event (to `NeoMutt->notify`).

Setting up the observing side should be pretty simple, e.g. 31d786e73

The final part will be to use the observed Event to update (add/remove) user headers from the current Envelope.

matthewhughes934:
I've got a working (though still WIP) approach for redrawing the compose screen when 'my_hdr' is called under devel/matthugh/my-hdr-notifications. I need to give some thought into how best to handle sharing some logic currently in `parse_my_hdr` around finding a node in a header list and creating/updating as needed.

I'm yet to look into handling calls to 'unmy_hdr' here, but I think I will need to do some similar sharing for `parse_unmy_hdr` as I will do for `parse_my_hdr`, so hopefully once I have that done this will happen simply (famous last words)

flatcap:
That's looking good!

matthewhughes934:
[...]

PR #2614, opened 2020-08-26
Display custom headers in compose

matthewhughes934:
* **What does this PR do?**

Allows displaying customer headers under a "Headers: " prompt in the compose screen

* **Screenshots (if relevant)**

Appearance with no custom headers (unchanged):

```
        From: 
          To: 
          Cc:
         Bcc:
     Subject: 
    Reply-To:
         Fcc: 
```

Appearance with a single custom header:

```
        From:
          To:
          Cc:
         Bcc:
     Subject:
    Reply-To:
         Fcc:
     Headers: X-First-Header: 1

```

Appearance with five custom headers:

```
        From:
          To:
          Cc:
         Bcc:
     Subject:
    Reply-To:
         Fcc:
     Headers: X-First-Header: 1
              X-Second-Header: 2
              X-Third-Header: 3
              X-Fouth-Header: 4
              X-Fifth-Header: 5
```

Appearance with more than 5 custom headers:

```
        From:
          To:
          Cc:
         Bcc:
     Subject:
    Reply-To:
         Fcc:
     Headers: X-First-Header: 1
              X-Second-Header: 2
              X-Third-Header: 3
              X-Fouth-Header: 4
              ...
```

* **Does this PR meet the acceptance criteria?**

   - Documentation created/updated (you have to edit
     [doc/manual.xml.head](https://www.github.com/neomutt/neomutt/blob/master/doc/manual.xml.head)
     for that) - not sure if these docs need updating

* **What are the relevant issue numbers?**

#2582

PR #2626, opened 2020-08-28
Add show user headers config

matthewhughes934:
* **What does this PR do?**

Add `C_ComposeShowUserHeaders/compose_show_user_headers` which defaults to
true and determines whether user headers should be displayed in the
compose screen.

(See PR #2614 for more details on these custom headers)

This is the first part of #2622

### DIFF
--- a/compose.c
+++ b/compose.c
@@ -99,6 +99,7 @@
 char *C_ComposeFormat; ///< Config: printf-like format string for the Compose panel's status bar
 char *C_Ispell; ///< Config: External command to perform spell-checking
 unsigned char C_Postpone; ///< Config: Save messages to the #C_Postponed folder
+bool C_ComposeShowUserHeaders; ///< Config: Whether to display user-defined headers
 
 static const char *There_are_no_attachments = N_("There are no attachments");
 
@@ -494,7 +495,7 @@ static int calc_security(struct Email *e, short *rows)
 }
 
 /**
- * calc_user_hdrs - Calculate how many rows are needed for user headers
+ * calc_user_hdrs - Calculate how many rows are needed for user-defined headers
  * @param hdrs Header List
  * @retval num Rows needed, limited to #MAX_USER_HDR_ROWS
  */
@@ -542,7 +543,8 @@ static int calc_envelope(struct ComposeRedrawData *rd)
     rows += calc_address(&env->bcc, &rd->bcc_list, cols, &rd->bcc_rows);
   }
   rows += calc_security(e, &rd->sec_rows);
-  rows += calc_user_hdrs(&env->userhdrs);
+  if (C_ComposeShowUserHeaders)
+    rows += calc_user_hdrs(&env->userhdrs);
 
   return rows;
 }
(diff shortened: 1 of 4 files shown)


matthewhughes934:
I'm not sure whether the desired defaulting should be `true` or `false`.

I'm also not sure about the name of the config variable. But I don't think I've be sure of a variable name in my life...

flatcap:
> I'm not sure whether the desired defaulting is `true` or `false` (currently `true`)

I think `true` is a good choice, so existing users will see the new feature.

> I'm also not sure about the name of the config variable
> `$compose_show_user_headers`

The name looks good -- a scope and a descriptive name

mutt_config.c
```diff
@@ -250,6 +250,9 @@ struct ConfigDef MainVars[] = {
   { "compose_format", DT_STRING|R_MENU, &C_ComposeFormat, IP "-- NeoMutt: Compose  [Approx. msg size: %l   Atts: %a]%>-", 0, NULL,
     "printf-like format string for the Compose panel's status bar"
   },
+  { "compose_show_user_headers", DT_BOOL, &C_ComposeShowUserHeaders, true, 0, NULL,
+    "Controls whether or not customer headers are shown in the compose envelope"
```
flatcap, line 254:
```suggestion
    "Controls whether or not custom headers are shown in the compose envelope"
```

docs/config.c
```diff
@@ -567,6 +567,12 @@
 ** information on how to set $$compose_format.
 */
 
+{ "compose_show_user_headers", DT_BOOL, true },
+/*
+** .pp
+** When \fIset\fP, Neomutt will display custom headers on the compose screen.
```
flatcap, line 573: Can you add a sentence mentioning the `my_hdr` command.
If you quote it as `$my_hdr` it will be auto-linked in the manual.

flatcap (changes requested):

mutt_config.c
matthewhughes934: Well spotted, fixed with 22d5c38

docs/config.c
matthewhughes934: Added with 179d074. Also here used 'user-defined headers' over 'custom headers' (the former is more common in existing docs)

compose.c
```diff
@@ -99,6 +99,7 @@
 char *C_ComposeFormat; ///< Config: printf-like format string for the Compose panel's status bar
 char *C_Ispell; ///< Config: External command to perform spell-checking
 unsigned char C_Postpone; ///< Config: Save messages to the #C_Postponed folder
+bool C_ComposeShowUserHeaders; ///< Config: Whether to display customer user headers
```
matthewhughes934, line 102: Ughh I made the same typo here: 'customer'->'custom'. Following my changes to the docs I might just update this (and elsewhere in this file) to 'user-defined headers'. Thoughts/objections @flatcap ?

flatcap: Good idea

flatcap (approved):

Issue #2621, opened 2020-08-28
Address Book enhancements


flatcap:
**To Do**:

- [ ] Move the Address config into libaddress
- [ ] Add patterns for the Address fields
- [ ] Upgrade search to allow using patterns
- [ ] Add a `<limit>` function
- [ ] Move the function data to become Window data
- [ ] Eliminate config global variables

**Advanced**:

- Add functions for parity with other dialogs:
  `<tag-pattern>`, `<delete-pattern>`, `<undelete-pattern>`
- Use some formatted comments, e.g. `# label=red,blue` for labelling
  These would allow users to group their aliases
- Add a sidebar of labels, including "All", "Self" (from alternates), plus each user label

flatcap:
## Address Book - Add Patterns

Currently, the search works my matching a regex against the fulltext displayed
on the screen.  Adding the following patterns would allow more accurate
searching.

- `~f EXPR` - From (alias)
- `~t EXPR` - To (address)
- `~b EXPR` - Body (comment)
- `~Y EXPR` - Tag (label) -- not yet implemented

If the search didn't contain a `~`, then we'd use the old search method.

## Add a Limit Command

In the index, you can limit the view by using a pattern.

The Address Book is started via auto-completion on an Address.
If you type `joh<tab>` then only matching Addresses are shown.
Once in this limited view, you can't get back to the full list.

Rather than filtering the list, I suggest starting with a limit set.
This would allow the user to see the full list, by `<limit>.<enter>`

## Add Pattern Functions

Using the patterns, we can now add functions:

- `<tag-pattern>`
- `<delete-pattern>`
- `<undelete-pattern>`

RomeuG:
Hey @flatcap!
I would like to ask you how you would approach the `Address Book - Add Patterns` feature.

I have been trying to understand how the Patterns in the index window work. And I think I did figure it out.
My problem at the moment is how to actually fit the Address Book into this.

I know that when we search in the index window, `mutt_search_command` is executed.
Inside it, if there is no cache, `mutt_pattern_comp` will be executed and it will define the op and tag of the search input (e.g. if I use `~t Romeu` as the input, it will search for the next e-mail that has `Romeu` in the `To:` field.).
Finally, `mutt_pattern_exec` will be used to check if it matches or not.

[...]

PR #2627, opened 2020-08-28
Issue #2621: Move address config into libaddress

RomeuG:
* **What does this PR do?**

This PR moves Address configuration into its own configuration inside `alias`.
Then it is initialized through `CONFIG_INIT_VARS`.

The relevant issue number is #2621 , being this PR related to the first bullet point.

Hope everything okay. I did test it, everything seems to be working, sorting, alias file, etc.

### DIFF
--- a/alias/config.c
+++ b/alias/config.c
@@ -1,0 +1,78 @@
+/**
+ * @file
+ * Config used by libaddress
+ *
+ * @authors
+ * Copyright (C) 2020 Richard Russon <rich@flatcap.org>
+ *
+ * @copyright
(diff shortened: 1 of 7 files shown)


alias/config.c
```diff
@@ -0,0 +1,54 @@
+#include "config.h"
```
flatcap, line 1: Can you add a Copyright header block here.
We also need a Doxygen @page block, which needs linking up in `alias/lib.h`

RomeuG: Sorry! My bad.

Done! :-)

flatcap (approved):

PR #2633, opened 2020-09-03
Issue #2621: Add patterns to alias/query

RomeuG:
* **What does this PR do?**

This PR adds the possibility of using patterns (e.g.: `~b EXPR` to search comments`) in the Alias and Query menus.
In this PR there are three different patterns that can be used:

- `~f EXPR` - Alias
- `~t EXPR` - Address
- `~b EXPR` - Comment

The relevant issue number where this was discussed is #2621.

### DIFF
--- a/alias/dlgalias.c
+++ b/alias/dlgalias.c
@@ -41,6 +41,7 @@
 #include "gui/lib.h"
 #include "mutt.h"
 #include "lib.h"
+#include "pattern/lib.h"
 #include "alias.h"
 #include "format_flags.h"
 #include "gui.h"
@@ -195,6 +196,7 @@ static void dlg_select_alias(char *buf, size_t buflen, struct AliasMenuData *mda
   dlg->help_menu = MENU_ALIAS;
 
   menu->make_entry = alias_make_entry;
+  menu->custom_search = true;
   menu->tag = alias_tag;
   menu->title = _("Aliases");
   menu->max = ARRAY_SIZE(mdata);
@@ -284,6 +286,17 @@ static void dlg_select_alias(char *buf, size_t buflen, struct AliasMenuData *mda
 
         break;
       }
+      case OP_SEARCH_REVERSE:
+      case OP_SEARCH_NEXT:
+      case OP_SEARCH_OPPOSITE:
+      case OP_SEARCH:
+        menu->current = mutt_search_alias_command(menu, menu->current, op);
+        if (menu->current == -1)
+          menu->current = menu->oldcurrent;
+        else
+          menu->redraw |= REDRAW_MOTION;
+        break;
+
       case OP_GENERIC_SELECT_ENTRY:
         t = menu->current;
         if (t >= ARRAY_SIZE(mdata))
--- a/alias/dlgquery.c
+++ b/alias/dlgquery.c
@@ -40,6 +40,7 @@
 #include "gui/lib.h"
 #include "mutt.h"
 #include "lib.h"
+#include "pattern/lib.h"
 #include "send/lib.h"
 #include "alias.h"
 #include "format_flags.h"
@@ -321,6 +322,7 @@ static void dlg_select_query(char *buf, size_t buflen, struct AliasList *all, bo
 
   menu->make_entry = query_make_entry;
   menu->search = query_search;
+  menu->custom_search = true;
   menu->tag = query_tag;
   menu->title = title;
   menu->max = ARRAY_SIZE(&mdata);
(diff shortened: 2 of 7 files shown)


gahr:
Please use the ARRAY API to access AliasMenuData items. See f3f108f626a2193eb54b60bededb46b528dcde5b.

flatcap:
It's looking very nice so far.

I was right about 'and' and 'or', see c40e4432f.
The two cases get hit when searching for complex patterns.

If you search for `~f a ~t e`, then the generated `SearchPattern` will look something like:
```
     PAT_AND
     /     \
PAT_FROM   PAT_TO
  "a"        "e"
```

`~f a | ~t e` will generate the same with `PAT_OR` at the top.

The `and()` and `or()` functions will perform the logic and recurse to `mutt_pattern_alias_exec()` to perform the `~f` and `~t` matching.

flatcap:
Thinking further ahead (don't change anything yet)...
I think `mutt_pattern_alias_exec()` should live in `alias/`, and `mutt_pattern_exec()` should live somewhere email-y.
That would mean `pattern/` is just the machinery of pattern matching.

RomeuG:
@flatcap @gahr 

Hey!

So now it is using the ARRAY API (I think everything is correct).
And I also did some changes because of complex patterns.

I still didn't change anything related to where the functions should be... tell me what you think! :-)

flatcap:
The ARRAY changes look good.
I have a few more small requests, please.

- Move the pattern functions to `alias/gui.c` (as they're Alias-specific)
  `mutt_pattern_alias_exec()`, `perform_alias_and()`, `perform_alias_or()`

- Rename the `struct AliasView` parameters to `av`.
  I'm starting to sort out the naming (see also [Data and Views](https://github.com/neomutt/gfx/blob/master/arch/README.md#data-and-views))
  The bare object might get a single letter name, then it would be suffixed with 'l' (list), 'v' (view)

- Rename `AliasView.searched` and `.matched` to `.is_searched` and `.is_matched`
  In some code `deleted` is a bool and others it's the number of deleted messages

- Change `AliasView.is_tagged` and `.is_deleted` into `: 1` bitfields
  (that's my mistake, not yours)

- Add some general patterns (for parity with the Index)
  (this should be another quick cut'n'paste job)
  - `~A` (`MUTT_ALL`)
  - `~D` (`MUTT_DELETED`) -> `Alias.is_deleted`
  - `~m` (`MUTT_PAT_MESSAGE`) -> `Alias.num`
  - `~T` (`MUTT_TAG`) -> `Alias.is_tagged`

Thanks.

---

The Alias code is quickly becoming the perfect prototype for all dialogs in NeoMutt :+1:

RomeuG:
@flatcap 

Just made the changes you pointed out.

The most confusing point was moving the pattern code that was specific to Alias, to the Alias folder. I made some decisions that I don't know if they are to your liking.

Thanks!

flatcap:
hmm... OK, it's a mess (but that's not your fault :-)
I'm not sure how to break up the pattern code, yet.
Suggestions welcome :-)

RomeuG:
Hey!

I am sorry, I am trying... :disappointed: 

Generic and general parts of the pattern code can surely be in libpattern.
Regarding the more specific parts... they could still be there to be honest, but probably in different files?

What should I do to progress this PR?

Thanks!

flatcap:
> Regarding the more specific parts...
> but probably in different files?

hmm... yes.

Sorry to be a pain.
Please can you put the Alias Pattern code back in `pattern/exec.c`, `pattern/pattern.c`

We can work out how to break up the deps another day.

> What should I do to progress this PR?

The code works, so please can you just add some docs.
I suggest duplicating these three rows and creating Alias-specific entries:

https://github.com/neomutt/neomutt/blob/fad7b512b62075b077da23157ec5972dfaf90646/docs/manual.xml.head#L7949-L7955

https://github.com/neomutt/neomutt/blob/fad7b512b62075b077da23157ec5972dfaf90646/docs/manual.xml.head#L8023-L8029

https://github.com/neomutt/neomutt/blob/fad7b512b62075b077da23157ec5972dfaf90646/docs/manual.xml.head#L8247-L8253

RomeuG:
Reverted the commit where I did move the functions and I added documentation!
I was not sure how to structure the documentation, so I added a new table that is specific for Alias Pattern modifiers.

Thanks!

flatcap:
> Reverted the commit where I did move the functions

Ah!  About that.  That commit also contained some other changes
(such as renaming the AliasView vars to `av` and the members to `is_*`)
I've restored the changes.

> and I added documentation!
> I added a new table that is specific for Alias Pattern modifiers.

Oh, that's much better.

Creating this PR has meant that I've had to read and re-read all the docs for the Address Book.
Everything you need to know about **one** dialog, is scattered over an 86000 word document.

I'll raise an issue to discuss what the perfect doc, for this dialog, would look like.

flatcap:
I picked the patterns to try to fit in with the existing letters.
That was a bit strained.

I think we should change one final thing...

Change `~b` (body) to `~c` (comment)

RomeuG:
I think this is it! :-)

flatcap (approved):

PR #2639, opened 2020-09-05
build: Add the Address Sanitizer

flatcap:
I'm not sure why we didn't think to add this sooner.

However, I couldn't figure out how to check if ASAN was actually installed.

I tried:
```
if [cc-with { -libs -fsanitize=address }] {
```
and
```
if [cctest -link 1 -libs -fsanitize=address] { 
```

Nothing quite worked.
Those that didn't fail, also meant that all the later checks had the Sanitizer compiles flags, too.

### DIFF
--- a/auto.def
+++ b/auto.def
@@ -113,8 +113,9 @@ options {
 # Misc
   include-path-in-cflags=1  => "Remove include paths from CFLAGS in the output of neomutt -v"
 # Testing
-  testing=0                 => "Enable Unit Testing"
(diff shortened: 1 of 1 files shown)


gahr:
I can look into making it fail if it’s not available tomorrow, or we can merge it as is. It’s simple and off by default, I don’t see much of a problem.

gahr (approved):

flatcap:
> I can look into making it fail if it’s not available tomorrow

Please, if you can spare it ten minutes (no longer, though :-)

PR #2645, opened 2020-09-09
alias: fix simple searches

flatcap:
I've changed the "simple search" case for Aliases..

If the user doesn't put a pattern in the search, the code used `C_SimpleSearch` to expand the pattern.
Unfortunately, that isn't meaningful for Aliases (wrong `~` patterns).

It now expands to `~f %s | ~t %s | ~c %s`

This is hard-coded to match on **any** Alias field.
I'm not sure if it's worth making this into a config variable.

Any thoughts?

### DIFF
--- a/pattern/exec.c
+++ b/pattern/exec.c
@@ -339,7 +339,7 @@ static int perform_alias_or(struct PatternList *pat, PatternExecFlags flags,
 }
 
 /**
- * match_addrlist - Match a Pattern against and Address list
+ * match_addrlist - Match a Pattern against an Address list
  * @param pat            Pattern to find
  * @param match_personal If true, also match the pattern against the real name
  * @param n              Number of Addresses supplied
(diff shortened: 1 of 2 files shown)


RomeuG:
Hey! 

It makes total sense.

But, in my opinion, I would a new global variable in `mutt_globals.h` with a name along the lines of `C_SimpleSearchAlias` and then have a new added element to `MainVars` in `mutt_config.c`.

With what exists, I guess it could be the ideal? If patterns are added to other areas, this could be the way of creating an area-specific simple search variable.

Regardless, this is just a suggestion of mine, even though I do not have that much experience in this project, yet.

I hope I helped! :-)

flatcap:
> I would add a new global variable

I'll leave it for now.
For the Aliases, we want to search **all** the possible fields.
For the Index, we only search a couple, but the user might configure more.

Issue #2641, opened 2020-09-07
Refactor the Commands list


flatcap:
All of NeoMutt's commands are stored in one large static array: [`Commands`](https://github.com/neomutt/neomutt/blob/master/mutt_commands.c#L50).

Each time a new feature adds a command, this central point needs to be updated.
This makes the `mutt_commands.c` file into a dependency nightmare.

## Task 1

Create a new Command dynamic array, using the new [`ARRAY` functions](https://github.com/neomutt/neomutt/blob/master/mutt/array.h)

- Create setup and shutdown functions to control its lifespan (and free any memory)

- Initialise the `ARRAY` with the contents of the original `Commands`

- After adding to the `ARRAY`, sort it by command name

## Task 2

Break up the dependencies.

There are four features that define their own commands
(look for `USE_COMP_MBOX`, `USE_IMAP`, `USE_LUA` and `USE_SIDEBAR`)

All of the commands in these `#ifdef`s should be moved out to the feature's source dir.
Each feature will need an `init()` function that registers their commands.

louis77:
I've commited the first draft of this. Commands for the following features were moved:

- `USE_COMP_MBOX`
- `USE_IMAP`
- `USE_SIDEBAR`

Questions so far:

1. Is the overall approach ok?
2. What about `USE_NOTMUCH`?
3. `USE_LUA` is different that the other features. Should I move the commands to `mutt_lua.c`?

flatcap:
Yes, that's looking very nice.

1. Yes
2. Yes.  How did I miss that from the list?
3. Yes

Thanks :-)

PR #2647, opened 2020-09-09
Devel/commands refactor

louis77:
* **What does this PR do?**

Refactor the Commands list: feature specific commands are now moved into the feature source. 
Mutt_commands.c/h provides a facility for features to register their commands on init into a new dynamic Commands array. Commands for the following features were moved:

- `USE_COMP_MBOX`
- `USE_IMAP`
- `USE_LUA`
- `USE_NOTMUCH`
- `USE_SIDEBAR`

#2641

### DIFF
--- a/mutt_commands.c
+++ b/mutt_commands.c
@@ -46,23 +46,17 @@
 #include "sidebar/lib.h"
 #endif
 
-// clang-format off
-const struct Command Commands[] = {
+const struct Command mutt_commands[] = {
+  // clang-format off
   { "account-hook",        mutt_parse_hook,        MUTT_ACCOUNT_HOOK },
   { "alias",               parse_alias,            0 },
   { "alternates",          parse_alternates,       0 },
   { "alternative_order",   parse_stailq,           IP &AlternativeOrderList },
-#ifdef USE_COMP_MBOX
-  { "append-hook",         mutt_parse_hook,        MUTT_APPEND_HOOK },
-#endif
   { "attachments",         parse_attachments,      0 },
   { "auto_view",           parse_stailq,           IP &AutoViewList },
   { "bind",                mutt_parse_bind,        0 },
   { "cd",                  parse_cd,               0 },
   { "charset-hook",        mutt_parse_hook,        MUTT_CHARSET_HOOK },
-#ifdef USE_COMP_MBOX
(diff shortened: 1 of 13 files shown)


mutt_commands.h
```diff
@@ -72,6 +73,11 @@ enum MuttSetCommand
 /* parameter to parse_mailboxes */
 #define MUTT_NAMED   (1 << 0)
 
-extern const struct Command Commands[];
+/* command registry functions */
+#define COMMANDS_REGISTER(cmds) commands_register(cmds, sizeof(cmds) / sizeof(cmds[0]))
```
gahr, line 77: We have `mutt_array_size1 in `mutt/memory.h`

mutt_commands.c
```diff
@@ -147,18 +129,54 @@ const struct Command Commands[] = {
...
-  { "virtual-mailboxes",   parse_mailboxes,        MUTT_NAMED },
-#endif
-  { NULL,                  NULL,                   0 },
 };
+
 // clang-format on
+
+ARRAY_HEAD(CommandsData, struct Command);
+struct CommandsData commands = ARRAY_HEAD_INITIALIZER;
+const struct Command *Commands;
```
gahr, line 140: Why do you keep this around?

mutt_commands.h
louis77: Using it now, thanks for the hint.

mutt_commands.c
louis77: I've eliminated the redundant stuff. Let me know if it looks better now.

gahr: Uhm I'm still wondering why you need the global `Commands` pointer. Can't we have APIs to access the local `commands` array?

louis77: As discussed with @gahr and @flatcap on IRC remove the global `Commands` array and provider further API to access the `commands` array.

louis77:
Moved to `mutt_commands.c`:
- `mutt_command_get`
- `mutt_commands_apply`

Removed global `Commands` array, replacement:
- `mutt_commands_array`

The following functions use the above:
- `mutt_get_hook_type`
- `mutt_command_complete`
- `mutt_parse_rc_buffer`

- [ ] The latter could be optimized by calling `mutt_command_get` instead

flatcap (approved):

PR #2650, opened 2020-09-10
trans: split up usage and options strings

flatcap:
I've split up the two massive strings in `main.c` (and split up all the translation files too).

https://github.com/neomutt/neomutt/blob/4bcb54f97eb7bcbaa0e9490fd34ddc9c2f74c209/main.c#L153-L170

https://github.com/neomutt/neomutt/blob/4bcb54f97eb7bcbaa0e9490fd34ddc9c2f74c209/main.c#L173-L212

Each option is now in a separate translation string (47 strings total).
This means that if we add a new CLI option to NeoMutt, it will be a simple job to update the translation.

However... Having separate strings will make it harder to keep the text aligned.

@neomutt/translators 
Thoughts please.


po/lt.po
```diff
@@ -3207,129 +3207,234 @@ msgstr "Rašau %s..."
 
 #. L10N: Try to limit to 80 columns
 #: main.c:153
+msgid "usage:"
+msgstr "usage:"
```
mgedmin, line 3211: Well, at least one good thing came out of this PR: I noticed that this bit was never translated!

mgedmin (approved):
LGTM!  Alignment doesn't seem too hard, at least if the translation keeps the original alignment instead of shifting everything.

(For vim users: `:set cursorcolumn` can help.)

flatcap:
Great!

PR #2658, opened 2020-09-13
Add <limit> function to Alias/Query dialogs

RomeuG:
* **What does this PR do?**

The purpose of this PR is to present a working and complete `<limit>` function in the Alias/Query menus.

The final version of the PR will take everything into account regarding Documentation, Formatting, etc.

#2621

### DIFF
--- a/alias/dlgalias.c
+++ b/alias/dlgalias.c
@@ -121,8 +121,8 @@ static const char *alias_format_str(char *buf, size_t buflen, size_t col, int co
  */
 static void alias_make_entry(char *buf, size_t buflen, struct Menu *menu, int line)
 {
-  const struct AliasMenuData *mdata = menu->mdata;
-  const struct AliasView *av = ARRAY_GET(mdata, line);
+  const struct AliasViewArray *ava = &((struct AliasMenuData *) menu->mdata)->ava;
+  const struct AliasView *av = ARRAY_GET(ava, line);
 
   mutt_expando_format(buf, buflen, 0, menu->win_index->state.cols, NONULL(C_AliasFormat),
                       alias_format_str, IP av, MUTT_FORMAT_ARROWCURSOR);
@@ -133,8 +133,8 @@ static void alias_make_entry(char *buf, size_t buflen, struct Menu *menu, int li
  */
 static int alias_tag(struct Menu *menu, int sel, int act)
 {
-  struct AliasMenuData *mdata = (struct AliasMenuData *) menu->mdata;
-  struct AliasView *av = ARRAY_GET(mdata, sel);
+  const struct AliasViewArray *ava = &((struct AliasMenuData *) menu->mdata)->ava;
+  struct AliasView *av = ARRAY_GET(ava, sel);
 
   bool ot = av->is_tagged;
 
@@ -160,16 +160,25 @@ static int alias_alias_observer(struct NotifyCallback *nc)
 
   if (nc->event_subtype == NT_ALIAS_NEW)
   {
-    menu_data_alias_add(mdata, alias);
+    alias_array_alias_add(&mdata->ava, alias);
+
+    if (alias_array_count_visible(&mdata->ava) != ARRAY_SIZE(&mdata->ava))
+      mutt_pattern_alias_func(MUTT_LIMIT, NULL, _("Aliases"), mdata, menu);
   }
   else if (nc->event_subtype == NT_ALIAS_DELETED)
   {
-    menu_data_alias_delete(mdata, alias);
-  }
-  menu_data_sort(mdata);
+    alias_array_alias_delete(&mdata->ava, alias);
 
-  menu->max = ARRAY_SIZE(mdata);
+    int vcount = alias_array_count_visible(&mdata->ava);
+    if ((menu->current > (vcount - 1)) && (menu->current > 0))
+      menu->current--;
+  }
+
+  alias_array_sort(&mdata->ava);
+
+  menu->max = alias_array_count_visible(&mdata->ava);
   menu->redraw = REDRAW_FULL;
+
   return 0;
 }
 
@@ -181,7 +190,7 @@ static int alias_alias_observer(struct NotifyCallback *nc)
  */
 static void dlg_select_alias(char *buf, size_t buflen, struct AliasMenuData *mdata)
 {
-  if (ARRAY_EMPTY(mdata))
+  if (ARRAY_EMPTY(&mdata->ava))
   {
     mutt_warning(_("You have no aliases"));
     return;
@@ -198,9 +207,9 @@ static void dlg_select_alias(char *buf, size_t buflen, struct AliasMenuData *mda
   menu->make_entry = alias_make_entry;
   menu->custom_search = true;
   menu->tag = alias_tag;
-  menu->title = _("Aliases");
-  menu->max = ARRAY_SIZE(mdata);
+  menu->max = alias_array_count_visible(&mdata->ava);
   menu->mdata = mdata;
+  menu->title = menu_create_alias_title(_("Aliases"), mdata->str);
 
   notify_observer_add(NeoMutt->notify, NT_ALIAS, alias_alias_observer, menu);
   notify_observer_add(NeoMutt->notify, NT_CONFIG, alias_config_observer, mdata);
@@ -208,10 +217,10 @@ static void dlg_select_alias(char *buf, size_t buflen, struct AliasMenuData *mda
 
   mutt_menu_push_current(menu);
 
-  ARRAY_SORT(mdata, alias_get_sort_function(C_SortAlias));
+  ARRAY_SORT(&mdata->ava, alias_get_sort_function(C_SortAlias));
 
   struct AliasView *avp = NULL;
-  ARRAY_FOREACH(avp, mdata)
+  ARRAY_FOREACH(avp, &mdata->ava)
   {
     avp->num = ARRAY_FOREACH_IDX;
   }
@@ -225,7 +234,7 @@ static void dlg_select_alias(char *buf, size_t buflen, struct AliasMenuData *mda
       case OP_UNDELETE:
         if (menu->tagprefix)
         {
-          ARRAY_FOREACH(avp, mdata)
+          ARRAY_FOREACH(avp, &mdata->ava)
           {
             if (avp->is_tagged)
               avp->is_deleted = (op == OP_DELETE);
@@ -234,7 +243,7 @@ static void dlg_select_alias(char *buf, size_t buflen, struct AliasMenuData *mda
         }
         else
         {
-          ARRAY_GET(mdata, menu->current)->is_deleted = (op == OP_DELETE);
+          ARRAY_GET(&mdata->ava, menu->current)->is_deleted = (op == OP_DELETE);
           menu->redraw |= REDRAW_CURRENT;
           if (C_Resolve && (menu->current < menu->max - 1))
           {
@@ -297,9 +306,22 @@ static void dlg_select_alias(char *buf, size_t buflen, struct AliasMenuData *mda
           menu->redraw |= REDRAW_MOTION;
         break;
 
+      case OP_MAIN_LIMIT:
+      {
+        int result = mutt_pattern_alias_func(MUTT_LIMIT, _("Limit to messages matching: "),
+                                             _("Aliases"), mdata, menu);
+        if (result == 0)
+        {
+          alias_array_sort(&mdata->ava);
+          menu->redraw = REDRAW_FULL;
+        }
+
+        break;
+      }
+
       case OP_GENERIC_SELECT_ENTRY:
         t = menu->current;
-        if (t >= ARRAY_SIZE(mdata))
+        if (t >= ARRAY_SIZE(&mdata->ava))
           t = -1;
         done = true;
         break;
@@ -309,7 +331,7 @@ static void dlg_select_alias(char *buf, size_t buflen, struct AliasMenuData *mda
     }
   }
 
-  ARRAY_FOREACH(avp, mdata)
+  ARRAY_FOREACH(avp, &mdata->ava)
   {
     if (avp->is_tagged)
     {
@@ -320,7 +342,7 @@ static void dlg_select_alias(char *buf, size_t buflen, struct AliasMenuData *mda
 
   if (t != -1)
   {
-    mutt_addrlist_write(&ARRAY_GET(mdata, t)->alias->addr, buf, buflen, true);
+    mutt_addrlist_write(&ARRAY_GET(&mdata->ava, t)->alias->addr, buf, buflen, true);
   }
 
   notify_observer_remove(NeoMutt->notify, alias_alias_observer, menu);
@@ -328,7 +350,10 @@ static void dlg_select_alias(char *buf, size_t buflen, struct AliasMenuData *mda
   notify_observer_remove(NeoMutt->notify, alias_color_observer, menu);
 
   mutt_menu_pop_current(menu);
+
+  FREE(&menu->title);
   mutt_menu_free(&menu);
+
   dialog_destroy_simple_index(&dlg);
 }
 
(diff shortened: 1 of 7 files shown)


RomeuG:
@flatcap 

I am a bit worried about setting the menu title.

I am setting it in the `mutt_pattern_alias_func` in `pattern/pattern.c`.

```
  char *tmp_str = NULL;
  char *new_title = NULL;

  mutt_str_asprintf(&tmp_str, _("Limit: %s"), buf->data);
  mutt_str_asprintf(&new_title, "Aliases - %s", tmp_str);

  m->title = new_title;
```

I am worried because of possible memory leaks after creating the new string.

flatcap:
> I am worried because of possible memory leaks

Good :-)
We'll need to `FREE()` any existing string in `Menu.title`, and also free it before we `menu_free()`

flatcap:
In your example you'd need to free `tmp_str` too.

alias/gui.c
```diff
@@ -53,7 +53,31 @@ int alias_sort_name(const void *a, const void *b)
...
-  int r = mutt_str_coll(av_a->alias->name, av_b->alias->name);
+  int r;
+
+  if (av_a->is_visible == av_b->is_visible) {
+    r = mutt_str_coll(av_a->alias->name, av_b->alias->name);
+  } else {
+    r = (av_a->is_visible == av_b->is_visible) ? 0 : av_a->is_visible ? -1 : 1;
+  }
+
+  return RSORT(r);
```
flatcap, line 64: I think we need to deal with the `is_visible` case first, and `return`.
The code above applies `RSORT(r)` to the result, even if hidden.
If `sort = reverse-name`, I _think_ we might get:
```
hidden1
hidden2
zebra
yak
xebu
```

RomeuG: Most likely I am misunderstanding something.

How would you """intersect""" two different values in a sorting function?
If I deal with the `is_visible` case first, what about the `name` sorting?

flatcap: If only one is visible, then `$sort` doesn't matter.
Make sure the hidden Alias is moved towards the end.

```c
if (av_a->is_visible != av_b->is_visible)
  return av_a->is_visible ? -1 : 1;
```

Check for both being invisible:

```c
if (!av_a->is_visible)
  return 0;
```

Then sort by name (honouring `reverse-`):

```c
int r = mutt_str_coll(av_a->alias->name, av_b->alias->name);
return RSORT(r);
```

(I haven't tested this)

RomeuG: I _think_ I got it working... but I will apply the same logic to the other sorting functions.
I still have to apply all these changes to the query menu, which is something that I will do tomorrow. :-)

RomeuG:
@flatcap 

Hi! I hope you are well!

This isn't totally finished yet, I still have to fix the menu title.
And there is still an issue about your recommendation about using a new struct, because I did not find any uses for the other two attributes (string and Pattern).

For what I see, I think everything seems to work OK. I did some tests on my end, didn't find anything relevant to point out.

Could you do some tests yourself? Maybe you have better ideas on how to test it.

Thanks! :smile:

alias/dlgalias.c
```diff
@@ -297,9 +298,21 @@ static void dlg_select_alias(char *buf, size_t buflen, struct AliasMenuData *mda
...
         break;
 
+      case OP_MAIN_LIMIT:
+      {
+        int result = mutt_pattern_alias_func(
+            MUTT_LIMIT, _("Limit to messages matching: "), &mdata->marray, menu);
+
+
+        menu_data_sort(&mdata->marray);
+        menu->redraw = REDRAW_FULL;
```
flatcap, line 308: these should only be run if the limit succeeds `(result == 0)`

alias/dlgquery.c
```diff
@@ -502,6 +504,17 @@ static void dlg_select_query(char *buf, size_t buflen, struct AliasList *all, bo
           menu->redraw |= REDRAW_MOTION;
         break;
 
+      case OP_MAIN_LIMIT:
+      {
+        int result = mutt_pattern_alias_func(
+            MUTT_LIMIT, _("Limit to messages matching: "), &mdata.marray, menu);
+
+        menu_data_sort(&mdata.marray);
+        menu->redraw = REDRAW_FULL;
```
flatcap, line 513: these should only be run if the limit succeeds `(result == 0)`

alias/dlgquery.c
```diff
@@ -302,13 +302,15 @@ static int query_run(char *s, bool verbose, struct AliasList *al)
  */
 static void dlg_select_query(char *buf, size_t buflen, struct AliasList *all, bool retbuf)
 {
-  struct AliasMenuData mdata = ARRAY_HEAD_INITIALIZER;
+  //struct AliasMenuArray mdata = ARRAY_HEAD_INITIALIZER;
```
flatcap, line 305: dead

alias/dlgquery.c
```diff
@@ -364,12 +366,12 @@ static void dlg_select_query(char *buf, size_t buflen, struct AliasList *all, bo
         struct Alias *tmp = NULL;
         TAILQ_FOREACH_SAFE(np, &al, entries, tmp)
         {
-          menu_data_alias_add(&mdata, np);
+          menu_data_alias_add(&mdata.marray, np);
           TAILQ_REMOVE(&al, np, entries);
           TAILQ_INSERT_TAIL(all, np, entries); // Transfer
         }
         menu_data_sort(&mdata);
```
flatcap, line 373: This should be `menu_data_sort(&mdata.marray);`

alias/gui.c
```diff
@@ -53,11 +53,33 @@ int alias_sort_name(const void *a, const void *b)
   const struct AliasView *av_a = a;
```
flatcap, line 53: As this is a standard `sort` callback function, we can be brief in the Doxygen comment.
Add ` - Implements ::sort_t` to the description, and drop all the `@param`s and `@retval`s.

Add a `@note` mentioning that non-visible Aliases will be sorted to the end

alias/gui.c
```diff
@@ -53,11 +53,33 @@ int alias_sort_name(const void *a, const void *b)
...
   const struct AliasView *av_b = b;
 
+  if (av_a->is_visible != av_b->is_visible) {
+    return av_a->is_visible ? -1 : 1;
+  }
+
```
flatcap, line 59: We're missing the 'both are invisible` check
```c
if (!av_a->is_visible)
  return 0;
```

alias/gui.c
```diff
@@ -53,11 +53,33 @@ int alias_sort_name(const void *a, const void *b)
...
   int r = mutt_str_coll(av_a->alias->name, av_b->alias->name);
 
   return RSORT(r);
 }
 
+/**
+ * alias_sort_visibility - Compare two Aliases by their visibility
```
flatcap, line 66: This function isn't used

alias/gui.c
```diff
@@ -53,11 +53,33 @@ int alias_sort_name(const void *a, const void *b)
...
+  const struct AliasView *av_b = b;
+
+  int r = (av_a->is_visible == av_b->is_visible) ? 0 : av_a->is_visible ? 1 : -1;
+
+  return RSORT(r);
+}
+
 /**
  * alias_sort_address - Compare two Aliases by their Addresses
  * @param a First  Alias to compare
```
flatcap, line 85: Add `Implements`, drop `@param`, drop `@retval`, add `@note` about invisibility

alias/gui.c
```diff
@@ -68,8 +90,16 @@ int alias_sort_name(const void *a, const void *b)
...
+  const struct AliasView *av_a = a;
+  const struct AliasView *av_b = b;
+
+  const struct AddressList *al_a = &av_a->alias->addr;
+  const struct AddressList *al_b = &av_b->alias->addr;
+
+  if (av_a->is_visible != av_b->is_visible)
+  {
+    return av_a->is_visible ? -1 : 1;
+  }
```
flatcap, line 102: We're missing the 'both are invisible` check
```c
if (!av_a->is_visible)
  return 0;
```

alias/gui.c
```diff
@@ -107,6 +137,11 @@ int alias_sort_unsort(const void *a, const void *b)
   const struct AliasView *av_a = a;
```
flatcap, line 137: Add `@note` about invisibility

alias/gui.c
```diff
@@ -107,6 +137,11 @@ int alias_sort_unsort(const void *a, const void *b)
...
   const struct AliasView *av_b = b;
 
+  if (av_a->is_visible != av_b->is_visible)
+  {
+    return av_a->is_visible ? -1 : 1;
+  }
```
flatcap, line 143: We're missing the 'both are invisible` check
```c
if (!av_a->is_visible)
  return 0;
```

alias/gui.h
```diff
@@ -40,17 +40,25 @@ struct AliasView
...
   bool is_tagged   : 1; ///< Is it tagged?
   bool is_deleted  : 1; ///< Is it deleted?
+  bool is_visible  : 1; ///< Is visible?
   struct Alias *alias;  ///< Alias
 };
 
-ARRAY_HEAD(AliasMenuData, struct AliasView);
+ARRAY_HEAD(AliasMenuArray, struct AliasView);
+
+struct AliasMenuData
```
flatcap, line 49: Add Doxygen comment block describing `struct`

alias/gui.h
```diff
@@ -40,17 +40,25 @@ struct AliasView
...
+{
+  char* str;
+  struct Pattern *pat;
+  struct AliasMenuArray marray;
```
flatcap, line 53: Add Doxygen comments `///<` for each

pattern/lib.h
```diff
@@ -183,6 +183,8 @@ int mutt_which_case(const char *s);
 int mutt_is_list_recipient(bool all_addr, struct Envelope *e);
 int mutt_is_subscribed_list_recipient(bool all_addr, struct Envelope *e);
 int mutt_pattern_func(int op, char *prompt);
+int mutt_pattern_alias_func(int op, char *prompt, struct AliasMenuArray *marray,
+                            struct Menu *m);
```
flatcap, line 187: Build warnings from using `struct AliasMenuArray` without declaring it.
Add, this after the `#include`s
```c
struct AliasMenuArray;
```

pattern/pattern.c
```diff
@@ -223,6 +223,87 @@ bool mutt_limit_current_thread(struct Email *e)
   return true;
 }
 
+int mutt_pattern_alias_func(int op, char *prompt, struct AliasMenuArray *marray, struct Menu *m)
```
flatcap, line 226: Add Doxygen comment block

flatcap (changes requested):
Lots of requests, all of them minor.
Building your changes generated a lot of warnings!

Besides the comments I've made, I'd like you to do some renaming.

When you moved the Alias array into a separate struct you didn't change the variables' names to match, so now they're a bit confusing.

I suggest renaming:
`AliasMenuArray` to `AliasArray` (nothing Menu-y about it)

In `struct AliasMenuData` rename:
`struct AliasMenuArray marray` to `struct AliasArray alias_arr`

Elsewhere, rename all `struct AliasArray` parameters and variables to `aa`.

`mdata` should only be used to refer to `struct AliasMenuData`

Thanks.

flatcap (changes requested):
Bugs:

- The Limit display in the status bar should show what the user typed in
  (rather than the generated pattern)
- When Limiting for a second time:
  - The prompt should show the existing limit string
  - If new text is typed in, this should create a different limit, not add to the existing one
- ~Adding / Removing Aliases (`:alias`, `:unalias`) causes a crash~

alias/gui.c
RomeuG: Unfortunately, adding that checks makes it impossible to do any kind of sorting without any active limit.

flatcap: Hmm... good point.
If we set `is_visible = true` in `menu_data_alias_add()` that should help.

RomeuG:
Did some work today, here is what I still have to do:
- Fix Alias/Query menu limit string
- Following prompts should show the existing limit string
- Discuss about the additional invisibility check (code review)

Also, I did not understand this:

`The Limit display in the status bar should show what the user typed in (rather than the generated pattern)`

If I type `~c test`, shouldn't the display show something along the lines of `Limit: ~c test` ?

flatcap:
Nice work!  That's much tidier.

> If I type `~c test`, shouldn't the display show something along the lines of `Limit: ~c test` ?

Yes, that case is OK, but if I limit to `john`, I see `Limit: ~f "john" | ~s "john"`
I'd expect to see just `Limit: "john"` (That's how `%V` in `$status_format` works)

pattern/pattern.c
```diff
@@ -223,6 +223,112 @@ bool mutt_limit_current_thread(struct Email *e)
...
+  mdata->str = strdup(simple);
+
+  m->max = vcounter;
+  m->current = 0;
+
+  char *tmp_str = NULL;
+  char *new_title = NULL;
+
+  mutt_str_asprintf(&tmp_str, _("Limit: %s"), simple);
+  mutt_str_asprintf(&new_title, "Aliases - %s", tmp_str);
```
flatcap, line 311: If `match_all` is set, we don't want this

flatcap (changes requested):
Looking good (apart from the crash :-)

If you have Doxygen installed, please can you run it -- it'll show where the function comments don't match:

`doxygen doxygen/doxygen.conf`

RomeuG:
Done!

Pretty ashamed about that crash though! :laugh:

flatcap:
Looking good :+1:

There's one last thing that'd be _nice to have_ (if it's not too much work).

If you enter the Alias dialog with a completion: <kbd>john</kbd> + <kbd>\<tab\></kbd>, would it be possible to make this a _limit_?
(so all the other Aliases would be there, just hidden)

The initial view shows all the `john`s, but hitting `<limit>.<enter>` would show all the Aliases.

How much work would this be?

RomeuG:
I don't think it is too much work. It is surely feasible.

But the thing is, I have to use `mutt_pattern_alias_func()` inside `alias_complete()`.
The problem is that `mutt_pattern_alias_func()` needs `Menu` as a parameter.

I think there are two ways of doing this:
- If you think it is better to continue using `mutt_pattern_alias_func()` in this situation, I guess it should take an AliasMenuData instead of AliasArray. Since `Menu` will be NULL, then we just have to check it to see if we're going to set the title or not.
- We create a new function similar to `mutt_pattern_alias_func()` just to accomodate these needs.

I prefer option one. What do you think, @flatcap ?

flatcap:
> I prefer option one

Yes, me too.  Please can you do that.

RomeuG:
@flatcap 

Just did the change. What do you think of it?

flatcap:
I hate to keep making changes, but you're **so** close now.
On entry, the two limits don't quite line up.

- <kbd>jo</kbd><kbd>\<tab\></kbd>
- `Limit: jo`

I think we should change the limit string to `~f ^jo` so that `<limit><enter>` doesn't change the results.

- <kbd>grammy</kbd><kbd>\<tab\></kbd>
- `Limit: grammy`

In this case, there are zero matching Aliases.
We show a limit of `grammy`, but **all** Aliases are displayed.
I think we should apply the limit here.

This would mean that auto-completion of the Alias is unchanged (and predictable),
but that we can quickly search for arbitrary strings.

alias/dlgalias.c
```diff
@@ -198,20 +207,36 @@ static void dlg_select_alias(char *buf, size_t buflen, struct AliasMenuData *mda
...
+    mutt_str_asprintf(&new_title, _("Aliases - %s"), tmp_str);
+
+    FREE(&tmp_str);
+
+    menu->title = new_title;
+  }
+  else
+  {
+    menu->title = strdup(_("Aliases"));
+  }
```
flatcap, line 228: This block of code is repeated in `pattern/pattern.c`
I suggest a adding small function taking `(AliasMenuData *mdata, char *limit)` where `limit` could be `NULL`.

Also, the string `"Aliases"` is already translated, but `"Aliases - %s"` isn't.
If we can use the first one _creatively_, then we save the translators a job.

flatcap:
I've rebase onto `master` and added a few small commits to tidy up.

RomeuG:
Ok, I will check your commits and re-evaluate what I did wrong.
Thanks!

flatcap:
A few final changes.

I renamed `AliasArray` to `AliasViewArray` (and `aa` to `ava`).
I wasn't happy with the name I suggested before.

I've renamed four `menu_*()` functions that actually take an `AliasViewArray` now.

flatcap (approved):

Issue #1949, opened 2019-11-07
When opening an email with an attachment, it downloads it twice


thomergil:
## Expected Behaviour

Open an email with an attachment. I expect neomutt to download the attachment once, then display the email.

## Actual Behaviour

After starting neomutt, open an email with an attachment. Notice that progress ("Fetching message...") goes from 0% to 100% twice--slowly for large attachments. After the first time, it only downloads the attachment once. Close neomutt, start it again, open the same email with the attachment, it downloads it twice again.

## Steps to Reproduce

See above.

## How often does this happen?

- Always

## When did it start to happen?

- When I upgraded to 20191102

## NeoMutt Version

```
NeoMutt 20191102-2513-1d5e12
Copyright (C) 1996-2016 Michael R. Elkins and others.
NeoMutt comes with ABSOLUTELY NO WARRANTY; for details type 'neomutt -vv'.
NeoMutt is free software, and you are welcome to redistribute it
under certain conditions; type 'neomutt -vv' for details.

System: Darwin 18.7.0 (x86_64)
ncurses: ncurses 5.7.20081102 (compiled with 5.7.20081102)
libiconv: 1.11
libidn: 1.35 (compiled with 1.35)
hcache backends: tokyocabinet, lmdb

Compiler:
Apple clang version 11.0.0 (clang-1100.0.33.12)
Target: x86_64-apple-darwin18.7.0
Thread model: posix
InstalledDir: /Applications/Xcode.app/Contents/Developer/Toolchains/XcodeDefault.xctoolchain/usr/bin

Configure options: --gss --lmdb --notmuch --sasl --tokyocabinet --with-ui=ncurses --disable-doc --disable-nls --with-ssl=/usr/local/opt/openssl@1.1 --prefix=/Users/macuser/install/

Compilation CFLAGS: -g -O2 -std=c99 -D_ALL_SOURCE=1 -D_GNU_SOURCE=1 -D__EXTENSIONS__ -I/Users/macuser/install//include -DNCURSES_WIDECHAR -I/usr/local/opt/openssl@1.1/include

Default options:
  +attach_headers_color +compose_to_sender +compress +cond_date +debug 
  +encrypt_to_self +forgotten_attachments +forwref +ifdef +imap +index_color 
  +initials +limit_current_thread +multiple_fcc +nested_if +new_mail +nntp +pop 
  +progress +quasi_delete +regcomp +reply_with_xorig +sensible_browser +sidebar 
[...]

PR #2661, opened 2020-09-17
Show error when message_cachedir is invalid

louis77:
**What does this PR do?**

- If message_cachedir is not a valid directory, create a temporary directory
- Prevents remote backends (imap, pop, nntp) to repeatedly fetch messages during the same session

A _suggestion_ to ease symptoms described in #1949.


gahr:
This is a config error and should in my opinion be handled when the config is parsed: how about a validator that checks if the path exists and it's a directory?

I don't like to create random directories under tmp (at least we should use `C_Tmpdir`) that are left behind when NeoMutt quits.

gahr (approved):

PR #2662, opened 2020-09-18
build: run shellcheck on hcachever.sh

flatcap:
Tidy the `hcachever.sh` build script.
The output is unchanged.

- Replace ` `` `s with `$()`
- Tidy `while`, `if`, `case` layout
- Quote variables to prevent globbing
- Add `-r` to `read`

~One check has been disabled: SC2181.~
~It allows us to check `$?` separately from the command.~

  ~https://github.com/koalaman/shellcheck/wiki/SC2181~

### DIFF
--- a/hcache/hcachever.sh
+++ b/hcache/hcachever.sh
@@ -20,12 +20,10 @@ getstruct () {
     *';') return ;;
   esac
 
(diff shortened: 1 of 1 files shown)


gahr (approved):
Thanks!

Issue #2665, opened 2020-09-22
Add a Contract to the Mailbox API


flatcap:
## MXAPI (Mailbox API)

The MXAPI (Mailbox API) is a set of functions that each backend (Imap, Maildir, etc) supports.
They're documented in `struct MxOps`:

https://github.com/neomutt/neomutt/blob/d9cb4825bec1c833229d51acc9a0976e6c31ca7e/mx.h#L104-L291

The frontend calls wrapper functions in `mx.c` that call the relevant backend function. e.g.

The frontend:
https://github.com/neomutt/neomutt/blob/d9cb4825bec1c833229d51acc9a0976e6c31ca7e/index.c#L753

calls:
https://github.com/neomutt/neomutt/blob/d9cb4825bec1c833229d51acc9a0976e6c31ca7e/mx.c#L400

ending up in the backend
https://github.com/neomutt/neomutt/blob/d9cb4825bec1c833229d51acc9a0976e6c31ca7e/imap/imap.c#L1909-L1912

## Contract

Each function is documented, but I'd like to go further and add a Contract to the docs.

In most functions, it's important to check pointer parameters before using them.
For these API functions, we know that they'll only be called by the wrappers in `mx.c`
We can write a Contract in the docs declaring that certain conditions will be met.

Here's one from `config/set.h`
https://github.com/neomutt/neomutt/blob/d9cb4825bec1c833229d51acc9a0976e6c31ca7e/config/set.h#L127-L141

When `native_set()` is called, we guarantee that `cs`, `var` and `cdef` will not be NULL.

In my [devel/path](https://github.com/neomutt/neomutt/tree/devel/path) branch, you can see a more complex example:

https://github.com/neomutt/neomutt/blob/5f0f022752aa895a8d40ccbb8c586831d3b4a49d/mx.h#L306-L322

## Research

There are 19 functions in `MxOps`:

- Create a list of the API functions in `MxOps`
- For each, find where it's called from (in `mx.c`).

There are 9 backends implementing `MxOps`:

- Create a list of all the API users (Imap, etc)
- For each, list which functions are called

## Tasks

- For each caller, check that the pointers cannot be `NULL`
- Add a Contract for each pointer
- For each backend function, remove any unnecessary checks

brunnre8:
Hi Rich, please assign me to it.

flatcap:
Thanks, that looks good -- It'll take me a little while to review it.

The next challenge is to add to the MXAPI.

[...]

PR #2673, opened 2020-09-27
Add a Contract to the Mailbox API

brunnre8:
Implements #2665 

For some of the functions I've added the guard all the backends were doing anyhow at the api call site so that we can enforce a stricter contract.

### DIFF
--- a/mx.c
+++ b/mx.c
@@ -1157,10 +1157,8 @@ int mx_mbox_check(struct Mailbox *m)
  */
 struct Message *mx_msg_open(struct Mailbox *m, int msgno)
 {
-  if (!m)
+  if (!m || !m->emails || (msgno < 0) || (msgno >= m->msg_count))
     return NULL;
-
-  struct Message *msg = NULL;
 
   if (!m->mx_ops || !m->mx_ops->msg_open)
   {
@@ -1168,7 +1166,7 @@ struct Message *mx_msg_open(struct Mailbox *m, int msgno)
     return NULL;
   }
 
-  msg = mutt_mem_calloc(1, sizeof(struct Message));
+  struct Message *msg = mutt_mem_calloc(1, sizeof(struct Message));
   if (m->mx_ops->msg_open(m, msg, msgno) < 0)
     FREE(&msg);
 
@@ -1184,7 +1182,7 @@ struct Message *mx_msg_open(struct Mailbox *m, int msgno)
  */
 int mx_msg_commit(struct Mailbox *m, struct Message *msg)
 {
-  if (!m || !m->mx_ops || !m->mx_ops->msg_commit)
+  if (!m || !m->mx_ops || !m->mx_ops->msg_commit || !msg)
     return -1;
 
   if (!(msg->write && m->append))
@@ -1305,7 +1303,7 @@ int mx_check_empty(const char *path)
  */
 int mx_tags_edit(struct Mailbox *m, const char *tags, char *buf, size_t buflen)
 {
-  if (!m)
+  if (!m || !buf)
     return -1;
 
   if (m->mx_ops->tags_edit)
@@ -1325,7 +1323,7 @@ int mx_tags_edit(struct Mailbox *m, const char *tags, char *buf, size_t buflen)
  */
 int mx_tags_commit(struct Mailbox *m, struct Email *e, char *tags)
 {
-  if (!m)
+  if (!m || !e || !tags)
     return -1;
 
   if (m->mx_ops->tags_commit)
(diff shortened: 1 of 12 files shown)


brunnre8:
after review it might make sense to squash them... but the choice is of course up to you

flatcap:
The separate commits has definitely simplified the review process

> after review it might make sense to squash them

Yes, I probably will

mx.h
```diff
@@ -212,13 +250,20 @@ struct MxOps
    * @param msg Message to close
    * @retval  0 Success
    * @retval -1 Failure
+   *
+   * **Contract**
+   * - @a m   is not NULL
+   * - @a msg is not NULL
```
flatcap, line 256: Add `*msg` too

mx.h
```diff
@@ -184,6 +210,10 @@ struct MxOps
    * @param msgno Index of message to open
    * @retval  0 Success
    * @retval -1 Error
+   *
+   * **Contract**
+   * - @a m   is not NULL
+   * - @a msg is not NULL
```
flatcap, line 216: What about `msgno`?
Perhaps: `0 <= msgno < m->msg_count`

mx.h
```diff
@@ -212,13 +250,20 @@ struct MxOps
...
 
   /**
    * msg_padding_size - Bytes of padding between messages
    * @param m Mailbox
    * @retval num Bytes of padding
+   *
+   * **Contract**
+   * - @a m is not NULL
    */
   int (*msg_padding_size)(struct Mailbox *m);
```
flatcap, line 269: No contract for `msg_save_hcache()`, below

mx.h
```diff
@@ -250,6 +298,9 @@ struct MxOps
    * @param buf Buffer containing tags
    * @retval  0 Success
    * @retval -1 Failure
+   *
+   * **Contract**
+   * - @a m is not NULL
    */
```
flatcap, line 304: What about `e` and `buf`?
Also, we could check that `buf` isn't empty (`buf[0] != '\0')
See check in `nm_tags_commit()`

mx.h
```diff
@@ -240,6 +285,9 @@ struct MxOps
    * @retval -1 Error
    * @retval  0 No valid user input
    * @retval  1 Buf set
+   *
+   * **Contract**
+   * - @a m is not NULL
    */
```
flatcap, line 291: What about `buf`?

flatcap (changes requested):
I've found a few things to fix.
Some I've listed here, because GitHub won't allow me to review code that you haven't changed :+1: 

Kill some checks:
https://github.com/brunnre8/neomutt/blob/mxops_pr/mbox/mbox.c#L885-L886

and these lines in a comment:
https://github.com/brunnre8/neomutt/blob/mxops_pr/nntp/nntp.c#L2379-L2381

delete the first check:
https://github.com/brunnre8/neomutt/blob/mxops_pr/compmbox/compress.c#L825

mx.h
brunnre8: hm, what's the point there? Can a valid pointer actually point to struct Message which is Null?
We aren't passing a double pointer to struct Message, but only a struct Message*

mx.h
brunnre8: I generally only added contract for existing invariants, I didn't add more checks (minus the two cases where that was the only sensible input)

I can do that, but then we should probably fail in mx.c if the invariant isn't met right?

mx.h
brunnre8: yes, as the function as is didn't have any guarantees.
It only checks if it has the mx_ops and then calls the worker function.
in fact, the only thing guaranteed right now is that the mailbox isn't null and only because the application would blow up if it was ;)

Shall I add the contract (and the guards)?

mx.h
brunnre8: handled the !e and !buf cases.... but we can't use an empty buffer as a failure. The imap worker uses the absence of any tags to clear all flags (iiuc).

As in, it doesn't have notmuch +/-tag semantics but rather just replaces all tags with whatever the user was typing, including ""

Or do I misunderstand how this works?

flatcap: you're right, we don't want the (buf[0] != '\0') check

mx.h
flatcap: Sorry, ignore that.
We're only **closing** the Message handle, not freeing it.

brunnre8:
that should address all the review comments I think, please have a look (at your leisure naturally)

mx.c
```diff
@@ -1168,6 +1168,11 @@ struct Message *mx_msg_open(struct Mailbox *m, int msgno)
     return NULL;
   }
 
+  if (!(0 <= msgno < m->msg_count))
```
flatcap, line 1171: Ah, I _did_ suggest that, but it's not valid C.
You'll have check the two conditions separately.

brunnre8: sorry... habit from other languages where this is perfectly valid code...

PR #2682, opened 2020-10-03
build: improve ASAN flags

flatcap:
Change the build flags:

- Replace `-O2` with `-O0` (turn off optimisations)
- Add `-fno-omit-frame-pointer -fno-common`
  as suggested here, https://github.com/google/sanitizers/wiki/AddressSanitizerFlags

### DIFF
--- a/auto.def
+++ b/auto.def
@@ -1214,7 +1214,7 @@ if {[get-define want-asan]} {
     user-error "Unable to compile with -fsanitize=address"
   }
   msg-result "yes"
-  define-append CFLAGS -fsanitize=address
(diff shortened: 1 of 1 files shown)


gahr (commented):
Clever, perhaps a bit too much? How about just append -O0 so it doesn’t break when/if -O2 is not there?

flatcap:
> Clever, perhaps a bit too much?

Yeah, you're right.
We default to adding `-O2`, but I didn't think of the user overriding them: `CFLAGS="-O3" ./configure`

> How about just append -O0

Done

gahr (approved):
Thanks :)

PR #2683, opened 2020-10-04
Move mx_check_empty into mx_ops

brunnre8:
Move `mx_check_empty()` into the mx_ops and rename to `mx_path_is_empty`.
The actual business logic is moved to the backends.

### DIFF
--- a/mbox/mbox.c
+++ b/mbox/mbox.c
@@ -1721,6 +1721,14 @@ static int mbox_path_parent(char *buf, size_t buflen)
 }
 
 /**
+ * mbox_path_is_empty - Is the mailbox empty - Implements MxOps::path_is_empty()
+ */
+static int mbox_path_is_empty(const char *path)
+{
+  return mutt_file_check_empty(path);
+}
+
+/**
  * mmdf_msg_commit - Save changes to an email - Implements MxOps::msg_commit()
  */
 static int mmdf_msg_commit(struct Mailbox *m, struct Message *msg)
@@ -1827,6 +1835,7 @@ struct MxOps MxMboxOps = {
   .path_canon       = mbox_path_canon,
   .path_pretty      = mbox_path_pretty,
   .path_parent      = mbox_path_parent,
+  .path_is_empty    = mbox_path_is_empty,
 };
 
 /**
@@ -1856,5 +1865,6 @@ struct MxOps MxMmdfOps = {
   .path_canon       = mbox_path_canon,
   .path_pretty      = mbox_path_pretty,
   .path_parent      = mbox_path_parent,
+  .path_is_empty    = mbox_path_is_empty,
 };
 // clang-format on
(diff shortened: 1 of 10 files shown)


brunnre8:
sorry, messed up the rebase, should be ready for review now

flatcap:
Heads up: I'm looking at some more MXAPI enhancements (once this is merged).
Currently the browser allows the use to create, rename and delete Mailboxes (but only IMAP).

1. Add new API functions and fit them into the browser
2. Implement the functions for the Maildir and Mbox backends

If you're interested, start looking at the existing functions and think about how the API functions should work.

mbox/mbox.c
```diff
@@ -1866,6 +1866,7 @@ struct MxOps MxMboxOps = {
   .path_canon       = mbox_path_canon,
   .path_pretty      = mbox_path_pretty,
   .path_parent      = mbox_path_parent,
+  .path_is_empty    = mutt_file_check_empty,
```
flatcap, line 1869: This will sound inefficient, but can you create an `mbox_path_is_empty()` function to call.

The APIs have an arrays of function pointers.
When debugging them, it's useful if the functions have unique names,
so in `mx_path_is_empty()` the pointer `ops->path_is_empty` is mbox-specific.

brunnre8:
want me to duplicate the function body? Or would you rather have library function called from both?

flatcap:
One function called by both, please.

mbox/mbox.c
brunnre8: want me to duplicate the function body? Or would you rather have library function called from both?

rewrote here... replying via mail breaks the threading apparently

brunnre8: @flatcap so... this is an issue over almost all the MxOps structs... not only in that field.

it's pretty prominent in mh vs maildir... they both share quite a lot of code and have random methods of the other one in their ops.
In fact, the only backends that consistently have that are compress (not that this is hard...) / imap and nntp

Do you really want me to wrap those all in function calls?

brunnre8: MMDF will be especially painfull... it shares 90% of the code with mbox

flatcap: Hang on!
I just wanted a single function wrapper around the `mutt_file_check_empty()` call.
Nothing else.

So that all the functions in MxMboxOps and MxMmdfOps have names beginning either `mbox_` or `mmdf_`.

brunnre8: yes. However most other backends have also methods from other backends embedded. The obly places that have this consistently is imap, nntp and  compress.

If your issue is debugging you still have cases where say the notmuch backend does stuff but the function pointer is called maildirsdomething, which was your original complaint no?

flatcap: > which was your original complaint no?

Yes, but one thing at a time :-)
I'll get to the other backends eventually.

flatcap (approved):

PR #2688, opened 2020-10-05
update/correct the sasl authentication methods

flatcap:
Some of the methods we cut'n'pasted into `sasl_authenticators[]` weren't meaningful.
_mea culpa_

@grawity
Sorry it's taken a while to get to.
Please can you let me know what you think.

Thanks.

**References**:
- [cyrus sasl](https://www.cyrusimap.org/sasl/sasl/authentication_mechanisms.html)
- [wikipedia](https://en.wikipedia.org/wiki/Simple_Authentication_and_Security_Layer)

### DIFF
--- a/conn/sasl.c
+++ b/conn/sasl.c
@@ -100,9 +100,27 @@ struct SaslSockData
  * sasl_authenticators - Authenticaion methods supported by Cyrus SASL
  */
 static const char *const sasl_authenticators[] = {
-  "ANONYMOUS", "CRAM-MD5",   "DIGEST-MD5",  "EXTERNAL", "G2",
-  "GSSAPI",    "GSS-SPNEGO", "KERBEROS_V4", "LOGIN",    "NTLM",
-  "OTP",       "PASSDSS",    "PLAIN",       "SCRAM",    "SRP"
+  "ANONYMOUS",
+  "CRAM-MD5",
+  "DIGEST-MD5",
+  "EXTERNAL",
+  "GS2-IAKERB",
+  "GS2-KRB5",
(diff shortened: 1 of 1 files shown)


grawity:
Looks good to me. Though the `-PLUS` variants require channel bindings (i.e. giving libsasl the TLS session information) – I think neomutt currently doesn't do that, so in that case the -PLUS mechanisms shouldn't be in the list.

(Not 100% sure of the purpose of the validation in general. Is it mainly to be able to show an error message without having to invoke libsasl?)

flatcap:
> -PLUS mechanisms shouldn't be in the list.

done

> Not 100% sure of the purpose of the validation in general

The idea is to validate the config as early as possible.
When config variables are set, their value is checked to see if it's valid for their type.

Some config variables have a _validator_ too, which can add stricter rules,
e.g. `$wrap_headers` (a number) which must be between 78 and 998.

It was a nice idea, but it's getting a bit complicated, now :-)

Neustradamus:
Thanks to @yousefakbar, @flatcap, @grawity about SCRAM:
- https://github.com/neomutt/neomutt/pull/2624 + https://github.com/neomutt/neomutt/commit/b7bcb9aef103943251305c05840cbe96c09a6508
- https://github.com/neomutt/neomutt/pull/2688 + https://github.com/neomutt/neomutt/commit/85437597de9e692fa8c0aeab7eaaf27e43c563c6

Linked to:
- https://github.com/neomutt/neomutt/issues/1846
- https://gitlab.com/muttmua/mutt/-/issues/174

PR #2695, opened 2020-10-10
libcompose remove config variables

matthewhughes934:
* **What does this PR do?**

Removes global 'C_*' config variables from libcompose, replacing them with access to the corresponding ConfigSubset element.

#2622


matthewhughes934:
Still want to do some deeper testing, I've manually test each added `cs_subset*` at least runs (i.e. I avoided any heinous typos), but still want to double check variables are respected (e.g. change the value of each). The only thing I missed (so far) in my testing is at `compose/compose.c:543` because I haven't played around enough with Newsgroups in Neomutt.

flatcap:
I've just added some `{}`s and moved a comment to keep the compiler happy.

flatcap (changes requested):
Now we can eliminate the `C_*` variables completely (from compose),
and update the flags to `cs_register_variables()`
See: 6f79d959b2 for details.

flatcap (approved):

Issue #2692, opened 2020-10-08
Entries in query_command result list are sorted by name


ftilde:
## Expected Behaviour

Entries in the query_command menu should be displayed in the order that the query_command produced them. This is useful because the query command may have more information about the desired order the addresses should appear in. For example, in [maq](https://github.com/ftilde/maq) email addresses are sorted by how frequently they appear in the searched mailboxes. If the user *does* want to sort by display name, they can just pipe their command through `sort`.

## Actual Behaviour

Entries in the query_command menu are always sorted by the display name.

## Steps to Reproduce

In your config:
```
set query_command = "echo '\n foo@bar.com\tfoo \n foo@bar.com\tbar'"
```
Try to complete an address and notice that "bar" is listed before "foo" even though "foo" was emitted first.

## How often does this happen?

- Always

## When did it start to happen?

0c5851047af0b45408a78f9ce8408672f414bccf appears to be the culprit.

## NeoMutt Version

All versions since 0c5851047af0b45408a78f9ce8408672f414bccf

flatcap:
Yes.  In modernising the Queries, they now use the same config.
```
set sort_alias = "unsorted"
```
Will leave your addresses in the order you want.

If there's the demand for it, we _could_ add a `$sort_query`, but I'd prefer not to add more config.

ftilde:
On 08.10.2020 14:35, Richard Russon wrote:
>```
>set sort_alias = "unsorted"
>```
I see! That does indeed resolve my immediate problem.

>If there's the demand for it, we _could_ add a `$sort_query`, but I'd 
>prefer not to add more config.
I don't think that there is a need for a separate config item. However, 
I still feel (although not too strongly) that the results of the 
query_command should just not be sorted (as this can just be done in the 
query_command itself if desired).

Feel free to disagree, though! In that case I think we should update the 
documentation of `sort_alias` and also `query_command`.

I would be willing to implement either solution.

flatcap:
> I still feel that the results of the query_command should just not be sorted

OK, noted.

There is another possibility, to **not** sort either by default.
The user now has the ability to sort/search/limit the results,
and the default behaviour for `name<tab>` is smarter to limit the matches.

I'll leave changing the defaults for now and I'll mark this issue as "discuss", to see if anyone has any other ideas / strong views :-)

In the meantime, if you'd like to improve the docs, that'd be greatly appreciated.
The two relevant sections are:

[...]

PR #2698, opened 2020-10-12
Document that $sort_alias affects the query menu

ftilde:
* **What does this PR do?**
Document (in manual and man page) the fact that the alias and query menu are now identical and that consequently `$sort_alias` also affects the query menu.

#2692


flatcap (approved):
I just tweaked your change to put it in `<note>` tags.
Thanks

flatcap:
btw, which is your preferred address: "tamepointer" or "protonmail"?

ftilde:
On 13.10.2020 05:02, Richard Russon wrote:
>I just tweaked your change to put it in `<note>` tags.
That makes sense. Thanks!

ftilde:
On 13.10.2020 05:08, Richard Russon wrote:
>btw, which is your preferred address: "tamepointer" or "protonmail"?
tamepointer. I've recently(ish) moved away from protonmail because of 
the annoying smpt/imap situation.

PR #2706, opened 2020-10-16
feat(notmuch): show originating folder in index

### LINKED ISSUE #1849: Show originating/source mail folder in vfolder-from-query?
Hi. I searched around but couldn't find an answer to this. 

Is there a way to show the source mail folder (maildir in my case) for each message in the index of a  vfolder built from a notmuch query or in the pager when viewing a message in a vfolder?

Austin-Ray:
Changes the behavior of '%b' in 'index_format' to display the email's
folder instead of the notmuch query if it exists. In a notmuch mailbox,
the current behavior of index_format's %b' expando displays the query
portion of a notmuch mailbox declaration.

The reference documentation for 'index_format' says '%b' shows "Filename
of the original message folder (think mailbox)". By my interpretation,
displaying the notmuch query is incorrect as it _is not_ the original
message folder. The underlying maildir folder _is_ the original message
folder. Therefore it should be shown instead.

[0] https://neomutt.org/guide/reference.html#3-169-%C2%A0index_format

* **Some decision considerations**

1. This brings the `%b` behavior in line with other local mailboxes; however, it only displays the last part of the folder. So if there's emails from `email1@email.com/INBOX` and `email2@email.com/INBOX` then it will only display `INBOX`. Is this something we want to changes? How far back do we go?

Closes #1849

### DIFF
--- a/hdrline.c
+++ b/hdrline.c
@@ -56,6 +56,9 @@
 #include "mutt_thread.h"
 #include "muttlib.h"
 #include "sort.h"
+#ifdef USE_NOTMUCH
+#include "notmuch/lib.h"
+#endif
 
 /* These Config Variables are only used in hdrline.c */
 struct MbTable *C_CryptChars; ///< Config: User-configurable crypto flags: signed, encrypted etc.
@@ -540,6 +543,15 @@ static const char *index_format_str(char *buf, size_t buflen, size_t col, int co
       if (m)
       {
         p = strrchr(mailbox_path(m), '/');
+#ifdef USE_NOTMUCH
+        if (m->type == MUTT_NOTMUCH)
+        {
+          char *rel_path = nm_email_get_folder_rel_db(m, e);
+          if (rel_path)
+            p = rel_path;
+        }
+#endif
+
         if (p)
           mutt_str_copy(buf, p + 1, buflen);
         else
--- a/notmuch/lib.h
+++ b/notmuch/lib.h
@@ -58,6 +58,7 @@ void  nm_db_debug_check          (struct Mailbox *m);
 void  nm_db_longrun_done         (struct Mailbox *m);
 void  nm_db_longrun_init         (struct Mailbox *m, bool writable);
 char *nm_email_get_folder        (struct Email *e);
+char *nm_email_get_folder_rel_db (struct Mailbox *m, struct Email *e);
 int   nm_get_all_tags            (struct Mailbox *m, char **tag_list, int *tag_count);
 bool  nm_message_is_still_queried(struct Mailbox *m, struct Email *e);
 void  nm_parse_type_from_query   (struct NmMboxData *mdata, char *buf);
--- a/notmuch/notmuch.c
+++ b/notmuch/notmuch.c
@@ -1676,6 +1676,29 @@ char *nm_email_get_folder(struct Email *e)
 }
 
 /**
+ * nm_email_get_folder_rel_db - Get the folder for a Email from the same level as the notmuch database
+ * @param m Mailbox containing Email
+ * @param e Email
+ * @retval ptr  Folder containing email from the same level as the notmuch db
+ * @retval NULL Error
+ *
+ * Instead of returning a path like /var/mail/account/Inbox, this returns
+ * account/Inbox. If wanting the full path, use nm_email_get_folder().
+ */
+char *nm_email_get_folder_rel_db(struct Mailbox *m, struct Email *e)
+{
+  char *full_folder = nm_email_get_folder(e);
+  if (!full_folder)
+    return NULL;
+
+  const char *db_path = nm_db_get_filename(m);
+  if (!db_path)
+    return NULL;
+
+  return full_folder + strlen(db_path);
+}
+
+/**
  * nm_read_entire_thread - Get the entire thread of an email
  * @param m Mailbox
  * @param e   Email


flatcap:
> How far back do we go?

Do we have access to the Notmuch config, in particular:
```ini
[database]
path=/home/user/mail
```

We could use that as the base.

Failing that, is the Notmuch database always in the root dir of the mail?

jindraj:
> Do we have access to the Notmuch config, in particular:
> 
> ```ini
> [database]
> path=/home/user/mail
> ```

yes, `nm_default_url` in neomuttrc should match `database.path` in notmuch-config

----

There's one edge case I can think about:
But what about multiple same messages indexed?

e.g.: offlineimap users syncing *INBOX*, *archive* , *some-mailbox* - all mailboxes having the same message.
maildirs: **~/mail/{INBOX,archive,some-mailbox}/{cur,new,tmp}**
having `database.path` set to **~/mail/** would lead to notmuch record with file index in all 3 mailboxes.

brunnre8:
I'm not really fond of this... I think having the notmuch query displayed is actually what you want.
I manage multiple accounts in one notmuch DB, the original underlying maildir store is actually not used at all and has no utility, considering that you wouldn't be moving messages between maildirs anyhow but just tagging them differently to put them in a different "virtual mailbox".

The "folder" the message is in is the virtual folder, so the current behavior seems correct to me.

flatcap:
> I think having the notmuch query displayed is actually what you want.

Is it?  This is the Index we're talking about.
Knowing which folder an Email came from _might_ be useful.

Surely the 'notmuch query' will be the same for every email (effectively the Mailbox name).
It'd make sense to display the query in the Status Bar.

(Note: **I know nothing about Notmuch**, so feel free to shoot down anything I say :-)

jindraj:
> Knowing which folder an Email came from _might_ be useful.

There are people asking for this:
 - [How can I find out/display the mailbox that a particular message is in with neomutt+notmuch?](https://unix.stackexchange.com/q/363929/85046)
 - [In notmuch search results, how to find which maildir contains each of the emails?](https://stackoverflow.com/questions/55042384/in-notmuch-search-results-how-to-find-which-maildir-contains-each-of-the-emails)

brunnre8:
>Surely the 'notmuch query' will be the same for every email

The same can be said if you have the maildir worker (backend?) active, it'll show a constant folder as you can only have a single dir open at the time or am I wrong?

In my view it's ok if backends change their behavior a bit depending on what makes sense in the view of what you are doing.

As jindraj said, you can have a single entry in the notmuch DB which points to multiple files (it'll simply merge the headers etc, say if you have a copy in sent and one copy from the mailing list if you have that option active).

I'm not sure what the value is in knowing one of the files (you can't show all of them).  
From a use case perspective what would you want to do with it?

Normally you don't use the underlying maildir at all, only to sync mails.
Once you start using the tags that's pretty much the only organization you can do.

brunnre8:
hm the people in the threads jindraj linked seem to use notmuch in a rather strange way... they only use it sporadically for searching it seems and want to actually have maildir semantics iiuc.

That's probably not what most people using notmuch do, at least not as far as I could tell from the mailing list of notmuch or their official website.
Generally all their docs assume that you use notmuch to organize the emails, not just search them... at least that's what I took away from it.

flatcap:
> > Surely the 'notmuch query' will be the same for every email

> The same can be said if you have the maildir worker (backend?) active,
> it'll show a constant folder as you can only have a single dir open at the time or am I wrong?

Ah, yes.
Hmm... Changing `%b` will also affect some other displays, as `index_format_str()` is used elsewhere.
See: `mutt_make_string_flags()`, `mutt_make_string_info()`.

jindraj:
> hm the people in the threads jindraj linked seem to use notmuch in a rather strange way... they only use it sporadically for searching it seems and want to actually have maildir semantics iiuc.
> Generally all their docs assume that you use notmuch to organize the emails, not just search them... at least that's what I took away from it.

I agree. These people are using notmuch differently than it's supposed to be used. Only for searching and probably not using tags to organize their messages. 

----
The edge case I talked about in my first comment:
I created another maildir with a duplicate message.

`notmuch search --output=files id:neomutt/neomutt/pull/2706/c710810027@github.com`
```
/Users/jindraj/.mail/gmail/mail/cur/17536285a0e93eb5:2,S
/Users/jindraj/.mail/gmail/duplicate/cur/17536285a0e93eb5:2,S
```
then `%b` expando displays only `duplicate` – the last file from output above.

----
The correct approach is to have a single notmuch database for multiple maildirs.
```
~/mail/account1/archive/{cur,new,tmp}
~/mail/account2/archive/{cur,new,tmp}
```
and notmuch-config with `database.path = ~/.mail`

The notmuch search would still produce output with two files belonging to the message, but in notmuch database it's only a single record, linking these two files + tags you've applied to the message. Example:
```
notmuch tag +account1 -- folder:account1
notmuch tag +account2 -- folder:account2
```
then `%g` expando displays both `account1` and `account2` + any other tag applied on the message.

Austin-Ray:
> These people are using notmuch differently than it's supposed to be used. Only for searching and probably not using tags to organize their messages.

Any discussion about _how_ someone should use `notmuch` is irrelevant. Notmuch bills itself as "a fast, global-search and tag-based email system" so if someone wants to use it solely for ad-hoc searching, that's their prerogative. It may suit their workflow better. There's too many assumptions that a pure `notmuch`  workflow is correct, and that our `notmuch` workflows are the correct ones.

---

The fact remains that the documentation describes a behavior that all other backends (to my knowledge) implement correctly. Either the documentation needs to be updated to reflect the difference or the behavior needs to be fixed. As evidence by the second StackExchange post linked by @jindraj, users have tried `%b` expecting the documented behavior. This inconsistency reflects poorly on us.

Here's what the documentation says:
> Filename of the original message folder (think mailbox)

The emphasis is on "original". Given `notmuch`'s many-to-many behavior, it's hard to see any notmuch mailbox as the original folder, especially for `vfolder-from-query`. If a single email can be loaded from two different notmuch queries then neither is the original. That loaded email **can** be traced to a single file therefore its parent folder **is** the original folder.

Now the many-to-many behavior is related to @jindraj 's edge-case, which is a user experience design consideration. That edge-case shows the internal workings of neomutt. The notmuch backend loads the last filename it sees for a message-id.

https://github.com/neomutt/neomutt/blob/26e48ff0f70aaadbd3606fd8a5cf747e87499338/notmuch/notmuch.c#L831-L842

So displaying only one filename is internally consistent with neomutt since that is the file neomutt used for the email. The copy code has the same behavior.

However, is it confusing to the users? And is it more confusing that inconsistent documentation?

---

There's also a lot of discussion about multiple accounts, but none about the maildir folders associated with the accounts. If I use this PR and run a `vfolder-from-query` with `tag:inbox` as my query and `%b` as my `index_format`, I see folders like `INBOX`, `archive`, `sent`, and `spam`. With the improvement to go based off the mail folder root, it would be `email1/INBOX`, `email2/INBOX`, `email1/archive`, etc. That seems pretty useful to me for ad-hoc searches.

It's even beneficial for a single account that may have dozens of folders (like a really old email account of mine.)

To tag each folder individually is very tedious.

Austin-Ray:
Cleaned up the git history. Please review.

Austin-Ray:
ping

flatcap (approved):

PR #2710, opened 2020-10-20
Change postpone mode to write Date header too

bket:
This will ensure that postponed mails are written with a Date header.
rfc5322 says the Date field must be there for an email to be valid.


gahr (approved):
Thanks!

PR #2725, opened 2020-11-06
Fix sidebar indentation issue for multiple accounts configuration

Dvbnhbq:
After v.20200806 sidebar indentation was broken for all mailboxes names
which are not relative to the path set by the `folder` parameter.

Only successfully abbreviated mailboxes names were indented correctly.
Here is how the sidebar was displayed in the configuration with multiple
accounts:

```
[=-account1-=]     <<------ `folder` parameter is pointing to this account
INBOX
  Newsletters
Sent
Drafts
          [=-account2-=]    <<----- other mailboxes are all extra indented
          INBOX                     to the depth value being calculated
            Newsletters             from the absolute path of the
          Sent                      mailbox, but not in relation to the
          Drafts                    base path of the account
          [=-account3-=]
          INBOX
            Newsletters
            ...
```

There is `sidebar_component_depth` parameter which decreases the depth
of the actual path of all mailboxes, so that all those common path parts
'/home/Mail/xxxxxxxxx/' could be omitted when calculating the
indentation needed, this worked like so:

```
  if (C_SidebarComponentDepth > 0)
    entry->depth -= C_SidebarComponentDepth;
```

But the problem was that, this depth correction was excessively applied
to the abbreviated mailboxes to which the `folder` parameter was
pointing, which is not needed, because all the depth had been removed by
`abbrev_folder` already.

```
[=-account1-=]     <<------ `folder` parameter is pointing to this account
INBOX                       { This line should be indented, but because
Newsletters    <<---------- { of decrease of depth set by
Sent                        { `sidebar_component_depth` all mailboxes
Drafts                      { are de-indented
[=-account2-=]
INBOX
  Newsletters      <<----- other mailboxes are indented correctly
Sent
Drafts
[=-account3-=]
INBOX
  Newsletters
...
```

So, it looks like, instead of
    ```
    if (C_SidebarFolderIndent && short_path_is_abbr)
      if (C_SidebarComponentDepth > 0)
        entry->depth -= C_SidebarComponentDepth;
    ```
it should do
    ```
    if (C_SidebarFolderIndent && !short_path_is_abbr)
      ...
    ```

Also, in this block:
```
    const char *abbr = abbrev_folder(path, C_Folder, m->type);
    ...
    const char *short_path = abbr ? abbr : path;
    ...
    const bool short_path_is_abbr = (short_path != path);
```
it looks like `abbr` is never `false`, because `abbrev_folder` always returns a string, either abbreviation or full path, so this ternary ? operator always returns `abbr` value. This needs a modification.

So, after those corrections, with `sidebar_component_depth` applied accordingly to the mailboxes
common path depth, the result looks like this:

```
[=-account1-=]     <<------ `folder` parameter is pointing to this account
INBOX
  Newsletters
Sent
Drafts
[=-account2-=]     <<----- other mailboxes are all extra indented
INBOX
  Newsletters
Sent
Drafts
[=-account3-=]
INBOX
  Newsletters
...
```

### DIFF
--- a/sidebar/window.c
+++ b/sidebar/window.c
@@ -693,7 +693,7 @@ int sb_recalc(struct MuttWindow *win)
     const char *last_part = abbr;
     entry->depth = calc_path_depth(abbr, C_SidebarDelimChars, &last_part);
 
-    const bool short_path_is_abbr = (short_path == abbr);
+    const bool short_path_is_abbr = (short_path != path);
     if (C_SidebarShortPath)
     {
       short_path = last_part;
@@ -701,7 +701,7 @@ int sb_recalc(struct MuttWindow *win)
 
     // Don't indent if we were unable to create an abbreviation.
     // Otherwise, the full path will be indent, and it looks unusual.
-    if (C_SidebarFolderIndent && short_path_is_abbr)
+    if (C_SidebarFolderIndent && !short_path_is_abbr)
     {
       if (C_SidebarComponentDepth > 0)
         entry->depth -= C_SidebarComponentDepth;


gahr:
This is a known problem with the current sidebar logic, but I don't think your change fixes it *in general* (although I can well imagine that it fixes it in *your particular use case*).

Here's why it's difficult to get it right. Given this setup:

```
set folder = imaps://user@server.com
mailboxes +Inbox
mailboxes +Lists
mailboxes +Lists.Neomutt
mailboxes /home/gahr/Mail/gmail/Inbox
mailboxes /home/gahr/Mail/gmail/[Gmail]/Important
```

what is the correct indentation of the /home/gahr mailboxes, and how do you figure it? Note that I might want to change their relative order, and indentation should not change:

```
mailboxes /home/gahr/Mail/gmail/[Gmail]/Important
mailboxes /home/gahr/Mail/gmail/Inbox
```

I say the shortened / indented view should be (assuming `sidebar_folder_indent=yes sidebar_short_path=no`

```
Inbox # the IMAP one
  Lists.NeoMutt
Inbox # the /home/gahr/Mail/gmail one
  [Gmail]/Important
```

I have been dodging this issue for a long time because I think fixing it would require an O(N²) algorithm that compare every mailbox with all others to figure what would be a "common path".

Dvbnhbq:
> This is a known problem with the current sidebar logic, but I don't think your change fixes it _in general_ (although I can well imagine that it fixes it in _your particular use case_).
> 
> Here's why it's difficult to get it right. Given this setup:
> 
> ```
> set folder = imaps://user@server.com
> mailboxes +Inbox
> mailboxes +Lists.Neomutt
> mailboxes /home/gahr/Mail/gmail/Inbox
> mailboxes /home/gahr/Mail/gmail/[Gmail]/Important
> ```
> 
> what is the correct indentation of the /home/gahr mailboxes, and how do you figure it? Note that I might want to change their relative order, and indentation should not change:
> 
> ```
> mailboxes /home/gahr/Mail/gmail/[Gmail]/Important
> mailboxes /home/gahr/Mail/gmail/Inbox
> ```
> 
> I say the shortened / indented view should be (assuming `sidebar_folder_indent=yes sidebar_short_path=no`
> 
> ```
> Inbox # the IMAP one
>   Lists.NeoMutt
> Inbox # the /home/gahr/Mail/gmail one
>   [Gmail]/Important
> ```
> 
> I have been dodging this issue for a long time because I think fixing it would require an O(N²) algorithm that compare every mailbox with all others to figure what would be a "common path".

Excuse me, if I am missing someting here, but as I mentioned in the commit description, should the `sidebar_component_depth` option solve this problem? What I do, I set `set sidebar_component_depth = 5` for the *common part* of the  absolute path of all mailboxex (considering they all have a common part at least `/home/gahr/Mail/<account name>/` followed by the mailbox name in each account). This should work with my fix, beause I have the same tree sturcture of my multiple account mailboxes. It just works.

What `sidebar_component_depth` do, is, it decreeses the indentation exactly by 5. This was written in your original code, but I just slightly tweaked the logic to fit exactly the tree you noticed here. I consider this setting as a way to say, how to identify the common part of the maibox path, if `folder` is unknown for the particular mailbox. As all mailboxes have identical path structure, we need to know olny this common depth number, not the common path itself. I would suggest this scheme as a temporary commonly usable workaround for the configuration with multiple accounts. Otherwise, more complex refactoring is needed to find the common path for multiple accounts.

gahr:
I don't think you should set `set sidebar_component_depth = N` where `N` is the number of slashes in your paths not relative to `folder`. That's not the way I see its purpose. I would like to see a solution that does the right thing.

gahr:
Also, `sidebar_component_depth` and `sidebar_short_path` (which gives you the abbreviated names) are orthogonal: I want to be be able to have:

```
# component_depth=0, short_path=no
Inbox
  Lists.NeoMutt
```

```
# component_depth=1, short_path=no
Inbox
Lists.NeoMutt
```

```
# component_depth=0, short_path=yes
Inbox
  NeoMutt
```

```
# component_depth=1, short_path=yes
Inbox
NeoMutt
```

Dvbnhbq:
All right, but this is exactly the thing, that this option does, based on code, and I don't see what else this option does, and is there another purpose except for saying what is the relative path offset.

And here is a thing, the `folder` value works for ONE account at a time ONLY,  as it stores the only value at a time. It just cannot be used for multiple accounts, so when it is set to one of the accounts, all other mailboxes are extra indented in the sidebar.

The problem is, if you want to use `folder-hook` settings to switch the settings between the accounts, then you cannot split this current `folder` parameter value for the other accounts, as it is not storing multiple of them at a time. This is what the problem is. 

I had a look at the code, trying to find a better thing to do, but as I am not the author, it is hard for me to refactor it significantly enough in the more appropriate way. Just wanted to fix this quickly in the evident and most painless way without major changes. Would be happy if you could find some time to do some better things.

gahr:
> And here is a thing, the folder value works for ONE account at a time ONLY, as it stores the only value at a time. 

Yep, that's what's broken now. I want indentation and shortening to work based on relative common paths of a subset of the mailboxes. But that requires proper thinking, which I can't do at the moment :(

Dvbnhbq:
> Also, `sidebar_component_depth` and `sidebar_short_path` (which gives you the abbreviated names) are orthogonal: I want to be be able to have:
> 
> ```
> # component_depth=0, short_path=no
> Inbox
>   Lists.NeoMutt
> ```
> 
> ```
> # component_depth=1, short_path=no
> Inbox
> Lists.NeoMutt
> ```
> 
> ```
> # component_depth=0, short_path=yes
> Inbox
>   NeoMutt
> ```
> 
> ```
> # component_depth=1, short_path=yes
> Inbox
> NeoMutt
> ```



> Also, `sidebar_component_depth` and `sidebar_short_path` (which gives you the abbreviated names) are orthogonal: I want to be be able to have:
> 
> ```
> # component_depth=0, short_path=no
> Inbox
>   Lists.NeoMutt
> ```
> 
> ```
> # component_depth=1, short_path=no
> Inbox
> Lists.NeoMutt
> ```
> 
> ```
> # component_depth=0, short_path=yes
> Inbox
>   NeoMutt
> ```
> 
> ```
> # component_depth=1, short_path=yes
> Inbox
> NeoMutt
> ```

This should work, but as you've mentioned above, you need to say, how to abbreviate mailbox from the absolute path and what is the offset... You have to find some way or a setting to set this up. Or you just need to find the common part in mailbox paths, which is slightly harder to implement.

In this case, you could `set sidebar_component_depth=5` or `6`, and this will produce the result requested.

Dvbnhbq:
> which I can't do at the moment :(

The same thing here, as some major refactoring and changes needed :)

PR #2731, opened 2020-11-22
imap/browse: avoid redundant function call

brunnre8:
Cleans up (in my view) a redundant str len call after copying a string.

Was there a particular reason why the size variable was a short instead of a size_t? Seems roundabout to me, even if the destination buffer size fits in a short as is.

### DIFF
--- a/imap/browse.c
+++ b/imap/browse.c
@@ -382,7 +382,6 @@ int imap_mailbox_create(const char *path)
   struct ImapAccountData *adata = NULL;
   struct ImapMboxData *mdata = NULL;
   char name[1024];
-  short n;
 
   if (imap_adata_find(path, &adata, &mdata) < 0)
   {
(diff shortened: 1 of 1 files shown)


Austin-Ray (commented):
Looks good to me.

The uses of `n` have bound checks so we won't have to worry about `size_t` being larger than `name`, and call removal doesn't appear to affect the execution logic (introduction commit https://github.com/neomutt/neomutt/commit/20f567b326613ed74b0481f144c74790ac4ac88b doesn't give a reason for extra call so it may have been implementation convenience.)

I'll let @gahr give the final approval in case I missed something as he's more familiar with the IMAP code.

> Was there a particular reason why the size variable was a short instead of a size_t? Seems roundabout to me, even if the destination buffer size fits in a short as is.

Not sure if any of the current maintainers can answer that. It comes from a >20 year old patch: https://github.com/neomutt/neomutt/commit/b4020643fe8b51e1ab50739cbfbe08e2da9751d9#

brunnre8:
>Not sure if any of the current maintainers can answer that. It comes from a >20 year old patch

Well, for starters who am I to judge how old the people in here ;) ?  
Second, I'm a C rookie, might have been a common idiom not yet known to me.

gahr (approved):
Thanks - while you're touching this part, would you mind moving the definition at the point of initialization and making it const?

brunnre8:
as in `size_t n = mutt_str_copy(name, mdata->real_name, sizeof(name));` ?
Is that what you mean?

gahr:
> as in `size_t const n = mutt_str_copy(name, mdata->real_name, sizeof(name));` ?
> Is that what you mean?

Yup, although we usually spell it as `const size_t n..`

brunnre8:
I can't make it const, as the value gets modified when the delimiter gets added.  
But I've moved the declaration in line

brunnre8:
or well, I can... but it doesn't make the code cleaner I think... would mean indexing with n+1 in 2 cases

brunnre8:
I've now made it const. Thanks for the explanations Gahr.

gahr (approved):

PR #2733, opened 2020-11-25
crash: alias_file is a path, not a string

gahr:
* **What does this PR do?**
Fix a crash when trying to access the config variable `alias_file` as a string instead of as a path.

This was introduced in 6f79d959b2d2ca88898ab0c24e4caa70af56967c.

### DIFF
--- a/alias/alias.c
+++ b/alias/alias.c
@@ -482,7 +482,7 @@ retry_name:
   alias_reverse_add(alias);
   TAILQ_INSERT_TAIL(&Aliases, alias, entries);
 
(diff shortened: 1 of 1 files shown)


RomeuG:
My bad, thank you for fixing it. I thought it was a string.
Where do you check that?

gahr:
here (see `DT_PATH`):
https://github.com/neomutt/neomutt/blob/6abb1404ec00ee15c0437babfea9f1d96194d970/alias/config.c#L47

RomeuG:
Okay, I understand now! Thank you once again.

PR #2746, opened 2020-12-02
Always recreate a mailbox after folder-hook

### LINKED ISSUE #2743: Mailboxes list in the sidebar has gone haywire after upgrading
Hi,

For several years I've had multiple accounts set up utilizing `folder-hook`s to source different files with settings unique for each account. Recently I also separated the `mailboxes` for each account and include `unmailboxes *` above them to clear the previous accounts list. This has been working fine until ugrading to 2020-11-20 when the mailbox list in the sidebar started behaving strangely.

By strangely I mean when I switch accounts I will end up with duplicate Inboxes in the sidebar, then if I change folders I might have 2 Sent folders and 2 Inbox folders, for example, and if I continue to switch between accounts or change folders I might end up with 3, 4 or 5 Inbox, Sent or Archive folders etc. in the sidebar.

I've noticed if I uncomment `unmailboxes *` from each file then things settle down but then I may as well just have a global `mailboxes` list instead (as switching accounts appends the new mailboxes to the bottom of the previous accounts in the sidebar).

Downgrading to 2020-09-25 fixes the weirdness.

_Oh, first time reporting a possible issue and hope I've explained things okay and correctly._

gahr:
A folder-hook might have various effects on the mailboxes.

1. opening of an existing mailbox: `unmailboxes ...` might make it
   invalid (the previous code was guarding against this case)
2. opening of a new mailbox: `mailboxes ...` might create and link to an
   account the same mailbox we're trying to open (the new code guards
   against this case too)
3. both of the above, which resulted in a crash in the sidebar

Fixes #2743

### DIFF
--- a/index.c
+++ b/index.c
@@ -741,11 +741,10 @@ static void change_folder_mailbox(struct Menu *menu, struct Mailbox *m, int *old
      * of the function. */
     notify_observer_remove(m->notify, mailbox_index_observer, &m);
   }
-  else
-  {
-    // Recreate the Mailbox (probably because a hook has done `unmailboxes *`)
-    m = mx_path_resolve(dup_path);
-  }
+
+  // Recreate the Mailbox as the folder-hook might have invoked `mailboxes`
+  // and/or `unmailboxes`.
+  m = mx_path_resolve(dup_path);
   FREE(&dup_path);
   FREE(&dup_name);
 


Austin-Ray (approved):
LGTM

flatcap (approved):

PR #2748, opened 2020-12-02
Restore correct `sort_re` documentation

### LINKED ISSUE #2721: Wrong documentation for configuration variable sort_re
## Bug
The documentation for configuration variable [`sort_re`](https://neomutt.org/guide/reference#sort-re) is wrong. It (correctly) states that the variable is a boolean. But it (incorrectly) contains:

> Valid values:
> * count (all message count)
> * desc (virtual mailbox description)
> * new (new message count)
> * path
> * unsorted

This part of the text cannot be right, as a boolean does not have values like `path` or `unsorted`. The error raises suspicions about the correctness of the _entire_ text of the `sort_re` documentation (including the part not cited above).

## Analysis
The bug has been [introduced](https://github.com/neomutt/neomutt/commit/8ffa473227cd121a88aeb7512569dc651a361992#diff-240a2eb5bc025bdda3a6df846f922305301d94642582dbeb61ac89b1eada1ffcR3275-R3287) in 8ffa473227cd121a88aeb7512569dc651a361992 as part of the `notmuch` feature, in file [`init.h`](https://github.com/neomutt/neomutt/blob/8ffa473227cd121a88aeb7512569dc651a361992/init.h). I’m under the impression that the new documentation for `sort_re` which has been introduced in that commit is in some way related to `DT_SORT` and the like, or may be one of the `SortSidebarMethods`. Perhaps it belongs to the `sidebar` feature instead?

## Suggestions
1. Restore the original documentation.
1. Find out what the introduced documentation was intended for; possibly re-introduce it in the proper place.

gahr:
The documentation was bogus'd in 8ffa473227cd121a88aeb7512569dc651a361992.

Fixes #2721

### DIFF
--- a/docs/config.c
+++ b/docs/config.c
@@ -4699,15 +4699,14 @@
 { "sort_re", DT_BOOL, true },
 /*
 ** .pp
-** This variable is only useful when sorting by mailboxes in sidebar. By default,
-** entries are unsorted.  Valid values:
-** .il
-** .dd count (all message count)
-** .dd desc  (virtual mailbox description)
-** .dd new (new message count)
-** .dd path
-** .dd unsorted
-** .ie
+** This variable is only useful when sorting by threads with
+** $$strict_threads \fIunset\fP.  In that case, it changes the heuristic
+** mutt uses to thread messages by subject.  With $$sort_re \fIset\fP, mutt will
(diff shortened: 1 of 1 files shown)


Austin-Ray (approved):
LGTM

PR #2760, opened 2020-12-08
MX API review

gahr:
Small API tidying.

The structure `MxOps` in mx.h defines the interface that mailbox backends such as mbox, maildir, or IMAP implement. The scope of this PR is to review and clean up this API, e.g.,
* functions returning `int` with values `0` (success) and `-1` (failure) can be changed to return `bool`
* functions returning `int` with values `0` (something), `1` (something else), `2` (something different) can be changed to return an `enum`

### DIFF
--- a/command_parse.c
+++ b/command_parse.c
@@ -974,7 +974,7 @@ enum CommandResult parse_mailboxes(struct Buffer *buf, struct Buffer *s,
       }
     }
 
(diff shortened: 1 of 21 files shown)


flatcap (approved):

flatcap:
Rebased over master

PR #2764, opened 2020-12-17
Refactor: use ctx_mailbox(Context)

### LINKED ISSUE #2762: Use more ctx_mailbox
In #2761 I introduced a new API, `struct Mailbox *ctx_mailbox(struct Context* ctx)`, which has the sole purpose to avoid boilerplate like `if (Context && Context->mailbox)`.

I would like to use the API more extensively, whenever `Context->mailbox` is used. This needs to be done with care as `Context` is a global variable and its contents might change when calling other functions.

This would be a nice small refactoring task.

rprimus:
* Refactor (Context && Context->mailbox) to ctx_mailbox(Context)
* Resolves #2762

### DIFF
--- a/index.c
+++ b/index.c
@@ -4,6 +4,7 @@
  *
  * @authors
  * Copyright (C) 1996-2000,2002,2010,2012-2013 Michael R. Elkins <me@mutt.org>
+ * Copyright (C) 2020 R Primus <rprimus@gmail.com>
  *
  * @copyright
  * This program is free software: you can redistribute it and/or modify it under
@@ -689,23 +690,24 @@ static void change_folder_mailbox(struct Menu *menu, struct Mailbox *m, int *old
   if (!m)
     return;
 
+  struct Mailbox *m_ctx = ctx_mailbox(Context);
   /* keepalive failure in mutt_enter_fname may kill connection. */
-  if (Context && Context->mailbox && (mutt_buffer_is_empty(&Context->mailbox->pathbuf)))
+  if (m_ctx && (mutt_buffer_is_empty(&m_ctx->pathbuf)))
     ctx_free(&Context);
 
-  if (Context && Context->mailbox)
+  if (m_ctx)
   {
     char *new_last_folder = NULL;
 #ifdef USE_INOTIFY
     int monitor_remove_rc = mutt_monitor_remove(NULL);
 #endif
 #ifdef USE_COMP_MBOX
-    if (Context->mailbox->compress_info && (Context->mailbox->realpath[0] != '\0'))
-      new_last_folder = mutt_str_dup(Context->mailbox->realpath);
+    if (m_ctx->compress_info && (m_ctx->realpath[0] != '\0'))
+      new_last_folder = mutt_str_dup(m_ctx->realpath);
     else
 #endif
-      new_last_folder = mutt_str_dup(mailbox_path(Context->mailbox));
-    *oldcount = Context->mailbox->msg_count;
+      new_last_folder = mutt_str_dup(mailbox_path(m_ctx));
+    *oldcount = m_ctx->msg_count;
 
     int check = mx_mbox_close(&Context);
     if (check != 0)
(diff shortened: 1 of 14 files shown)


gahr:
This is great, thanks! A next step could be to change instaces of
* check `ctx_mailbox(Context)`
* conditionally use `Context->mailbox`
 
into instances of

* `(const) Mailbox * m = ctx_mailbox(Context);`
* conditionally use `m`

rprimus:
Thu Dec 17 17:02:35 GMT 2020

There are 12 occurences of:
```
 if (!Context || !Context->mailbox)
```
eg `pattern/compile.c:177`

For consistency, should these be changed to:
```
! ctx_mailbox(Context)
```

rprimus:
Fri Dec 18 11:26:49 GMT 2020

In index.c:694, where is (the global) Context initialised and how is Context->mailbox different from the *m that is passed to the function (change_folder_mailbox)?

gahr:
`Context->mailbox` is the current mailbox, `change_folder_mailbox` gets passed the target mailbox you want to open.

rprimus:
Sat Dec 19 04:05:50 GMT 2020

In the `Mailbox` struct definition (`core/mailbox.h`), all numeric values are `int`s.
```
  int msg_count;                      ///< Total number of messages
```

I've noticed (eg in index.c:788), there are comparisons between these struct member  `int` values and `size_t` values.
```
          for (size_t i = 0; i < m->msg_count; i++)
```

What's the view on these?

rprimus:
Sat Dec 19 09:13:46 GMT 2020

Should be all done.  Have (mosty) left `index.c:mutt_index_menu` out of 2nd stage refactor (conditionally use `m`).

Locally, build is successful, however, not `make all-test` - link failure for `pattern/{pattern,compile}.c`: `_ctx_mailbox` symbol: "not found".

Looking into how `Makefile.sutosetup` should be updated.

flatcap:
The fix I've just pushed is a bit ugly (no alternative).
We've got some tests for the Pattern code, but that code references `struct Context` leading to some unpleasant dependencies.

rprimus:
Sat Dec 19 14:31:43 GMT 2020

@flatcap - Thank you for the build fix!

index.c
```diff
@@ -689,23 +690,25 @@ static void change_folder_mailbox(struct Menu *menu, struct Mailbox *m, int *old
   if (!m)
     return;
 
+  struct Mailbox *cm = ctx_mailbox(Context);
```
flatcap, line 693: Please can you rename this to `m_ctx`.
I've given all the "main" structs standard names:
- `struct Mailbox *m`
- `struct Email *e`
etc

Where there are multiple instances, I've added suffixes, e.g. `m_sel`, `m_cur`.

monitor.c
```diff
@@ -542,7 +543,9 @@ int mutt_monitor_remove(struct Mailbox *m)
     goto cleanup;
   }
 
-  if (Context && Context->mailbox)
+  struct Mailbox *cm = ctx_mailbox(Context);
```
flatcap, line 546: Please can you rename this to `m_ctx`

sidebar/window.c
```diff
@@ -288,8 +289,9 @@ static const char *sidebar_format_str(char *buf, size_t buflen, size_t col, int
   if (!m)
     return src;
 
-  bool c = Context && Context->mailbox &&
-           mutt_str_equal(Context->mailbox->realpath, m->realpath);
+  struct Mailbox *cm = ctx_mailbox(Context);
```
flatcap, line 292: Please can you rename this to `m_ctx`

test/pattern/dummy.c
```diff
@@ -259,3 +262,8 @@ void mutt_buffer_select_file(struct Buffer *file, SelectFileFlags flags,
                              char ***files, int *numfiles)
 {
 }
+
+struct Mailbox *ctx_mailbox(struct Context *ctx)
+{
+  return Context ? Context->mailbox : NULL;
+}
```
flatcap, line 269: Please can you add a Doxygen function comment block (for complete-ness).
It doesn't need to say much.

flatcap (changes requested):
Looks good at a first glance.
I'll have a deeper dive, later.

Thanks

sidebar/window.c
rprimus: Mon Dec 21 13:22:10 GMT 2020

Done.  See 15bd1a7

test/pattern/dummy.c
rprimus: Mon Dec 21 13:37:12 GMT 2020

Done - see a12f1ff.

flatcap (approved):

PR #2766, opened 2020-12-20
fix links of Reto

brunnre8:
I guess that Reto is me?  
If so we should probably put my names in ;)


flatcap:
Oh.  Sorry about that.

That file is auto-generated from [`neomutt.txt`](https://github.com/neomutt/management/blob/master/git/neomutt.txt) in the [management repo](https://github.com/neomutt/management/tree/master/git)
I've just updated it.

The "quoted" text mirrors your GitHub login so that the link has a nice tooltip.

flatcap (approved):

Issue #2387, opened 2020-06-25
Better support for <group-alternatives> after attaching other files


dcpurton:
I've been experimenting with `<group-alternatives>` to send multipart/alternative mail.

I'd like to be able to construct the email as I normally would and then create the multipart/alternative version and send it in one keystroke.

I can do this with a macro like this:

```
macro compose y "<first-entry><enter-command>set wait_key=no<enter>| msg2html.sh<enter><attach-file>/tmp/neomutt-alternative.html<enter><tag-entry><first-entry><tag-entry><group-alternatives><enter-command>set wait_key=yes<enter><send-message>" "Send multipart/alternative message"
```

(`msg2html.sh` is a script that calls `pandoc`).

This works well, except if I attach other files before running the macro above. In this case the multipart/alternative message ends up being the last part in the list of mime parts. It seems not possible to make it the first part again. Neomutt returns the error: 'The fundamental part can't be moved'. Other how-tos I've seen suggest creating the multipart/alternative email, then adding other attachments, then sending. But, it would be nice to be able to do things in a more streamlined way.

The email still seems readable by the clients I tested, but not having the actual message as the first part seems untidy. (Neo)Mutt, of course, then places the attached files at the top, rather than the bottom of the email when reading it.

I don't know the technical reasons why 'The fundamental part can't be moved', but perhaps a command like `<make-fundamental>` could be added. Then this could be added to my macro just after `<group-alternatives>` to make the multipart/alternative mime part the first mime part in the email.

dcpurton:
I had a go at a preliminary implementation of a `<make-fundamental>` feature. See https://github.com/dcpurton/neomutt/tree/make-fundamental.

It seems to work OK for me in my tests for what I want.

My macro to send a `multipart/alternative` email is now:

```
macro compose y "<first-entry><enter-command>set wait_key=no<enter>| msg2html.sh<enter><attach-file>/tmp/neomutt-alternative.html<enter><toggle-unlink><tag-entry><first-entry><tag-entry><group-alternatives><toggle-unlink><edit-description>\Ca\Ck<enter><make-fundamental><enter-command>set wait_key=yes<enter><send-message>" "Send multipart/alternative message"
```

[...]

PR #2770, opened 2021-01-10
Move attachment to top of attachment list in compose window

dcpurton:
* **What does this PR do?**

    This patch creates a new function `<move-top>` which allows any attachment to be moved to the top of the list of attachments in the compose window.

    The first commit is a pre-requisite bug fix which syncs the `e->body` and `actx->idx` attachment lists. These lists seem to get out of sync after grouping attachments (either alternatives or multilingual).

     for that)

     Yes.

     I'm not sure how to do this.

     I think so.

     I think so.

    This PR implements feature request #2387.

### DIFF
--- a/compose/compose.c
+++ b/compose/compose.c
@@ -1272,6 +1272,40 @@ static void compose_attach_swap(struct Body *msg, struct AttachPtr **idx, short 
 }
 
 /**
+ * compose_attach_move_top - Move an attachment to the top of the attachment list
+ * @param[in]  e   Email
+ * @param[out] idx Array of Attachments
+ * @param[in]  new Index of attachment to make the body of the email
+ */
+static void compose_attach_move_top(struct Email *e, struct AttachPtr **idx, short new)
+{
+  struct Body *saved_part = e->body;
+
+  /* Reorder Body pointers.
+   * Must traverse msg from top since Body has no previous ptr.  */
+  for (struct Body *part = e->body; part; part = part->next)
+  {
+    if (part->next == idx[new]->body)
+    {
+      part->next = idx[new]->body->next;
+      e->body = idx[new]->body;
+      e->body->next = saved_part;
+      break;
+    }
+  }
+
+  /* Reorder index and ptr->num */
+  struct AttachPtr *saved = idx[new];
+  for (short i = new; i > 0; i--)
+  {
+    idx[i] = idx[i - 1];
+    idx[i]->num += 1;
+  }
+  idx[0] = saved;
+  idx[0]->num = 0;
+}
+
+/**
  * cum_attachs_size - Cumulative Attachments Size
  * @param menu Menu listing attachments
  * @retval num Bytes in attachments
--- a/compose/compose.c
+++ b/compose/compose.c
@@ -1873,6 +1873,10 @@ int mutt_compose_menu(struct Email *e, struct Buffer *fcc, struct Email *e_cur,
               actx->idx[j] = actx->idx[j + 1];
               actx->idx[j + 1] = NULL; /* for debug reason */
             }
+
+            if (i > 0)
+              actx->idx[i - 1]->body->next = actx->idx[i]->body;
+
             actx->idxlen--;
           }
           else
@@ -1892,6 +1896,10 @@ int mutt_compose_menu(struct Email *e, struct Buffer *fcc, struct Email *e_cur,
         struct AttachPtr *gptr = mutt_mem_calloc(1, sizeof(struct AttachPtr));
         gptr->body = group;
         update_idx(menu, actx, gptr);
+
+        /* update e->body pointer */
+        e->body = actx->idx[0]->body;
+
         menu->redraw |= REDRAW_INDEX;
         break;
       }
@@ -1970,6 +1978,10 @@ int mutt_compose_menu(struct Email *e, struct Buffer *fcc, struct Email *e_cur,
               actx->idx[j] = actx->idx[j + 1];
               actx->idx[j + 1] = NULL; /* for debug reason */
             }
+
+            if (i > 0)
+              actx->idx[i - 1]->body->next = actx->idx[i]->body;
+
             actx->idxlen--;
           }
           else
@@ -1989,6 +2001,10 @@ int mutt_compose_menu(struct Email *e, struct Buffer *fcc, struct Email *e_cur,
         struct AttachPtr *gptr = mutt_mem_calloc(1, sizeof(struct AttachPtr));
         gptr->body = group;
         update_idx(menu, actx, gptr);
+
+        /* update e->body pointer */
+        e->body = actx->idx[0]->body;
+
         menu->redraw |= REDRAW_INDEX;
         break;
       }
--- a/compose/compose.c
+++ b/compose/compose.c
@@ -1853,6 +1853,17 @@ int mutt_compose_menu(struct Email *e, struct Buffer *fcc, struct Email *e_cur,
         menu->current++;
         break;
 
+      case OP_COMPOSE_MOVE_TOP:
+        if (menu->current == 0)
+        {
+          mutt_error(_("Attachment is already at top"));
+          break;
+        }
+        compose_attach_move_top(e, actx->idx, menu->current);
+        menu->redraw |= REDRAW_INDEX;
+        menu->current = 0;
+        break;
+
       case OP_COMPOSE_GROUP_ALTS:
       {
         if (menu->tagged < 2)
--- a/compose/compose.c
+++ b/compose/compose.c
@@ -1825,10 +1825,12 @@ int mutt_compose_menu(struct Email *e, struct Buffer *fcc, struct Email *e_cur,
         }
         if (menu->current == 1)
         {
-          mutt_error(_("The fundamental part can't be moved"));
-          break;
+          compose_attach_move_top(e, actx->idx, menu->current);
         }
-        compose_attach_swap(e->body, actx->idx, menu->current - 1);
+        else
+        {
+          compose_attach_swap(e->body, actx->idx, menu->current - 1);
+        }
         menu->redraw |= REDRAW_INDEX;
         menu->current--;
         break;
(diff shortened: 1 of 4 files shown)


gahr:
Why is this needed? I understand you cannot move any arbitrary attachment to the top now because of "The fundamental part can't be moved", but if you want to remove that check, aren't move-up and move-down enough?

dcpurton:
@gahr When you call `<group-alternatives>`, the resulting multipart/alternative attachment is placed at the bottom of the attachment list.

For the email to be well formed, the multipart/alternative part of the message should come first.

You are right that it could be moved up using *n* calls to `<move-up>`, however if other attachments have already been added to the message before the text/plain and text/html parts are grouped, then *n* can not be predicted in advance.

I use a macro to make the the multipart/alternative part and send the message all in one go. This is my macro:

```
macro compose y "<first-entry><enter-command>set wait_key=no<enter>| msg2html.sh<enter><attach-file>/tmp/neomutt-alternative.html<enter><toggle-unlink><tag-entry><first-entry><tag-entry><group-alternatives><toggle-unlink><edit-description>\Ca\Ck<enter><move-top><enter-command>set wait_key=yes<enter><send-message>" "send the message as 'multipart/alternative'"
```

Notice the use of `<move-top>` to move the grouped part to the top. Without `<move-top>`, this macro would need an unknown number of `<move-up>` commands to skip over any other attachments which may be present.

I guess I could just include a large number which would probably work for most cases, but this seems untidy to me and using a `<move-top>` command is neater.

gahr:
I see, perhaps `<group-alternatives>` should be changed to move the grouped parts to the top, or is there a valid use case (which I'm not seeing) for having a multipart/alternative group elsewhere?

dcpurton:
@gahr Good question. Certainly the most common use case for `<group-alternatives>` would be to create an alternative part for the message body and in this case, you do want it at the top (or better, in the same position as the first of the two tagged attachments).

`<group-alternatives>` is pretty basic. There's no concept of one attachment being tagged as the first alternative or any ability to reorder the alternative parts once they are grouped. `<group-alternatives>` just makes the tagged message highest in the attachment list the first part and drops the group at the end of the list of attachments.

I didn't consider just making `<group-alternatives>` automatically put the group attachment at the top as this PR does not change current behaviour and so will likely not break anything for others.

It wouldn't be difficult to change `<group-alternatives>` (and `<group-multilingual>`) to do this. Commit  https://github.com/neomutt/neomutt/pull/2770/commits/f2fb08391994a163a597814c063bc9c7013ec9de would still be needed as I'm pretty sure this is a bug. Then similar logic as used in https://github.com/neomutt/neomutt/pull/2770/commits/4804bfc33755d8526a88a810faec48a08370ea4c could be used to make sure the grouped attachment was left in the right place (say the location of the highest placed attachment to be grouped).

I'm not wedded to there being a `<move-top>` function. It's just a means to an end for me. I want to be able to create the multipart message and send it with one macro. As I said, I only made the `<move-top>` function because this makes what I want possible without changing current behaviour and so potentially breaking things for others.

If you prefer, I can change this PR to only modify `<group-alternatives>` and `<group-multilingual>` (I think it makes sense to do both). Although maybe it is better to open a new PR with this?

gahr:
mhh let me think about it and discuss with @flatcap too. I want to try to get the full picture, I'm sure there are things I'm not seeing yet. Thanks!

flatcap:
Thanks for diving in, @dcpurton (sorry it's taken me a while to respond).

I had a look at the attachment handling a while back, but I only got as far as some analysis (and diagrams).
See:
- https://flatcap.org/mutt/group/lang2.svg
- https://flatcap.org/mutt/group/recv-email.svg
- https://flatcap.org/mutt/group/

The first two show some grouped attachments, showing how the `struct Body`s connect up.
(green arrows (right) are `Body->next`, red arrows (down) are `Body->parts`)

I understand the problem you've found.
`<move-top>` is an OK solution, but I think your "group in place" suggestion is better.
It'll need a little care if the tagged entries aren't contiguous.

There's also the display problem -- once grouped, you can't see the contents of the group.
I didn't get to the root of that.

If you're willing to keep hacking, I'm happy to guide you.

Cheers, @flatcap.

flatcap:
I've pushed some more debug code to `master`.
The added functions include:
- `dump_graphviz_email()` -- dump an Email to a graphviz file
- `dump_graphviz_attach_ctx()` -- dump an attachment context to a graphviz file
- `dump_email()` -- log a LOT of detail about an Email

The graphviz files (`.gv`) can be viewed with ImageMagick's `display` command,
or converted to another format using GraphViz's `dot` command:

`dot -Tsvg email.gv > email.svg`

Note: the memory addresses in the `.gv` files are **real** -- you can use the images for debugging!

dcpurton:
Thanks @flatcap . I'll see if I can put together some code that does a group in place.

I did try using the graphviz things, but without much success. I'll try again with changes you've made.

dcpurton:
@flatcap I can't get `dump_graphviz_email` to look like your pictures.

If I insert `dump_graphviz_email(e)` at the top of the `OP_COMPOSE_GROUP_ALTS` case (before any code has run) I get this:

16:41-email

The output of `dump_graphviz_attach_ctx(actx)` seems better:

16:41-actx

And I can't get `dump_email(e)` to link.

This is my configure command:

```
./configure --mandir=/usr/share/man --libexecdir=/usr/libexec --with-mailpath=/var/mail --gpgme --lua --notmuch --with-ui --gnutls --gss --idn --mixmaster --sasl --tokyocabinet --debug-email --debug-graphviz
```

What am I doing wrong?

dcpurton:
Just an FYI, although commit https://github.com/neomutt/neomutt/pull/2770/commits/f2fb08391994a163a597814c063bc9c7013ec9de works, I think it can be simplified slightly.

flatcap:
> I can't get `dump_graphviz_email` to look like your pictures.

Ah.  I've seen that happen when I also built using ASAN (address sanitizer).
It messes around with memory in a way that I don't understand
(and the graphviz code is very stupid :-)

> And I can't get `dump_email(e)` to link.

Sorry.  I was using `dump_email()` from somewhere else, so the linking order didn't matter too much.
I've fixed the link order and pushed to `master`.

I've also rebased your branch over `master` (hope that's ok).

dcpurton:
> > I can't get `dump_graphviz_email` to look like your pictures.
> 
> Ah. I've seen that happen when I also built using ASAN (address sanitizer).
> It messes around with memory in a way that I don't understand
> (and the graphviz code is very stupid :-)

That's with the `--asan` configure option? I'm not using that though, am I?

> > And I can't get `dump_email(e)` to link.
> 
> Sorry. I was using `dump_email()` from somewhere else, so the linking order didn't matter too much.
> I've fixed the link order and pushed to `master`.

OK, that links now.

> I've also rebased your branch over `master` (hope that's ok).

I also adjusted the first commit to simplify slightly (Hope I rebased and push in the right way).

flatcap:
> (Hope I rebased and push in the right way)

Looks good.  GitHub's fairly forgiving of force-pushing.

> I can't get dump_graphviz_email to look like your pictures.

> That's with the --asan configure option? I'm not using that though, am I?

That option definitely causes problems, but it's off by default.
I don't know what else might be happening.

Both gcc and clang seem to work for me (on Linux).

I've just tested a minimal build: `./configure --disable-doc --debug-graphviz && make`

Then I compose, attach two files and `<group-alternatives>`, giving me:



https://flatcap.org/mutt/group/15%3A06-email.svg

dcpurton:
Woot! I built with `clang` and now I get good output. Hopefully can make some progress now :). Who knows what is going on with `gcc` (I run Debian Testing for what it's worth).

dcpurton:
@flatcap Now that I've seen a bit more clearly how things hang together, I think I can do much better than this PR. With not much code I think it should be possible to get things working really well:

neomutt

flatcap:
> ... now I get good output

lovely :-)

> I think I can do much better than this PR.
> With not much code I think it should be possible to get things working really well:

brilliant!

Any questions, don't hesitate to ask.
(I'm also on IRC, `#neomutt` on `irc.freenode.net`, though our time difference may prove tricky -- I'm in England)

dcpurton:
Here's some proof of concept code:

https://github.com/dcpurton/neomutt/tree/group-in-place

It works and allows me to send a multipart/alternative email. I can add attachments before adding the HTML part and grouping. When grouped, the multipart/alternative part is placed at the position of the first tagged message. And the attachment tree displays as expected. PGP signing and encryption works.

Of course this has significant implications for *many* other functions. (This is one advantage of the current grouping code, making the group a black box allows for other functions to remain unchanged.)

Currently moving attachments around after grouping doesn't work at all and tends to result in a corrupt email and corrupt memory. Neither can you edit the message with `<edit-message>` (although you can use `<edit-mime>` to edit individual parts). And it's safe to say that you shouldn't try and add another attachment after grouping either. There's almost certainly other issues I haven't tried.

Nonetheless, I think this is a good way forward. Fixing the above issues doesn't seem too difficult. It just means you have to take a bit more care in tracking groups and levels.

Also, this is this the best place to discuss this now? This is really heading in a completely different direction to the initial (far more modest) goal of this PR. Shall I open a new issue to discuss there? Or a new PR? Or something else? Or just keep working on things myself and open a new PR when I'm closer to something useful?

You can probably close this pull request if you like.

flatcap:
> Here's some proof of concept code:

My first test worked.  Woo!

> Currently ...

Yeah, beyond that things go a bit pear-shaped :-)

> Fixing the above issues doesn't seem too difficult

Great.  I'll start thinking about a list of tests.
I like comprehensive lists.

> Also, this is this the best place to discuss this now?
> Or something else?

I suggest creating new branch of the form `devel/SHORT-NAME` on the `neomutt` repo
(I've just sent you a project invite)

That'll be **yours** to do whatever you like.
push, force-push, rebase, refactor -- anything

Having the branch on the `neomutt` repo means a couple of things:

- The branch will be automatically built (by Travis and Cirrus)
- Other devs can access it (and contribute) easily

In this case, I think you should create a PR now so that we can discuss development and testing.

> You can probably close this pull request if you like.

OK

Cheers!

dcpurton:
@flatcap Great!

I gather you mean create a `devel/group-in-place` branch, add my code as it currently stands, and then create a pull request from that branch into master (marked as draft or for discussion somehow)?

flatcap:
> I gather you mean create a `devel/group-in-place` branch
> add my code as it currently stands
> and then create a pull request from that branch into master

Exactly

> (marked as draft or for discussion somehow)?

When you create the PR, there are two buttons: <kbd>Create pull request</kbd> or <kbd>Create draft PR</kbd>.
Being a draft PR can be toggled at any time.
It also reduces the number of automatic test builds (to be kind to the free services that we're using).

PR #2771, opened 2021-01-11
Use --libdir to locate dependency libraries

markuspeloquin:
The final component of libdir (from --libdir or computed by autosetup)
should be used in place of 'lib' in auto.def.

This assumes that all libraries use the same library directory name,
which will be true when all libraries are installed in the same
location.

* **What does this PR do?**

This adapts the configure script to work on multilib systems. On multilib Gentoo installations, /usr/lib64 contains native 64-bit libraries and /usr/lib contains 32-bit libraries. OpenSUSE and Slackware do the same apparently from a few minutes of research. (I don't agree with this hierarchy, but it is what it is.)

More specifically, the basename of the `--libdir` argument (or the computed value if `--libdir` not provided, i.e. lib, lib32, or lib64) is used in place of 'lib' within auto.def.

N/A

I believe so.

N/A

### DIFF
--- a/auto.def
+++ b/auto.def
@@ -169,21 +169,25 @@ if {1} {
 ###############################################################################
 # Paths
 define BINDIR           [get-define bindir]
+define LIBDIR           [get-define libdir]
 define MUTTLOCALEDIR    [get-define datadir]/locale
 define PKGDATADIR       [get-define datadir]/neomutt
 define PKGDOCDIR        [opt-val docdir [get-define datadir]/doc/neomutt]
 define SYSCONFDIR       [get-define sysconfdir]
 ###############################################################################
 
+set libdir_name [file tail [get-define libdir]]
+
 ###############################################################################
 # Helper functions
 if {1} {
   # Check for a header file and a function in a library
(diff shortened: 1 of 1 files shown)


markuspeloquin:
In the second revision cd34608 I added `global libdir` in one proc. I don't actually know tcl.

markuspeloquin:
Third revision 6eae153, used `global` again within an `apply` block.

gahr:
Thanks, what you are trying to do makes sense. Can you please follow the same pattern used for defining other paths, see the block starting at line 170 of auto.def

markuspeloquin:
Well, I made a `LIBDIR` following the pattern on lines 170-..., but I can't access the 'define' naturally, so I kept the variable, but renamed to `$libdir_name`.

gahr:
What's wrong with `[get-define libdir]`? Also, please use it full, we don't want `--libdir=/opt/lib` to be changed into $prefix/lib

gahr:
hi @markuspeloquin , are you interested in finishing this up? Please ask questions if anything is not clear from my comments above.

PR #2773, opened 2021-01-12
Make the location of /tmp configurable

brunnre8:
Termux (Android) and Debian packagers would like to be able to configure the
/tmp location to somewhere else. Add a corresponding --with-tmpdir option.

The docs of neomutt (https://neomutt.org/dev/build/autosetup) say that if you set a default option eg `with-mailpath:=/var/mail   => "Location of the spool mailboxes"` you should be able to get the default by using `[opt-val with-mailpath]`. However at least on my machine that isn't actually working and I need to repeat the default argument there.

### DIFF
--- a/auto.def
+++ b/auto.def
@@ -110,6 +110,7 @@ options {
   with-zstd:path            => "Location of Zstandard"
 # System
   with-sysroot:path         => "Target system root"
+  with-tmpdir:=/tmp         => "location of the tmp directory"
 # Misc
   include-path-in-cflags=1  => "Remove include paths from CFLAGS in the output of neomutt -v"
 # Testing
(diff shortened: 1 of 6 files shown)


brunnre8:
Hm, do we actually need to use the same tmp dir for the build itself? Meaning autotools?
Or should this just apply for the actual neomutt code?

brunnre8:
note: docs/config.c:5101 is now not necessarily correct. But I don't quite see how I can do a macro expansion in a C comment?

gahr:
> The docs of neomutt (https://neomutt.org/dev/build/autosetup) say that if you set a default option eg with-mailpath:=/var/mail => "Location of the spool mailboxes" you should be able to get the default by using [opt-val with-mailpath]. However at least on my machine that isn't actually working and I need to repeat the default argument there.

The doc might not be too clear, I fear. Blame me. Also, we're abusing a little the `with-foo:=default` sometimes, and I think we did that for backwards compatibility with the old autoconf options. The way `--with-foo:=default` is supposed to be used is to enable/disable a feature, and have a default value if you want the feature but don't care about specifying a value. Think `with-ssl:=/usr/local`

`./configure` -> `[opt-val with-ssl]` = empty
`./configure --with-ssl` -> `[opt-val with-ssl]` = `/usr/local`
`./configure --with-ssl=/opt/local/lib` -> `[opt-val with-ssl]` = `opt/local/lib`

TL;DR: what you are doing is correct, according to the way we are (ab)using autosetup :)

gahr (approved):

flatcap (approved):

PR #2779, opened 2021-01-22
allow aborting multipart alternative question

ngortheone:
This is work on 

https://gist.github.com/flatcap/9d701d2ec21057dd77c6f685c56647c6#file-0001-allow-to-abort-on-question-about-multipart-alternati-patch


ngortheone:
@flatcap 
It looks like we need to port entire `generate_multipart_alternative()` function

PR #2783, opened 2021-01-24
fix mut_buffer_alloc

ngortheone:
While working on https://github.com/neomutt/neomutt/pull/2782 I found some suspiciously looking things in `mutt_buffer_alloc()`

- #2782 expects Buffer to have non-NULL data after calling `mutt_buffer_alloc()` and explicitly passes a buffer with NULL data
- `mutt_buffer_alloc()` does not perform any checks if `buf->data == NULL` and does things like `buf->dptr = buf->data + offset;` 

I am not a C expert, but NULL + offset does not result in a valid memory address in my head.
I added a few more checks to be more defensive about edge cases.

In addition, I think a few other buffer manipulation functions had forgot to indicate the end of the data with `\0`

### DIFF
--- a/mutt/buffer.c
+++ b/mutt/buffer.c
@@ -65,6 +65,7 @@ struct Buffer mutt_buffer_make(size_t size)
   {
     buf.dptr = buf.data = mutt_mem_calloc(1, size);
     buf.dsize = size;
+    *(buf.dptr) = '\0';
   }
   return buf;
 }
@@ -82,6 +83,7 @@ void mutt_buffer_reset(struct Buffer *buf)
     return;
   memset(buf->data, 0, buf->dsize);
   buf->dptr = buf->data;
+  *(buf->dptr) = '\0';
 }
 
 /**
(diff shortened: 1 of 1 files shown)


ngortheone:
Failing test:
`Test test_mutt_buffer_add_printf...             realloc(): invalid next size`

mutt/buffer.c
```diff
@@ -65,6 +65,7 @@ struct Buffer mutt_buffer_make(size_t size)
   {
     buf.dptr = buf.data = mutt_mem_calloc(1, size);
     buf.dsize = size;
+    *(buf.dptr) = '\0';
```
gahr, line 68: `mutt_mem_calloc` initialized the memory to 0.

mutt/buffer.c
```diff
@@ -82,6 +83,7 @@ void mutt_buffer_reset(struct Buffer *buf)
     return;
   memset(buf->data, 0, buf->dsize);
   buf->dptr = buf->data;
+  *(buf->dptr) = '\0';
```
gahr, line 86: `memset` sets the memory to 0

mutt/buffer.c
```diff
@@ -264,27 +266,48 @@ bool mutt_buffer_is_empty(const struct Buffer *buf)
...
 {
+
+  // no buffer - no work
   if (!buf)
   {
     return;
   }
 
+  // size must be positive
+  if (new_size <= 0)
```
gahr, line 277: `new_size` is an unsigned type, it cannot be negative.

mutt/buffer.c
```diff
@@ -264,27 +266,48 @@ bool mutt_buffer_is_empty(const struct Buffer *buf)
...
+    return;
+  }
+
+  // if data pointer is screwed then we fix set by setting it to data beginning
   if (!buf->dptr)
   {
     buf->dptr = buf->data;
   }
 
-  if ((new_size > buf->dsize) || !buf->data)
```
gahr, line 277: here's the check on `buf->data == NULL` you're looking for...

mutt/buffer.c
```diff
@@ -264,27 +266,48 @@ bool mutt_buffer_is_empty(const struct Buffer *buf)
...
-  {
-    size_t offset = (buf->dptr && buf->data) ? buf->dptr - buf->data : 0;
```
gahr, line 279: which results in offset being set to 0...

mutt/buffer.c
```diff
@@ -264,27 +266,48 @@ bool mutt_buffer_is_empty(const struct Buffer *buf)
...
-
-    buf->dsize = new_size;
-    mutt_mem_realloc(&buf->data, buf->dsize);
-    buf->dptr = buf->data + offset;
```
gahr, line 283: and `buf->dptr` being set to `buf->data`.

gahr:
I don't see a bug in the code as it's currently written. Please can you read my comments and see if they answer your doubts.

gahr (changes requested):

PR #2789, opened 2021-01-27
Add mutt_buffer_seek() function

ngortheone:
upstream patch: https://gist.github.com/flatcap/9d701d2ec21057dd77c6f685c56647c6#file-0006-add-mutt_buffer_rewind-function-patch

This makes it a bit clearer what the assignment is doing, and reduces
direct dptr manipulation a bit.

### DIFF
--- a/mutt/buffer.c
+++ b/mutt/buffer.c
@@ -81,7 +81,7 @@ void mutt_buffer_reset(struct Buffer *buf)
   if (!buf || !buf->data || (buf->dsize == 0))
     return;
   memset(buf->data, 0, buf->dsize);
-  buf->dptr = buf->data;
+  mutt_buffer_seek(buf, 0);
 }
 
 /**
@@ -183,7 +183,7 @@ void mutt_buffer_fix_dptr(struct Buffer *buf)
   if (!buf)
     return;
 
-  buf->dptr = buf->data;
+  mutt_buffer_seek(buf, 0);
 
   if (buf->data && (buf->dsize > 0))
   {
@@ -271,7 +271,7 @@ void mutt_buffer_alloc(struct Buffer *buf, size_t new_size)
 
   if (!buf->dptr)
   {
-    buf->dptr = buf->data;
+    mutt_buffer_seek(buf, 0);
   }
 
   if ((new_size > buf->dsize) || !buf->data)
@@ -280,7 +280,7 @@ void mutt_buffer_alloc(struct Buffer *buf, size_t new_size)
 
     buf->dsize = new_size;
     mutt_mem_realloc(&buf->data, buf->dsize);
-    buf->dptr = buf->data + offset;
+    mutt_buffer_seek(buf, offset);
     /* This ensures an initially NULL buf->data is now properly terminated. */
     if (buf->dptr)
       *buf->dptr = '\0';
@@ -453,3 +453,20 @@ size_t mutt_buffer_copy(struct Buffer *dst, const struct Buffer *src)
 
   return mutt_buffer_addstr_n(dst, src->data, mutt_buffer_len(src));
 }
+
+
+/** 
+ * mutt_buffer_seek - set current read/write position to offset from beginning
+ * @param buf    Buffer to use
+ * @param offset Distance from the beginning
+ *
+ * This is used for cases where the buffer is read from
+ * A value is placed in the buffer, and then b->dptr is set back to the
+ * beginning as a read marker instead of write marker.
+ */
+void mutt_buffer_seek (struct Buffer *buf, size_t offset)
+{
+  if (buf)
+    buf->dptr = buf->data + offset;
+}
+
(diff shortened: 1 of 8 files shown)


gahr:
If you allow an offset argument, I'd go for `seek` instead of `rewind`.

mutt/buffer.c
```diff
@@ -265,22 +265,18 @@ bool mutt_buffer_is_empty(const struct Buffer *buf)
...
   if (!buf)
-  {
     return;
-  }
 
   if (!buf->dptr)
-  {
-    buf->dptr = buf->data;
-  }
+    mutt_buffer_rewind(buf, 0);
```
gahr, line 271: I'd keep those parentheses

mutt/buffer.c
```diff
@@ -453,3 +449,19 @@ size_t mutt_buffer_copy(struct Buffer *dst, const struct Buffer *src)
...
+ * @param buffer Buffer to rewind
+ * @param offset size
+ *
+ * This is used for cases where the buffer is read from
+ * A value is placed in the buffer, and then b->dptr is set back to the
+ * beginning as a read marker instead of write marker.
+ */
+void mutt_buffer_rewind (struct Buffer *buf, size_t offset)
+{
+  buf->dptr = buf->data + offset;
```
gahr, line 465:
```
if (!buf)
{
  return;
}
```

mutt/buffer.c
```diff
@@ -264,14 +264,14 @@ bool mutt_buffer_is_empty(const struct Buffer *buf)
  */
 void mutt_buffer_alloc(struct Buffer *buf, size_t new_size)
 {
-  if (!buf)
+  if (!buf)
```
flatcap, line 267: Careful!  Some trailing whitespace has crept in.

Running `clang-format` on your changed code will fix this
(and check other things like indentation and `{}`-placement.

mutt/buffer.c
```diff
@@ -264,14 +264,14 @@ bool mutt_buffer_is_empty(const struct Buffer *buf)
...
 void mutt_buffer_alloc(struct Buffer *buf, size_t new_size)
 {
-  if (!buf)
+  if (!buf) 
   {
     return;
   }
 
-  if (!buf->dptr)
+  if (!buf->dptr)
```
flatcap, line 272: trailing whitespace

mutt/buffer.h
```diff
@@ -51,6 +51,7 @@ size_t         mutt_buffer_len          (const struct Buffer *buf);
 struct Buffer  mutt_buffer_make         (size_t size);
 void           mutt_buffer_reset        (struct Buffer *buf);
 char *         mutt_buffer_strdup       (const struct Buffer *buf);
+void           mutt_buffer_seek         (struct Buffer *, size_t offset);
```
flatcap, line 54: Please can you name the `Buffer` parameter.
(upstream are slack about that)

flatcap (changes requested):
Looks good, just a few trivial suggestions.

mutt/buffer.h
ngortheone: done

mutt/buffer.c
```diff
@@ -453,3 +453,20 @@ size_t mutt_buffer_copy(struct Buffer *dst, const struct Buffer *src)
 
   return mutt_buffer_addstr_n(dst, src->data, mutt_buffer_len(src));
 }
+
+
+/** 
+ * mutt_buffer_seek - set current read/write position to offset from beginning
+ * @param buffer Buffer to use
+ * @param offset size position from beginning to set
```
gahr, line 461: "size position" reads weird

gahr (approved):

mutt/buffer.c
ngortheone: I was using the "@param {{param_name}} {{param_type}} {{explanation}}" pattern

I am open to suggestions how to improve this. English is not my best skill :)

flatcap:
Sorry I keep missing you on IRC

> I think it will be smarter to do assert(buf)

I have mixed feelings here.

If a user manages to trigger a bug, what's the result?
- `if (!ptr)`
  A command/function doesn't work, but NeoMutt keeps running
- `assert(ptr)`
  NeoMutt crashes

Both are likely to get reported, but the **perception** of the first case is better.

We've been through some periods of instability.
Once people start saying, "NeoMutt crashes", it's hard to recover.

---

One possible compromise might be to create `mutt_assert()`.
We already have code that can dump the call stack; we could encourage the packagers to enable it.
`./configure --debug-backtrace` (uses libunwind)

Our "assert" could log the call stack, print an error message and return failure.
Even if the user isn't running with logging enabled, they could view the info using the `<show-log-messages>` function.

flatcap (approved):

ngortheone:
@flatcap 

> We've been through some periods of instability.
> Once people start saying, "NeoMutt crashes", it's hard to recover.

The idea is that assert statements are compiled out in release builds. People who run debug builds know what they are doing. 
Also from my point of view crashing with a clear message and stacktrace is better than silently not doing anything.

We can combine approaches and have
```c
mutt_assert(ptr); // crashes on debug builds and gives stacktrace
if (!ptr) return;  //  on release builds no crash

```

PR #2792, opened 2021-01-28
 Abort IMAP open if condstore/qresync updates fetch fails

ngortheone:
Upstream patch https://gist.github.com/flatcap/9d701d2ec21057dd77c6f685c56647c6#file-0010-abort-imap-open-if-condstore-qresync-updates-fetch-f-patch

Abort IMAP open if condstore/qresync updates fetch fails

An error in imap_cmd_step() was not being properly returned to the
caller.


gahr (approved):

flatcap (approved):
I've altered the commit message to reference the upstream commit, rather than my gist'd copy.

PR #2800, opened 2021-01-30
build: add debug and release  build-types

ngortheone:
This patch adds various build types:

- debug:      (-O0 -g)
- release: (-O2 -DNDEBUG)

To preserve compatibility with old behavior release build type is set
by default. Debug build types also compile in asserts to allow verifying
invariants in the code.

Usage example: configure --build-type=debug

### DIFF
--- a/auto.def
+++ b/auto.def
@@ -26,6 +26,7 @@ set subdirs {po docs contrib}
 ###############################################################################
 # Add any user options here
 options {
+  build-type:=release       => "Select build type: release or debug"
 # Curses / S-Lang
   with-ui:=ncurses          => "Select ncurses or slang for the UI"
   with-ncurses:path         => "Location of ncurses"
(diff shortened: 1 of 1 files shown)


gahr:
The code in the  autosetup directory is not ours, please avoid changing it. How about a single `--debug` flag that just adds `-O0 -ggdb` and adding `-DNDEBUG` to the default build? You can leave `-O2` in the default cflags: arguments you append will take precedence.

flatcap:
> `--debug`

Unfortunately, that's use by autosetup: "display debugging output as autosetup runs"

ngortheone:
@gahr I have reverted changes to `autosetup/cc.tcl`

gahr:
> > `--debug`
> 
> Unfortunately, that's use by autosetup: "display debugging output as autosetup runs"

ah, sure, how about a default on `--release` flag then?

gahr:
My point here is that we're introducing an effectively boolean flag (debug / release) as a string config switch, just to check later that it's either of the two accepted values ;)

ngortheone:
originally I was planning to have more build types. Like MinimumSizeRelease and ReleaseWithDebug

ngortheone:
Similar to this
https://github.com/neovim/neovim/blob/master/contrib/local.mk.example#L13-L26

gahr:
honestly, I'm not sure.. tools over policies: you can already `make EXTRA_CFLAGS="-O3 -ggdb -Wall -wathever"` so perhaps any hard-coded build mode is superfluous anyway.

ngortheone:
motion dismissed. no consensus

PR #2801, opened 2021-01-31
buffer: replace mutt_b2s macro with mutt_buffer_string function

ngortheone:
Macro

```c
#define mutt_b2s(buf) ((buf)->data ? (const char *) (buf)->data : "")
```

- poorly named, does not follow mutt_buffer_action naming pattern
- can be confused with a function call [1]
- is not type-safe
- allows mutating underlying buffers data

A new function is introduced instead - `mutt_buffer_string()`
- follows naming pattern
- function name clearly communicates the contract
- can still be inlined by the compiler
- does not allow mutating the data

And macroses are evil in general. So get rid of them.

[1] https://drewdevault.com/2021/01/19/Spooky-code-at-a-distance.html

### DIFF
--- a/mutt/buffer.c
+++ b/mutt/buffer.c
@@ -454,7 +454,7 @@ size_t mutt_buffer_copy(struct Buffer *dst, const struct Buffer *src)
   return mutt_buffer_addstr_n(dst, src->data, mutt_buffer_len(src));
 }
 
-/** 
+/**
  * mutt_buffer_seek - set current read/write position to offset from beginning
  * @param buf    Buffer to use
  * @param offset Distance from the beginning
--- a/mutt/buffer.h
+++ b/mutt/buffer.h
@@ -37,9 +37,6 @@ struct Buffer
   size_t dsize; ///< Length of data
 };
 
-/* Convert a buffer to a const char * "string" */
-#define mutt_b2s(buf) ((buf)->data ? (const char *) (buf)->data : "")
-
 #define MoreArgs(buf) (*(buf)->dptr && (*(buf)->dptr != ';') && (*(buf)->dptr != '#'))
 
 void           mutt_buffer_alloc        (struct Buffer *buf, size_t size);
@@ -67,5 +64,22 @@ int            mutt_buffer_printf       (struct Buffer *buf, const char *fmt, ..
 size_t         mutt_buffer_strcpy       (struct Buffer *buf, const char *s);
 size_t         mutt_buffer_strcpy_n     (struct Buffer *buf, const char *s, size_t len);
 size_t         mutt_buffer_substrcpy    (struct Buffer *buf, const char *beg, const char *end);
+
+/**
+ * mutt_buffer_string - Convert a buffer to a const char * "string"
+ * @param buf Buffer to that is to be converted
+ * @retval ptr String inside the Buffer
+ *
+ * This method exposes Buffer's underlying data field
+ *
+ * @note Returns an empty string if Buffer isn't initialised
+ */
+static inline const char *mutt_buffer_string(const struct Buffer *buf)
+{
+  if (!buf || !buf->data)
+    return "";
+
+  return buf->data;
+}
 
 #endif /* MUTT_LIB_BUFFER_H */
--- a/alias/dlgquery.c
+++ b/alias/dlgquery.c
@@ -247,10 +247,10 @@ static int query_run(char *s, bool verbose, struct AliasList *al,
   const char *query_command = cs_subset_string(sub, "query_command");
   mutt_buffer_file_expand_fmt_quote(cmd, query_command, s);
 
-  pid_t pid = filter_create(mutt_b2s(cmd), NULL, &fp, NULL);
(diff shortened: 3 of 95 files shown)


mutt/buffer.c
```diff
@@ -468,3 +468,20 @@ void mutt_buffer_seek(struct Buffer *buf, size_t offset)
...
 }
+
+
+/**
+ * mutt_buffer_as_rostr - Convert a buffer to a const char * "string" 
+ * @param buf Buffer to that is to be converted
+ *
+ * This method exposes Buffer's underlying data field
+ */
+const char * const mutt_buffer_as_rostr (const struct Buffer *buf)
```
gahr, line 479: A few comments:

1. you probably want `const char *` in the return type, as returning a variable that cannot be reassigned to makes little sense
2. I don't particularly like the name: how about `mutt_buffer_string`?
3. I don't think you get much chance for inlining, unless you've got some clever link-time optimization in place: how about defining this as inline in the header file?

ngortheone: 1. fixed
2. fixed
3. fixed
4. I am not sure what NONULL is. Can you explain? Especially in the context of C99 which our compile flags seem to target explicitly. ( I suspect this is some kind of gcc-only magic? )

gahr: > I am not sure what NONULL is.

It's this little macro of ours: https://github.com/neomutt/neomutt/blob/master/mutt/string2.h#L37

mutt/buffer.h
```diff
@@ -37,35 +37,48 @@ struct Buffer
...
-size_t         mutt_buffer_strcpy       (struct Buffer *buf, const char *s);
-size_t         mutt_buffer_strcpy_n     (struct Buffer *buf, const char *s, size_t len);
-size_t         mutt_buffer_substrcpy    (struct Buffer *buf, const char *beg, const char *end);
+size_t             mutt_buffer_concat_path  (struct Buffer *buf, const char *dir, const char *fname);
+size_t             mutt_buffer_concatn_path (struct Buffer *dst, const char *dir, size_t dirlen, const char *fname, size_t fnamelen);
+size_t             mutt_buffer_copy         (struct Buffer *dst, const struct Buffer *src);
+int                mutt_buffer_printf       (struct Buffer *buf, const char *fmt, ...);
+size_t             mutt_buffer_strcpy       (struct Buffer *buf, const char *s);
+size_t             mutt_buffer_strcpy_n     (struct Buffer *buf, const char *s, size_t len);
+size_t             mutt_buffer_substrcpy    (struct Buffer *buf, const char *beg, const char *end);
```
gahr, line 66: I think most of the reindentation in this file is superfluous.

mutt/buffer.c
gahr: However, I now realize that mutt_b2s did the same thing, so yeah, I think it's fine to fallback to an empty string, unless we can prove that callers don't use the feture.

mutt/buffer.h
ngortheone: fixed

flatcap:
I'm having trouble with this.
Conceptually it's fine, but my build keeps failing.

```sh
export EXTRA_CFLAGS='-O0'
./configure --disable-doc CC=gcc
make CC=gcc
```

fails with:

```
/usr/bin/ld: browser.o: in function `mailbox_path':
/home/mutt/work/neo/core/mailbox.h:206: undefined reference to `mutt_buffer_string'
/usr/bin/ld: browser.o: in function `link_is_dir':
/home/mutt/work/neo/browser.c:332: undefined reference to `mutt_buffer_string'
/usr/bin/ld: browser.o: in function `examine_directory':
/home/mutt/work/neo/browser.c:771: undefined reference to `mutt_buffer_string'
/usr/bin/ld: /home/mutt/work/neo/browser.c:783: undefined reference to `mutt_buffer_string'
/usr/bin/ld: browser.o: in function `examine_mailboxes':
/home/mutt/work/neo/browser.c:870: undefined reference to `mutt_buffer_string'
```

and the same with `CC=clang`

```
/usr/bin/ld: browser.o: in function `mutt_browser_select_dir':
/home/mutt/work/neo/browser.c:1129: undefined reference to `mutt_buffer_string'
/usr/bin/ld: browser.o: in function `mutt_buffer_select_file':
/home/mutt/work/neo/browser.c:1194: undefined reference to `mutt_buffer_string'
/usr/bin/ld: /home/mutt/work/neo/browser.c:1198: undefined reference to `mutt_buffer_string'
/usr/bin/ld: /home/mutt/work/neo/browser.c:1209: undefined reference to `mutt_buffer_string'
/usr/bin/ld: /home/mutt/work/neo/browser.c:1216: undefined reference to `mutt_buffer_string'
```

flatcap (approved):
`mutt_b2s()` was certainly flawed, but its beauty was that it was short.
The new name has meant a lot a lines needed to be re-wrapped and a few `{}`s inserted for clarity (done).

You can see the separate changes in [[devel/buffers-tidy]](https://github.com/neomutt/neomutt/tree/devel/buffers-tidy)

PR #2802, opened 2021-02-01
Add <view-pager> function to attachment menu

ngortheone:
This uses a copiousoutput mailcap entry, or falls back to raw text.

The existing functions do not provide a way to use copiousoutput
mailcap entries exclusively.  The default <view-attach> will use one
for internally supported types if auto_view is configured, but there
is no way to do so for other mime types (such as application/pdf).

Upstream-commit: https://gitlab.com/muttmua/mutt/commit/474d368e859aa86b4adf9e41f1048488ece99c42
https://gist.github.com/flatcap/9d701d2ec21057dd77c6f685c56647c6#file-0004-add-view-pager-function-to-attachment-menu-patch

### DIFF
--- a/docs/manual.xml.head
+++ b/docs/manual.xml.head
@@ -10251,6 +10251,87 @@ macro index ,a "&lt;save-message&gt;=archive&lt;enter&gt;&lt;enter-command&gt;ec
           See table <xref linkend="tab-attachment-bindings" /> for all
           available functions.
         </para>
+
+        <sect3 id="attach-viewers">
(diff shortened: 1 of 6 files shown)


flatcap (approved):
I've made a couple of minor changes:
- `clang-format`
- Bracketed logic (a personal preference)

C allows code
```c
if (x == 1 && y == 2 && z == 3)
```

I prefer to have any non-trivial logic bracketed (e.g. if it contains whitespace)
```c
if ((x == 1) && (y == 2) && (z == 3))
```

PR #2804, opened 2021-02-02
Add <skip-headers> to skip past message headers in pager

ngortheone:
This is based on `<skip-quoted>`, and is bound to 'H' by default.

I've added an error message in case there is no text past the headers,
just to cover the logical case.  I don't think Mutt allows that to
happen for text passed to the pager, but just in case.

Upstream-commit: https://gitlab.com/muttmua/mutt/commit/42331065e432c03903eac2fc8cc2aa94f7da501e

### DIFF
--- a/docs/manual.xml.head
+++ b/docs/manual.xml.head
@@ -2525,18 +2525,12 @@ color sidebar_divider   color8  default     <emphasis role="comment"># Dark grey
           </varlistentry>
           <varlistentry>
             <term>
-              <literal>&lt;toggle-quoted&gt;</literal> (default: T)
-              <anchor id="toggle-quoted" />
+              <literal>&lt;skip-headers&gt;</literal> (default: H) 
+              <anchor id="skip-headers"/>
             </term>
             <listitem>
               <para>
-                The pager uses the
-                <link linkend="quote-regex">$quote_regex</link> variable to
-                detect quoted text when displaying the body of the message.
-                This function toggles the display of the quoted material in the
-                message. It is particularly useful when being interested in
-                just the response and there is a large amount of quoted text in
-                the way.
+                This function will skip past the headers of the current message.
               </para>
               <para>
                 The variable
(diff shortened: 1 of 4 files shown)


ngortheone:
@flatcap this is still in progress, I `rd.pager_window` is probably somewhere else, I didn't have time to find it..

pager.c
```diff
@@ -2926,6 +2926,47 @@ int mutt_pager(const char *banner, const char *fname, PagerFlags flags, struct P
...
+            mutt_error(_("Already skipped past headers."));
+            break;
+          }
+ 
+          while ((new_topline < rd.last_line ||
+                  (0 == (dretval = display_line(rd.fp, &rd.last_pos, &rd.line_info,
+                         new_topline, &rd.last_line, &rd.max_line, 
+                         MUTT_TYPES | (flags & MUTT_PAGER_NOWRAP),
+                         &rd.quote_list, &rd.q_level, &rd.force_redraw, 
+                         &rd.search_re, rd.pager_window))))
```
gahr, line 2951:
```suggestion
                         &rd.search_re, rd.extra->win_pager))))
```

docs/manual.xml.head
```diff
@@ -2578,6 +2572,21 @@ color sidebar_divider   color8  default     <emphasis role="comment"># Dark grey
...
             </listitem>
           </varlistentry>
+          <varlistentry>
+            <term>
+              <literal>&lt;toggle-quoted&gt;</literal><anchor id="toggle-quoted"/>
+              (default: T)
+            </term>
+            <listitem>
+              <para>
+                The pager uses the <link linkend="quote-regexp">$quote_regexp</link>
```
flatcap, line 2582:
```suggestion
                The pager uses the <link linkend="quote-regex">$quote_regex</link>
```

We standardised on "regex" (no 'p') for the config and code.

pager.c
```diff
@@ -2926,6 +2926,47 @@ int mutt_pager(const char *banner, const char *fname, PagerFlags flags, struct P
...
+          int dretval = 0;
+          int new_topline = rd.topline;
+ 
+          if (!IS_HEADER (rd.line_info[new_topline].type))
+          {
+            /* L10N:
+               Displayed if <skip-headers> is invoked in the pager, but we are
+               already past the headers
+            */
+            mutt_error(_("Already skipped past headers."));
```
flatcap, line 2942: Mutt has a lot of **errors**.
If we can tell the user **why** their action failed (or what to do next time),
then I'd like to downgrade the `mutt_error()` to `mutt_warning()` or `mutt_message()`.

flatcap:
I've made one small tweak.
Rather then an `if (rd.has_types)` that spanned the entire `case`, I've inverted it to return early.
This reduces the indent of the code block.

flatcap (approved):

PR #2806, opened 2021-02-03
 Add a sanity check for qresync

ngortheone:
If it fails, force a redownload without qresync, to set the uid_seqset
to a correct state.

Note the headers should almost all be in the header cache, so this
isn't a huge penalty, but ensures no duplicates occur (the most common
issue if something goes wrong).

Upstream-commit: https://gitlab.com/muttmua/mutt/commit/1598baa2ffd1a62053fea10bf2550d6cfbd91694

https://gist.github.com/flatcap/9d701d2ec21057dd77c6f685c56647c6#file-0012-add-a-sanity-check-for-qresync-patch

### DIFF
--- a/imap/message.c
+++ b/imap/message.c
@@ -29,6 +29,7 @@
  */
 
 #include "config.h"
+#include <assert.h>
 #include <ctype.h>
 #include <stdbool.h>
 #include <stdint.h>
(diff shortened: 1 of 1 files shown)


ngortheone:
This one is not yet ready. `imap_verify_qresync()` needs to be refactored, as accepted parameter type was changed in neomutt.

imap/message.c
```diff
@@ -985,6 +985,77 @@ static int read_headers_condstore_qresync_updates(struct ImapAccountData *adata,
...
+
+/**
+ * imap_verify_qresync - Check to see if QRESYNC got jumbled. 
+ * If so, wipe the context and try again with a normal download. 
+ * @param m  Imap Selected Mailbox
+ *
+ * @retval  0 Success
+ * @retval -1 Error
+ */
+// TODO refactor this
```
flatcap, line 997: All the data we need for this function is in the `Mailbox`, or its private mdata `ImapMboxData`.
Data that was Account-level in Mutt, is now Mailbox-level.
The single `ctx` in Mutt is just the `m` Mailbox in NeoMutt.
Oh, and a bool name has been swapped for its opposite -- I'll leave you to discover which :-)

ngortheone:
This is ready for review now.

flatcap:
er, it doesn't compile

ngortheone:
:confused:  compiled on my machine after `make clean` :man_shrugging:

flatcap:
It fails if the header cache is configured:
```
imap/message.c:1459:7: error: ‘uid_validity’ undeclared
```

See the TravisCI logs, below

ngortheone:
my local build was configured without hcache :facepalm: 
should be better now

flatcap:
My default build (`build.sh`) includes most things,
and before a release, I'll run a **wide** build (`build-parallel.sh`) of config options.
https://github.com/neomutt/management/tree/master/build

If you haven't already installed it, I thoroughly using `ccache` to speed up your builds.
https://neomutt.org/dev/analysis/compiler

ngortheone:
I do have ccache, I need to install all the libraries needed for building with all dependencies. Maybe a docker container, as I am running a pretty spartan Gentoo system

imap/message.c
```diff
@@ -985,6 +986,74 @@ static int read_headers_condstore_qresync_updates(struct ImapAccountData *adata,
...
+  if (!adata || (adata->mailbox != m))
+    return -1;
+
+  const size_t max_msn = imap_msn_highest(&mdata->msn);
+
+  unsigned int msn;
+  unsigned int uid;
+  struct Email *e = NULL;
+  struct Email *uidh = NULL;
+
```
flatcap, line 1012: I've moved the `int i;` definition into the `for` (and below too), to reduce `i`'s scope.

imap/message.c
```diff
@@ -985,6 +986,74 @@ static int read_headers_condstore_qresync_updates(struct ImapAccountData *adata,
...
+  struct ImapMboxData *mdata = imap_mdata_get(m);
+  if (!adata || (adata->mailbox != m))
+    return -1;
+
+  const size_t max_msn = imap_msn_highest(&mdata->msn);
+
+  unsigned int msn;
+  unsigned int uid;
+  struct Email *e = NULL;
+  struct Email *uidh = NULL;
```
flatcap, line 1011: I've initialised all pointers to NULL.

flatcap (approved):

PR #2809, opened 2021-02-05
asan: fix leaking addrinfo for non-canonical domain names

### LINKED ISSUE #2808: [Low importance] ASAN detects leaks on exit
I am running master build with `--asan` configuration option.

NeoMutt works fine while I use it. When I exit NeoMutt I get this error. 

```
=================================================================
==2291964==ERROR: LeakSanitizer: detected memory leaks

Direct leak of 64 byte(s) in 1 object(s) allocated from:
    #0 0x7f94a7dec847 in malloc (/usr/lib/gcc/x86_64-pc-linux-gnu/10.2.0/libasan.so.6+0xaf847)
    #1 0x7f94a7bad9e7  (/lib64/libc.so.6+0xe39e7)
    #2 0x7f94a7baec44 in getaddrinfo (/lib64/libc.so.6+0xe4c44)
    #3 0x7f94a7dc66c6  (/usr/lib/gcc/x86_64-pc-linux-gnu/10.2.0/libasan.so.6+0x896c6)
    #4 0x7f94a7c89442  (/lib64/libanl.so.1+0x1442)

Indirect leak of 548 byte(s) in 8 object(s) allocated from:
    #0 0x7f94a7dec847 in malloc (/usr/lib/gcc/x86_64-pc-linux-gnu/10.2.0/libasan.so.6+0xaf847)
    #1 0x7f94a7bad9e7  (/lib64/libc.so.6+0xe39e7)
    #2 0x7f94a7baec44 in getaddrinfo (/lib64/libc.so.6+0xe4c44)
    #3 0x7f94a7dc66c6  (/usr/lib/gcc/x86_64-pc-linux-gnu/10.2.0/libasan.so.6+0x896c6)
    #4 0x7f94a7c89442  (/lib64/libanl.so.1+0x1442)

Indirect leak of 9 byte(s) in 1 object(s) allocated from:
    #0 0x7f94a7dec847 in malloc (/usr/lib/gcc/x86_64-pc-linux-gnu/10.2.0/libasan.so.6+0xaf847)
    #1 0x7f94a7b54f7a in strdup (/lib64/libc.so.6+0x8af7a)
    #2 0x6c6568646e6976ff  (<unknown module>)

SUMMARY: AddressSanitizer: 621 byte(s) leaked in 10 allocation(s).
```

No matter how long I use neomutt, the byte count and allocation count is always the same.

I don't know if this is normal, but to me it seems strange that ASAN reports leaks caused libc
I use Gentoo, `sys-libs/glibc-2.32-r7` 

This might be not related to neomutt too. I just don't have other ASAN enabled applications to confirm that this is more widespread.

ngortheone:
Fixes #2808

### DIFF
--- a/conn/getdomain.c
+++ b/conn/getdomain.c
@@ -27,6 +27,7 @@
  */
 
 #include "config.h"
+#include <assert.h>
 #include <netdb.h>
 #include <string.h>
 #include <sys/socket.h>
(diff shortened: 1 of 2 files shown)


conn/getdomain.c
```diff
@@ -99,8 +99,9 @@ int getdnsdomainname(struct Buffer *domain)
     mutt_buffer_strcpy(domain, ++p);
     rc = 0;
     mutt_debug(LL_DEBUG1, "Hostname: %s\n", mutt_buffer_string(domain));
-    freeaddrinfo(h);
   }
+  if (h)
+    freeaddrinfo(h);
```
flatcap, line 104: That'll do the trick, but I think I might have gone for:
```c
  if (h)
  {
    if (h->ai_canonname && (p = strchr(h->ai_canonname, '.')))
    {
      mutt_buffer_strcpy(domain, ++p);
      rc = 0;
      mutt_debug(LL_DEBUG1, "Hostname: %s\n", mutt_buffer_string(domain));
    }
    freeaddrinfo(h);
  }
```

gahr:
What I would probably do is make two functions, one for getaddrinfo_a and one for getaddrinfo, that take hints and return h or NULL. :)

conn/getdomain.c
gahr: The check on `h->ai_canonname` is useless, because of the `AI_CANONNAME` flag we're passing :)

ngortheone: done

flatcap (approved):
Ran `clang-format` and added Doxygen comment blocks.

PR #2810, opened 2021-02-06
Added Locale headers for OSX platforms

zakarouf:
Compiling on macos will give
```mx.c:1113:7: error: use of undeclared identifier 'locale_t'```
Error
To counteract that Added header for OSX.

* **What does this PR do?**

Added OSX Platform-specific headers

### DIFF
--- a/mx.c
+++ b/mx.c
@@ -80,8 +80,7 @@
 #include <libintl.h>
 #endif
 #ifdef __APPLE__
-# include <string.h>
-# include <xlocale.h>
+#include <xlocale.h>
 #endif
 
 /* These Config Variables are only used in mx.c */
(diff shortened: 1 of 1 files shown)

### CODE REVIEW

File: mx.c [hunk 1]
```diff
@@ -79,6 +79,10 @@
 #ifdef ENABLE_NLS
 #include <libintl.h>
 #endif
+#ifdef __APPLE__
+# include <string.h>
```
gahr, line 83: why string?

zakarouf: I used to have same problems in older version on OSX. So i included both to be safe
As of now, I tested it on 10.12. It is working with only <xlocale.h> included.
Shall i remove <string.h> then?
Thanks.

gahr (approved):

PR #2812, opened 2021-02-10
pager de-globalization phase 1: removal of Contex2

ngortheone:
This is PR aims to remove the usage of the global variable Context (marked as Contex2)

### DIFF
--- a/gui/curs_lib.c
+++ b/gui/curs_lib.c
@@ -731,7 +731,7 @@ int mutt_do_pager(const char *banner, const char *tempfile, PagerFlags do_color,
 
   if (!C_Pager || mutt_str_equal(C_Pager, "builtin"))
   {
-    rc = mutt_pager(banner, tempfile, do_color, info, NULL);
+    rc = mutt_pager(banner, tempfile, do_color, info);
   }
   else
   {
--- a/gui/curs_lib.c
+++ b/gui/curs_lib.c
@@ -731,7 +731,7 @@ int mutt_do_pager(const char *banner, const char *tempfile, PagerFlags do_color,
 
   if (!C_Pager || mutt_str_equal(C_Pager, "builtin"))
   {
-    rc = mutt_pager(banner, tempfile, do_color, info);
+    rc = mutt_pager(banner, tempfile, do_color, info, NULL);
   }
   else
   {
--- a/mutt_menu.h
+++ b/mutt_menu.h
@@ -118,7 +118,7 @@ struct Menu
    * custom_redraw - Redraw the menu
    * @param menu Menu to redraw
    */
-  void (*custom_redraw)(struct Menu *menu);
+  void (*custom_redraw)(struct Menu *menu, struct Mailbox *m, int msg_in_pager);
 
   void *redraw_data;
 };
--- a/mutt_menu.h
+++ b/mutt_menu.h
@@ -118,7 +118,7 @@ struct Menu
    * custom_redraw - Redraw the menu
    * @param menu Menu to redraw
    */
-  void (*custom_redraw)(struct Menu *menu, struct Mailbox *m, int msg_in_pager);
+  void (*custom_redraw)(struct Menu *menu);
 
   void *redraw_data;
 };
--- a/pager.c
+++ b/pager.c
@@ -2023,7 +2023,7 @@ static void pager_custom_redraw(struct Menu *pager_menu)
 
   struct Mailbox *m = NULL;
   int msg_in_pager = -1;
-  if (rd->extra && rd->extra->ctx) 
+  if (rd->extra && rd->extra->ctx)
   {
     m = rd->extra->ctx->mailbox;
     msg_in_pager = rd->extra->ctx->msg_in_pager;
@@ -2293,8 +2293,8 @@ int mutt_pager(const char *banner, const char *fname, PagerFlags flags, struct P
     flags |= MUTT_SHOWFLAT;
 
   int index_space = C_PagerIndexLines;
- 
-  if (extra->ctx && extra->ctx->mailbox) 
+
+  if (extra->ctx && extra->ctx->mailbox)
   {
     index_space = MIN(index_space, extra->ctx->mailbox->vcount);
     m = extra->ctx->mailbox;
@@ -3268,7 +3268,7 @@ int mutt_pager(const char *banner, const char *fname, PagerFlags flags, struct P
         break;
 
       case OP_MAIL:
-        CHECK_MODE(IsEmail(extra) && !IsAttach(extra));
+        CHECK_MODE(IsEmail(extra));
         CHECK_ATTACH;
         mutt_send_message(SEND_NO_FLAGS, NULL, NULL, extra->ctx, NULL, NeoMutt->sub);
         pager_menu->redraw = REDRAW_FULL;
@@ -3276,7 +3276,7 @@ int mutt_pager(const char *banner, const char *fname, PagerFlags flags, struct P
 
 #ifdef USE_NNTP
       case OP_POST:
-        CHECK_MODE(IsEmail(extra) && !IsAttach(extra));
+        CHECK_MODE(IsEmail(extra));
         CHECK_ATTACH;
         if (extra->ctx && (extra->ctx->mailbox->type == MUTT_NNTP) &&
             !((struct NntpMboxData *) extra->ctx->mailbox->mdata)->allowed && (query_quadoption(C_PostModerated, _("Posting to this group not allowed, may be moderated. Continue?")) != MUTT_YES))
@@ -3375,7 +3375,7 @@ int mutt_pager(const char *banner, const char *fname, PagerFlags flags, struct P
 
       case OP_RECALL_MESSAGE:
       {
-        CHECK_MODE(IsEmail(extra) && !IsAttach(extra));
+        CHECK_MODE(IsEmail(extra));
         CHECK_ATTACH;
         struct EmailList el = STAILQ_HEAD_INITIALIZER(el);
         emaillist_add_email(&el, extra->email);
--- a/pager.c
+++ b/pager.c
@@ -2021,8 +2021,13 @@ static void pager_custom_redraw(struct Menu *pager_menu)
   if (!rd)
     return;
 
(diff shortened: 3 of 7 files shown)


ngortheone:
This builds and works, surprisingly to me :)

But there is one place where I have no confidence that I did the right thing

gui/curs_lib.c
```diff
@@ -728,8 +728,9 @@ int mutt_do_pager(const char *banner, const char *tempfile, PagerFlags do_color,
 
   int rc;
 
-  if (!C_Pager || mutt_str_equal(C_Pager, "builtin"))
-    rc = mutt_pager(banner, tempfile, do_color, info);
+  if (!C_Pager || mutt_str_equal(C_Pager, "builtin")) {
+    rc = mutt_pager(banner, tempfile, do_color, info, NULL);
```
ngortheone, line 732: `mutt_pager()` seems to account for mailbox to be `NULL` but this still feels hacky.

I don't know how to test this code path.

flatcap: This shows all the routes into the pager:

mutt_pager

I don't think a `Context` make any sense for them.

ngortheone: Help certainly works, viewing messages too. So I guess I didn't break anything

pager.c
```diff
@@ -310,14 +308,21 @@ inline static bool assert_mailbox_writable(struct Mailbox *mailbox)
     break;                                                                     \
   }
 
-#define CHECK_ACL(aclbit, action)                                              \
-  if (!Contex2 || !(Contex2->mailbox->rights & aclbit))                        \
-  {                                                                            \
-    mutt_flushinp();                                                           \
-    /* L10N: %s is one of the CHECK_ACL entries below. */                      \
-    mutt_error(_("%s: Operation not permitted by ACL"), action);               \
-    break;                                                                     \
```
flatcap, line 311: Please add a Doxygen comment block for the function.

pager.c
```diff
@@ -286,13 +287,20 @@ static const struct Mapping PagerNewsHelp[] = {
...
   }
 
-#define CHECK_READONLY                                                         \
-  if (!Contex2 || Contex2->mailbox->readonly)                                  \
-  {                                                                            \
-    mutt_flushinp();                                                           \
-    mutt_error(_(Mailbox_is_read_only));                                       \
-    break;                                                                     \
-  }
+
```
flatcap: Please add a Doxygen comment block for the function.

pager.c
```diff
@@ -286,13 +287,20 @@ static const struct Mapping PagerNewsHelp[] = {
...
+inline static bool assert_mailbox_writable(struct Mailbox *mailbox)
```
flatcap, line 290: Do we really need `assert` in the function name?
Would `is_mailbox_writable()` do?

pager.c
```diff
@@ -286,13 +287,20 @@ static const struct Mapping PagerNewsHelp[] = {
...
+{
+  assert(mailbox);
+  if (mailbox->readonly)
+  {
+    mutt_flushinp();
+    mutt_error(_(Mailbox_is_read_only));
+    return false;
+  } 
+  else
+    return true;
```
flatcap, line 300: We can drop the `else` here (slightly flattening the code).
If the `if` statement is true, the function ends.

pager.c
```diff
@@ -310,14 +308,21 @@ inline static bool assert_mailbox_writable(struct Mailbox *mailbox)
...
-    mutt_error(_("%s: Operation not permitted by ACL"), action);               \
-    break;                                                                     \
+inline static bool assert_mailbox_permissions(struct Mailbox *m, AclFlags acl, char* action)
+{
+  assert(m);
+  if (m->rights & acl) 
+  {
+      return true;
+  }
+  else
```
flatcap, line 318: We can drop the `else` here (slightly flattening the code).
If the `if` statement is true, the function ends.

pager.c
```diff
@@ -3067,7 +3075,7 @@ int mutt_pager(const char *banner, const char *fname, PagerFlags flags, struct P
       case OP_PURGE_MESSAGE:
       case OP_DELETE:
         CHECK_MODE(IsEmail(extra));
-        CHECK_READONLY;
+        if (!assert_mailbox_writable(Contex2->mailbox)) break;
```
flatcap: You asked about these `break`s...
The look OK, but `clang-format` will wrap them.

pager.c
```diff
@@ -310,14 +308,21 @@ inline static bool assert_mailbox_writable(struct Mailbox *mailbox)
...
-#define CHECK_ACL(aclbit, action)                                              \
-  if (!Contex2 || !(Contex2->mailbox->rights & aclbit))                        \
-  {                                                                            \
-    mutt_flushinp();                                                           \
-    /* L10N: %s is one of the CHECK_ACL entries below. */                      \
-    mutt_error(_("%s: Operation not permitted by ACL"), action);               \
-    break;                                                                     \
+inline static bool assert_mailbox_permissions(struct Mailbox *m, AclFlags acl, char* action)
+{
+  assert(m);
```
flatcap, line 313: Should we `assert` on `action` too?
(it's dereferenced by `mutt_error()`)

pager.c
```diff
@@ -3077,7 +3082,7 @@ int mutt_pager(const char *banner, const char *fname, PagerFlags flags, struct P
         CHECK_MODE(IsEmail(extra));
         if (!assert_mailbox_writable(Contex2->mailbox)) break;
         /* L10N: CHECK_ACL */
-        CHECK_ACL(MUTT_ACL_DELETE, _("Can't delete message"));
+        if (!assert_mailbox_permissions(Contex2->mailbox, MUTT_ACL_DELETE, _("Can't delete message"))) break;
```
flatcap: When I ran `clang-format`, some of these statements ran over three lines, e.g.
```c
if (really_long_condition &&
    something_else)
  break;
```

If this happens, then the `break` needs `{}`s for clarity.

```c
if (really_long_condition &&
    something_else)
{
  break;
}
```

(or we can try to abbreviate the name of the helper function)

mutt_menu.h
```diff
@@ -118,7 +118,7 @@ struct Menu
    * custom_redraw - Redraw the menu
    * @param menu Menu to redraw
    */
-  void (*custom_redraw)(struct Menu *menu);
+  void (*custom_redraw)(struct Menu *menu, struct Mailbox *m, int msg_in_pager);
```
flatcap, line 121: Hmm...  This is tricky.
We don't want the `Menu` to know any more than it needs to, so I'm very suspicious of `msg_in_pager`.
Can we put this info in `Menu->redraw_data`, making it private to the Pager?

Another problem is that both Index and Compose use `custom_redraw()`, but you haven't altered their callback functions!

pager.c
ngortheone: done

pager.c
ngortheone: done

pager.c
ngortheone: done

pager.c
ngortheone: done

flatcap:
I've done a bit of juggling...

- Create branch [`[devel/refactor]`](https://github.com/neomutt/neomutt/tree/devel/refactor) from my unmerged refactoring work
- Merge your macro-expansion commits into `[devel/refactor]`
- Rebase your remaining work over `[devel/refactor]`
- Set the base of the PR to `[devel/refactor]`

flatcap (changes requested):
I've attached comments to all the relevant places in the code.
If read in the correct order, they should be helpful :-)

ngortheone:
thanks @flatcap ! Tomorrow I might actually have some time work on this!

pager.c
```diff
@@ -2265,27 +2271,35 @@ static void pager_custom_redraw(struct Menu *pager_menu)
...
     flags |= MUTT_SHOWFLAT;
 
   int index_space = C_PagerIndexLines;
-  if (extra->ctx && extra->ctx->mailbox)
+ 
+  // Do we even consire an option when pager is invoked without mailbox?
+  assert(extra);
+  assert(extra->ctx);
+  assert(extra->ctx->mailbox);
+  if (extra->ctx && extra->ctx->mailbox)
```
ngortheone, line 2298: @flatcap can what do you think about the above?

ngortheone: this is apparently not a good idea. This breaks helper since mailbox is NULL in that case. I will remove the last two asserts

pager.c
```diff
@@ -2265,27 +2271,35 @@ static void pager_custom_redraw(struct Menu *pager_menu)
  */
 int mutt_pager(const char *banner, const char *fname, PagerFlags flags, struct Pager *extra)
 {
+
+  struct Mailbox *m;
```
flatcap, line 2275: Please init all pointers to `NULL`

pager.c
```diff
@@ -2265,27 +2271,35 @@ static void pager_custom_redraw(struct Menu *pager_menu)
...
   static char searchbuf[256] = { 0 };
   char buf[1024];
   int ch = 0, rc = -1;
   bool first = true;
   int searchctx = 0;
   bool wrapped = false;
-
   struct Menu *pager_menu = NULL;
+  int msg_in_pager = 0;
```
flatcap, line 2283: This has become orphaned.
It's set in a few places, but not read.

Rather than using this local, I think we should be using `Context->msg_in_pager`.
There are three places where it's set/used.

flatcap (changes requested):
Just a quick review; we're getting closer now.
There are a couple of `Context`s left.
One is:
```c
ctx_free(&Contex2);
```
You removed this line.
This feels wrong, but I haven't investigated it yet.

The other is:
```c
mutt_update_index(..., Contex2, ...);
```

(GitHub won't allow me to review code that you haven't changed :-)

ngortheone:
`mutt_update_index(..., Contex2, ...);`
Leads down the rabbit hole. I think that one deserves a separate PR, and discussion. I don't think I am prepared to deal with it right now.

ngortheone:
@flatcap 
This still looks ugly, but I think we are getting closer. 
I think this PR got a bit out of hand, lets merge what can be merged, and split the rest into a more scoped PRs.
And we should discuss tactics about how to attack `mutt_update_index(..., Contex2, ...);`

PR #2813, opened 2021-02-11
Renamed config names

paniash:
* **What does this PR do?**
Rename config names.


flatcap:
That was quick!
That gives us the **new** names, but we still want to support the old ones.
For each old name, we need a synonym definition, like this:

https://github.com/neomutt/neomutt/blob/aa2b0c5810065ad4701da6b91bc85aaf1454acf0/mutt_config.c#L760

conn/config.c
```diff
@@ -131,7 +131,7 @@ struct ConfigDef ConnVars[] = {
     "(ssl) Use TLSv1.3 for authentication"
   },
 #ifdef USE_SSL_OPENSSL
-  { "ssl_usesystemcerts", DT_BOOL, &C_SslUsesystemcerts, true, 0, NULL,
+  { "ssl_use_system_certs", DT_BOOL, &C_SslUsesystemcerts, true, 0, NULL,
```
gahr, line 134: `C_SslUseSystemCerts` too, while at it?

contrib/samples/sample.neomuttrc-tlr
```diff
@@ -194,12 +194,12 @@ set   abort_unmodified=no               # Let me send empty messages
 set   alias_file=~/.mutt/aliases        # Where to store aliases
 unset allow_8bit                        # Produce correct MIME
 unset arrow_cursor                      # Use the bar cursor
-set   askcc                             # Ask me about CCs
+set   ask_cc                             # Ask me about CCs
```
gahr, line 197: perhaps you should try to keep the `#` aligned

paniash:
> That was quick!

Good 'ol `sed` to the rescue. ;-)

> That gives us the **new** names, but we still want to support the old ones.
> For each old name, we need a synonym definition, like this:
> 
> https://github.com/neomutt/neomutt/blob/aa2b0c5810065ad4701da6b91bc85aaf1454acf0/mutt_config.c#L760

If I'm understanding this right, the format is something along the lines of:

```c
{ "old_name",        DT_SYNONYM, NULL, IP "new_name", },
```

Is this right?

conn/config.c
paniash: Done!

contrib/samples/sample.neomuttrc-tlr
paniash: Apologies for that! It's resolved now. :-)

flatcap:
> ```c
> { "old_name",        DT_SYNONYM, NULL, IP "new_name", },
> ```
> 
> Is this right?

Yep, that's right

paniash:
@flatcap Looks like the formatting is out of place now. Looks fine on my editor but not very much on github. Any idea? (I'm using spaces and not tabs).

flatcap:
It's going to take a little while to review this.

A couple of things for now.
Please can you:
- revert the change to `autosetup/jimsh0.c` -- that's a third-party source file
- split up the list of synonyms, by the location of the real config name.
  It's important that the synonyms are in the same file, so they are enabled/disabled along with the **real** config item.

Thanks

paniash:
> Please can you:
> 
>     * revert the change to `autosetup/jimsh0.c` -- that's a third-party source file
> 
>     * split up the list of synonyms, by the location of the real config name.
>       It's important that the synonyms are in the same file, so they are enabled/disabled along with the **real** config item.

Done! Let me know if anything else is needed. btw any idea why all the FreeBSD related checks keep failing? 

Thanks!

ncrypt/config.c
```diff
@@ -348,10 +348,14 @@ struct ConfigDef NcryptVars[] = {
     "Save the cleartext Subject with the headers"
   },
 
-  { "pgp_create_traditional", DT_SYNONYM, NULL, IP "pgp_autoinline",    },
-  { "pgp_self_encrypt_as",    DT_SYNONYM, NULL, IP "pgp_default_key",   },
-  { "pgp_verify_sig",         DT_SYNONYM, NULL, IP "crypt_verify_sig",  },
-  { "smime_self_encrypt_as",  DT_SYNONYM, NULL, IP "smime_default_key", },
+  { "crypt_confirmhook",      DT_SYNONYM, NULL, IP "crypt_confirm_hook", },
+  { "pgp_autoinline",	      DT_SYNONYM, NULL, IP "pgp_auto_inline",    },
+  { "pgp_clearsign_command",  DT_SYNONYM, NULL, IP "pgp_clear_sign_command",     },
```
flatcap, line 353:
```suggestion
#ifdef CRYPT_BACKEND_CLASSIC_PGP
  { "pgp_clearsign_command",  DT_SYNONYM, NULL, IP "pgp_clear_sign_command",     },
#endif
```

ncrypt/config.c
```diff
@@ -348,10 +348,14 @@ struct ConfigDef NcryptVars[] = {
...
+  { "pgp_create_traditional", DT_SYNONYM, NULL, IP "pgp_auto_inline",    },
+  { "pgp_getkeys_command",	  DT_SYNONYM, NULL, IP "pgp_get_keys_command",       },
```
flatcap, line 355:
```suggestion
#ifdef CRYPT_BACKEND_CLASSIC_PGP
  { "pgp_getkeys_command",	  DT_SYNONYM, NULL, IP "pgp_get_keys_command",       },
#endif
```

flatcap (changes requested):

conn/config.c
```diff
@@ -131,7 +131,7 @@ struct ConfigDef ConnVars[] = {
     "(ssl) Use TLSv1.3 for authentication"
   },
 #ifdef USE_SSL_OPENSSL
-  { "ssl_usesystemcerts", DT_BOOL, &C_SslUsesystemcerts, true, 0, NULL,
+  { "ssl_use_system_certs", DT_BOOL, &C_SslUseSystemCerts, true, 0, NULL,
```
flatcap, line 134: You've renamed the global variable here: `C_SslUseSystemCerts`, but it exists in a few other places!
Build `./configure --ssl` fails.

paniash: Done!

paniash:
Dumb me apparently, erased the contents instead of changing tabs to spaces. Sorry about that, working on fixing it.

flatcap:
> What's wrong with this?
> 
> ```shell
> expand -t 4 filename > filename
> ```

Ah, that's a Unix thing.
The redirection means that the first thing the shell does is truncate `filename` to zero length.

You could use a temporary file, or you could install the `moreutils` package (name may vary with distro).

```shell
expand -t 4 filename | sponge filename
```

The util `sponge` magically allows you to overwrite a file, for tools that can't work in-place.

paniash:
> > What's wrong with this?
> > ```shell
> > expand -t 4 filename > filename
> > ```
> 
> Ah, that's a Unix thing.
> The redirection means that the first thing the shell does is truncate `filename` to zero length.

I thought this simply overwrites the contents of the file with the standard output. :/


> ```shell
> expand -t 4 filename | sponge filename
> ```
> 
> The util `sponge` magically allows you to overwrite a file, for tools that can't work in-place.

Great! Worked like a charm (hopefully).

paniash:
@flatcap I only need to change spaces back into tabs for all the Makefiles right?

flatcap:
> @flatcap I only need to change spaces back into tabs for all the Makefiles right?

`Makefile*` and `mutt/queue.h`

flatcap (approved):

PR #2818, opened 2021-02-13
notmuch: allow nm_record in non-notmuch mailboxes

### LINKED ISSUE #1867: nm_record and nm_record_tags
Hi,

I am trying to get neomutt to add sent e-mail to the notmuch database but so far I am quite unsuccessful. If I understand the relevant documentation correctly, this should be achieved for example by the following settings:

```
set record = $MAIL/sent
set nm_record = yes
set nm_record_tags = sent
```

After sending a mail, I can indeed see it being appended to the file `$MAIL/sent` however executing the command `notmuch search tag:sent` does not show the message. In fact the message is not present in the notmuch database at all.
Is there something I am misunderstanding or did not consider?

I am using:
  notmuch version 0.29.1,
  current neomutt-git build from the AUR

Austin-Ray:
The 'nm_record' functionality doesn't work in non-notmuch mailboxes
because 'nm_record_message()' requires notmuch-specific account data.
Other mailbox types lack the data so 'nm_record_message()' fails.
NeoMutt's documentation does not limit this functionality to notmuch
mailboxes.

To improve behavior consistency and user experience, modify
'nm_record_message()' to access a default notmuch mailbox.
When a user sends an email from a non-notmuch mailbox, the notmuch
backend uses the default mailbox for database access to append the
message.

The backend does not FREE the default mailbox for two reasons:
1) The user may define it in their configuration.
2) For future calls to prevent repeated allocations/deallocations.

The code contains inline warnings repeating the above rationale for
future modifications to 'nm_record_message()'

Fixes #1867

### DIFF
--- a/notmuch/notmuch.c
+++ b/notmuch/notmuch.c
@@ -287,6 +287,34 @@ struct NmEmailData *nm_edata_get(struct Email *e)
 }
 
 /**
+ * nm_get_default_url - Get a default Notmuch URL
+ * @retval Notmuch URL
+ * @retval NULL Error, unable to create default URL.
+ *
+ * @note Caller must free the returned string
+ */
+static char *nm_get_default_url(void)
+{
+  // path to DB + query + url "decoration"
+  size_t len = PATH_MAX + 1024 + 32;
+  char *url = mutt_mem_malloc(len);
+
+  // Try to use `$nm_default_url` or `$folder`.
+  // If neither are set, it is impossible to create a Notmuch URL.
+  if (C_NmDefaultUrl)
+    snprintf(url, len, "%s", C_NmDefaultUrl);
+  else if (C_Folder)
+    snprintf(url, len, "notmuch://%s", C_Folder);
+  else
+  {
+    FREE(url);
+    return NULL;
+  }
+
+  return url;
+}
+
+/**
  * nm_get_default_data - Create a Mailbox with default Notmuch settings
  * @retval ptr  Mailbox with default Notmuch settings
  * @retval NULL Error, it's impossible to create an NmMboxData
@@ -294,18 +322,14 @@ struct NmEmailData *nm_edata_get(struct Email *e)
 static struct NmMboxData *nm_get_default_data(void)
 {
   // path to DB + query + url "decoration"
-  char url[PATH_MAX + 1024 + 32];
-
-  // Try to use `$nm_default_url` or `$folder`.
-  // If neither are set, it is impossible to create a Notmuch URL.
-  if (C_NmDefaultUrl)
-    snprintf(url, sizeof(url), "%s", C_NmDefaultUrl);
-  else if (C_Folder)
-    snprintf(url, sizeof(url), "notmuch://%s", C_Folder);
-  else
+  char *url = nm_get_default_url();
+  if (!url)
     return NULL;
 
-  return nm_mdata_new(url);
+  struct NmMboxData *default_data = nm_mdata_new(url);
+  FREE(&url);
+
+  return default_data;
 }
 
 /**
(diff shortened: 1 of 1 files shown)


flatcap:
@Austin-Ray 
Hello stranger ;-)

@neomutt/notmuch-users
Please can I have a second opinion from some Notmuch users.

Thanks

jindraj:
> @neomutt/notmuch-users
> Please can I have a second opinion from some Notmuch users.
I don't use `nm_record` because GMail automatically puts sent mail into [GMail]/Sent and it's fetched and labeled in next sync. Hopefully someone else uses it and will shed some light this value of this PR.

Austin-Ray:
I had hoped someone else would chime, but since no one else has.

I've been using `nm_record` as having NeoMutt add to the notmuch database when sending is more convenient than sending + pulling from mail provider or an external call to `notmuch new`. It's a better user experience for my workflow.

In addition, for new users, our documentation doesn't limit `nm_record` to notmuch mailboxes so to improve their experience, we should either enable functionality in non-notmuch mailboxes or update the docs calling out the limitation. 

Limiting to only notmuch mailboxes hampers our notmuch users who use its search integration, but not the `virtual-mailboxes`

notmuch/notmuch.c
```diff
@@ -287,22 +287,41 @@ struct NmEmailData *nm_edata_get(struct Email *e)
...
 /**
- * nm_get_default_data - Create a Mailbox with default Notmuch settings
- * @retval ptr  Mailbox with default Notmuch settings
- * @retval NULL Error, it's impossible to create an NmMboxData
+ * nm_get_default_url - Get a default Notmuch URL
+ * @retval Notmuch URL
+ * @retval NULL Error, unable to create default URL.
  */
-static struct NmMboxData *nm_get_default_data(void)
+static char *nm_get_default_url()
```
flatcap, line 294: In C, `()` doesn't mean `(void)`

notmuch/notmuch.c
```diff
@@ -287,22 +287,41 @@ struct NmEmailData *nm_edata_get(struct Email *e)
...
 {
   // path to DB + query + url "decoration"
-  char url[PATH_MAX + 1024 + 32];
+  size_t len = PATH_MAX + 1024 + 32;
+  char *url = mutt_mem_malloc(len);
```
flatcap, line 298: We `alloc` the string here...

notmuch/notmuch.c
```diff
@@ -287,22 +287,41 @@ struct NmEmailData *nm_edata_get(struct Email *e)
...
+ * @retval NULL Error, it's impossible to create an NmMboxData
+ */
+static struct NmMboxData *nm_get_default_data(void)
+{
+  // path to DB + query + url "decoration"
+  char *url = nm_get_default_url();
+  if (!url)
     return NULL;
 
   return nm_mdata_new(url);
```
flatcap, line 327: ... but nobody frees it.

notmuch/notmuch.c
```diff
@@ -2065,6 +2084,23 @@ static enum MxStatus nm_mbox_check_stats(struct Mailbox *m, uint8_t flags)
...
 
+/**
+ * get_default_mailbox - Get Mailbox for notmuch without any parameters.
+ * @retval ptr Mailbox pointer.
+ */
+static struct Mailbox *get_default_mailbox()
+{
+    // Create a new notmuch mailbox from scratch and add plumbing for DB access.
+    char *default_url = nm_get_default_url();
+    struct Mailbox *m = mx_path_resolve(default_url);
```
flatcap, line 2095: ... nor here.

notmuch/notmuch.c
```diff
@@ -2065,6 +2084,23 @@ static enum MxStatus nm_mbox_check_stats(struct Mailbox *m, uint8_t flags)
   return rc;
 }
 
+/**
+ * get_default_mailbox - Get Mailbox for notmuch without any parameters.
+ * @retval ptr Mailbox pointer.
+ */
+static struct Mailbox *get_default_mailbox()
```
flatcap, line 2091: `(void)`

flatcap (changes requested):

notmuch/notmuch.c
Austin-Ray: Fixed

notmuch/notmuch.c
Austin-Ray: Fixed

flatcap (approved):
- clang-format'd
- rebased
- squashed

PR #2829, opened 2021-02-27
 pager: remove CHECK_ATTACH macro

ngortheone:
CHECK_ATTACH macro was obscuring the execution flow by hiding break
keyword;

New function that replaces the macro is better named and does not
obscure the flow;

### DIFF
--- a/pager.c
+++ b/pager.c
@@ -355,13 +355,24 @@ static inline bool assert_mailbox_writable(struct Mailbox *mailbox)
   return true;
 }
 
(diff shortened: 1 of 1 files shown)


flatcap:
I suggest inverting the logic (to reduce the mental load):

Rather than:
```c
if (!assert_not_attach_msg_mode(OptAttachMsg))
  break;
```

Drop the double-negative:

```c
if (assert_attach_msg_mode(OptAttachMsg))
  break;
```

ngortheone:
fixed

PR #2833, opened 2021-03-01
pager: check_sig minor improvements

ngortheone:
* **What does this PR do?**

Minor readability and type improvements

### DIFF
--- a/pager.c
+++ b/pager.c
@@ -325,7 +325,6 @@ static const struct Mapping PagerNewsHelp[] = {
   (pager && (pager)->fp && (pager)->body && (pager)->body->email)
 #define IsEmail(pager) (pager && (pager)->email && !(pager)->body)
 
-#define NUM_SIG_LINES 4
 
 /**
  * assert_pager_mode - Check that pager is in correct mode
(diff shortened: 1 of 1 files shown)


pager.c
```diff
@@ -203,6 +203,10 @@ struct PagerRedrawData
 
 /* hack to return to position when returning from index to same message */
 static int TopLine = 0;
+
+
+/* The amount of lines a signature takes */
+static const unsigned int NUM_SIG_LINES = 4;
```
gahr, line 209: Can we make it local to the function that uses it?

ngortheone: done

gahr (approved):
LGTM - I am curious on the choice of unsigned int, though. We're probably fine with `int` if you just need an integral type, or `size_t` if you really need a counter.
There's been discussions lately on whether we should use `int` or `size_t` for loop indices and counters that we know cannot be negative. There are pros and cons to each, and people have pretty strong opinions (e.g., only use `unsigned` if you want modulo airthmetics). What's your rationale for `unsigned int` here?

ngortheone:
@gahr `size_t` is probably a good choice. Don't forget that I am a homegrown C hacked with a few days of experience

PR #2834, opened 2021-03-01
pager: libpager

ngortheone:
* **What does this PR do?**

Moves pager to a separate directory. Creates libpager


flatcap:
Hmm... Sorry, `display_line()` is a nightmare.
Thanks for investigating, but I don't think this is going anywhere yet.

--- 

There are a few globals:
- `AttachmentMarker` in `check_attachment_marker()`
- `ProtectedHeaderMarker` in `check_protected_header_marker()`
- `Colors` in 7 functions

Config is used in 4 functions.

My quick analysis looked like this:


Source: https://gist.github.com/flatcap/3f5f474c1bff666bfd54d7fed3702766

ngortheone:
Marching orders:

```
flatcap | 1) like the other libraries, move Pager's config from mutt_config.c to pager/config.c
flatcap | 2) then eliminate the C_* variables, by using the ConfigSet helper functions
flatcap | the end of this page gives a little more detail: https://neomutt.org/dev/config
```

mutt_config.c
```diff
@@ -817,6 +787,7 @@ static void init_variables(struct ConfigSet *cs)
   CONFIG_INIT_VARS(cs, helpbar);
   CONFIG_INIT_VARS(cs, history);
   CONFIG_INIT_VARS(cs, index);
+  CONFIG_INIT_VARS(cs, pager);
```
flatcap, line 790: For all the OCD sufferers out there -- this list is alphabetically sorted :-)

pager/lib.h
```diff
@@ -81,4 +81,6 @@ void mutt_buffer_strip_formatting(struct Buffer *dest, const char *src, bool str
 
 void mutt_clear_pager_position(void);
 
+bool config_init_pager(struct ConfigSet *cs);
```
flatcap, line 84: Actually, this isn't needed.
The macro in `mutt_config.c` declares the function for you:
https://github.com/neomutt/neomutt/blob/8ff8cd7f5e5f773c257522218583e9d36324c237/mutt_config.c#L71-L73

pager/lib.h
```diff
@@ -81,4 +81,6 @@ void mutt_buffer_strip_formatting(struct Buffer *dest, const char *src, bool str
...
+
 #endif /* MUTT_PAGER_H */
```
flatcap, line 86: Please can you tidy up a few things I forgot to do in my libpager commit:

- Fix the `#include` guard name to match the file: `MUTT_PAGER_LIB_H`

- Update `po/POTFILES.in` to match the new files/locations
  Here's a [one-liner command](https://neomutt.org/run/release/translate#popotfilesin) to do it

- Add a Doxygen library header block
  Here's the Sidebar's:

https://github.com/neomutt/neomutt/blob/8ff8cd7f5e5f773c257522218583e9d36324c237/sidebar/lib.h#L25-L40

flatcap (changes requested):
looking good

pager/lib.h
ngortheone: I was mimicking `index/lib.h`

flatcap: Hmm... a quick grep shows 17 `config_init_*()` functions.
We don't need any of them.  Please can you delete them.

flatcap:
Next task: Please can you have a look at this

## Eliminate `static int TopLine`

Some functions of the Pager are actually performed by the Index.
The Pager closed, the Index performs the function and the Pager is re-opened.
Hopefully the use doesn't notice.

```
/* hack to return to position when returning from index to same message */
static int TopLine = 0;
```

This is needs to die.

One solution might be to make it part of the `struct PagerRedrawData`
The Index could create this and pass it to the Pager.
That way, it would persist across multiple calls.

---

A simple test is to open a long email in the pager, then `<toggle-write>` (<kbd>%</kbd>).
This is performed by the Index (and causes the Index Bar (status) to update).

pager/lib.h
ngortheone: I removed declarations. Do we need to remove definitions too?

pager/lib.h
ngortheone: done

mutt_config.c
ngortheone: fixed. But how do OCD people feel that line 777 is `config_init_main(cs);` and not `CONFIG_INIT_VARS(cs, main);`
That breaks the pattern! :open_mouth: :smile:

ngortheone:
@flatcap 
All your comments addressed.
I think we should merge this one and split TopLine into a separate PR.
And can you save that graph with a legend somewhere? I'd like to find it again !

flatcap (approved):

flatcap:
> All your comments addressed

Thanks, merged.

> split TopLine into a separate PR.

yep

> And can you save that graph with a legend somewhere? I'd like to find it again !

https://gist.github.com/flatcap/3f5f474c1bff666bfd54d7fed3702766
(includes the graphviz source)

PR #2843, opened 2021-03-08
test(notmuch): test string_to_query_type

Austin-Ray:
Create a unit test for the 'string_to_query_type()' function.

To do this, refactor it into a new 'util.c' file so it doesn't contain
any dependencies on other backends or global state, which complicates
test compilations.  Next, write a unit test covering all three outcomes
of the function. Last, update the CI pipeline to compile with notmuch
support.

This marks the first step towards unit testing the notmuch backend.
While its scope is only on utility functions, having confidence in those
allows for writing high-level code in a more confident manner.


flatcap (commented):
I've pushed some small fixes:
- missing `#include`s
- clang-format
- cut'n'paste typos
- doxygen headers
- add new file to `po/POTFILES.in`

flatcap (commented):
Now you've exposed `string_to_query_type()`,
how would you feel about renaming it to `nm_string_to_query_type()`?

gahr:
Thanks for the test, I think it's a good step forward, but can we please talk about

> To do this, refactor it into a new 'util.c' file so it doesn't contain
any dependencies on other backends or global state, which complicates
test compilations.

I don't buy this argument: the test suite is linked against `$(MUTTLIBS)` (which includes our `libnotmuch.a`) and `$(LIBS)` (which includes upstream's `-lnotmuch`) when the NotMuch feature is enabled anyway.
Moving the function to a dedicated `util.c` file brings you absolutely nothing in terms of (in)dependency from global data or state. Quite the opposite, it gives you the illusion of indipendene or stand-alone-ness, which it doesn't have.
Plus, sorry, util is a very bad name: the only information it conveys is "dunno where this belongs, let me dump it in here".

I propose to move the function declaration to `notmuch/lib.h` and rename it as @flatcap suggests.

Thanks!

Austin-Ray:
@flatcap :
> Now you've exposed string_to_query_type(), how would you feel about renaming it to nm_string_to_query_type()?

Sure, but please note that this function is exposed for testing purposes only, and isn't intended for use outside the `notmuch/` module. It's the visible for testing pattern.

---

@gahr :

> I propose to move the function declaration to `notmuch/lib.h`

The function is only intended for internal use in within the `notmuch` module. It shouldn't be included as part of the backend's public API. Since we're following the pattern of public APIs in `lib.h` and private in other header files, let's pick a more accurate name. 

Would `notmuch/query.h` and `notmuch/query.c` be more accurate names?

> Can we rename notmuch.c -> private.c and push nm_string_to_query_type to (the new) notmuch.c file?

I'm not sure I understand the rationale for this change. `notmuch.c` contains the bulk of the implementation for the module's _public_ API. Renaming to `private.c` is a misnomer.

Furthermore, the name falls to the same argument you used against `util`.

flatcap:
> note that this function is exposed for testing purposes only

Yes, thanks.  I checked that :-)
There are some functions, usually callbacks, that are static, but still used outside of source file / component.
I think it's helpful to prefix the names in situations like that.

> > Would `notmuch/query.h` and `notmuch/query.c` be more accurate names?
> >
> LGTM.

Great.

---

Things for another day...

`notmuch.c` is BIG.
Please can you have a think about how we might split it.
On IRC, I wondered about splitting backends along Account/Mailbox/Email lines.
Suggestions welcome.

flatcap (approved):

PR #2846, opened 2021-03-09
Move MX APIs into core

gahr:


flatcap:
This may be optimistic, but can we remove `#include "mx.h"` from any of the backend source?

Austin-Ray:
> This may be optimistic, but can we remove #include "mx.h" from any of the backend source?

Looks possible for the `notmuch` backend based on a quick compile. I don't think I missed any `mx.h` statements.

flatcap (approved):

PR #2848, opened 2021-03-11
Greeting

vcheburkin:
* **Insert greetings at the beginning of the message**

Adaptation of my patch for mutt for neomut.
Insert greetings at the beginning of the message, based on formatted
string "greeting" setting in config:

set greeting="Hi, %v!\n"

make string like:

Hi, Viktor!

before attribution.

Sorry, this is my first experience with git and github.

### DIFF
--- a/docs/config.c
+++ b/docs/config.c
@@ -1440,6 +1440,23 @@
 ** "Franklin" to "Franklin, Steve".
 */
 
+{ "greeting", DT_STRING, 0 },
+/*
+** .pp
+** When set, this is the string that will precede every message as a
+** greeting phrase to the recipients.
+** .pp
+** "Format strings" are similar to the strings used in the "C"
+** function printf to format output (see the man page for more detail).
+** The following sequences are defined in NeoMutt:
+** .pp
+** .dl
+** .dt %n .dd Recipient's real name
+** .dt %u .dd User (login) name of recipient
+** .dt %v .dd First name of recipient
+** .de
+*/
+
 #ifdef USE_NNTP
 { "group_index_format", DT_STRING, "%4C %M%N %5s  %-45.45f %d" },
 /*
--- a/mutt_config.c
+++ b/mutt_config.c
@@ -349,6 +349,9 @@ struct ConfigDef MainVars[] = {
   { "gecos_mask", DT_REGEX, &C_GecosMask, IP "^[^,]*", 0, NULL,
     "Regex for parsing GECOS field of /etc/passwd"
   },
+  { "greeting", DT_STRING, &C_Greeting, 0, 0, NULL,
+    "Greeting string added to the top of all messages"
+  },
   { "header", DT_BOOL, &C_Header, false, 0, NULL,
     "Include the message headers in the reply email (Weed applies)"
   },
--- a/send/send.c
+++ b/send/send.c
@@ -86,6 +86,15 @@
 #endif
 
 /**
+ * struct GreetingInfo - Data passed to greeting_string()
+ */
+struct GreetingInfo
+{
+  struct Mailbox *mailbox;
+  struct Email *email;
+};
+
+/**
  * append_signature - Append a signature to an email
  * @param fp  File to write to
  * @param sub Config Subset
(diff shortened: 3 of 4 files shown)


flatcap:
Hi, Viktor!

Welcome to NeoMutt and thanks for the PR.

> this is my first experience with git and github.

Everything looks good :-)

> Insert greetings at the beginning of the message

That seems a nice and simple idea, but `$attribution` can be used to do that.

`$attribution` uses the same expandos (%-things) as your `$greeting` and it can contain `\n`s to insert newlines.

It's not quite as pretty as a separate config variable, but this would work:

```
set attribution = "Hi, %v\n\nOn %d, %n wrote:"
```

There are ~450 config variables already (the manual is a beast), so we'd prefer not to add any more if possible.
Sorry.

However, would you mind updating the docs for `$attribution` to suggest your greeting idea?
I'm sure lots of people would be interested -- we devs had never thought of it before :-)

https://github.com/neomutt/neomutt/blob/9ce71e41ad85cd01a883ad3fcaa0cf6730e9f556/docs/config.c#L294-L300

Thanks.

vcheburkin:
Hi, Richard!

$attribution is a good idea, but not if you write new message, not reply.

flatcap:
> not if you write new message

Hmm... fair point.
Let me think about that for a moment.

flatcap:
OK, you convinced us :-)

> > this is my first experience with git and github.
> 
> Everything looks good :-)

Ah, that was a polite lie.
Your code isn't actually attached to anything yet.

We need to call `greeting_string()` from somewhere in `mutt_send_message()`.
Just before here looks good:

https://github.com/neomutt/neomutt/blob/9ce71e41ad85cd01a883ad3fcaa0cf6730e9f556/send/send.c#L2314-L2319

We'll want some of the same checks: `(!(flags & (SEND_KEY | SEND_BATCH)) && c_editor)`

**Note:**
We're in the process of changing how the config variables are used.
Even though you've created `C_Greeting`, we'll read it by using a helper function: `cs_subset_string()`.

Next, some tidying.  Please can you:
- Move `greeting_string()` to `send/send.c`, as it's not used anywhere else
- Add a comment block to the function -- You can strip down [index_format_str()'s](https://github.com/neomutt/neomutt/blob/master/hdrline.c#L411)

Give it a test:
- Set the greeting
- Reply to an email
- Compose a new email
- Send an email from the command line -- greeting should be disabled

---

You've done well to get this far -- The code is still a complicated mess.
Feel free to ask lots of questions.

If you use IRC, come say 'hello'.  We're in `#neomutt` on `irc.freenode.net`

vcheburkin:
Ok.
I also noticed that not all the changes were in the PR - my mistake, I forgot that send.c is now in the send directory.

flatcap:
> send.c is now in the send directory

Yes, since our split from Mutt, we've been tidying/refactoring pretty heavily.

vcheburkin:
All done: tidying, use index_format_str(), not insert greetings in batch mode.

send/send.c
```diff
@@ -608,6 +617,85 @@ void mutt_make_attribution(struct Mailbox *m, struct Email *e, FILE *fp_out,
   fputc('\n', fp_out);
 }
 
+static const char *greeting_string(char *buf, size_t buflen, size_t col, int cols,
+                                   char op, const char *src, const char *prec,
+                                   const char *if_str, const char *else_str,
+                                   intptr_t data, MuttFormatFlags flags)
```
flatcap, line 623: Please can you add a Doxygen function comment.
You can strip down [index_format_str()'s](https://github.com/neomutt/neomutt/blob/master/hdrline.c#L411)

docs/config.c
```diff
@@ -1440,6 +1440,14 @@
 ** "Franklin" to "Franklin, Steve".
 */
 
+{ "greeting", DT_STRING, "" },
```
flatcap, line 1443: Empty strings are represented as `NULL` internally.

mutt_config.c
```diff
@@ -349,6 +349,9 @@ struct ConfigDef MainVars[] = {
   { "gecos_mask", DT_REGEX, &C_GecosMask, IP "^[^,]*", 0, NULL,
     "Regex for parsing GECOS field of /etc/passwd"
   },
+  { "greeting", DT_STRING, &C_Greeting, IP "", 0, NULL,
```
flatcap, line 352: `IP ""` -> `NULL` here too

send/send.c
```diff
@@ -608,6 +617,85 @@ void mutt_make_attribution(struct Mailbox *m, struct Email *e, FILE *fp_out,
...
+{
+  struct GreetingInfo *gi = (struct GreetingInfo *) data;
+  struct Envelope *env = gi->email->env;
+  char *p, buf2[256];
```
flatcap, line 627: Please can you split these two and initialise `p` to `NULL`.

send/send.c
```diff
@@ -608,6 +617,85 @@ void mutt_make_attribution(struct Mailbox *m, struct Email *e, FILE *fp_out,
...
+
+/**
+ * mutt_make_greeting - Add greetings string
+ * @param m      Mailbox
+ * @param e      Email
+ * @param fp_out File to write to
+ * @param sub    Config Subset
+ */
+void mutt_make_greeting(struct Mailbox *m, struct Email *e, FILE *fp_out,
+                        struct ConfigSubset *sub)
```
flatcap, line 683: This function can be `static` as it's not used anywhere else.

send/send.h
```diff
@@ -62,6 +62,7 @@ void            mutt_fix_reply_recipients(struct Envelope *env, struct ConfigSub
 void            mutt_forward_intro(struct Mailbox *m, struct Email *e, FILE *fp, struct ConfigSubset *sub);
 void            mutt_forward_trailer(struct Mailbox *m, struct Email *e, FILE *fp, struct ConfigSubset *sub);
 void            mutt_make_attribution(struct Mailbox *m, struct Email *e, FILE *fp_out, struct ConfigSubset *sub);
+void            mutt_make_greeting(struct Mailbox *m, struct Email *e, FILE *fp_out, struct ConfigSubset *sub);
```
flatcap, line 65: This can be removed as it's not used elsewhere.

flatcap (changes requested):
Argh!  What an idiot **I am**.
`index_format_str()` isn't any use, because some of the expandos use the 'from' address not the 'to' address.

I've restored your code and left comments for some very minor tidying.

Thanks!

send/send.c
vcheburkin: > Argh! What an idiot **I am**.
> `index_format_str()` isn't any use, because some of the expandos use the 'from' address not the 'to' address.

No-no-no, in tests everything works as expected.
Снимок экрана 2021-03-12 в 9 40 48

mutt_config.c
vcheburkin: Compiler show warning here, when I changed `IP ""` to `NULL`:
```
mutt_config.c:352:41: warning: initialization of ‘long int’ from ‘void *’ makes integer from pointer without a cast [-Wint-conversion]
   { "greeting", DT_STRING, &C_Greeting, NULL, 0, NULL,
                                         ^~~~
```

vcheburkin: `0` is better, but the meaning is lost, as it seems to me

send/send.c
vcheburkin: It might be logically more correct to leave greeting_string().

flatcap (approved):
Lovely, thanks!

flatcap:
I noticed that a couple of your commits were from a different address 'hostmaster'.
GitHub didn't recognise those as your commits.
You can register other addresses on your [profile page](https://github.com/settings/profile).

flatcap:
Hi @vcheburkin,

I sent you a couple of emails welcoming you to NeoMutt
and suggesting things you might be interested in helping with.

Did they arrive?

Cheers, @flatcap

PR #2104, opened 2020-01-30
refactor path handling

flatcap:
### About

Encapsulate Mailbox Paths:

```c
struct Path
{
  char *desc;      ///< Descriptive name    
  char *orig;      ///< User-entered path    
  char *canon;     ///< Canonical path    
  int type;        ///< Path type, enum MailboxType    
  PathFlags flags; ///< Flags describing what's known about the path    
};

```

Starting from an 'unknown' state, such as 'user entry', a Mailbox Path can undergo several transformations:

- **Resolve**
  Expand '+' to `$folder`, '!' to the spoolfile, `@alias` to a username
  This is performed by NeoMutt

- **Tidy**
  Light tidying of the path, e.g. fixing the case; removing duplicate path separators
  This is performed by the backend

- **Canon**
  Create the canonical path, e.g. follow symlinks; sort parameters
  This is performed by the backend

### Done

- mxapi: tidy stat(2) usage
  Add `bool MxOps.is_local` so NeoMutt knows when `stat(2)` is appropriate

- config: add path type
  Encapsulate a Mailbox 'path'
  
- test: add config/path
  Test coverage for 'path' config type

- libmutt: split tilde from canon
  Expand `~` before saving into a config variable
  This means we don't need to pre-process a variable before calling `fopen()`
  
- config: convert path variables
  Change 24 config variables to `DT_PATH` (using `Path`)
  Each variable is subtyped as either `DT_PATH_DIR` or `DT_PATH_FILE`

- mxapi: add new path functions
  Add stubs for all the new path functions
  Each backend now has a `path.c` file which shouldn't have any dependencies

### To Do

- Fill out the path functions (mostly cut'n'paste from the existing ones)
- Decide what a 'tidy' path means for each backend
- Decide what a 'canonical' path means for each backend
- Replace each `char` array, representing a Mailbox, with `Path`
- Strip out old MxOps path functions
- Rename new path functions
- Strip out redundant backend path functions

PR #2853, opened 2021-03-16
WIP:Add create maildir function

matthewhughes934:
First attempt at adding a create mailbox function for maildir. Some test are under https://github.com/neomutt/test-library/tree/create-maildir-tests.


maildir/maildir.c
```diff
@@ -1609,6 +1611,57 @@ static enum MailboxType maildir_path_probe(const char *path, const struct stat *
...
+struct Mailbox *maildir_create_mailbox(struct Account *a, const char *path, const char *name)
+{
+  if (maildir_create_maildir(path) != 0)
+    return NULL;
+  struct Mailbox *m = mailbox_new();
+
+  mutt_buffer_strcpy(&m->pathbuf, path);
+  m->realpath = mutt_str_dup(path);
+  m->name = mutt_str_dup(name);
+  m->type = MUTT_MAILDIR;
```
matthewhughes934, line 1660: I'm not entirely sure on all the setup we need here, should there also be `mx_ac_add(a, m);` (test-library didn't play nice with this one)

maildir/maildir.c
```diff
@@ -1609,6 +1611,57 @@ static enum MailboxType maildir_path_probe(const char *path, const struct stat *
...
+ */
+static int maildir_create_maildir(const char *path)
+{
+  char tmp[PATH_MAX];
+
+  for (int i = 0; i < mutt_array_size(maildir_subdirs); i++)
+  {
+    snprintf(tmp, sizeof(tmp), "%s/%s", path, maildir_subdirs[i]);
+
+    if (mutt_file_mkdir(tmp, S_IRWXU) != 0)
```
matthewhughes934, line 1629: Per our conversation on IRC: we probably want to prompt users in the case of `EEXIST` here. Maybe this is best done in `maildir_create_mailbox` with e.g. `maildir_path_probe`? To keep this function purely for creation

command_parse.c
```diff
@@ -942,17 +942,14 @@ enum CommandResult parse_mailboxes(struct Buffer *buf, struct Buffer *s,
...
+    if (new_account)
     {
       a = account_new(NULL, NeoMutt->sub);
       a->type = m->type;
-      new_account = true;
-    }
-
-    if (!new_account)
-    {
+    } else {
```
matthewhughes934, line 952: Ignore this, just some random refactoring I forgot to remove :facepalm:

maildir/maildir.c
```diff
@@ -1610,6 +1612,57 @@ static enum MailboxType maildir_path_probe(const char *path, const struct stat *
...
+ *  maildir_create_mailbox - Create a maildir mailbox - Implements - MxOps::mbox_create()
+ */
+struct Mailbox *maildir_create_mailbox(struct Account *a, const char *path, const char *name)
+{
+  if (maildir_create_maildir(path) != 0)
+    return NULL;
+  struct Mailbox *m = mailbox_new();
+
+  mutt_buffer_strcpy(&m->pathbuf, path);
+  m->realpath = mutt_str_dup(path);
```
matthewhughes934, line 1659: I'm not sure I understand the difference in `pathbuf` vs `realpath`?

Austin-Ray: > I'm not sure I understand the difference in pathbuf vs realpath?

I'm also confused about the difference. In `mailbox.h`, `pathbuf` is the only undocumented field and `realpath`'s documentation is its usage, not what it is.

Is the difference that one is fully-qualified and one isn't?

flatcap: Originally it was just `path`, a big char array.

Then `realpath` came along to store the 'canonical' path.
In local terms that means following symlinks, for network paths the full `proto://user@host.com/path`.

To avoid lots of large char arrays, upstream replaced `path` with `pathbuf` (`struct Buffer`).

The other thing in the mix, is the descriptive name, for `named-mailboxes`.

---

I spent several months working out a better scheme for the paths, see #2104
It documents how paths should be handled (it creates a Path API).
Unfortunately, I couldn't figure out how to fit it into the existing mailboxes without breaking things.
Now that we're busy breaking things, it might be time to revisit paths.

flatcap:
I've pushed a couple of commits to `test-library`.
The first allocates a full set of objects in `test_open_and_close()`.

The second adds a call to `dump_graphviz()` which dumps the objects as a `.gv` (dot) file.
You'll need to install graphviz and ImageMagick packages to view them, e.g.
```
display 12:34-index.gv
dot -Tsvg 12:34-index.gv > 12:34-index.svg
```

maildir/maildir.c
flatcap: Yes.  Ideally the 'business' decisions should be made by the front-end code.

From the Browser, we might see:
- User starts browser: `mutt_buffer_select_file()`
- User calls function `<create-mailbox>`
  - Prompt: "Enter name:"
- Browser calls `mx_mbox_create()`
- The MXAPI
  - selects backend based on the `Account`
  - calls `maildir_mbox_create()`
- The backend returns `EEXISTS`
- `mx_mbox_create()` relays the error back to the browser
- Browser tells the user "Folder 'X' already exists"

core/mxapi.h
```diff
@@ -361,6 +361,14 @@ struct MxOps
...
    */
   int (*path_is_empty)     (const char *path);
+
+  /**
+   * mbox_create - create a new mailbox
+   * @param a Account to create the mailbox on
+   * @param path
+   * @param name The name of the mailbox
+  */
+  struct Mailbox *(*mbox_create)       (struct Account *a, const char *path, const char *name);
```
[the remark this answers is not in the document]
matthewhughes934: > The Mailbox tells the Account where to create the new folder
>  (passing NULL would mean create a top-level folder)

Where do we get the path info in this case? Is there some path info in the Account object already? (at a glance I didn't see much useful `Account.adata` for maildir)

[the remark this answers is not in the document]
flatcap: Ah! You've got me there.
I'm thinking of my `account` command (not yet implemented).

We can probably use `$folder` as the root -- create `name` beneath it.
Hmm... but the backends shouldn't know about `$folder`.
I'll keep thinking.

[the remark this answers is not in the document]
flatcap: I think for now, we do need to pass the entire path (rather than the name) for Maildir.
We can ask the user for the name, then `if (m->type == MUTT_MAILDIR)` we can swap it for the full path.

Currently, the Maildir backend doesn't have its own AccountData.
I'll have a think about how it should be set up.

matthewhughes934:
@flatcap I've updated my approach following some of our discussions on IRC. I'm working on updating the tests under `test-library` too.

browser.c
```diff
@@ -1606,21 +1606,60 @@ void mutt_buffer_select_file(struct Buffer *file, SelectFileFlags flags,
...
+            mutt_error(_("Mailbox creation failed"));
+            break;
         }
 
-        if (imap_mailbox_create(mutt_buffer_string(&LastDir)) == 0)
+#ifdef USE_IMAP
+        if (created)
         {
           /* TODO: find a way to detect if the new folder would appear in
            *   this window, and insert it without starting over. */
```
matthewhughes934, line 1665: Not sure how much (currently none) of this I should be using. There is some coupling to IMAP, but I think that could be filtered?

browser.c
```diff
@@ -1606,21 +1606,60 @@ void mutt_buffer_select_file(struct Buffer *file, SelectFileFlags flags,
           mutt_message("%s", ARRAY_GET(&state.entry, menu->current)->name);
         break;
 
-#ifdef USE_IMAP
-      case OP_BROWSER_TOGGLE_LSUB:
-        bool_str_toggle(NeoMutt->sub, "imap_list_subscribed", NULL);
+      case OP_CREATE_MAILBOX:
```
matthewhughes934, line 1609: Is there a preference for this code to be moved to a separate function instead?

browser.c
```diff
@@ -1606,21 +1606,60 @@ void mutt_buffer_select_file(struct Buffer *file, SelectFileFlags flags,
...
+      {
+        int create_rc;
+        bool created = false;
+        struct Mailbox *m = mailbox_find(CurrentFolder);
+        /* PATH_MAX used for local mailboxes, e.g. maildir, but should also be
+         * sufficient for remote mailboxes */
```
matthewhughes934, line 1615: I'm still no sufficiently familiar with the other use cases to be confident in this assumption, so I'd be happy to hear what others think.

browser.c
flatcap: yes, good idea

browser.c
flatcap: yes, `PATH_MAX` (usually 4096) should be fine

browser.c
flatcap: Hmm...  I think we'll need most of it, to start with.

When create/delete/rename Mailbox succeeds, we need to refresh our view.
This code does it by completely tearing down the view and rebuilding it.

The architecturally-cleaner solution, which we can get to later, is:
- Browser sets up observers for the objects it's displaying
- User performs a function
- Backend performs the actual work
- Backend sends a notification of change
  e.g. Account: new Mailbox
- Browser receives notification and updates itself (non-destructively :-)

matthewhughes934:
Split things out into a separate function, I also realised the separation for the IMAP case needed a bit more attention: I didn't realise 'state.imap_browse' was conditional on a '#ifdef USE_IMAP' so rearrange things to ensure we're safe in this regard. I leant towards making things easier for myself, rather than more maintainable since most of this extra handling should be removed when we implement mbox_create for IMAP

flatcap:
From IRC
> is there a way to override a define set in audo.def?

No.
We decided ages ago that features that don't add any build dependencies should always be built.

It's simple to workaround, though, because `config.h` is only built once, during `./configure`

```sh
./configure [OPTIONS]
sed -i '/USE_IMAP/d' config.h
make
```

matthewhughes934:
Opening this (finally) for review. The last bit I'm not certain on is the desired browser behaviour after creation, should we be open in the folder we just created? I still need to poke around the IMAP case a bit more (does anyone have a nice setup for a local IMAP server for testing?)

browser.c
```diff
@@ -1127,6 +1127,77 @@ static int file_tag(struct Menu *menu, int sel, int act)
...
+    return false;
+
+  if (m_current->mx_ops->mbox_create == NULL)
+  {
+    mutt_error(_("Create is not supported for %s mailboxes"), m_current->mx_ops->name);
+    return false;
+  }
+
+  mutt_get_field(_("Create mailbox: "), path, sizeof(path), MUTT_COMP_NO_FLAGS,
+                 false, NULL, NULL);
```
matthewhughes934, line 1171: Need to handle return values here

browser.c
matthewhughes934: > Hmm... I think we'll need most of it, to start with.

Yeah, I think I'll need to dive into `imap_browse` .... it seems to be doing most of the setup here

flatcap:
Rebased over `master`

flatcap (commented):
The code looks ok, however I've found a bug.

The user is calling `<create-mailbox>` in the browser,
but the new folders are always created in the **current** mailbox, not the **selected** mailbox.

matthewhughes934:
> but the new folders are always created in the **current** mailbox, not the **selected** mailbox.

Ah right, so I should be getting the name of the current entry in the browser->extract the mailbox, like what is done for `OP_DESCEND_DIRECTORY` https://github.com/neomutt/neomutt/blob/05a57304c869ee9b5349c99fcc4fd7401aa7dcc9/browser.c#L1493 ?

flatcap:
> like what is done for `OP_DESCEND_DIRECTORY`

yes

matthewhughes934:
> but the new folders are always created in the **current** mailbox, not the **selected** mailbox.

If I understand correctly, I _think_ this might be how it is handled for the existing IMAP case (which I've just copied). If that were the case, would the preference be for separate behaviour for the new case, or to update the existing case as well?

matthewhughes934:
> > like what is done for `OP_DESCEND_DIRECTORY`
> 
> yes

Apologies for the delay on this one (:see_no_evil: I'm back to finish this off if it's still desired) I gave this a shot with 6e66c8b

browser/browser.c
```diff
@@ -1013,6 +1013,83 @@ static int file_tag(struct Menu *menu, int sel, int act)
...
+    case MX_CREATE_ERROR:
+    case -1: /* imap_mailbox_create */
+      mutt_error(_("Mailbox creation failed"));
+      break;
+  }
+
+  /* Creation failed, cleanup */
+  mutt_buffer_pool_release(&path);
+  mutt_mem_free(&new_mailbox);
+  return 1;
```
matthewhughes934, line 1090: my C has become a bit rusty. Maybe this is cleaner with a flag to guard cleanup/put all the exits closer together:

```suggestion
  bool create_ok = false;
  switch (create_rc)
  {
    case MX_CREATE_OK: /* = 0 (for imap_mailbox_create) */
      mutt_message(_("Mailbox created"));
      create_ok = true
    case MX_CREATE_EXISTS:
      mutt_error(_("A %s mailbox already exists under %s"),
                 m_current->mx_ops->name, mutt_buffer_string(path));
      break;
    case MX_CREATE_BAD_NAME:
      mutt_error(_("Bad name '%s' (is it empty?)"), mutt_buffer_string(path));
      break;
    case MX_CREATE_SYS_ERROR:
      mutt_error(_("Mailbox creation failed: %s"), strerror(errno));
      break;
    case MX_CREATE_ERROR:
    case -1: /* imap_mailbox_create */
      mutt_error(_("Mailbox creation failed"));
      break;
  }
  mutt_buffer_pool_release(&path);

  if (create_ok) {
    return 0
  } else {
    mutt_buffer_pool_release(&path);
    mutt_mem_free(&new_mailbox);
    return 1;
  }
```

flatcap:

flatcap:
Archived to: https://github.com/neomutt/neomutt-old/tree/add-creat-maildir-function

PR #2854, opened 2021-03-17
Provide a mean to turn off wrapping output of mailcap commands

gahr:
* **What does this PR do?**

~Disable wrapping in the pager if the data is being generated by a mailcap command.~

This implements a new mailcap extension flag, `x-neomutt-nowrap`, to disable wrapping in the pager for particular mailcap entries.

### DIFF
--- a/commands.c
+++ b/commands.c
@@ -378,7 +378,7 @@ int mutt_display_message(struct MuttWindow *win_index, struct MuttWindow *win_ib
 
     pview.mode = PAGER_MODE_EMAIL;
     pview.banner = NULL;
-    pview.flags = MUTT_PAGER_MESSAGE;
+    pview.flags = MUTT_PAGER_MESSAGE | (e->body->nowrap ? MUTT_PAGER_NOWRAP : 0);
     pview.win_ibar = win_ibar;
     pview.win_index = win_index;
     pview.win_pbar = win_pbar;
--- a/docs/manual.xml.head
+++ b/docs/manual.xml.head
@@ -10895,6 +10895,18 @@ text/html; lynx %s
                 <screen>text/html; firefox %s &amp; x-neomutt-keep</screen>
               </listitem>
             </varlistentry>
+            <varlistentry>
+              <term>x-neomutt-nowrap</term>
+              <listitem>
+                <para>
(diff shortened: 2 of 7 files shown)


jindraj:
makes sense to me

flatcap (commented):
We discussed this on IRC.
The general feeling was that we don't need this.

NeoMutt calls external programs to format data for display in the pager.
e.g. Attachments, using mailcap entries.

When NeoMutt calls the program, it sets the env var `$COLUMNS` to the correct size for the display window.  The allows the mailcap program to format the data correctly.

By turning off wrapping in the Pager, we risk information being lost off-screen.
Worse still, there would be no indication to the user that data had been lost.

gahr:
The more I think of an "x-neomutt-nowrap" extension, the more I like it :)

flatcap:
**If** we were to change anything, I'd want to see a table of all the possible outcomes of mailcap and pager behaviour.

For a single line of text,

The **mailcap** program can:
- Do nothing
- Wrap the line
- Truncate the line

The **pager** can:
- Do nothing (line fits on screen)
- Wrap the line
- Truncate the line

If the pager were to truncate, I think we'd want some feedback such as Vim's `listchars` -- `extends:▶,precedes:◀`
and possibly the ability to scroll right/left to access the offscreen info.

gahr:
Turns out this only works when viewing attachments or anyway not the primary part of a message. When viewing the primary part, we take a different code path:
```
mutt_display_message
  -> mutt_copy_message
    -> mutt_body_handler
      -> run_decode_and_handler
        -> autoview_handler
          -> mailcap_lookup
  -> mutt_pager
```

We need to make the info we have in `mailcap_lookup` available to `mutt_display_message` so it can set the proper flags for `mutt_pager` :(

flatcap:
I had a quick look...
Notes: https://gist.github.com/flatcap/04d2ca9a7993eaedd26650986049b8de

I wondered about attaching the `struct MailcapEntry` to `struct Body`.
Probably a bit messy.

Simpler would be to add `Body.mailcap_flags`

gahr:
> I wondered about attaching the `struct MailcapEntry` to `struct Body`.

I've had exactly the same idea :)

flatcap (approved):

PR #2855, opened 2021-03-17
Devel/notmuch tests

Austin-Ray:
* **What does this PR do?**

Adds two new unit tests for the notmuch backend. While they're not particularly complicated, it increases test coverage while I work out a way to test the more complicated logic.

### DIFF
--- a/notmuch/query.c
+++ b/notmuch/query.c
@@ -33,6 +33,20 @@
 #include "query.h"
 
 /**
+ * nm_query_type_to_string - Turn a query type into a string
+ * @param query_type Query type
+ * @retval ptr String
+ *
+ * @note This is a static string and must not be freed.
+ */
+const char *nm_query_type_to_string(enum NmQueryType query_type)
+{
+  if (query_type == NM_QUERY_TYPE_THREADS)
+    return "threads";
+  return "messages";
+}
+
+/**
  * nm_string_to_query_type - Lookup a query type
  *
  * If there's no query type, default to NM_QUERY_TYPE_MESGS.
--- a/notmuch/query.c
+++ b/notmuch/query.c
@@ -85,18 +85,37 @@ const char *nm_query_type_to_string(enum NmQueryType query_type)
 /**
  * nm_string_to_query_type - Lookup a query type
  *
- * If there's no query type, default to NM_QUERY_TYPE_MESGS.
+ * If there's an unknown query type, default to NM_QUERY_TYPE_MESGS.
  *
  * @param str String to lookup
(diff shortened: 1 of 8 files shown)


flatcap:
I've added a commit to add Notmuch code coverage.
If you have `lcov` installed, then you can:
```
./configure --testing --coverage --notmuch
make coverage
```

This will produce code coverage stats in `coverage/index.html`

notmuch/query.c
```diff
@@ -32,6 +32,56 @@
...
+
+      mutt_istr_remall(buf, variants[i]);
+    }
+  }
+
+  return query_type;
+}
+
+/**
+ * query_type_to_string - Turn a query type into a string
```
flatcap, line 72:
```suggestion
 * nm_query_type_to_string - Turn a query type into a string
```

notmuch/query.c
```diff
@@ -32,6 +32,56 @@
...
+const char *nm_query_type_to_string(enum NmQueryType query_type)
+{
+  if (query_type == NM_QUERY_TYPE_THREADS)
+    return "threads";
+  return "messages";
+}
+
 /**
  * nm_string_to_query_type - Lookup a query type
  *
```
flatcap, line 87: I've had another think about this one...
I'm not happy with the error message being in it.

I think the function should return an error value if the lookup fails,
and allow the caller to decide whether an error message should be shown.

(this would also mean that we don't need to mess around with `MuttLogger` in the tests)

flatcap (commented):
The new tests look good.

notmuch/query.c
Austin-Ray: Gah! I always miss one renaming.

notmuch/query.c
Austin-Ray: > I'm not happy with the error message being in it.

I'm of the same opinion. Utility functions should

> I think the function should return an error value if the lookup fails,

What's the best way to accomplish that? Extend `NmQueryType` with `NM_QUERY_TYPE_UNKNOWN` and return that for higher-level code to handle it? Or am I missing something obvious?

flatcap: > Extend `NmQueryType` with `NM_QUERY_TYPE_UNKNOWN`

Yes, that's seem best.

> return that for higher-level code to handle it?

Exactly.  A small wrapper function that, on error, logs and chooses the default.

flatcap:
> Experimentally add log capture test

Wow!  That's pretty out-there!

> If not, it can be dropped

Yep, sorry.

PR #2856, opened 2021-03-18
Cleanup account data when closing the last mailbox

gahr:
Imagine the sequence `unmailboxes *`; `change-folder`. The currently open
mailbox cannot be removed by `unmailboxes *`, which only marks it as
hidden. Then, on `change-folder`, the current mailbox is closed, but
nobody notifies the Account that the mailbox is gone, so in case of IMAP
the connection to the old server stays around.


Austin-Ray (commented):
This looks like a sensible change to me. Since `imap/imap.c` is the only caller of `mx_ac_remove()`, we don't have to worry about anyone else expecting the mailbox to be free'd.

In addition, removing the free seems like a good idea because anyone just reading the docstring might be surprised when their mailbox disappeared instead of just unlinking the account.

flatcap (approved):

PR #2858, opened 2021-03-18
config: fix a couple of small bugs

flatcap:
- 619f926b18 config: deprecate $visual
  Config variable hasn't been used in a long time

- 16923700c4 config: fix editor defaults
  `$EDITOR` env var wasn't being honoured, when `$editor` neomutt config wasn't set

- 3d42e5706d config: fix signature warning
  Only warn about a missing signature file **IF** the user set the config

### DIFF
--- a/docs/config.c
+++ b/docs/config.c
@@ -1066,7 +1066,7 @@
 ** ignored for interoperability reasons.
 */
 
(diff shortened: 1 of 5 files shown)


Austin-Ray (approved):
LGTM

PR #2861, opened 2021-03-22
Eliminate Config Global Variables (all of them)

flatcap:
- **41** commits
- **235** files changed, **5407** insertions(+), **4675** deletions(-)

This is a horrifyingly large diff -- I'm not expecting anyone to read it.
In brief, it eliminates **all** of the `C_*` global variables that are behind Config items.

`C_Sort` would be replaced with a call to one of the [helper functions](https://github.com/neomutt/neomutt/blob/master/config/helpers.h#L31-L42).
```c
const short c_sort = cs_subset_sort(NeoMutt->sub, "sort");
```

This is a bit verbose, but it **guarantees** that any changes to the Config must go through similar functions, which will result in **notifications**.

## Prep

The first 8 commits do a little preparatory work:

- 057412334c config: deprecate $visual 
- 7774687646 config: fix editor defaults                                                                                                                                      
- 847224df19 config: fix signature warning 
- b8376f6250 config: use correct name for synonyms 
  If you ask `:set askbcc?`, NeoMutt will reply with `set ask_bcc = yes`
- 1ac9cc5aef config: eliminate C_* from startup code 
- 0aaf6a31f6 config: delay init 
- d1f05fe31e config: fix setting initial string 
- fc7f18294a test: add shared NeoMutt/Config init
  This also adds an `acutest` destructor function, which is used to free the BufferPool.
  Now test functions can use the BufferPool without leaking memory on exit.

## Libraries

These 14 commits work on libraries.
Generally, it's lots of simple replacements.
Very little code actually **set**s config.

- Replace `C_X` with a helper
- Eliminate `C_X` from the config definition

The notable exceptions are around `$sort` and `$sort_aux`.
Some code saves their values, performs some work, then restore the values.
Now that set/reset action causes notifications which may cause extra work in the short term.

- 41fbb8db5d config: imap   
- 18826fa29d config: conn   
- 3b97f1d239 config: nntp 
- bde058bb2c config: maildir  
- c4a0d5e552 config: notmuch 
- 8c4c42d674 config: history 
- 7fa136cba9 config: pop 
- e8c3d83c00 config: autocrypt 
- 5cd4cc130f config: email 
- 725f9c860d config: pager 
- ec70dad14d config: mbox 
- 0a572871b1 config: hcache 
- 032f6fe5ea config: gui 
- ff11f9339a config: misc libraries (⚠️ Big)

## Main Code

These 15 commits fix the main body of code in NeoMutt,
including (@paniash's work on `hdrline.c` :heart: )

The second commit is a ugly hack (put into a separate commit for easy identification).
There are four places in `mutt_thread.c` that alter `$sort` which cause ugly multiple refreshes of the screen.
By saving and restoring the pseudo global `OptNeedResort`, we avoid this problem.

- 33ab748d4b config: mutt_thread.c
- c587a83121 threads: prevent background sorting from causing refreshes
- abfeff6659 config: muttlib.c
- e3acfa3c9b config: menu.c
- 6f6598f4bf config: recvattach.c
- aad601848d config: mx.c
- 8d1fabc891 config: browser.c
- bd783a9ced config: commands.c
- 7fafd12465 config: postpone.c
- c25d128327 config: recvcmd.c
- f09a213a0f config: handler.c
- 8aa29f44eb config: main.c
- 59079caa81 config: mutt_logging.c
- 4b6e42149b config: hdrline.c
- 83c37cdfa8 config: misc files (⚠️ Big)

## Config Refactoring

The final four commits do some tidying/refactoring.

Notably, I've reorganised `struct ConfigDef` which defines the config variables.
Before, the config looked like:
```c
{ "mail_check_stats", DT_BOOL, &C_MailCheckStats, false, 0, NULL, "Periodically check for new mail" },
```

The global variable has been eliminated, but we still need that spot in the struct to store the actual value.
Rather than every config having a `NULL` in the spot, the struct has been rearranged to move it to the end.
Where it can be dropped altogether

```c
{ "mail_check_stats", DT_BOOL, false, 0, NULL, "Periodically check for new mail", NULL },
{ "mail_check_stats", DT_BOOL, false, 0, NULL, "Periodically check for new mail" },
```

- ac1b3813e1 config: eliminate final Config variables (⚠️ Big)
- 2569fc6a7c config: reorg ConfigDef to hide config variable (⚠️ Big)
- 185e9e55bd config: change config var type to intptr_t
  This also lead to a little simplification of the config code
- 608330474f config: eliminate variables from config tests (⚠️ Big)
  This is a huge commit as the config test code used a HUGE amount of config variables!


flatcap:
For testing, I've
- Built locally using ~21 configure combinations
- Built using Cirrus CI
- Run NeoMutt
  - seems OK, no crashes or visual quirks
  - ASAN is happy
- All tests run correctly
- All coverage tests seem OK
- I've independently checked that each: (config name|config type|helper function) triple is correct

alias/config.c
```diff
@@ -70,5 +70,5 @@ struct ConfigDef AliasVars[] = {
  */
 bool config_init_alias(struct ConfigSet *cs)
 {
-  return cs_register_variables(cs, AliasVars, DT_NO_VARIABLE);
+  return cs_register_variables(cs, AliasVars, 0);
```
Austin-Ray, line 73: I see the pattern of `cs_register_variables(cs, ..., 0);` throughout all the `config.c` files. However, I don't see a single usage of `cs_register_variables()` that doesn't provide `0`. Do we need to keep it?

Austin-Ray:
I only got through the first 7 files (started late), but I'll continue my review tomorrow.

pop/auth.c
```diff
@@ -343,10 +348,11 @@ static enum PopAuthRes pop_auth_oauth(struct PopAccountData *adata, const char *
   snprintf(auth_cmd, auth_cmd_len, "AUTH OAUTHBEARER %s\r\n", oauthbearer);
   FREE(&oauthbearer);
 
+  const short c_debug_level = cs_subset_number(NeoMutt->sub, "debug_level");
```
gahr, line 351:
```suggestion

```

pop/auth.c
```diff
@@ -343,10 +348,11 @@ static enum PopAuthRes pop_auth_oauth(struct PopAccountData *adata, const char *
...
   int ret = pop_query_d(adata, auth_cmd, strlen(auth_cmd),
 #ifdef DEBUG
                         /* don't print the bearer token unless we're at the ungodly debugging level */
-                        (C_DebugLevel < MUTT_SOCK_LOG_FULL) ?
+                        (c_debug_level < MUTT_SOCK_LOG_FULL) ?
```
gahr, line 355:
```suggestion
                        (cs_subset_number(NeoMutt->sub, "debug_level") < MUTT_SOCK_LOG_FULL) ?
```

To pet the "unused variable" warning in non-debug builds.

gahr:
I am running this branch, I'll report here whatever I find:

- [ ] delete an email + sync the mailbox results in `sort` being reset to `unsorted`.

alias/config.c
flatcap: The param was added for the the `NO_VAR` case, but I had other ideas for it.
My favourite was `DT_DISABLED` -- mostly a developer feature, but might be useful for packagers.

`DT_DISABLED` config would be accepted when reading a config file, but have no effect.
Libraries that were disabled by `./configure` would register their config, but no other code would be built.

As a developer this would allow you to have large config files that are quietly accepted when you don't have features enabled.

(disabled config would be logged)

flatcap:
> `:exec edit-or-view-raw-message` -> `Error`

I can't seem to repeat this.
I've tried IMAP, Maildir and Maildir (with zero config).

gahr:
> > `:exec edit-or-view-raw-message` -> `Error`
> 
> I can't seem to repeat this.
> I've tried IMAP, Maildir and Maildir (with zero config).

yeah I'm almost certain it *has* something to do with some config ;)

editmsg.c
```diff
@@ -66,13 +66,14 @@ static int ev_message(enum EvMessage action, struct Mailbox *m, struct Email *e)
   struct Buffer *fname = mutt_buffer_pool_get();
   mutt_buffer_mktemp(fname);
 
-  enum MailboxType otype = C_MboxType;
-  C_MboxType = MUTT_MBOX;
+  unsigned char c_mbox_type = cs_subset_enum(NeoMutt->sub, "mbox_type");
+  enum MailboxType otype = c_mbox_type;
+  c_mbox_type = MUTT_MBOX;
```
gahr, line 71: So my `mbox_type` is maildir, but we need `mbox` here to create a single temporary file. However, you're not overriding the global value `C_MboxType` of the `mbox_type` variable anymore, but just the local copy `c_mbox_type`. Then `mx_mbox_open` fetches `mbox_type` again, finds `maildir`, and the temporary file becomes a temporary directory with `cur`, `tmp`, `new`. At line 111 we stat it thinking it's a file, but because it's a directory we're confused and boom.

gahr: I fear this is a subtle bug that might have crept in in several places, not easily spotted by static analysis :(

flatcap: Ah, I see.  Thanks.
Fixed in 7985e4d043.

flatcap: > I fear this is a subtle bug that might have crept in in several places,

You're probably right...

> not easily spotted by static analysis

Actually, it should be possible.

To make the refactoring quicker, I generated example `variable = helper_function()` for every config variable.
I **meant** to make them all `const` variables to avoid situations like this.
Looks like I missed some.

I'll review the constness of all of them,
and re-review the ~180 places where we set variables
(to make sure noone relies on an old variable).

gahr: >  I **meant** to make them all `const` variables to avoid situations like this.

ah yes, that makes sense!

PR #2862, opened 2021-03-23
Allow a -noregex param to folder and mbox hooks

gahr:
This is helpful if we want to match mailbox that comes from a variable exactly (or we simply can't be bothered to escape the whole thingie).
As an example, in `folder-hook -noregex $trash set my_foo=bar` we don't want an expansion of `$trash` such as `imaps://user@gmail.com/[Gmail]/Trash` to be treated as a regex containing a character class `[Gmail]`.

An alternative approach would be to create different commands with an additional flag
```
{ "folder-hook-noregex",  mutt_parse_hook,  MUTT_FOLDER_HOOK | MUTT_HOOK_NOREGEX}`
```
and handle the flag in `mutt_parse_hook`. I'm fine with implementing either approach.

### DIFF
--- a/docs/manual.xml.head
+++ b/docs/manual.xml.head
@@ -4559,8 +4559,11 @@ bind index gg first-entry
       </para>
       <cmdsynopsis>
         <command>folder-hook</command>
+        <arg choice="opt">
+          <replaceable class="parameter">-noregex</replaceable>
+        </arg>
         <arg choice="plain">
-          <replaceable class="parameter">regex</replaceable>
+          <replaceable class="parameter">pattern</replaceable>
         </arg>
         <arg choice="plain">
           <replaceable class="parameter">command</replaceable>
@@ -4570,9 +4573,11 @@ bind index gg first-entry
         It is often desirable to change settings based on which mailbox you are
         reading. The <command>folder-hook</command> command provides a method
         by which you can execute any configuration command.
-        <emphasis>regex</emphasis> is a regular expression specifying in which
-        mailboxes to execute <emphasis>command</emphasis> before loading. If
-        a mailbox matches multiple <command>folder-hook</command>s, they are
+        The <emphasis>command</emphasis> is executed before loading any mailboxes
+        matching <emphasis>pattern</emphasis>. The <emphasis>-noregex</emphasis>
+        switch controls whether <emphasis>pattern</emphasis> is matched using
+        a simple string comparison or a full regex match.
+        If a mailbox matches multiple <command>folder-hook</command>s, they are
         executed in the order given in the <literal>.neomuttrc</literal>.
       </para>
       <para>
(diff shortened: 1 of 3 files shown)


flatcap:
Hmm... that sounds useful.
I had a similar idea for the `set` command.

Some other possibilties:

- `noeval set var = literal`
  - **Prefix** that could be applied to many different commands
  - Literal ends at end-of-line
  - Zero or one space character after `=` is allowed

- `set var == literal`
  - Define a new **operator** to mean exactly equal
  - Literal ends at end-of-line
  - Zero or one space character after `=` is allowed

- `set var = /literal/`
  - Use Vim-style symbol **delimiters** rather than single `'` / double `"` quotes
  - Any non-alphanumeric character will do: `/` `!` `:` `|` etc

docs/manual.xml.head
```diff
@@ -4560,8 +4560,11 @@ bind index gg first-entry
       </para>
       <cmdsynopsis>
         <command>folder-hook</command>
+        <arg choice="opt">
+          <replaceable class="parameter">-noregex</replaceable>
+        </arg>
         <arg choice="plain">
-          <replaceable class="parameter">regex</replaceable>
+          <replaceable class="parameter">pattern</replaceable>
```
flatcap, line 4567: Hmm... I'm not sure about this.
In the docs 'pattern' generally refers to a NeoMutt pattern, e.g. `~f addr`

Perhaps: `string_to_match`, `match_string`
Hmm... I'm not convinced by those.

hook.c
```diff
@@ -88,8 +88,10 @@ enum CommandResult mutt_parse_hook(struct Buffer *buf, struct Buffer *s,
   struct Hook *hook = NULL;
   int rc = MUTT_CMD_ERROR;
   bool pat_not = false;
+  bool noregex = false;
```
flatcap, line 91: Could we invert this to `bool use_regex = true`
...

hook.c
```diff
@@ -145,7 +158,7 @@ enum CommandResult mutt_parse_hook(struct Buffer *buf, struct Buffer *s,
 
     struct Buffer *tmp = mutt_buffer_pool_get();
     mutt_buffer_copy(tmp, pattern);
-    mutt_buffer_expand_path_regex(tmp, true);
+    mutt_buffer_expand_path_regex(tmp, !noregex);
```
flatcap, line 161: ... then we eliminate the double-negative here...

hook.c
```diff
@@ -156,7 +169,14 @@ enum CommandResult mutt_parse_hook(struct Buffer *buf, struct Buffer *s,
       goto cleanup;
     }
 
-    mutt_buffer_copy(pattern, tmp);
+    if (noregex)
```
flatcap, line 172: ... and flip this around

flatcap (commented):
Some thoughts.
Feel free to disregard them.

flatcap (approved):

PR #2869, opened 2021-03-29
refactor: commands that require a Context

flatcap:
When NeoMutt parses the three commands `subjectrx`, `attachments` and `alternates`,
it uses the global `Context` to make updates to the current `Mailbox`.

These commits refactor the commands, to:
- Move the the parsing code and global lists into a separate source file
- Create a new notification to inform the Index of change
- Create an observer in the Index to act on changes

For now, this doesn't eliminate the use of the global `Context`,
but it moves it into the Index which **does** have access to the current `Mailbox`.

---

- 257b582693 refactor: 'subjectrx' command handling -- **`subjectrx.c`**
- 927b842154 refactor: 'attachments' command handling -- **`attachments.c`**
- 60462b1d35 refactor: 'alternates' command handling -- **`alternates.c`**


gahr (approved):

flatcap:
> Perhaps I would change this to return either `e->env->disp_subj` or `e->env->subject`,
> so we can just ` const char *subj = subjrx_apply_mods(e->env);`

Yeah, I wondered about that.

I'm think I'm going to leave it as is:
- The function handles the **action** of generating the string
- The caller decides on the **business logic** of which string to display

PR #2870, opened 2021-03-29
refactor: Don't use the global Context when not necessary

flatcap:
These commits reduce function dependencies.
They mostly eliminate the use of the global `Context`.

- 3d3d009b07 strip gui dep from `mutt_check_traditional_pgp()`
  Crypto functions should know about GUI flags

- b86af1a2ae context: eliminate from pattern functions
  Don't pass a Context parameter when only a Mailbox is needed

- 2ac84e2974 context: eliminate from alias/query
  The alias/query patterns don't need a Context/Mailbox

- 0f44eb4d17 context: eliminate from `mutt_save_message_ctx()`
  Pass a Context paramater, rather than using the global

- 237cb9f811 context: reduce complexity of `complete_all_nm_tags()`
  Simple tidying

### DIFF
--- a/alias/dlgalias.c
+++ b/alias/dlgalias.c
@@ -42,6 +42,7 @@
 #include "lib.h"
 #include "pattern/lib.h"
 #include "alias.h"
+#include "context.h"
 #include "format_flags.h"
 #include "gui.h"
 #include "keymap.h"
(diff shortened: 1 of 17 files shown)


gahr (approved):

flatcap:
The odd line-splitting is `clang-format` doing its thing.
Fortunately, a later commit causes the splits to be reunited.

PR #2871, opened 2021-03-30
Eliminate `Menu->redraw_data`

flatcap:
`struct Menu` has two data members, `mdata` and `redraw_data`.
Compose used to need both, but since its redesign, it only needs one.

- 91acfec71d compose: tidy code

- 1b4986082b compose: merge data structs
  Use `struct ComposeRedrawData` for all our needs

- d1d14b20e7 menu: eliminate redraw_data member

### DIFF
--- a/mutt_menu.h
+++ b/mutt_menu.h
@@ -115,8 +115,6 @@ struct Menu
    * @param menu Menu to redraw
    */
   void (*custom_redraw)(struct Menu *menu);
-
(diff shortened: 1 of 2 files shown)


ngortheone (approved):

ngortheone:
I like this. Seems like a nice simplification.

gahr:
Nice!

Issue #2865, opened 2021-03-25
Assert failure in mutt_prex_capture


l-mb:
On starting neomutt 20210205, it began parsing my Maildir folders and immediately crashes with:

```
neomutt: mutt/prex.c:338: mutt_prex_capture: Assertion `(h->re->re_nsub == (h->nmatches - 1)) && "Regular expression and matches enum are out of sync"' failed.
```

Taking a look at the backtrace:

```
(gdb) bt
#0  0x00007ffff78ef4a5 in raise () at /lib64/libc.so.6
#1  0x00007ffff78d8864 in abort () at /lib64/libc.so.6
#2  0x00007ffff78d8749 in _nl_load_domain.cold () at /lib64/libc.so.6
#3  0x00007ffff78e79b6 in  () at /lib64/libc.so.6
#4  0x000055555566d592 in mutt_prex_capture (which=<optimized out>, str=0x5555558d8612 "Sun, 14 Jun 2020 17:44:02 +0200") at mutt/prex.c:338
#5  0x0000555555668032 in mutt_date_parse_date (s=0x5555558d8612 "Sun, 14 Jun 2020 17:44:02 +0200", tz_out=0x0) at mutt/date.c:452
#6  0x000055555565f86b in mutt_rfc822_parse_line
    (env=env@entry=0x5555558d3040, e=e@entry=0x5555558b4e80, line=line@entry=0x5555558d8500 "Received", p=p@entry=0x5555558d850a "XXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXX"..., user_hdrs=user_hdrs@entry=false, weed=weed@entry=false, do_2047=true) at email/parse.c:876
#7  0x000055555566074f in mutt_rfc822_parse_line
    (do_2047=true, weed=false, user_hdrs=false, p=0x5555558d850a "XXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXX"..., line=0x5555558d8500 "Received", e=0x5555558b4e80, env=0x5555558d3040) at email/parse.c:608
#8  mutt_rfc822_read_header (fp=<optimized out>, e=0x5555558b4e80, user_hdrs=<optimized out>, weed=<optimized out>) at email/parse.c:1210
#9  0x00005555555fb288 in maildir_parse_stream
[...]

PR #2882, opened 2021-04-05
Fix pcre2-posix runtime crash on openSUSE

scosu:
* **What does this PR do?**

Fixes the runtime assertion when pcre2-posix leaks in, see the commit messages of the individual commits

#2865

### DIFF
--- a/.cirrus.yml
+++ b/.cirrus.yml
@@ -26,7 +26,7 @@ task:
         - ${PKG_INSTALL} ${PKG_COMMON}
         - git clone --depth 1 https://github.com/neomutt/neomutt-test-files.git ${NEOMUTT_TEST_DIR}
         - (cd ${NEOMUTT_TEST_DIR} && ./setup.sh)
-      configure_script: ./configure --autocrypt --disable-doc --disable-idn --idn2 --notmuch --testing
+      configure_script: ./configure --autocrypt --disable-doc --disable-idn --idn2 --notmuch --testing --pcre2
       build_script: make
       version_script: ./neomutt -v
       test_script: make test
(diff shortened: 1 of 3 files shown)


gahr:
This requires using `--pcre2` whenever `libpcre2-posix.so` is installed, which I don't want to do.

I can see from the [changelog here](https://rpmfind.net/linux/RPM/opensuse/tumbleweed/i586/libncurses5-6.2.20210313-16.1.i586.html) that libncursesw on OpenSUSE recently started to require/depend on PCRE2.

I don't think it's reasonable to require all consumers of libncursesw to be patched to switch from libc's regex API to actively use PCRE2's APIs because of libncursesw's needs.

If that's the case on OpenSUSE, I think this is something that needs to be done in the OpenSUSE port of NeoMutt, as was done by turning on `--pcre2` for NeoMutt.

Is it really the case that on OpenSUSE one can't both i) link against libncursesw, and ii) use libc's regex API expecting it to behave as [OpenGroup mandates](https://pubs.opengroup.org/onlinepubs/9699919799/basedefs/regex.h.html)?

scosu:
Yes I think using '--pcre2' is the correct solution on openSUSE side.

But I think something like this should be detected during the build and not at runtime. Runtime is basically too late as you are having code that is using an unexpected library that does some things slightly differently. If this is not changed, there is be a big chance that the same thing happens with the next distribution or the next person that compiles it on openSUSE.

I would also be happy about any other solution that wouldn't fail at a random assert at runtime but shows the issue immediately during the build.

gahr:
Thanks, I took your suggestion and did a new PR in #2883.

Issue #2884, opened 2021-04-06
ASAN reported memory leak


ngortheone:
How to reproduce:

1. Start neomutt
2. while in index do `exec view-raw-message` on any message
3. close your EDITOR, you should pop back to neomutt index
4. exit neomott

```
=================================================================
==1190179==ERROR: LeakSanitizer: detected memory leaks

Direct leak of 272 byte(s) in 1 object(s) allocated from:
    #0 0x7f7351ee8a0f in calloc (/usr/lib/gcc/x86_64-pc-linux-gnu/10.2.0/libasan.so.6+0xafa0f)
    #1 0x55cf44df1328 in mutt_mem_calloc mutt/memory.c:55
    #2 0x55cf44d963c3 in mailbox_new core/mailbox.c:54
    #3 0x55cf44bdab4d in mx_path_resolve /usr/local/src/github.com/neomutt/neomutt/mx.c:1711
    #4 0x55cf44b52b49 in ev_message /usr/local/src/github.com/neomutt/neomutt/editmsg.c:73
    #5 0x55cf44b547ed in mutt_ev_message /usr/local/src/github.com/neomutt/neomutt/editmsg.c:288
    #6 0x55cf44c26a96 in mutt_index_menu index/index.c:3384
    #7 0x55cf44b9d36d in main /usr/local/src/github.com/neomutt/neomutt/main.c:1261
    #8 0x7f7351be9e39 in __libc_start_main (/lib64/libc.so.6+0x23e39)

Indirect leak of 200 byte(s) in 1 object(s) allocated from:
    #0 0x7f7351ee8a0f in calloc (/usr/lib/gcc/x86_64-pc-linux-gnu/10.2.0/libasan.so.6+0xafa0f)
    #1 0x55cf44df1328 in mutt_mem_calloc mutt/memory.c:55
    #2 0x55cf44d9645f in mailbox_new core/mailbox.c:60
    #3 0x55cf44bdab4d in mx_path_resolve /usr/local/src/github.com/neomutt/neomutt/mx.c:1711
    #4 0x55cf44b52b49 in ev_message /usr/local/src/github.com/neomutt/neomutt/editmsg.c:73
    #5 0x55cf44b547ed in mutt_ev_message /usr/local/src/github.com/neomutt/neomutt/editmsg.c:288
    #6 0x55cf44c26a96 in mutt_index_menu index/index.c:3384
[...]

PR #2886, opened 2021-04-07
fix memory leaks

flatcap:
Fix a few memory leaks.

1. In `main()` use the Mailbox returned from `mutt_index_menu()` and free it
2. Free a couple of strings in `mutt_send_message()`
3. Free the Mailbox in `mx_fastclose_mailbox()`
   This function calls `mx_ac_remove()`, but that doesn't free the Mailbox any more

Change `mx_mbox_close()` and `mx_fastclose_mailbox()` to take `Mailbox**`,
so that if they close a Mailbox, they also `NULL` the pointer.

Fixes: #2884

### DIFF
--- a/main.c
+++ b/main.c
@@ -1258,9 +1258,10 @@ int main(int argc, char *argv[], char *envp[])
       mutt_debug(LL_NOTIFY, "NT_MAILBOX_SWITCH: %p\n", m);
       notify_send(dlg->notify, NT_MAILBOX, NT_MAILBOX_SWITCH, &ev_m);
 
-      mutt_index_menu(dlg, m);
+      m = mutt_index_menu(dlg, m);
       dialog_pop();
       mutt_window_free(&dlg);
+      mailbox_free(&m);
       log_queue_empty();
       repeat_error = false;
     }
--- a/mbox/mbox.c
+++ b/mbox/mbox.c
@@ -1131,7 +1131,7 @@ static enum MxStatus mbox_mbox_check(struct Mailbox *m)
   /* fatal error */
 
   mbox_unlock_mailbox(m);
-  mx_fastclose_mailbox(m);
+  mx_fastclose_mailbox(&m);
   mutt_sig_unblock();
   mutt_error(_("Mailbox was corrupted"));
   return MX_STATUS_ERROR;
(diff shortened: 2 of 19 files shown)


main.c
```diff
@@ -1256,10 +1256,11 @@ int main(int argc, char *argv[], char *envp[])
       struct EventMailbox em = { m };
       notify_send(dlg->notify, NT_MAILBOX, NT_MAILBOX_SWITCH, &em);
 
-      mutt_index_menu(dlg, m);
+      m = mutt_index_menu(dlg, m);
```
matthewhughes934, line 1259: Could this be assigned to a separate variable?

matthewhughes934 (approved):
LGTM (for what that's worth :wink:). Out of curiosity, what's the reason behind the difference in naming convention `mx_mbox_close` vs `mbox_fastclose_mailbox`?

flatcap:
> what's the reason behind the difference in naming convention?
> `mx_mbox_close()` vs `mx_fastclose_mailbox()`

Purely history.
'fastclose' is a relic of the old code that we haven't managed to tidy away.

The Mailbox API (`struct MxOps`) has a strict naming convention
and so do the immediate functions around it such as the backend implementations and the frontend wrappers.
After that it gets a bit looser.

main.c
flatcap: Actually, no.

The original Mailbox, `m`, might get closed by the Index, e.g. if the user changes Account.
Or it might not.

When `mutt_index_menu()` finishes, it returns the 'current' Mailbox, which we must `mailbox_free()`.

mbox/mbox.c
```diff
@@ -1801,11 +1801,12 @@ static enum MxStatus mbox_mbox_check_stats(struct Mailbox *m, uint8_t flags)
...
     mx_mbox_open(m, MUTT_QUIET | MUTT_NOSORT | MUTT_PEEK);
-    mx_mbox_close(m);
-    m->peekonly = old_peek;
+    mx_mbox_close(&m);
+    if (m)
+      m->peekonly = old_peek;
   }
 
-  if (m->has_new || m->msg_new)
+  if (m && (m->has_new || m->msg_new))
```
gahr, line 1809: This is called through `mutt_mailbox_check` -> `mailbox_check` on all mailboxes -> `mx_mbox_check_stats` -> `m->mx_ops->mbox_check_stats`, which is `mbox_mbox_check_stats` for mbox. This means we'll come in with unopened mailboxes most of the times. I suggest to save the return code before calling `mx_mbox_close`.

flatcap:
This is probably still the right idea, but too ambitious for now.
It will return.

PR #2889, opened 2021-04-08
Set invalidity flags for gpgme/smime keys

ftilde:
* **What does this PR do?**

Previously, flags that show keys to be revoked/expired/disabled were
not transfered from the gpgme_key_t to KeyFlags for smime, even though
they were for pgp keys.

Not sure if there are any. I've just noticed that neomutt kept showing me a list of smime keys for contacts where all but one of them were expired and I went to see why. Turns out that there is code that checks for/filters valid keys (in particular paths that use `pgp_show_unusable`), but the flags that indicate invalidity were never set.

### DIFF
--- a/ncrypt/crypt_gpgme.c
+++ b/ncrypt/crypt_gpgme.c
@@ -3484,6 +3484,13 @@ static struct CryptKeyInfo *get_candidates(struct ListHead *hints, SecurityFlags
       if (key_check_cap(key, KEY_CAP_CAN_SIGN))
         flags |= KEYFLAG_CANSIGN;
 
+      if (key->revoked)
+        flags |= KEYFLAG_REVOKED;
+      if (key->expired)
+        flags |= KEYFLAG_EXPIRED;
+      if (key->disabled)
+        flags |= KEYFLAG_DISABLED;
+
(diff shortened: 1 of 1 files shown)


ngortheone (approved):

gahr (approved):

flatcap:
Thanks @ftilde!
Few people understand the crypto code well, so if you have any other ideas we'd love to hear them.

You can also find us on IRC:  `#neomutt` on `irc.freenode.net`  ([web client](https://webchat.freenode.net/?channels=neomutt))

ftilde:
On 09.04.2021 02:44, Richard Russon wrote:
>Thanks @ftilde!
>Few people understand the crypto code well, so if you have any other ideas we'd love to hear them.
>
>You can also find us on IRC:  `#neomutt` on `irc.freenode.net`  ([web client](https://webchat.freenode.net/?channels=neomutt))

Oh well, I'm not really a cryptography expert myself. I just noticed 
while debugging that a key that was definitely expired did not have the 
appropriate flag set and went looking for places that _do_ set this 
flag. Well, the pgp case of gpgme did, so I figured that we should do 
the exact same thing for smime...

PR #2892, opened 2021-04-11
limit observer callbacks to a single type

flatcap:
Filter out just the notifications types the user wants.
Observing `NT_ALL` will give an unfiltered view.

This should reduce the torrent of callbacks.
(I _thought_ I'd already done this)


Austin-Ray (approved):
LGTM

PR #2893, opened 2021-04-12
Fix 2891

### LINKED ISSUE #2891: neomutt takes forever "Sorting mailbox..." on termux
## Expected Behaviour

Opening neomutt

## Actual Behaviour

Takes (seemingly) forever "Sorting mailbox..."
I waited for about 15 minutes on a mailbox that is about 500 messages.
No crash, though.

## Steps to Reproduce

* neomutt with notmuch
* in termux
* open neomutt without flags

- Always

- When I setup neomutt on termux (never had it running before)

## NeoMutt Version

```
$ neomutt -v
NeoMutt 20210205

System: Linux 4.14.180-perf-g302b184 (aarch64)
ncurses: ncurses 6.2.20200725 (compiled with 6.2.20200725)
libiconv: 1.16
OpenSSL: OpenSSL 1.1.1i  8 Dec 2020
libnotmuch: 5.3.0
storage: gdbm
compression: zstd

Configure options: --host=aarch64-linux-android --prefix=/data/data/com.termux/files/usr --with-mailpath=/data/data/com.termux/files/usr/var/mail --notmuch --disable-gpgme --disable-idn --zstd --sasl --ssl --gdbm --gss

Compilation CFLAGS:  -fstack-protector-strong -Oz -std=c99 -D_ALL_SOURCE=1 -D_GNU_SOURCE=1 -D__EXTENSIONS__ -I/data/data/com.termux/files/usr/include -DNCURSES_WIDECHAR

Default options:

Compile options:
  +getaddrinfo -gnutls -gpgme +gss +hcache -homespool -idn +inotify 
  -locales_hack -lua +meta -mixmaster +nls +notmuch +openssl +pgp +regex +sasl 
  +smime -sqlite +start_color +sun_attachment +typeahead 
MAILPATH="/data/data/com.termux/files/usr/var/mail"
PKGDATADIR="/data/data/com.termux/files/usr/share/neomutt"
SENDMAIL="/usr/sbin/sendmail"
SYSCONFDIR="/data/data/com.termux/files/usr/etc"

```

## Extra Info

> Operating System and its version

Android 10, termux from fdroid (up to date as of time of writing this)

> Were you using multiple copies of NeoMutt at once? 

No

> Were you using 'screen' or 'tmux'?

Reproducible with and without tmux. 

> Is your email local (maildir) or remote (IMAP)?

Local, notmuch.

---

If someone can guide me through the steps of debugging this, I'd love to help out! Maybe some termux-specific knowledge is required tho.

Austin-Ray:
Validates `nm_default_url` to ensure its in the correct form, and changes sleep time from 8.33 minutes to the correct 1/2 second.

~~Commit 22c5f2297 changed `nanosleep()` to `mutt_date_sleep_ms()`, but
didn't alter the valaue so a half second sleep became 8 minutes and 20
seconds.~~

Fixes #2891 

~~Partial fix for #2891~~

~~This fixes the long  timeout since instead of 5 second timeout, we get an 83 minute and 20 second timeout. However, it won't correct the invalid `notmuch` URI issue.~~

### DIFF
--- a/notmuch/config.c
+++ b/notmuch/config.c
@@ -30,13 +30,39 @@
 #include <stddef.h>
 #include <config/lib.h>
 #include <stdbool.h>
+#include "notmuch/private.h"
+
+/**
+ * is_valid_notmuch_url - Checks that a URL is in required form.
+ * @retval true url in form notmuch://<absolute path>
+ * @retval false url is not in required form.
+ */
+static bool is_valid_notmuch_url(const char *url)
+{
+  return mutt_istr_startswith(url, NmUrlProtocol) && (url[NmUrlProtocolLen] == '/');
+}
+
+/**
+ * nm_default_url_validator - Ensure nm_default_url is of the form notmuch://<absolute path> - Implements ConfigDef::validator()
+ */
+static int nm_default_url_validator(const struct ConfigSet *cs, const struct ConfigDef *cdef, intptr_t value, struct Buffer *err)
+{
+  const char *url = (const char *) value;
+  if (!is_valid_notmuch_url(url))
+  {
+    mutt_buffer_printf(err, _("nm_default_url must be: notmuch://<absolute path> . Current: %s"), url);
+    return CSR_ERR_INVALID;
+  }
+
+  return CSR_SUCCESS;
+}
 
 static struct ConfigDef NotmuchVars[] = {
   // clang-format off
   { "nm_db_limit", DT_NUMBER|DT_NOT_NEGATIVE, 0, 0, NULL,
     "(notmuch) Default limit for Notmuch queries"
   },
-  { "nm_default_url", DT_STRING, 0, 0, NULL,
+  { "nm_default_url", DT_STRING, 0, 0, nm_default_url_validator,
     "(notmuch) Path to the Notmuch database"
   },
   { "nm_exclude_tags", DT_STRING, 0, 0, NULL,
--- a/notmuch/private.h
+++ b/notmuch/private.h
@@ -40,6 +40,7 @@ struct Mailbox;
    (LIBNOTMUCH_MAJOR_VERSION == (major) &&                                        \
     LIBNOTMUCH_MINOR_VERSION == (minor) && LIBNOTMUCH_MICRO_VERSION >= (micro)))
 
+extern const char NmUrlProtocol[];
 extern const int NmUrlProtocolLen;
 
 notmuch_database_t *nm_db_do_open     (const char *filename, bool writable, bool verbose);
--- a/notmuch/db.c
+++ b/notmuch/db.c
@@ -119,7 +119,7 @@ notmuch_database_t *nm_db_do_open(const char *filename, bool writable, bool verb
 
     if (verbose && ct && ((ct % 2) == 0))
       mutt_error(_("Waiting for notmuch DB... (%d sec)"), ct / 2);
-    mutt_date_sleep_ms(500000); /* Half a second */
+    mutt_date_sleep_ms(500); /* Half a second */
     ct++;
   } while (true);
 


notmuch/config.c
```diff
@@ -30,14 +30,40 @@
...
 #include <config/lib.h>
 #include <stdbool.h>
+#include "notmuch/private.h"
+
+/**
+ * is_valid_notmuch_url - Checks that a URL is in required form.
+ * @retval true url in form notmuch://<absolute path>
+ * @retval false url is not in required form.
+ */
+bool is_valid_notmuch_url(const char *uri)
```
gahr, line 40: static?

notmuch/private.h
```diff
@@ -40,6 +40,7 @@ struct Mailbox;
    (LIBNOTMUCH_MAJOR_VERSION == (major) &&                                        \
     LIBNOTMUCH_MINOR_VERSION == (minor) && LIBNOTMUCH_MICRO_VERSION >= (micro)))
 
+extern const char NmUrlProtocol[];
```
gahr, line 43: Does it need to be extern?

notmuch/config.c
```diff
@@ -30,14 +30,40 @@
...
+{
+  return mutt_istr_startswith(uri, NmUrlProtocol) && uri[NmUrlProtocolLen] == '/';
+}
+
+/**
+ * validate_nm_default_url - Ensure nm_default_url is in form notmuch://<absolute path>
+ */
+int validate_nm_default_url(const struct ConfigSet *cs, const struct ConfigDef *cdef, intptr_t value, struct Buffer *err)
```
gahr, line 48: static?

notmuch/config.c
Austin-Ray: Fixed

notmuch/config.c
Austin-Ray: Fixed

notmuch/private.h
Austin-Ray: My understanding is that it does since it's used in `notmuch/notmuch.c` and now `notmuch/config.c`. Am I missing something obvious?

notmuch/config.c
```diff
@@ -30,14 +30,40 @@
...
+
+/**
+ * nm_default_url_validator - Ensure nm_default_url is in form notmuch://<absolute path> - Implements ConfigDef::validator()
+ */
+static int nm_default_url_validator(const struct ConfigSet *cs, const struct ConfigDef *cdef, intptr_t value, struct Buffer *err)
+{
+  const char *uri = (const char *) value;
+  if (!is_valid_notmuch_url(uri))
+  {
+    mutt_buffer_printf(err, _("nm_default_url must be in form: notmuch://<absolute path> . Current value %s"), uri);
```
flatcap, line 53: It'd be nice to show the user the current value, but mind if we shorten it to just the hint?

On an 80 char screen, you're likely to see:
```
|                                                                                |
|nm_default_url must be in form: notmuch://<absolute path> . Current value notmuc|
|                                                                                |
```

notmuch/private.h
gahr: No, I was :)

notmuch/config.c
flatcap: Hmm...  That looks nice, but...

I wasn't thinking about config files, just the on-screen look.
Unfortunately, I've just tested that with newlines and it doesn't work.

There is a hack `mutt_window_reflow_message_rows()`, but intercepting the message would be a pain.

Austin-Ray: Ah, I see what you mean. When a user runs `:set nm_default_url`, right? I'll trim up the message.

flatcap: thanks

Austin-Ray: Trimmed the message a bit to reduce character count by 13.

`set nm_default_url="notmuch://"` now returns:

 ```
nm_default_url must be: notmuch://<absolute path> . Current: notmuch://
```

from:

```
nm_default_url must be in form: notmuch://<absolute path> . Current value notmuch://
```

flatcap (approved):

PR #2894, opened 2021-04-12
Show error if nm_default_uri{,/.notmuch} doesn't point to a valid dir

jindraj:
* **What does this PR do?**
Show error when directory provided in `nm_default_uri` doesn't exist, or isn't directory. The same applies for subdir `./.notmuch`

     for that)

#2891

### DIFF
--- a/notmuch/db.c
+++ b/notmuch/db.c
@@ -27,6 +27,7 @@
  */
 
 #include "config.h"
+#include <errno.h>
 #include <limits.h>
 #include <notmuch.h>
 #include <stdbool.h>
@@ -87,6 +88,19 @@ const char *nm_db_get_filename(struct Mailbox *m)
  */
 notmuch_database_t *nm_db_do_open(const char *filename, bool writable, bool verbose)
 {
+  struct stat st1;
+  char buf[PATH_MAX];
+  if (stat(mutt_path_concat(buf, filename, ".notmuch", sizeof(buf)), &st1) != 0)
+  {
+    mutt_error(_("Can't stat %s: %s"), buf, strerror(errno));
+    return NULL;
+  }
+  else if (!S_ISDIR(st1.st_mode))
+  {
+    mutt_error(_("%s is not a directory"), buf);
+    return NULL;
+  }
+
   notmuch_database_t *db = NULL;
   int ct = 0;
   notmuch_status_t st = NOTMUCH_STATUS_SUCCESS;


notmuch/db.c
```diff
@@ -102,6 +103,18 @@ notmuch_database_t *nm_db_do_open(const char *filename, bool writable, bool verb
...
+  char buf[PATH_MAX];
+  if (stat(filename, &st1) != 0)
+    mutt_error(_("Can't stat %s: %s"), filename, strerror(errno));
+  else if (!S_ISDIR(st1.st_mode))
+    mutt_error(_("%s is not a directory"), filename);
+  if (stat(mutt_path_concat(buf, filename, ".notmuch", sizeof(buf)), &st2) != 0)
+    mutt_error(_("Can't stat %s: %s"), filename, strerror(errno));
+  else if(!S_ISDIR(st2.st_mode))
+    mutt_error(_("%s is not a directory"), buf);
+
```
flatcap, line 117: A few suggestions:

- If any of the checks fail, I think we should error message, then return
- Move the checks to the top of the function
- Do we need both `struct stat`?

flatcap (changes requested):

notmuch/db.c
```diff
@@ -102,6 +103,18 @@ notmuch_database_t *nm_db_do_open(const char *filename, bool writable, bool verb
...
+  struct stat st2;
+  char buf[PATH_MAX];
+  if (stat(filename, &st1) != 0)
+    mutt_error(_("Can't stat %s: %s"), filename, strerror(errno));
+  else if (!S_ISDIR(st1.st_mode))
+    mutt_error(_("%s is not a directory"), filename);
+  if (stat(mutt_path_concat(buf, filename, ".notmuch", sizeof(buf)), &st2) != 0)
+    mutt_error(_("Can't stat %s: %s"), filename, strerror(errno));
+  else if(!S_ISDIR(st2.st_mode))
+    mutt_error(_("%s is not a directory"), buf);
```
Austin-Ray, line 116: Whenever I hit this conditional, NeoMutt locks up and must be killed externally. It appears to be in an infinite loop of trying to open the database, hitting this conditional, then trying again.

Here's a small script to reproduce it. Just change the paths accordingly.
```
mkdir -p /home/aray/tmp
touch /home/aray/tmp/.notmuch
echo 'set nm_default_uri="notmuch:///home/aray/tmp"' >>.neomuttrc.test
echo 'set spoolfile="notmuch://?query=tag:inbox"' >> .neomuttrc.test
./neomutt -n -F .neomuttrc.test
```

notmuch/db.c
Austin-Ray, line 117: Just a general comment, but we should reword the actual error messages to let the user know it's due to notmuch. Maybe something along the lines of `Unable to open notmuch database: can't stat /some/path`?

notmuch/db.c
flatcap: I've added some `return`s on the error paths.
NeoMutt doesn't lock up using your example any more.

notmuch/db.c
flatcap: @jindraj reused existing messages to avoid adding extra translations (there are 1500 already! :-)

notmuch/db.c
Austin-Ray: Awesome! I'll mark this as resolved.

notmuch/db.c
Austin-Ray: Hmm, that's a good point regarding translations. 

However, if I have `mailboxes "notmuch:///?query=tag:inbox` (note the extra `/`), it produces a rather unhelpful message `Can't stat /: No such file or directory`.  That could be caused by anything.

Perhaps it would be best to always stat the `.notmuch` folder so the message is `Can't stat /.notmuch: No such file or directory`. That at least gives the user an indication it's notmuch-related without increasing translation count.

Austin-Ray (approved):
LGTM. Just need to clean up the git history

flatcap (approved):

Issue #2895, opened 2021-04-12
 pager/pager.c:2386: mutt_pager: Assertion `pview->pdata->email' failed.


sstark:
## Expected Behaviour

View the message I have just finished writing in the builtin pager.

## Actual Behaviour

neomutt exits with:

`pager/pager.c:2386: mutt_pager: Assertion 'pview->pdata->email' failed.`

## Steps to Reproduce

1. Checkout neomutt master from git
2. `make distclean`
3. `./configure --prefix=/usr/local --gpgme --notmuch --with-ui --gnutls --gss --idn --sasl --kyotocabinet --disable-doc`
4. `make`
5. `./neomutt`
6. create new email ('m')
7. add recipient
8. write email in vim, save and exit vim
9. press 'ret' to view message 

## How often does this happen?

- Always

## When did it start to happen?

After bisecting, first bad commit turns out to be 04c6ddb371ee2f7ffdf5eaaca6d9c0fd8a612835 (pager: refactoring part  1)

e. g. 2b8e6b74142e18fbac302beb2972a7914765d31b does not show this behaviour.

## NeoMutt Version

NeoMutt 20210205-233-b2ba20
Copyright (C) 1996-2020 Michael R. Elkins and others.
NeoMutt comes with ABSOLUTELY NO WARRANTY; for details type 'neomutt -vv'.
NeoMutt is free software, and you are welcome to redistribute it
under certain conditions; type 'neomutt -vv' for details.

System: Linux 5.8.0-48-generic (x86_64)
ncurses: ncurses 6.2.20200212 (compiled with 6.2.20200212)
libidn: 1.33 (compiled with 1.33)
GPGME: 1.13.1-unknown
GnuTLS: 3.6.13
libnotmuch: 5.2.0
storage: kyotocabinet

Configure options: --prefix=/usr/local --gpgme --notmuch --with-ui --gnutls --gss --idn --sasl --kyotocabinet --disable-doc

Compilation CFLAGS: -g -O2 -std=c99 -D_ALL_SOURCE=1 -D_GNU_SOURCE=1 -D__EXTENSIONS__ -I/usr/local/include -DNCURSES_WIDECHAR -isystem /usr/include/mit-krb5

Default options:
  +attach_headers_color +compose_to_sender +compress +cond_date +debug 
  +encrypt_to_self +forgotten_attachments +forwref +ifdef +imap +index_color 
[...]

PR #2896, opened 2021-04-12
compose: pass Email to pager

flatcap:
The Pager is expecting an Email when viewing an attachment.

Fixes: #2895


ngortheone:
Either this bug was always there, only we didn't have assert to verify it or ...

@flatcap thanks!

PR #2899, opened 2021-04-16
formats: drop Mailbox param

flatcap:
Some Config items have format strings that allow Mailbox properties:

```sh
# Message to start a reply, 'On DATE, PERSON wrote:'
set attribution = "On %d, %n wrote:"

# Prefix message for forwarded messages
set forward_attribution_intro = "----- Forwarded message from %f -----"

# Suffix message for forwarded messages
set forward_attribution_trailer = "----- End forwarded message -----"

# printf-like format string to control the subject when forwarding a message
set forward_format = "[%a: %s]"

# String used to indent 'reply' text
set indent_string = "> "

# Suffix message to add after reply text
set post_indent_string = ""
```

Ultimately, these end in calls to `index_format_str()`.
By dropping the Mailbox parameter, we remove the possibility of using a few expandos

| Expando | Description                                                        |
| :------ | :----------------------------------------------------------------- |
| `%b`    | Filename of the original message folder (think mailbox)            |
| `%B`    | The list to which the email was sent, or else the folder name (%b) |
| `%m`    | Total number of message in the mailbox                             |

They aren't really meaningful for these Config items.
Only Email-specific attributes are, such as `%a` Author.

### DIFF
--- a/attachments.c
+++ b/attachments.c
@@ -252,6 +252,9 @@ static int count_body_parts(struct Body *body)
  */
 int mutt_count_body_parts(struct Mailbox *m, struct Email *e)
 {
+  if (!m || !e)
+    return 0;
+
   bool keep_parts = false;
 
   if (e->attach_valid)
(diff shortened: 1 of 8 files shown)


ngortheone:
Kill it with fire

ngortheone (approved):

gahr (approved):

PR #2906, opened 2021-04-20
Keep the Message alive in the Pager

### LINKED ISSUE #2751: Attachments downloaded twice
I use neomutt version 20191207, which comes by default on Ubuntu 20.04.

When I hit Enter on a message header, the attachments get downloaded (the increasing percentage is shown).

Then, when I attempt to open the attachments (using the "v" keyboard shortcut), they get downloaded all over again. This is an undesirable behavior, and increases bandwidth requirement and waiting time when the attachments are large. Is there anyway to get neomutt to save its downloaded attachment somewhere?

Or alternatively, is there a way to tell neomutt to not download attachments when one hits Enter on the message header, and let it download only when we hit "v"?

Thank you.

gahr:
So we can avoid re-fetching the message to view the attachments

This finally fixes #2751.

### DIFF
--- a/commands.c
+++ b/commands.c
@@ -213,7 +213,12 @@ int mutt_display_message(struct MuttWindow *win_index, struct MuttWindow *win_ib
   int res;
 
   struct Message *msg = mx_msg_open(m, e->msgno);
-  mutt_parse_mime_message(m, e, msg);
+  if (!msg)
+  {
+    return -1;
+  }
+
+  mutt_parse_mime_message(m, e, msg->fp);
   mutt_message_hook(m, e, MUTT_MESSAGE_HOOK);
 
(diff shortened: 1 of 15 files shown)


commands.c
```diff
@@ -373,6 +373,7 @@ int mutt_display_message(struct MuttWindow *win_index, struct MuttWindow *win_ib
     struct PagerView pview = { &pdata };
 
     pdata.email = e;
+    pdata.msg = msg;
```
flatcap, line 376: We'll need to set this to `NULL`, when we later:
```c
mx_msg_close(m, &msg);
```

(I can't add comments to code you haven't changed :-)

flatcap (changes requested):

PR #2908, opened 2021-04-21
Try to avoid redundant mx_msg_open calls

gahr:
This reduces the number of calls to mx_msg_open when viewing a message,
its attachments, or limiting the index, by trying to call mx_msg_open as
soon as possible and pass the resulting Message or FILE pointer down.
This has impact mostly on IMAP w/o `message_cachedir`.

Some net results:
- `<display-message>` then `<view-attachments>` fetches the message once
  instead of twice
- limiting with `|` or `&` with patterns that require a full message
  (e.g., `~M`) fetches the message once instead of once per branch of
  the conditional

### DIFF
--- a/attachments.c
+++ b/attachments.c
@@ -248,9 +248,10 @@ static int count_body_parts(struct Body *body)
  * mutt_count_body_parts - Count the MIME Body parts
  * @param m Mailbox
  * @param e Email
+ * @param fp File to parse
  * @retval num Number of MIME Body parts
  */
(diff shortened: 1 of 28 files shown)


flatcap (approved):
Tidied: [clang-format, doxy comments](https://github.com/neomutt/neomutt/compare/688d7797ceeee7aea71d2b2197a1eb324c55d060..ad86a726a46b7188267d51c03fb2add957829027)

PR #2917, opened 2021-04-23
Refactor notmuch window queries, improve documentation, and fix #2900

### LINKED ISSUE #2900: using nm_query_window_duration setting results in time window duration +1
## Expected Behaviour

yesterday i decided to give the nm_query_window_* settings a shot. i like the idea but the result was somewhat unexpected.
i used the following settings:

```
set nm_query_type = "threads"
set nm_query_window_duration = 1
set nm_query_window_timebase = "month"
bind index < vfolder-window-backward
bind index > vfolder-window-forward
```

i expect to see 1 months worth of emails when viewing a notmuch tag.
when using '<' and '>' to navigate from one month to the next, i expect to see 1 month of emails.

## Actual Behaviour

i always see 2 months worth of emails.
when viewing in april i see the following emails:
beginning of march to end of april

navigate back one time window:
beginning of february to end of march

navigate back one time window:
beginning of january to end of february

hence i see every month in duplicate when using the vfolder-window-* keys to navigate into the past.

to get the results i expected using notmuch search, i need to use:
```notmuch search date:0months..``` to list all emails in 'this month'.
this is also true for the other time units such as week. unfortunately it is impossible to set ```nm_query_window_duration = 0``` as that disables the feature.

when setting ```nm_query_window_duration = 2``` i see 3 months worth of emails and navigating 'back in time' moves 2 months at a time.

perhaps there is a setting i should have set or perhaps i misunderstood something in the manual.

## Steps to Reproduce

```
set nm_query_type = "threads"
set nm_query_window_duration = 1
set nm_query_window_timebase = "month"
bind index < vfolder-window-backward
bind index > vfolder-window-forward
```

- Always

- When I changed my config

```
set nm_query_type = "threads"
set nm_query_window_duration = 1
set nm_query_window_timebase = "month"
bind index < vfolder-window-backward
bind index > vfolder-window-forward
```

## NeoMutt Version

```
NeoMutt 20210205

System: Linux 5.11.10-gentoo-dist (x86_64)
GPGME: 1.14.0
OpenSSL: OpenSSL 1.1.1j  16 Feb 2021
libnotmuch: 5.2.0
storage: gdbm, bdb, lmdb

Configure options: --prefix=/usr --build=x86_64-pc-linux-gnu --host=x86_64-pc-linux-gnu --mandir=/usr/share/man --infodir=/usr/share/info --datadir=/usr/share --sysconfdir=/etc --localstatedir=/var/lib --docdir=/usr/share/doc/neomutt-20210205 --with-sysroot=/ --libdir=/usr/lib64 CCACHE=none --full-doc --enable-nls --enable-notmuch --enable-autocrypt --enable-gpgme --disable-pgp --disable-smime --enable-bdb --enable-gdbm --disable-kyotocabinet --disable-qdbm --disable-tokyocabinet --disable-idn --disable-gss --enable-lmdb --enable-sasl --with-ui=ncurses --sysconfdir=/etc/neomutt --enable-ssl --disable-gnutls --disable-testing

Compilation CFLAGS: -march=skylake -O2 -pipe -std=c99 -fno-delete-null-pointer-checks -D_ALL_SOURCE=1 -D_GNU_SOURCE=1 -D__EXTENSIONS__ -I//usr/include -DNCURSES_WIDECHAR -I/include -I//usr/include/

Default options:

Compile options:
  +getaddrinfo -gnutls +gpgme -gss +hcache -homespool -idn +inotify 
  -locales_hack -lua +meta -mixmaster +nls +notmuch +openssl -pgp +regex +sasl 
  -smime +sqlite +start_color +sun_attachment +typeahead 
MAILPATH="/var/mail"
SENDMAIL="/usr/sbin/sendmail"
SYSCONFDIR="/etc/neomutt"
```

## Extra Info

* Operating System and its version: gentoo up to date as of 2021-04-17
* Were you using multiple copies of NeoMutt at once?: no
* Were you using 'screen' or 'tmux'?: no
* Is your email local (maildir) or remote (IMAP)?: local

### PR DESCRIPTION
**Pull request by Austin-Ray**
This PR is a bit larger than I'd like, but I got ahead of myself. It's best understood by reading each commit message. However, there are four goals in this PR:

 - Refactor window query logic into two components, high-level and low-level, then add tests to protect against regressions.
 - Improve documentation of the feature
 - Fix #2900
 - Don't break any existing configurations so we're not burning our users' good will

Please note that commit 5b01866 is shown out of order on GitHub, it is the **last** commit in this series. I reordered the branch so the refactor comes before fixing #2900 for easier reverting if need be.

---

@neomutt/notmuch-users Please test this and let me know if I've broken a use-case, I've tried my best to test manually, but errors are always possible. If you can contribute a unit test for any broken use-case, I'd greatly appreciate it!

### DIFF
--- a/notmuch/config.c
+++ b/notmuch/config.c
@@ -108,6 +108,10 @@ static struct ConfigDef NotmuchVars[] = {
   { "nm_query_type", DT_STRING, IP "messages", 0, NULL,
     "(notmuch) Default query type: 'threads' or 'messages'"
   },
+  {
+    "nm_query_window_enable", DT_BOOL, false, 0, NULL,
+    "(notmuch) Enable query windows"
+  },
   { "nm_query_window_current_position", DT_NUMBER, 0, 0, NULL,
     "(notmuch) Position of current search window"
   },
--- a/notmuch/config.c
+++ b/notmuch/config.c
@@ -33,6 +33,7 @@
 #include <stdint.h>
 #include "mutt/lib.h"
 #include "notmuch/private.h"
+#include "notmuch/query.h"
 
 /**
  * is_valid_notmuch_url - Checks that a URL is in required form.
@@ -55,6 +56,32 @@ static int nm_default_url_validator(const struct ConfigSet *cs, const struct Con
   {
     mutt_buffer_printf(
         err, _("nm_default_url must be: notmuch://<absolute path> . Current: %s"), url);
+    return CSR_ERR_INVALID;
+  }
+
+  return CSR_SUCCESS;
+}
+
+/**
+ * nm_query_window_timebase_validator - Ensures nm_query_window_timebase matches allowed values - Implements ConfigDef::validator()
+ *
+ * Allowed values:
+ * - hour
+ * - day
+ * - week
+ * - month
+ * - year
+ */
+static int nm_query_window_timebase_validator(const struct ConfigSet *cs,
+                                              const struct ConfigDef *cdef,
+                                              intptr_t value, struct Buffer *err)
+{
+  const char *timebase = (const char *) value;
+  if (!nm_query_window_check_timebase(timebase))
+  {
+    mutt_buffer_printf(
+        err, _("Invalid nm_query_window_timebase value (valid values are: "
+               "hour, day, week, month or year)"));
     return CSR_ERR_INVALID;
   }
 
@@ -90,7 +117,7 @@ static struct ConfigDef NotmuchVars[] = {
   { "nm_query_window_duration", DT_NUMBER|DT_NOT_NEGATIVE, 0, 0, NULL,
     "(notmuch) Time duration of the current search window"
   },
-  { "nm_query_window_timebase", DT_STRING, IP "week", 0, NULL,
+  { "nm_query_window_timebase", DT_STRING, IP "week", 0, nm_query_window_timebase_validator,
     "(notmuch) Units for the time duration"
   },
   { "nm_record_tags", DT_STRING, 0, 0, NULL,
--- a/notmuch/query.c
+++ b/notmuch/query.c
@@ -121,7 +121,7 @@ enum NmQueryType nm_string_to_query_type_mapper(const char *str)
 }
 
 /**
- * query_window_check_timebase - Checks if a given timebase string is valid
+ * nm_query_window_check_timebase - Checks if a given timebase string is vali_
  * @param[in] timebase: string containing a time base
  * @retval true The given time base is valid
  *
@@ -130,7 +130,7 @@ enum NmQueryType nm_string_to_query_type_mapper(const char *str)
  *
  *     nm_query_window_timebase
  */
-static bool query_window_check_timebase(const char *timebase)
+bool nm_query_window_check_timebase(const char *timebase)
 {
   if ((strcmp(timebase, "hour") == 0) || (strcmp(timebase, "day") == 0) ||
       (strcmp(timebase, "week") == 0) || (strcmp(timebase, "month") == 0) ||
@@ -197,7 +197,7 @@ enum NmWindowQueryRc nm_windowed_query_from_query(char *buf, size_t buflen,
   int beg = duration * (cur_pos + 1);
   int end = duration * cur_pos;
 
-  if (!query_window_check_timebase(timebase))
+  if (!nm_query_window_check_timebase(timebase))
   {
     return NM_WINDOW_QUERY_INVALID_TIMEBASE;
   }
--- a/notmuch/query.c
+++ b/notmuch/query.c
@@ -119,3 +119,99 @@ enum NmQueryType nm_string_to_query_type_mapper(const char *str)
 
   return NM_QUERY_TYPE_UNKNOWN;
 }
+
+/**
+ * query_window_check_timebase - Checks if a given timebase string is valid
+ * @param[in] timebase: string containing a time base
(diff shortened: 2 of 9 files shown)

### CODE REVIEW

File: notmuch/config.c [hunk 1]
```diff
@@ -61,6 +62,32 @@ static int nm_default_url_validator(const struct ConfigSet *cs, const struct Con
...
+static int nm_query_window_timebase_validator(const struct ConfigSet *cs,
+                                              const struct ConfigDef *cdef,
+                                              intptr_t value, struct Buffer *err)
+{
+  const char *timebase = (const char *) value;
+  if (!nm_query_window_check_timebase(timebase))
+  {
+    mutt_buffer_printf(
+        err, _("Invalid nm_query_window_timebase value (valid values are: "
+               "hour, day, week, month or year)"));
```
flatcap, line 84: This needs some care.
I suggest quoting the valid strings, then adding an `L10N` comment for the translators,
as it's crucial that they don't translate 'hour', 'day', etc.

An alternative to a `DT_STRING`, is a `DT_ENUM`, currently only used by `$mbox_type`
https://github.com/neomutt/neomutt/blob/cd15d4e6ec3467d36c9cd6cc146ea6a64bb94267/mutt_config.c#L411
https://github.com/neomutt/neomutt/blob/cd15d4e6ec3467d36c9cd6cc146ea6a64bb94267/mx.c#L88-L101

flatcap (commented):
I'll have a close look tomorrow

notmuch/config.c
Austin-Ray: > This needs some care.
I suggest quoting the valid strings, then adding an L10N comment for the translators, as it's crucial that they don't translate 'hour', 'day', etc.

Do you have an example code segment that I could look at for correctness? I actually lifted that string from `notmuch.c`.

>  An alternative to a DT_STRING, is a DT_ENUM, currently only used by $mbox_type

That's actually what I mentioned on IRC for a "good first issue".  It's an easy enough change for someone who isn't super familiar with the codebase to do.

notmuch/query.c
```diff
@@ -119,3 +119,112 @@ enum NmQueryType nm_string_to_query_type_mapper(const char *str)
 
   return NM_QUERY_TYPE_UNKNOWN;
 }
+
+/**
+ * query_window_check_timebase - Checks if a given timebase string is valid
```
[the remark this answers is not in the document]
Austin-Ray: fixup'd

notmuch/config.c
flatcap: > don't translate 'hour'

Thinking about it a bit more, perhaps we should go for a more enum-compatible error message.
How about:
```
nm_query_window_timebase: value must be one of: hour, day, week, month, year
```

We can code this error as:

```c
  /* L10N: Warning message when setting an enumeration, e.g.
     "mbox_type: value must be one of: mbox, mmdf, mh, maildir" */
  _("%s: value must be one of: %s"), "nm_query_window_timebase", "hour, day, week, month, year"
```

The `L10N` comment is a free-form instruction that's copied verbatim to the `po/*.po` files.
The config name and comma-separated values are not translated.

> "good first issue"

Ah!  Good :-)
And a good second issue would be to improve the error message for enums
by automatically generating the warning message above.

Austin-Ray: > Thinking about it a bit more, perhaps we should go for a more enum-compatible error message.
> The config name and comma-separated values are not translated.

While the values aren't translated, how do we convey the equivalent time concept to a non-English speaker? Or should we assume some level of English understanding for non-English speakers?

For `mbox_type`, its values are all technical terms so the enum message works perfectly. However, `nm_query_window_timebase` should let the user know that week means, well, a week even though we only accept the English word for it.

Our existing translation aren't error proof in their implementation. Translations, such as German (below) and Japanese, include accepted English values with translated values next to them while others (such as Czech and Spanish) translate the values.

German:
```
Ungültiger Wert für nm_query_window_timebase (gültige Werte: hour (Stunde), day (Tag), week (Woche), month (Monat), year (Jahr))
```

Austin-Ray: After further thought on the enumerations translations, we can include a `L10N` comment for concepts requiring translation but keeping accepted values untranslated.

For example:
```
/* L10N: Warning message when setting an enumeration.
   Please kept untranslated values alongside translation e.g. hour (Stunde), day (Tag), week (Woche), month (Monat), year (Jahr)*/
```

That seems a perfect use-case for `L10N` comment calling focus to an edge-case in translation.

flatcap:
I rebased this PR and made a few minor non-code fixes.
Unfortunately, I can't update your branch.

I suspect you unticked the box "**Allow edits by maintainers**"
It's at the bottom of the right-hand sidebar.

Austin-Ray:
> a few minor non-code fixes.

What changes are needed? I can make them

flatcap:
I've pushed the changes to [[devel/issue-2900]](https://github.com/neomutt/neomutt/tree/devel/issue-2900)

Austin-Ray:
Updated to modifications in that branch

Austin-Ray:
Anything required for this PR?

Austin-Ray:
Ping

flatcap (approved):
Sorry this has taken so long.

flatcap:
I can't update this branch, so I've rebased and merged it locally.
Next time, **please** create the branch on the neomutt repo,
or at least ensure that the "Enable maintainer updates" tickbox is ticked.

PR #2921, opened 2021-04-28
index: Escape '+' in Message-Id for <mark-message>

jpoimboe:
If the Message-Id has '+' in it, `<mark-message>` fails to work because
the '+' is interpreted as a regex operator.

Escape the '+' to make the Message-Id string a valid regex.

### DIFF
--- a/index/index.c
+++ b/index/index.c
@@ -3830,11 +3830,21 @@ struct Mailbox *mutt_index_menu(struct MuttWindow *dlg, struct Mailbox *m_init)
               buf2[0])
           {
             char str[256], macro[256];
+	    char *s = macro;
             const char *const c_mark_macro_prefix =
                 cs_subset_string(shared->sub, "mark_macro_prefix");
             snprintf(str, sizeof(str), "%s%s", c_mark_macro_prefix, buf2);
             snprintf(macro, sizeof(macro), "<search>~i \"%s\"\n",
                      shared->email->env->message_id);
+
+            /* Escape '+' to make the Message-Id string a valid regex. */
+            while (*s) {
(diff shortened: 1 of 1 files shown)


gahr:
Thanks for spotting this. Unfortunately, the issue is wider than this. Message IDs can also contain other regex-sensitive characters, such as `^`. Also, the escapes must be places inside a single-quoted string. See #2925 for a proposed fix.

jpoimboe:
In practice, I've never seen `^` in Message-Id.  I'm not sure I get your point about escapes needing to be inside a single-quoted string.  My patch fixed the issue.  But your approach is clearly better.  I didn't know about mutt_file_sanitize_regex().

Thanks!

gahr:
Yeah, not saying I get message ids with weird chars everyday, but as we're fixing it, let's fix the whole of it. See the productions for msg-id here: https://tools.ietf.org/html/rfc5322#section-3.6.4.

Thanks!

gahr:
>  I'm not sure I get your point about escapes needing to be inside a single-quoted string

We need the escapes to go all the way down to the regex call. If we put them in double-quotes, the NeoMutt parser will unescape them. For non-special characters, `\<char>` will be translated into `<char>`, so the regex engine will not see the backslashes.

jpoimboe:
> We need the escapes to go all the way down to the regex call. If we put them in double-quotes, the NeoMutt parser will unescape them. For non-special characters, `\<char>` will be translated into `<char>`, so the regex engine will not see the backslashes.

I see what you mean.  It's cleaner to use single quotes in the macro, though my patch still worked with double quotes in the macro because I escaped the `\` characters.

Issue #2928, opened 2021-05-05
neomutt dies with SIGFPE when trying to open mail


l29ah:
## Actual Behaviour

```
#0  0x000055d9bb027315 in mutt_color_quote (q=q@entry=0) at gui/color.c:1486
#1  0x000055d9bb006443 in classify_quote (quote_list=quote_list@entry=0x7ffd30798238, 
    qptr=0x55d9bd147610 "> On Thu, Mar 25, 2021 at 01:12:38AM +0300, Alexey Belovushchenko wrote:\n", length=1, 
    force_redraw=force_redraw@entry=0x7ffd30798228, q_level=q_level@entry=0x7ffd30798234) at pager/pager.c:655
#2  0x000055d9bb007b5b in resolve_types (
    buf=0x55d9bd147610 "> On Thu, Mar 25, 2021 at 01:12:38AM +0300, Alexey Belovushchenko wrote:\n", 
    raw=0x55d9bd15cd50 "> On Thu, Mar 25, 2021 at 01:12:38AM +0300, Alexey Belovushchenko wrote:\n", 
    line_info=0x55d9bd14bbd0, n=n@entry=15, last=16, quote_list=quote_list@entry=0x7ffd30798238, 
    q_level=0x7ffd30798234, force_redraw=0x7ffd30798228, q_classify=true) at pager/pager.c:1138
#3  0x000055d9bb00872d in display_line (fp=0x55d9bd122140, last_pos=last_pos@entry=0x7ffd30798240, 
    line_info=line_info@entry=0x7ffd307983a0, n=15, last=last@entry=0x7ffd3079821c, max=max@entry=0x7ffd30798218, 
    flags=66, quote_list=0x7ffd30798238, q_level=0x7ffd30798234, force_redraw=0x7ffd30798228, 
    search_re=0x7ffd30798258, win_pager=0x55d9bd146ee0) at pager/pager.c:1789
#4  0x000055d9bb009268 in pager_custom_redraw (pager_menu=0x55d9bd14c3c0) at pager/pager.c:2182
#5  0x000055d9bb00a3c2 in mutt_pager (pview=pview@entry=0x7ffd30798d10) at pager/pager.c:2570
#6  0x000055d9bafbf0b0 in mutt_display_message (win_index=0x55d9bd11a550, win_ibar=0x55d9bd11a5f0, 
[...]

PR #2929, opened 2021-05-05
fix crash with uncoloured quotes

flatcap:
Fixes: #2928

If the user hasn't configured any quoted-email colours, then NeoMutt tries to calculate a value modulo zero!


Austin-Ray (approved):
LGTM

PR #2930, opened 2021-05-07
GH Actions: Added workflows and Dockerfiles

nixMark:
**Currently this PR is in draft state, further additions are expected.**

* **What does this PR do?**
This PR is adding actions capabilities to the repository to move from Circle CI to GitHub actions. As part of this migration this PR moves into docker based builds to ensure easier migration to another build platforms in case of need, docker ensures platform agnostic build system.
 
N/A

     for that)

### DIFF
--- a/.github/workflows/base.yml
+++ b/.github/workflows/base.yml
@@ -1,0 +1,30 @@
+name: Base image build
+
+on: 
+  workflow_dispatch:
+    branches: 
+     - '*'
+     - '!flatcap'
+
+jobs:
+  build-and-push:
+    runs-on: ubuntu-latest
+    steps:
+    - name: Checkout
+      uses: actions/checkout@v2
+      with:
+        fetch-depth: 1  
+    - name: Login to GitHub Container Registry
+      uses: docker/login-action@v1 
+      with:
+        registry: ghcr.io
+        username: ${{ github.repository_owner  }}
+        password: ${{ secrets.GH_TOKEN }}
+    - name: Build and push
+      uses: docker/build-push-action@v2
+      with:
+        file: base.Dockerfile
+        platforms: linux/amd64
+        push: true
+        tags: |
+            ghcr.io/nixmark/neomutt-base:latest
--- a/.github/workflows/base.yml
+++ b/.github/workflows/base.yml
@@ -19,7 +19,7 @@ jobs:
       with:
         registry: ghcr.io
         username: ${{ github.repository_owner  }}
-        password: ${{ secrets.GH_TOKEN }}
+        password: ${{ secrets.GITHUB_TOKEN }}
     - name: Build and push
       uses: docker/build-push-action@v2
       with:
--- a/.github/workflows/base.yml
+++ b/.github/workflows/base.yml
@@ -8,7 +8,7 @@ on:
 
 jobs:
   build-and-push:
-    runs-on: ubuntu-latest
+    runs-on: ubuntu-20.04
     steps:
     - name: Checkout
       uses: actions/checkout@v2
@@ -27,4 +27,4 @@ jobs:
         platforms: linux/amd64
         push: true
         tags: |
-            ghcr.io/nixmark/neomutt-base:latest
+            ghcr.io/neomutt/neomutt-base:latest
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -1,0 +1,28 @@
+name: neomutt CI
+
+on: 
+  workflow_dispatch:
+    branches: 
+     - '*'
+     - '!flatcap'
+
+jobs:
+  build:
+    runs-on: ubuntu-latest
+    steps:    
+    - name: Login to GitHub Container Registry
+      uses: docker/login-action@v1 
+      with:
+        registry: ghcr.io
+        username: ${{ github.repository_owner  }}
+        password: ${{ secrets.GH_TOKEN }}
+    - name: Checkout
+      uses: actions/checkout@v2
+      with:
+        fetch-depth: 1  
+    - name: Build and push
+      uses: docker/build-push-action@v2
+      with:
+        file: build.Dockerfile
+        platforms: linux/amd64
+        push: false
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -8,7 +8,7 @@ on:
 
 jobs:
   build:
-    runs-on: ubuntu-latest
+    runs-on: ubuntu-20.04
     steps:    
     - name: Login to GitHub Container Registry
       uses: docker/login-action@v1 
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -15,7 +15,7 @@ jobs:
       with:
         registry: ghcr.io
         username: ${{ github.repository_owner  }}
-        password: ${{ secrets.GH_TOKEN }}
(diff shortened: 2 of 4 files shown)


.github/workflows/base.yml
```diff
@@ -0,0 +1,30 @@
...
+    - name: Checkout
+      uses: actions/checkout@v2
+      with:
+        fetch-depth: 1  
+    - name: Login to GitHub Container Registry
+      uses: docker/login-action@v1 
+      with:
+        registry: ghcr.io
+        username: ${{ github.repository_owner  }}
+        password: ${{ secrets.GH_TOKEN }}
```
Austin-Ray, line 22: Is this a user-defined secret? The docs recommend using `GITHUB_SECRET` for authentication[0]. That looks like a less error-prone authentication method.

[0] https://docs.github.com/en/actions/reference/authentication-in-a-workflow

.github/workflows/main.yml
```diff
@@ -0,0 +1,28 @@
...
+jobs:
+  build:
+    runs-on: ubuntu-latest
+    steps:    
+    - name: Login to GitHub Container Registry
+      uses: docker/login-action@v1 
+      with:
+        registry: ghcr.io
+        username: ${{ github.repository_owner  }}
+        password: ${{ secrets.GH_TOKEN }}
```
Austin-Ray, line 18: Same question about authentication

base.Dockerfile
```diff
@@ -0,0 +1,17 @@
+FROM  ubuntu
```
Austin-Ray, line 1: We should probably pin this to a specific tag so the CI environment stays constant across all runs. It's possible for the Ubuntu image to update and break the CI.

.github/workflows/base.yml
```diff
@@ -0,0 +1,30 @@
...
+    - name: Build and push
+      uses: docker/build-push-action@v2
+      with:
+        file: base.Dockerfile
+        platforms: linux/amd64
+        push: true
+        tags: |
+            ghcr.io/nixmark/neomutt-base:latest
```
Austin-Ray, line 30: Shouldn't the namespace be `neomutt` ?

build.Dockerfile
```diff
@@ -0,0 +1,11 @@
+FROM  ghcr.io/nixmark/neomutt-base:latest
```
Austin-Ray, line 1: Same namespace comment

Austin-Ray (commented):
Do GitHub actions run in parallel? If so, won't that create a race condition between the `base` and `main` workflows where the base image may update but the main workflow uses an old image?

.github/workflows/base.yml
nixMark: It is the same thing, you have to get this token from GH with needed access rights and put into repository secrets.

base.Dockerfile
nixMark: It is picking latest, but it can be changed to preferred one. 20.04 LTS may be?

.github/workflows/base.yml
nixMark: Yes, this is my bad. I tested everything in my repo, and forgot to fix this. I will fix in next commit.

nixMark:
For the first time, we will have to run the base image build and push to the registry. For later builds we will run it once a month or so.

base.Dockerfile
Austin-Ray: > It is picking latest, but it can be changed to preferred one.

It's less about preference and more about stability of the CI pipeline. The `latest` tag points to `20.04`[0], but Canonical can always change that to a version with a breaking change (say change in package naming) and now we have to fix our CI pipeline, taking up valuable developer time. 

> 20.04 LTS may be?

An LTS is fine.

[0] https://hub.docker.com/_/ubuntu/

Austin-Ray:
> For the first time, we will have to run the base image build and push to the registry. For later builds we will run it once a month or so.

Ah, okay, I see. I misunderstood the `workflow_dispatch` keyword. I thought it'd trigger both pipeline for every build.

Thanks for putting in the work on this!

flatcap:
What's needed to get this working?
Have you set up the secrets?
What triggers the building of the docker image?
Will that get us an "Actions" tab in this PR?
Or will that only happen when it's committed?
If so, how do we test it?

.github/workflows/base.yml
nixMark: Yes.

base.Dockerfile
nixMark: changing it to 20.04 LTS

nixMark:
_> What's needed to get this working?_ 
Adding token should be enough.
_> Have you set up the secrets?_ 
I will do it today.
_> What triggers the building of the docker image?_ 
As of this PR, it is manual, for testing, later can be changed to pull requests or commits.
_> Will that get us an "Actions" tab in this PR?_ 
Yes, these workflows should appear in Actions tab.
_> Or will that only happen when it's committed?_
After the PR is merged.
_> If so, how do we test it?_
After merge we will have to run base first to create the base image for second, build workflow. Names can be changed as wished.

flatcap:
> I will do it today.

Great!  Thanks!
I can't wait to see this working.

I'll be around all day if you have any questions.

nixMark:
Austin was right. My mistake. Creating Personal Access Token for Organization is not possible. `GITHUB_TOKEN` is the token that is created by default and has access to push to the GH Packages registry. I am testing it in my private repo to test if it works as expected.

nixMark:
As of now I am facing an issue with pushing process of image. If I won't be able to find a fix to this with the plugins, will move everything to custom commands.

flatcap:
> an issue with pushing process of image

I don't see any secrets yet, might that be the cause?
(either on the neomutt repo, or the neomutt org)

nixMark:
Documentation says: 

> GitHub automatically creates a GITHUB_TOKEN secret to use in your workflow. You can use the GITHUB_TOKEN to authenticate in a workflow run.

Actions is able to login through it: (Don't mind the nixMark part, I am checking in my own account)
```
Run docker/login-action@v1
  with:
    registry: ghcr.io
    username: nixMark
    password: ***
    logout: true
Logging into ghcr.io...
Login Succeeded!
```
But when push phase comes:

```
#7 pushing nixmark/neomutt-base:latest with docker
#7 sha256:3986290f677b135ffa22d379bf2ac92d4dabae8447b199f72e3c4ddeded96657
#7 pushing layer 87a46fcc7b39
#7 pushing layer feef05f055c9
#7 pushing layer 87a46fcc7b39 0.2s done
#7 pushing layer feef05f055c9 0.2s done
#7 ERROR: denied: requested access to the resource is denied
------
 > pushing nixmark/neomutt-base:latest with docker:
------
error: denied: requested access to the resource is denied
Error: buildx failed with: error: denied: requested access to the resource is denied
```

I was using the kinda plugins, but will do it with direct commands.

flatcap:
Hi @nixMark,

Good news and bad news.

Good news: @rprimus took your configs and got them all working.
(it took a couple of days' reading of the docs)

We now have plain actions and a Docker action:

- https://github.com/neomutt/neomutt-docker-build/actions
- https://github.com/neomutt/neomutt/tree/master/.github/workflows
- https://github.com/neomutt/neomutt/actions

Bad news: We don't need this PR any more.

:heart: Thank you very much for your **hard work**.
It's really appreciated :-)

PR #2941, opened 2021-05-26
notmuch: refactor tag parsing

Austin-Ray:
Refactors the three spots of tag parsing into a single low-level, unit tested function. This removes redundancy and makes the higher-level functions easier to understand.

This is the first time I've used the array API so hopefully it's correct.

### DIFF
--- a/Makefile.autosetup
+++ b/Makefile.autosetup
@@ -504,7 +504,7 @@ $(PWD)/nntp:
 LIBNOTMUCH=	libnotmuch.a
 LIBNOTMUCHOBJS=	notmuch/config.o notmuch/db.o notmuch/notmuch.o \
 		notmuch/adata.o notmuch/edata.o notmuch/mdata.o \
-		notmuch/query.o
+		notmuch/query.o notmuch/tag.o
 CLEANFILES+=	$(LIBNOTMUCH) $(LIBNOTMUCHOBJS)
 ALLOBJS+=	$(LIBNOTMUCHOBJS)
 
--- a/notmuch/lib.h
+++ b/notmuch/lib.h
@@ -34,6 +34,7 @@
  * | notmuch/mdata.c    | @subpage nm_mdata   |
  * | notmuch/notmuch.c  | @subpage nm_notmuch |
  * | notmuch/query.c    | @subpage nm_query   |
+ * | notmuch/tag.c      | @subpage nm_tag     |
  */
 
 #ifndef MUTT_NOTMUCH_LIB_H
--- a/notmuch/notmuch.c
+++ b/notmuch/notmuch.c
@@ -1094,36 +1094,23 @@ static void sync_email_path_with_nm(struct Email *e, notmuch_message_t *msg)
 
 /**
  * update_tags - Update the tags on a message
- * @param msg  Notmuch message
- * @param tags String of tags (space separated)
+ * @param msg     Notmuch message
+ * @param tag_str String of tags (space separated)
  * @retval  0 Success
  * @retval -1 Failure
  */
-static int update_tags(notmuch_message_t *msg, const char *tags)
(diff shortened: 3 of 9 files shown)


Austin-Ray:
> the tests leak memory
> (build with --asan to see)

That's a helpful configure flag. 

Fixed half the messages through `ARRAY_FREE()`, but to get the final messages to go away, I had to loop and free each element. Is that to be expected with the array API or just when using `char *` elements?

> please add the new file to po/POTFILES.in
> (list of all source that could contain translations)

Done

flatcap:
> > (build with --asan to see)
> 
> That's a helpful configure flag.

Yeah.  ([More detail about asan](https://neomutt.org/dev/analysis/asan))

> I had to loop and free each element.
> Is that to be expected with the array API or just when using `char *` elements?

That's expected when you alloc memory (`strdup()`).

The `ARRAY` is quite simple, but that makes it quite flexible.
It maintains a set of "boxes" in a contiguous block of memory.
This is efficient and allows us to `qsort()` the entries.

In your case the "box" just contains a pointer to an alloc'd string.

By being agnostic to its contents, the `ARRAY` could equally contain pointer to strings that are owned by someone else -- (no alloc or dealloc required).

Note, the "box" could be any size.
In some places we have an `ARRAY` of an entire `struct` (See, [`AliasViewArray`](https://github.com/neomutt/neomutt/blob/master/alias/gui.h#L31-L46))

To free the `ARRAY`, perhaps we could add an `array_free_members()` function.
It would take a 'free' function as a parameter.

Austin-Ray:
Hmm, I think my array issue arises from the API documentation. `ARRAY_FREE` states

> ARRAY_FREE - Release **all** memory

So I assumed it would free any heap-allocated memory.

Would that function be better described as a way to empty the array of its elements? If so, I'll add some documentation clarifying it (in this PR or another, your call) so future users of the API won't make the same mistake. 

> To free the ARRAY, perhaps we could add an array_free_members() function.

I think that's a valuable addition to the API. I'm interested on implementation since I'd like to learn more about macros.

flatcap:
> ... your call ...

It's @gahr's baby, so he gets the casting vote.

flatcap:
The oracle has spoken!  (and I just about understand :-)

> we could add an `array_free_members()` function

`mutt/array.h` doesn't have any functions, just macros, so my first idea isn't any good.

Adding a macro that performs an action on each member, would be possible, but could be tricky to implement.
See how the scope is managed in `ARRAY_FOREACH_FROM_TO()`.

> I assumed it would free any heap-allocated memory

It's just freeing the block of memory that **is** the array (and zeroing the sizes).
Feel free to make the comment a bit clearer.

> FOREACH ... free(tag)

@gahr offered a different solution...

What's the lifespan of the tags expected to be?
Will new tags be added to this array (or removed)?
If not, then how about:

```c
struct TagArray
{
  ARRAY_HEAD(, char *) tags;
  char *str;
};
```

When you set up the array, you
- `strdup()` the original tag string
- Save this string in the struct (`str`)
- punch holes (`NUL` chars) into the string to split it
- Add pointers to the split string into the array

To free the struct, simply:
```c
ARRAY_FREE(tags.tags);
FREE(&tags.str);
```

Austin-Ray:
> @gahr offered a different solution...

Good solution, I'll implement it.

Austin-Ray:
Pushed two more fixup commits. They move the manual clean up by the callers into a dedicated function. Makes the clean-up less error prone as users of `TagArray` can call this function instead of remembering if they free'd everything.

Austin-Ray:
@flatcap: Is there anything else needed for this PR or should I rebase the history for merging?

flatcap:
No, go ahead

Austin-Ray:
> No, go ahead

Done

flatcap (approved):
Just a bit of light tidying to keep doxygen happy.

PR #2946, opened 2021-05-31
create libprogress

flatcap:
Refactor the Progress Bar so that is has an explicit creation and destruction.

The progress bar is one of the **active** users of the Message Window.
When there's no active user, the Message Window may be used for informational messages.

- 6df764e7d create libprogress
  Backends shouldn't use code that isn't in a library (some day :-)

- a9c557b22 tidy naming
  shorten a few function names

- a1caef8a8 add new, free
  Dynamically allocate/destroy a progress bar

- 874529f3b use new/free
  Convert the simple callers to use new/free

- 58fea6430 convert notmuch to use new/free
  Notmuch's use of the progress bar was non-trivial (notes below)

- 502742b1f hide implementation
  Make the progress bar into an opaque type

---

### Conversions

There are over 30 users of the progress bar.
Most of the conversions were simple: create the bar, use it, free it.

Some of the converted functions had multiple exits, so I needed to free the bar in several places.

`smtp_data()` had a lot of `return` statements.
This needed a little refactoring to keep the code simple.

### Notmuch

The changes to Notmuch have actually simplified the code somewhat.

The only function that updates the progress bar is `append_message()`.
The call tree to that function looks like:

[source gv](https://gist.github.com/flatcap/393f6e222f29209026fd048dec1dfb41/raw/ce8e0a3f7f660a5cef4ba9dd16a150841092e282/n1.gv)

The progress bar pointer is held in `Mailbox->mdata`.
Before, the code had a bunch of variables to decide whether the progress bar was being used.
Now, we're more explicit -- if the pointer is set, then use it.

In keeping with the old code:
- `nm_mbox_check()` doesn't use the progress bar
- `nm_mbox_open()` uses the progress bar
- `nm_read_entire_thread()` uses the progress bar


gahr:
> Now, we're more explicit -- if the pointer is set, then use it.

At the cost of having to allocate the Progress structure on the heap and explicitely free it. What is the downside you were seeing in having it by value? I can think of
```c
struct Progress progress = { 0 };
...
progress_init(&progress, ...); // something inside progress is made "active"
progress_update(&progress, ...); // checks this "active" state
...
// no more free
```

flatcap:
**tl;dr**
- gui stuff in the struct
- need to know when progress is finished

---

> What is the downside you were seeing in having it by value?

There are two problems that will affect me soon: dependencies; ownership of the Message Window.

I'm busy refactoring how the windows are painted.  Generally, each window will have some data and be responsible for repainting itself, say if its geometry changes.

The Message Window is a tricky exception.  Much of the time, it's just passively relaying error messages, or showing "tag-" when you hit `<tag-prefix>`.  However, there are some active users like `mutt_yesorno()`.  The active users take control of the Message Window and have their own event loop.  They wait for user input and make sure that the screen is refreshed.

The progress bar is a bit of a pain.  It's not quite active (no event loop), but it does require refreshing the screen frequently.  I'd like to use `progress_update()` like an event loop.

**Dependencies**

The progress bar is used by lots of code deep in the backends.
Currently, it doesn't contain anything that would cause problems, but I'd like to add some gui members to the struct.
(at least a pointer to the window)

**Ownership**

When the active windows take control of the Message Window, they `window_set_focus()` it. (in my work branch)
This does two things: stop the Message Window from drawing itself; tell the Help Bar to update itself.
When the active window is done, it restores the focus.

The progress bar needs to restore the focus when it's done.  It needs an explicit finish.

gahr (approved):

Austin-Ray (approved):
LGTM

PR #2948, opened 2021-06-01
msgwin: factor out MessageWindow

flatcap:
### Overview

Make the MessageWindow self-contained.
**This PR** is quite simple, but it explains the changes coming in the next PR.

- 6ea96d9ca factor out MessageWindow
  Move the Message Window into separate source file.

- 4cba5ca3b let the MessageWindow paint itself
  Give the MessageWindow `recalc()` and `repaint()` functions.

- 17dac7ca2 add helper functions
  Future changes to the MessageWindow will be done using these functions.

- 03119e38b Convert the code to use `msgwin_set_height()`

- 7c61d8c01 Convert the code to use `msgwin_get_width()`

- 4a488b733 Convert the code to use `msgwin_clear_text()`

- 65965f8f5 Convert the code to use `msgwin_set_text()`

- 17bbd4e5f ask for a MessageWindow pointer
  The active users must ask for a pointer to the MessageWindow.

- 4161a1377 convert the Progress Bar to use `msgwin_get_window()`

- fb682b86a hide the details of the MessageWindow
  Nobody needs the MessageWindow pointer any more.

---

The MessageWindow is used by a lot of different parts of the code.
However, the users fall into two types: Active, Passive.

### Passive Users

Passive users don't care about the details of the MessageWindow.
They just want to display some text (and clear it up later).

Examples:
- Logging system (`mutt_message()`, etc)
- Index printing 'tag-' on `<tag-prefix>`

Passive users expect that if the screen is resized or needs repainting, that it will "just happen".

### Active Users

Active users are characterised by an event loop.
They take the focus and wait for user input.

Examples:
- `mutt_yesorno()` - "Are you sure [Y/n]?"
- `mutt_buffer_get_field()` - "To: "

Active users are expected to handle:
- user input
- `SIGWINCH` (window resizing)
- Re-flowing, calculating, painting of all the windows

### Progress Bar

Technically, the progress bar is passive -- it just displays some text.
However, calling `progress_update()` forces a screen update, so it's behaving like an event loop.

As an example, imagine the IMAP code reading all the headers.
Here's how I expect it to work:

- Code calls `progress_new()`
  Progress Bar does `set_focus(MessageWindow)`

- Code loops
  - Code does some work
  - Code calls `progress_update()`
    Progress Bar updates itself and calls `window_redraw()`

  - Code does some work
  - Code calls `progress_update()`
    Progress Bar updates itself and calls `window_redraw()`

  - ....

  - User unexpectedly resizes the window

  - Code does some work
  - Code calls `progress_update()`
    Progress Bar detects `SIGWINCH` and calls:
    - `mutt_resize_screen()`
    - `mutt_window_set_root()`
    - `mutt_window_reflow()`
    - `window_notify_all()`
    Progress Bar updates itself and calls `window_redraw()`
      - Relevant windows get `recalc()` called
      - Relevant windows get `repaint()` called

  - ...

- Code calls `progress_free()`
  Progress Bar is destroyed, focus is returned to the previous state

### DIFF
--- a/gui/msgwin.c
+++ b/gui/msgwin.c
@@ -35,6 +35,8 @@
 #include "mutt_curses.h"
 #include "mutt_window.h"
 
+struct MuttWindow *MessageWindow = NULL; ///< Message Window, ":set", etc
+
 /**
  * struct MsgWinPrivateData - Private data for the Message Window
  */
@@ -119,6 +121,7 @@ static int msgwin_window_observer(struct NotifyCallback *nc)
       return 0;
 
     notify_observer_remove(NeoMutt->notify, msgwin_window_observer, win_msg);
+    MessageWindow = NULL;
     mutt_debug(LL_DEBUG5, "window delete done\n");
   }
   return 0;
@@ -169,6 +172,8 @@ struct MuttWindow *msgwin_create(void)
   win->repaint = msgwin_repaint;
 
   notify_observer_add(win->notify, NT_WINDOW, msgwin_window_observer, win);
+
+  MessageWindow = win;
   return win;
 }
 
--- a/gui/msgwin.c
+++ b/gui/msgwin.c
@@ -1,0 +1,43 @@
+/**
+ * @file
+ * Message Window
+ *
+ * @authors
+ * Copyright (C) 2021 Richard Russon <rich@flatcap.org>
+ *
+ * @copyright
+ * This program is free software: you can redistribute it and/or modify it under
+ * the terms of the GNU General Public License as published by the Free Software
+ * Foundation, either version 2 of the License, or (at your option) any later
+ * version.
+ *
+ * This program is distributed in the hope that it will be useful, but WITHOUT
+ * ANY WARRANTY; without even the implied warranty of MERCHANTABILITY or FITNESS
+ * FOR A PARTICULAR PURPOSE.  See the GNU General Public License for more
+ * details.
+ *
+ * You should have received a copy of the GNU General Public License along with
+ * this program.  If not, see <http://www.gnu.org/licenses/>.
+ */
+
+/**
+ * @page gui_msgwin Message Window
+ *
+ * Message Window for displaying messages and text entry.
+ */
+
+#include "config.h"
+#include "mutt_window.h"
+
+/**
+ * msgwin_create - Create the Message Window
+ * @retval ptr New Window
+ */
+struct MuttWindow *msgwin_create(void)
+{
+  struct MuttWindow *win =
+      mutt_window_new(WT_MESSAGE, MUTT_WIN_ORIENT_VERTICAL, MUTT_WIN_SIZE_FIXED,
+                      MUTT_WIN_SIZE_UNLIMITED, 1);
+
+  return win;
+}
--- a/gui/msgwin.c
+++ b/gui/msgwin.c
@@ -27,7 +27,130 @@
  */
 
 #include "config.h"
+#include "mutt/lib.h"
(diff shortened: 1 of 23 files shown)


gui/msgwin.c
```diff
@@ -0,0 +1,249 @@
...
+  if (!mutt_window_is_visible(win))
+    return 0;
+
+  struct MuttWindow *win_focus = window_get_focus();
+  if (!win_focus)
+    return 0;
+
+  // If the MessageWindow is focussed, then someone else is actively using it.
+  if (win_focus == win)
+    return 0;
```
Austin-Ray, line 87: Since this block of code appears twice and neither occurrences uses `win_focus` outside of the checks, it makes sense to move it to a well-named function so easier to focus on the important logic:

```C
static bool mutt_window_is_focused(MuttWindow *win)
{
   struct MuttWindow *win_focus = window_get_focus();
    if (!win_focus)
      return false;

    // If the MessageWindow is focussed, then someone else is actively using it.
    if (win_focus == win)
      return false;
        
    return true;
}

static int msgwin_repaint(struct MuttWindow *win)
{
  if (!mutt_window_is_visible(win) || !mutt_window_is_focused(win))
    return 0;

  struct MsgWinPrivateData *priv = win->wdata;

  mutt_window_move(win, 0, 0);

  mutt_curses_set_color(priv->color);
  mutt_window_move(win, 0, 0);
  mutt_paddstr(win, win->state.cols, priv->text);
  mutt_curses_set_color(MT_COLOR_NORMAL);

  mutt_debug(LL_DEBUG5, "repaint done\n");
  return 0;
}
```

flatcap: Good call.  Done.

My defensive style of programming means I've put in a huge number of checks.
Soon, I may be able to remove a lot of them.

For `recalc()` and `repaint()`, I'll guarantee `win` to be non-`NULL` as a contract of the function.

In the future, the Window functions won't get called if the Window isn't visible.

Finally, many of the observers watch `NeoMutt->notify` which leads to a LOT of notifications that aren't of interest.  By changing the observed object to a **single** object, more checks can be removed.

gui/msgwin.c
```diff
@@ -0,0 +1,233 @@
...
+  return 0;
+}
+
+/**
+ * msgwin_repaint - Redraw the Message Window
+ * @param win Message Window
+ */
+static int msgwin_repaint(struct MuttWindow *win)
+{
+  if (!mutt_window_is_visible(win) || !window_is_focused(win))
```
Austin-Ray, line 70: Realized the code I gave you was logically incorrect. Conditional should be:

```C
if (!mutt_window_is_visible(win) || window_is_focused(win))
  return 0;
```

Since we want to make sure the window **isn't** focused.

gui/mutt_window.c
```diff
@@ -769,6 +749,26 @@ void window_redraw(struct MuttWindow *win)
...
+
+  struct MuttWindow *win_focus = window_get_focus();
+  if (!win_focus)
+    return false;
+
+  if (win_focus == win)
+    return false;
+
+  return true;
+}
```
Austin-Ray, line 770: Similarly, here is also logically incorrect. It will return `false` if the window **is** focused. We can simplify the logic down to:

```C
return win_focus == win
```

gui/msgwin.c
flatcap: urgh!  yes :-)
fixed

gui/mutt_window.c
flatcap: done, thanks

Issue #2954, opened 2021-06-08
Add read_delay variable to control how long pager must remain on message before it gets marked read


ebblake:
## What would you like to be able to do?

I'm used to a UI feature present in several other mailers (both local such as Thunderbird, and web-based such as RoundCube) where a message is not marked as read merely because it was previewed, but only after a user-configured delay has passed without moving on to another message.  With such a feature, it becomes easier to open up pager mode with pager_index_lines set, and then jump to nearby-but-not-adjacent messages in a thread using repeated j/k (or other navigation bindings) based on the index preview, without marking all of the intermediate messages as read merely because they were between the current message and the target message (and possibly fewer keystrokes than typing in the numeric message id to skip to the desired message).

## What needs to change in NeoMutt?
The pager needs a new neomuttrc variable, I'm thinking 'read_delay' but am open to naming suggestions for consistency. It would default to 0 (existing behavior is that the pager visiting a message instantly marks it read), but if set to non-zero, it controls the number of seconds that must elapse while the pager is visiting a given message before that message is marked as read.  In addition to the code patch, this needs documentation updates.

IRC suggestions:
&lt;flatcap&gt; the MUTT_READ flag is set here: https://github.com/neomutt/neomutt/blob/master/pager/pager.c#L2427
&lt;flatcap&gt; moving that into the pager event loop and checking a variable, shouldn't be too hard: https://github.com/neomutt/neomutt/blob/master/pager/pager.c#L2514

## Have you tried any alternative methods?

- Macros
- Hooks
- Scripts
I'm not seeing any existing hooks that fire at the right time to make it as easy as a single config variable would be.  Maybe it could be cobbled together by adding a timeout-hook and figuring out a way to log the current message being visited by the pager on each timeout, such that a message is only marked as read if consecutive timeout-hook invocations are on the same message (such a workaround may be hairier in order to actually undo marking messages as read when they are first visited, only to remark them as read at the right timeout).

ebblake:
More from IRC:
&lt;gahr&gt; read_ is already a bit overloaded in read_inc and read_only, which are totally unrelated :)
&lt;gahr&gt; how about a new name for this feature? preview_delay? pager_mark_as_read_after?
&lt;gahr&gt; I wouldn't mind longer and more talking setting names
&lt;flatcap&gt; yeah, prefixes based on component
&lt;flatcap&gt; pager_read_delay, pager_email_read_delay

PR #2958, opened 2021-06-15
pager: add pager_read_delay

ebblake:
implements https://github.com/neomutt/neomutt/issues/2954
- config: add pager_read_delay config var
- docs: document it
- pager: delay marking new message as read

* **What does this PR do?**
* Adds a delay before the pager marks a message as read; if you navigate away prior to that delay, the message remains unread.  This is comparable to a feature present in at least Thunderbird, and allows for easier previewing of messages you want to focus on now vs. later.

     for that)

2954

### DIFF
--- a/pager/pager.c
+++ b/pager/pager.c
@@ -2303,7 +2303,38 @@ static bool jump_to_bottom(struct PagerRedrawData *rd, struct PagerView *pview)
 }
 
 /**
- * mutt_pager - Display a an email, attachment, or help, in a window
+ * pager_check_read_delay - Is it time to mark the message read?
+ * @param pview      Pager view
+ * @param mailbox    Mailbox of message being visited
+ * @param timestamp  Time when message should be marked read, or 0
+ * @param final      True if is the final call for the given message
+ * @retval 0 if delay has expired, or @timestamp if delay is not yet met
+ */
+static uint64_t pager_check_read_delay(struct PagerView *pview,
+                                       struct Mailbox *m, uint64_t timestamp,
+                                       bool final)
+{
+  if (timestamp == 0)
+    return 0;
+
+  const uint64_t now = mutt_date_epoch_ms();
+  if (now > timestamp)
+  {
+    if (final)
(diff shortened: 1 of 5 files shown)


pager/pager.c
```diff
@@ -2513,6 +2522,17 @@ int mutt_pager(struct PagerView *pview)
...
+    {
+      const uint64_t now = mutt_date_epoch_ms();
+      if (now - delay_begin > c_pager_read_delay * 1000)
+      {
+        mutt_debug(LL_DEBUG4, "delay elapsed while still on message, marking message read\n");
+        mutt_set_flag(m, pview->pdata->email, MUTT_READ, true);
+        delay_begin = 0;
+      }
+    }
+
```
gahr, line 2535: Do you actually need this block both when entering and when exiting the `while (op != -1)` loop?

gahr: `mutt_pager` is a huge function, would you mind factoring your block into a dedicated function?

gahr:
Thanks! A couple of comments are inline.

pager/pager.c
ebblake: > `mutt_pager` is a huge function, would you mind factoring your block into a dedicated function?

Sure, will do in v2.

ebblake: > Do you actually need this block both when entering and when exiting the `while (op != -1)` loop?

I think so. Not all actions navigate away from a message, so it is nicer to make the switch at the first possible action after enough time has elapsed if that will be observable by later actions while still in the pager.  But if you think otherwise, that a single check after the while() loop exits is sufficient, that's an easy enough change.  I'm also thinking towards the future: right now, the 'N' indicator in the index line (if you have pager_index_lines set) remains displayed.  Cooler would be having the 'N' indicator automatically clear itself at the first redraw after the message has been marked new, which requires clearing the unread bit inside the while loop.

ebblake:
Dropping this PR in favor of https://github.com/neomutt/neomutt/pull/2960/ which looks at a more reasonable branch, rebased to more recent activity

PR #2960, opened 2021-06-15
pager: add pager_read_delay

ebblake:
implements https://github.com/neomutt/neomutt/issues/2954
- config: add pager_read_delay config var
- docs: document it
- dlg_pager: pager_check_read_delay: new helper function
- dlg_pager: mutt_pager: use it to delay marking new message as read

* **What does this PR do?**
Supersedes https://github.com/neomutt/neomutt/pull/2958

     for that)

### DIFF
--- a/pager/dlg_pager.c
+++ b/pager/dlg_pager.c
@@ -2307,7 +2307,23 @@ static bool jump_to_bottom(struct PagerPrivateData *priv, struct PagerView *pvie
 }
 
 /**
- * mutt_pager - Display a an email, attachment, or help, in a window
+ * check_read_delay - Is it time to mark the message read?
+ * @param timestamp Time when message should be marked read, or 0
+ * @retval  true Message should be marked read
+ * @retval false No action necessary
+ */
+static bool check_read_delay(uint64_t *timestamp)
+{
+  if ((*timestamp != 0) && (mutt_date_epoch_ms() > *timestamp))
+  {
+    *timestamp = 0;
+    return true;
+  }
+  return false;
+}
+
+/**
+ * mutt_pager - Display an email, attachment, or help, in a window
  * @param pview Pager view settings
  * @retval  0 Success
  * @retval -1 Error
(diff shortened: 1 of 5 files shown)


flatcap (approved):
Very nice!  Thanks!

I've pushed a few minor fixes (97b883ac2).
I'll squash them before merging.

pager/dlg_pager.c
```diff
@@ -2306,7 +2306,48 @@ static bool jump_to_bottom(struct PagerPrivateData *priv, struct PagerView *pvie
...
- * mutt_pager - Display a an email, attachment, or help, in a window
+ * pager_check_read_delay - Is it time to mark the message read?
+ * @param pview      Pager view
+ * @param m          Mailbox of message being visited
+ * @param timestamp  Time when message should be marked read, or 0
+ * @param final      True if is the final call for the given message
+ * @retval 0 if delay has expired, or @a timestamp if delay is not yet met
+ */
+static uint64_t pager_check_read_delay(struct PagerView *pview, struct Mailbox *m,
+                                       uint64_t timestamp, bool final)
```
gahr, line 2317: Suggestion: 

```c
static uint64_t pager_check_read_delay(struct Mailbox *m, struct Email *e, uint64_t timestamp);
```

* nothing else from `struct PagerView` is needed
* let the caller do the logging based on `final`, if really needed, or just drop it

ebblake: If we drop logging altogether, I no longer need the long function name (that was a result of making the log useful); and if we separate checking timestamp from marking a message read, we reduce down to:
```
static bool check_read_delay(uint64_t timestamp)
{
  if (timestamp == 0)
    return false;
  return mutt_date_epoch_ms() > timestamp;
}

```
and the two call-sites become:
```
  // act 3
  while (op != -1)
  {
    if (check_read_delay(delay_read_timestamp))
      mutt_set_flag(m, pview->pdata->email, MUTT_READ, true);
    mutt_curses_set_cursor(MUTT_CURSOR_INVISIBLE);
    ...
  }
  // end act 3
  if (check_read_delay(delay_read_timestamp))
    mutt_set_flag(m, pview->pdata->email, MUTT_READ, true);
  mutt_file_fclose(&priv->fp);


```

gahr: Yep, except you'd be calling `mutt_set_flag` many times (once `mutt_date_epoch_ms() > timestamp` is true, it never becomes false again). That's why I was suggesting to pass `delay_read_timestamp` by pointer and let the function set it to 0 once done.

gahr:
```c
static bool check_read_delay(uint64_t *timestamp)
{
  if (*timestamp != 0 && mutt_date_epoch_ms() > *timestamp)
  {
    *timestamp = 0;
    return true;
  }
  return false
}

```

ebblake: Dropped logging and simplified the check_read_delay() function as discussed in the latest version of the commit

PR #2963, opened 2021-06-16
Fix too strict assertion when viewing an attachment

gahr:
We don't necessarily need to be sitting on an open mailbox (hence, have
a Context) when composing and email and viewing attachments: we might be
sending from the command line.

Reported by @guckes

### DIFF
--- a/pager/dlg_pager.c
+++ b/pager/dlg_pager.c
@@ -2359,8 +2359,6 @@ int mutt_pager(struct PagerView *pview)
       //  - fp and body->email in special case of viewing an attached email.
       assert(pview->pdata->email); // This should point to the top level email
       assert(pview->pdata->body);
-      assert(pview->pdata->ctx);
-      assert(pview->pdata->ctx->mailbox);
       if (pview->pdata->fp && pview->pdata->body->email)
       {
         // Special case: attachment is a full-blown email message.


pager/dlg_pager.c
```diff
@@ -2359,8 +2359,6 @@ int mutt_pager(struct PagerView *pview)
       //  - fp and body->email in special case of viewing an attached email.
       assert(pview->pdata->email); // This should point to the top level email
       assert(pview->pdata->body);
-      assert(pview->pdata->ctx);
-      assert(pview->pdata->ctx->mailbox);
```
Austin-Ray, line 2363: I'm not super familiar with command-line sending so please feel free to correct my assumptions.

Could this be represented as:
```C
// TUI assertions
if (pview->pdata->ctx)
{
    assert(pview->pdata->ctx->mailbox);
}
```
so the assertion isn't too strict for command-line sending, but ensures a consistent state for TUI usage?

gahr: I thought I checked and we would never need `pview->pdata->ctx` or `pview->pdata->ctx->mailbox` in this mode, but I will double check!

Austin-Ray: ah, the commit message didn't mention that. if that's the case then it makes sense to drop them.

flatcap (approved):
Looks good.
In mode `PAGER_MODE_ATTACH` there are a couple of places which use `pview->pdata->ctx` as a param to `el_add_tagged()`, but it checks for `NULL`.

Issue #2937, opened 2021-05-19
config file `set my_var+=val` doesn't extend custom vars


ebblake:
## Expected Behaviour

Split the assignment of a long line for a my_* variable in .neomuttrc, by writing two set lines for shorter portions to be concatenated into the final value.

The documentation in 'man neomuttrc' claims:

       set { variable=value | variable+=increment | variable-=decrement } [ ... ]

              NeoMutt  knows  four  basic types of variables: boolean, number,
              string, string list and quadoption. Boolean variables can be set
              (true),  unset  (false), or toggled. Number variables can be as‐
              signed a positive integer value. Value of number  variables  can
              be  incremented "+=" and decremented "-=". String list variables
              use "+=" for appending increment to the string list and "-=" for
              removal decrement from the string list.

              String  variables  consist of any number of printable characters
              and must be enclosed in quotes if they contain spaces  or  tabs.
              You  may also use the escape sequences “\n” and “\t” for newline
              and tab, respectively. Content of a string variables can be  ex‐
              tended using "+="

## Actual Behaviour

The docs are misleading; "four basic types" is incompatible with a list of 5 elements "boolean, number, string, string list, and quadoption".
And that's a grammar error, either you want "Content of string variables" or "Content of a string variable".
But those doc problems are secondary; I'm filing this because the docs say that string variables can be extended using +=, but in practice, I encountered overwrites rather than appending.

If NeoMutt crashed, did your OS create a 'coredump' file?

no crash

## Steps to Reproduce

```
$ cat tmp
set folder = /dev/null
set my_foo = "abc"
set my_foo += "def"
$ MAIL=/dev/null neomutt -n -Ftmp -z -e 'echo $my_foo\n' -d5
Debugging at level 5 to file '/home/eblake/.neomuttdebug0'
Mailbox is empty
$ cat ~/.neomuttdebug0
[2021-05-19 16:57:06] NeoMutt-20210205 debugging at level 5
[2021-05-19 16:57:06]<M> log_file_open() Debugging at level 5 to file '/home/eblake/.neomuttdebug0'

WARNING:
    Logging at this level can reveal personal information.
    Review the log carefully before posting in bug reports.

[2021-05-19 16:57:06]<1> log_translation() Translation: NONE
[2021-05-19 16:57:06]<2> mutt_window_reflow() entering
[2021-05-19 16:57:06]<2> source_rc() Reading configuration file '/home/eblake/.mutt/tmp'
[...]

PR #2966, opened 2021-06-17
myvar: Allow += modification of my_ variables.

ebblake:
Fixes: https://github.com/neomutt/neomutt/issues/2937

- myvar: add append parameter to myvar_set
- command_parse: use it in parse_set
- docs: tweak docs to match

* **What does this PR do?**
* Make it possible to use 'set my_var += "string value"'
* Error out on 'set my_var -= anything' instead of silently treating it like '='

2937

### DIFF
--- a/myvar.c
+++ b/myvar.c
@@ -121,6 +121,25 @@ void myvar_set(const char *var, const char *val)
 }
 
 /**
+ * myvar_append - Append to the value of a "my_" variable
+ * @param var Variable name
+ * @param val Value to append
+ */
+void myvar_append(const char *var, const char *val)
+{
+  struct MyVar *myv = myvar_find(var);
+
+  if (myv)
+  {
+    mutt_str_append_item(&myv->value, val, '\0');
+    return;
+  }
+
+  myv = myvar_new(var, val);
+  TAILQ_INSERT_TAIL(&MyVars, myv, entries);
+}
+
+/**
  * myvar_del - Unset the value of a "my_" variable
  * @param var Variable name
  */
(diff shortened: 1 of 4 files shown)


myvar.c
```diff
@@ -85,18 +85,26 @@ const char *myvar_get(const char *var)
...
 /**
  * myvar_set - Set the value of a "my_" variable
- * @param var Variable name
- * @param val Value to set
+ * @param var    Variable name
+ * @param val    Value to set
+ * @param append True to append to an existing value, if any
  */
-void myvar_set(const char *var, const char *val)
+void myvar_set(const char *var, const char *val, bool append)
```
gahr, line 92: I would avoid boolean parameters, and have

```C
void myvar_set(const char *var, const char *val);
void myvar_append(const char *var, const char *val);
```

perhaps both implemented in terms of a
```C
static struct MyVar *myvar_find(const char *var);
```

ebblake: Sure, will do in v2.

myvar.h
```diff
@@ -40,7 +40,7 @@ extern struct MyVarList MyVars; ///< List of all the user's custom config variab
 
 void        myvar_del(const char *var);
 const char *myvar_get(const char *var);
-void        myvar_set(const char *var, const char *val);
+void        myvar_set(const char *var, const char *val, bool append);
```
ebblake, line 43: clang-format wants to change the layout not only of my changed line, but also the neighboring lines.  Should I prefer prevailing style, or allow the changes?

flatcap: For the header files, don't worry.

We mandate `clang-format` for the source, but humans benefit from some extra whitespace in the headers.

flatcap (approved):

Issue #2969, opened 2021-06-17
use-after-free when quitting


ebblake:
## Expected Behaviour
Quitting neomutt should not trigger any ASAN use-after-free errors.

## Actual Behaviour
from ASAN
```
=================================================================
==402015==ERROR: AddressSanitizer: heap-use-after-free on address 0x60300024c640 at pc 0x7fadf37a1828 bp 0x7ffdd74a3d20 sp 0x7ffdd74a34d0
READ of size 1 at 0x60300024c640 thread T0
    #0 0x7fadf37a1827 in __interceptor_strcmp.part.0 (/lib64/libasan.so.6+0x8d827)
    #1 0x73d51f in mutt_str_cmp mutt/string.c:569
    #2 0x7252c6 in cmp_string_key mutt/hash.c:59
    #3 0x7263f2 in union_hash_delete mutt/hash.c:226
    #4 0x7270dd in mutt_hash_delete mutt/hash.c:425
    #5 0x6e4a26 in cs_uninherit_variable config/set.c:336
    #6 0x6e9ca8 in cs_subset_free config/subset.c:125
    #7 0x6d7912 in mailbox_free core/mailbox.c:103
    #8 0x6d725a in account_mailbox_remove core/account.c:113
    #9 0x6d749b in account_free core/account.c:138
    #10 0x6d9133 in neomutt_account_remove core/neomutt.c:125
    #11 0x6d892b in neomutt_free core/neomutt.c:71
    #12 0x487891 in main /home/eblake/neomutt/main.c:1306
    #13 0x7fadf2da2b74 in __libc_start_main (/lib64/libc.so.6+0x27b74)
    #14 0x40c26d in _start (/home/eblake/neomutt/neomutt+0x40c26d)

0x60300024c640 is located 0 bytes inside of 19-byte region [0x60300024c640,0x60300024c653)
freed by thread T0 here:
    #0 0x7fadf37c2647 in free (/lib64/libasan.so.6+0xae647)
    #1 0x72fcd5 in mutt_mem_free mutt/memory.c:75
    #2 0x6e375b in cs_hashelem_free config/set.c:70
    #3 0x72651f in union_hash_delete mutt/hash.c:230
    #4 0x7270dd in mutt_hash_delete mutt/hash.c:425
    #5 0x6e4a26 in cs_uninherit_variable config/set.c:336
    #6 0x6e9ca8 in cs_subset_free config/subset.c:125
[...]

PR #2974, opened 2021-06-18
hash: fix crash on element deletion

flatcap:
Fixes: #2969

When NeoMutt frees a ConfigSubset, it iterates through the set passing the name (key) of each item to mutt_hash_delete().

Unfortunately, passing the key name from `struct Inheritance` doesn't work.
The HashTable uses the name after deleting the item to check for duplicate items.

### DIFF
--- a/mutt/hash.c
+++ b/mutt/hash.c
@@ -421,8 +421,10 @@ void mutt_hash_delete(struct HashTable *table, const char *strkey, const void *d
   if (!table || !strkey)
     return;
   union HashKey key;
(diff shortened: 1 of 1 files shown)


gahr:
Shouldn't `union_hash_delete` handle this itself, by `strdup`ing `key.strkey` before the iteration?

flatcap:
> Shouldn't `union_hash_delete` handle this itself

Yes.  Done.

gahr (approved):

Issue #2973, opened 2021-06-17
failure to sync notmuch database when files moved between maildirs behind neomutt's back


ebblake:
## Expected Behaviour
If an email is moved by external programs to an alternative file while neomutt still has pending changes to that file's name (such as adding the S flag to change it from inbox/new/... unread to inbox/cur/...,S read), it should still be possible to query notmuch where the message id lives (and perform the rename on its new location instead), and failing that, assume that the user has deleted the email from notmuch's view and therefore syncing flag states on that message is no longer essential.  At any rate, failure to rename a message shouldn't lock me out of my other mail.

## Actual Behaviour
When the file moves behind neomutt's back, the notmuch.c code notices the easy attempt at maildir rename() failed, and should fall back to reopening the notmuch database to do its fallback. From notmuch/notmuch.c:nm_mbox_sync at commit 9a99804bbbcd3:
```
    bool ok = maildir_sync_mailbox_message(m, i, h);
    if (!ok)
    {
      // Syncing file failed, query notmuch for new filepath.
      notmuch_database_t *db = nm_db_get(m, true);
      if (db)
      {
        notmuch_message_t *msg = get_nm_message(db, e);

        sync_email_path_with_nm(e, msg);

        ok = maildir_sync_mailbox_message(m, i, h);
      }
      nm_db_release(m);
    }
```
but from my gdb session, the fallback fails:
```
Run till exit from #0  rename (
    old=0x61900000eb80 "/home/eblake/.Maildir/work/inbox/new/1623958045.413130_573.blue,U=38735:2,", 
    new=0x61900000dc80 "/home/eblake/.Maildir/work/inbox/cur/1623958045.413130_573.blue,U=38735:2,S") at ../sysdeps/unix/sysv/linux/rename.c:27
maildir_sync_message (m=0x612000002d40, msgno=2700) at maildir/maildir.c:453
453	    if (rename(mutt_buffer_string(oldpath), mutt_buffer_string(fullpath)) != 0)
Value returned is $11 = -1
(gdb) n
455	      mutt_perror("rename");
(gdb) 
456	      rc = -1;
(gdb) 
457	      goto cleanup;
(gdb) 
463	  mutt_buffer_pool_release(&newpath);
(gdb) 
464	  mutt_buffer_pool_release(&partpath);
(gdb) 
465	  mutt_buffer_pool_release(&fullpath);
(gdb) 
466	  mutt_buffer_pool_release(&oldpath);
(gdb) 
[...]

PR #2976, opened 2021-06-18
notmuch: really tolerate file renames behind neomutt's back

ebblake:
Resolves https://github.com/neomutt/neomutt/issues/2973

In commit 731a3d9c, we noted that aborting a mailbox sync on a
notmuch failure is not helpful (the user is now stuck in that mailbox,
and must quit neomutt to recover); it was deemed better to notify the
user of the failure, but to allow sync to claim success after all.
But the logic in that patch was flawed: the final value of rc depends
on the value of rc for the last message visited in the mailbox: if an
earlier message fails and a later one resets rc back to 0, we get the
desired behavior, but if the last message fails, rc is an error.

Later, commit c3afc987 cleaned up the error return code to make rc
sticky: now it gets set to a non-zero value on ANY failed message, not
just the last one in the notmuch mailbox being synced.

In the meantime, commit 84d9944f tried to get smarter about the
situation: just because the filename neomutt knows didn't work doesn't
mean notmuch doesn't know what to do - after all, notmuch's philosophy
is that REALLY deleting emails is an infrequent operation, more
likely, you just move them to a different folder or tag them with
something so they no longer appear in default queries, but the
underlying mail is still known to the database.  However, that code
never worked!  In order for mh_sync_mailbox_message() to succeed, we
first had to munge m->type to MUTT_MAILDIR, but nm_db_get() returns
NULL unless m->type is MUTT_NOTMUCH.  So if we're going to mess with
m->type, we have to mess with it in multiple places.

- nm_mbox_sync: try fallback to notmuch for real
- nm_mbox_sync: ignore sync failures for real

* **What does this PR do?**

     for that)

### DIFF
--- a/notmuch/notmuch.c
+++ b/notmuch/notmuch.c
@@ -2250,6 +2250,7 @@ static enum MxStatus nm_mbox_sync(struct Mailbox *m)
     if (!ok)
     {
       // Syncing file failed, query notmuch for new filepath.
+      m->type = MUTT_NOTMUCH;
       notmuch_database_t *db = nm_db_get(m, true);
       if (db)
       {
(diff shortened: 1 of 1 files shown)


Austin-Ray (approved):
LGTM

Issue #2984, opened 2021-06-21
stack-buffer-overflow with odd status_format


ebblake:
## Expected Behaviour
I was playing with unusual status_format values (because I'm considering adding a new expando to match my proposal to add a new $sort_thread config), and managed to crash things.  While the resulting status line may not make much sense, it must not crash.

## Actual Behaviour
In an ASAN build, `:set status_format = '%?l? %? %& & %& %?? %s'` crashed neomutt with:

If NeoMutt crashed, did your OS create a 'coredump' file?
ASAN log:
```
=================================================================
==754242==ERROR: AddressSanitizer: stack-buffer-overflow on address 0x7ffd4931bc80 at pc 0x00000048d143 bp 0x7ffd4931b980 sp 0x7ffd4931b978
WRITE of size 1 at 0x7ffd4931bc80 thread T0
    #0 0x48d142 in mutt_expando_format /home/eblake/neomutt/muttlib.c:989
    #1 0x4dd335 in menu_status_line /home/eblake/neomutt/status.c:435
    #2 0x507b95 in ibar_recalc index/ibar.c:64
    #3 0x595022 in window_recalc gui/mutt_window.c:596
    #4 0x5950b0 in window_recalc gui/mutt_window.c:602
    #5 0x5950b0 in window_recalc gui/mutt_window.c:602
    #6 0x5950b0 in window_recalc gui/mutt_window.c:602
    #7 0x5950b0 in window_recalc gui/mutt_window.c:602
    #8 0x5950b0 in window_recalc gui/mutt_window.c:602
    #9 0x59530e in window_redraw gui/mutt_window.c:640
    #10 0x59d1bc in menu_redraw_full menu/draw.c:342
    #11 0x4e98c8 in index_custom_redraw index/dlg_index.c:1100
    #12 0x4ebf57 in mutt_index_menu index/dlg_index.c:1336
    #13 0x487540 in main /home/eblake/neomutt/main.c:1261
    #14 0x7fda1ce6bb74  (/lib64/libc.so.6+0x27b74)
    #15 0x40c26d in _start (/home/eblake/neomutt/neomutt+0x40c26d)

Address 0x7ffd4931bc80 is located in stack of thread T0 at offset 544 in frame
    #0 0x48b873 in mutt_expando_format /home/eblake/neomutt/muttlib.c:782

  This frame has 16 object(s):
    [48, 52) 'pw' (line 1107)
    [64, 68) 'pw' (line 1183)
[...]

PR #2985, opened 2021-06-21
status: fix buffer overflow

flatcap:
Fixes: #2984 

Off-by-one caused a buffer overflow when formatting `$status_format`.

### DIFF
--- a/muttlib.c
+++ b/muttlib.c
@@ -981,7 +981,7 @@ void mutt_expando_format(char *buf, size_t buflen, size_t col, int cols, const c
         src++;
         cp = prefix;
         count = 0;
-        while ((count < sizeof(prefix)) && (*src != '?'))
+        while ((count < (sizeof(prefix) - 1)) && (*src != '\0') && (*src != '?'))
         {
           *cp++ = *src++;
           count++;
@@ -995,8 +995,7 @@ void mutt_expando_format(char *buf, size_t buflen, size_t col, int cols, const c
         /* eat the format string */
         cp = prefix;
         count = 0;
-        while ((count < sizeof(prefix)) && (isdigit((unsigned char) *src) || (*src == '.') ||
-                                            (*src == '-') || (*src == '=')))
+        while ((count < (sizeof(prefix) - 1)) && strchr("0123456789.-=", *src))
         {
           *cp++ = *src++;
           count++;


ebblake:
With this applied, the crash is gone, but the code is still hairy.  With the mentioned `set status_format = '%?l? %? %& & %& %?? %s'`, the first mutt_expando_format() results in a call to
```
 status_format_str (buf=0x7fff71826520 "", buflen=1024, col=0, cols=110, 
    op=108 'l', src=0x7fff718263ec " %& %?? %s", prec=0x7fff71826200 "", 
    if_str=0x7fff718262a0 " %> %", else_str=0x7fff71826340 " ", 
```
It looks like mutt rewrote the first %? into %&gt;  when building the if_str (my attempts to escape the ? were futile, it was rewritten as if it were the terminator of the overall %?l?ifstr? expression), used the & of %& as the swap to else_str, then turned the bare &amp; into the termination of else_str.  Then from the callback, a second call to mutt_expando_format() is passed `" %> %"` as its src, which is not a well-formed format (it would be right-justify and pad with space, followed by an unterminated %), and that in turn triggers a call to mutt_expando_format() on `"%"`, which finally detects the invalid format.

I have no idea if it is good that we rewrite ? into ? before forming if_str, but that appears to be pre-existing and independent of your off-by-one fix.

However, I'm also concerned that your patch is NOT fixing the REAL problem: you are iterating past the end of src.
```
984	        while ((count < (sizeof(prefix) - 1)) && (*src != '?'))
(gdb) 
986	          *cp++ = *src++;
(gdb) p src
$21 = 0x7fff718263fc ""
(gdb) n
987	          count++;
(gdb) 
984	        while ((count < (sizeof(prefix) - 1)) && (*src != '?'))
(gdb) 
986	          *cp++ = *src++;
(gdb) 
987	          count++;
(gdb) p src
$22 = 0x7fff718263fe ""
```
Using sizeof(prefix)-1 (the size of cp) means we don't write past the end of cp, but does nothing about the fact that we are reading garbage beyond the end of src.

muttlib.c
```diff
@@ -995,8 +995,9 @@ void mutt_expando_format(char *buf, size_t buflen, size_t col, int cols, const c
         /* eat the format string */
         cp = prefix;
         count = 0;
-        while ((count < (sizeof(prefix) - 1)) && (isdigit((unsigned char) *src) || (*src == '.') ||
-                                            (*src == '-') || (*src == '=')))
+        while ((count < (sizeof(prefix) - 1)) && (*src != '\0') &&
+               (isdigit((unsigned char) *src) || (*src == '.') ||
+                (*src == '-') || (*src == '=')))
```
ebblake, line 1000: Technically, the check for (*src != '\0') is redundant here, because you are guaranteed that is the case if (isdigit || '.' || '-' || '=') succeeds.  But for symmetry with line 984, where the check is NOT redundant, I like keeping it in both places.
LGTM

ebblake (approved):

muttlib.c
```diff
@@ -995,8 +995,9 @@ void mutt_expando_format(char *buf, size_t buflen, size_t col, int cols, const c
         /* eat the format string */
         cp = prefix;
         count = 0;
-        while ((count < sizeof(prefix)) && (isdigit((unsigned char) *src) || (*src == '.') ||
-                                            (*src == '-') || (*src == '=')))
+        while ((count < (sizeof(prefix) - 1)) && (*src != '\0') &&
+               (isdigit((unsigned char) *src) || (*src == '.') ||
+                (*src == '-') || (*src == '=')))
```
gahr, line 1000: I would avoid any extra checks and try to simplify the logic as much as possible.

```c
        while ((count < (sizeof(prefix) - 1)) && (isdigit(*src) || strchr(".-=", *src)))
```

or even

```c
        while ((count < (sizeof(prefix) - 1)) && strchr("0123456789.-=", *src))
```

flatcap: Nice!  Done.

PR #2986, opened 2021-06-21
menu: fix collapse-all

flatcap:
This may solve the problem of the cursor position in the index after functions such as `<collapse-all>`.

Calling `<collapse-all>` for a second time, opens up all the folded threads.
NeoMutt then calculates the new position for the cursor.
However, the new menu-cursor setting code range-checks the value and the menu hasn't been resized to match.

This failure led to the erratic positioning of the cursor.

### DIFF
--- a/index/dlg_index.c
+++ b/index/dlg_index.c
@@ -268,6 +268,7 @@ static void collapse_all(struct Context *ctx, struct Menu *menu, int toggle)
 
   /* Restore the cursor */
   mutt_set_vnum(ctx->mailbox);
+  menu->max = ctx->mailbox->vcount;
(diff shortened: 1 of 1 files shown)


ebblake (commented):
This appears to have fixed `<uncollapse-all>`, but did not fix `<next-new-then-unread>` jumping to the root message instead of the (previously-hidden) new message inside a collapsed thread where the root is read but a submessage is not.  `<entire-thread>` is also not landing on the right message, if the additional messages loaded in are parents to the message I was on prior to loading the extra messages.

ebblake:
okay, elsewhere we determined that you've fixed the jump target of `<collapse-all>`, but still need further work on interaction with `$uncollapse_jump`.  We'll get there

Issue #2988, opened 2021-06-23
regression with `set sort_aux=...` and similar taking immediate effect


ebblake:
## Expected Behaviour
Any time the index is shown and the user can alter config commands (whether by the `o` (`<sort-mailbox>`) command, or by `set ...`), various configure settings that can result in alternate ordering of the index should trigger immediate re-sorting.

## Actual Behaviour
While the `o` command still works, we accidentally lost the ability for `$duplicate_threads`, `$sort_re`, `$sort_aux`, `$strict_threads`, `$thread_received` to update the index on the fly.

## Steps to Reproduce
Asciinema recording attached demonstrating the difference between 20210205 and latest git.  In the recording, both versions visit the 31-message mbox of the March 2021 neomutt-devel archives, execute `ot` to sort by threads, then `:set sort_aux=last-date` which SHOULD rearrange some of the subthreads: pre-`set`, the thread at message 23-26 appears in date order where siblings are sorted solely by their own date and not descendants; post-`set`, that thread has a more recent reply and sinks lower in the overall setting (28-31), as well as having its two subthreads rearranged for the same reason.  (Ignore the memory leaks reported by asan at the end; that is a known issue covered elsewhere)
[sort_bug.asciinema.txt](https://github.com/neomutt/neomutt/files/6703814/sort_bug.asciinema.txt)

## How often does this happen?
- Always

## When did it start to happen?
Commit faf75108 introduced the bug when it changed the menu config observer.

## NeoMutt Version

```
NeoMutt 20210205-532-5c2f1b
Copyright (C) 1996-2020 Michael R. Elkins and others.
NeoMutt comes with ABSOLUTELY NO WARRANTY; for details type 'neomutt -vv'.
NeoMutt is free software, and you are welcome to redistribute it
under certain conditions; type 'neomutt -vv' for details.

System: Linux 5.12.10-300.fc34.x86_64 (x86_64)
ncurses: ncurses 6.2.20200222 (compiled with 6.2.20200222)
libidn2: 2.3.1 (compiled with 2.3.1)
GPGME: 1.15.1
GnuTLS: 3.7.2
libnotmuch: 5.3.0
PCRE2: 10.36 2020-12-04
storage: tokyocabinet, lmdb
compression: lz4, zlib, zstd

[...]

PR #2989, opened 2021-06-23
Devel/sort fix

ebblake:
* **What does this PR do?**
Fixes easier issues I found along the way towards my eventual goal of resolving #2342. In particular, this fixes a regression documented in #2988.

     for that)

#2988

### DIFF
--- a/mutt_thread.c
+++ b/mutt_thread.c
@@ -43,8 +43,6 @@
 #include "protos.h"
 #include "sort.h"
 
-static sort_t sort_func = NULL;
-
 /**
  * struct ThreadsContext - The "current" threading state
  */
@@ -53,6 +51,7 @@ struct ThreadsContext
   struct Mailbox *mailbox; ///< Current mailbox
   struct MuttThread *tree; ///< Top of thread tree
   struct HashTable *hash;  ///< Hash table for threads
+  short c_sort;            ///< Sort method
 };
 
 /**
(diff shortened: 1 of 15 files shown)


sort.c
```diff
@@ -436,6 +439,8 @@ void mutt_sort_headers(struct Mailbox *m, struct ThreadsContext *threads,
   }
   else
   {
+    OptSortReverse = c_sort_aux & SORT_REVERSE;
```
ebblake, line 442: Copy+paste error, that should be c_sort, not c_sort_aux

commands.c
```diff
@@ -830,18 +830,18 @@ void mutt_print_message(struct Mailbox *m, struct EmailList *el)
...
   enum SortType sort = SORT_DATE;
 
   switch (mutt_multi_choice(reverse ?
-                                /* L10N: The highlighted letters must match the "Sort" options */
+                                /* L10N: The highlighted letters must match the "Rev-Sort" options */
                                 _("Rev-Sort "
                                   "(d)ate,(f)rm,(r)ecv,(s)ubj,t(o),(t)hread,(u)"
                                   "nsort,si(z)e,s(c)ore,s(p)am,(l)abel?") :
-                                /* L10N: The highlighted letters must match the "Rev-Sort" options */
+                                /* L10N: The highlighted letters must match the "Sort" options */
```
flatcap, line 844: Actually, these were correct.
What they're saying is:
The highlighted letters in the string **below**, should match those in the string **mentioned** in the comment.

`L10N` comments are copied to the `.po` files where the order of the strings may not match the code.

mutt_thread.c
```diff
@@ -679,29 +680,28 @@ static int compare_threads(const void *a, const void *b)
...
-  int i, array_size, sort_top = 0;
+  int i, array_size;
+  bool sort_top = false;
 
   /* we put things into the array backwards to save some cycles,
    * but we want to have to move less stuff around if we're
    * resorting, so we sort backwards and then put them back
    * in reverse order so they're forwards */
-  short c_sort = cs_subset_sort(NeoMutt->sub, "sort");
+  short c_sort = cs_subset_sort(NeoMutt->sub, "sort_aux");
```
flatcap, line 698: Please can you change `c_sort` to `c_sort_aux`

flatcap (changes requested):

commands.c
ebblake: okay, that makes more sense. I've removed this hunk of the change in my latest forced push.

mutt_thread.c
ebblake: updated

flatcap (commented):
First three commits merged.  Thanks!

ebblake:
Hmm. Configure on FreeBSD failed to find qsort_s, then used qsort_r with the wrong signature. We'll have to improve auto.def probing

flatcap:
I tweaked the `.cirrus.yml` to get a bit more info: https://cirrus-ci.com/task/5053252086530048
(I don't have access to FreeBSD box / vm)

Snippet from `config.log`
```
Failed: cc -g -O2 -std=c99 -D_ALL_SOURCE=1 -D_GNU_SOURCE=1 -D__EXTENSIONS__ -std=c99 -g -O2 -std=c99 -D_ALL_SOURCE=1 -D_GNU_SOURCE=1 -D__EXTENSIONS__ conftest__.c -o conftest__
ld: error: undefined symbol: qsort_s
>>> referenced by conftest__.c:3
>>>               /tmp/conftest__-acba91.o:(main)
>>> did you mean: qsort_b
>>> defined in: /lib/libc.so.7
cc: error: linker command failed with exit code 1 (use -v to see invocation)
============
The failed code was:
extern void qsort_s(void);
int main(void) {
qsort_s();
return 0;
}
```

Snippet from `config.h`
```
#define HAVE_QSORT_R 1
/* #undef HAVE_QSORT_S */
```

ebblake:
Looks like FreeBSD wants us to do this first:

  #define __STDC_WANT_LIB_EXT1__ 1

flatcap (approved):
Very nice.  Thanks!

Issue #2996, opened 2021-06-29
`<attach-message>` with notmuch dumps core


ebblake:
## Expected Behaviour
open a nested index view to tag a message that should be attached as part of a reply to another message

## Actual Behaviour
I was trying to reply to a mail in a notmuch folder, and attach another message from a different notmuch query as part of my composition.
neomutt: mutt_thread.c:1099: mutt_sort_threads: Assertion `tctx->tree' failed.

If NeoMutt crashed, did your OS create a 'coredump' file?
Running under gdb produced:
```
Program received signal SIGABRT, Aborted.
0x00007f361d5d02a2 in ?? () from /lib64/libc.so.6
(gdb) bt
#0  0x00007f361d5d02a2 in  () at /lib64/libc.so.6
#1  0x00007f361d5b98a4 in  () at /lib64/libc.so.6
#2  0x00007f361d5b9789 in  () at /lib64/libc.so.6
#3  0x00007f361d5c8a16 in  () at /lib64/libc.so.6
#4  0x00000000004ab04e in mutt_sort_threads (tctx=0x603000195310, init=true)
    at mutt_thread.c:1099
#5  0x00000000004da7cb in mutt_sort_headers
    (m=0x612000004e40, threads=0x603000195310, init=true, vsize=0x60700047c9c0)
    at sort.c:403
#6  0x00000000004353a3 in ctx_update (ctx=0x60700047c9c0) at context.c:214
#7  0x000000000043440f in ctx_new (m=0x612000004e40) at context.c:96
#8  0x00000000004e98d0 in mutt_index_menu
    (dlg=0x60e000076940, m_init=0x612000004e40) at index/dlg_index.c:1134
#9  0x00000000006c9942 in mutt_compose_menu
    (e=0x60f00020c410, fcc=0x7ffebcd954d0, flags=0 '\000', sub=0x604000000a50)
    at compose/compose.c:1345
#10 0x00000000006ab7d5 in mutt_send_message
    (flags=1, e_templ=0x60f00020c410, tempfile=0x0, m=0x612000004e40, el=0x7ffebcd95d10, sub=0x604000000a50) at send/send.c:2726
#11 0x0000000000524895 in mutt_pager (pview=0x7ffebcd968f0)
    at pager/dlg_pager.c:3758
#12 0x00000000004207c5 in mutt_display_message
[...]

PR #2997, opened 2021-06-29
sort: avoid assertion under <attach-message>

ebblake:
Partially Fixes: https://github.com/neomutt/neomutt/issues/2996

Commit b698a6ab added an assert that tctx->tree was set and dropped an
'if', because I could not at the time find any scenario where we would
try sorting threads but not find any.  But in the meantime, I found
(at least) one: when using <attach-message> and selecting the same
folder as already being visited, and when no new mail has arrived in
the meantime, then ALL messages already have e->thread set, and there
is no need to sort subthreads.  But the comment behind the assert was
correct: in that code path, we forgot to restore the value of $sort
after temporarily munging it to the value of $sort_aux.

Later, commit aaa2ece04 changed things so that we no longer munge
$sort in the first place.  So now all we need to do is remove the
bogus assertion, while still restoring the 'if' to avoid unnecessary
work if there are no new threads that need subthread sorting.

Note that the issue of potentially leaving $sort munged to $sort_aux
if ctx->tree remains unset after searching for new threads is present
even in upstream mutt.  But upstream never saw the problem with
<attach-message> because it always opened a NEW mailbox context,
passing init=true.  It is only in neomutt (and possibly only with
notmuch folders, since I didn't manage to reproduce with maildir),
where we try harder to reuse an existing mailbox and pass init=false
while tctx->tree is still NULL.

* **What does this PR do?**

     for that)

### DIFF
--- a/mutt_thread.c
+++ b/mutt_thread.c
@@ -1096,15 +1096,17 @@ void mutt_sort_threads(struct ThreadsContext *tctx, bool init)
 
   /* if $sort_aux or similar changed after the mailbox is sorted, then
    * all the subthreads need to be resorted */
-  assert(tctx->tree);
-  mutt_sort_subthreads(tctx, init || OptSortSubthreads);
-  OptSortSubthreads = false;
+  if (tctx->tree)
+  {
+    mutt_sort_subthreads(tctx, init || OptSortSubthreads);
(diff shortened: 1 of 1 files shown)


Austin-Ray (approved):
LGTM

Issue #2972, opened 2021-06-17
allow windowed notmuch queries to include additional messages, jump back to present


ebblake:
## What would you like to be able to do?
I love the idea of notmuch windowed query, and use (among others):
```
set nm_query_type = threads
set nm_query_window_duration = 6
set nm_query_window_timebase = month
bind index < vfolder-window-backward
bind index > vfolder-window-forward
```
However, there are some messages that I intentionally flag or leave unread to remind myself that _someday_ I need to revisit it, even if they have aged longer than my default window (this works best on a folder that tries to mark all messages read once they are no longer important to revisit, but where preserving the message for archival reasons is a good idea).

Right now, the notmuch windowed behavior is basically:
'...existing query...' AND date:start..end
but I would rather have:
'...existing query...' AND (date:start..end OR (tag:flagged or tag:unread))

Of course, the decision on _which_ messages should be shown always in spite of windowing is itself something that needs to be customized, and something that I should be able to change on the fly.

## What needs to change in NeoMutt?
I'm envisioning a new config variable, maybe named nm_query_window_augment, which if specified, is an additional query to OR alongside the windowing date limiter, so that my windowed view is all messages meeting the date window or my augmented query.  Then it becomes easy to:
set nm_query_window_augment = 'tag:flagged or tag:unread'

Also, I've noticed that while browsing backwards in time to find that old thread I remember, repeated vfolder-window-backward is nice; but when I'm done searching for that old message, I usually want to return to the present; it would be nice to have a vfolder-window-reset command that immediately jumped to the present, instead of having to repeat as many vfolder-window-forward commands as I previously went backwards.

## Have you tried any alternative methods?

- Macros
- Hooks
- Scripts
Instead of using neomutt's built-in date windowing, I can always write my mailbox notmuch:// queries to do manual windowing myself, but that becomes a lot more invasive for rewriting a folder query every time I want to adjust the window being visited.

There may be ways to write a macro that temporarily changes nm_query_window_duration and/or nm_query_window_timebase to more easily jump to the present by one large forward jump (say 30 years) without having to worry about how many smaller backwards I went through, although I haven't played with it.  That is because the whole point of windowing is to view only a few hundred recent emails, rather than the unwindowed entire history of thousands of emails, where there is a performance drag any time notmuch presents the entire history of tagged messages without a date limiter.

PR #2998, opened 2021-06-30
notmuch: implement query window reset functionality

Austin-Ray:
This PR adds a user function to reset the current query window position to the present. It's a quality of life improvement for notmuch users as there's no documented way to reset the position. While it's possible to use the undocumented config. variable `nm_query_window_current_position`, it's used for internal state keeping so not something a user should access. It's much better to give the user a function.

In addition to the feature, there's a refactor merging all the query window `index` code into a single code block because the only difference between them is the notmuch backend call. There's no need for duplication at this time.

---

@ebblake can you test this functionality to see if it matches your request in #2972

### DIFF
--- a/index/dlg_index.c
+++ b/index/dlg_index.c
@@ -2429,6 +2429,33 @@ struct Mailbox *mutt_index_menu(struct MuttWindow *dlg, struct Mailbox *m_init)
         change_folder_notmuch(priv->menu, buf, sizeof(buf), &priv->oldcount, shared, false);
         break;
       }
+
+      case OP_MAIN_WINDOWED_VFOLDER_RESET:
+      {
+        if (!prereq(shared->ctx, priv->menu, CHECK_IN_MAILBOX))
+          break;
+        const short c_nm_query_window_duration =
+            cs_subset_number(shared->sub, "nm_query_window_duration");
+        const bool c_nm_query_window_enable =
+            cs_subset_bool(shared->sub, "nm_query_window_enable");
+        if (!c_nm_query_window_enable && (c_nm_query_window_duration <= 0))
+        {
+          mutt_message(_("Windowed queries disabled"));
+          break;
+        }
+        const char *const c_nm_query_window_current_search =
+            cs_subset_string(shared->sub, "nm_query_window_current_search");
+        if (!c_nm_query_window_current_search)
+        {
+          mutt_message(_("No notmuch vfolder currently loaded"));
+          break;
+        }
+        nm_query_window_reset();
+        char buf[PATH_MAX] = { 0 };
+        mutt_str_copy(buf, c_nm_query_window_current_search, sizeof(buf));
+        change_folder_notmuch(priv->menu, buf, sizeof(buf), &priv->oldcount, shared, false);
+        break;
+      }
 #endif
 
 #ifdef USE_SIDEBAR
--- a/index/dlg_index.c
+++ b/index/dlg_index.c
@@ -2377,53 +2377,10 @@ struct Mailbox *mutt_index_menu(struct MuttWindow *dlg, struct Mailbox *m_init)
       }
 
       case OP_MAIN_WINDOWED_VFOLDER_BACKWARD:
-      {
-        if (!prereq(shared->ctx, priv->menu, CHECK_IN_MAILBOX))
-          break;
-        if (!nm_query_window_available())
-        {
-          mutt_message(_("Windowed queries disabled"));
-          break;
-        }
-        const char *const c_nm_query_window_current_search =
-            cs_subset_string(shared->sub, "nm_query_window_current_search");
-        if (!c_nm_query_window_current_search)
-        {
-          mutt_message(_("No notmuch vfolder currently loaded"));
-          break;
-        }
-        nm_query_window_backward();
-        char buf[PATH_MAX] = { 0 };
-        mutt_str_copy(buf, c_nm_query_window_current_search, sizeof(buf));
-        change_folder_notmuch(priv->menu, buf, sizeof(buf), &priv->oldcount, shared, false);
-        break;
-      }
-
       case OP_MAIN_WINDOWED_VFOLDER_FORWARD:
-      {
-        if (!prereq(shared->ctx, priv->menu, CHECK_IN_MAILBOX))
-          break;
-        if (!nm_query_window_available())
-        {
-          mutt_message(_("Windowed queries disabled"));
-          break;
-        }
-        const char *const c_nm_query_window_current_search =
-            cs_subset_string(shared->sub, "nm_query_window_current_search");
-        if (!c_nm_query_window_current_search)
-        {
-          mutt_message(_("No notmuch vfolder currently loaded"));
-          break;
-        }
-        nm_query_window_forward();
-        char buf[PATH_MAX] = { 0 };
-        mutt_str_copy(buf, c_nm_query_window_current_search, sizeof(buf));
-        change_folder_notmuch(priv->menu, buf, sizeof(buf), &priv->oldcount, shared, false);
-        break;
-      }
-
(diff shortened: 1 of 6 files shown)


index/dlg_index.c
```diff
@@ -2380,11 +2380,7 @@ struct Mailbox *mutt_index_menu(struct MuttWindow *dlg, struct Mailbox *m_init)
...
           break;
-        const short c_nm_query_window_duration =
-            cs_subset_number(shared->sub, "nm_query_window_duration");
-        const bool c_nm_query_window_enable =
-            cs_subset_bool(shared->sub, "nm_query_window_enable");
-        if (!c_nm_query_window_enable && (c_nm_query_window_duration <= 0))
+        if (!nm_query_window_available())
         {
           mutt_message(_("Windowed queries disabled"));
           break;
```
ebblake: Is it worth putting the mutt_message() call inside nm_query_window_available(), rather than repeating it at each call-site?

notmuch/notmuch.c
```diff
@@ -1607,6 +1607,19 @@ char *nm_url_from_query(struct Mailbox *m, char *buf, size_t buflen)
...
+ *  nm_query_window_available - Are windowed queries enabled for use?
+ */
+bool nm_query_window_available(void)
+{
+  const short c_nm_query_window_duration =
+      cs_subset_number(NeoMutt->sub, "nm_query_window_duration");
+  const bool c_nm_query_window_enable =
+      cs_subset_bool(NeoMutt->sub, "nm_query_window_enable");
+
+  return c_nm_query_window_enable || (c_nm_query_window_duration > 0);
```
ebblake, line 1620: This is the opposite logic of the code it is replacing, but looks like you are fixing the pre-existing bug I pointed out on the previous commit.  Maybe call out the bug fix in the commit message?

ebblake (changes requested):
typo in the commit message for 1/3 (s/additonal/additional)
The new function does what I expect. Missing docs patch to call out the availability of the function.

index/dlg_index.c
```diff
@@ -2429,6 +2429,33 @@ struct Mailbox *mutt_index_menu(struct MuttWindow *dlg, struct Mailbox *m_init)
...
+
+      case OP_MAIN_WINDOWED_VFOLDER_RESET:
+      {
+        if (!prereq(shared->ctx, priv->menu, CHECK_IN_MAILBOX))
+          break;
+        const short c_nm_query_window_duration =
+            cs_subset_number(shared->sub, "nm_query_window_duration");
+        const bool c_nm_query_window_enable =
+            cs_subset_bool(shared->sub, "nm_query_window_enable");
+        if (!c_nm_query_window_enable && (c_nm_query_window_duration <= 0))
```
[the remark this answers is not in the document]
Austin-Ray: > nm_query_window_duration is marked DT_NOT_NEGATIVE, so the <=0 could be ==0

That conditional pre-dates the introduction configuration validation framework. It used to be possible to set a negative value without NeoMutt's complaint. It's better to leave the working code alone than make modifications as the initial implementation assumptions may differ from our current understanding.

> Should that be || instead of &&

No, that would disable query windows whenever `nm_query_window_enable=false` or `nm_query_window_duration=0`. 

This is not desired behavior as `nm_query_window_enable` is a new config. variable to allow query windows when `nm_query_window_duration=0`. This allows windows over a single timebase (e.g. `date:1month..1month`). Before that the explicit enable `nm_query_window_duration=0` disabled the query windows.

It was implemented this way so we wouldn't break existing user setups while supporting new functionality. It's a better user experience.

index/dlg_index.c
Austin-Ray: > repeating it at each call-site?

This duplication is removed in a later commit in this PR, which separates the new feature from a refactor so it's easier to identify root causes of breakage during a bisect. Also allows reverting the refactor while keeping the feature intact.

notmuch/notmuch.c
Austin-Ray: > This is the opposite logic of the code it is replacing

This logic is negated as this function checks if query windows **are** available while the original conditional checks if query windows **are not** available. The call sites use `!nm_query_window_available` to preserve logical equivalence. It's very common to separate out negations into the language keywords instead of function names. Prevents code that clumsy looking like "if not not_available()"

It also makes the logic easier to follow, "if explicitly enabled then available otherwise check the legacy enable condition (duration > 0)"

> you are fixing the pre-existing bug

There is no bug. If a user explicitly enables query windows then we won't check the duration for enable/disable condition.

Austin-Ray:
> typo in the commit message for 1/3 (s/additonal/additional)

Fixed. My grammer/spelling linting setup broke for some reason. Thanks for catching that.

> The new function does what I expect. Missing docs patch to call out the availability of the function.

I'm not sure what you mean by this. In the past, I've never been asked to modify any doc file other than `docs/manual.xml.head`. What am I missing?

index/dlg_index.c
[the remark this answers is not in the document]
ebblake: okay, I missed DeMorgan's rules at a distance. The new code is !(call()), where call does 'enable || duration &gt; 0)', but moving the ! inside the call is indeed (!enable) && (!(duration &gt; 0)).

ebblake:
> > The new function does what I expect. Missing docs patch to call out the availability of the function.
> 
> I'm not sure what you mean by this. In the past, I've never been asked to modify any doc file other than `docs/manual.xml.head`. What am I missing?

Prior to your patch:
```
$ git grep vfolder-window docs
docs/manual.xml.head:                <entry><literal>&lt;vfolder-window-forward&gt;</literal></entry>
docs/manual.xml.head:                <entry><literal>&lt;vfolder-window-backward&gt;</literal></entry>
docs/manual.xml.head:bind index &lt; vfolder-window-backward
docs/manual.xml.head:bind index &gt; vfolder-window-forward
```
You touched one place, but not the other; although figuring out what key to recommend to users who want to bind it in their own rc file may be harder, so just listing it as a comment might do.
In general, we have very little information about the vfolder-window functionality - just the feature page description of it. However, I'm okay with adding yet another function to this family of underdocumented set without mandating that you also give more examples of the overall usage of windowing (although that won't hurt if you add it, it is more work than your original patch).

Austin-Ray:
> You touched one place, but not the other; although figuring out what key to recommend to users who want to bind it in their own rc file may be harder,

That second place is an example RC file in the notmuch section, and I intentionally left out the binding because:
 1) Bindings are a fundamental concept of NeoMutt covered elsewhere so I think the example RC should demonstrate non-trivial config. specific to the notmuch backend.
 2) The comment next to the binding would repeat the documentation in the _Functions_ section
 3) The example RC seems like more of a demonstration and not a source of truth for documentation.
 4) I couldn't think of a good binding (`0` could work)
 
What are some arguments for having it in example RC file?

> In general, we have very little information about the vfolder-window functionality - just the feature page description of it.

I'm happy to add documentation; however, I personally don't use the feature and understand its usefulness. As a user of query windows, would you be willing to write the documentation explaining its function, useful qualities, and why a user might consider it?

---
> this family of underdocumented

As an aside, if you think that's bad, the associated variables weren't documented in `manual.xml.head` outside of that example RC file until 23461dfa0. They were introduced in 426f0c27b

ebblake:
> > In general, we have very little information about the vfolder-window functionality - just the feature page description of it.
> 
> I'm happy to add documentation; however, I personally don't use the feature and understand its usefulness. As a user of query windows, would you be willing to write the documentation explaining its function, useful qualities, and why a user might consider it?

Yes, that seems like a reasonable request. At any rate, I don't see a documentation patch as a reason to hold up this PR, so I'm marking my review as approved - you've addressed all my immediate comments.

ebblake (approved):

PR #3002, opened 2021-07-08
notmuch: show additional mail in query windows

### LINKED ISSUE #2972: allow windowed notmuch queries to include additional messages, jump back to present

Austin-Ray:
Sometimes a user may wish to include certain messages in a query window
even if it exists outside the date range. For example, always showing
flagged messages.

To support this, extend the query window code to emit a window
containing an `OR` statement with additional search terms if a user
configures the `nm_query_window_additional_terms` variable.

For example, using the mailbox `notmuch://?query=tag:inbox` as the
mailbox and `tag:flagged and tag:unread` as the additional search terms
will produce a query window such as:

`notmuch://?query=tag:inbox and (date:... or (tag:flagged and tag:unread))`

The `date` value was omitted for brevity. The resulting mailbox contains
mail with tag inbox within the date range and mail with tags inbox,
flagged, and unread.

Closes #2972

---

Some notes to focus on in the review:

 1. How do you feel about the new variable name? I had trouble coming up with an adequately descriptive name
 2. Is the documentation descriptive enough for this feature?

### DIFF
--- a/docs/manual.xml.head
+++ b/docs/manual.xml.head
@@ -15947,6 +15947,20 @@ virtual-mailboxes "My INBOX" "notmuch://?query=tag:inbox"
                 </entry>
               </row>
               <row>
+                <entry><literal>nm_query_window_or_terms</literal></entry>
+                <entry>string</entry>
+                <entry><literal>(empty)</literal></entry>
+                <entry>
+                  Additional notmuch search terms to always include in the
+                  window even if they're outside the date range. This turns the
+                  window from <literal>date:...</literal> to 
+                  <literal>date:... or (additional search terms.)</literal>
+
+                  For example, to always include flagged, unread emails, set to
+                  <literal>tag:flagged and tag:unread</literal>
+                </entry>
+              </row>
+              <row>
                 <entry><literal>nm_query_window_timebase</literal></entry>
                 <entry>string</entry>
                 <entry><literal>week</literal></entry>
@@ -16164,6 +16178,8 @@ set virtual_spool_file = no
 set nm_query_window_enable=yes
 set nm_query_window_duration=2
 set nm_query_window_timebase="week" # or "hour", "day", "week", "month", "year"
+<emphasis role="comment"># Extend query window to always show mail matching these terms.</emphasis>
+set nm_query_window_or_terms="tag:unread and tag:flagged"
 <emphasis role="comment"># --------------------------------------------------------------------------</emphasis>
 <emphasis role="comment"># FUNCTIONS – shown with an example mapping</emphasis>
 <emphasis role="comment"># --------------------------------------------------------------------------</emphasis>
--- a/notmuch/config.c
+++ b/notmuch/config.c
@@ -122,6 +122,9 @@ static struct ConfigDef NotmuchVars[] = {
   { "nm_query_window_enable", DT_BOOL, false, 0, NULL,
     "(notmuch) Enable query windows"
   },
+  { "nm_query_window_or_terms", DT_STRING, 0, 0, NULL,
+    "(notmuch) Additional notmuch search terms for messages to be shown regardless of date"
+  },
   { "nm_query_window_timebase", DT_STRING, IP "week", 0, nm_query_window_timebase_validator,
     "(notmuch) Units for the time duration"
   },
--- a/docs/config.c
+++ b/docs/config.c
@@ -2621,6 +2621,12 @@
 ** Accepted values all non negative integers. A value of 0 disables the feature.
 */
 
+{ "nm_query_window_enable", DT_BOOL, false },
+/*
+** .pp
+** This variable enables windowed notmuch queries even if window duration is 0.
+*/
+
 { "nm_query_window_or_terms", DT_STRING, 0 },
 /*
 ** .pp
--- a/docs/config.c
+++ b/docs/config.c
@@ -2621,6 +2621,20 @@
 ** Accepted values all non negative integers. A value of 0 disables the feature.
 */
 
+{ "nm_query_window_or_terms", DT_STRING, 0 },
+/*
+** .pp
+** This variable contains additional notmuch search terms for messages to be
+** shown regardless of date.
+** .pp
+** Example:
+** .pp
+** Using "notmuch://?query=tag:inbox" as the mailbox and "tag:flagged and
+** tag:unread" as the or terms, NeoMutt will produce a query window such as:
+** .pp
+** notmuch://?query=tag:inbox and (date:... or (tag:flagged and tag:unread))
+*/
+
 { "nm_query_window_timebase", DT_STRING, "week" },
 /*
 ** .pp
--- a/notmuch/notmuch.c
+++ b/notmuch/notmuch.c
@@ -282,6 +282,8 @@ static bool windowed_query_from_query(const char *query, char *buf, size_t bufle
       cs_subset_string(NeoMutt->sub, "nm_query_window_current_search");
   const char *const c_nm_query_window_timebase =
       cs_subset_string(NeoMutt->sub, "nm_query_window_timebase");
+  const char *const c_nm_query_window_or_terms =
+      cs_subset_string(NeoMutt->sub, "nm_query_window_or_terms");
 
   /* if the query has changed, reset the window position */
   if (!c_nm_query_window_current_search ||
@@ -293,7 +295,7 @@ static bool windowed_query_from_query(const char *query, char *buf, size_t bufle
   enum NmWindowQueryRc rc = nm_windowed_query_from_query(
       buf, buflen, c_nm_query_window_enable, c_nm_query_window_duration,
       c_nm_query_window_current_position, c_nm_query_window_current_search,
-      c_nm_query_window_timebase);
+      c_nm_query_window_timebase, c_nm_query_window_or_terms);
 
   switch (rc)
   {
--- a/notmuch/query.c
+++ b/notmuch/query.c
@@ -167,6 +167,7 @@ bool nm_query_window_check_timebase(const char *timebase)
  * @param[in]  cur_search Current notmuch search
  * @param[in]  timebase Timebase for `date:` search term. Must be: `hour`,
  *                      `day`, `week`, `month`, or `year`
+ * @param[in]  or_terms Additional notmuch search terms
  * @retval NM_WINDOW_QUERY_SUCCESS  Prepended `buf` with `date:` search term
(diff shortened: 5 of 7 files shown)


docs/manual.xml.head
```diff
@@ -15958,6 +15958,20 @@ virtual-mailboxes "My INBOX" "notmuch://?query=tag:inbox"
                   <literal>year</literal>
                 </entry>
               </row>
+              <row>
+                <entry><literal>nm_query_window_additional_terms</literal></entry>
```
ebblake, line 15962: The name feels long, but is certainly descriptive.  Do you like `nm_query_window_extras` any better?  Shorter, but not as precise.

notmuch/config.c
```diff
@@ -110,6 +110,10 @@ static struct ConfigDef NotmuchVars[] = {
   { "nm_query_type", DT_STRING, IP "messages", 0, NULL,
     "(notmuch) Default query type: 'threads' or 'messages'"
   },
+  {
+    "nm_query_window_additional_terms", DT_STRING, 0, 0, NULL,
+    "(notmuch) additional notmuch search terms for query windows to match on instead of the date."
+  },
```
ebblake, line 116: Maybe:
"(notmuch) additional notmuch search terms for messages to be shown regardless of date."

ebblake (approved):
Some minor comments above, but overall looks good.
I tested on my inbox, and enjoyed being able to see older flagged messages regardless of whether they normally landed in my query window.

I did note that even though I have `$nm_query_type` set to threads, setting `$nm_query_window_additional_terms` to `tag:flagged` picked up _only_ the tagged message and not its out-of-date siblings, rather than the entire thread containing the tagged message.  But using `<entire-thread>` pulled in the related out-of-date messages without too much bother, so I don't know if it is something this patch needs to work on.

I also noticed that the effect of setting the variable on the fly was not immediate, but only took effect the next time I reopened a mailbox or switched the current window.  We may want to add a config observer that recomputes the window whenever any of the `$nm_query_window_*` variables are modified, rather than waiting for the next `<vfolder-window-*>` command.  But that can be done on top, and may depend on @flatcap's event/observer efforts.

brunnre8:
Additional as a description seems wrong. Addition would imply AND in a boolean sense, you are ORing it.

Exclusion is more akin to what you are actually doing

Austin-Ray:
> Additional as a description seems wrong

Please propose an alternative name. I don't think the current one is descriptive (while concise) enough to convey function, but I'm struggling to find a better one.

> Addition would imply AND in a boolean sense, you are ORing it.

Boolean OR is addition[0][1] while boolean AND is multiplication[1][2].

[0] https://en.wikibooks.org/wiki/Practical_Electronics/Logic/Boolean_Arithmetic#Addition
[1] https://bob.cs.sonoma.edu/IntroCompOrg-RPi/sec-balgebra.html
[2] https://en.wikibooks.org/wiki/Practical_Electronics/Logic/Boolean_Arithmetic#Multiplication

docs/manual.xml.head
Austin-Ray: > The name feels long

Agreed.

>  Do you like `nm_query_window_extras` any better?

I think that reduces the descriptiveness too much. Maybe something like `nm_query_window_or_terms` instead? It hints more towards an `OR` statement in the window. I think users will need to look at the manual regardless for more details.

notmuch/config.c
Austin-Ray: That's an improvement. I'm happy to replace this version.

Though I showed both descriptions to a friend and they were still confused about what the variable did. Should we take a more direct approach in the documentation and say something along the lines of:
"(notmuch) convert window to "date:... or (additional terms)" statement."

brunnre8: maybe with an example of what the term looks like after expansion?

>For example, using the mailbox notmuch://?query=tag:inbox as the
mailbox and tag:flagged and tag:unread as the additional search terms
will produce a query window such as:
>
>notmuch://?query=tag:inbox and (date:... or (tag:flagged and tag:unread))


Made it very obvious

Austin-Ray: > maybe with an example of what the term looks like after expansion?

That's a great idea! I'll incorporate that into the PR when I'm back at my dev machine.

flatcap:
> nm_query_window_**additional_terms**

This implies (to me) that we're adding more terms (conditions) to the query.
Logical AND.

We need some language that's a bit more **OR**-y.
Perhaps:

- 'nm_query_window_**merge_terms**'
- 'nm_query_window_**merge_results**'
- 'nm_query_window_**merge_query**'

brunnre8:
not sure merge is a word I'd associate with ORing something... then again I'm bad at giving names.

As long as it's documented with an example we should be fine I think.
The behavior itself is pretty obvious once you see it in action

Austin-Ray:
@flatcap 

> We need some language that's a bit more OR-y.

I dropped `nm_query_window_or_terms` as a possible name further up in the PR comments. With an example, it be clear to user's what its purpose is.

---

@brunnre8 

> I'm bad at giving names.

There's only two hard problem in programming: naming, cache invalidation, and off-by-one errors.

flatcap:
> I also noticed that the effect of setting the variable on the fly was not immediate   

That's easily fixed, as you reasoned, with a config observer.

But...  Is that how we want the windows to work?
(I hadn't really thought about the details before)

If the user creates several windowed queries, surely they should be independent.
A new query would use the config settings and store them in `struct NmMboxData`.
The window functions would allow the user to move the window on _that_ Mailbox.

flatcap:
> I also noticed that the effect of setting the variable on the fly was not immediate

A config observer could be added to either the Account or each Mailbox.

### Account-level

The Account data, `struct NmAccountData`, is created like this:

- `parse_mailboxes()`
- `account_new()`
- `mx_ac_add()`
- `nm_ac_add()`
- `nm_adata_new()`


### Mailbox-level

The Mailbox data, 
**struct NmMboxData**:

- `mx_mbox_open()`
- `nm_mbox_open()`
- `init_mailbox()`
- `nm_mdata_new()`

nm_mdata_free() -- doesn't have a Mailbox ptr
need to alter Mailbox.mdata_free() in mailbox_free()

Austin-Ray:
@flatcap : Are multi-line documentation strings possible in `config.c`? Or must those go elsewhere?

flatcap:
> Are multi-line documentation strings possible in `config.c`?

No.  I only imagined them as a quick _hint_ rather than documentation.

Currently, they're only used in the command line `-O` (for `-D` and `-Q`).
They're probably going to stay, but if they do then at some point they'll need translating
(which at 462 entries would be quite a burden for the translators)

Note: Upstream have just added a <kbd>?</kbd> option to Yes/No and Quad option questions
and that would benefit from adding our one-liner help text.

Austin-Ray:
Incorporated PR comments (please let me know if I missed any by accident), rebased onto master, and add one commit documenting `nm_query_window_enable` in the manual. I had mistakenly thought `notmuch/config.c` was pulled into the manual.

docs/manual.xml.head
```diff
@@ -16164,6 +16178,8 @@ set virtual_spool_file = no
 set nm_query_window_enable=yes
 set nm_query_window_duration=2
 set nm_query_window_timebase="week" # or "hour", "day", "week", "month", "year"
+<emphasis role="comment"># Extend query window to always show mail matching these terms.</emphasis>
+set nm_query_window_or_terms="tag:unread and tag:unread"
```
ebblake, line 16182: Two tag:unread there, one should be tag:flagged

Austin-Ray:
> Two tag:unread there, one should be tag:flagged

Thanks for catching that. Fixed!

On 07/15, Eric Blake wrote:
> @ebblake commented on this pull request.
> 
> 
> 
> > @@ -16164,6 +16178,8 @@ set virtual_spool_file = no
>  set nm_query_window_enable=yes
>  set nm_query_window_duration=2
>  set nm_query_window_timebase="week" # or "hour", "day", "week", "month", "year"
> +<emphasis role="comment"># Extend query window to always show mail matching these terms.</emphasis>
> +set nm_query_window_or_terms="tag:unread and tag:unread"
> 
> Two tag:unread there, one should be tag:flagged
> 
> -- 
> You are receiving this because you authored the thread.
> https://github.com/neomutt/neomutt/pull/3002#pullrequestreview-707947095

-- 
https://austinray.io
Open Source Maintainer, Software Engineer, Keyboard Enthusiast
GPG: 0127 ED83 B939 CCC9 8082  476E 1AA0 B115 C8AC 2C9E

Austin-Ray:
anything pending?

flatcap (approved):

PR #3030, opened 2021-08-31
notmuch: handle missing libnotmuch version bumps

Austin-Ray:
Sometimes libnotmuch's version isn't bumped when new APIs are available.
Some new features require updating our integration to support users
using them, such as separate database and email directories, but we're
unable to use the standard `LIBNOTMUCH_CHECK_VERSION` macro. The current
approach is #ifdef on a function's existence; however, this is
inconsistent with our usage of the version check macro.

Instead, move the function existence checks into
`LIBNOTMUCH_CHECK_VERSION`. Since we know which libnotmuch version
introduced a function, we're able to check the major and minor version
against the function's existence.

There are two libnotmuch version bumps NeoMutt corrects for:
 - libnotmuch 5.1 released in notmuch 0.26 but not bumped until 0.26.1
 - libnotmuch 5.4 released in notmuch 0.32 but not bumped until 0.32.3

---

Note to reviewer: as I'm not super familiar with macros, please take extra care reviewing the change.

### DIFF
--- a/notmuch/private.h
+++ b/notmuch/private.h
@@ -33,12 +33,28 @@ struct Mailbox;
 #undef LIBNOTMUCH_CHECK_VERSION
 #endif
 
-/* The definition in <notmuch.h> is broken */
-#define LIBNOTMUCH_CHECK_VERSION(major, minor, micro)                             \
-  (LIBNOTMUCH_MAJOR_VERSION > (major) ||                                          \
-   (LIBNOTMUCH_MAJOR_VERSION == (major) && LIBNOTMUCH_MINOR_VERSION > (minor)) || \
-   (LIBNOTMUCH_MAJOR_VERSION == (major) &&                                        \
-    LIBNOTMUCH_MINOR_VERSION == (minor) && LIBNOTMUCH_MICRO_VERSION >= (micro)))
+#ifndef HAVE_NOTMUCH_DATABASE_INDEX_FILE
+#define HAVE_NOTMUCH_DATABASE_INDEX_FILE 0
+#endif
+
+#ifndef HAVE_NOTMUCH_DATABASE_OPEN_WITH_CONFIG
+#define HAVE_NOTMUCH_DATABASE_OPEN_WITH_CONFIG 0
+#endif
+
+/*
+ * The definition in <notmuch.h> is broken.
+ *
+ * Corrects for libnotmuch releases with missing version bumps:
+ * - libnotmuch 5.4 released with notmuch 0.32. notmuch 0.32.3 fixed version.
+ * - libnotmuch 5.1 released with notmuch 0.26. notmuch 0.26.1 fixed version.
+ */
+#define LIBNOTMUCH_CHECK_VERSION(major, minor, micro)                              \
+  (major == 5 && minor == 4 && HAVE_NOTMUCH_DATABASE_OPEN_WITH_CONFIG)          || \
+  (major == 5 && minor == 1 && HAVE_NOTMUCH_DATABASE_INDEX_FILE)                || \
+  ((LIBNOTMUCH_MAJOR_VERSION > (major) ||                                          \
+   (LIBNOTMUCH_MAJOR_VERSION == (major) && LIBNOTMUCH_MINOR_VERSION > (minor))  || \
+   (LIBNOTMUCH_MAJOR_VERSION == (major) &&                                         \
+    LIBNOTMUCH_MINOR_VERSION == (minor) && LIBNOTMUCH_MICRO_VERSION >= (micro))))
 
 extern const char NmUrlProtocol[];
 extern const int NmUrlProtocolLen;
--- a/notmuch/db.c
+++ b/notmuch/db.c
@@ -119,7 +119,7 @@ notmuch_database_t *nm_db_do_open(const char *filename, bool writable, bool verb
 
   do
   {
-#if LIBNOTMUCH_CHECK_VERSION(5, 4, 0) || defined(HAVE_NOTMUCH_DATABASE_OPEN_WITH_CONFIG)
+#if LIBNOTMUCH_CHECK_VERSION(5, 4, 0)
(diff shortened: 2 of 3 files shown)


notmuch/private.h
```diff
@@ -34,11 +34,15 @@ struct Mailbox;
...
-    LIBNOTMUCH_MINOR_VERSION == (minor) && LIBNOTMUCH_MICRO_VERSION >= (micro)))
+// notmuch 0.32 introduced libnotmuch 5.4, but didn't include version bump until 0.32.3
+// notmuch 0.26 introduced libnotmuch 5.1, but didn't include version bump until 0.26.1
+#define LIBNOTMUCH_CHECK_VERSION(major, minor, micro)                              \
+  (major == 5 && minor == 4 && defined(HAVE_NOTMUCH_DATABASE_OPEN_WITH_CONFIG)) || \
+  (major == 5 && minor == 1 && defined(HAVE_NOTMUCH_DATABASE_INDEX_FILE))       || \
+  ((LIBNOTMUCH_MAJOR_VERSION > (major) ||                                          \
+   (LIBNOTMUCH_MAJOR_VERSION == (major) && LIBNOTMUCH_MINOR_VERSION > (minor))  || \
+   (LIBNOTMUCH_MAJOR_VERSION == (major) &&                                         \
+    LIBNOTMUCH_MINOR_VERSION == (minor) && LIBNOTMUCH_MICRO_VERSION >= (micro))))
```
flatcap, line 45: The `defined()` preprocessor directives, in a `#define` are causing build problems:

- gcc 11.2.1: `notmuch/db.c:108:37: error: this use of "defined" may not be portable`
- clang 12.0.1: `notmuch/db.c:108:5: error: macro expansion producing 'defined' has undefined behavior`

Perhaps:
```c
#ifndef HAVE_NOTMUCH_DATABASE_INDEX_FILE
#define HAVE_NOTMUCH_DATABASE_INDEX_FILE 0
#endif

#ifndef HAVE_NOTMUCH_DATABASE_OPEN_WITH_CONFIG
#define HAVE_NOTMUCH_DATABASE_OPEN_WITH_CONFIG 0
#endif

#define LIBNOTMUCH_CHECK_VERSION(major, minor, micro)                              \
  (major == 5 && minor == 4 && HAVE_NOTMUCH_DATABASE_OPEN_WITH_CONFIG) ||          \
  (major == 5 && minor == 1 && HAVE_NOTMUCH_DATABASE_INDEX_FILE)       ||          \
  ((LIBNOTMUCH_MAJOR_VERSION > (major) ||                                          \
   (LIBNOTMUCH_MAJOR_VERSION == (major) && LIBNOTMUCH_MINOR_VERSION > (minor))  || \
   (LIBNOTMUCH_MAJOR_VERSION == (major) &&                                         \
    LIBNOTMUCH_MINOR_VERSION == (minor) && LIBNOTMUCH_MICRO_VERSION >= (micro))))
```

I'll keep thinking

gahr:
> The current approach is #ifdef on a function's existence; however, this is
inconsistent with our usage of the version check macro.

If we can't rely on the versioning, let's just check on the function existence. Mixing the two is confusing in my opinion.

flatcap:
> Mixing the two is confusing in my opinion.

I agree, but after the next version bump, we won't need the feature check.
I'd favour checking for the higher version number.
Any users in the middle ground (old version + new function) would have to wait for the upgrade.

gahr:
Until the next screwup

Austin-Ray:
> If we can't rely on the versioning, let's just check on the function existence. Mixing the two is confusing in my opinion.
> Until the next screwup

Versioning has 99% reliability (not scientifically measured) as there's been 2 version screw-ups in ~3.5 years. Releases are complicated and we've screwed up our releases in worse ways. Let's be forgiving and understanding of minor mistakes; they're a part of software development.

My developer rationale for correcting it in our macro is:
 1. A single macro for version-specific code improves code uniformity. Mixing the two is confusing as you said.
 2. It's more informative of libnotmuch versions than function existence (I had to look-up when libnotmuch introduced `notmuch_database_index_file()`)
 3. It abstracts the unnecessary detail of a bad dependency release away from code contributors. Only those touching the macro implementation would need to be aware.

---

> I'd favour checking for the higher version number.
> Any users in the middle ground (old version + new function) would have to wait for the upgrade.

I'm against abandoning the middle ground users as it's a bad user experience for them. They're not going to want to use NeoMutt if the experience is bad. 

In addition, they're not using NeoMutt in a vacuum. Package maintainers have a say in their experience, and there's a myriad of reasons why they choose a version or take a while to update.

It's a near zero maintenance cost for a fix on our end.

---

> Until the next screwup

This will happen again as humans are error prone. Do we want a consistent way to handle future incidents?

gahr:
Ok your future self comes around and modified neomutt to start using a new API from notmuch. I am very much in favour of guarding this usage by the function being available than by a version check.

This fixes all the problems: we won't have mismatches between APIs and versions; users upgrading will get the new feature as soon as it's available; you don't need complicated version checking macros.

It's as simple as this: check for the function in auto.def, use it if it's available.

flatcap:
OK, then...
That's everyone's opinions out of the way :-)

@Austin-Ray It's your code and your decision, so go with what you're happiest with, and I'll merge it.

Thanks

gahr:
> @Austin-Ray It's your code and your decision, so go with what you're happiest with, and I'll merge it.

As long as it's valid C. See n1256, 6.10.1  Conditional inclusion, for where `defined` is valid.

_"If the  token defined is generated as a result of this replacement process or use of the defined unary operator does not match one of the two s pecified forms prior to macro replacement, the behavior is undefined."_

notmuch/private.h
Austin-Ray: Incorporated change

Austin-Ray:
Modified code per @flatcap's suggestion.

Also rephrased macro comment and git commit to clarify that version bump was omitted by accident and which version of notmuch fixed that omission.

flatcap (approved):

Issue #3047, opened 2021-09-30
notmuch: mails marked as old even though mark_old is unset


brunnre8:
Current master (15ebea34b2f9f5dac49bece47368f159aa879cac) ignores the mark_old setting for notmuch mailboxes

>3.184. mark_old
>Type: boolean
>Default: yes
>
>Controls whether or not NeoMutt marks new unread messages as old if you exit a mailbox without reading them. With this option set , the next time you start NeoMutt, the messages will show up with an "O" next to them in the index menu, indicating that they are old.

Even though mark_old is unset, when you open up neomutt you will see items marked as old, rather than unread if they have previously been in the inbox but you didn't read them.

This is caused by the well intentioned commit 
>commit 0c09ab6de3baae0c17338cc70f09a4e356598f63
>Author: Eduardo Habkost (@ehabkost) <ehabkost@redhat.com>
>Date:   Fri Apr 9 19:37:59 2021 -0400
>
>      notmuch: Update Email.old when a maildir file path is detected

It doesn't actually check the corresponding setting though, all other backends first do the check on their own, example maildir:
```
int maildir_parse_dir(struct Mailbox *m, struct MdEmailArray *mda,
                      const char *subdir, struct Progress *progress)
{
  struct dirent *de = NULL;
  int rc = 0;
  bool is_old = false;
  struct MdEmail *entry = NULL;
  struct Email *e = NULL;

  struct Buffer *buf = mutt_buffer_pool_get();

  mutt_buffer_printf(buf, "%s/%s", mailbox_path(m), subdir);
  const bool c_mark_old = cs_subset_bool(NeoMutt->sub, "mark_old");
  is_old = c_mark_old ? mutt_str_equal("cur", subdir) : false;

  DIR *dirp = opendir(mutt_buffer_string(buf));
  if (!dirp)
  {
    rc = -1;
    goto cleanup;
  }

  while (((de = readdir(dirp))) && !SigInt)
  {
    if (*de->d_name == '.')
      continue;

    /* FOO - really ignore the return value? */
    mutt_debug(LL_DEBUG2, "queueing %s\n", de->d_name);

    e = email_new();
    e->edata = maildir_edata_new();
    e->edata_free = maildir_edata_free;

    e->old = is_old;
```

Is the right approach to do the same in notmuch? I could try to fix it up

gahr:
> Is the right approach to do the same in notmuch? I could try to fix it up

[...]

PR #3050, opened 2021-10-03
notmuch: respect mark_old

brunnre8:
Only mark messages as old when the user has set mark_old.

Fixes https://github.com/neomutt/neomutt/issues/3047

Note, that introduces a new user of the global config object, that's what maildir does.
Would you rather have the function accept the config as a parameter?


flatcap (approved):
That looks sensible.

> Would you rather have the function accept the config as a parameter?

Ideally, yes, but I haven't investigated how much work that would be to achieve.

> ... user of the global config object, that's what maildir does.

Yes, this PR is fine.

PR #3055, opened 2021-10-06
fix leaks in address groups

flatcap:
Address groups were cached but not fully freed on exit.

- Refactor the creation/deletion of `struct Group`.
- Add a `HashTable` destructor for `Groups`

---

Simple test case, **test.rc**
```
alternates -group me address1 address2
alternates -group me -group work address3
```

Under ASAN, **run**
```
neomutt -n -F test.rc -B
```


flatcap:
@ebblake 
This should mop up most of the leaks you mentioned.

PR #3056, opened 2021-10-06
fix leak of hidden mailbox

flatcap:
If the current Mailbox isn't in the sidebar, then it would be leaked when exiting NeoMutt.

**Setup**:
- Using several mailboxes: **apple**, **banana**, **cherry**, **damson**
- Config: `mailboxes apple banana`

**Tests**:

1. `set spoolfile = apple`
   `<quit>`

1. `set spoolfile = apple`
   `<change-folder>banana`
   `<quit>`

1. `set spoolfile = cherry`
   `<quit>`

1. `set spoolfile = cherry`
   `<change-folder>damson`
   `<quit>`

1. `set spoolfile = apple`
   `<change-folder>cherry`
   `<quit>`

1. `set spoolfile = cherry`
   `<change-folder>apple`
   `<quit>`

### DIFF
--- a/main.c
+++ b/main.c
@@ -1365,7 +1365,10 @@ int main(int argc, char *argv[], char *envp[])
       mutt_debug(LL_NOTIFY, "NT_MAILBOX_SWITCH: %p\n", m);
       notify_send(dlg->notify, NT_MAILBOX, NT_MAILBOX_SWITCH, &ev_m);
 
-      mutt_index_menu(dlg, m);
+      m = mutt_index_menu(dlg, m);
+      if (m->flags == MB_HIDDEN)
+        mailbox_free(&m);
+
       dialog_pop();
       mutt_window_free(&dlg);
       log_queue_empty();


flatcap:
> free all mailboxes at `main_exit`?

Most are.
Every Mailbox that's configured with a `mailboxes` command will be owned by an `Account`.
`Account`s are owned by the `NeoMutt` object, which is freed on exit.

However, there are quite a few places where there are hidden `Mailbox`es (`Mailbox->flags == MB_HIDDEN`).
e.g. counting postponed emails

We flag them `MB_HIDDEN` so that we know to free them after.

Austin-Ray:
> However, there are quite a few places where there are hidden Mailboxes (Mailbox->flags == MB_HIDDEN).
e.g. counting postponed emails
>
> We flag them MB_HIDDEN so that we know to free them after.

So these are `Account`-less mailboxes created at runtime?

flatcap:
> So these are `Account`-less mailboxes created at runtime?

Yes

One place is `mx_path_resolve()`, e.g. when changing folder.
It scans the `Account`s for a matching `Mailbox`.
If it fails, it creates a new hidden `Mailbox`.

Austin-Ray:
> Yes

Understood. So it's an exception to the architectural changes. It was my understanding that every mailbox was owned by an account.

So why is that this single mailbox isn't free'd unlike other hidden mailboxes?

Issue #3053, opened 2021-10-04
next-unread-mailbox only works every mail_check_stats_interval seconds


vuori:
## Expected Behaviour

`next-unread-mailbox` should go to next unread mailbox.

## Actual Behaviour

If called more often than `mail_check_stats_interval` seconds, function says that there are no unread mailboxes.

## Steps to Reproduce

1. Wait for mail to arrive in a mailbox which is not the current one.
2. Press key bount to `next-unread-mailbox`, note that it will usually work once.
3. Move to another mailbox (without reading the new mail), try it again.
4. "No new messages" will be printed.

## How often does this happen?

- Always

## When did it start to happen?

Switched to master from Ubuntu package (20191207+dfsg.1-1.1) (maybe, I'm not sure if I tried to use the function when using the packaged release).

## NeoMutt Version

```
NeoMutt 20210205-804-7958fd-dirty
Copyright (C) 1996-2020 Michael R. Elkins and others.
NeoMutt comes with ABSOLUTELY NO WARRANTY; for details type 'neomutt -vv'.
NeoMutt is free software, and you are welcome to redistribute it
under certain conditions; type 'neomutt -vv' for details.

System: Linux 5.10.60+ (x86_64)
ncurses: ncurses 6.2.20200212 (compiled with 6.2.20200212)
libidn: 1.33 (compiled with 1.33)
GPGME: 1.13.1-unknown
OpenSSL: OpenSSL 1.1.1f  31 Mar 2020
libnotmuch: 5.4.0
PCRE2: 10.34 2019-11-21
storage: kyotocabinet
compression: lz4, zlib
```

Mail check settings:
```
set mail_check = 5
set mail_check_recent = yes
set mail_check_stats = yes
set mail_check_stats_interval = 60
```

## Extra Info

Using an all-Maildir configuration.

The problem is that `mutt_mailbox_check()` sets `MailboxCount` to 0 even in cases when it's not going to actually check stats on mailboxes due to `mail_check_stats_interval` not having expired, except on the first time it sees a given mailbox.

Diff below fixes the immediate problem, though it may lead to `MailboxCount` being wonky after a new mailbox is added. In practice `first_check_stats_done` being false after the first call to the function sounds rare, but I think a canonical fix would require always either checking stats on all mailboxes or checking none of them. Basically `first_check_stats_done` would need to be checked in a separate pass and `check_stats` set to `true` if at least one mailbox needed a check I guess.

```diff
[...]

Issue #669, opened 2017-07-11
Change in "change folder" due to drop of VirtIncoming


scottkosty:
After 1116748c, if I press 'c', I am given the following prompt:

    Open mailbox ('?' for list): notmuch:///home/scott/.mail?query=date:yesterday..

where before I was given

    Open mailbox ('?' for list):

Was this change in behavior intended? I wasn't sure, but noticed the commit message that the changes should not have a visible effect.

I am on Ubuntu 16.04 and use the following prepare command:

    ./prepare --enable-imap --with-ssl --with-sasl --enable-notmuch --enable-smtp --enable-debug --with-tokyocabinet

If you cannot reproduce, let me know and I can spend some more time figuring out a minimal .muttrc to post.

flatcap:
> changes should not have a visible effect

That was the plan.  The virtual folders _used_ to be kept in a separate list.
I have a plan to group the mailboxes (but there are a few other steps first).

I'll have a think about the prompt.  It'll be easy to filter out the virtual folders, but...
I'm wondering if a notmuch user might _want_ to be offered a virtual folder.

New mail would appear in one real and possibly several virtual folders.
Which would a user rather be prompted with?
Which would _you_ prefer?

Thanks for letting me know.

scottkosty:
> Which would a user rather be prompted with?
> Which would you prefer?

I'm really not sure. I liked having the previous prompt so that I could easily press "tab" and be given a list of folders to change to. But I can easily add a macro that does this. Perhaps we could ask some other notmuch users? @julian-klode @bbenne10 @paratje @WPettersson

bbenne10:
I've somewhat embraced the use of tags and queries over folders and so I don't even *use* this menu. (I have "bind index,pager / vfolder-from-query" set to do a search and show me my results rather than use defined folders of them)

I imagine that the macro is easy enough to add, but I don't particularly have a deeper opinion on the matter at all. While the concept of "folder" is key to mutt, I tend to create and throw away virtual folders pretty regularly...

bbenne10:
(that being said - 1116748 looks to have a lot of negative code, which is awesome!)

scottkosty:
> (that being said - 1116748 looks to have a lot of negative code, which is awesome!)

+1

scottkosty:
I've been thinking about this the last couple of days. I think the reason I like the previous behavior is that I don't like seeing notmuch anywhere unless I call it. I'm not sure where the output

[...]

PR #3058, opened 2021-10-06
fix(index): find next mailbox with unread mail

austin-ray:
Existing implementation of `next-unread-mailbox` finds the next mailbox
with new mail. This contradicts the function's name and
documentation[0], which states:

   The (by default unbound) function `<next-unread-mailbox>` in the
   index can be used to immediately open the next folder with unread
   mail (if any).

To bring implementation in line with the name and documentation,
introduce 'mutt_mailbox_next_unread()', which mirrors
'mutt_mailbox_next()' except it for unread check mail and omission of
stat checking, and call it.

[0] https://neomutt.org/guide/advancedusage#13-2-%C2%A0polling-for-new-mail

-- 

Partial fix for #3053 

Per discussion on IRC, this PR is meant to ensure `next-unread-mailbox` works for the next release. Future PRs will introduce `next-new-mailbox` (more design consideration needed), fix `sidebar-next/prev-new` naming, and update the error message (omitted from this commit so translate work isn't need for PR.)

### DIFF
--- a/index/functions.c
+++ b/index/functions.c
@@ -1342,7 +1342,7 @@ static int op_main_next_undeleted(struct IndexSharedData *shared,
 }
 
 /**
- * op_main_next_unread_mailbox - open next mailbox with new mail - Implements ::index_function_t - @ingroup index_function_api
+ * op_main_next_unread_mailbox - open next mailbox with unread mail - Implements ::index_function_t - @ingroup index_function_api
  */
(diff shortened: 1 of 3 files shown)


gahr:
> It might also be worth revisiting the behavior introduced to #669 to display virtual mailbox names so we can drop that skip all together.

Let's do that, as a separate PR, thanks!

Issue #3066, opened 2021-10-13
more memory leaks found by ASAN, may be notmuch related


ebblake:
## Expected Behaviour
memory leaks should be avoided

## Actual Behaviour
running with ASAN, a couple lingering memory leaks were flagged even after #3055 and friends were fixed.
```

=================================================================
==257422==ERROR: LeakSanitizer: detected memory leaks

Direct leak of 6865791 byte(s) in 305878 object(s) allocated from:
    #0 0x7f669b35fcb8 in __interceptor_realloc (/lib64/libasan.so.6+0xaecb8)
    #1 0x735df8 in mutt_mem_realloc mutt/memory.c:131
    #2 0x7425ea in mutt_str_append_item mutt/string.c:474
    #3 0x7098c1 in driver_tags_getter email/tags.c:66
    #4 0x709e57 in driver_tags_get email/tags.c:147
    #5 0x599c23 in update_email_tags notmuch/notmuch.c:490
    #6 0x5a20ac in nm_mbox_check notmuch/notmuch.c:2170
    #7 0x4a7c4d in mx_mbox_check /home/eblake/neomutt/mx.c:1124
    #8 0x4ce472 in mutt_index_menu index/dlg_index.c:1131
    #9 0x482ab3 in main /home/eblake/neomutt/main.c:1368
    #10 0x7f669a93fb74 in __libc_start_main (/lib64/libc.so.6+0x27b74)

Direct leak of 2750965 byte(s) in 144423 object(s) allocated from:
    #0 0x7f669b35fcb8 in __interceptor_realloc (/lib64/libasan.so.6+0xaecb8)
    #1 0x735df8 in mutt_mem_realloc mutt/memory.c:131
    #2 0x7425ea in mutt_str_append_item mutt/string.c:474
    #3 0x7098c1 in driver_tags_getter email/tags.c:66
    #4 0x709e57 in driver_tags_get email/tags.c:147
    #5 0x599c23 in update_email_tags notmuch/notmuch.c:490
    #6 0x5a20ac in nm_mbox_check notmuch/notmuch.c:2170
    #7 0x4a7c4d in mx_mbox_check /home/eblake/neomutt/mx.c:1124
    #8 0x4f071b in mutt_pager pager/dlg_pager.c:697
    #9 0x501711 in mutt_display_message pager/message.c:436
    #10 0x4d27cc in op_display_message index/functions.c:325
    #11 0x4e4c88 in index_function_dispatcher index/functions.c:3016
[...]

PR #3067, opened 2021-10-14
fix: leak of notmuch tags

flatcap:
#3066 reports an unbelievable amount of leaks (454653).
The 8 unique call-stacks in the ASAN report all end:

- **`update_email_tags()`** https://github.com/neomutt/neomutt/blob/613e2c78fe0a077d7cf154053bbe3af06a0f51f9/notmuch/notmuch.c#L490
- **`driver_tags_get()`** https://github.com/neomutt/neomutt/blob/613e2c78fe0a077d7cf154053bbe3af06a0f51f9/email/tags.c#L147
- **`driver_tags_getter()`** https://github.com/neomutt/neomutt/blob/613e2c78fe0a077d7cf154053bbe3af06a0f51f9/email/tags.c#L66
- **`mutt_str_append_item()`** https://github.com/neomutt/neomutt/blob/613e2c78fe0a077d7cf154053bbe3af06a0f51f9/mutt/string.c#L474

The fix seems to simply be to free the `old_tags` in `update_email_tags()`.

Fixes: #3066

### DIFF
--- a/notmuch/notmuch.c
+++ b/notmuch/notmuch.c
@@ -496,6 +496,7 @@ static int update_email_tags(struct Email *e, notmuch_message_t *msg)
     mutt_debug(LL_DEBUG2, "nm: tags unchanged\n");
     return 1;
   }
+  FREE(&old_tags);
 
   /* new version */
   driver_tags_replace(&e->tags, new_tags);


flatcap:
In the Index, calling `<modify-tags>` doesn't leak any more.

PR #3071, opened 2021-10-18
fix: new mail notification

flatcap:
Hopefully, this will fix all the New Mail issues.

- 85764f13f ibar: observe `NT_MAILBOX`
  Ensure that the Index Bar (status) gets updated.
  Its display can be affected by any Mailbox changing.

- 021d2558f fix: new mail notification
  Send a notification from `mx_mbox_check_stats()` when the Mailbox changes.
  Recalculate `MailboxCount` every time, rather than when stats are checked.

**Tests**:
- New mail arrives in current mailbox (immediate effect)
- New mail arrives in different mailbox (seen within `$mail_check_stats_interval` seconds)
- `<mailbox-list>` shows mailboxes with new mail
- `<change-folder>` defaults to first mailbox with new mail
- Index Bar (status), `Inc: %b` shows number of mailboxes with new mail
- Changing folder to a mailbox with new mail decreases `%b` and removes it from the `<mailbox-list>`

### DIFF
--- a/imap/command.c
+++ b/imap/command.c
@@ -920,6 +920,9 @@ static void cmd_parse_status(struct ImapAccountData *adata, char *s)
   // force back to keep detecting new mail until the mailbox is opened
   if (m->has_new)
     mdata->uid_next = oldun;
+
+  struct EventMailbox ev_m = { m };
+  notify_send(m->notify, NT_MAILBOX, NT_MAILBOX_CHANGE, &ev_m);
 }
 
 /**


flatcap:
> I was thinking within the docstring itself

The pic was the result of a Doxygen, `@note`.  Is that what you meant?
It would appear in the code docs (even if we turn off displaying the code).

```c
/**
 * mx_mbox_check_stats - Check the statistics for a mailbox - Wrapper for MxOps::mbox_check_stats()
 * 
 * @note Emits: #NT_MAILBOX_CHANGE  
 */
```

PR #3081, opened 2021-10-20
fix: move 'check_stats' check to MUTT_NOTMUCH case

austin-ray:
This commits moves the check introduced in commit 0bef239d6 to the
notmuch case. This restores any functionality broken by 0bef239d6 while
keeping the goal of not leaking high-level details into the notmuch
'mbox_check_stats()' implementation.

This commit does not fix the fact that all other 'mbox_check_stats()'
implementations are aware they're called every 'mail_check' and that
they compute additional statistics not listed in the API docs nor
'mail_check_stats' docs. In addition, it does not fix the fact that
'mbox_check_stats()' is called more often than
'mail_check_stats_interval', but all non-notmuch backends handle this
with the 'check_stats' parameter.


Austin-Ray:
Rebased onto master

PR #3082, opened 2021-10-21
Replace NULL with "" in call to notmuch 5.4 call to fix segmentation fault

LarsHaalck:
I haven't found a relevant issue, but in the new release 2021-10-15 together with notmuch 0.33.2, neomutt crashes with a segmentation fault in xapian. After debugging the issue, I found that the call to `notmuch_database_open_with_config` returns the status code `NOTMUCH_STATUS_NO_CONFIG` which in turn leads to a segmentation fault later on. It should not be fatal error, but apparently this sets `config_path` in the query struct to the default value `$HOME/.notmuch-config` instead of setting `xapian_path`.

According the notmuch documentation, `notmuch_database_open_verbose` has been replaced with `notmuch_database_open_with_config` and `config_path=''` instead of `NULL`. 

Setting it to an empty string, fixes this issue for me.

### DIFF
--- a/notmuch/db.c
+++ b/notmuch/db.c
@@ -121,7 +121,7 @@ notmuch_database_t *nm_db_do_open(const char *filename, bool writable, bool verb
   {
 #if LIBNOTMUCH_CHECK_VERSION(5, 4, 0)
     // notmuch 0.32-0.32.2 didn't bump libnotmuch version to 5.4.
-    st = notmuch_database_open_with_config(filename, mode, NULL, NULL, &db, &msg);
+    st = notmuch_database_open_with_config(filename, mode, "", NULL, &db, &msg);
 #elif LIBNOTMUCH_CHECK_VERSION(4, 2, 0)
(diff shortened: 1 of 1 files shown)


bremner:
Lars Haalck ***@***.***> writes:

> Okay I see. I think the issue arises because I don't have a
> `.notmuch-config` in my home folder. Instead I have separate configs
> for different accounts and was using `nm_default_url` in a folder hook
> to switch between the databases. If I copy one of the configs to my
> home folder, it doesn't crash.

Someone would have to translate that into calls to libnotmuch for me to
comment sensibly.

d

LarsHaalck:
Okay, thanks for the quick feedback. I will submit another PR with the changes.

I would also be interested in adding proper config/profile support in the future, but I will implement this fix first.

bremner:
David Bremner ***@***.***> writes:

> This should be treated as fatal by callers, since we didn't succeed in
> opening a Xapian database.

patch applied (with an extra *, thanks to check-notmuch-commit) to
release and master.

PR #3115, opened 2021-10-30
Add url corresponding to $query.

falsifian:
The "query" section doesn't appear in the neomuttrc man page, so the text 'See
"query" for more information wasn't useful there.

NB: I haven't tried re-building the docs with this change (man page or website). Also, it presumably makes the web version of the docs a bit worse since the URL is redundant there, so not sure if it should be merged.

* **What does this PR do?**

Makes the `neomuttrc`  man page more useful.


Austin-Ray (approved):
This seems reasonable to me.

PR #3118, opened 2021-11-01
refactor the quoted colour code

flatcap:
- 894dbf0560 color: move functions to library
- af3075d76d color: rename functions for consistency
- 698d2366d6 color: refactor quoted handling

The first two commits have no functional changes: move; rename.
The third does a little light refactoring:

- Create/use `qstyle_new()`, `qstyle_free()`
- Replace `calloc()`, `strncpy()` with `mutt_strn_dup()`
- Factor out `class_color_new()`


Austin-Ray:
Before digging into the code, is it possible to put some unit tests in place? I see neither existing nor new tests.

color/quoted.c
```diff
@@ -122,3 +122,402 @@ bool quoted_colors_parse_color(enum ColorId color, uint32_t fg, uint32_t bg,
...
+
+/**
+ * qclass_new - Create a new QClass
+ * @retval ptr New QClass
+ */
+static struct QClass *qclass_new(void)
+{
+  struct QClass *qc = mutt_mem_calloc(1, sizeof(struct QClass));
+
+  return qc;
```
Austin-Ray, line 167: As there's no complicated initialization logic and the function providers type details, any objection to `return mutt_mem_calloc(...)`?

color/quoted.h
```diff
@@ -29,16 +29,40 @@
...
+ * and the Email has three levels of indent, then the tree will contain two
+ * siblings each with a child and grandchild.
+ */
+struct QClass
+{
+  int quote_n;                  ///< The quoteN colour index for this level
+  int color;                    ///< Curses colour pair
+  char *prefix;                 ///< Prefix string, e.g. "> "
+  size_t prefix_len;            ///< Length of the prefix string
+  struct QClass *prev, *next;   ///< Different quoting styles at the same level
```
Austin-Ray, line 54: Can we provide any example in the docstring? I'm not 100% sure what this means. 
Is it `> > > Some quote` where `prev` and `next` would one of the `>`s?

flatcap:
> is it possible to put some unit tests in place?

Probably, yes.
But I'm still refactoring the colour code, so it ought to wait a bit.

color/quoted.h
```diff
@@ -29,16 +29,40 @@
...
 
+/**
+ * struct QClass - Style of quoted text
+ *
+ * NeoMutt will store a tree of all the different email quoting levels it
+ * detects in an Email.  If $quote_regex matches, say both "> " and "| ",
+ * and the Email has three levels of indent, then the tree will contain two
+ * siblings each with a child and grandchild.
+ */
+struct QClass
```
Austin-Ray, line 48: As a reader, `QClass` doesn't much context to the struct's function. Any objections to change it to `Quote` or something along those lines?

color/quoted.c
```diff
@@ -122,3 +122,402 @@ bool quoted_colors_parse_color(enum ColorId color, uint32_t fg, uint32_t bg,
...
+
+  FREE(&qc->prefix);
+  FREE(ptr);
+}
+
+/**
+ * qclass_free_tree - Free a tree of QClass
+ * @param[out] quote_list Quote list to free
+ */
+void qclass_free_tree(struct QClass **quote_list)
```
Austin-Ray, line 145: Are there benefits to naming this `qclass_free_tree()` instead of `qclass_free()`? It breaks the expected pattern of `<struct>_free()` for cleanup that the codebase follows. I can see contributors calling `qclass_free()` by mistake.

nit: the `QClass` uses a bidirectional graph instead of a tree.

color/quoted.c
```diff
@@ -122,3 +122,402 @@ bool quoted_colors_parse_color(enum ColorId color, uint32_t fg, uint32_t bg,
   *rc = MUTT_CMD_SUCCESS;
   return true;
 }
+
+/**
+ * qclass_free - Free a single QClass
+ * @param ptr QClass to free
+ */
```
Austin-Ray, line 129: We should note in the docstring that `qclass_free_tree()` should be called instead of `qclass_free()` otherwise we'll leak memory. I can easily see contributors making this mistake and it slipping past review.

Austin-Ray (changes requested):
Changes look good to me so far. Have a few comments on it.

nit: can we include a list of changed names in the commit message for  d76932f23d1ea6d9d90c112156b711d0169c741e. Makes it easier to see what changed without reading through all the code.

Austin-Ray:
> Probably, yes.
> But I'm still refactoring the colour code, so it ought to wait a bit.

Tests should be part of the refactoring process to ensure consistent behavior (or explicit changes from existing behavior.) Easier to catch bugs early in the process.

color/quoted.c
flatcap: done

color/quoted.h
flatcap: I've added some GraphViz to the docstring.
It'll get rendered as something like:
image

color/quoted.h
flatcap: Changed to `QuoteStyle`

color/quoted.c
flatcap: It means that `_new()` and `_free()` are equal and opposite.
I've added `@note`s emphasising the difference between `_free()` and `_free_tree()`

color/quoted.c
flatcap: see above

flatcap:
> Tests should be part of the refactoring process 

That's a lovely sentiment, but in this case, no.
The current colour code is essentially untestable -- too many dependencies and poor design.

My current work branch sorts out the design, a bit, and abstracts more away from the caller.
_That_ might be amenable to testing -- if you're volunteering.

color/quoted.h
Austin-Ray: What does that picture mean? I don't understand what it's trying to convey.

flatcap: That's the tree of `QuoteStyle` objects representing the quoted lines found in the email.

It's a complicated example, but it represents and email with many replies and two styles of reply-indent:
- People replying by adding `> `
- People replying by adding `| `

As there have been many replies to replies, some lines have multiple `> ` indents (and some multiple `| ` indents).

Each QuoteStyle (blue box) stores the pattern to match, e.g. `> >`, and has a `color` that the text should be displayed.

Austin-Ray: So the fields are to different `prefixes` at the same quote depth?

flatcap: yes

PR #3121, opened 2021-11-03
Avoid unnecessary gettimeofday calls during progress update

gahr:
This gives me a 15% speedup at mailbox open time.

((please double check my boolean logic))

### DIFF
--- a/progress/progress.c
+++ b/progress/progress.c
@@ -176,16 +176,18 @@ static bool time_needs_update(const struct Progress *progress, size_t now)
  */
 void progress_update(struct Progress *progress, size_t pos, int percent)
 {
-  if (!progress || OptNoCurses)
+  if (OptNoCurses || !progress || (progress->inc == 0))
     return;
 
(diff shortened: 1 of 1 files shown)


ebblake (approved):
logic was:
if (a || b) return
if (c && (d || (e && f))) do stuff else return

by deMorgan's, we can rewrite the latter to:
if (!c || (!d && (!e || !f))) return else do stuff

then hoisting things around, we get:
if (b || a || c) return
if (!d && (!e || !f)) return else do stuff

using distributive law for another rewrite of the second gives:
if (!d && !e) return
if (!d && !f) return else do stuff

then one more deMorgan's on the last gives:
if (d || f) do stuff else return

the logic change is correct.

flatcap:
Note for Linux users:

In Linux `gettimeofday()` is handled by a magic shared library: `linux-vdso.so.1`
This avoids having to make a syscall, which would involve slow context switching.
For Linux, this PR doesn't measurably change the speed of the progress functions.

flatcap (approved):

PR #3124, opened 2021-11-06
Fix scanf/printf bugs (one critical, three less so) and constant string warning

nabijaczleweli:
See each commit for warnings from GCC 11.2.0-10.

Found these when building for x32, remaining warnings for a `--mandir=/usr/share/man --libexecdir=/usr/libexec --with-mailpath=/var/mail --gpgme --lua --notmuch --with-ui --gnutls --gss --idn --mixmaster --sasl --tokyocabinet --zlib --zstd --testing` build:
```
notmuch/notmuch.c: In function ‘rename_filename’:
notmuch/notmuch.c:1238:29: warning: ‘%s’ directive output may be truncated writing 3 bytes into a region of size between 0 and 4095 [-Wformat-truncation=]
 1238 |   snprintf(buf, buflen, "%s/%s/%s%s", folder,
      |                             ^~
notmuch/notmuch.c:1238:3: note: ‘snprintf’ output between 6 and 12291 bytes into a destination of size 4096
 1238 |   snprintf(buf, buflen, "%s/%s/%s%s", folder,
      |   ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
 1239 |            (e->read || e->old) ? "cur" : "new", filename, suffix);
      |            ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
mutt/regex.c: In function ‘mutt_replacelist_apply’:
mutt/regex.c:406:15: warning: ‘strncpy’ output may be truncated copying between 0 and 1023 bytes from a string of length 2047 [-Wstringop-truncation]
  406 |               strncpy(&dst[tlen], src, cpysize);
      |               ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
map autocrypt account undefined at ./docs/gen-map-doc line 64, <XML> line 26.
Error: no ID for constraint linkend: "tunnel".
Error: no ID for constraint linkend: "tunnel-is-secure".
Error: no ID for constraint linkend: "tunnel-is-secure".
Error: no ID for constraint linkend: "autocrypt".
Error: no ID for constraint linkend: "autocrypt-dir".
Error: no ID for constraint linkend: "autocrypt-dir".
Error: no ID for constraint linkend: "autocrypt-dir".
Error: no ID for constraint linkend: "autocrypt-dir".
Error: no ID for constraint linkend: "autocrypt-dir".
Error: no ID for constraint linkend: "autocrypt-dir".
Error: no ID for constraint linkend: "autocrypt-reply".
Error: no ID for constraint linkend: "autocrypt-account-map".
Error: no ID for constraint linkend: "autocrypt-dir".
Error: no ID for constraint linkend: "autocrypt-dir".
Error: no ID for constraint linkend: "autocrypt-dir".
Error: no ID for constraint linkend: "autocrypt-reply".
Error: no ID for constraint linkend: "net-inc".
Error: no ID for constraint linkend: "net-inc".
Error: no ID for constraint linkend: "net-inc".
```

The second one is verified above, but the first one is valid. I'm not sure why a path is hard-`PATH_MAX`-sized, but y'know. It should probably at least get overflow-checked and return -1 if it's passed verbatim to a syscall, or dynamically allocated if it isn't, but I don't use notmuch, so.

### DIFF
--- a/browser.c
+++ b/browser.c
@@ -1563,7 +1563,7 @@ void mutt_buffer_select_file(struct Buffer *file, SelectFileFlags flags,
               {
                 char *p = NULL;
                 if (lastdirlen > 1)
-                  p = strrchr(mutt_buffer_string(&LastDir) + 1, '/');
+                  p = strrchr(LastDir.data + 1, '/');
 
                 if (p)
                 {
--- a/nntp/newsrc.c
+++ b/nntp/newsrc.c
@@ -38,6 +38,7 @@
 #include <sys/stat.h>
 #include <time.h>
 #include <unistd.h>
+#include <inttypes.h>
 #include "private.h"
 #include "mutt/lib.h"
 #include "config/lib.h"
@@ -625,7 +626,7 @@ static int active_get_cache(struct NntpAccountData *adata)
   if (!fp)
     return -1;
 
-  if (!fgets(buf, sizeof(buf), fp) || (sscanf(buf, "%ld%4095s", &t, file) != 1) || (t == 0))
+  if (!fgets(buf, sizeof(buf), fp) || (sscanf(buf, "%" SCNd64 "%4095s", &t, file) != 1) || (t == 0))
   {
     mutt_file_fclose(&fp);
     return -1;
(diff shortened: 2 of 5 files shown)


flatcap (approved):
Nice work! Thanks

PR #3125, opened 2021-11-07
Support XDG Base Directory Specification ... a little more

darrel-glud:
Dear Neomutt developers

* **What does this PR do?**

Better start with the problem this PR tries to solve:
To follow the XDG Directory Specification [0] it is not simply reading its config from `~/.config/neomutt` but also to place associated files in the corresponding directory, e.g. cache files in $XDG_CACHE_HOME.
While I don't expect neomutt to support XDG out of the box, it would be nice if the user can at least configure it to behave that way. Let's take `header_cache` as an example. It should go into `$XDG_CACHE_HOME`

```
set header_cache = "~/.cache/neomutt/headercache"
```
but that doesn't do the job if `$XDG_CACHE_HOME` is something other than `~/.cache`. Likewise,
```
set header_cache = "$XDG_CACHE_HOME/neomutt/headercache"
```
fails if XDG_CACHE_HOME is not set (in which case `~/.cache/` should be used).
I'm sure you can do something with ifdef, shell, sourcing a script, ... but I figured it might be easier to provide the user with the XDG environment variables, so a simple $xdg_dir_you_want will do.

To this end this PR introduce config variables
* `$xdg_app_data_home`
* `$xdg_app_config_home`
* `$xdg_app_state_home`
* `$xdg_app_cache_home`

Which are their corresponding XDG environment variables substituted with defaults and "/neomutt/" appended, e.g. `$xdg_app_config_home` is `~/.config/neomutt`. The above task is now easily solved with
```
set header_cache = "$xdg_app_cache_home/headercache"
```

[0] https://specifications.freedesktop.org/basedir-spec/basedir-spec-latest.html

Additional notes:

1. `mutt_set_xdg_path()` could be simplified to a two-liner if the order of the config would be
```
confdir1/neomuttrc
confdir2/neomuttrc
...
confdir3/neomuttrc
confdir1/muttrc
confdir2/muttrc
...
confdir3/muttrc
```
currently it is alternating `neomuttrc` and `muttrc`. I don't expect to change that, just wanted to point out why I couldn't simplify that function further.

2. I considered moving `mutt_set_xdg_path()` into `xdg.c` but did not for two reasons.
  a) that function is not for general XDG dir handling but ensuring a particular order of config files
  b) wanted to keep the diff small.
 
3. I didn't know where to place the `xdg.c` / `xdg.h`. Maybe mutt/ would be a good place but then again I have no idea what the guidelines are for inclusion. So I placed it just "somewhere". Feel free to move that.

4. There is not documentation for the config variables yet, as they seem to be not documented in manual.xml.head. Hints on where to put them are welcome.

5. I named the config variables `xdg_app_data_dir` instead of `xdg_data_dir` (note the `APP`) because I added `neomutt` to the end and so this cannot be confused with a potential `$xdg_data_dir` which does not end in `neomutt/`.

6. The backend works for all environment variables defined by the XDG Base Directory Specification but only those containing a single directory are exported as config variables. I see little use in exporting the XDG environment variables which are a colon separated list of directories.

7. I don't usually program C so there might be some more efficient ways of achieving this. Nicely documented code base BTW.

8. In `config_init_main()` in mutt_config.c the call to `cs_register_variables()` returns `false`. This has nothing to do with this PR since it did it already before I started working. According to the doc this means that some variables could not be registered. This might or might not indicate a bug.

9. If you feel this is not the right way to expose XDG directories to the users config, please say so and close this PR.

10. How to set the default value for these new variables? Currently a `reset xdg_app_data_home` resets it to `./`.
      (This in particular applies to a potential `reset all` call).

Darrel

### DIFF
--- a/Makefile.autosetup
+++ b/Makefile.autosetup
@@ -72,7 +72,7 @@ NEOMUTTOBJS=	alternates.o commands.o command_parse.o \
 		mutt_logging.o mutt_mailbox.o mutt_signal.o mutt_socket.o \
 		mutt_thread.o mx.o mview.o myvar.o opcodes.o recvcmd.o \
 		resize.o rfc3676.o score.o sort.o status.o subjectrx.o \
-		system.o version.o
+		system.o version.o xdg.o
 
 @if !HAVE_TIMEGM
 NEOMUTTOBJS+=	timegm.o
--- a/docs/config.c
+++ b/docs/config.c
@@ -5619,5 +5619,82 @@
 ** name of original article author) to article that followuped to newsgroup.
 */
 #endif
+
+{ "xdg_app_data_home", DT_PATH, "$XDG_DATA_HOME/neomutt" },
+/*
+** .pp
+** The family of $$xdg_FOO config variables is a group of convenience variables
+** to get access to value of XDG_FOO environment variable.  The covered XDG
+** variables are:
+** .dl
+** .dt \fBName\fP      .dd \fBDefault value\fP
+** .dt XDG_DATA_HOME   .dd ~/.local/share
+** .dt XDG_CONFIG_HOME .dd ~/.config
+** .dt XDG_STATE_HOME  .dd ~/.local/state
+** .dt XDG_CACHE_HOME  .dd ~/.cache
+** .de
+** They all work the same, in the following we use $$xdg_app_cache_home as
+** example.
+** .pp
+** $$xdg_app_cache_home (note the "app" in the name) expands to
+** "XDG_CACHE_HOME/neomutt" if the environment variable XDG_CACHE_HOME is set,
+** otherwise is expands to the default as defined in the XDG Base Directory
+** Specification (see list above), in the case of XDG_CACHE_HOME it expands to
+** "~/.cache/neomutt".
+** .pp
+** The purpose of this can best be seen on an example:
+** .ts
+** set header_cache = "$$XDG_CACHE_HOME/neomutt/headers"
+** .te
+** The above works as intended if the environment variable XDG_CACHE_HOME is
+** set.  However, if it is not then the header cache would be
+** "/neomutt/headercache" which was not intended.  Using the
+** $$xdg_app_cache_home variable like so
+** .ts
+** set header_cache = "$$xdg_app_cache_home/headers"
+** .te
+** avoids this problem and correctly assigns "~/.cache/neomutt/headers" in the
+** case that XDG_CACHE_HOME is not set.
+** .pp
+** Note: NeoMutt does not use these variables internally, they are solely to
+** help writing your config file and should be considered read-only.  Setting
+** them will have no effect on NeoMutt or the XDG environment variables.  If
+** you want to change the XDG environment variables you have to do that outside
+** of NeoMutt before you launch NeoMutt.
+*/
+
+{ "xdg_app_config_home", DT_PATH, "$XDG_CONFIG_HOME/neomutt" },
+/*
+** .pp
+** See $$xdg_app_data_home for the documentation.
+*/
+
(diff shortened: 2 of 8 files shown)


flatcap:
Thanks for the PR, @darrel-glud.
It seems like a reasonable thing to do
(and a quick scan of the code looks good :-)

It's going to take little while to unpack and review it all.

darrel-glud:
Ahh, I forgot one thing:

10. How to set the default value for these new variables? Currently a `reset xdg_app_data_home` resets it to `./`.
      (This in particular applies to a potential `reset all` call).

(I also included it in the initial post, so you only have to skim over one list)

darrel-glud:
Dear @flatcap

> It's going to take little while to unpack and review it all.

Can you give me a status update on the progress of this?
(I don't mind the long time, it is just a friendly reminder to prevent this from falling into oblivion.)

Darrel

flatcap:
Hi Darrel,

Sorry for the long silences.  I'm still interested in the PR.

For over six months I've been deeply buried, refactoring the GUI code (I still am).
Despite that, I've made time to regularly rebase your code.

I'll make some time to understand how best to integrate it.

darrel-glud:
Hi @flatcap 

> For over six months I've been deeply buried, refactoring the GUI code (I still am).
> Despite that, I've made time to regularly rebase your code.
>
> I'll make some time to understand how best to integrate it.

No worries. :-)
Please, finish your current refactoring first before putting the next iron in the fire.  This PR can wait – and so can I.

darrel-glud:
It's been almost another year now. I've rebased the this PR to the newest master and added some documentation for the manual of the new variables.  They now also reset properly to their initial value (instead of `""`) when using reset. So from the daunting list of 10 above.  There is nothing left but ramblings.

The doc-build fails because I use `$` to indicate an environment variable in my documentation but `$` seems to be reserved for some kind of reference.  A hint how to get a literal `$` in the docs would be welcome. And finalise this PR on my part.

gahr:
Hi, sorry, but we're going to reject this one. We think this overly complicates things.

Your starting point is this:

> Likewise,
> set header_cache = "$XDG_CACHE_HOME/neomutt/headercache"
> fails if XDG_CACHE_HOME is not set (in which case ~/.cache/ should be used).

but we can assume a user controls *both* its environment variables and its NeoMutt config, so you can very well set both XDG_CACHE_HOME in your env and use it in neomutt.rc, or not set it in your env and not use it in neomutt.rc.

We think it's fine to default a few things to xdg-compliant paths, like we're already doing for some things, but introducing NeoMutt variables to map environment variables sounds a bit overkill.

darrel-glud:
Sad news as this means that NeoMutt configs are less portable and more fragile.

For instance, I have two systems which I'd like to share my NeoMutt config, one has (some) `XDG_*` variables, the other runs the usual default (like millions of others) of not having set any of the `XDG_*` environment variables.  After porting my config to use `$XDG_*`, I must not forget to explicitly set the XDG environment to their *defaults* on all current and any future machine otherwise my NeoMutt config will silently break as some paths become broken (`$XDG_*` gets replaced by "").

---

NB: You might want have a look at the first two commits which refactors some of the XDG code and have nothing to do with config variables.

gahr:
You can populate your settings by sourcing an external script

### ~/.config/neomutt/hcache-path.sh

```sh
#!/bin/sh

if [ -n "$XDG_CACHE_HOME" ]; then
    p="$XDG_CACHE_HOME"/neomutt/headercache
else
    p="~/.cache/neomutt/headercache"
fi

echo "set header_cache='$p'"
```

### ~/.config/neomutt/neomuttrc
```
source ~/.config/neomutt/hcache-path.sh
```

PR #3128, opened 2021-11-10
Fix back-reference parsing in mutt_replacelist_add

gahr:
A back-reference not at the end of the string was not properly parsed.
This is because our mutt_str_ato[li] functions explicitely check that
the full string was parsed. Perhaps we should do something about it.

Also, without the fix, the test would crash because `mutt_replacelist_match` keeps some static variables that are not properly reset.

### DIFF
--- a/test/regex/mutt_replacelist_match.c
+++ b/test/regex/mutt_replacelist_match.c
@@ -23,6 +23,7 @@
 #define TEST_NO_MAIN
 #include "config.h"
 #include "acutest.h"
+#include "test_common.h"
 #include "mutt/lib.h"
 
 void test_mutt_replacelist_match(void)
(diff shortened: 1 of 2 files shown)


flatcap:
> our mutt_str_ato[li] functions explicitly check that the full string was parsed.

Upstream recently(-ish) modified their `ato*` functions to behave consistently,
but that may have just been the function prototypes.

> Perhaps we should do something about it.

Sounds good.
There are currently 50 uses of `mutt_str_{atoi,atol,atos,atoui,atoul,atoull}()`.

test/regex/mutt_replacelist_match.c
```diff
@@ -44,4 +45,20 @@ void test_mutt_replacelist_match(void)
...
+    TEST_CHECK_STR_EQ("foo [foo-1234-bar] bar", buf);
+  }
+
+  {
+    struct ReplaceList replacelist = STAILQ_HEAD_INITIALIZER(replacelist);
+    mutt_replacelist_add(&replacelist, "foo-([^-]+)-bar", "foo [%1] bar", NULL);
+    char buf[32] = { 0 };
+    TEST_CHECK(mutt_replacelist_match(&replacelist, buf, sizeof(buf), "foo-1234-bar"));
+    TEST_CHECK_STR_EQ("foo [1234] bar", buf);
+  }
```
flatcap, line 63: Is it worth adding a failing test for an illegal back-reference: `%2`?
```c
  {
    struct ReplaceList replacelist = STAILQ_HEAD_INITIALIZER(replacelist);
    mutt_replacelist_add(&replacelist, "foo-([^-]+)-bar", "foo [%2] bar", NULL);
    char buf[32] = { 0 };
    TEST_CHECK(!mutt_replacelist_match(&replacelist, buf, sizeof(buf), "foo-1234-bar"));
  }
```

flatcap (approved):

Issue #3131, opened 2021-11-12
Neomuttrc man page broken


john-at-charpa:
Being ye old unix kid I often use the man page(s) for neomutt. Unfortunately the formatting for neomuttrc has been broken for a long time now. I finally got tired of the errors and fixed it. Due to the size of the table, simply fixing the formatting syntax isn't sufficient - you simply get a new error about it being too large. So I've split the table into two logical tables that fit nicely and should be legible.

I've submitted a PR #3132 for this.

PR #3132, opened 2021-11-12
Fix broken formatting and split tables to avoid formatting and length man errors

john-at-charpa:
* **What does this PR do?**
Fixes the formatting errors in the neomuttrc man page. There are problems with both formatting and the table size.

#3131


flatcap (approved):
Thanks for diving into the scary world of nroff :-)
It's much appreciated ❤️

john-at-charpa:
> Thanks for diving into the scary world of nroff :-)
> It's much appreciated ❤️

Happy to help where I can; thanks jumping on it :)

PR #3134, opened 2021-11-16
Improve API of strto* wrappers

gahr:
The signature of the `mutt_str_ato*` functions have been changed to return the endptr argument to the relative libc functions. In addition, I have added wrappers named `mutt_str_ato*_full` that return a bool indicating whether a successful conversion consumed the whole string. So, the changes at API level are

```diff
- if (mutt_str_ato*(...) < 0)
+ if (!mutt_str_ato*(...))
```

and

```diff
- if (mutt_str_ato*(...) != 0)
+ if (!mutt_str_ato*_full(...))
```

Additionally, whoever needs more control can use the returned endptr, see e.g., the changes in mutt/regex.c .

### DIFF
--- a/color/command.c
+++ b/color/command.c
@@ -325,8 +325,7 @@ static enum CommandResult parse_object(struct Buffer *buf, struct Buffer *s,
     int val = 0;
     if (buf->data[6] != '\0')
     {
-      rc = mutt_str_atoi(buf->data + 6, &val);
-      if ((rc != 0) || (val > COLOR_QUOTES_MAX))
+      if (!mutt_str_atoi_full(buf->data + 6, &val) || (val > COLOR_QUOTES_MAX))
       {
         mutt_buffer_printf(err, _("%s: no such object"), buf->data);
         return MUTT_CMD_WARNING;
@@ -639,7 +638,7 @@ static enum CommandResult parse_color(struct Buffer *buf, struct Buffer *s,
     {
       struct Buffer tmp = mutt_buffer_make(0);
       mutt_extract_token(&tmp, s, MUTT_TOKEN_NO_FLAGS);
-      if (mutt_str_atoui(tmp.data, &match) < 0)
+      if (!mutt_str_atoui_full(tmp.data, &match))
       {
         mutt_buffer_printf(err, _("%s: invalid number: %s"),
                            color ? "color" : "mono", tmp.data);
--- a/mutt/string.c
+++ b/mutt/string.c
@@ -175,35 +175,108 @@ size_t mutt_istr_startswith(const char *str, const char *prefix)
 }
 
 /**
- * mutt_str_atol - Convert ASCII string to a long
+ * str_atol_clamp - Convert ASCII string to a long and clamp
  * @param[in]  str String to read
+ * @param[in]  lmin Lower bound
+ * @param[in]  lmax Upper bound
  * @param[out] dst Store the result
- * @retval  0 Success
- * @retval -1 Error
- * @retval -2 Overflow
+ * @retval endptr
+ *
+ * endptr    == NULL -> no conversion happened, or overflow
+ * endptr[0] == '\0' -> str was fully converted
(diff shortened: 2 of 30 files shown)


mutt/string.c
```diff
@@ -247,118 +288,119 @@ int mutt_str_atos(const char *str, short *dst)
...
-  if (rc < 0)
-    return rc;
-  if ((res < INT_MIN) || (res > INT_MAX))
-    return -2;
-
-  if (dst)
-    *dst = (int) res;
+  long l;
+  const char *res = str_atol_clamp(str, &l, SHRT_MIN, SHRT_MAX);
+  dst && (*dst = res ? l : 0);
```
flatcap, line 298: gcc 11.2.1 warns about this (and the 4 others below)
```
mutt/string.c: In function ‘mutt_str_atos’:
mutt/string.c:298:7: error: value computed is not used
mutt/string.c: In function ‘mutt_str_atoi’:
mutt/string.c:320:7: error: value computed is not used
mutt/string.c: In function ‘mutt_str_atoui’:
mutt/string.c:341:7: error: value computed is not used
mutt/string.c: In function ‘mutt_str_atoul’:
mutt/string.c:362:7: error: value computed is not used
mutt/string.c: In function ‘mutt_str_atous’:
mutt/string.c:384:7: error: value computed is not used
```

color/command.c
```diff
@@ -639,7 +638,7 @@ static enum CommandResult parse_color(struct Buffer *buf, struct Buffer *s,
     {
       struct Buffer tmp = mutt_buffer_make(0);
       mutt_extract_token(&tmp, s, MUTT_TOKEN_NO_FLAGS);
-      if (mutt_str_atoui(tmp.data, &match) < 0)
+      if (!mutt_str_atoui(tmp.data, &match))
```
flatcap, line 641: `mutt_str_atoui_full()`

resize.c
```diff
@@ -80,14 +80,14 @@ void mutt_resize_screen(void)
   if (screenrows <= 0)
   {
     const char *cp = mutt_str_getenv("LINES");
-    if (cp && (mutt_str_atoi(cp, &screenrows) < 0))
+    if (cp && !mutt_str_atoi(cp, &screenrows))
```
flatcap, line 83: `mutt_str_atoi_full()`

resize.c
```diff
@@ -80,14 +80,14 @@ void mutt_resize_screen(void)
...
       screenrows = 24;
   }
 
   if (screencols <= 0)
   {
     const char *cp = mutt_str_getenv("COLUMNS");
-    if (cp && (mutt_str_atoi(cp, &screencols) < 0))
+    if (cp && !mutt_str_atoi(cp, &screencols))
```
flatcap, line 90: `mutt_str_atoi_full()`

score.c
```diff
@@ -141,7 +141,7 @@ enum CommandResult mutt_parse_score(struct Buffer *buf, struct Buffer *s,
     ptr->exact = true;
     pc++;
   }
-  if (mutt_str_atoi(pc, &ptr->val) < 0)
+  if (!mutt_str_atoi(pc, &ptr->val))
```
flatcap, line 144: `mutt_str_atoi_full()`

flatcap (changes requested):
So far I've worked through ~2/3 of the changes
(starting with the easiest, most-debuggable ones).

Most of the comments are where we're expecting a NUL-terminated string that should be a plain number.
It's a bit repetitive, sorry :-)

color/command.c
gahr: I agree, but this is changing the current behaviour, where `color status red default '\[([^]]+)\]' 1foobar` works.

flatcap: I vote for fixing the behaviour.
I can't imagine anyone relying on this mis-parsing.

(`color status fg bg pattern` is a NeoMutt extension, so we haven't inherited the code)

flatcap:
rebased, squashed, clang-formatted

flatcap (approved):

PR #3137, opened 2021-11-17
gpgme: fix crash with partially defined keys

flatcap:
If the key isn't held locally, `uid` may be `NULL`.

Situation:
- Using GPGME
- Decrypting an email
- Decryption key doesn't have a UID

If the encrypted email contains a PGP key, then NeoMutt will display a block, like this:
```
[-- BEGIN PGP PUBLIC KEY BLOCK --]  
pub EdDSA 255/ABC123DE 2020-11-29 John Doe <jd@example.com>
sub  ECDH 255/456DEF78 2020-11-29
sub EdDSA 255/12345678 2020-11-29
[-- END PGP PUBLIC KEY BLOCK --]
```

If the UID ("John Doe...") isn't defined, then `uid` in `pgp_gpgme_extract_keys()` will be `NULL` leading to a crash.

---

Thanks :heart: to Nick Johnson for his help debugging.

### DIFF
--- a/ncrypt/crypt_gpgme.c
+++ b/ncrypt/crypt_gpgme.c
@@ -2518,15 +2518,10 @@ static int pgp_gpgme_extract_keys(gpgme_data_t keydata, FILE **fp)
         fprintf(*fp, "sub %5.5s %u/%8s %s\n", gpgme_pubkey_algo_name(subkey->pubkey_algo),
                 subkey->length, shortid, date);
       }
-      else if (uid)
-      {
-        fprintf(*fp, "pub %5.5s %u/%8s %s %s\n", gpgme_pubkey_algo_name(subkey->pubkey_algo),
-                subkey->length, shortid, date, uid->uid);
-      }
       else
       {
-        fprintf(*fp, "pub %5.5s %u/%8s %s\n", gpgme_pubkey_algo_name(subkey->pubkey_algo),
-                subkey->length, shortid, date);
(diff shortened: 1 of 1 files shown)


ncrypt/crypt_gpgme.c
```diff
@@ -2518,11 +2518,16 @@ static int pgp_gpgme_extract_keys(gpgme_data_t keydata, FILE **fp)
...
+      else if (uid)
       {
         fprintf(*fp, "pub %5.5s %u/%8s %s %s\n", gpgme_pubkey_algo_name(subkey->pubkey_algo),
                 subkey->length, shortid, date, uid->uid);
       }
+      else
+      {
+        fprintf(*fp, "pub %5.5s %u/%8s %s\n", gpgme_pubkey_algo_name(subkey->pubkey_algo),
+                subkey->length, shortid, date);
+      }
```
gahr, line 2530: if you can live with an extra space at the end of the line:

```suggestion
      else
      {
        fprintf(*fp, "pub %5.5s %u/%8s %s %s\n", gpgme_pubkey_algo_name(subkey->pubkey_algo),
                subkey->length, shortid, date, (uid ? uid->uid : ""));
      }
```

flatcap: I thought about that...

> if you can live with an extra space at the end of the line:

The perfectionist in me shuddered at the thought!
The text is displayed to the user, but not used anywhere 'critical', so I can live with a non-visible space.

gahr (approved):

Issue #2936, opened 2021-05-18
SASL password field size might need bumping to at least 2048 bytes


ivucica:
## What would you like to be able to do?

Google's OAuth2 access token will soon be up to 2048 bytes.

I've just received the following email:

> Google Cloud Platform will roll out security and reliability improvements that will increase the sizes of OAuth 2.0 tokens. Please follow the instructions below to use the documented token size limits of 2048 bytes and to remove any logic in your services/code that restricts its ability to process access tokens of certain sizes before August 23, 2021.
>
> Hello Google Cloud Customer,
>
> We’re writing to let you know that on August 23, 2021, we will roll out security and reliability improvements that will increase the sizes of OAuth 2.0 access tokens that your services may be using to call Google Cloud Platform APIs.
>
> The access token size increase will only impact services that don’t conform to the access token size limits of 2048 bytes established on Google’s Developer guide and RFC 6749. If your service falls into this category, please continue reading and take action as outlined below.

This linked to https://developers.google.com/identity/protocols/oauth2#size

I did only some cursory reading of the code, but I believe NeoMutt (and likely Mutt) currently use at most 256 bytes for passwords.

## What needs to change in NeoMutt?

I believe this might be in https://github.com/neomutt/neomutt/blob/fdf8eae4f74bcbca76f1c26eb7e12f1cdf12ec20/imap/auth_login.c#L46 -- but I am not certain.

Password field was bumped from 128 to 256 in #454. I'm having a hard time tracking the exact commit, as #454 did not mention the commit. In short, a libsasl2 plugin that added the `XOAUTH2` client method needed a bit more than 127 bytes back then.

Bumping the relevant constants to 2049b will likely do the trick this time around. If they're already fixed, apologies for not digging deeper to confirm.

Same comments probably apply to the built-in `OAUTHBEARER` support.

## Have you tried any alternative methods?

- Macros
- Hooks
- Scripts

N/A (I believe)

gahr:
Thanks! Here's a relevant piece of code, too.

https://github.com/neomutt/neomutt/blob/732c046f60e0671d550e1135d418b86897b4b2d5/conn/connaccount.c#L197-L202

PR #3138, opened 2021-11-17
imap: increase max size of oauth2 token

flatcap:
A user reported a token size of 2304.
Ensure we have plenty of space.

oauth2 tokens can be pretty big.
Increase the buffer size to 4096 bytes.

Thanks :heart: to `neo43589367` on IRC for their help debugging.

### DIFF
--- a/conn/connaccount.c
+++ b/conn/connaccount.c
@@ -204,16 +204,16 @@ char *mutt_account_getoauthbearer(struct ConnAccount *cac, bool xoauth2)
     return NULL;
   }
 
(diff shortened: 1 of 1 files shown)


gahr:
Why don't we keep a reasonably small buffer (4500 **is** reasonably small for today's stack sizes) but fall back to allocating on the heap instead of failing if the token doesn't fit?

flatcap:
It's a reasonable idea, but I don't have a account to test this code.
Judging by #2936, the tokens probably aren't going to be much over 2048 bytes.

If you'd like to future-proof the code, go for it.
Otherwise, I'll leave it for now.

gahr (approved):

PR #3144, opened 2021-11-25
Fuzzing: Add CIFuzz

AdamKorcz:
**What does this PR do?**
Adds CIFuzz to neomutt's OSS-fuzz integration. In short, CIFuzz is a service offered by OSS-fuzz to run neomutt's fuzzers during the CI to prevent bugs from being introduced.
In this PR the fuzz time is set to 600 seconds but can be changed.

More about CIFuzz can be found here: https://google.github.io/oss-fuzz/getting-started/continuous-integration/

### DIFF
--- a/.github/workflows/cifuzz.yml
+++ b/.github/workflows/cifuzz.yml
@@ -10,7 +10,7 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        sanitizer: [address, undefined, memory]
+        sanitizer: [address, undefined]
 
     steps:
       - name: Build Fuzzers (${{ matrix.sanitizer }})
--- a/.github/workflows/cifuzz.yml
+++ b/.github/workflows/cifuzz.yml
@@ -1,0 +1,40 @@
+name: CIFuzz
+on:
+  pull_request:
+    branches:
+      - master
+
+jobs:
+  Fuzzing:
+    runs-on: ubuntu-latest
+    strategy:
+      fail-fast: false
+      matrix:
+        sanitizer: [address, undefined, memory]
+
+    steps:
+      - name: Build Fuzzers (${{ matrix.sanitizer }})
+        id: build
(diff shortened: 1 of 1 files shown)


flatcap:
Brilliant!  Thanks @AdamKorcz (fellow Oxford-dweller?)

Years ago, Joseph helped set up Fuzzing and it found loads of problems.
I didn't understand the process then, and after several more hours investigation yesterday, I'm not sure I understand today :-)

I pushed some fixes to the [fuzzer](https://github.com/neomutt/neomutt/tree/fuzzer) branch and rebased it.

To the results...
Your PR's found a memory problem immediately, but I don't understand it.
It points to the `&&` in L1123:
https://github.com/neomutt/neomutt/blob/963acf051fe2e8e451d2e58c19f109a5695b073a/email/parse.c#L1121-L1127

Any idea what it's telling us?

Thanks

DavidKorczynski:
@flatcap I believe the memory sanitizer issue reported is a false positive. Recently OSS-Fuzz had some upgrades that as a side effect caused MSAN to report false positives in many projects. I have now disabled MSAN for neomutt: @AdamKorcz could update this PR to not include "memory" as a sanitizer in the CI action?

flatcap:
> I believe the memory sanitizer issue reported is a false positive.

Thanks, I'd hoped it was :-)

> Recently OSS-Fuzz had some upgrades that as a side effect caused MSAN to report false positives in many projects.

Hmm... it's odd that the error hasn't occurred in the main OSS-Fuzz checking.

> I have now disabled MSAN for neomutt:

Thanks

flatcap:
Thanks @AdamKorcz

DavidKorczynski:
> Hmm... it's odd that the error hasn't occurred in the main OSS-Fuzz checking.
> 
> >

hmm, that is odd.  I built neomutt locally and ran the fuzzer - the issue came up instantly (on the first run of the fuzzer).

flatcap (approved):

PR #3147, opened 2021-11-30
Remove config_init_autocrypt dummy definition

gahr:
This was introduced in 6b893166b9d724e22db0b45803fc4a45ccb2fb17, I can't see why :/


flatcap (approved):
Ah, sorry.
The fuzzing code has been living out-of-tree for years, accumulating cruft.

PR #3153, opened 2021-12-11
build: configure --debug-queue

flatcap:
Enable debugging on mutt/queue.h macros:

- `LIST`, `SLIST`, `TAILQ`, `STAILQ`

This enables two symbols:

- `QUEUE_MACRO_DEBUG_TRACE`
  Store the last 2 places the queue element or head was altered

- `QUEUE_MACRO_DEBUG_TRASH`
  Invalidated pointers will be set to `0xffffffffffffffff`

---

These macros are reliable and very well tested, but they can be hard to use.
Defining these symbols will make it easier to spot mistakes in our code.

### Example of TRACE

In `gdb` examine the contents of a `TAILQ_HEAD`:

```
p *head
1: *head = {tqh_first = 0x850480, tqh_last = 0x850490,
    trace = {lastline = 215, prevline = 74, lastfile = 0x58a3d6 "email/parse.c", prevfile = 0x58a3b3 "email/parameter.c"}}
```
Here we can see the contents of the `TAILQ` **plus** the exact locations of the previous two changes to it.

### Example of TRASH

In `gdb` see the effect of a call to `TAILQ_REMOVE()`:

```
1: *np = {attribute = 0x6ad000 "charset", value = 0x84c640 "iso-8859-1",
    entries = {tqe_next = 0x0, tqe_prev = 0x81ba08,
    trace = {lastline = 215, prevline = 0, lastfile = 0x58a3d6 "email/parse.c", prevfile = 0x0}}}
```

```c
  TAILQ_REMOVE(head, np, entries);
```

`np`s prev/next pointers have become invalidated

```
1: *np = {attribute = 0x6ad000 "charset", value = 0x84c640 "iso-8859-1",
    entries = {tqe_next = 0xffffffffffffffff, tqe_prev = 0xffffffffffffffff,
    trace = {lastline = 71, prevline = 215, lastfile = 0x58ac2b "email/rfc2231.c", prevfile = 0x58a3d6 "email/parse.c"}}}
```

### DIFF
--- a/auto.def
+++ b/auto.def
@@ -127,6 +127,7 @@ options {
   debug-graphviz=0          => "DEBUG: Enable Graphviz dump"
   debug-notify=0            => "DEBUG: Enable Notifications dump"
   debug-parse-test=0        => "DEBUG: Enable 'neomutt -T' for config testing"
+  debug-queue=0             => "DEBUG: Enable TAILQ debugging"
   debug-window=0            => "DEBUG: Enable windows dump"
 # Deprecated stuff
   with-ui:=deprecated       => "Deprecated"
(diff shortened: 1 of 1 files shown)


flatcap:
Meanwhile, I've been digging into the Coverity defects associated with `TAILQ`.
They're definitely bogus and I don't think there's anything we can do about them.

The problem seems limited to `TAILQ_REMOVE()`.
Coverity only notices the case of removing the first element, because it's watching the `HEAD`.

`TAILQ_REMOVE()` uses the double-pointer from the **element**-being-removed to join the previous element (the `HEAD`) to the next.
Unfortunately, this is too many levels of indirection for Coverity.

It doesn't register that the following is affecting `HEAD`
```c
*(elm)->field.tqe_prev = TAILQ_NEXT((elm), field);
```

I tried splitting up the statement.  No change:
```c
__typeof__(elm) *__tmp = (elm)->field.tqe_prev; 
*__tmp = TAILQ_NEXT((elm), field);
```

Also, I looked into using a **model file**, but they only apply to **functions**.

gahr (approved):

PR #3155, opened 2021-12-13
refactor mutt_get_field()

flatcap:
This very repetitive PR refactors 36 functions to use `mutt_buffer_get_field()` rather than `mutt_get_field()`.
See "The Near Future", below.

The code to read some text from the user is too complicated.
It's a mix of functions that take `char []` or `struct Buffer`,
flags that are relevant to some functions but not others,
parameters (arrays of files) that _might_ be needed,
and originally a bunch of macro wrappers.

## The Past

The names have been abbreviated by `mutt_` for clarity.
The blue nodes are macros.

We eliminated the macros a while back, leading to:

## The Present

**Key**:

The numbers in `()`s show the number of callers of each function.

Worryingly, many of the functions pass `CompletionFlags` and an Array for the results of completion.
This means that `mutt_enter_string_full()` has to know a lot about completion methods.

## The Near Future?

The first step is to standardise the string passing, in favour of `struct Buffer`.
This PR is a step to unite `mutt_get_field()` and `mutt_buffer_get_field()`.
When `struct Buffer` is used throughout, the function names can be stripped of `buffer_`.

## The Far Future?

With a bit more refactoring, it may be possible to get to this state:

### DIFF
--- a/alias/alias.c
+++ b/alias/alias.c
@@ -371,10 +371,11 @@ struct AddressList *mutt_get_address(struct Envelope *env, const char **prefix)
  */
 void alias_create(struct AddressList *al, const struct ConfigSubset *sub)
 {
-  char buf[1024];
-  char fixed[1024];
-  char prompt[2048];
-  char tmp[1024] = { 0 };
+  struct Buffer *buf = mutt_buffer_pool_get();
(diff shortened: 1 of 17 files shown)


flatcap:
Images & source: https://flatcap.org/mutt/enter/

PR #3162, opened 2021-12-21
send: Allow an empty To: address prompt

laumann:
* **What does this PR do?**

This is a backport of commit 04a7f1d9585ca21894ec09283becd9425d1c1334
from [0]

[0]: https://github.com/neomutt/upstream-mutt/commit/04a7f1d9

### DIFF
--- a/send/send.c
+++ b/send/send.c
@@ -269,7 +269,7 @@ static int edit_envelope(struct Envelope *en, SendFlags flags, struct ConfigSubs
     const bool c_fast_reply = cs_subset_bool(sub, "fast_reply");
     if (TAILQ_EMPTY(&en->to) || !c_fast_reply)
     {
-      if ((mutt_edit_address(&en->to, _("To: "), true) == -1) || TAILQ_EMPTY(&en->to))
+      if ((mutt_edit_address(&en->to, _("To: "), true) == -1))
         goto done;
     }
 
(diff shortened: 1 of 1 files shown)


laumann:
Maybe "backport" is not the right word...

send/send.c
```diff
@@ -287,6 +287,9 @@ static int edit_envelope(struct Envelope *en, SendFlags flags, struct ConfigSubs
         goto done;
     }
 
+    if (TAILQ_EMPTY(&en->to) && TAILQ_EMPTY(&en->cc) && TAILQ_EMPTY(&en->bcc))
+      goto done;
```
flatcap, line 291: Elsewhere in `send.c`, we already have an error message for this situation.

```suggestion
      {
        mutt_warning(_("No recipients specified"));
        goto done;
      }
```

Please can you downgrade that one too.
https://github.com/neomutt/neomutt/blob/b77af02186a3c800ca04ddbee9fccf9188877dee/send/send.c#L2772

The use of errors vs warnings isn't consistent yet, but my rule-of-thumb is:
- **Warning**: If we can explain to the user how to avoid the issue next time
- **Error**: Something went wrong that the user couldn't avoid happening

flatcap (changes requested):
Thanks!  That looks good and works correctly.

For upstream commits, I like to have special format for the commit message:

- The **original message** (tweaked to match the new config names)
- **Upstream-commit**: a handy link to the upstream commit
- **Co-authored-by**: to give credit to the original author

```
If the user has $ask_cc or $ask_bcc configured, they may want to put
addresses in those without filling in the To field.

Instead check to make sure at least one of the fields is filled in
after all the prompt before aborting.

Upstream-commit: https://gitlab.com/muttmua/mutt/commit/04a7f1d9585ca21894ec09283becd9425d1c1334
    
Co-authored-by: Kevin McCarthy <kevin@8t8.us>
```

send/send.c
laumann: Done.

flatcap (approved):

PR #3171, opened 2021-12-29
Fix out of boundary write with long log lines

stoeckmann:
* **What does this PR do?**

mutt_perror calls can overflow the internal 1024 byte buffer when
errno string representation is added.

Since snprintf returns the amount of bytes it would write regardless
of the limitation of the second argument, the return value has to be
checked for such a truncation before incrementing the start position.

Proof of Concept (compile with -fsanitize=address):
```
echo "set mbox_type=MH" > muttrc
python -c 'print("set postponed=\"/tmp/"+1024*"a"+"\"")' >> muttrc
neomutt -F muttrc -d 1
```
Postpone a newly created mail to see the OOB write.

No issue created.

### DIFF
--- a/mutt/logging.c
+++ b/mutt/logging.c
@@ -411,7 +411,8 @@ int log_disp_queue(time_t stamp, const char *file, int line,
 
   if (level == LL_PERROR)
   {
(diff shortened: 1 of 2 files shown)


flatcap (approved):
Nice catch!
Thanks for the fix.

PR #3172, opened 2021-12-29
Improve log_file_running reliability

stoeckmann:
* **What does this PR do?**

Casting LogFileFP (FILE *) to bool truncates the pointer to one byte.
If LogFileFP is aligned properly, then log_file_running returns false
even though LogFileFP itself is not NULL.

Explicitly check for NULL to make log_file_running more reliable.

No issue created.

### DIFF
--- a/mutt/logging.c
+++ b/mutt/logging.c
@@ -229,7 +229,7 @@ void log_file_set_version(const char *version)
  */
 bool log_file_running(void)
 {
-  return LogFileFP;
+  return LogFileFP != NULL;
 }
 
 /**


flatcap:
Hmm... really?

The minimum build requirement for NeoMutt is a C99 compiler.
(we check for this in `./configure`)

We use `<stdbool.h>` which defines `bool` as `_Bool`.
Both the C99 and C11 specs say:

> 6.3.1.2 Boolean type
> When any scalar value is converted to _Bool, the result is 0 if the value compares equal to 0; otherwise, the result is 1.

stoeckmann:
Oh right, thanks for pointing out that c99 is actually enforced. In this case the code is safe and this pull request effectively a no-op.

flatcap:
No worries.
If you discover anything else that looks dodgy, don't hesitate to raise another issue.
Thanks.

PR #3173, opened 2021-12-30
Improve composition of multipart messages

dcpurton:
## Background

The present support for creating multipart emails is basic and suffers from functionality and usability problems. For example,

- Grouped attachments display as a single attachment which cannot be modified or ungrouped.
- Grouped attachments are appended to the list of attachments and can not be made the body of the email if other attachments have been added prior to grouping. See issue #2387.
- Grouping is lost if sending is aborted for any reason (e.g., crypt function fails or failed attachment check) or an email is postponed.

## Goals for this branch

The goal for this branch is to provide better support for both composing multipart/alternative and multipart/multilingual emails.

### Tasks:

- [x] `<group-attachment>` should place multipart/alternative group at position of first tagged attachment rather than end of attachment list.
- [x] Multipart/alternative attachments should display with correct tree structure.
- [x] Attachments (grouped and otherwise) should be able to be moved around in a natural way.
- [x] Support `<edit-message>` when the body of the email is a multipart/alternative attachment.
- [x] Support ungrouping of multipart attachments.
- [x] Preserve grouping when postponing a message.
- [x] Provide analogous support for multipart/multilingual attachments.
- [x] Amend documentation.
- [x] Support nested groups.
- [x] Create tests.

### DIFF
--- a/attach/recvattach.c
+++ b/attach/recvattach.c
@@ -952,6 +952,16 @@ int mutt_attach_display_loop(struct ConfigSubset *sub, struct Menu *menu, int op
       case OP_VIEW_ATTACH:
       {
         struct AttachPtr *cur_att = current_attachment(actx, menu);
+        if (!cur_att->fp)
+        {
+          if (cur_att->body->type == TYPE_MULTIPART)
+          {
+            struct Body *b = cur_att->body->parts;
+            while (b->parts)
+              b = b->parts;
+            cur_att = b->aptr;
+          }
+        }
         op = mutt_view_attachment(cur_att->fp, cur_att->body, MUTT_VA_REGULAR,
                                   e, actx, menu->win);
         break;
--- a/compose/compose.c
+++ b/compose/compose.c
@@ -285,21 +285,16 @@ static void gen_attach_list(struct AttachCtx *actx, struct Body *m, int parent_t
 {
   for (; m; m = m->next)
   {
+    struct AttachPtr *ap = mutt_mem_calloc(1, sizeof(struct AttachPtr));
+    mutt_actx_add_attach(actx, ap);
+    ap->body = m;
+    m->aptr = ap;
+    ap->parent_type = parent_type;
+    ap->level = level;
     if ((m->type == TYPE_MULTIPART) && m->parts &&
         (!(WithCrypto & APPLICATION_PGP) || !mutt_is_multipart_encrypted(m)))
     {
-      gen_attach_list(actx, m->parts, m->type, level);
-    }
-    else
-    {
-      struct AttachPtr *ap = mutt_mem_calloc(1, sizeof(struct AttachPtr));
-      mutt_actx_add_attach(actx, ap);
-      ap->body = m;
-      m->aptr = ap;
-      ap->parent_type = parent_type;
-      ap->level = level;
-
-      /* We don't support multipart messages in the compose menu yet */
+      gen_attach_list(actx, m->parts, m->type, level + 1);
     }
   }
 }
--- a/compose/functions.c
+++ b/compose/functions.c
@@ -405,6 +405,87 @@ static void compose_attach_swap(struct Body *msg, struct AttachPtr **idx, short 
   idx[first + 1]->num = i;
 }
 
+/**
+ * group_attachments - Group tagged attachments into a multipart group
+ * @param  shared        Shared compose data
+ * @param  multipart_tag String format for group description
+ * @param  subtype       MIME subtype
+ * @retval IR_SUCCESS    Success
+ * @retval IR_ERROR      Failure
+ */
+static int group_attachments(struct ComposeSharedData *shared,
+                             const char *multipart_tag, char *subtype)
+{
+  struct Body *group = mutt_body_new();
+  group->type = TYPE_MULTIPART;
+  group->subtype = mutt_str_dup(subtype);
+  group->disposition = DISP_INLINE;
+
+  struct Body *alts = NULL;
+  /* group tagged message into a multipart group */
+  struct Body *bptr = shared->email->body;
+  for (int i = 0; bptr;)
+  {
+    if (bptr->tagged)
+    {
+      bptr->tagged = false;
+      bptr->disposition = DISP_INLINE;
+
+      /* for first match, set group desc according to match */
+      if (!group->description)
+      {
+        char *p = bptr->description ? bptr->description : bptr->filename;
+        if (p)
+        {
+          group->description = mutt_mem_calloc(1, strlen(p) + strlen(multipart_tag) + 1);
+          sprintf(group->description, multipart_tag, p);
+        }
+      }
+
+      // append bptr to the alts list, and remove from the shared->email->body list
+      if (alts)
+      {
+        alts->next = bptr;
+        bptr = bptr->next;
+        alts = alts->next;
+        alts->next = NULL;
+      }
+      else
+      {
+        group->parts = bptr;
+        alts = bptr;
+        bptr = bptr->next;
+        alts->next = NULL;
+      }
+
+      for (int j = i; j < shared->adata->actx->idxlen - 1; j++)
+      {
+        shared->adata->actx->idx[j] = shared->adata->actx->idx[j + 1];
+        shared->adata->actx->idx[j + 1] = NULL; /* for debug reason */
+      }
+      shared->adata->actx->idxlen--;
+    }
+    else
+    {
+      bptr = bptr->next;
+      i++;
+    }
+  }
+
+  group->next = NULL;
+  mutt_generate_boundary(&group->parameter);
+
+  /* if no group desc yet, make one up */
+  if (!group->description)
+    group->description = mutt_str_dup("unknown multipart group");
+
+  struct AttachPtr *gptr = mutt_mem_calloc(1, sizeof(struct AttachPtr));
+  gptr->body = group;
+  update_idx(shared->adata->menu, shared->adata->actx, gptr);
+  menu_queue_redraw(shared->adata->menu, MENU_REDRAW_INDEX);
+  return IR_SUCCESS;
+}
+
 // -----------------------------------------------------------------------------
 
 /**
@@ -1029,74 +1110,7 @@ static int op_compose_group_alts(struct ComposeSharedData *shared, int op)
     return IR_ERROR;
   }
 
-  struct Body *group = mutt_body_new();
-  group->type = TYPE_MULTIPART;
-  group->subtype = mutt_str_dup("alternative");
-  group->disposition = DISP_INLINE;
-
-  struct Body *alts = NULL;
-  /* group tagged message into a multipart/alternative */
-  struct Body *bptr = shared->email->body;
-  for (int i = 0; bptr;)
-  {
-    if (bptr->tagged)
-    {
-      bptr->tagged = false;
-      bptr->disposition = DISP_INLINE;
-
-      /* for first match, set group desc according to match */
-      if (!group->description)
-      {
-        char *p = bptr->description ? bptr->description : bptr->filename;
-        if (p)
-        {
-          group->description = mutt_mem_calloc(1, strlen(p) + strlen(ALTS_TAG) + 1);
-          sprintf(group->description, ALTS_TAG, p);
-        }
-      }
-
-      // append bptr to the alts list, and remove from the shared->email->body list
-      if (alts)
-      {
-        alts->next = bptr;
-        bptr = bptr->next;
-        alts = alts->next;
-        alts->next = NULL;
-      }
-      else
-      {
-        group->parts = bptr;
-        alts = bptr;
-        bptr = bptr->next;
-        alts->next = NULL;
-      }
-
-      for (int j = i; j < shared->adata->actx->idxlen - 1; j++)
-      {
-        shared->adata->actx->idx[j] = shared->adata->actx->idx[j + 1];
-        shared->adata->actx->idx[j + 1] = NULL; /* for debug reason */
-      }
-      shared->adata->actx->idxlen--;
-    }
-    else
-    {
-      bptr = bptr->next;
-      i++;
-    }
-  }
-
-  group->next = NULL;
-  mutt_generate_boundary(&group->parameter);
-
-  /* if no group desc yet, make one up */
-  if (!group->description)
-    group->description = mutt_str_dup("unknown alternative group");
-
-  struct AttachPtr *gptr = mutt_mem_calloc(1, sizeof(struct AttachPtr));
-  gptr->body = group;
-  update_idx(shared->adata->menu, shared->adata->actx, gptr);
-  menu_queue_redraw(shared->adata->menu, MENU_REDRAW_INDEX);
-  return IR_SUCCESS;
+  return group_attachments(shared, ALTS_TAG, "alternative");
 }
 
 /**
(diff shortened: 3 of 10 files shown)


flatcap:
Wow!  This is looking pretty good!
I haven't come up with a solid strategy for testing, yet, but I've been giving it a good thrashing.

- Creating / breaking groups
- Creating / breaking multilingual
- Moving invididual parts up / down
- Moving groups up / down
- Postponing emails
- Restoring postponed emails

No crashes, no leaks :heart:

But, I've noticed a few oddities...

### Descriptions

The default descriptions are a bit wordy:

- Alternatives for "/home/mutt/work/compose/command_parse.c"
- Multilingual part for "/home/mutt/work/compose/editmsg.c"

How about using `mutt_path_basename()` on the filenames?

I also noticed that after postponing/restoring an email, then
group/lingual creates groups with `<no description>`

### Fundamental

This may be something for another PR, another day.

NeoMutt won't allow you to reorder the "fundamental" part of the email,
but it will allow you to delete it, making whatever's next the main part of the email (even if it's a group).
I'm not sure what would constitute sensible behaviour.

flatcap:
I've also been thinking about automated testing.
It's a no-go for now, but I can see a route to getting it working.

The new compose functions are a good step forward, but they still contain too
many dependencies, e.g. Windows / Menus.

The ideal situation would be to create separate "work" functions. e.g.

- `op_compose_group_alts()`
  - Perform some pre-requisite checks
  - Get input from the user, if necessary
  - Create a list of 'tagged' attachments
  - Call `work_compose_group_alts()`
  - Notify user, if necessary

- `work_compose_group_alts()`
  - Takes an `AttachCtx` and a list of attachments
  - Performs the work
  - Returns an error code, if necessary
  - Few dependencies
  - No interaction with the user

dcpurton:
Thanks for looking over it!

Hardly surprising there are quirks. I'll look at the description stuff.

Also check `<resend-message>`. This uses the same code as recalling postponed messages, but is potentially going to have more complex incoming messages.

Changing the fundamental part behaviour is probably outside the scope of this PR, but you're right that it isn't sensible to allow the fundamental part to be deleted if you can't move it.

It's not that hard to change things so you can move the fundamental part around, especially with my modified `compose_attach_swap` function.

My feeling is there is quite a few existing bugs in the existing code.

In fact, I just found one!

`op_compose_group_alts` checks to make sure there are at least two attachments tagged using `shared->adata->menu->tagged` (this pre-dates my changes). But this test is unreliable. If a tagged attachment is deleted, the menu->tagged isn't decremented.

I'll go through and see if I can find all these.

dcpurton:
> I also noticed that after postponing/restoring an email, then group/lingual creates groups with `<no description>`

@flatcap, I'm not exactly sure of the steps you took to get this behaviour, but I've slightly changed the description logic. See if it still does this.

I've also switched to using `mutt_path_basename()` as you suggested.

I've been thinking about your fundamental part comments. Perhaps it does make sense to alter this in this PR. Given it's possible to reorder parts of a group, the fundamental part could also be changed by making a group, reordering the parts, then breaking the group. Given you can also currently change the fundamental part by deleting it, it's probably just best to remove the (in my view arbitrary) restriction when moving attachments around. I'll put together a commit that allows this. It can always be left out of the final PR if you decide against it. 

And regarding you test comments and creating a `work_compose_group_alts()` function, is this something you envisage for this PR or for doing later?

> No crashes, no leaks :heart:

That's good, but I did manage to make it crash. It was while working on a recalled message and ungrouping and regrouping things. But I don't know the exact steps and haven't managed to reproduce it.

dcpurton:
Ugh.

It makes a right mess of `<resend-message>` for something like this:

```
q:Exit  s:Save  |:Pipe  p:Print  ?:Help
  I     1 <no description>                                              [multipart/alternative, 7bit, 29K]
  I     2 ├─><no description>                                            [text/plain, base64, utf-8, 6.8K]
  I     3 └─><no description>                                               [multipart/related, 7bit, 22K]
  I     4   ├─><no description>                                           [text/html, base64, utf-8, 6.8K]
  I     5   ├─><no description>                                                  [image/jpeg, base64, 10K]
  I     6   └─><no description>                                                 [image/jpeg, base64, 4.7K]
  I     7 <no description>                                                   [multipart/mixed, 7bit, 409K]
  A     8 └─>attachment.pdf                                       [application/octet-stream, base64, 409K]
```

Though to be fair the current code just presents the two top level attachments with all MIME parts inline, which is not terribly useful.

dcpurton:
OK! I added support for nesting groups.

I wasn't going to originally do this because it increases the chance of mistakes, but given that `<resend-message>` can receive arbitrary content, it needed to be done.

I *think* it's all OK, but you should try hard to break it.

With the addition of nested groups, some interesting new possibilities open up. You can create a multilingual group with alternative text and HTML parts (if any other MUA could handle it). But more interestingly, it's now very easy to add support for sending `multipart/related` emails and support inline images in the HTML part. You'd also need to add the ability to set the *Content-ID:* for an attachment, but that's easy. This means that if you felt the need to put a corporate logo in the the signature of the HTML part, this is doable with relatively little pain. Not for this PR, but once any issues are ironed out here, could be an interesting project.

dcpurton:
I also rebased to `master`. Hope that's OK :grimacing:

flatcap:
> I added support for nesting groups.

fantastic!

> given that `<resend-message>` can receive arbitrary content

ah, I see

> I _think_ it's all OK, but you should try hard to break it.

will do :-)

> I also rebased to `master`. Hope that's OK

Absolutely, any time.

flatcap:
I can crash NeoMutt with the following steps:

**test.rc**

```vim
unset record
set postpone = no
set editor="shuf -n 100 /usr/share/dict/words | fmt >> %s"

macro index   <F1> "<mail>john.doe@example.com<enter>test<enter>"
macro compose <F2> "aalternates.c<enter>acommand_parse.c<enter>acommands.c<enter>acomplete.c<enter>aconststrings.c<enter>acontext.c<enter>acopy.c<enter>adlg_postpone.c<enter>aeditmsg.c<enter>"
macro compose <F3> "aenriched.c<enter>aflags.c<enter>afunctions.c<enter>agit_ver.c<enter>ahandler.c<enter>ahdrline.c<enter>ahelp.c<enter>ahook.c<enter>aicommands.c<enter>ainit.c<enter>"
macro compose <F4> "3<enter><tag-entry><tag-entry><tag-entry><group-alternatives>"
macro compose <F5> "9<enter><tag-entry><tag-entry><tag-entry><group-alternatives>"
macro compose <F6> "3<enter><tag-entry>9<enter><tag-entry><group-alternatives>"

# vim: syn=neomuttrc
```

Run: `./neomutt -n -F test.rc`

Run macros <kbd>F1</kbd> to <kbd>F6</kbd> to compose, attach files and nest two groups.
Then <kbd>P</kbd> to `<postpone>`

---

ASAN reports: **heap-use-after-free**

```
ALLOC
main.c:1379:              main
index/dlg_index.c:1341:   mutt_index_menu
index/functions.c:3207:   index_function_dispatcher
index/functions.c:735:    op_mail
send/send.c:2810:         mutt_send_message
compose/compose.c:443:    mutt_compose_menu
compose/functions.c:2403: compose_function_dispatcher
compose/functions.c:1382: op_compose_group_alts
compose/functions.c:606:  compose_group_attachments
email/body.c:45:          mutt_body_new

FREE
main.c:1379:              main
index/dlg_index.c:1341:   mutt_index_menu
index/functions.c:3207:   index_function_dispatcher
index/functions.c:735:    op_mail
send/send.c:3046:         mutt_send_message
email/email.c:59:         email_free
email/body.c:95:          mutt_body_free
email/body.c:95:          mutt_body_free
email/body.c:96:          mutt_body_free

READ
main.c:1379:              main
index/dlg_index.c:1341:   mutt_index_menu
index/functions.c:3207:   index_function_dispatcher
index/functions.c:735:    op_mail
send/send.c:3046:         mutt_send_message
email/email.c:59:         email_free
email/body.c:95:          mutt_body_free
email/body.c:67:          mutt_body_free
```

flatcap:
ASAN Leak of three objects.

Using `test.rc` from above:

Run: `./neomutt -n -F test.rc`

Run macros <kbd>F1</kbd>, <kbd>F2</kbd>, <kbd>F4</kbd> to compose, attach files and nest one group.
Then `#` to `<ungroup-attachment>` and quit compose and NeoMutt.

----

Leaked objects were allocated, here:

```
main.c:1379:              main
index/dlg_index.c:1341:   mutt_index_menu
index/functions.c:3207:   index_function_dispatcher
index/functions.c:735:    op_mail
send/send.c:2810:         mutt_send_message
compose/compose.c:443:    mutt_compose_menu
compose/functions.c:2403: compose_function_dispatcher
compose/functions.c:1382: op_compose_group_alts
compose/functions.c:608:  compose_group_attachments
mutt/string.c:186:        mutt_str_dup

main.c:1379:              main
index/dlg_index.c:1341:   mutt_index_menu
index/functions.c:3207:   index_function_dispatcher
index/functions.c:735:    op_mail
send/send.c:2810:         mutt_send_message
compose/compose.c:443:    mutt_compose_menu
compose/functions.c:2403: compose_function_dispatcher
compose/functions.c:1382: op_compose_group_alts
compose/functions.c:702:  compose_group_attachments
send/multipart.c:92:      mutt_generate_boundary
email/parameter.c:131:    mutt_param_set
email/parameter.c:41:     mutt_param_new

main.c:1379:              main
index/dlg_index.c:1341:   mutt_index_menu
index/functions.c:3207:   index_function_dispatcher
index/functions.c:735:    op_mail
send/send.c:2810:         mutt_send_message
compose/compose.c:443:    mutt_compose_menu
compose/functions.c:2403: compose_function_dispatcher
compose/functions.c:1382: op_compose_group_alts
compose/functions.c:718:  compose_group_attachments
```

dcpurton:
Ooo! I should have thought about testing this way a while ago…

This will help. I'll go through every combination I can think of and put in a test config file.

dcpurton:
Oog. There are a *lot* of combinations, aren't there?

flatcap:
I've fixed the Graphviz problem with ASAN -- I should have investigated sooner.
I merged the fixes into master and I've rebased the PR over them.

flatcap:
I've found a case that causes an incorrect layout.

**Config**: https://github.com/neomutt/test-compose/blob/main/bad1/bad1.rc

Grouping some files seems to work correctly, but adding one extra file makes it go wrong.

### Good

<kbd>F1</kbd>, <kbd>F2</kbd>, <kbd>F4</kbd>, <kbd>F5</kbd>, <kbd>l</kbd>, <kbd>F6</kbd>, <kbd>l</kbd>
(**skipping** <kbd>F3</kbd>, using <kbd>l</kbd> to dump the layout)

gives me:



[source](https://raw.githubusercontent.com/neomutt/test-compose/main/bad1/01%3A08%3A23-email.svg)



[source](https://raw.githubusercontent.com/neomutt/test-compose/main/bad1/01%3A08%3A26-email.svg)

### Bad

<kbd>F1</kbd>, <kbd>F2</kbd>, <kbd>F3</kbd>, <kbd>F4</kbd>, <kbd>F5</kbd>, <kbd>l</kbd>, <kbd>F6</kbd>, <kbd>l</kbd>
(**including** <kbd>F3</kbd>, using <kbd>l</kbd> to dump the layout)

gives me:



[source](https://raw.githubusercontent.com/neomutt/test-compose/main/bad1/01%3A08%3A38-email.svg)



[source](https://raw.githubusercontent.com/neomutt/test-compose/main/bad1/01%3A08%3A40-email.svg)

dcpurton:
@flatcap with some inspiration from your tests and your one line body dump, it's possible to do something like this: https://github.com/neomutt/test-compose/tree/main/dcpurton

Some things to think about:

- I locally removed the `dump_graphviz_email()` from `<print-message>`, since you probably don't want a graphviz file for every test when running them automatically.
- I also changed the test for an alternative part to test for `body->parts` rather than `(mutt_str_startswith(b->description, "Alternatives")`, which is not reliable.

flatcap:
Rebased and lightly squashed

dcpurton:
It would be easier to squash if I had pulled the common code out of the two grouping functions *first*.

flatcap:
> It would be easier to squash if I had pulled the common code out of the two grouping functions _first_.

We can still do that.

There's a handful of small commits that fix the old code.  We could move them first.
For my own refactoring, I tend to do any renames or code-moving next -- things that don't make any **functional** changes.
(My aim is to make the **functional** changes as small as possible)

Then we can apply the functional refactoring on top.

dcpurton:
What's the easiest way to do that? Would it be best just to make a temporary branch and patch all the changes in a sensible order? That way I can check if the final diff is the same for the two branches. (I don't mind doing this.)

Or is there a better way? I'm at the limits of my git knowledge.

Would pulling the common code out of the grouping functions early mean most committee have to be manually updated anyway?

flatcap:
> Would it be best just to make a temporary branch and patch all the changes in a sensible order?

Yes, sort of.

The commits don't quite represent the perfect solution, so we can't just `git cherry-pick` them into the right order
(or my favourite operation, `git rebase --interactive master`)

Now I come to explain my workflow, it seems quite convoluted :-)

First, make sure everything in the source branch is how you want it.
Are you happy with:
- the naming of variables?
- the naming of functions?
- their position of the functions within the file?

If so:

```sh
# create a new branch
git switch -c comp2 master
# copy over the changes
git restore -s devel/compose-multipart .
```

Now you can 'stage' changes that you want to add to the commit, using either
`git add --patch` or the `vim-fugitive` command `:Gdiffsplit`

If you want to test the build of the staged changes, then you can clear out the other changes:
`git restore .` or put them back `git restore -s devel/compose-multipart .`

When you're happy, `git commit` and repeat.

> That way I can check if the final diff is the same for the two branches.

Exactly

> Would pulling the common code out of the grouping functions early mean most commits have to be manually updated anyway?

Here's magic of my workflow.
If your commits do:
- Refactor code
- Move code into separate function

I would:
- Clear out the changes
- Create the new function
- Move the old code into it
- Commit _that_

Then, 'restore' the new code on top: `git restore -s devel/compose-multipart .`
That gets us to the correct place without having to do any awkward merging.

> (I don't mind doing this.)

Great.  Have a play around, I find it fun :-)

flatcap:
I've found a couple of bugs.

### Simple

**Steps**:
- Attach one file
- Group the two parts (fundamental and attachment)
- Delete the grouped attachment

**Result** - No attachments at all!

I've pushed a possible fix as 4773a6647.
Feel free to delete it and do something better :-)

### Hard

A certain combination of grouping leads to mislinked attachments.
See the 2<sup>nd</sup> line, 4<sup>th</sup> Body.
It is linked from the left and above.



**Steps**:
- Open compose
- `source bad2.rc`

[bad.rc](https://github.com/neomutt/test-compose/blob/main/bad2/bad2.rc) in the [test-compose](https://github.com/neomutt/test-compose/blob/main/bad2/bad2.rc) repo

dcpurton:
I think I might do what I should have done a while ago: Write a `find_body_previous()` function. That kind of logic is used everywhere. It needs recursion and there's too many hacks throughout my code to try and do it in one loop.

flatcap:
> Write a `find_body_previous()` function

Sounds like a good idea.

I've just updated the [bad2.rc](https://github.com/neomutt/test-compose/tree/main/bad2#readme) test case to make it simpler.

flatcap:
I've rebased my automation code over your latest changes -- [[flatcap/compose]](https://github.com/neomutt/neomutt/tree/flatcap/compose)
I've made all the hard-coded stuff configurable.

### Instructions

**Config**:
```vim
# Minimise all the delays
set sleep_time = 0
set imap_keepalive = 0
set timeout = 1

# Number of complete tests (0 means run forever)
set compose_autorun = 2

# Macro to restart the compose dialog
macro index <F1> "<mail>john.doe@example.com<enter>test<enter>"
```

**Build**:
```sh
./configure --debug-compose
```

**Run**:
- Run `neomutt` as normal
- <kbd>F1</kbd> to start the Compose dialog
- <kbd>g</kbd> to Go! (function `<automate>`)

NeoMutt will now run ~100 random operations, then quit compose and use <kbd>F1</kbd> to repeat the test `$compose_autorun` times.
At the end of the tests, NeoMutt will close complete.

**Logging**:

All actions are logged with the label `ACTION:`
With a small amount of editing, they can be turned into test scripts.

Note: The <kbd>F1</kbd> macro is necessary until I figure out why pushing `<mail>` doesn't work.

flatcap:
It's just survived a 1000 itererations of 100 random functions! :tada:
Well done!

dcpurton:
@flatcap , I've finished putting the commits in order at https://github.com/neomutt/neomutt/tree/devel/compose-multipart-final.

I've tried to keep changes to a minimum in each commit and make sure that it doesn't break along the way.

There's a lot of memory leaks in the the old code. I started trying to fix these, but gave up. So the leaks don't go away until everything is in place.

Let me know what happens next!

flatcap (approved):
Fantastic job!
Thanks :heart:

PR #3178, opened 2021-12-31
Fix default display when preferred_languages unset (fixes #3174)

### LINKED ISSUE #3174: No multipart/multilingual part displayed if $preferred_languages is not set
The Neomutt manual [states](https://neomutt.org/guide/mimesupport.html#6-1-%C2%A0reading-multipart-multilingual-emails):

> If `$preferred_languages` is not set, it default to None, and the first part of the received multipart/multilingual email will be displayed.

However, in reality if `$preferred_languages` is not set, no attachment is displayed.

If `$preferred_languages` is set to `"en,fr"` or `"fr,en"`, the output is as expected.

## Example email

```
Date: Thu, 30 Dec 2021 16:25:37 +1030
From: David Purton <dcpurton@marshwiggle.net>
To: David Purton <dcpurton@marshwiggle.net>
Subject: multipart/multilingual test
MIME-Version: 1.0
Content-Type: multipart/multilingual; boundary="3modmaki7xn74i2n"
Content-Disposition: inline


--3modmaki7xn74i2n
Content-Language: en
Content-Disposition: inline

English


--3modmaki7xn74i2n
Content-Type: text/plain; charset=utf-8
Content-Language: fr
Content-Disposition: inline
Content-Transfer-Encoding: 8bit

Français


--3modmaki7xn74i2n--
```

dcpurton:
This PR ensures the `first_part` Body pointer of a `multipart/multilingual` email is set even if `$preferred_languages` is not set.

### DIFF
--- a/handler.c
+++ b/handler.c
@@ -1145,6 +1145,16 @@ static int multilingual_handler(struct Body *a, struct State *s)
   struct Body *zxx_part = NULL;
   struct ListNode *np = NULL;
 
+  while (b)
+  {
+    if (mutt_can_decode(b))
+    {
+      first_part = b;
+      break;
+    }
+    b = b->next;
+  }
+
   const struct Slist *c_preferred_languages =
       cs_subset_slist(NeoMutt->sub, "preferred_languages");
   if (c_preferred_languages)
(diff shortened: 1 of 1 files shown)


gahr:
Nice! What if it can't be decoded? Shouldn't we search for the first one we can display?

dcpurton:
> Nice! What if it can't be decoded? Shouldn't we search for the first one we can display?

Yeah, that probably makes sense and is easy to do. I'll update.

It's probably not worth losing too much sleep on `multipart/multilingual` emails though. I'm not aware of any other mail client capable of actually displaying them correctly, let alone producing such an email.

gahr (approved):

Issue #3180, opened 2022-01-01
Leak detected by ASAN when changing to a folder


scottkosty:
From current master, I get the following leak when I open mutt and `c<TAB>` and choose a folder:

```
=================================================================
==2117269==ERROR: LeakSanitizer: detected memory leaks

Direct leak of 1088 byte(s) in 17 object(s) allocated from:
    #0 0x7fde1bbfba37 in __interceptor_calloc ../../../../src/libsanitizer/asan/asan_malloc_linux.cpp:154
    #1 0x556df4b818a6 in mutt_mem_calloc mutt/memory.c:55
    #2 0x556df48e6252 in mutt_sort_threads /home/scott/IndependentInstalls/neomutt/mutt_thread.c:1102
    #3 0x556df490a823 in mutt_sort_headers /home/scott/IndependentInstalls/neomutt/sort.c:403
    #4 0x556df487a2e5 in ctx_update /home/scott/IndependentInstalls/neomutt/context.c:214
    #5 0x556df4879318 in ctx_new /home/scott/IndependentInstalls/neomutt/context.c:96
    #6 0x556df49146b5 in change_folder_mailbox index/dlg_index.c:657
    #7 0x556df4914e9f in change_folder_string index/dlg_index.c:758
    #8 0x556df491fee7 in op_main_change_folder index/functions.c:867
    #9 0x556df492f5a6 in index_function_dispatcher index/functions.c:3207
    #10 0x556df4919700 in mutt_index_menu index/dlg_index.c:1341
    #11 0x556df48cb094 in main /home/scott/IndependentInstalls/neomutt/main.c:1379
    #12 0x7fde1b3bbfcf in __libc_start_call_main ../sysdeps/nptl/libc_start_call_main.h:58

SUMMARY: AddressSanitizer: 1088 byte(s) leaked in 17 allocation(s).
```

## NeoMutt Version

```
NeoMutt 20211029-155-ec2bfc
Copyright (C) 1996-2020 Michael R. Elkins and others.
NeoMutt comes with ABSOLUTELY NO WARRANTY; for details type 'neomutt -vv'.
NeoMutt is free software, and you are welcome to redistribute it
under certain conditions; type 'neomutt -vv' for details.

System: Linux 5.15.5-051505-generic (x86_64)
ncurses: ncurses 6.2.20201114 (compiled with 6.2.20201114)
libidn: 1.33 (compiled with 1.33)
[...]

Issue #3108, opened 2021-10-29
Memory leak restoring and postponing postponed message


Austin-Ray:
Appears on latest master commit.

## Expected Behaviour

No memory leak

## Actual Behaviour

ASAN outputs:
```
=================================================================
==765551==ERROR: LeakSanitizer: detected memory leaks

Indirect leak of 128 byte(s) in 2 object(s) allocated from:
    #0 0x7f7ac0bb8957 in calloc (/usr/lib/gcc/x86_64-pc-linux-gnu/11.2.0/libasan.so.6+0xb6957)
    #1 0x5619b73b5665 in mutt_mem_calloc mutt/memory.c:55
    #2 0x5619b713ff99 in mutt_sort_threads /home/aray/builds/neomutt/neomutt/mutt_thread.c:1102
    #3 0x5619b7163ea1 in mutt_sort_headers /home/aray/builds/neomutt/neomutt/sort.c:403
    #4 0x5619b70ce58b in ctx_update /home/aray/builds/neomutt/neomutt/context.c:214
    #5 0x5619b70cd5c6 in ctx_new /home/aray/builds/neomutt/neomutt/context.c:96
    #6 0x5619b7154a47 in mutt_get_postponed /home/aray/builds/neomutt/neomutt/postpone.c:687
    #7 0x5619b7321e3f in mutt_send_message send/send.c:2190
    #8 0x5619b7178637 in op_mail index/functions.c:702
    #9 0x5619b7188745 in index_function_dispatcher index/functions.c:3017
    #10 0x5619b7172c49 in mutt_index_menu index/dlg_index.c:1341
    #11 0x5619b7124ffa in main /home/aray/builds/neomutt/neomutt/main.c:1377
    #12 0x7f7ac026e7fc in __libc_start_main (/lib64/libc.so.6+0x237fc)

SUMMARY: AddressSanitizer: 128 byte(s) leaked in 2 allocation(s).
```

## Steps to Reproduce

1. Have postponed email
2. Restore postpone email
3. Postpone email again
4. Exit NeoMutt

## How often does this happen?

Always

## NeoMutt Version

```
NeoMutt 20211022-22-c86df2
Copyright (C) 1996-2020 Michael R. Elkins and others.
NeoMutt comes with ABSOLUTELY NO WARRANTY; for details type 'neomutt -vv'.
NeoMutt is free software, and you are welcome to redistribute it
under certain conditions; type 'neomutt -vv' for details.

[...]

PR #3186, opened 2022-01-04
fix leak of threads

flatcap:
Fixes: #3180

When Emails don't have a Message-ID, we store them in the HashTable with a key of "".
Remove the check that prevented this.


Austin-Ray:
Also fixes #3108

PR #3189, opened 2022-01-10
macos_keychain: introduce a way to store credentials in Keychain

artagnon:
Hi 👋🏽

I would like to get neomutt to work with macOS's Keychain seamlessly. This is the first patch in the effort, and I'd like to get some feedback before going ahead.

Thanks.

### DIFF
--- a/auto.def
+++ b/auto.def
@@ -106,6 +106,8 @@ options {
   with-zlib:path            => "Location of zlib"
   zstd=0                    => "Enable Zstandard header cache compression support"
   with-zstd:path            => "Location of Zstandard"
+# MacOS
+  macoskeychain=1           => "Use MacOS's Keychain for authentication"
 # System
   with-sysroot:path         => "Target system root"
   with-tmpdir:=/tmp         => "location of the tmp directory"
@@ -155,8 +157,8 @@ if {1} {
     debug-notify debug-parse-test debug-queue debug-window doc everything
     fmemopen full-doc fuzzing gdbm gnutls gpgme gss homespool idn idn2
     include-path-in-cflags inotify kyotocabinet lmdb locales-fix lua lz4
-    mixmaster nls notmuch pcre2 pgp pkgconf qdbm rocksdb sasl smime sqlite ssl
-    testing tdb tokyocabinet zlib zstd
+    mixmaster nls notmuch macoskeychain pcre2 pgp pkgconf qdbm rocksdb sasl smime
+    sqlite ssl testing tdb tokyocabinet zlib zstd
   } {
     define want-$opt [opt-bool $opt]
   }
@@ -165,8 +167,8 @@ if {1} {
   # relative --enable-opt to true. This allows "--with-opt=/usr" to be used as
   # a shortcut for "--opt --with-opt=/usr".
   foreach opt {
-    bdb gdbm gnutls gpgme gss homespool idn idn2 kyotocabinet lmdb lua lz4 
-    mixmaster nls notmuch pcre2 qdbm rocksdb sasl sqlite ssl tdb tokyocabinet 
+    bdb gdbm gnutls gpgme gss homespool idn idn2 kyotocabinet lmdb lua lz4
+    mixmaster nls notmuch pcre2 qdbm rocksdb sasl sqlite ssl tdb tokyocabinet
     zlib zstd
   } {
     if {[opt-val with-$opt] ne {}} {
@@ -1194,6 +1196,15 @@ if {[get-define want-gss]} {
     define HAVE_HEIMDAL
   }
   define USE_GSS
+}
+
+###############################################################################
+# Use MacOS's Keychain for authentication
+if {[has-define {} __APPLE__]} {
+  if {[get-define want-macoskeychain]} {
+    define USE_MACOS_KEYCHAIN
+    define-append LDFLAGS -Wl,-framework -Wl,Security
+  }
 }
 
 ###############################################################################
--- a/conn/macos_keychain.c
+++ b/conn/macos_keychain.c
@@ -1,0 +1,228 @@
+/**
+ * @file
+ * Credential management via macOS Keychain
+ *
+ * @authors
+ * Copyright (C) 2022 Ramkumar Ramachandra <r@artagnon.com>
+ *
+ * @copyright
+ * This program is free software: you can redistribute it and/or modify it under
+ * the terms of the GNU General Public License as published by the Free Software
+ * Foundation, either version 2 of the License, or (at your option) any later
+ * version.
+ *
+ * This program is distributed in the hope that it will be useful, but WITHOUT
+ * ANY WARRANTY; without even the implied warranty of MERCHANTABILITY or FITNESS
+ * FOR A PARTICULAR PURPOSE.  See the GNU General Public License for more
+ * details.
+ *
+ * You should have received a copy of the GNU General Public License along with
+ * this program.  If not, see <http://www.gnu.org/licenses/>.
+ */
+
+/**
+ * @page conn_macos_keychain Read/write macOS Keychain credentials
+ *
+ * Called by mutt_account_getuser() and mutt_user_getpass() as an alternative to
+ * storing/retrieving credentials in the configutation file, or entering them in
+ * the prompt repeatedly. A dialog will automatically pop up requesting permissions
+ * on macOS the very first time.
+ *
+ * Disabled using the configure option '--disable-macoskeychain'.
+ */
+
+#include <Security/Security.h>
+#include <assert.h>
+#include <stdio.h>
+#include <stdlib.h>
+#include <string.h>
+#include "mutt/lib.h"
+#include "connaccount.h"
+#include "mutt_account.h"
+
+/**
+ * struct Credential - A bunch of pointers that act as a lens into ConnAccount,
+ *                     filtering out only the information relevant to keychain.
+ */
+struct Credential
+{
+  SecProtocolType protocol; ///< Connection type, e.g. HTTP
+  UInt16 port;              ///< Port to connect to
+  const char *host;         ///< Server to login to
+  const char *user;         ///< Username
+  const char *pass;         ///< Password
+};
+
+#define KEYCHAIN_ITEM(x) mutt_str_len(x), x
+
+/**
+ * find_username_in_item - Find the username in a given SecKeychainItem
+ * @param item A reference to the security keychain item
+ * @param user A pre-allocated buffer into which to write the username
+ * @param buflen The length of the pre-allocated user buffer
+ * @retval 0 on success, a specialized error-code on failure
+ */
+static int find_username_in_item(SecKeychainItemRef item, char *user, unsigned buflen)
+{
+  SecKeychainAttributeList list = { 0 };
+  SecKeychainAttribute attr = { 0 };
+
+  list.count = 1;
+  list.attr = &attr;
+  attr.tag = kSecAccountItemAttr;
+
+  int ret = SecKeychainItemCopyContent(item, NULL, &list, NULL, NULL);
+  if (ret != 0)
+    return ret;
+
+  assert(attr.length + 1 <= buflen &&
+         "Did you modify the keychain item by hand?");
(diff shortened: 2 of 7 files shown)


flatcap:
Welcome to NeoMutt, @artagnon!
Thanks for the PR.

That sounds like a useful thing to do.
I'll have a look through the code and try to figure out the best place to integrate.

Note: I've turned this into a draft PR, to reduce the number of automatic builds while we develop.

Thanks

flatcap:
A quick overview of the objects in NeoMutt.
Each `struct` owns the one below:
Many objects have private data shown to the right.

- `NeoMutt`
- `Account`, `ImapAccountData`, `NmAccountData`, etc
- `Mailbox`, `ImapMboxData`, `NmMboxData`, etc
- `Email`, `ImapEmailData`, `NmEmailData`, etc

Some of the `Account`s' private data holds a `Connection` which holds a `ConnAccount`.
`ConnAccount` caches the login credentials.

Each of IMAP, SMTP, NNTP, etc has its own config variables.
To fill the `ConnAccount`, each backend sets its own `get_field()` function.

When a backend, e.g. IMAP, needs credentials to login, it calls:
- `mutt_account_getuser()`
- `mutt_account_getpass()`

These will be the perfect place to introduce your keychain code.

Here's an object diagram (click to enlarge)


(lots more diagrams in the [gfx repo](https://github.com/neomutt/gfx#readme))

flatcap (changes requested):
I've undone the clang-format changes to the header file -- whitespace in headers is helpful for humans :-)
I've added a build option:
```
./configure --osxkeychain
```
that defines `USE_OSX_KEYCHAIN`, adds your linker flags and conditionally builds `conn/osxkeychain.c`

`auto.def` checks for the existence of `__APPLE__`, but I haven't any way to test if that's correct.

All the keychain code in `conn/connaccount.c` needs to be protected by `#ifdef USE_OSX_KEYCHAIN`

Finally, when things are working, all the code in `conn/osxkeychain.c` will need commenting :-)

flatcap:
I've pushed an updated Mac check -- I'm more hopeful this time.

flatcap:
I've pushed a bunch of changes; they're split into four commits.

**Bugs** 5a8c5b5fb

You'll see

**Doxygen** e639b6b48

Doc everything and link them up
We don't run `clang-format` on the headers, so there's no need to turn it off

**Init objects** 4e25bf11d

All pointers need to be initialised, including tricky ones like `SecKeychainItemRef`.
I've also initialised the objects and a `UInt32` where they're being used as out-params for functions.

**Style** 71d412916

The hardest part of running any project -- agreeing on style.
I've split up a few compound assignment/test, e.g.

```diff
-if (ret = function())
+ret = function();
+if (ret)
```

The compiler doesn't care, but it make is a little easier to read (imho), and a lot easier to debug.

The other change is to some functions that return 0 as success
This is more of a personal preference, than a rigid guideline.

```diff
-if (!my_function())
+if (my_function() == 0)
   do_something();
```

When will `do_something()` be run?
The second `if` gives more of a hint that a **successful** outcome will lead to `do_something()` being run.

---

See the [Code Style](https://neomutt.org/dev/code) guide for more tips.

conn/osxkeychain.c
```diff
@@ -0,0 +1,209 @@
...
+  const char *pass;         ///< Password
+};
+
+static struct Credential cred;
+
+#define KEYCHAIN_ITEM(x) (x ? strlen(x) : 0), x
+#define KEYCHAIN_ARGS                                                          \
+  NULL,                                  /* default keychain */                \
+      KEYCHAIN_ITEM(cred.host), 0, NULL, /* account domain */                  \
+      KEYCHAIN_ITEM(cred.user), 0, NULL, cred.port, cred.protocol, kSecAuthenticationTypeDefault
```
flatcap, line 61: These save a bit of space, but macros can make debugging harder.
I'd prefer to see them written out fully.
Note: `mutt_str_len()` does `NULL`-checking for us.

conn/osxkeychain.c
```diff
@@ -0,0 +1,209 @@
...
+  UInt32 len = 0;
+  SecKeychainItemRef item = NULL;
+  int ret = 0;
+
+  ret = SecKeychainFindInternetPassword(KEYCHAIN_ARGS, &len, &buf, &item);
+  if (ret != 0)
+    return ret;
+
+  // Keychain does not send the password with null-terminator
+  mutt_str_copy(pass, buf, len + 1);
```
flatcap, line 105: OK, that's a bug!
A have a fixed-length buffer, `pass`, but we don't check if `len + 1` might overflow.

conn/osxkeychain.c
```diff
@@ -0,0 +1,209 @@
...
+struct Credential
+{
+  SecProtocolType protocol; ///< Connection type, e.g. HTTP
+  UInt16 port;              ///< Port to connect to
+  const char *host;         ///< Server to login to
+  const char *user;         ///< Username
+  const char *pass;         ///< Password
+};
+
+static struct Credential cred;
```
flatcap, line 55: What's the lifespan of `cred`?
What if the user has multiple `Account`s?

I see that is doesn't **own** the objects, `host`, etc...
What happens if the `Account` goes away?

flatcap (changes requested):

artagnon:
Thanks a lot! I’m a bit busy today, but a couple of quick comments on the changes:

1. Is there some issue with #pragma once? Isn’t it cleaner than the #ifdef?
2. Is () versus (void) for function arguments a bug? As far as I know, () is valid C99. Stylistic issue, perhaps?

Warm regards,
Ram

flatcap:
> I’m a bit busy today

No worries

> #pragma once
> Isn’t it cleaner than the #ifdef?

It's certainly cleaner, but not everything supports it.
Some people compile (neo)mutt on very old hardware.

 > Is () versus (void) for function arguments a bug?

Not really, but...

In C++, `function()` means `function(void)` -- it's explicitly mentioned in the spec.
However in C, it means the arguments aren't specified and could be anything.

auto.def
```diff
@@ -106,6 +106,8 @@ options {
   with-zlib:path            => "Location of zlib"
   zstd=0                    => "Enable Zstandard header cache compression support"
   with-zstd:path            => "Location of Zstandard"
+# MacOS
+  osxkeychain=0             => "Use MacOS's Keychain for authentication"
```
gahr, line 110: I think we should use `macos` instead of `osx`, both here and in the file names.

conn/osxkeychain.c
gahr: Let's not do this and have conn_account_to_cred return a Credential object instead.

auto.def
artagnon: Agreed. Will make the change.

artagnon:
Okay, all done with cleaned-up history. Please eyeball the changes and test locally.

flatcap (changes requested):
Just one small fix; the rest looks OK to me (but I can't test it).
You're welcome to squash my commit in

conn/macos_keychain.c
```diff
@@ -0,0 +1,217 @@
...
+  SecKeychainItemFreeContent(NULL, buf);
+  return ret;
+}
+
+/**
+ * add_internet_password - Add a credential to the keychain
+ * @param cred A credential object to insert
+ * @retval 0 on success, a specialized error-code on failure
+ */
+static int add_internet_password(struct Credential cred)
```
gahr, line 123: How about passing it by const* ?

artagnon:
Okay, I've made the change that @flatcap suggested, but not the one that @gahr suggested. It should be merge-ready: it would be nice if @gahr could test it locally.

conn/macos_keychain.h
```diff
@@ -0,0 +1,31 @@
...
+
+#ifndef MUTT_CONN_MACOS_KEYCHAIN_H
+#define MUTT_CONN_MACOS_KEYCHAIN_H
+
+struct ConnAccount;
+
+int mutt_account_read_keychain (struct ConnAccount *account);
+int mutt_account_write_keychain(struct ConnAccount *account);
+
+#endif /* MUTT_CONN_OSX_KEYCHAIN_H */
```
gahr, line 31:
```suggestion
#endif /* MUTT_CONN_MACOS_KEYCHAIN_H */
```

conn/macos_keychain.c
```diff
@@ -0,0 +1,217 @@
...
+  ret = SecKeychainFindInternetPassword(NULL, /* default keychain */
+                                        KEYCHAIN_ITEM(cred.host), 0, NULL, /* account domain */
+                                        KEYCHAIN_ITEM(cred.user), 0, NULL, cred.port,
+                                        cred.protocol, kSecAuthenticationTypeDefault,
+                                        &len, &buf, &item);
+  if (ret != 0)
+    return ret;
+
+  // Keychain does not send the password with null-terminator
+  assert(len + 1 <= sizeof(pass) &&
```
gahr, line 105: I don't understand this assertion. What's the point of sizeof(char*)?

artagnon: It's actually `char[256]` -- see the caller.

conn/macos_keychain.c
```diff
@@ -0,0 +1,217 @@
...
+         "Did you modify the keychain item by hand?");
+  mutt_str_copy(pass, buf, len + 1);
+  if (!cred.user && (ret = find_username_in_item(item, user)))
+  {
+    SecKeychainItemFreeContent(NULL, buf);
+    return ret;
+  }
+
+  SecKeychainItemFreeContent(NULL, buf);
+  return ret;
```
gahr, line 115:
```suggestion
  if (!cred.user)
  {
    ret = find_username_in_item(item, user);
  }

  SecKeychainItemFreeContent(NULL, buf);
  return ret;
```

conn/macos_keychain.c
gahr: It's not, as far as find_internet_password is concerned. You get a pointer to char.

artagnon: Hm, do you suggest changing the signature of `find_internet_password`?

artagnon:
Okay, I've hard-coded 256 in a proposed fixup. Does it look okay?

conn/macos_keychain.c
gahr: I suggest to pass the length as a parameter

gahr: find_internet_password(account->user, sizeof(account->user), account->pass, sizeof(account->pass), cred);

gahr:
Because this doesn't add any additional dependency, I think I'd probably always compile the module in when in a mac, and add a config variable to say whether we want to use the keychain.

gahr:
I tried to test it on my macos but it didn't work. I'm not familiar with keychain, but I do have my mail account set up in Mail, so I guess it should work? When I try to `neomutt -f imaps://user@host`, I see that `user` and `host` are correctly passed to `SecKeychainFindInternetPassword`, but I get back a len of `0` and an empty `buf`. Sorry I can't debug on this machine right now.

conn/macos_keychain.c
```diff
@@ -0,0 +1,223 @@
...
+    case errSecDuplicateItem:
+      mutt_error(_("Duplicate item in keychain"));
+    default:
+      mutt_error(_("Error code from keychain add: %d"), ret);
+  }
+  return -1;
+}
+
+/**
+ * mutt_account_write_keychain - Read user/pass into a paritially-filled
```
gahr, line 202:
```suggestion
 * mutt_account_read_keychain - Read user/pass into a paritially-filled
```

gahr:
You can add 
```
if {[string match "darwin*" [get-define build_os]]} {
  options-add {
  # MacOS
    macoskeychain=1 => "Use MacOS's Keychain for authentication"
  }
}
```
before `options {...}` to limit the `--macoskeychain` config to macos platforms.

flatcap:
Archived to https://github.com/neomutt/neomutt-old/tree/osxkeychain

the-x-at:
Hi, I was looking forward to see this become part of `neomutt`, and so I am a little bit sad that this feature has become retired. @artagnon appears to be busy with other stuff, so I would volunteer to help out here.

PR #3190, opened 2022-01-10
date: humanize timestamps in index

artagnon:
We introduce a new configuration option `humanize_time`, a boolean that
is set to false by default. When set to true, the index will display
timestamps like "Just now", "A few hours ago", and "Earlier today".

Signed-off-by: Ramkumar Ramachandra <r@artagnon.com>

Capture d’écran 2022-01-11 à 08 01 44

### DIFF
--- a/mutt/date.c
+++ b/mutt/date.c
@@ -33,6 +33,7 @@
 #include <stdlib.h>
 #include <sys/time.h>
 #include <time.h>
+#include "mutt/message.h"
 #include "date.h"
 #include "buffer.h"
 #include "logging.h"
@@ -55,6 +56,19 @@ static const char *const Months[] = {
   "Jan", "Feb", "Mar", "Apr", "May", "Jun",
   "Jul", "Aug", "Sep", "Oct", "Nov", "Dec",
 };
+
+/**
+ * Special cases for humanized timestamps
+ */
+static const unsigned ONE_HOUR = 3600;         ///< In terms of seconds
+static const unsigned ONE_DAY = 24 * ONE_HOUR; ///< Again, expressed in seconds
+static const char *const E0h = N_("just now"); ///< Within the hour
+static const char *const E1h = N_("an hour ago"); ///< Between an hour and two hours
+static const char *const E2h = N_("few hours ago"); ///< Between two and four hours
+static const char *const E6h = N_("earlier today"); ///< Over four hours ago, the same day
+static const char *const E1d = N_("yesterday"); ///< Anytime the previous day
+static const char *const ENd = N_("days ago");  ///< N days ago; be precise
+// Anything beyond this range doesn't need to be humanized
 
 /**
  * TimeZones - Lookup table of Time Zones
@@ -695,6 +709,36 @@ size_t mutt_date_localtime_format(char *buf, size_t buflen, const char *format, 
 
   struct tm tm = mutt_date_localtime(t);
   return strftime(buf, buflen, format, &tm);
+}
+
+/**
+ * @param  buf     Buffer to store humanized time string
+ * @param  buflen  Buffer size
+ * @param  t       Time to format
+ * @retval num     Number of bytes added to the buffer, excluding NUL byte
+ */
+size_t mutt_date_humanize(char *buf, size_t buflen, time_t t)
+{
+  if (!buf)
+    return 0;
+
+  double diff = difftime(mutt_date_epoch(), t);
+  struct tm now = mutt_date_localtime(mutt_date_epoch());
+  struct tm tm = mutt_date_localtime(t);
+  if (diff < ONE_HOUR)
+    return sprintf(buf, "%13s", _(E0h));
+  else if (diff < 2 * ONE_HOUR)
+    return sprintf(buf, "%13s", _(E1h));
+  else if (diff < 4 * ONE_HOUR)
+    return sprintf(buf, "%13s", _(E2h));
+  else if (diff < ONE_DAY && now.tm_mday == tm.tm_mday)
+    return sprintf(buf, "%13s", _(E6h));
+  else if (diff < 2 * ONE_DAY && now.tm_mday - tm.tm_mday == 1)
+    return sprintf(buf, "%13s", _(E1d));
+  else if (diff < 7 * ONE_DAY)
+    return sprintf(buf, "%4d %s", now.tm_mday - tm.tm_mday, _(ENd));
+  else
+    return strftime(buf, buflen, "  %b %d %Y", &tm);
 }
 
 /**
--- a/hdrline.c
+++ b/hdrline.c
@@ -855,6 +855,13 @@ static const char *index_format_str(char *buf, size_t buflen, size_t col, int co
       break;
     }
 
+    case 'h':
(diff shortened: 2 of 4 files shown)


flatcap:
I'm not sure what's happened to the FreeBSD builds, but it's not your fault :-)
I'll look into to.

---

I've found a bug, with an old email - it shows "**Earlier today**".
It's dated Dec 10<sup>th</sup> and today's Jan 10<sup>th</sup>.

```
From jd@example.com Fri Dec 10 19:06:11 2021
Return-Path: <jd@example.com>
	by example.com (Postfix) with ESMTP id 99C2B43E
	for <flatcap-mail@example.com>; Fri, 10 Dec 2021 19:06:11 +0000 (GMT)
From: John Doe <jd@example.com>
Subject: Test email
To: rich@example.com
Date: Fri, 10 Dec 2021 19:06:05 +0000

testing
```

You'll probably need to adjust the dates by the time you read this.

artagnon:
Hi @flatcap,

Thanks for reviewing. I've fixed the bug now, and changed the capitalization a bit.

mutt/date.c
```diff
@@ -56,6 +57,19 @@ static const char *const Months[] = {
...
+ * Special cases for humanized timestamps
+ */
+static const unsigned ONE_HOUR = 3600;          // In terms of seconds
+static const unsigned ONE_DAY = 24 * ONE_HOUR;  // Again, expressed in seconds
+static const char *const E0h = "just now";      // Within the hour
+static const char *const E1h = "an hour ago";   // Between an hour and two hours
+static const char *const E2h = "few hours ago"; // Between two and four hours
+static const char *const E6h = "earlier today"; // Over four hours ago, the same day
+static const char *const E1d = "yesterday";     // Anytime the previous day
+static const char *const ENd = "days ago";      // N days ago; be precise
```
flatcap, line 70: Static strings need to be marked for translation using `N_("...")`

flatcap, line 70: Please can you change the comments to `///<`
This string is magic to Doxygen, meaning: This comment applies to the thing before it.

See: https://neomutt.org/code/

flatcap (changes requested):

flatcap:
Sorry for not remembering this before: [`index-format-hook`](https://neomutt.org/guide/configuration.html#index-format-hook)

We can approximate your PR using this config:
```vim
set index_format="%4C %-12@date@ %-15.15F %Z (%4c) %s"

index-format-hook  date  "~d<1H"    "just now"
index-format-hook  date  "~d<2H"    "an hour ago"
index-format-hook  date  "~d<4H"    "few hours ago"
index-format-hook  date  "~d<1d"    "earlier today"
index-format-hook  date  "~d<2d"    "yesterday"
index-format-hook  date  "~d<7d"    "this week"
index-format-hook  date  "~d<1m"    "%[%a %d]"
index-format-hook  date  "~d<1y"    "%[%b %d]"
index-format-hook  date  "~A"       "%[%m/%y]"
```

Your PR gives more accurate results though (and translations)
e.g. "earlier today" checks for the "same day" rather than "less than 24 hours".

If we go ahead, then I think we should fit your format into `%h` (human) rather than using a config variable.
Also, we could extend the date range, "this month", "last year", etc

Let me have a think first

flatcap:
OK, let's go for it.

**To do**:
- [ ] Add `%h` expando to `index_format_str()`
- [ ] Drop the config var `humanize_time`
- [ ] Extend the date range to years.  Perhaps:
  - last month
  - N months ago
  - this year
  - last year
  - N years ago

(suggestions welcome)

Thanks!

## Notes

This will affect:

**Config**:
- `$attribution`
- `$forward_attribution_intro`
- `$forward_attribution_trailer`
- `$forward_format`
- `$indent_string`
- `$index_format`
- `$message_format`
- `$pager_format`
- `$post_indent_string`

**Hooks**:
- `fcc-hook`
- `save-hook`

This means you could configure: `save-hook artagnon ~/mail/%h`
and have emails from you automatically saved in a `~/mail/ealier today` folder ;-)

artagnon:
Thanks for detailed notes, Rich. I've, however, decided that I'm not in favor of extending the date ranges: it's not very "human" to show a vague "this year" or "last year"; I'd want to know the exact dates beyond the ranges that I've already sent in the patch.

gahr:
FWIW, I'm in favour of the `index-format-hook` solution and I'd avoid hard-coding any decision on the C code: we do have a way to make it configurable the way the user wants. Why do we want to hard-code a subjective decision instead?

artagnon:
Hm, it's not a lot of code, and I thought it would be a nice feature to have with accurate humanization and translations. Yes, the choices in humanization are subjective, but surely, the community can pitch in, and we can have a consensus?

For what it's worth, all my patches to neomutt are in the same direction: ease of adoption, for people like me, who have experience using web-based email for many years. Personally, I don't like to maintain a long and complicated config file, and would like some conveniences provided to me by a modern mail client.

Another thing perhaps worth noting: I'm a macOS user, and am used to good defaults, with some amount of configurability. I realize that there aren't many users of neomutt like me, but perhaps we can attract more macOS-based users/contributors?

gahr:
> For what it's worth, all my patches to neomutt are in the same direction: ease of adoption, for people like me, who have experience using web-based email for many years. Personally, I don't like to maintain a long and complicated config file, and would like some conveniences provided to me by a modern mail client.

I think in general I favour different tools for different jobs and configurability, rather than pushing more features into a single program. As they say, tools over policies. This is a great case where something can be easily done via configuration.

flatcap:
OK, I've made a decision. **`%h` wins**.

Here are the date ranges.
They're pretty subjective, but I don't want a long debate over them.

| Date Range | Singular       | Plural         | Description                       |
| :--------- | :------------- | :------------- | :-------------------------------- |
| `< 1H`     | just now       | -              | Within the hour                   |
| `< 2H`     | an hour ago    | -              | Between an hour and two hours     |
| `< 4H`     | few hours ago  | -              | Between two and four hours        |
| today      | earlier today  | -              | Over four hours ago, the same day |
| yesterday  | yesterday      | -              | Anytime the previous day          |
| `< 7d`     | a day ago      | %d days ago    | N days ago                        |
| `< 8w`     | a week ago     | %d weeks ago   | N weeks ago                       |
| `< 12m`    | a month ago    | %d months ago  | N months ago                      |
| `>= 12m`   | a year ago     | %d years ago   | N years ago                       |

The strings contain the `%d` and appear in singular and plural forms.
This is to support languages with non-trivial plural forms.
(This will require a change to use `ngettext()`)

To display only _some_ of the strings, `index-format-hook` can be used.

flatcap:
Here's an example of `ngettext()`:
https://github.com/neomutt/neomutt/blob/ea4ac7cca5205672dca53cb50021da42bc7474bb/attach/recvattach.c#L517-L518

It can look a bit repetitive in use, but it allows the translators to handle complicated plurals:
```c
ngettext("Attachment saved", "%d attachments saved", saved_attachments)
```

This returns a translated string (which may or may not contain `%d`), which is fed to `mutt_message()`.

flatcap:
Archived to https://github.com/neomutt/neomutt-old/tree/humanize-dates

PR #3191, opened 2022-01-12
prex: fix bug in regex; pass wasn't captured in url

artagnon:
Hi,

This is a bug I spotted while investigating the way authentication is done in neomutt for my osxkeychain branch.

Thanks.

PR #3192, opened 2022-01-12
newsrc: bugfix; nntp_user and nntp_pass ignored

artagnon:
When connecting to an nntp server such as news.eternal-september.org,
there are three ways to connect: First, without a username/password pair,
in which case, it displays a few posts on the server with information
about how to connect correctly. Second, with the username/password
encoded in the url, which works perfectly after #3191. And lastly, by
specifying a username/password pair in nntp_user and nntp_pass, in which
case, the variables are completely ignored in the connection to the
server. The behavior is puzzling, and this patch proposes a fix.

Signed-off-by: Ramkumar Ramachandra <r@artagnon.com>


flatcap:
> The behavior is puzzling

hehe, yes.

NNTP has been maintained out-of-tree from Mutt for over a decade.
It's never really integrated well.

It seems that we already ask for the username/password, but much later, in `nntp_auth()`.

https://github.com/neomutt/neomutt/blob/0338b9c45145ff34f365307ba9680b6b0d890199/nntp/nntp.c#L437-L439

This shows the code paths leading to `nntp_auth()`


It won't hurt to ask for it in both places, for now.
Please can you change `nntp_select_server()` to use  `mutt_account_getuser()` and `mutt_account_getpass()`.
That way your keychain code will integrate properly.

Thanks

artagnon:
Hm, that's a bit problematic. If the user doesn't need authentication, it'd be unpleasant to prompt for a username/password before even connecting to the server. Yeah, I can see that nntp is a bit unmaintained, but what choice do we have at this point, until someone does a big cleanup?

flatcap:
Archived to https://github.com/neomutt/neomutt-old/tree/nntp-pass

PR #3194, opened 2022-01-13
lens: introduce a new Mailbox type

artagnon:
neomutt faces one stark defeciency compared to modern web-based email:
in an email thread for a received email, the emails that are sent by the
user are missing. In order to fix this, we propose a new Mailbox type
dubbed "lens", that merely acts as a lens into multiple existing
Mailboxes.

This patch is in its very early stages, and is more of a mockup than any
functioning code.

### DIFF
--- a/lens/lens.c
+++ b/lens/lens.c
@@ -1,0 +1,79 @@
+#include <stddef.h>
+#include <string.h>
+#include "mutt/hash.h"
+#include "email/envelope.h"
+#include "core/neomutt.h"
+#include "lib.h"
+#include "queue.h"
+
+static int mailbox_to_lens(struct Mailbox *mailbox, struct MailboxLens *lens)
+{
+  // TODO: Dumb shallow-ref
+}
+
+static int lens_add_email(struct LensEmailList el, struct Email *e)
+{
+  if (!e)
+    return -1;
+
+  struct LensEmail *en = mutt_mem_calloc(1, sizeof(*en));
+  en->email = e;
+  // TODO: en->mailbox
+  STAILQ_INSERT_TAIL(&el, en, entries);
+
+  return 0;
+}
+
+static int merge_mailboxes(struct Mailbox *primary, struct MailboxList ml,
+                           struct LensMailbox *merged)
+{
+  struct MailboxNode *np = NULL;
+  int alloc = 0;
+  if (!primary)
+    return -1;
+  mailbox_to_lens(primary, merged);
+
+  // Prep: Find the number of elements in all our hash tables
+  STAILQ_FOREACH(np, &ml, entries)
+  {
+    if (!np->mailbox)
+      continue;
+
+    alloc += np->mailbox->id_hash->num_elems;
+  }
+
+  // Step 1: Copy out id_hash into id_hash_aux, converting every key into a LensEmail
+  // TODO
+
+  // Step 2: Merge all the hash tables
+  struct HashTable *table = mutt_hash_new(alloc, MUTT_HASH_NO_FLAGS);
+  STAILQ_FOREACH(np, &ml, entries)
+  {
+    mutt_hash_merge(table, np->mailbox->id_hash);
+  }
+
+  // Now the real work: match in-reply-to against the msg-ids in table
+  struct LensEmail *le = NULL;
+  char *ref = NULL;
+  struct Email *match = NULL;
+  STAILQ_FOREACH(le, &merged->emails, entries)
+  {
+    STAILQ_FOREACH(ref, &le->email->env->in_reply_to, entries)
+    {
+      // TODO: This step should directly return a LensEmail
+      if ((match = mutt_hash_find(table, ref)))
+        lens_add_email(merged->emails, match);
+    }
+  }
+
+  mutt_hash_free(&table);
+  return 0;
+}
+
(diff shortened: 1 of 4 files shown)


georgalis:
is this complexity really desirable? what is the advantage verses setting a default Bcc: {from address} on all your clients?

dcpurton:
I use Maildir and I do this using [`unionfs-fuse`](https://github.com/rpodgorny/unionfs-fuse). 

My use case was to combine INBOX, Sent, and Archive together. I made it read only just in case Neomutt got confused.

mschilli87:
Why do you need the Sent folder at all in this case? Couldn't you simply set `record = '+INBOX'` or use an `fcc-save-hook` to directly drop sent mails in your `INBOX`? This way, there would only be one copy of the sent mail on the IMAP server in the same mailbox as the one it was replying to.

artagnon:
> Why do you need the Sent folder at all in this case? Couldn't you simply set record = '+INBOX' or use an fcc-save-hook to directly drop sent mails in your INBOX? This way, there would only be one copy of the sent mail on the IMAP server in the same mailbox as the one it was replying to.

First, the web-email experience would degrade: the `Sent` tab on the interface would be empty. And more importantly, replying to email using the app or web-email UI would put stuff in `Sent`, and opening that thread in neomutt would display missing items.

mschilli87:
I am sure there would be a way to make emails send from the web go to INBOX as well. If not on their end, then via chron job or even some sort of neomutt hook over IMAP. Regarding the 'Sent` directory: Just delete it and replace it by a query for your own email in the 'From' field in case you really ever need a full list of all emails send by you. Could be done using notmuch in neomutt and I am sure whatever web client you use will offer an indexed search as well. :wink:
This is not arguing against you patch but rather pointing out an alternative.workaround in case this doesn't end up in a release in the end. :wink:

artagnon:
> This is not arguing against your patch but rather pointing out an alternative.workaround in case this doesn't end up in a release in the end.

I appreciate it, but I badly want this feature (and prefer writing C code to jumping through hoops, setting up cron jobs etc.), and I think it's doable with some help from the community :)

gahr:
I smell feature bloat. I am not sure this should be done within NeoMutt itself as opposed to any of the suggested alternatives.

artagnon:
Hm, I'm actually leaning towards making this part of core neomutt, modifying the `Mailbox` and `Email` data structures directly to support this use case elegantly. It will involve some refactoring work, but at the very least, there won't be a separate Mailbox type called lens, which duplicates everything, and does a lot of grunt work. The question would then be turned into: is this a common enough use case to support directly in core?

gahr:
> The question would then be turned into: is this a common enough use case to support directly in core?

I have different questions: how flexible is this?, and can this be achieved via a dedicated tool? I fear the answers are "not very" and "yes, with notmuch". Notmuch can do what you're trying to achieve here (multiple mailboxes into a single view) and much more. Please, I kindly ask you to consider trying that and see if it fits your use case before re-implementing part of it in NeoMutt.

artagnon:
I've decided that I'm not happy with this mockup either. Until we find an elegant way to achieve the same thing, I'm closing this PR.

flatcap:
The trick is polymorphism.
I imagine a Lens Mailbox being indistinguishable from any other Mailbox.

I still have a lot to think through, but here's a possible walkthrough.
(all ideas are open to suggestion)

`struct Mailbox` holds the data common to all Mailbox types.
An array of `Email`, numbers of new messages, etc.

Any data that's specific to a Mailbox type is stored in `Mailbox->mdata`.

We want to create a new `MailboxType`: `MUTT_LENS`.
This type is associated with the URL form: `lens://`

The Lens Mailbox will store any private data in `struct LensMboxData`.
It needs to hold a list of children `Mailbox`es
and it needs to remember which `Email` belongs to which `Mailbox`.

config: `lens-mailbox my-inbox imap://inbox imap://sent`

`parse_lens_mailbox()` would create a Mailbox of type `MUTT_LENS` 
and `lens_add_mailbox(Inbox)` and `(Sent)`.

`lens_add_mailbox()` will:
- Add the `Mailbox` to the list of children
- Copy the child's `Email`s into the Lens `Mailbox`
- Set up an observer of the child `Mailbox`

After configuration, most of the behaviour of this `Mailbox` will be controlled using the MXAPI.

User opens `lens://my-inbox`
The index is given the Lens Mailbox and parses it as normal.
It sorts and threads the `Email`s in `Mailbox->emails`.
The Index observes the `Mailbox`.

A new mail arrives in `imap://inbox.`
A notification event is raised on the `Email`.
The event is passed to the `Mailbox` (and then to the `Account`).

The Lens Mailbox is observing `imap://inbox` and sees the new mail.
It inserts the new Email into the array, and updates its 'who's who' list.
The Mailbox sends out a notification that it has changed.

The Inbox receives the notification of change and re-sorts / re-threads the `Email`s.

---

I'll make this all clearer, when I understand it myself.

flatcap:
Have a look at `struct MxOps` (the MXAPI).
Each backend implements most of those functions.

Also see how the backends store their private data.
Search for `@extends Mailbox` and `@extends Email`

flatcap:
Archived to https://github.com/neomutt/neomutt-old/tree/lens

PR #3197, opened 2022-01-15
Add support for composing 'multipart/related' emails

dcpurton:
## What does this PR do?

Adds support for composing `multipart/related` emails.

## Example

run `./neomutt -n -F multipart-related.rc`

#### `multipart-related.rc`

```
# vim: syn=neomuttrc

# filename: multipart-related.rc

# run from build directory:
#   ./neomutt -n -F multipart-related.rc

# key in <F1> through <F6> to construct the email

unset record
unset signature

set attach_format = "%u%D%I %t%4n %T%.40d%> [%.15m/%.15M, %.6e%?C?, %C?, %s]"
set editor = "echo '![Neomutt Logo](cid:neomutt-256.png)' >> %s"
set from = "john.doe@example.com"
set help = no
set postpone = no
set postponed = "postponed.mbox"
set realname = "John Doe"
set recall=no
set spoolfile = "inbox.mbox"

macro index   <F1> "<mail>John Doe <john.doe@example.com><enter>test<enter>"
macro compose <F2> "<attach-file>contrib/logo/neomutt-256.png<enter>"
macro compose <F3> "<rename-attachment><enter><toggle-disposition>"
macro compose <F4> "<attach-file>html-part.html<enter>"
macro compose <F5> "<toggle-disposition><tag-entry>1<enter><tag-entry><group-alternatives>"
macro compose <F6> "<tag-entry>4<enter><tag-entry><group-related>"
```

#### `html-part.html`

```HTML
<!DOCTYPE html>
<html lang="en">
<head>
  <meta charset="utf-8">
  <meta name="generator" content="pandoc">
  <meta name="viewport" content="width=device-width, initial-scale=1.0, user-scalable=yes">
  <title> </title>
</head>
<body>
<div style="max-width: 15cm;">
<p style="margin: 1ex 0; padding: 0;"><img src="cid:neomutt-256.png" alt="Neomutt Logo"></p>
</div>
</body>
</html>
```

### Final email structure

```
  I     1 <no description>                                                   [multipart/related, 7bit, 0K]
  I     2 ├─><no description>                                            [multipart/alternative, 7bit, 0K]
- I     3 │ ├─>/tmp/neomutt-hostname-XXXX-XXXXXX-XXXXXXXXX              [text/plain, 7bit, us-ascii, 0.1K] 
  I     4 │ └─>html-part.html                                          [text/html, 8bit, iso-8859-1, 0.4K] 
  I     5 └─>neomutt-256.png                                                      [image/png, base64, 19K] 
```

## Behaviour at receiving end

| MUA | PGP: Clear | PGP: Signed | PGP: Signed and Encrypted |
|---|:---:|:---:|:---:|
| Gmail (Android) | ✔️ | ✔️ | n/a |
| Gmail (Android, Exchange) | ✔️¹ | ❌² | n/a |
| Gmail (iOS) | ✔️ | ✔️ | n/a |
| Gmail (Web) | ✔️ | ✔️ | n/a |
| Mail App (Windows) | ✔️ | ✔️ | n/a |
| Outlook (Android) | ✔️ | ✔️ | n/a |
| Outlook (Web) | ✔️ | ✔️ | n/a |
| Outlook (iOS) | ✔️ | ✔️ | n/a |
| Outlook (Web) | ✔️ | ✔️ | n/a |
| Outlook (Windows) | ✔️ | ✔️ | n/a |
| Proton (Android) | ✔️ | ❌³ | ❌³ |
| Proton (iOS) | ✔️ | ✔️⁴ | ✔️⁴ |
| Proton (Web) | ✔️ | ✔️ | ✔️ |
| Thunderbird (Linux) | ✔️ | ✔️ | ✔️ |

¹ Also displays embedded image as attachment (but emails from other clients display like this too).
² Embedded image is not displayed and isn't available in attachment list. (Also fails on messages from Thunderbird.)
³ Embedded image is not displayed, but is available in attachment list. (Also fails on messages from Thunderbird.)
⁴ Thinks there is remote content for some reason.

## Notes on constructing emails

- Attachments in the `multipart/related` group should probably have inline disposition. (Microsoft Outlook shows also shows them as a separate attachment if they have disposition attachment, but Gmail doesn't.) So ensure embedded images have the header `Content-Disposition: inline; filename="image.png"`. The attachment needs to be given a `d_filename` (with `<rename-attachment>`) for the `filename` field to be included.
- Actual attachments should be placed outside of the `multipart/related` group.

### DIFF
--- a/compose/functions.c
+++ b/compose/functions.c
@@ -123,6 +123,27 @@ static char *gen_cid(void)
 }
 
 /**
+ * check_cid - Check if a Content-ID is valid
+ * @param  cid   Content-ID to check
+ * @retval true  Content-ID is valid
+ * @retval false Content-ID is not valid
+ */
+static bool check_cid(const char *cid)
+{
+  static const char *check = "^[-\\.0-9@A-Z_a-z]+$";
+  struct Buffer buf = mutt_buffer_make(0);
+
+  struct Regex *check_cid_regex = mutt_regex_new(check, 0, &buf);
+  mutt_buffer_dealloc(&buf);
+
+  const bool valid = mutt_regex_match(check_cid_regex, cid);
+
+  mutt_regex_free(&check_cid_regex);
+
+  return valid;
+}
+
+/**
  * count_attachments - Count attachments
  * @param  body    Body to start counting from
  * @param  recurse Whether to recurse into groups or not
--- a/compose/functions.c
+++ b/compose/functions.c
@@ -107,6 +107,22 @@ static bool check_count(struct AttachCtx *actx)
 }
 
 /**
+ * gen_cid - Generate a random Content ID
+ * @retval ptr Content ID
+ *
+ * @note The caller should free the string
+ */
+static char *gen_cid(void)
+{
+  char rndid[MUTT_RANDTAG_LEN + 1];
+
+  mutt_rand_base32(rndid, sizeof(rndid) - 1);
+  rndid[MUTT_RANDTAG_LEN] = 0;
+
+  return mutt_str_dup(rndid);
+}
+
+/**
  * count_attachments - Count attachments
  * @param  body    Body to start counting from
  * @param  recurse Whether to recurse into groups or not
--- a/compose/functions.c
+++ b/compose/functions.c
@@ -1131,6 +1131,61 @@ static int op_attachment_detach(struct ComposeSharedData *shared, int op)
 }
 
 /**
+ * op_attachment_edit_content_id - Edit the 'Content-ID' of the attachment - Implements ::compose_function_t - @ingroup compose_function_api
+ */
+static int op_attachment_edit_content_id(struct ComposeSharedData *shared, int op)
+{
+  if (!check_count(shared->adata->actx))
+    return IR_NO_ACTION;
+
+  int rc = IR_NO_ACTION;
+  struct Buffer *buf = mutt_buffer_pool_get();
+  struct AttachPtr *cur_att =
+      current_attachment(shared->adata->actx, shared->adata->menu);
+
+  char *id = mutt_param_get(&cur_att->body->parameter, "content-id");
+  if (id)
+  {
+    mutt_buffer_strcpy(buf, id);
+  }
+  else
+  {
+    id = gen_cid();
+    mutt_buffer_strcpy(buf, id);
+    FREE(&id);
+  }
+
+  if (mutt_buffer_get_field("Content-ID: ", buf, MUTT_COMP_NO_FLAGS, false,
+                            NULL, NULL, NULL) == 0)
+  {
+    if (!mutt_str_equal(id, mutt_buffer_string(buf)))
+    {
+      if (check_cid(mutt_buffer_string(buf)))
+      {
+        mutt_param_set(&cur_att->body->parameter, "content-id", mutt_buffer_string(buf));
+        menu_queue_redraw(shared->adata->menu, MENU_REDRAW_CURRENT);
+        notify_send(shared->notify, NT_COMPOSE, NT_COMPOSE_ATTACH, NULL);
+        mutt_message_hook(NULL, shared->email, MUTT_SEND2_HOOK);
+        rc = IR_SUCCESS;
+      }
+      else
+      {
+        mutt_error(
+            _("Content-ID can only contain the characters: -.0-9@A-Z_a-z"));
+        rc = IR_ERROR;
+      }
+    }
+  }
+
+  mutt_buffer_pool_release(&buf);
+
+  if (rc != IR_ERROR)
+    mutt_clear_error();
+
+  return rc;
+}
+
+/**
  * op_attachment_edit_description - Edit attachment description - Implements ::compose_function_t - @ingroup compose_function_api
  */
 static int op_attachment_edit_description(struct ComposeSharedData *shared, int op)
@@ -2450,6 +2505,7 @@ struct ComposeFunction ComposeFunctions[] = {
   { OP_ATTACHMENT_ATTACH_MESSAGE,         op_attachment_attach_message },
   { OP_ATTACHMENT_ATTACH_NEWS_MESSAGE,    op_attachment_attach_message },
   { OP_ATTACHMENT_DETACH,                 op_attachment_detach },
+  { OP_ATTACHMENT_EDIT_CONTENT_ID,        op_attachment_edit_content_id },
   { OP_ATTACHMENT_EDIT_DESCRIPTION,       op_attachment_edit_description },
   { OP_ATTACHMENT_EDIT_ENCODING,          op_attachment_edit_encoding },
   { OP_ATTACHMENT_EDIT_LANGUAGE,          op_attachment_edit_language },
--- a/compose/functions.c
+++ b/compose/functions.c
@@ -1747,6 +1747,39 @@ static int op_attachment_toggle_unlink(struct ComposeSharedData *shared, int op)
 }
 
 /**
+ * op_attachment_group_related - Group tagged attachments as 'multipart/related' - Implements ::compose_function_t - @ingroup compose_function_api
+ */
+static int op_attachment_group_related(struct ComposeSharedData *shared, int op)
+{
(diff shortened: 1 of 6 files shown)


flatcap:
It works!

dcpurton:
Using a Neomutt macro to pipe the primary attachment (in Markdown format) to a sufficiently complex script does allow you to construct and send a `multipart/alternative` email wrapped in a `multipart/related` group if required.

Consider initial email structure like this:

```
- I     1 /tmp/neomutt-hostname-XXXX-XXXXXX-XXXXXXXXXX                  [text/plain, 7bit, us-ascii, 0.1K]
  A     2 image.png                                                               [image/png, base64, 19K]
  A     3 attachment.pdf                                                   [application/pdf, quoted, 7.1K]
```

`image.png` can be referenced in the main file as:

```
![Alt Text](cid:image.png)
```

We can send the message with a single macro:

```
macro compose y "<first-entry><enter-command>set wait_key=no<enter><pipe-entry>msg2html.sh<enter>
  <attach-file>/tmp/neomutt-alternative.html<enter><toggle-unlink><toggle-disposition><tag-entry>
  <first-entry><tag-entry><group-alternatives><enter-command>set wait_key=yes<enter>
  <enter-command>source /tmp/neomutt.rc<enter><send-message>"
```

`msg2html.sh` produces `/tmp/neomutt-alternative.html` using `pandoc` and parses the Markdown looking for `(cid:filename)` and creates `/tmp/neomutt.rc` with the commands to tag the appropriate files and create the `related/multipart` group. If no `(cid:filename)` is found, then leave `/tmp/neomutt.rc` empty.

In the above case `/tmp/neomutt.rc` would contain:

```
push "<group-related>"
push "<search>image.png<enter><tag-entry>"
push "<tag-entry>"
```

The final email structure before sending will be:

```
  I     1 <no description>                                                   [multipart/related, 7bit, 0K]
  I     2 ├─><no description>                                            [multipart/alternative, 7bit, 0K]
- I     3 │ ├─>/tmp/neomutt-hostname-XXXX-XXXXXX-XXXXXXXXXX             [text/plain, 7bit, us-ascii, 0.1K]
- I     4 │ └─>/tmp/neomutt-alternative.html                                [text/html, 8bit, utf-8, 0.6K]
  A     5 └─>image.png                                                            [image/png, base64, 19K]
  A     6 attachment.pdf                                                   [application/pdf, quoted, 7.1K]
```

At send time Neomutt then wraps all this in a `multipart/mixed` group and signs/encrypts it if required.

And provided you haven't forgot to attach anything, misspelled anything, or omitted the `cid:` … it will all work fine at the receiving end.

send/header.c
```diff
@@ -784,6 +786,13 @@ int mutt_write_mime_header(struct Body *a, FILE *fp, struct ConfigSubset *sub)
       struct Parameter *cont = NULL;
       TAILQ_FOREACH(cont, &pl_conts, entries)
       {
+        if (mutt_istr_equal(cont->attribute, "content-id"))
+        {
+          // Content-ID: gets its own header
+          id = mutt_str_dup(cont->value);
```
flatcap, line 792: Can the `Parameter` change during this function?
If not, then `id = cont->value;` would do (and we don't need to `FREE()` later)

compose/functions.c
```diff
@@ -1488,6 +1527,43 @@ static int op_compose_group_lingual(struct ComposeSharedData *shared, int op)
...
+  if (shared->adata->menu->tagged < 2)
+  {
+    mutt_error(_("Grouping 'related' requires at least 2 tagged messages"));
+    return IR_ERROR;
+  }
+
+  // ensure Content-ID is set for tagged attachments
+  for (struct Body *b = shared->email->body; b; b = b->next)
+  {
+    if (b->tagged)
```
flatcap, line 1545: This `if` doesn't have an `else` clause, so we can "return early"

```c
if (!b->tagged)
  continue;
```

This will reduce the indent of the code and mean that the reader doesn't need to hunt for an `else` if the `Body` isn't tagged.

compose/functions.c
```diff
@@ -1488,6 +1527,43 @@ static int op_compose_group_lingual(struct ComposeSharedData *shared, int op)
...
+    {
+      char *id = mutt_param_get(&b->parameter, "content-id");
+      if (!id)
```
flatcap, line 1548: Same here:
```c
if (id)
  continue;
```

email/parse.c
```diff
@@ -1379,6 +1379,17 @@ struct Body *mutt_read_mime_header(FILE *fp, bool digest)
         mutt_str_replace(&p->description, c);
         rfc2047_decode(&p->description);
       }
+      else if (mutt_istr_equal("id", line + plen))
+      {
+        // strip <angle braces> from Content-ID: header
```
flatcap, line 1384: Is it OK to assume there are `<>`s?
Should we check?

email/parse.c
```diff
@@ -1379,6 +1379,17 @@ struct Body *mutt_read_mime_header(FILE *fp, bool digest)
...
+        char *id = NULL;
+        if (mutt_str_len(c) > 2)
+        {
+          id = c + 1;
+          id[mutt_str_len(id) - 1] = '\0';
```
flatcap, line 1389: `strlen()` isn't very costly, but could we do it just once? :-)

flatcap (changes requested):
- Rebased
- Clang-formatted
- A couple of minor tweaks to the docs

Looking good

send/header.c
dcpurton: I think I was originally going to add the angle braces in to `id`, but later just used the `fprintf` and forgot I didn't need to make a copy.

email/parse.c
dcpurton: It can't harm to check. Can it be assumed that there is no leading or trailing white space at this point?

dcpurton:
@flatcap , your discussion on saving out `multipart/related` groups and the format of `Content-ID` got me thinking.

It would be better for Neomutt to produce a RFC compliant `Content-ID` when sending email (accepting email is a different matter and Neomutt should be generous with what is accepted).

But there are some complexities:

- For the case of an automatically generated `Content-ID`, the domain name can be appended easily enough and this can be documented. I actually considered doing this, but it made it slightly difficult for a script to predict the `Content-ID`, since it can change based on configuration.
- For the case of a user specified `Content-ID`, what should happen if the user does not specify a RFC compliant `Content-ID`?
    - Should Neomutt change it?
        - But this is confusing and annoying for a user.
    - Should Neomutt warn?
        - Potentially adds configuration options on whether to warn or not and what should be done by default.
    - Should Neomutt just allow the user to do what they want?
- For the case of a resent/draft/postponed message containing a non-compliant `Content-ID` (or even a compliant one, but with the wrong domain), what should Neomutt do?
    - Change the `Content-ID` (as is done for the `Message-ID`)?
        - But this involves modifying attachments in the group, which seems like a bad idea.
    - Just use the `Content-ID` as is and ignore the fact that it isn't RFC compliant.

A penny for your thoughts.

flatcap:
Rebased over a bunch of huge Compose-function renames.

flatcap (approved):
Oh!  It was going so well... and you blew it :-)
It doesn't work; the `Content-ID` that gets sent is garbage.

However, I've pushed some fixes.

In `mutt_write_mime_header()`, you `id = cont->attribute;` but you actually wanted `cont->value`

Just below that, `pl_conts` gets freed: `mutt_param_free(&pl_conts);` which means that `id` is no longer valid.  I've added a `strdup()` and `free()` to work around this.

I've added a `mutt_buffer_dealloc()` to `check_cid()` to cover the possibility of an error.

Finally, in the docs it says: "an HTML part ... needs to contain ... ``"
I've removed the reference to `alt="Alt Text"` as that's not _required_.

dcpurton:
> Oh! It was going so well... and you blew it :-) It doesn't work; the `Content-ID` that gets sent is garbage.

Oops! It definitely worked at one stage.

> However, I've pushed some fixes.

Great.

> Finally, in the docs it says: "an HTML part ... needs to contain ... ``" I've removed the reference to `alt="Alt Text"` as that's not _required_.

Fine.

PR #3198, opened 2022-01-16
hcache-bench: adjust README.md linebreaks

me-and:
Some Markdown implementations (e.g. the GitHub one, as can be seen at [0]) can treat a line starting with `-` as a list item.  Move a line break to avoid that happening unnecessarily.

[0]: https://github.com/neomutt/neomutt/blob/1ce876f0c2935832ed6a2c11ff85dd5708f11daa/contrib/hcache-bench/README.md

### DIFF
--- a/contrib/hcache-bench/README.md
+++ b/contrib/hcache-bench/README.md
@@ -9,8 +9,8 @@ benchmark the NeoMutt hcache backends.
 
 In order to run the benchmark, you must have a directory in maildir format at
 hand. NeoMutt will load messages from there and populate the header cache with
-them. Please note that you'll need a reasonable large number of messages - >50k
-- to see anything interesting.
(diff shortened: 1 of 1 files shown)


flatcap:
Ah!   Good spot.
Unfortunately, you've fallen into a different trap.
By moving the `>50` to the beginning of the line, [GitHub thought it was an indented quote](https://github.com/me-and/neomutt/blob/220fb2b46898b20bae3352117a8813d6c94556a1/contrib/hcache-bench/README.md)!

I've reworded it to be unambiguous.

Thanks

flatcap (approved):

me-and:
Thank you! I thought I'd checked it rendered correctly, but clearly whatever I checked, it wasn't that!

PR #3223, opened 2022-01-29
Support for inserting attachments

dcpurton:
## What does this PR do?

Add support for inserting attachments before and after the currently selected one.

- [x] attaching existing files before and after the current entry.

  - `<attach-file-after>` (Alt+a) command
  - `<attach-file-before>` (Alt+b) command

- [x] creating new files positioned before and after the current entry.

  - `<mime-new-after>` (Alt+A) command
  - `<mime-new-before>` (Alt+B) command

## Discussion

When attaching multiple files, order is reversed to match with `<attach-file>`.

### DIFF
--- a/compose/functions.c
+++ b/compose/functions.c
@@ -732,6 +732,15 @@ static int op_attachment_attach_file_after(struct ComposeSharedData *shared, int
                               menu_get_index(shared->adata->menu), true);
 }
 
+/*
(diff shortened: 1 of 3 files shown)


flatcap:
I've pushed some changes to `master` that affect how functions are defined and keys are bound.
I've rebased your PR over them.

Issue #3217, opened 2022-01-26
Mails that need PGP decryption (and passphrase) leave altmode


nabijaczleweli:
## Expected Behaviour

That, after GPG asks for my passphrase, neomutt gets back into altmode.

## Actual Behaviour

i be scrollin
![image](https://user-images.githubusercontent.com/6709544/151162912-821fa08c-8c7e-4956-85cb-13cd20100d61.png)

## Steps to Reproduce

1. Make sure you don't have GPG credentials cached
2. Click into an encrypted mail
3. Back out of the prompt
4. Scroll: you're no longer in altmode
5. Press E, back out of `$EDITOR`: you're back in altmode

## How often does this happen?

- Sometimes

I think it's just when it's not cached.

## When did it start to happen?

Always, but I receive more encrypted mail nowadays.

## NeoMutt Version

```
NeoMutt 20211029
Copyright (C) 1996-2020 Michael R. Elkins and others.
NeoMutt comes with ABSOLUTELY NO WARRANTY; for details type 'neomutt -vv'.
NeoMutt is free software, and you are welcome to redistribute it
under certain conditions; type 'neomutt -vv' for details.

System: Linux 5.10.0-10-amd64 (x86_64)
ncurses: ncurses 6.2.20201114 (compiled with 6.2.20201114)
libidn: 1.33 (compiled with 1.33)
GPGME: 1.14.0-unknown
GnuTLS: 3.7.1
libnotmuch: 5.3.0
storage: tokyocabinet, lmdb
compression: zlib, zstd

Configure options: --build=x86_64-linux-gnu --prefix=/usr {--includedir=${prefix}/include} {--mandir=${prefix}/share/man} {--infodir=${prefix}/share/info} --sysconfdir=/etc --localstatedir=/var --disable-option-checking --disable-silent-rules {--libdir=${prefix}/lib/x86_64-linux-gnu} {--libexecdir=${prefix}/lib/x86_64-linux-gnu} --disable-maintainer-mode --disable-dependency-tracking --mandir=/usr/share/man --libexecdir=/usr/libexec --with-mailpath=/var/mail --gpgme --lua --notmuch --gnutls --gss --idn --mixmaster --sasl --tokyocabinet --lmdb --zlib --zstd --testing

[...]

PR #3224, opened 2022-01-29
fix screen mode after attempting decryption

flatcap:
When decrypting an email, gpg can change the screen mode.
This can confuse ncurses, causing it to write to the "normal", rather than the "alternate" screen.

Forcibly switch to the alternate screen after gpg operations.

Fixes: #3217

### DIFF
--- a/gui/curs_lib.c
+++ b/gui/curs_lib.c
@@ -135,6 +135,10 @@ void mutt_refresh(void)
  */
 void mutt_need_hard_redraw(void)
 {
+  // Forcibly switch to the alternate screen.
+  // Using encryption can leave ncurses confused about which mode it's in.
+  fputs("\033[?1049h", stdout);
+  fflush(stdout);
   keypad(stdscr, true);
   clearok(stdscr, true);
   window_redraw(NULL);


gui/curs_lib.c
```diff
@@ -135,6 +135,9 @@ void mutt_refresh(void)
  */
 void mutt_need_hard_redraw(void)
 {
+  // Forcibly switch to the alternate screen.
+  // Using encryption can leave ncurses confused about which mode it's in.
+  fputs("\033[?1049h", stderr);
```
gahr, line 140: Why stderr?

flatcap: good question :-)
With lots of redirections going on during pin-entry, I wanted to ensure that it was actually emitted.

My slight worry is that there's a CLI code path that can get here that leaves the user in the alternate screen.

gahr: meaning it would be bogus to run neomutt with stderr redirected to file anyway?

flatcap: updated: `stdout`, `fflush()`

gahr (approved):

PR #3245, opened 2022-02-08
Support compiling under Solaris 10

pekdon:
* Add support compiling under Solaris 10

Add compatibility functions and use existing wrappers to get neomutt running under Solaris 10.

### DIFF
--- a/Makefile.autosetup
+++ b/Makefile.autosetup
@@ -13,6 +13,7 @@ CFLAGS_FOR_BUILD=	@CFLAGS_FOR_BUILD@ $(EXTRA_CFLAGS_FOR_BUILD)
 LDFLAGS_FOR_BUILD=	@LDFLAGS_FOR_BUILD@ $(EXTRA_LDFLAGS_FOR_BUILD)
 
 PWD=		@PWD@
+SH?=		sh
 AR=		@AR@
 CC=		@CC@
 CPP=		@CPP@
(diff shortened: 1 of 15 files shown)


gahr:
I think we once decided we'd target C90 as a minimum. That's allowed us to get rid of a number of shims to support very old systems. What C standard does the Solaris 10 compiler support?

pekdon:
The shipped compiler is GCC-3.4.3 assuming with some Sun patches, I'm however currently using GCC 8.4.0.
POSIX.1-2001 is the highest supported POSIX version.

The auto.defs requires c99 it seems (and compiles with -std=c99)

gahr:
Yup, I meant to write C99, not C90 :)

gahr:
What C99 feature are you missing on gcc 8.4.0?

pekdon:
I'm afraid that I don't quite follow or know what you're asking about. The functions added in this PR are timegm (GNU extension, now available on BSD), strsep (BSD), strcasestr also non standard.

strncasecmp might actually be possible to rid from the PR however, I'll try that out tomorrow now when the wrappers are used.

As for the features, C99 from a language perspective should be complete however don't get a new libc with the compiler (as it is with C++ and a new libstdc++) so stuck with whatever was on the system before.

flatcap:
Minor changes:
- Rebased over master
- clang-format
- Doxygen
  - function comments
  - link up new page

flatcap (approved):

PR #3246, opened 2022-02-09
Refactor Sidebar Functions

flatcap:
In the beginning, the Index and Pager had to handle all the sidebar function in a big `switch`.
Next came the separate op functions -- better but still lots of duplication.

This PR creates a Sidebar-specific function dispatcher.
Now, the Index and Pager are _almost_ unaware of the Sidebar.

- 694a76159 sidebar: reorder functions
- 669f5a5d4 sidebar: rename functions
- b808899be sidebar: refactor op_sidebar_next/prev()
- 8490dd35f sidebar: add the window to wdata
- 33842b594 sidebar: convert work functions
- 7f3d28177 sidebar: create functions for open and toggle
- beef03872 sidebar: add a function dispatcher
- 1736fbbb0 index/pager: convert to use the sidebar dispatcher

The changes are broken into small pieces.
The first couple don't make any functional changes.
Then refactor the functions to the right format, and create the dispatcher.

---

In the future, I'd like the dispatchers to be connected to the windows.
When an event happened, the event loop would search for a handler starting with the focused window.

### DIFF
--- a/index/dlg_index.c
+++ b/index/dlg_index.c
@@ -1428,3 +1428,28 @@ struct MuttWindow *index_pager_init(void)
 
   return dlg;
 }
+
+/**
+ * dlg_change_folder - Change the current folder, cautiously
+ * @param dlg Dialog holding the Index
+ * @param m   Mailbox to change to
+ */
+void dlg_change_folder(struct MuttWindow *dlg, struct Mailbox *m)
+{
+  if (!dlg || !m)
+    return;
+
+  struct IndexSharedData *shared = dlg->wdata;
+  if (!shared)
+    return;
+
+  struct MuttWindow *panel_index = window_find_child(dlg, WT_INDEX);
+  if (!panel_index)
+    return;
+
+  struct IndexPrivateData *priv = panel_index->wdata;
+  if (!priv)
+    return;
+
+  change_folder_mailbox(priv->menu, m, &priv->oldcount, shared, false);
+}
--- a/index/dlg_index.c
+++ b/index/dlg_index.c
@@ -97,6 +97,9 @@
 #ifdef USE_INOTIFY
 #include "monitor.h"
 #endif
+#ifdef USE_SIDEBAR
+#include "sidebar/lib.h"
+#endif
 
 /// Help Bar for the Index dialog
 static const struct Mapping IndexHelp[] = {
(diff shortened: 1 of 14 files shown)


flatcap:
This means that the Sidebar library is reduced to just four entry points:
```c
void sb_init(void);
void sb_shutdown(void);
bool config_init_sidebar(struct ConfigSet *cs);
int sb_function_dispatcher(struct MuttWindow *win_sidebar, int op);
```

There's a small amount of sidebar data outside of the library.
- functions
- colours
- sort methods

These still rely on static lists.

Once the library is initialised, the Sidebar Window manages itself.
It listens for notifications, e.g. `NT_MAILBOX_CHANGE` and responds to events like `NT_WINDOW_STATE`.

index/dlg_index.c
```diff
@@ -1338,8 +1341,16 @@ struct Mailbox *mutt_index_menu(struct MuttWindow *dlg, struct Mailbox *m_init)
     nm_db_debug_check(shared->mailbox);
 #endif
 
-    int rc = index_function_dispatcher(priv->win_index, op);
+    int rc = IR_UNKNOWN;
 
+#ifdef USE_SIDEBAR
+    struct MuttWindow *win_sidebar = window_find_child(dlg, WT_SIDEBAR);
+    rc = sb_function_dispatcher(win_sidebar, op);
+    if ((rc != IR_ERROR) && (rc != IR_UNKNOWN))
```
gahr, line 1349: If you continue on `IR_ERROR` you'll get a `Key is not found` on, e.g., `unmailboxes *; sidebar-first`.

pager/dlg_pager.c
```diff
@@ -831,7 +834,16 @@ int mutt_pager(struct PagerView *pview)
       continue;
     }
 
-    int rc = pager_function_dispatcher(priv->pview->win_pager, op);
+    int rc = IR_UNKNOWN;
+
+#ifdef USE_SIDEBAR
+    struct MuttWindow *win_sidebar = window_find_child(dlg, WT_SIDEBAR);
+    rc = sb_function_dispatcher(win_sidebar, op);
+    if ((rc != IR_ERROR) && (rc != IR_UNKNOWN))
```
gahr, line 842: Same as above

gahr (commented):
Great job! Minor comments inline.

index/dlg_index.c
gahr: On the other hand, if you don't, you won't be able to quit the `messages` window anymore.. doh..

flatcap:
**Updates**:

- Sidebar dispatcher is tried **after** the Index/Pager's own dispatcher
  This should fix any differences in behaviour

- Sidebar dispatcher only returns success or unknown
  The callers only need to know whether to continue searching, or not

- Dispatchers log the outcome
  `Handled OP_PAGER_BOTTOM (179) -> no action`

gahr:
Works fine, thanks!

gahr (approved):

Issue #3259, opened 2022-02-18
Does $header_cache_backend really need to be set for header caching?


cbrownstein:
According to the NeoMutt Guide Reference section, no header caching will be used if [`$header_cache_backend`](https://neomutt.org/guide/reference.html#header-cache-backend) isn't set.

The same is said in the NeoMutt Guide Optional Features section under the [Header Caching](https://neomutt.org/guide/optionalfeatures.html#header-caching) heading:

> Additionally, [$header_cache_backend](https://neomutt.org/guide/reference.html#header-cache-backend) **must** be set to specify which backend to use. The list of available backends can be specified at configure time with a set of --with-<backend> options. Currently, the following backends are supported: bdb, gdbm, kyotocabinet, lmdb, qdbm, rocksdb, tdb, tokyocabinet.

Despite what the documentation says, header caching works for me with only `$header_cache` set (to a directory). I do not have `$header_cache_backend` set.

So… does `$header_cache_backend` really need to be set for header caching? It doesn't seem so.

This is probably a trivial question I'm asking but, IMO, the documentation should accurately state under what conditions it is actually necessary to set `$header_cache_backend`.

flatcap:
> So… does `$header_cache_backend` really need to be set for header caching?
> It doesn't seem so.

You're right.  That needs fixing.

In fact since we split up the hcache (442faa55a five years ago),
NeoMutt's picked a default backend if `$header_cache_backend` wasn't set.

The code defaults to the first entry...
https://github.com/neomutt/neomutt/blob/f3b9af50105daaa18563e929d4d39bd7a88de372/store/store.c#L107-L114

... in this table (dependent on `./configure` options
https://github.com/neomutt/neomutt/blob/f3b9af50105daaa18563e929d4d39bd7a88de372/store/store.c#L50-L76

flatcap:
FYI: We tend to recommend **lmdb** as a fast modern database.

cbrownstein:
>> So… does $header_cache_backend really need to be set for header caching? It doesn't seem so.
> 
> You're right. That needs fixing.

I volunteer to update the doc, now that I understand better how NeoMutt uses `$header_cache_backend`.

flatcap:
> I volunteer to update the doc

Brilliant!  Thanks

The docs are auto-generated form multiple sources
(if you want to know how complicated that is, have a look at the [right-hand diagram](https://neomutt.org/dev/build/make))

The two you'll need are:
[...]

PR #3260, opened 2022-02-18
Update documentation wrt. need to set $header_cache_backend

cbrownstein:
* **What does this PR do?**
  This PR updates (corrects) the documentation concerning the need to set `$header_cache_backend` for header caching to work. Reality is that `$header_cache_backend` doesn't need to be set unless the user wants to explicitly specify which backend should be used for header caching.

  Also part of this PR is a small change (26fbd795c0f55782d016507cf33b047afd087b51) to fix a broken link in the GitHub pull request template.

  #3259


flatcap (approved):
Lovely.  Thanks.

PR #3266, opened 2022-02-22
color: fix dupe regex crash

flatcap:
Fix a crash where a `mono` follows a `color` with the same regex, e.g.

```
color body      red green "signature from"
mono  body bold           "signature from"
```

Thanks to:
- Vladimir Zakharov (@z2v)
- Sam Axe (@SamAxe)
- Страхиња Радић (@Strahinja)

### DIFF
--- a/color/regex.c
+++ b/color/regex.c
@@ -412,14 +412,15 @@ bool regex_colors_parse_uncolor(enum ColorId cid, const char *pat, bool uncolor)
 
   if (!pat) // Reset all patterns
   {
-    bool rc = STAILQ_FIRST(cl);
+    if (STAILQ_EMPTY(cl))
+      return true;
 
     mutt_debug(LL_NOTIFY, "NT_COLOR_RESET: [ALL]\n");
     struct EventColor ev_c = { cid, NULL };
     notify_send(ColorsNotify, NT_COLOR, NT_COLOR_RESET, &ev_c);
 
     regex_color_list_clear(cl);
-    return rc;
+    return true;
   }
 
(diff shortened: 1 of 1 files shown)


color/regex.c
```diff
@@ -231,11 +231,14 @@ static enum CommandResult add_pattern(struct RegexColorList *rcl, const char *s,
     // different colours
     if (cc && ((cc->fg != fg) || (cc->bg != bg)))
     {
-      attr_color_clear(&rcol->attr_color);
       cc = curses_color_new(fg, bg);
```
gahr, line 234: Would it be unreasonable to handle unset/default too, so we're sure that this function always return a valid pointer?

flatcap: The `CursesColor` maps to an ncurses colour in the terminal.
We're only allowed `COLORS` (256) of those.
`cc->index` tells us which we've used.

If we allocated empty `CursesColor`s, we'd just make allocating them more complicated.

(I'll add colours to the glossary :-)

gahr: Can't we have a default one we use as a fallback? Or you'll have to check every call.. e.g., what happens at line 284?

flatcap: At L284, we don't care whether there's a colour, or not.

If `ac->curses_color` is non-`NULL`, then we have a colour.
This is tested in quite a few place, notably when we want to merge colours.

Having an empty `CursesColor` would mean something like:
```diff
-  if (ac->curses_color)
+  if (ac->curses_color != CurseColorEmpty)
```

Issue #3219, opened 2022-01-26
mutt html attachments don't persist


wgmayer0:
When I open an attachment in mutt, the attachment opens fine.

Here is my mailcap file:
```
text/html; ( nohup qutebrowser %s > /dev/null 2>&1 & ); nametemplate=%s.html;
text/plain; nohup less %s > /dev/null 2>&1 &
image; ( nohup feh %s > /dev/null 2>&1 & )
audio/video; ( nohup mpv %s > /dev/null 2>&1 & )
application/pdf; ( nohup zathura %s > /dev/null 2>&1 & )
application/octet-stream; ( nohup zathura %s > /dev/null 2>&1 & )
```

When I close mutt, qutebrowser turns completely blank. This behavior is not consistent. *Sometimes* qutebrowser will *not* turn blank. But more times than not it does.

Here is a before and after screenshot:
![before](https://user-images.githubusercontent.com/73801421/151235573-128060e2-d5a7-4b65-bdb0-7f1e96c2e334.png)
![after](https://user-images.githubusercontent.com/73801421/151235591-dcbac579-0ebd-46fa-aaec-c57a43897c20.png)

*Right now it is working, but as soon as I see it not work, I will post some more screenshots*

wgmayer0:
It just happened. I had 3 qutebrowser windows open and they all crashed once I closed neomutt. They just stay blank and blink. I tried opening up another qutebrowser session and it would not open. It also was blank and blinked. Same thing happens on firefox (untested because I cannot easily reproduce). What is causing these applications to crash?

gahr:
NeoMutt will remove temporary files on exit. If you want to persist them, you need to script that on your mailcap entry.

wgmayer0:
Ok what is the best way to script it?

wgmayer0:
Also this does not explain why sometimes it stays and other times it crashes into a blank page.

gahr:
> Also this does not explain why sometimes it stays and other times it crashes into a blank page.

It might be that we don't always manage to delete the file, or the browser sometimes doesn't notice that it's gone or doesn't care, perhaps because you don't change focus or something. Can you please check whether the temporary file is always deleted when you quit neomutt?

> Ok what is the best way to script it?

something like this?

```
text/html; call_browser.sh %s
```

```
% cat call_browser.sh
TMP="$(mktemp)"
cp $1 "$TMP" # it's up to you to get rid of $TMP now...

[...]

PR #3267, opened 2022-02-22
Don't overwrite with zeroes after unlinking the file

jakubkaczor:
* **What does this PR do?**
Removes overwriting file contents with zeroes in the function for unlinking
files.

neomutt/neomutt#3219

### DIFF
--- a/mutt/file.c
+++ b/mutt/file.c
@@ -216,18 +216,8 @@ void mutt_file_unlink(const char *s)
     return;
   }
 
-  FILE *fp = fdopen(fd, "r+");
-  if (fp)
-  {
-    unlink(s);
-    char buf[2048] = { 0 };
-    while (st.st_size > 0)
-    {
-      fwrite(buf, 1, MIN(sizeof(buf), st.st_size), fp);
-      st.st_size -= MIN(sizeof(buf), st.st_size);
-    }
-    mutt_file_fclose(&fp);
-  }
+  unlink(s);
+  close(fd);
 }
 
 /**


gahr:
Opening/closing the file can die too, I guess.

xHire:
BTW, shouldn’t the function also call `close(fd);` at the end?

jakubkaczor:
@xHire I don't think this commit changes the behavior within this regard, does it? I know next to nothing about the programming language C, but I felt this task was simple enough I could do it. 😅 Feel free to argue/edit.

xHire:
@jakubkaczor You’re right, your commit doesn’t change this behaviour. But I thought it would be worth pointing it out while this function is being changed after all those years. :·)

Your changes look good to me.

gahr (approved):

gahr:
@jakubkaczor we noticed that we're leaking the `fd` now. You should `close(fd)` it on all paths.

mutt/file.c
```diff
@@ -216,17 +216,10 @@ void mutt_file_unlink(const char *s)
...
   {
-    unlink(s);
-    char buf[2048] = { 0 };
-    while (st.st_size > 0)
-    {
-      fwrite(buf, 1, MIN(sizeof(buf), st.st_size), fp);
-      st.st_size -= MIN(sizeof(buf), st.st_size);
-    }
-    mutt_file_fclose(&fp);
+    mutt_file_fclose(&fd);
```
gahr, line 222: `mutt_file_fclose` takes a `FILE**`, you have an `int fd` :) you can just `close(fd)` it.

jakubkaczor:
> @jakubkaczor we noticed that we're leaking the `fd` now. You should `close(fd)` it on all paths.

I've made the changes. Is it ok, now?

mutt/file.c
jakubkaczor: Right. Corrected. 😊

gahr:
> > @jakubkaczor we noticed that we're leaking the `fd` now. You should `close(fd)` it on all paths.
> 
> @gahr, I've made the changes. Is it ok, now?

You don't want a condition on `if (fd)`: the file descriptor `fd` is a non-negative integer. Whatever it is, it needs to be closed :)

gahr (approved):
Thanks a lot!

jakubkaczor:
Thank you! 😊

PR #3271, opened 2022-02-25
build: Fix `make update-po`

flatcap:
Two small fixes for `make update-po`

- Fix the build, so that `make update-po` works out-of-tree
- Auto-generate `po/POTFILES` rather than expecting people to remember to update it

### Tests

The two tests should result in `po/*.po` being updated.
 
Normal build:
```sh
./configure
make update-po
```

Out-of-tree build:
```sh
mkdir temp
cd temp
../configure
make update-po
```


Austin-Ray (approved):
LGTM

PR #3282, opened 2022-03-03
Make free'ing a mailbox easier for the caller

### LINKED ISSUE #3098: Review mailbox lifetime APIs
There are several ways to get ahold of a `Mailbox`, including `mx_mbox_find` that only returns an existing instance or `NULL` and `mx_path_resolve` which creates one from scratch if it can't find an existing one. However, there is no clear API to _dispose_ of a `Mailbox`, so we need to hack around it as in #3097 by using the hidden flag as an indication that the mailbox returned by `mx_path_resolve` is a newly created one.

This issue is to think about a better API to make `Mailbox` lifetime management easier.

gahr:
Let's just try and free a mailbox whenever we're done with it, and
leave it to mailbox_free to decide when it's actually safe to do.

In this PoC, safe means the mailbox isn't open and isn't visible. This
can probably cause a few leaks, e.g., when quitting NeoMutt.

Fixes #3098

### DIFF
--- a/browser/browser.c
+++ b/browser/browser.c
@@ -504,7 +504,7 @@ static void add_folder(struct Menu *menu, struct BrowserState *state,
                        const char *name, const char *desc,
                        const struct stat *st, struct Mailbox *m, void *data)
 {
-  if ((!menu || state->is_mailbox_list) && m && (m->flags & MB_HIDDEN))
+  if ((!menu || state->is_mailbox_list) && m && !m->visible)
   {
     return;
   }
--- a/command_parse.c
+++ b/command_parse.c
@@ -601,9 +601,12 @@ enum CommandResult parse_mailboxes(struct Buffer *buf, struct Buffer *s,
       }
     }
 
-    if (!mx_ac_add(a, m))
+    if (mx_ac_add(a, m))
     {
-      //error
+      m->flags = MB_NORMAL; // make it finally visible
+    }
+    else
+    {
       mailbox_free(&m);
       if (new_account)
       {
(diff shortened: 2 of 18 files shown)


flatcap:
Hmm... it's leaking badly, but only with IMAP.

If I open and immediately close, I get **1** leak here:

```
main.c:1373:             main
mx.c:388:                mx_mbox_open
imap/imap.c:1991:        imap_mbox_open
imap/command.c:1141:     imap_cmd_step
imap/command.c:1023:     cmd_handle_untagged
imap/command.c:705:      cmd_parse_lsub
init.c:1054:             mutt_parse_rc_line
init.c:1019:             mutt_parse_rc_buffer
command_parse.c:544:     parse_mailboxes
core/mailbox.c:70:       mailbox_new
```

If I open five different mailboxes, I get a further **220** leaks.
(it's leaking all 44 of my mailboxes, five times)

```
main.c:1391:             main
index/dlg_index.c:1354:  mutt_index_menu
sidebar/functions.c:383: sb_function_dispatcher
sidebar/functions.c:231: op_sidebar_open
index/dlg_index.c:1472:  dlg_change_folder
index/dlg_index.c:659:   change_folder_mailbox
mx.c:388:                mx_mbox_open
imap/imap.c:1991:        imap_mbox_open
imap/command.c:1141:     imap_cmd_step
imap/command.c:1023:     cmd_handle_untagged
imap/command.c:705:      cmd_parse_lsub
init.c:1054:             mutt_parse_rc_line
init.c:1019:             mutt_parse_rc_buffer
command_parse.c:544:     parse_mailboxes
core/mailbox.c:70:       mailbox_new
```

`master` doesn't leak performing the same actions.

flatcap:
My original fix wasn't good enough, so there was a second fix.
I've reverted both in your branch and rebased over `master`.

gahr:
Thanks, I’ll work some more on this

flatcap (approved):

Issue #1294, opened 2018-07-17
racecondition()


RootUp:
Team,

muttlib.c

The above file used `access()` in multiple place such as line number 627, I believe this indicates a security issue, If attacker could change anything along the path between the call `access()` and the files actually used, attacker may exploit a race condition over here.

scottkosty:
Do you see the same issue with mutt? If so, you should probably report it there as well. I checked, and mutt's [muttlib.c](https://gitlab.com/muttmua/mutt/blob/master/muttlib.c) has 3 instances of `access()` compared to 5 instances in neomutt's muttlib.c.

RootUp:
Thanks @flatcap and @scottkosty for taking look into this and yes the issue exits in Mutt as well.

klartext:
Opened a PR for this issue: https://github.com/neomutt/neomutt/pull/3289

It addresses the access()-calls in mutt_check_overwrite().
There is another function below, mutt_set_xdg_path() which also uses access() twice.
But in betnween the bufer is changed and so I don't know what happens there.
Is the second access() call correct? Or should that function also be refactored?
What does mutt_buffer_expand_path() do?

klartext:
The 'fname' is changed between the two calls of access().
```mutt_buffer_concat_path(fname, path, mutt_buffer_string(tmp));```

So I'm not sure this issue holds.

klartext:
Closing, because both access()-calls are needed; see this comment on the closed PR:
https://github.com/neomutt/neomutt/pull/3289#issuecomment-1059942619

PR #3289, opened 2022-03-05
mutt_check_overwrite(): access() only used once now (for issue #1294)

klartext:
Using flags 'file_exists' and 'is_dir' to make intention clear.
Added { } in one place, where readability needs it.

The changes are only in _muttlib.c_  in function _mutt_check_overwrite()_.

This PR was created to address https://github.com/neomutt/neomutt/issues/1294

### DIFF
--- a/muttlib.c
+++ b/muttlib.c
@@ -627,12 +627,24 @@ int mutt_check_overwrite(const char *attname, const char *path, struct Buffer *f
 {
   struct stat st = { 0 };
 
+  bool file_exists = false;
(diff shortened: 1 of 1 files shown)


flatcap:
After a bit more investigation, it seems that we **do** need the second `access()` in `mutt_check_overwrite()`.

klartext:
Overlooked that 'fname' was reused - changed in the meantime - before the second access() call.
It looks like Issue #1294 might be invalid - needs to be checked.

PR #3291, opened 2022-03-08
docs: Improve description of some notmuch variables.

### LINKED ISSUE #3281: Improve documentation for nm_record and nm_record_tags
The [documentation for nm_record](https://neomutt.org/guide/reference.html#nm-record) is a bit scant, and unclear.

I believe it would be more complete (from what I understand), to say something like:"When writing a just-sent message to the Neomutt record, also add it to the notmuch DB. Replies inherit tags from the original messages."

The comment to `nm_record` in the `neomuttrc` example in the [notmuch feature page](https://neomutt.org/feature/notmuch) could also be made to match (it is, IMHO, less clear than the sentence above). 

The [documentation for nm_record_tags](https://neomutt.org/guide/reference.html#nm-record-tags) is equally vague and potentially a bit misleading. It also contains an extra spurious sentence ("When set to 0...") that should be removed.

A better description could be: "Tag modifications to be applied to just-sent messages added to the Neomutt record. See the [documentation for <modify-labels>](https://neomutt.org/feature/custom-tags#table-custom-tags-functions) for the syntax." I'm inferring this part from the `neomuttrc` example in the [notmuch feature page](https://neomutt.org/feature/notmuch), which shows a tag preceded by "-", which implies removal. I admit I haven't tested it.

rionda:
Fixes #3281

### DIFF
--- a/docs/config.c
+++ b/docs/config.c
@@ -2667,21 +2667,27 @@
 { "nm_record", DT_BOOL, false },
 /*
 ** .pp
-** This variable specifies if the NeoMutt record should indexed by notmuch.
+** This variable specifies whether, when writing a just-sent message to the
+** $$record, the message should also be added to the notmuch DB. Replies inherit
+** the notmuch tags from the original message. See $$nm_record_tags for how to
+** modify the set of notmuch tags assigned to sent messages written to the
+** record.
 */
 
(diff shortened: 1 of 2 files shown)


flatcap (approved):
Lovely, thanks!

Issue #3300, opened 2022-03-13
Various <query> and <complete-query> issues


dcpurton:
Here's a test `rc` file exhibiting a couple of issues with `<query>` and `<complete-query>`.

These are all regressions with your recent refactoring changes.

```
# vim: syn=neomuttrc

# filename: query-test.rc

# run from build directory:
#   ./neomutt -n -F query-test.rc

set folder = "./"

set query_command = "echo '%s: 3 matches\na@example.com\tA\texample.com\nb@example.com\tB\texample.com\nc@example.com\tC\texample.com\n'"

# success
macro index <F1> "<query>example<enter><tag-entry><tag-entry><tag-entry><tag-prefix><enter>"

# success
macro index <F2> "<query>example<enter><tag-entry><tag-prefix><enter>"

# success
macro index <F3> "<query>example<enter><enter>"

# failure
# produces    To: example, , C <c@exmaple.com>
# expected    To: A <a@example.com>, B <b@example.com>, C <c@example.com>
macro index <F4> "<mail>example<complete-query><tag-entry><tag-entry><tag-entry><tag-prefix><enter>"

# success
macro index <F5> "<mail>example<complete-query><enter>"

# failure
# produces    To: example, A <a@example.com>
# expected    To: A <a@example.com>
macro index <F6> "<mail>example<complete-query><tag-entry><tag-prefix><enter>"

# segfault
macro index <F7> "<query>example<enter><enter>\cu<enter><query>"

# unrelated failures
# indicator line disappears
macro index <F8> "<query>example<enter><tag-entry><tag-entry><tag-entry>"
macro index <F9> "<mail>example<complete-query><tag-entry><tag-entry><tag-entry>"
```

dcpurton:
GDB backtrace for segfault:

```c
Program received signal SIGSEGV, Segmentation fault.
0x0000000000707ebc in mutt_buffer_get_field (field=0xa10020 <str> "Query: ", buf=0x0, complete=0, 
    multiple=false, m=0x0, files=0x0, numfiles=0x0) at gui/curs_lib.c:310
310	   ret = mutt_enter_string_full(buf->data, buf->dsize, col, complete, multiple,
(gdb) bt
#0  0x0000000000707ebc in mutt_buffer_get_field (field=0xa10020 <str> "Query: ", buf=0x0, complete=0, multiple=false, m=0x0, files=0x0, numfiles=0x0) at gui/curs_lib.c:310
[...]

PR #3301, opened 2022-03-13
query: fix crashes

flatcap:
Fix crashes introduced by recent refactoring.

Tests:

- `<query>`, no matches
- `<query>`, matches
- `<mail>` `<complete>`, no matches
- `<mail>` `<complete>`, matches
- `<query>`, matches, `<query>`, no matches
- `<query>`, matches, `<query>`, matches
- `<mail>` `<complete>`, matches, `<query>`, no matches
- `<mail>` `<complete>`, matches, `<query>`, matches
- `<query>`, matches, `<query-append>`, no matches
- `<query>`, matches, `<query-append>`, matches
- `<mail>` `<complete>`, matches, `<query-append>`, no matches
- `<mail>` `<complete>`, matches, `<query-append>`, matches

Fixes: #3300

### DIFF
--- a/alias/dlg_alias.c
+++ b/alias/dlg_alias.c
@@ -301,15 +301,16 @@ static struct MuttWindow *alias_dialog_new(struct AliasMenuData *mdata)
 
 /**
  * dlg_select_alias - Display a menu of Aliases
- * @param buf    Buffer for expanded aliases
- * @param mdata  Menu data holding Aliases
+ * @param buf   Buffer for expanded aliases
+ * @param mdata Menu data holding Aliases
+ * @retval true Selection was made
  */
-static void dlg_select_alias(struct Buffer *buf, struct AliasMenuData *mdata)
+static bool dlg_select_alias(struct Buffer *buf, struct AliasMenuData *mdata)
 {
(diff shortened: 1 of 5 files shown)


dcpurton:
Looking good. Two things I notice.

`set auto_tag = yes` seems to have no effect. From your comments the other day I was under the impression that I should be able to:

```
<mail>example<complete-query><tag-entry><tag-entry><enter>
```

And both addresses would be inserted if `auto_tag = yes` is in my config.

The other quirk is that If I select an address via `<query>example<enter>` the window drops back to the index (with To: prompt), but if I select an address via `<mail>example<complete-query><enter>` then the window remains showing the alias completion dialog (with To: prompt). It makes sense to me to make them do the same.

flatcap:
> `set auto_tag = yes` seems to have no effect

yep, you're right.  I'll fix that

> ... the window remains showing the alias completion dialog ...

yes.  that's a bit more interesting.

In the case where you see the Index: that's the "correct" behaviour.
The dialog has closed and the Index has become visible again.

In the case where the Alias/Query Dialog is still visible,
(I think) it's because we're missing a repaint of the Index.

I'll see what I can do.

PR #3307, opened 2022-03-17
Use $openssl in hcachever.sh for consistency

pekdon:
* **What does this PR do?**

Small online fix for consistency, actually use the openssl looked up on the line before and that will be returned.


flatcap (approved):
Good spot.  Thanks!

PR #3314, opened 2022-03-20
mutt_parse_charset_hook() / mutt_parse_iconv_hook() added

klartext:
- mutt_parse_charset_hook()
  - mutt_parse_charset_hook() implemented (hook.c)
  - added prototype to hook.h
  - mutt_parse_charset_hook() activated as
    parser for "charset-hook" in mutt_commands.c

- mutt_parse_iconv_hook()
  - mutt_parse_iconv_hook() implemented (hook.c)
  - added prototype to hook.h
  - mutt_parse_iconv_hook() activated as
    parser for "iconv-hook" in mutt_commands.c


klartext:
One case is not handled: too many args.
Should add that also.

How to handle too many args?
- Ignore them?
- write warning message?

flatcap:
> One case is not handled: too many args.

Good point.  Nor did the original code.

For the `set` command, it's possible to chain multiple `set`s together:
`set apple=orange black=white`
but that doesn't make sense for the hooks.
If users really want to chain commands, they can use a `;`

Please check for too many args, there's a message you can reuse.

flatcap:
I think the separate charset and iconv functions is a step too far.
They differ by just the parameter to `mutt_ch_lookup_add()`.
I suggest merging them and using the flag in the `data` param to distinguish them.

hook.c
```diff
@@ -73,6 +74,84 @@ static struct HookList Hooks = TAILQ_HEAD_INITIALIZER(Hooks);
...
+
+  retval += mutt_extract_token(alias, s, MUTT_TOKEN_NO_FLAGS);
+  retval += mutt_extract_token(charset, s, MUTT_TOKEN_NO_FLAGS);
+
+  if (mutt_buffer_is_empty(alias))
+  {
+    mutt_buffer_printf(err, _("%s: too few arguments"), buf->data);
+    rc = MUTT_CMD_WARNING;
+  }
+  else if (mutt_buffer_is_empty(charset))
```
flatcap, line 136: The two `if`s have the same result.
Could we merge them?

hook.c
```diff
@@ -73,6 +74,84 @@ static struct HookList Hooks = TAILQ_HEAD_INITIALIZER(Hooks);
...
+    mutt_buffer_printf(err, _("%s: too few arguments"), buf->data);
+    rc = MUTT_CMD_WARNING;
+  }
+  else if (mutt_ch_lookup_add(MUTT_LOOKUP_ICONV, mutt_buffer_string(alias),
+                              mutt_buffer_string(charset), err))
+  {
+    rc = MUTT_CMD_SUCCESS;
+  }
+
+  rc = (retval == 0) && (rc == MUTT_CMD_SUCCESS) ? MUTT_CMD_SUCCESS : MUTT_CMD_ERROR;
```
flatcap, line 147: Pardon? :-)
Is it possible to make this more legible?

flatcap (changes requested):

klartext:
> I think the separate charset and iconv functions is a step too far. They differ by just the parameter to `mutt_ch_lookup_add()`. I suggest merging them and using the flag in the `data` param to distinguish them.

OK, no problem, now they are one thing.

hook.c
klartext: Heheh. I'm also not happy with this code.

It was partially driven by the confusion I got from not getting an error after calling mutt_extract_token(), even though no more text was available in the buffer.
I just wanted to retain the error message in case of refactoring of the mutt_extract_token(), if it might have shown up as necessary. But maybe I'm not familiar enough with the concept of these parser functions.

The documentation does not clarify it: ```mutt_extract_token - Extract one token from a string```.
```@retval  0 Success``` and ```@retval -1 Error```.

For me extracting a token means to get something back. But if it's not possible, because there is nothing left to be tokenized, I would expect an error of some kind. It seems error for this function means a more radical issue than just not extracting something. Maybe there should be more than two cases: success, parse-error and more-fundamental error (like NULL ptr or such stuff).

The other thing I had questions on, was: when to return ```MUTT_CMD_ERROR``` and when ```MUTT_CMD_WARNING```.
Would an eror from ```mutt_extract_token()``` justify the error case? I woul think so.
But would parse error also justify that? I would rather think the warning case would be enough.
But there might be cases or philosophies that would also vote for the error case here.
For example "fail early": in case of misconfigured rc-file don't start up and insist on correct config.

So a lot of unanswered questions. that lead to that "just in case" code.

Given the behaviour of ```mutt_extract_token()``` as it is, I can rewrite my code and look out for a better solution.

**Comment by klartext** [comment]
Do I need special handling for cases like these (picked from JindraJ's examples in #2686)?

```
charset-hook      "windows-1257" "CP1257"
charset-hook      windows-1258   CP1258
charset-hook      ^us-ascii$     cp1252
charset-hook      ^iso-8859-1$   cp1252
iconv-hook        iso-8859-1     8859-1
```

Double quotes and ```^``` and ```$```, are they processed in the hook-parser, or is this done elsewhere? Are regex's are allowed at all, or would these be cases for a warning? Can I just ignore that?
Or does ```mutt_extract_token()``` handle this already and I can ignore that?

File: hook.c (hunk 2 above)
Comment (flatcap) [reply hook.c]: > not getting an error after calling mutt_extract_token(),
> even though no more text was available in the buffer

Hmm...
I'm not against fixing the behaviour, but it's used in ~90 places, so we'd need a good plan first.

> when to return MUTT_CMD_ERROR and when MUTT_CMD_WARNING

Mutt doesn't have a `mutt_warning()`.
I introduced it do reduce the number of "errors" shown to the user.

- Warnings: When we can tell the user what they did wrong,
  so they can avoid the problem next time

- Errors: When the user couldn't have avoided the problem

> Would an eror from mutt_extract_token() justify the error case?

Probably, yes.
It's most likely triggered if there's a mistake in the code.

**Comment by flatcap** [comment]
> Double quotes and `^` and `$`, are they processed in the hook-parser, or is this done elsewhere?

By default, `mutt_extract_token()` splits out one space-delimited token.
However, it understands quoting, so `"a  b"` is one token.

Once we have a regex in a Buffer, we pass it to `mutt_ch_lookup_add()`,
which **compiles** the regex string into a regex object.

> Are regex's are allowed at all, or would these be cases for a warning?

If you entered an invalid regex, then `mutt_ch_lookup_add()` would return false.
e.g. `charset-hook * utf-8`

This puts a regex-specific error message in `err`.
(This is automatically translated by the system).
We should display this as a warning.

File: hook.c [hunk 3]
```diff
@@ -73,6 +74,58 @@ static struct HookList Hooks = TAILQ_HEAD_INITIALIZER(Hooks);
...
+    mutt_buffer_printf(err, _("%s: too few arguments"), buf->data);
+    rc = MUTT_CMD_WARNING;
+  }
+  else if (mutt_ch_lookup_add(type, mutt_buffer_string(alias),
+                              mutt_buffer_string(charset), err))
+  {
+    rc = MUTT_CMD_SUCCESS;
+  }
+
+  if (MoreArgs(s))
```
flatcap, line 116: We want to check for more args before we call `mutt_ch_lookup_add()`.
Currently, the hook is accepted and we get a warning.

hook.c
```diff
@@ -73,6 +74,58 @@ static struct HookList Hooks = TAILQ_HEAD_INITIALIZER(Hooks);
...
+  {
+    mutt_buffer_printf(err, _("%s: too many arguments"), buf->data);
+    mutt_buffer_reset(s); // clean up buffer to avoid a mess with further rcfile processing
+    rc = MUTT_CMD_WARNING;
+  }
+
```
flatcap, line 122: We could put a `done:` label here

hook.c
```diff
@@ -73,6 +74,58 @@ static struct HookList Hooks = TAILQ_HEAD_INITIALIZER(Hooks);
...
+  retval += mutt_extract_token(alias, s, MUTT_TOKEN_NO_FLAGS);
+  retval += mutt_extract_token(charset, s, MUTT_TOKEN_NO_FLAGS);
+
+  if (retval != 0)
+  {
+    mutt_buffer_pool_release(&alias);
+    mutt_buffer_pool_release(&charset);
+    return MUTT_CMD_ERROR;
```
flatcap, line 102: We could make this tidier by simply `goto done`.
rc is already set and the buffers are freed.

flatcap (changes requested):

hook.c
```diff
@@ -73,6 +74,58 @@ static struct HookList Hooks = TAILQ_HEAD_INITIALIZER(Hooks);
...
+  struct Buffer *charset = mutt_buffer_pool_get();
+
+  int rc = MUTT_CMD_ERROR;
+  int retval = 0;
+
+  enum LookupType type = MUTT_LOOKUP_CHARSET;
+  if (data & MUTT_ICONV_HOOK)
+  {
+    type = MUTT_LOOKUP_ICONV;
+  }
```
gahr, line 93: I like single assignments better, YMMV:

```suggestion
  const enum LookupType type = (data & MUTT_ICONV_HOOK) ? MUTT_LOOKUP_ICONV : MUTT_LOOKUP_CHARSET;
```

flatcap: yours had two conditions in it, gahr's has only one :-)

hook.c
klartext: I called it afterwards, because I added it later - and then supected side effects if calling it before the parsing was ready.

hook.c
klartext: I generally avoid Labels and goto.
```Go To Statement Considered Harmful```

It would be better to allocate the temp buffers from the caller, and give them as parameters to the parsing function.

I think, mutt_parse_hook() could be reestablished as the entry point into the parsing, looks up the data-flags, allocates the temp buffers, calling the sub-parsers with these args added and deallocates the temp buffers.
This would make things cleaner: the sub-paresrs don't need to do the alloaction and deallocation.

**Comment by klartext** [comment]
> I think the separate charset and iconv functions is a step too far. They differ by just the parameter to `mutt_ch_lookup_add()`. I suggest merging them and using the flag in the `data` param to distinguish them.

Merging both functions had the result, that the temporary buffers are now correctly named for one functionality, but bad named for the other functionality.
Names could be switched ot ```tmp1``` and ```tmp2``` which is **not** enhancing readability.

So, saving code for one function made the merged function worse, when considering good naming a good idea.

File: hook.c (hunk 4 above)
Comment (flatcap) [reply hook.c]: Yes, I'm familiar with Dijkstra :-) however...
In C, it's common to reduce complexity by grouping all the cleanup in one place -- at the end.

When we move to C++, life will be simpler
(Don't ask, it's not going to be any time soon).

Comment (klartext) [reply hook.c]: > Yes, I'm familiar with Dijkstra :-) however... In C, it's common to reduce complexity by grouping all the cleanup in one place -- at the end.

1) goto not needed for cleanup at the end
2) if such constructs like the goto's are needed, IMO that's a smell.
3) is it "common" in C? Not sure, this is  a language issue.

See my new commit, how the return value is gathered from the call to the subparser (and the allocation/cleanup is completely out of the sup-barsers - this one and potentially all others that will be implemented in the future).
Doing things similar for the other sub-parsers would make the code much cleaner.
if/else instead of goto.
Even more abstraction would be possible (function pointers...).

> 
> When we move to C++, life will be simpler (Don't ask, it's not going to be any time soon).

I'm completely happy with C. And using, what I proposed could offer enough potential for abstraction, even just with C. No goto needed. And no C++ needed :-)

**Comment by flatcap** [comment]
Thanks for your work on this @klartext, but I've made an executive decision.
Two days and three people is enough for a quick refactoring :-)
There are more important and exciting problems we could be solving.

I've pushed 7ece02994 - flatcap's adequate solution.
It's not perfect; it doesn't satisfy everyone. Sorry.

Unless anyone has any strong (and immediate) objections, I'll merge it in the morning.

Thanks

**Comment by klartext** [comment]
> Thanks for your work on this @klartext, but I've made an executive decision. Two days and three people is enough for a quick refactoring :-) There are more important and exciting problems we could be solving.
> 
> I've pushed [7ece029](https://github.com/neomutt/neomutt/commit/7ece0299497dc52231f67fbf51a01168fb63f403) - flatcap's adequate solution. It's not perfect; it doesn't satisfy everyone. Sorry.
> 
> Unless anyone has any strong (and immediate) objections, I'll merge it in the morning.
> 
> Thanks

Oooops. Did the word "smell" trigger your decision?
Did you looked at 24d4ed541780441afdbe7cd36b9f5f09179c9d96 ?
IMO it's not so far away from being perfect... :-P

**Comment by flatcap** [comment]
> Did you look at 24d4ed5?

Yes and that looked a lot more complicated.
Lots more parameters and it put back the charset code into `mutt_parse_hook()` -- the very thing we we trying to factor out.

**Comment by klartext** [comment]
> > Did you look at [24d4ed5](https://github.com/neomutt/neomutt/commit/24d4ed541780441afdbe7cd36b9f5f09179c9d96)?
> 
> Yes and that looked a lot more complicated. Lots more parameters and it put back the charset code into `mutt_parse_hook()` -- the very thing we were trying to factor out.


- TWO more parameters.
- FOUR function calls less (2 x allocate, 2 x deallocate)
- the big problem with mutt_parse_hook() was it's complicated structure, not that it just exists
- the proposed refactoring was boiling down the complicated structure in mutt_parse_hook() to just flag-to-function-call decisions
- "it put back the charset code" ?????
```
allocate_two_tmp_buffers(); // 2 calls (2 tmp buffers) - not needed in the 9 subparser-functions - 2 calls, not 18
if (flag_x)
 {
   rc = parser_x(...); // the parsing logic goes in the called subparser - it's NOT back in mutt_parse_hook()
 }
else if (flag_y)
{
  rc = parser_y(...); // also no y-specific code here
}
deallocate_the_tmp_buffers(); // 2 calls - not needed in each of the 9 subparser-functions -> 2 calls, not 18
```
replacing the nested if/else-structure from the former ```mutt_parse_hook()```.
```mutt_parse_hook()``` was just a wrapper - and could be replaced later, by providing the tmp Buffers via mutt_commands.c later - just not now (to not change too much at the same time).

flatcap:
> * "it put back the charset code" ?????

Yes, you removed the references to charset from `mutt_parse_hook()`, then 24d4ed541 put them back.
Ideally the charset functions wouldn't even be in `hook.c` -- they have almost nothing in common with the rest of the code there.

The current state of the PR is:
- It's a simple function
- It works
- The complexity of `mutt_parse_hook()` is reduced

That's good enough for me, for now.
I can't afford to spend any more time debating this one function.

flatcap (approved):

PR #3316, opened 2022-03-23
Do not insert an actual key mapping for noop

### LINKED ISSUE #3288: Characters in a sequence of a macro are "eaten" in the definition of another macro
## Expected Behaviour

Use a macro that adds a notmuch tag "starred" to a message, while also having a macro whose sequence is `dd`.

## Actual Behaviour

The tag "starre" is added instead, and other 'd' characters in the definition of the macro are "eaten".

## Steps to Reproduce

Have the following in your `.muttrc`:
```
# We use d as prefix for the dd "trash tag toggle" macro
bind index d   noop
# Macro to star/unstar
folder-hook somefolder "unmacro * S"
folder-hook somefolder "macro index,pager S \"<modify-labels>!starred<enter><sync-mailbox><enter-command>echo 'Toggled starred tag'<enter>\""
# Macro to move message to/from trash
folder-hook somefolder "unmacro * dd"
folder-hook somefolder "macro index,pager dd \"<modify-labels>!trash<enter><sync-mailbox><enter-command>echo 'Toggled trash tag'<enter>\""
```

Try using the `S` macro. See how the `starre` tag is added and the message `Toggle starre tag` is printed: the 'd's were eaten.

If one comments out the definition of the `dd` macro, everything works.

The order of the macro definitions is irrelevant.

- Always

- When I changed my config
  Can you narrow down **what** you changed?

  I added the above excerpt.

## NeoMutt Version

```
NeoMutt 20211029

System: Darwin 21.3.0 (x86_64)
libiconv: 1.16
libidn2: 2.3.2 (compiled with 2.3.2)
GPGME: 1.16.0-unknown
OpenSSL: OpenSSL 3.0.1 14 Dec 2021
libnotmuch: 5.5.0
storage: kyotocabinet
compression: zlib

Configure options: --prefix=/opt/local --enable-doc --disable-idn --with-ncurses=/opt/local --with-nls=/opt/local --with-ssl=/opt/local --with-zlib=/opt/local --with-gpgme=/opt/local --with-idn2=/opt/local --with-kyotocabinet=/opt/local --with-notmuch=/opt/local --with-sasl=/opt/local

Compilation CFLAGS: -pipe -Os -isysroot/Library/Developer/CommandLineTools/SDKs/MacOSX12.sdk -arch x86_64 -std=c99 -D_ALL_SOURCE=1 -D_GNU_SOURCE=1 -D__EXTENSIONS__ -I/opt/local/include -DNCURSES_WIDECHAR

Default options:

Compile options:
  -autocrypt +fcntl -flock -fmemopen +futimens +getaddrinfo -gnutls +gpgme -gss
  +hcache -homespool +idn -inotify -locales_hack -lua -mixmaster +nls +notmuch

MAILPATH="/var/mail"
PKGDATADIR="/opt/local/share/neomutt"
SENDMAIL="/usr/sbin/sendmail"
SYSCONFDIR="/opt/local/etc"

```

## Extra Info

* Operating System and its version: MacOS Monterey
* Were you using 'screen' or 'tmux'? tmux
* Is your email local (maildir) or remote (IMAP)? maildir

gahr:
`unmacro * dd` adds a binding from `dd` to `OP_NULL` (i.e., `noop`) to
all menus, including the editor. This `dd` mapping will keep matching
text typed into the editor, so if we type `abcde` the `d` will be
matched and eaten, resulting in `abce`.

This is a quick fix that solves the `unmacro *` case of #3288, but the
problem is more profound. If we want the editor to jump to the
begin-of-line on `bl`, we might try to `macro editor bl <bol>`. This
works, but this also eats any `b` we type in the editor. The editor is
special in this regard, because unlike other dialogs we want typed keys
to actually appear. ~I don't have a solution for this yet.~ 
Edit: I think [26d8c59](https://github.com/neomutt/neomutt/pull/3316/commits/26d8c5901b0241c979633ab66bc7ae85f8aa7dcb) might be a reasonable fix for this.
Edit2: this bug (eating all but the last char of a non-matching sequence) exists upstream too.  

Fixes #3288

### DIFF
--- a/keymap.c
+++ b/keymap.c
@@ -613,12 +613,13 @@ static void generic_tokenize_push_string(char *s, void (*generic_push)(int, int)
  */
 static int retry_generic(enum MenuType mtype, keycode_t *keys, int keyslen, int lastkey)
 {
+  if (lastkey)
+    mutt_unget_event(lastkey, 0);
+  for (; keyslen; keyslen--)
+    mutt_unget_event(keys[keyslen - 1], 0);
+
   if ((mtype != MENU_EDITOR) && (mtype != MENU_GENERIC) && (mtype != MENU_PAGER))
   {
-    if (lastkey)
-      mutt_unget_event(lastkey, 0);
-    for (; keyslen; keyslen--)
-      mutt_unget_event(keys[keyslen - 1], 0);
     return km_dokey(MENU_GENERIC);
   }
   if (mtype != MENU_EDITOR)
@@ -626,6 +627,8 @@ static int retry_generic(enum MenuType mtype, keycode_t *keys, int keyslen, int 
     /* probably a good idea to flush input here so we can abort macros */
     mutt_flushinp();
   }
+
+  LastKey = mutt_getch().ch;
   return OP_NULL;
 }
 
--- a/keymap.c
+++ b/keymap.c
@@ -426,17 +426,24 @@ static enum CommandResult km_bind_err(const char *s, enum MenuType mtype, int op
     }
   }
 
-  if (last) /* if queue has at least one entry */
+  if (map->op == OP_NULL)
   {
-    if (STAILQ_NEXT(last, entries))
-      STAILQ_INSERT_AFTER(&Keymaps[mtype], last, map, entries);
-    else /* last entry in the queue */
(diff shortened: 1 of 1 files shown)


flatcap (approved):
Very nice!

PR #3318, opened 2022-03-23
Get rid of the LastKey global variable

gahr:


flatcap:
**Aside**:
Whilst stripping down `struct Menu`, I've noticed that `Menu.type` is barely used.

The main users are to do with key handling (~13 places) and they happen in the dialogs' event loops:
- `op = km_dokey(menu->type);`
- `km_error_key(menu->type);`

The two exceptions handled by Menu are `search` and `op_help`.

flatcap (approved):

PR #3325, opened 2022-03-28
sidebar: fix selection/highlight position

flatcap:
Using `<change-folder>` to a non-visible mailbox would leave the sidebar with the wrong selection.

- b8c37df64 sidebar: fix highlight position
  If the user changes to a non-visible mailbox, clear the selection but leave the highlight where it was.

- 78fa452b0 drop: `NT_MAILBOX_SWITCH`
  This is now handled by `NT_INDEX`


gahr (approved):
Works just nicely, thanks!

Issue #3328, opened 2022-03-29
Add functionality to dump the screen


gahr:
ncurses has functionality to dump a window to file (and restore it later), see [man pages](https://invisible-island.net/ncurses/man/curs_util.3x.html#h3-putwin_getwin).
It could be nice to implement a NeoMutt function to dump the current screen to file.
As a first step, we might use it to test functionalities by comparing screens (e.g., compare the dumps of `<next-entry><next-entry><display-message>` and `<display-message><next-entry><next-entry>` to see that navigating the index works both in the index and the pager).
Next, we could implement a framework to test for particular features we expect on the screen by parsing the dumps.

flatcap:
Perhaps:
```
./configure --debug-screen
```

That'd enable a generic function `<write-screen>FILENAME<enter>`

I initially thought about an automatic timestamp filename, but that'd make it hard for the caller to use.

---

That might enable us to do a form of graphical coverage testing.

We could create a set of macros / config files that would navigate to every possible dialog and dump the contents.

flatcap:
I've been thinking some more about coverage.
It'd be good to dump every dialog (possibly in every configuration), but I'm starting to realise that'll be a LOT of tests.

Besides checking the visuals, it might be worth turning on **code** coverage as well.
I've no idea how much code will be covered by just _navigating_ to every dialog.

In [[flatcap/dump]](https://github.com/neomutt/neomutt/tree/flatcap/dump), I've started a [list of dialogs](https://github.com/neomutt/neomutt/blob/flatcap/dump/dialogs.txt).  (There are a few dupes, atm)

I've listed each caller of the Pager separately.
Do we want to consider the effects of `$help`, `$status_on_top`, `$pager_index_lines`, `$sidebar_visible`, `$sidebar_on_right`, etc

PR #3332, opened 2022-03-31
Implement dump-screen

gahr:
This introduces a new `dump-screen` function that dumps the content of the
screen in a ncurses-specific way into a file. The file is controlled
by the config `dump_screen_file` and is rotated each time `dump-screen`
is called.

For now, I have set a maximum of 100 rotations. We can decide to
configure it or we can implement unlimited rotation too.

Implements #3328


mutt_logging.c
```diff
@@ -68,45 +68,6 @@ static void error_pause(void)
...
-  struct Buffer *new_file = mutt_buffer_pool_get();
-
-  /* rotate the old debug logs */
-  for (count -= 2; count >= 0; count--)
-  {
-    mutt_buffer_printf(old_file, "%s%d", file, count);
-    mutt_buffer_printf(new_file, "%s%d", file, count + 1);
-
-    mutt_buffer_expand_path(old_file);
-    mutt_buffer_expand_path(new_file);
```
gahr, line 99: I can't think of a reason why we're expanding these files each time.

flatcap: Hmm... anything to do with the `~` in the default `$debug_file`?

gahr: Sure, but this doesn't change as we increment a counter to it

PR #3337, opened 2022-04-05
Index: refactor resolve

flatcap:
Refactor all the resolve code (advancing the cursor after a function).

- 8623759f1 create resolve_email()
- 29a43813b index: refactor resolve

This reduces the code duplication and gives us a central point to send notifications.

### DIFF
--- a/index/functions.c
+++ b/index/functions.c
@@ -90,6 +90,65 @@ static const char *Not_available_in_this_menu =
     N_("Not available in this menu");
 
 /**
+ * enum ResolveMethod - How to advance the cursor
+ */
+enum ResolveMethod
+{
+  RESOLVE_NEXT_EMAIL,     ///< Next email, whatever its state
+  RESOLVE_NEXT_UNDELETED, ///< Next undeleted email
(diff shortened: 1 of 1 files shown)


index/functions.h
```diff
@@ -50,7 +51,19 @@ struct IndexFunction
...
+enum ResolveMethod
+{
+  RESOLVE_NEXT_EMAIL,        ///< Next email, whatever its state
+  RESOLVE_NEXT_UNDELETED,    ///< Next undeleted email
+  RESOLVE_NEXT_THREAD,       ///< Next top-level thread
+  RESOLVE_NEXT_SUBTHREAD,    ///< Next sibling sub-thread
+};
+
 int index_function_dispatcher(struct MuttWindow *win, int op);
+bool resolve_email(struct Menu *menu, struct IndexSharedData *shared, enum ResolveMethod rm);
```
gahr, line 66: Do you plan on using this from outside index/functions.c?

flatcap: Good point.
I thought I might need it in the Postponed, but I'll probably upgrade that to use the index function dispatcher.

I'll internalise `resolve_email()`.

PR #3351, opened 2022-04-11
Fix forwarding text attachments when honor_disposition is set

### LINKED ISSUE #3346: text/plain attachments not forwarded with `honor_disposition` set
## Expected Behaviour

With `honor_disposition` set, text/plain attachments will be forwarded.

## Actual Behaviour

text/plain attachments are not attached to the forwarded message (other MIME types are).

## Steps to Reproduce

With the following `neomuttrc` try to forward a message which has attachments of text/plain MIME type.
```neomuttrc
set imap_user = username
set imap_pass = password
set folder = "imaps://imap.example.net"
set honor_disposition
```
Works as expected with `mutt`.

Always

I have only tested this with the current version of `neomutt`.

## NeoMutt Version

```
NeoMutt 20220408
Copyright (C) 1996-2022 Michael R. Elkins and others.

System: Linux 5.15.32_1 (x86_64)
libidn: 1.38 (compiled with 1.38)
GPGME: 1.15.1
OpenSSL: OpenSSL 1.1.1n  15 Mar 2022
libnotmuch: 5.6.0
storage: rocksdb, gdbm, tdb
compression: zlib

Configure options: --ssl --gpgme --notmuch --gdbm --lua --sasl --zlib --tdb --rocksdb --testing

Compilation CFLAGS: -fstack-clash-protection -D_FORTIFY_SOURCE=2 -mtune=generic -O2 -pipe  -D_FILE_OFFSET_BITS=64 -DLARGEFILE_SOURCE=1  -fdebug-prefix-map=/builddir/neomutt-20220408=. -std=c99 -fno-delete-null-pointer-checks -D_ALL_SOURCE=1 -D_GNU_SOURCE=1 -D__EXTENSIONS__ -I/usr/include -I/usr/include/lua5.4 -DNCURSES_WIDECHAR

Default options:

Compile options:
  -autocrypt +fcntl -flock -fmemopen +futimens +getaddrinfo -gnutls +gpgme -gss 
  +hcache -homespool +idn +inotify -locales_hack +lua -mixmaster +nls +notmuch 

MAILPATH="/var/mail"
SENDMAIL="/usr/sbin/sendmail"
SYSCONFDIR="/etc"

```

## Extra Info

I'm running `neomutt` on Void Linux with musl libc.

gahr:
`mutt_can_decode` is used in `inline_forward_attachments` to decide
which parts of a message need to be attached when forwarding an email.
If we don't consider skipped types such as signatures, this set must be
equal to the set of parts that wasn't included in the email text.
However, using `mutt_can_decode` is not enough, because parts with
disposition = attachments that could be decoded might have been skipped
during the formation of the email text because of `honor_disposition`.
Hence, we need a common piece of code to decide if a part is handled
inline or in an attachment.

An alternative to this would be to mark the parts that were handled in
the main body and include all others. I am not sure what is the best
approach.

While at it, use `b` for bodies.

Also, please double check my logic here.

Fixes #3346

### DIFF
--- a/handler.c
+++ b/handler.c
@@ -1727,12 +1727,7 @@ int mutt_body_handler(struct Body *b, struct State *s)
     }
   }
 
-  const bool c_honor_disposition =
-      cs_subset_bool(NeoMutt->sub, "honor_disposition");
-  /* only respect disposition == attachment if we're not
-   * displaying from the attachment menu (i.e. pager) */
-  if ((!c_honor_disposition || ((b->disposition != DISP_ATTACH) || OptViewAttach)) &&
-      (plaintext || handler))
+  if ((plaintext || handler) && !mutt_prefer_as_attachment(b))
   {
     /* Prevent encrypted attachments from being included in replies
      * unless $include_encrypted is set. */
@@ -1750,6 +1745,8 @@ int mutt_body_handler(struct Body *b, struct State *s)
    * if we're not already being called from there */
   else if (s->flags & MUTT_DISPLAY)
   {
+    const bool c_honor_disposition =
+        cs_subset_bool(NeoMutt->sub, "honor_disposition");
     struct Buffer msg = mutt_buffer_make(256);
 
     if (!OptViewAttach)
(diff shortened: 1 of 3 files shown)


flatcap:
> While at it, use `b` for bodies.

Woo!

flatcap (approved):

PR #3353, opened 2022-04-12
pager: fix $pager_stop

flatcap:
Fix the `$pager_stop` behaviour.

If `pager_stop=no`, then using `<next-page>` or `<half-down>` at the end of an email will jump to the next undeleted email.

This was broken by the Index/Pager refactoring.


gahr:
> If pager_stop=yes, then using <next-page> or <half-down> at the end of an email will jump to the next undeleted email.

ITYM pager_stop=no?

gahr (approved):

PR #3354, opened 2022-04-12
color: disable mono command

flatcap:
Historically the `mono` and `unmono` commands were only enabled if:
- Mutt was compiled **without** support for colour,
- Your terminal didn't support colour.

Both these conditions are very unlikely, these days.

The `color` command can take attributes in addition to the colours, e.g.

```
color body bold underline red default "neomutt"
```

During the refactoring, I made the decision to turn `mono` commands into a stripped down `color` command.
This would have allowed the user to tweak the attributes of an object, without having to specify the colour info.

```
# Equivalent
color body bold default default "neomutt"
mono body bold "neomutt"
```

Unfortunately, this means that all those `mono` commands that were laying dormant in users' config files have started interfering with their themes.
(including our own `/etc/neomuttrc`)

Conclusion.  Kill `mono` and encourage users to drop it from their config.

### DIFF
--- a/color/command.c
+++ b/color/command.c
@@ -678,12 +678,8 @@ enum CommandResult mutt_parse_uncolor(struct Buffer *buf, struct Buffer *s,
 enum CommandResult mutt_parse_unmono(struct Buffer *buf, struct Buffer *s,
                                      intptr_t data, struct Buffer *err)
 {
-  if (OptNoCurses)
-  {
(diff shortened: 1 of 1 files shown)


gahr:
Nice. I think we should doc it

gahr (approved):

flatcap:
Great.

> Nice. I think we should doc it

Yes, I'll add that to this PR.

PR #3372, opened 2022-04-17
Use DT_SLIST for charset variables

elshize:
* **What does this PR do?**

Changes charset config type to DT_SLIST

#2066


docs/config.c
```diff
@@ -191,7 +191,7 @@
 */
 #endif
 
-{ "assumed_charset", DT_STRING, 0 },
+{ "assumed_charset", DT_SLIST|SLIST_SEP_COLON|SLIST_ALLOW_EMPTY, 0 },
```
flatcap, line 194:
```suggestion
{ "assumed_charset", DT_SLIST, 0 },
```

The docs only need the base type, for now.

mutt/charset.c
```diff
@@ -314,16 +314,13 @@ int mutt_ch_convert_nonmime_string(char **ps)
   if (ulen == 0)
     return 0;
 
-  const char *c1 = NULL;
-
-  const char *const c_assumed_charset = cs_subset_string(NeoMutt->sub, "assumed_charset");
+  const struct Slist *const c_assumed_charset = cs_subset_slist(NeoMutt->sub, "assumed_charset");
   const char *const c_charset = cs_subset_string(NeoMutt->sub, "charset");
-  for (const char *c = c_assumed_charset; c; c = c1 ? c1 + 1 : 0)
+  const struct ListNode *np;
```
flatcap, line 319:
```suggestion
  const struct ListNode *np = NULL;
```

Please initialise all pointers.

mutt_config.c
```diff
@@ -164,7 +164,7 @@ static struct ConfigDef MainVars[] = {
   { "ask_cc", DT_BOOL, false, 0, NULL,
     "Ask the user for the carbon-copy recipients"
   },
-  { "assumed_charset", DT_STRING, 0, 0, charset_validator,
+  { "assumed_charset", DT_SLIST|SLIST_SEP_COLON|SLIST_ALLOW_EMPTY, 0, 0, charset_validator,
```
flatcap, line 167: As you mentioned on IRC, the validator needs updating.
Calling _this_ validator will probably crash (it'll be passed `struct Slist*`).
Let's set it to `NULL`, for now.

```suggestion
  { "assumed_charset", DT_SLIST|SLIST_SEP_COLON|SLIST_ALLOW_EMPTY, 0, 0, NULL,
```

flatcap (changes requested):
Very nice!
I've left a few small suggestions.
Also, please can you fix the typo in the commit message `s/TD_SLIST/DT_SLIST/` :-)
Thanks

elshize:
Thanks for the pointers! I'll fix it and continue with the rest this week.

elshize:
@flatcap could you explain why `charset_validator` returns `CSR_SUCCESS` if `value == 0`? What's the semantic meaning? Does it mean that the value is not set, and because it's not mandatory, then it is valid? Would it be the same for `Slist` case?

elshize:
I addressed your comments, and had a go at implementing the Slist validator.

I tested it locally, but I'd also want to write some unit tests for that. But let me know if the validator makes sense.

flatcap:
Yes.

NeoMutt stores empty config values as `NULL`, where possible.

> returns `CSR_SUCCESS` if `value == 0`? What's the semantic meaning?
> Does it mean that the value is not set, and because it's not mandatory, then it is valid?

Exactly.
If we set the `DT_NOT_EMPTY` flag, then the core validator code will reject the value before it gets to our validator.

> Would it be the same for Slist case?

Yes.  I've documented the details in the `@page` comment in each config type file:

https://github.com/neomutt/neomutt/blob/d88d7b78338d77858d4e2a42751ec17176b0cafa/config/slist.c#L26-L32

And you can see the check in the `set` function:

https://github.com/neomutt/neomutt/blob/d88d7b78338d77858d4e2a42751ec17176b0cafa/config/slist.c#L67-L69

---

Historically, Mutt's `parse_set()` had lots of special cases for _this_ variable or _that_.
Where there are common cases, NeoMutt has flags, e.g. `DT_NOT_NEGATIVE`.
If that's not enough we have validator functions.

flatcap:
> had a go at implementing the Slist validator

I'll take a look

> I'd also want to write some unit tests for that

That'd be great!

config/charset.c
```diff
@@ -70,3 +70,39 @@ int charset_validator(const struct ConfigSet *cs, const struct ConfigDef *cdef,
...
+  if (value == 0)
+    return CSR_SUCCESS;
+
+  const struct Slist *list = (const struct Slist *) value;
+
+  if ((cdef->type & DT_CHARSET_SINGLE) && list->count > 1)
+  {
+    mutt_buffer_printf(err, _("'charset' must contain exactly one character set name"));
+    return CSR_ERR_INVALID;
+  }
```
flatcap, line 90: We'll probably only apply the `DT_CHARSET_SINGLE` to variables of type `DT_STRING`.
So we probably don't need this section.

flatcap (changes requested):
Validator looks good

config/charset.c
elshize: Right, that makes sense, if it's single, it probably shouldn't be a list.

elshize:
Ok, I added some test. Let me know what you think.

Also, do you prefer merging this one change or do you want me to continue with other options in the same PR? It's all the same to me.

flatcap:
> Ok, I added some tests. Let me know what you think.

They look OK.  The tests always look a lot more complicated than they are.

My only query is "grapple"?  **GRAPPLE**???
Haven't you learnt the list of [approved fruits](https://github.com/neomutt/sample-data/blob/main/fruit.txt) ;-)

> Also, do you prefer merging this one change
> or do you want me to continue with other options in the same PR?

Let's continue in this PR, for now.  One commit per config.

In fact, if you're happy to mess with the history, I suggest:
splitting out `charset_validator_slist()` and its tests into a separate commit and putting it first in the PR.

elshize:
> My only query is "grapple"? GRAPPLE???
> Haven't you learnt the list of [approved fruits](https://github.com/neomutt/sample-data/blob/main/fruit.txt) ;-)

Lol, no, but now I know :joy: 

> In fact, if you're happy to mess with the history, I suggest:
splitting out charset_validator_slist() and its tests into a separate commit and putting it first in the PR.

Sure, I can do that. I'll make a commit for the validator and its test, then one per updated variable.

Here's a question: for the `charset` variable, that is single value, should that just be left as string then? If so, then we can also eventually change its validation to only check if the value is correct charset, right?

flatcap:
> for the `charset` variable, that is single value, should that just be left as string then?

Yes

> we can also eventually change its validation to only check if the value is correct charset, right?

Yes

`grep`ping the code, I see that `$charset` is the only variable using `DT_CHARSET_SINGLE`.
We could probably drop that and just comment in `charset_validator()`.

flatcap:
@elshize 
Good spot on IRC:

- definitely **multiple** values: `$assumed_charset`, `$attach_charset`, `$send_charset`
- definitely **single** values: `$charset`, `$config_charset`, `$newsgroups_charset`

elshize:
@flatcap so here are my concerned about `$attach_charset` I mentioned in IRC.

In this function here: https://github.com/neomutt/neomutt/blob/main/send/sendlib.c#L379 to quote the comment there:

> if fromcode is zero then fromcodes is assumed to be the name of a
> single charset even if it contains a colon.

Which, looking at the code, seems correct; there's a check and if `!fromcode`, then we're assuming one charset.

Now, the only place I found this function being called is here: https://github.com/neomutt/neomutt/blob/main/send/sendlib.c#L514 and here `fromcode` is not set anywhere or passed anywhere after it is initialized to `NULL` here: https://github.com/neomutt/neomutt/blob/main/send/sendlib.c#L471 This means that `fromcodes` passed to this function will always be treated as a single charset. And under certain conditions, we are passing `c_attach_charset` to `convert_file_from_to` as `fromcodes`. Which in turn means that we're assuming that `$attach_charset` is one charset.

This also seems to be the only place where `$attach_charset` is used in the entire project.

I haven't had the time to really dig into the logic of these functions, but judging by what I mention above, it looks like either man pages + variable definition are wrong, or the implementation is incorrect.

Let me know if I'm missing something here. I should have some time to dig deeper and write some code this weekend.

flatcap:
> there's a check and if `!fromcode`, then we're assuming one charset

Nope, but it took me a little while to spot why not.
`char **from` (is a double pointer, the **address** of `fromcode` in `mutt_get_content_info()`.
aka, an "out-param".  This is the result of the function.

So... in `mutt_get_content_info()`

https://github.com/neomutt/neomutt/blob/7df877bf67f8eb8e33473fad380846436943b143/send/sendlib.c#L471

we pass the address of this variable into `convert_file_from_to()`:

https://github.com/neomutt/neomutt/blob/7df877bf67f8eb8e33473fad380846436943b143/send/sendlib.c#L511-L515

Therefore, in `convert_file_from_to()`, this is true:

https://github.com/neomutt/neomutt/blob/7df877bf67f8eb8e33473fad380846436943b143/send/sendlib.c#L412

`fromcode` is an **address**, `*fromcode` is `NULL`

so, we parse our way through `fromcodes` (plural).
And here, we store the "best match":

https://github.com/neomutt/neomutt/blob/7df877bf67f8eb8e33473fad380846436943b143/send/sendlib.c#L428

Back  in `mutt_get_content_info()`, the result gets used here:

https://github.com/neomutt/neomutt/blob/7df877bf67f8eb8e33473fad380846436943b143/send/sendlib.c#L524

elshize:
Ah, yes! Completely missed the double pointer there. Thanks for having a look, it makes sense now.

convert/convert.h
```diff
@@ -0,0 +1,38 @@
+/**
+ * @file
```
flatcap, line 2: Please rename this to `convert/lib.h` -- all libraries have a single header file as an entry point

convert/convert.h
```diff
@@ -0,0 +1,38 @@
...
+ *
+ * This program is distributed in the hope that it will be useful, but WITHOUT
+ * ANY WARRANTY; without even the implied warranty of MERCHANTABILITY or FITNESS
+ * FOR A PARTICULAR PURPOSE.  See the GNU General Public License for more
+ * details.
+ *
+ * You should have received a copy of the GNU General Public License along with
+ * this program.  If not, see <http://www.gnu.org/licenses/>.
+ */
+
```
flatcap, line 22: Add a `@page lib_convert` here (cut'n'paste from other `lib.h`)

You'll also need to as `lib_convert` to the list of libraries in `main.c`

convert/convert.h
```diff
@@ -0,0 +1,38 @@
...
+#ifndef MUTT_LIB_CONVERT_H
+#define MUTT_LIB_CONVERT_H
+
+#include <wchar.h>
+
+#include "config/subset.h"
+#include "email/content.h"
+#include "email/body.h"
```
flatcap, line 30: We include all libraries using their `lib.h` rather than cherry-picking

email/content.h
```diff
@@ -45,4 +46,16 @@ struct Content
...
+/**
+ * struct ContentState - Info about the body of an email
+ */
+struct ContentState
+{
+  bool from;
+  int whitespace;
+  bool dot;
+  int linelen;
+  bool was_cr;
```
flatcap, line 58: This has just been moved not created, so for **optional** extra brownie points, please document the members :heart:

mutt/queue.h
```diff
@@ -354,6 +354,12 @@ struct {								\
 	   (var);							\
 	   (var) = STAILQ_NEXT((var), field))
 
+#define	STAILQ_FOREACH_ENUMERATE(var, index, head, field) \
```
flatcap, line 357: Hmm... now this is tricky.  `mutt/queue.h` isn't ours to alter.
It's a standard used by Linux and the BSDs.
It's a nice invention, but for now I'd prefer to use `STAILQ_FOREACH()` and a simple `index++`

test/convert/mutt_get_content_info.c
```diff
@@ -0,0 +1,109 @@
...
+  { NULL },
+  // clang-format on
+};
+
+void test_mutt_get_content_info(void)
+{
+  // struct Content *mutt_get_content_info(const char *fname, struct Body *b, struct ConfigSubset *sub);
+
+  char *fname = malloc(L_tmpnam);
+  TEST_CHECK(tmpnam_r(fname) != NULL);
```
flatcap, line 62: Sorry, FreeBSD isn't happy with `tmpnam_r()`
Our `mutt_file_mkstemp()` function unlinks the temp file, but you could cut'n'paste the rest

flatcap (changes requested):

test/convert/mutt_get_content_info.c
elshize: Oops, missed that failure.

mutt/queue.h
elshize: Ahh, I was not aware of that! Will change.

email/content.h
elshize: Good point, will document.

Makefile.autosetup
```diff
@@ -706,6 +706,19 @@ $(LIBQUESTION): $(PWD)/question $(LIBQUESTIONOBJS)
...
 	$(MKDIR_P) $(PWD)/question
 
+###############################################################################
+# libconvert
+LIBCONVERT=	libconvert.a
+LIBCONVERTOBJS=convert/convert.o
+CLEANFILES+=	$(LIBCONVERT) $(LIBCONVERTOBJS)
+ALLOBJS+=	$(LIBCONVERTOBJS)
+
+$(LIBCONVERT): $(PWD)/progress $(LIBCONVERTOBJS)
```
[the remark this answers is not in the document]
elshize: @flatcap where is the typo? what am I missing?

convert/convert.c
```diff
@@ -0,0 +1,463 @@
+#include <errno.h>
```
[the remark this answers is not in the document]
elshize: What should I put in `@authors` section? What year (current? range?), what name(s)?

convert/convert.h
elshize: If there is only one file `lib.c` in the lib, do I omit the file table under `@page`?

Makefile.autosetup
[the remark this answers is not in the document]
flatcap: "progress" in the "convert" section
(result of cut'n'paste)

convert/convert.c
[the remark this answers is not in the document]
flatcap: put your own name, 2022, for now -- it's always a dilemma what to put in refactored files

Makefile.autosetup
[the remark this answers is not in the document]
elshize: ahh, right, thanks!

convert/convert.h
flatcap: I keep meaning to docment this...

- Library XYZ
- `main.c` -- `@ref lib_xyz`
- `xyz/*.c` -- `@page xyz_foo`, `@page xyz_bar`, ...
- `xyz/lib.h` -- table `| xyz/foo.c | @subpage xyz_foo`, ... (**Note**: **sub**page)

That links everything up, so Doxygen can generate the ["Libraries" tab](https://neomutt.org/code/).

test/convert/mutt_get_content_info.c
elshize: So I actually just changed this to `tmpnam()` which is POSIX (while `tmpnam_r` isn't), and the difference as far as I can tell is only when passing `NULL` to the function, which I'm not doing.

Anyway, the builds are passing, so I think that should be fine.

flatcap (approved):
Very thorough job!  Thanks!

Issue #3374, opened 2022-04-19
Bulk flags manipulation in mbox make messages to duplicate


jindraj:
## Expected Behaviour

Just work

## Actual Behaviour

When messages are marked for deletion, messages reappear with some duplicates as a bonus. Sometimes several are deleted, but example below with just changing the flags makes duplicates.

## Steps to Reproduce
1. mbox file with some initial messages
```
neomutt -n -F /dev/null  -f MBOX -e "set use_threads = reverse" -e 'push "<tag-pattern>.<enter><tag-prefix><clear-flag>o<sync-mailbox><quit>"'
```
## How often does this happen?
This is the first time I encountered this behaviour. Here is the confirmed minimum config: (thanks Richard for eliminating it to a single line)
```
set use_threads         = reverse
```

## When did it start to happen?
bisected by @flatcap to 0874db3cb

I don't usually work with mbox mailboxes. In the beginning there were a lot of messages (one per minute) in my system mailbox. Every message was there only once and with unique message-id - I believe, since it was sent by postfix. After Tag all, Delete, Sync mailbox I've got duplicates, that drove me crazy. I repeated the procedure and there they was again with more and more duplicates. Turned me in into berserk mode. 

## NeoMutt Version

```
NeoMutt 20220415-15-7df877
Copyright (C) 1996-2022 Michael R. Elkins and others.
NeoMutt comes with ABSOLUTELY NO WARRANTY; for details type 'neomutt -vv'.
NeoMutt is free software, and you are welcome to redistribute it
under certain conditions; type 'neomutt -vv' for details.

System: Darwin 21.2.0 (arm64)
ncurses: ncurses 6.3.20211021 (compiled with 6.3.20211021)
libiconv: 1.11
libidn2: 2.3.2 (compiled with 2.3.2)
GPGME: 1.17.1
OpenSSL: OpenSSL 3.0.2 15 Mar 2022
libnotmuch: 5.6.0
storage: lmdb
compression: zstd

[...]

PR #3381, opened 2022-04-22
mbox: fix sync duplicates

flatcap:
Running <sync-mailbox> when use_threads=reverse caused duplicate emails.

When you sync a mailbox, the backend, in this case `mbox_mbox_sync()`,
sorts the mailbox into its original order before writing it to disk.

The fix, is to ensure that both `$use_threads` and `$sort` are reset before sorting.

Fixes: #3374 

---

**Steps**:
- Create an mbox file with a few emails in it
- Open the mailbox `neomutt -n -F /dev/null -f MBOX -e 'set use_threads = reverse'`
- Flag some emails as new
- Quit NeoMutt
- Re-run NeoMutt

**Results**:
- Mailbox has the same number of emails as before
- The emails that were marked new are still new

### DIFF
--- a/mbox/mbox.c
+++ b/mbox/mbox.c
@@ -54,6 +54,7 @@
 #include "copy.h"
 #include "mutt_globals.h"
 #include "mutt_header.h"
+#include "mutt_thread.h"
 #include "muttlib.h"
 #include "mx.h"
 #include "protos.h"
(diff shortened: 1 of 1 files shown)


gahr (approved):
Subtle one!

Issue #3168, opened 2021-12-29
Open email containing html attachment with Qutebrowser


wgmayer0:
Is there a way to open Neomutt attachments automatically? For example, here are my emails:

![1640741592](https://user-images.githubusercontent.com/73801421/147619123-1ed6e926-194c-4308-9122-78520f2c33ae.png)

And here is an attachment view of the first email:

![1640741689](https://user-images.githubusercontent.com/73801421/147619216-27fa3dcf-6a2e-4a0c-8227-384108636204.png)

As you can see, one of the attachments has the extension `text/html`.

I would like Neomutt to automatically open any and all html files in Qutebrowser if they exist. (https://github.com/qutebrowser/qutebrowser) If they don't exist, I just want it to open whatever text is associated with it.

Here is my muttrc:
<details>
  <summary>Click to expand</summary>
```
source /home/wgm/.mutt/aliases
source /home/wgm/.mutt/keybindings/unbindkeys
source /home/wgm/.mutt/keybindings/keybindings.index
source /home/wgm/.mutt/keybindings/keybindings.attach
source /home/wgm/.mutt/keybindings/keybindings.pager

# set the mailcap file
set mailcap_path = ~/.mutt/mailcap
set implicit_autoview = yes

# stop the "press any key to continue"
set wait_key=no

# appearance
set help=off

# enable threading
set sort=threads

# editor
set editor=vim

# pager
set pager=vim

# delete
set delete=yes

# turn on autotagging messages
set auto_tag=yes

# format the status bar
color status white black
set status_format=''

# disable history
set history=0

# use mbsync
set mbox_type=Maildir
set folder=~/.mutt/mail
set spoolfile="+INBOX"
unset postponed
unset record
mailboxes +INBOX

# use msmtp
set sendmail = "/usr/bin/msmtp"
set realname = "W. Grant Mayer"
set use_from=yes
set envelope_from=yes
set edit_headers = yes
set autoedit

# format dates
set index_format=' %<[y?%<[m?%<[d?%[%H:%M ]&%[%a %d]>&%[%b %d]>&%[%m/%y ]>  %S  %-14.14n  %s '

# set autoview up for html
auto_view text/html

```
</details>

mschilli87:
Untested, but I'd try something like 

```
text/html; qutebrowser '%s' &
```

in `mailcap` combined with 

```
auto_view text/html
```

in `.muttrc`.

wgmayer0:
[...]

PR #3383, opened 2022-04-24
Support viewing text/html attachments with embedded images

dcpurton:
## What does this PR do?

When viewing a `text/html` part within a `multipart/related` group, this PR adds support for writing out all attachments with a Content-ID header and mapping `cid` references to filenames in the `text/html` part.

### Implementation notes

- The logic of `save_cid_attachment()` is similar to the code that saves the attachment in `mutt_view_attachment()`.
- The logic of `map_cid_to_filename()` is similar to the code that edits an attachment in `mutt_rfc3676_space_unstuff_attachment()`.

### Linked issues and discussions

- #3209
- #3168

## Testing

Call the following `rc` file using `./neomutt -n -F view-related.rc` from within the build directory.

A `multipart/related` email with an embedded image is created, postponed, and then its `text/html` part viewed.

If it worked (and you have an appropriate `mailcap` entry for `text/html`), you should see:

neomutt-256

```
# vim: syn=neomuttrc

# filename: view-related.rc

# run from build directory:
#   ./neomutt -n -F view-related.rc

unset record
unset signature

set attach_format = "%u%D%I %t%4n %T%.40d%> [%.15m/%.15M, %.6e%?C?, %C?, %s]"
set folder = "./"
set from = "john.doe@example.com"
set help = no
set postpone = no
set postponed = "postponed.mbox"
set realname = "John Doe"
set recall = no
set spoolfile = "inbox.mbox"
set wait_key = no

set editor = "echo '![](cid:neomutt-256.png)' >> %s"

push "<view-attachments>3<enter><enter>"
push "<postpone-message><change-folder>postponed.mbox<enter>"
push "<group-related>"
push "<tag-entry>4<enter><tag-entry>"
push "<group-alternatives>"
push "1<enter><tag-entry>"
push "<toggle-disposition><toggle-unlink><tag-entry>"
push "<attach-file>/tmp/neomutt-alternative.html<enter>"
push "1<enter><pipe-entry>pandoc -o /tmp/neomutt-alternative.html<enter>"
push "<edit-content-id>\Cuneomutt-256.png<enter>"
push "<rename-attachment><enter>"
push "<toggle-disposition>"
push "<attach-file>contrib/logo/neomutt-256.png<enter>"
push "view multipart/related test<enter>"
push "<mail>John Doe <john.doe@example.com><enter>"
```

### DIFF
--- a/attach/cid.c
+++ b/attach/cid.c
@@ -1,0 +1,84 @@
+/**
+ * @file
+ * Attachment Content-ID header functions
+ *
+ * @authors
+ * Copyright (C) 2022 David Purton <dcpurton@marshwiggle.net>
+ *
+ * @copyright
+ * This program is free software: you can redistribute it and/or modify it under
+ * the terms of the GNU General Public License as published by the Free Software
+ * Foundation, either version 2 of the License, or (at your option) any later
+ * version.
+ *
+ * This program is distributed in the hope that it will be useful, but WITHOUT
+ * ANY WARRANTY; without even the implied warranty of MERCHANTABILITY or FITNESS
+ * FOR A PARTICULAR PURPOSE.  See the GNU General Public License for more
+ * details.
+ *
+ * You should have received a copy of the GNU General Public License along with
+ * this program.  If not, see <http://www.gnu.org/licenses/>.
+ */
+
+/**
+ * @page attach_cid Attachment Content-ID header functions
+ *
+ * Attachment Content-ID header functions
+ */
+
+#include "config.h"
+#include <stddef.h>
+#include "cid.h"
+
+/**
+ * cid_map_free - Free a CidMap
+ * @param[out] ptr CidMap to free
+ */
+void cid_map_free(struct CidMap **ptr)
+{
+  if (!ptr)
+    return;
+
+  struct CidMap *cid_map = *ptr;
+
+  FREE(&cid_map->cid);
+  FREE(&cid_map->fname);
+
+  FREE(ptr);
+}
+
+/**
+ * cid_map_new - Initialise a new CidMap
+ * @param  cid      Content-ID to replace including "cid:" prefix
+ * @param  filename Path to file to replace Content-ID with
+ * @retval ptr      Newly allocated CidMap
+ */
+struct CidMap *cid_map_new(const char *cid, const char *filename)
+{
+  if (!cid || !filename)
+    return NULL;
+
+  struct CidMap *cid_map = mutt_mem_calloc(1, sizeof(struct CidMap));
+
+  cid_map->cid = mutt_str_dup(cid);
+  cid_map->fname = mutt_str_dup(filename);
+
+  return cid_map;
+}
+
+/**
+ * cid_map_list_clear - Empty a CidMapList
+ * @param cid_map_list List of Content-ID to filename mappings
+ */
+void cid_map_list_clear(struct CidMapList *cid_map_list)
+{
+  if (!cid_map_list)
+    return;
+
+  while (!STAILQ_EMPTY(cid_map_list))
+  {
+    struct CidMap *cid_map = STAILQ_FIRST(cid_map_list);
+    STAILQ_REMOVE_HEAD(cid_map_list, entries);
+    cid_map_free(&cid_map);
+  }
+}
--- a/attach/cid.c
+++ b/attach/cid.c
@@ -30,11 +30,13 @@
 #include <stddef.h>
 #include <stdbool.h>
 #include <stdio.h>
+#include <string.h>
 #include "email/lib.h"
 #include "cid.h"
 #include "attach.h"
 #include "mailcap.h"
 #include "mutt_attach.h"
+#include "muttlib.h"
 
 /**
  * cid_map_free - Free a CidMap
@@ -157,4 +159,93 @@ void cid_save_attachments(struct Body *body, struct CidMapList *cid_map_list)
     else
       cid_save_attachment(b, cid_map_list);
   }
+}
+
+/**
+ * cid_to_filename - Replace Content-IDs with filenames
+ * @param filename     Path to file to replace Content-IDs with filenames
+ * @param cid_map_list List of Content-ID to filename mappings
+ */
+void cid_to_filename(struct Buffer *filename, const struct CidMapList *cid_map_list)
+{
+  if (!filename || !cid_map_list)
+    return;
+
+  FILE *fp_out = NULL;
+  char *pbuf = NULL;
+  char *searchbuf = NULL;
+  char *buf = NULL;
+  char *cid = NULL;
+  size_t blen = 0;
+  struct CidMap *cid_map = NULL;
+
+  struct Buffer *tmpfile = mutt_buffer_pool_get();
+  struct Buffer *tmpbuf = mutt_buffer_pool_get();
+
+  FILE *fp_in = mutt_file_fopen(mutt_buffer_string(filename), "r");
+  if (!fp_in)
+    goto bail;
+
+  /* ensure tmpfile has the same file extension as filename otherwise an
+   * HTML file may be opened as plain text by the viewer */
+  const char *suffix = mutt_strn_rfind(mutt_buffer_string(filename),
+                                       mutt_buffer_len(filename), ".");
+  if (suffix && *(suffix++))
+    mutt_buffer_mktemp_pfx_sfx(tmpfile, "neomutt", suffix);
+  else
+    mutt_buffer_mktemp(tmpfile);
+  fp_out = mutt_file_fopen(mutt_buffer_string(tmpfile), "w+");
+  if (!fp_out)
+    goto bail;
+
+  /* Read in lines from filename into buf */
+  while ((buf = mutt_file_read_line(buf, &blen, fp_in, NULL, MUTT_RL_NO_FLAGS)) != NULL)
+  {
+    if (mutt_str_len(buf) == 0)
+    {
+      fputs(buf, fp_out);
+      continue;
+    }
+
+    /* copy buf to searchbuf because we need to edit multiple times */
+    searchbuf = mutt_str_dup(buf);
+    mutt_buffer_reset(tmpbuf);
+
+    /* loop through Content-ID to filename mappings and do search and replace */
+    STAILQ_FOREACH(cid_map, cid_map_list, entries)
+    {
+      pbuf = searchbuf;
+      while ((cid = strstr(pbuf, cid_map->cid)) != NULL)
+      {
+        mutt_buffer_addstr_n(tmpbuf, pbuf, cid - pbuf);
+        mutt_buffer_addstr(tmpbuf, cid_map->fname);
+        pbuf = cid + mutt_str_len(cid_map->cid);
+        mutt_debug(LL_DEBUG2, "replaced \"%s\" with \"%s\" in file \"%s\"\n",
+                   cid_map->cid, cid_map->fname, mutt_buffer_string(filename));
+      }
+      mutt_buffer_addstr(tmpbuf, pbuf);
+      FREE(&searchbuf);
+      searchbuf = mutt_buffer_strdup(tmpbuf);
+      mutt_buffer_reset(tmpbuf);
+    }
+
+    /* write edited line to output file */
+    fputs(searchbuf, fp_out);
+    fputs("\n", fp_out);
+    FREE(&searchbuf);
+  }
+
+  mutt_file_set_mtime(mutt_buffer_string(filename), mutt_buffer_string(tmpfile));
+
+  /* add filename to TempAtachmentsList so it doesn't get left lying around */
+  mutt_add_temp_attachment(mutt_buffer_string(filename));
+  /* update filename to point to new file */
+  mutt_buffer_copy(filename, tmpfile);
+
+bail:
+  FREE(&buf);
+  mutt_file_fclose(&fp_in);
+  mutt_file_fclose(&fp_out);
+  mutt_buffer_pool_release(&tmpfile);
+  mutt_buffer_pool_release(&tmpbuf);
 }
--- a/attach/cid.c
+++ b/attach/cid.c
@@ -28,7 +28,13 @@
 
 #include "config.h"
 #include <stddef.h>
+#include <stdbool.h>
+#include <stdio.h>
+#include "email/lib.h"
 #include "cid.h"
+#include "attach.h"
+#include "mailcap.h"
+#include "mutt_attach.h"
 
 /**
  * cid_map_free - Free a CidMap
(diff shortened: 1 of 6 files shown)


attach/mutt_attach.c
```diff
@@ -366,6 +378,185 @@ void mutt_check_lookup_list(struct Body *b, char *type, size_t len)
...
+      goto bail;
+    has_tempfile = true;
+    mutt_debug(LL_DEBUG2, "attachment with \"Content-ID: %s\" saved to file \"%s\"\n",
+               id, mutt_buffer_string(tmpfile));
+
+    /* Add Content-ID to filename mapping to list */
+    struct CidMap *cid_map = mutt_mem_malloc(sizeof(struct CidMap));
+    mutt_buffer_printf(cid, "cid:%s", id);
+    cid_map->cid = mutt_str_dup(mutt_buffer_string(cid));
+    cid_map->fname = mutt_str_dup(mutt_buffer_string(tmpfile));
```
gahr, line 424: May I suggest a `struct CidMap *cid_map_new(const char *cid, const char *tmpfile)` function?

flatcap: Seconded.
And a `_free()` function (as it's non-trivial)
And `calloc`, to be safe

dcpurton:
@flatcap Any thoughts on putting these `cid` functions in a separate file? If a `<save-html-bundle>` command is created they will get reused. Or can leave where they are for now and change in a future PR.

flatcap:
> Any thoughts on putting these `cid` functions in a separate file?



**Key**:
- Red: Called from outside `attach/`
- Green: Called from within `attach/`
- Grey: static functions

It looks like these make sense:
- `cid_map_free()`
- `cid_map_new()`
- `cid_save_attachment()`
- `cid_save_attachments()`                                                                                           
- `cid_to_filename()`
- `mutt_add_temp_attachment()`

(we can always change our mind later)

attach/cid.c
```diff
@@ -0,0 +1,235 @@
...
+ *
+ * This program is distributed in the hope that it will be useful, but WITHOUT
+ * ANY WARRANTY; without even the implied warranty of MERCHANTABILITY or FITNESS
+ * FOR A PARTICULAR PURPOSE.  See the GNU General Public License for more
+ * details.
+ *
+ * You should have received a copy of the GNU General Public License along with
+ * this program.  If not, see <http://www.gnu.org/licenses/>.
+ */
+
```
flatcap, line 22: We need an `@page` Doxygen comment here.
Something minimal will do.

In `main.c`, these comments link us to `attach/lib.h`
https://github.com/neomutt/neomutt/blob/af90c326a3156b509cb371f6af908fd73a72a31c/main.c#L50-L60

and in `attach/lib.h`, these comments link us to the `@page` in each source file
https://github.com/neomutt/neomutt/blob/af90c326a3156b509cb371f6af908fd73a72a31c/attach/lib.h#L28-L37

attach/cid.c
```diff
@@ -0,0 +1,235 @@
...
+{
+  if (!ptr)
+    return;
+
+  struct CidMap *cid_map = *ptr;
+
+  FREE(&cid_map->cid);
+  FREE(&cid_map->fname);
+
+  FREE(ptr);
```
flatcap, line 49: I changed this from
```c
FREE(&cid_map);
*ptr = NULL;
```
to just
```c
FREE(ptr);
```
That's the pattern for our free functions.

attach/cid.c
```diff
@@ -0,0 +1,235 @@
...
+}
+
+/**
+ * cid_map_new - Initialise a new CidMap
+ * @retval ptr Newly allocated CidMap
```
flatcap, line 54: Need `@param`s for `cid` and `filename`

attach/cid.c
```diff
@@ -0,0 +1,235 @@
...
+  if (!body || !cid_map_list)
+    return;
+
+  char *id = mutt_param_get(&body->parameter, "content-id");
+  struct Buffer *tmpfile = mutt_buffer_pool_get();
+  struct Buffer *cid = mutt_buffer_pool_get();
+  bool has_tempfile = false;
+  FILE *fp = NULL;
+
+  if (id)
```
flatcap, line 88: If `!id`, then we don't do anything, so we could move this line above the two `mutt_buffer_pool_get()` lines.
```c
char *id = mutt_param_get(&body->parameter, "content-id");
if (!id)
  return;
```

That'd simplify the error path.

attach/cid.c
```diff
@@ -0,0 +1,235 @@
...
+  struct Buffer *tmpfile = mutt_buffer_pool_get();
+  struct Buffer *tmpbuf = mutt_buffer_pool_get();
+
+  FILE *fp_in = mutt_file_fopen(mutt_buffer_string(filename), "r");
+  if (!fp_in)
+    goto bail;
+
+  const char *suffix = mutt_strn_rfind(mutt_buffer_string(filename),
+                                       mutt_buffer_len(filename), ".");
+  if (++suffix)
```
flatcap, line 167: I think you're missing a `*`
`mutt_strn_rfind()` returns a pointer, or `NULL`.
If you prefix-increment the pointer, it will be moved on one place.
If you increment `NULL`, you'll get 0x0001, which is `true`, but it'll crash.

flatcap (changes requested):

attach/cid.c
dcpurton: Ugh! That is a bad error.

What I was intending was to ensure that filename didn't end in a `.` (i.e., next character is `\0`).

It never went wrong because in theory `filename` should always have an extension given to it already.

I think this should be safe…

```C
  if (suffix && *(suffix++))
```

attach/cid.c
dcpurton: I left the `goto` in, but could clean up and return if `mutt_save_attachment()` fails if you think better.

flatcap: The `goto`'s fine

attach/cid.c
flatcap: yeah, that looks a bit better

dcpurton:
@flatcap There's still changes requested, but I can't seem to see what they are.

flatcap:
> There's still changes requested, but I can't seem to see what they are

No, they're all done, but I need to cancel my review to turn off the red message above.

attach/lib.c
```diff
@@ -103,6 +105,38 @@ bool attach_body_parent(struct Body *start, struct Body *start_parent,
...
+ * attach_body_ancestor - Find the ancestor of a body with specified subtype
+ * @param[in]  start         Body to start search from
+ * @param[in]  body          Body to find ancestor of
+ * @param[in]  subtype       Mime subtype of ancestor to find
+ * @param[out] body_ancestor Body ancestor if found
+ * @retval     true          Ancestor body found
+ * @retval     false         Ancestor body not found
+ */
+bool attach_body_ancestor(struct Body *start, struct Body *body,
+                          const char *subtype, struct Body **body_ancestor)
```
flatcap, line 118: This purpose of this function is to find a `Body`,
so it'd be simpler to just return the ancestor and drop the `bool`.

attach/mutt_attach.c
```diff
@@ -487,6 +489,26 @@ int mutt_view_attachment(FILE *fp, struct Body *a, enum ViewAttachMode mode,
 
     mutt_rfc3676_space_unstuff_attachment(a, mutt_buffer_string(tmpfile));
 
+    /* check for multipart/related and save attachments with a Content-ID */
+    struct Body *related_ancestor = NULL;
+    bool has_related = false;
```
flatcap, line 494: We don't need this `bool`, we can just test if `related_ancestor` is set

attach/mutt_attach.c
```diff
@@ -487,6 +489,26 @@ int mutt_view_attachment(FILE *fp, struct Body *a, enum ViewAttachMode mode,
 
     mutt_rfc3676_space_unstuff_attachment(a, mutt_buffer_string(tmpfile));
 
+    /* check for multipart/related and save attachments with a Content-ID */
+    struct Body *related_ancestor = NULL;
```
flatcap, line 493: This is only used in the `if` clause, so we can reduce the scope of the variable

attach/mutt_attach.c
```diff
@@ -487,6 +489,26 @@ int mutt_view_attachment(FILE *fp, struct Body *a, enum ViewAttachMode mode,
...
+    bool has_related = false;
+    struct CidMapList cid_map_list;
+    STAILQ_INIT(&cid_map_list);
```
flatcap, line 496: We can reduce the scope of `cid_map_list`, too.
And optimise the initialisation:
```suggestion
    struct CidMapList cid_map_list = STAILQ_HEAD_INITIALIZER(&cid_map_list);
```

attach/cid.c
```diff
@@ -0,0 +1,244 @@
...
+   * HTML file may be opened as plain text by the viewer */
+  const char *suffix = mutt_strn_rfind(mutt_buffer_string(filename),
+                                       mutt_buffer_len(filename), ".");
+  if (suffix && *(suffix++))
+    mutt_buffer_mktemp_pfx_sfx(tmpfile, "neomutt", suffix);
+  else
+    mutt_buffer_mktemp(tmpfile);
+  fp_out = mutt_file_fopen(mutt_buffer_string(tmpfile), "w+");
+  if (!fp_out)
+    goto bail;
```
flatcap, line 183: If this fails, then `fp_in` is leaked, see below

attach/cid.c
```diff
@@ -0,0 +1,244 @@
...
+
+    /* write edited line to output file */
+    fputs(searchbuf, fp_out);
+    fputs("\n", fp_out);
+    FREE(&searchbuf);
+  }
+
+  FREE(&buf);
+  mutt_file_fclose(&fp_in);
+  mutt_file_fclose(&fp_out);
```
flatcap, line 224: I suggest moving these three cleanup lines after `bail:`

attach/cid.c
```diff
@@ -0,0 +1,244 @@
...
+  mutt_buffer_pool_release(&tmpfile);
+  mutt_buffer_pool_release(&tmpbuf);
+
+  /* free Content-ID to filename mapping list */
+  while (!STAILQ_EMPTY(cid_map_list))
+  {
+    cid_map = STAILQ_FIRST(cid_map_list);
+    STAILQ_REMOVE_HEAD(cid_map_list, entries);
+    cid_map_free(&cid_map);
+  }
```
flatcap, line 243: As this is non-trivial, I suggest a `cid_map_list_clear()`
(`_clear()` rather than `_free()` because it empties the object, but doesn't free it)

flatcap, line 243: Also, I don't think this belongs here.

You create `cid_map_list` in `cid_save_attachments()`
It gets populated with a call to `cid_save_attachments()`
The contents get used, here in `cid_to_filename()`
I think the freeing should happen in `mutt_view_attachment()`
(and we can make the param const)

flatcap (changes requested):
It's a lot of words, but the changes are all small.

flatcap:
I've made a few minor tidies.
I added a check in `cid_map_list_clear()`, so that passing `NULL` is safe.
I removed the checks before the three frees you moved.
It's safe to call `FREE()` and `mutt_file_fclose()` with `NULL` arguments.

flatcap (approved):

nickwynja:
@dcpurton Does this only work with `multipart/related`? What about embedded images in `text/html`?

dcpurton:
@nickwynja , you misunderstand the concepts going on. I'll see if I can explain :).

An email has two parts

1. Headers (containing `To:`, `From:`, `Subject:`, etc.)
2. Body containing *exactly one* MIME part.

So we could represent a normal text only email like this:

```
Email Headers
└─> text/plain Body
```

Now if we wanted to attach a PNG image to our email, there must still be *exactly one* MIME part for the body. To do this we make that part a `multipart` MIME part. Think of a `multipart` MIME part as a container which can contain other MIME parts. For a simple attachment, we use a `multipart/mixed` MIME type email body. `multipart/mixed` is a container which can hold MIME parts with different types. So the email becomes:

```
Email Headers
└─> multipart/mixed Body
    ├─> text/plain part containing email message
    └─> image/png containing attached PNG image
```

Now many email clients send email containing the message in two formats: Plain text and HTML. But the content is the same, so the receiving email client only needs to display one based on preference. To do this we can use a `mutlipart/alternative` MIME type for the email. The structure of a simple email might be:

```
Email Headers
└─> multipart/alternative Body
    ├─> text/plain part containing email message in plain text
    └─> text/html part containing the same email message, but in HTML
```

Now say we wanted to send a simple HTML email with no plain text alternative. We could do it like this:

```
Email Headers
└─> text/html Body
```

This is all good since there is only one MIME part for the email body. But what if we wanted to put an image into the HTML?

There's two options available to us:

1. We could link to an image somewhere on the Internet: `` and the receiving email client would request the image from `example.com` and display it. Most mass marketing emails include pictures in this way and they are easy to display in NeoMutt, you just save the HTML part and open it in a browser. The browser requests any images from the Internet and shows you the result.
2. We could include the image in the email message itself and reference it from the `text/html` part. This is slightly more complex and viewing this kind of email is what this PR supports. Let me explain how it works.

In this case the image must be included in the email and again it must be included in a `multipart` container along with the HTML part since the email can still only have one main MIME part. We use a `multipart/related` container MIME type. This type means that the containing parts are related to each other and can reference each other. In order for MIME parts to find each other they need to be given a `Content-ID:` header which can be used as a reference. The email might look like this:

```
Email Headers
└─> multipart/related Body
    ├─> text/html part containing email message
    └─> image/png part containing an image which can be referenced from the text/html part
```

The `image/png` part needs to have a Content-ID header looking something like: `Content-ID: <image@id>`. This can then be referenced in the `text/html` part like this: ``

So if we want to view the `text/html` part of an email like this we need to do more than save the `text/html` part and open it in a browser. We need to save the `text/html` part and the `image/png` part and then fix up the image `src` reference in the saved HTML file to point to the saved PNG file. This PR does this for a the `text/html` part and any other MIME parts in the relevant `multipart/related` container so the references in the `text/html` correctly resolve when it is opened in a browser.

Hope that all makes sense!

dcpurton:
@nickwynja , assuming the email has a `multipart/related` MIME part at the top level, this email structure is supposed to work.

> I'm opening using `<view-mailcap>`. In my mailcap, I've tried a couple things. Usually I use a variation of this old [view_attachment.sh script](https://gist.github.com/Integralist/cc2616ece918fdd8239d16cca62e37de) but currently am just testing with `text/html; open %s`, which just opens a text editor with the HTML.

Using a script like this is possibly going to cause issues. The script only copies the main HTML file to a new location, not any images that are also saved. So if something is removing the files after they are copied and before they are opened then you'll get a broken link. Moreover, this PR uses full paths for the image source, so even if your script moved all files, then the HTML file will still point to the original location for images.

However I don't think this is your problem.

> The HTML file gets temporarily saved in `~/.mutt/temp/`. The relevant HTML is this: ``.

Does the `img` tag look like this after it's saved and opened in your editor? This indicates that the code to save all the attachments and fix up the links isn't running at all (as you suspected).

It ought to look something like: ``.

It might be worth opening an issue with more info on your set up to properly investigate. Probably my code is buggy. :grin:

dcpurton:
@nickwynja I'd also be interested in seeing if the testing method in the main description of this PR works for you.

That might help narrow down if it is a problem with a particular email or a problem with Mac OS.

nickwynja:
@dcpurton It appears like the testing method is not working. I tried the `view-related.rc` from above with the addition of pointing to my mailcap file with `set mailcap_path`. I'm using `less -s -M +Gg %s` for these mimetypes for testing, if that's at all relevant.

This is what the message looks like when I view it:

```
From john.doe@example.com Fri Jun  3 08:54:29 2022
Date: Fri, 3 Jun 2022 08:54:29 -0400
From: John Doe <john.doe@example.com>
To: John Doe <john.doe@example.com>
Subject: view multipart/related test
MIME-Version: 1.0
Content-Type: multipart/related; boundary="drqt24jhsr33uabg"
Content-Disposition: inline
X-Mutt-Fcc:
Status: RO
Content-Length: 26934
Lines: 390

[-- Attachment #1 --]
[-- Type: multipart/alternative, Encoding: 7bit, Size: 0.3K --]

![](cid:neomutt-256.png)

[-- Attachment #2: neomutt-256.png --]
[-- Type: image/png, Encoding: base64, Size: 25K --]

[-- image/png is unsupported (use 'v' to view this part) --]
```

And the attachments looks like this:

```
  I     1 <no description>                   [multipart/alternative, 7bit, 0.3K]
  I     2 ├─><no description>                 [text/plain, 7bit, us-ascii, 0.1K]
  I     3 └─><no description>                  [text/html, 7bit, us-ascii, 0.1K]
  I     4 neomutt-256.png                               [image/png, base64, 25K]
```


The result from `<view-mailcap>` on the `text/html` is:

```
<p><img src="cid:neomutt-256.png" /></p>
```

This temp file is saved at `/var/folders/rd/m1_dzqv562zd498blb2yvtmr0000gn/T/neomutt-Nicks-MacBook-Pro-501-92173-10959897006356855227`. No `neomutt-256.png` is saved in that directory when I `<view-mailcap>` on the `multipart/alternative`. I've also tried updating the `tmpdir` to something in my user directory and I see the same outcome.

Happy to make a new issue to track down the specific issue as well.

dcpurton:
@nickwynja yes, please open an issue and fill in all the info the bug template asks for.

I do have access to a Mac, so maybe I can build with your settings and see what is going on.

You could also run the test above with `-d 4`, as there are a few debugging lines on the code. Look through the log and see if there are any errors listed.

And if you can run in a debugger that would help too.

dcpurton:
@nickwynja yes, please open an issue and fill in all the info the bug template asks for.

I do have access to a Mac, so maybe I can build with your settings and see what is going on.

You could also run the test above with `-d 4`, as there are a few debugging lines on the code. Look through the log and see if there are any errors listed.

And if you can run in a debugger that would help too.

Oh and I know it's obvious, but double check you are actually running a version containing this code. It was only merged a few weeks ago. Perhaps you have an older version in your path.

dcpurton:
@nickwynja I managed to install NeoMutt using brew under Mac OS and successfully viewed an email with an inline image.

I installed NeoMutt with `brew install NeoMutt --HEAD` in order to get the latest commits.

I made a `~/.mailcap` file with one entry:

```
text/html; open '%s'; nametemplate=%s.html
```

I found the macro behaviour slightly different than under Linux, but putting `neomutt-256.png` into current directory and running `NeoMutt -n -F view-related.rc` successfully opened the attachment in Safari and showed the NeoMutt logo. The TEMP directory in use is similar to yours.

This was the `view-related.rc` that worked for me:

```
# vim: syn=neomuttrc

# filename: view-related.rc

# run from directory containing neomutt-256.png:
#   ./neomutt -n -F view-related.rc

unset record
unset signature

set attach_format = "%u%D%I %t%4n %T%.40d%> [%.15m/%.15M, %.6e%?C?, %C?, %s]"
set folder = "./"
set from = "john.doe@example.com"
set help = no
set postpone = no
set postponed = "postponed.mbox"
set realname = "John Doe"
set recall = no
set spoolfile = "inbox.mbox"
set wait_key = no

set editor = "echo '![](cid:neomutt-256.png)' >> %s"

push "<view-attachments>3<enter><enter>"
push "<postpone-message><change-folder>postponed.mbox<enter>"
push "5<enter><edit-content-id>neomutt-256.png<enter>"
push "<group-related>"
push "<tag-entry>4<enter><tag-entry>"
push "<group-alternatives>"
push "1<enter><tag-entry>"
push "<edit-type>text/html<enter>"
push "<toggle-disposition><toggle-unlink><tag-entry>"
push "<attach-file>/tmp/neomutt-alternative.html<enter>"
push "1<enter><pipe-entry>pandoc -o /tmp/neomutt-alternative.html<enter>"
push "<edit-type>image/png<enter>"
push "<rename-attachment>neomutt-256.png<enter>"
push "<toggle-disposition>"
push "<attach-file>neomutt-256.png<enter>"
push "view multipart/related test<enter>"
push "<mail>John Doe <john.doe@example.com><enter>"
```

nickwynja:
> Oh and I know it's obvious, but double check you are actually running a version containing this code. It was only merged a few weeks ago. Perhaps you have an older version in your path.

@dcpurton Well, this is embarrassing. I _had_ double checked but it took the triple check to find my issue. I had `git pull`ed a couple times to make sure I was on the latest but didn't realize until now that I was checked out on old `master` and not `main` so _hadn't_  gotten the latest commits. 😞 

It is working for me. Thank you so much for taking the time to explain the concept for me and help me track things down. I'm sorry to have wasted so much of your time.

dcpurton:
@nickwynja That's good to hear! No need to apologise and I'm glad things work OK under Mac.

HaxNet:
Hi guys,

So sorry - but I am having a hard time figuring this out. how would I even call this out when viewing an html with images? Do I have to create a macro, pager?
I am still not getting the logo of the neomutt-256.png when I run the command in terminal.

dcpurton:
> So sorry - but I am having a hard time figuring this out. how would I even call this out when viewing an html with images? Do I have to create a macro, pager?

@HaxNet you don't need a macro, it should work transparently when you view the `text/html` attachment as you normally would.

> I am still not getting the logo of the neomutt-256.png when I run the command in terminal. 

Please submit a bug report with as much detail as you can and I'll take a look.

PR #3385, opened 2022-04-25
Fix parsing of sendmail cmd

clkamp:
Steps to reproduce:
- set sendmail to "x -- a"
- try to send a mail
- NeoMutt hangs


flatcap (approved):
Good find!
and thank you very much for fixing it :heart:

PR #3404, opened 2022-05-13
Store credentials in the MacOS Keychain

flatcap:
This is PR #3189, reopened.

On macOS, all passwords are managed centrally in a program known as Keychain.
This patch proposes to integrate neomutt with Keychain, to avoid storing even encrypted passwords in plain text files.
The code has been adapted from a similar (successful) effort in git.git, originally authored by Jeff King.

### DIFF
--- a/Makefile.autosetup
+++ b/Makefile.autosetup
@@ -273,6 +273,9 @@ LIBCONNOBJS=	conn/config.o conn/connaccount.o conn/getdomain.o conn/raw.o \
 @if HAVE_SASL
 LIBCONNOBJS+=	conn/sasl.o
 @endif
+@if USE_MACOS_KEYCHAIN
+LIBCONNOBJS+=	conn/macos_keychain.o
+@endif
 @if USE_SSL
 LIBCONNOBJS+=	conn/dlg_verifycert.o
 @endif
--- a/auto.def
+++ b/auto.def
@@ -106,6 +106,8 @@ options {
   with-zlib:path            => "Location of zlib"
   zstd=0                    => "Enable Zstandard header cache compression support"
   with-zstd:path            => "Location of Zstandard"
+# MacOS
+  macoskeychain=0           => "Use MacOS's Keychain for authentication"
 # System
   with-sysroot:path         => "Target system root"
   with-tmpdir:=/tmp         => "location of the tmp directory"
@@ -156,7 +158,7 @@ if {1} {
     debug-graphviz debug-notify debug-parse-test debug-queue debug-window doc
     everything fmemopen full-doc fuzzing gdbm gnutls gpgme gss homespool idn
     idn2 include-path-in-cflags inotify kyotocabinet lmdb locales-fix lua lz4
-    mixmaster nls notmuch pcre2 pgp pkgconf qdbm rocksdb sasl smime sqlite ssl
+    mixmaster nls notmuch macoskeychain pcre2 pgp pkgconf qdbm rocksdb sasl smime sqlite ssl
     testing tdb tokyocabinet zlib zstd
   } {
     define want-$opt [opt-bool $opt]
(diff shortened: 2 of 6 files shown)


flatcap:
@the-x-at
Thanks for your offer of help!

The code works, I'm told (I don't have a Mac),
but there were some unanswered questions about _how_ it should work.

Should it always be built for Macs? (it doesn't add any dependencies)
Should it always be enabled? (or should we have a config variable?)

Please can you give it a test and tell us your opinion on the questions.
(I'll try to find some more Mac users)

If we need to make any changes, I can guide you through the code.
Thanks

gahr:
I'd like to ask the question: how many different backends do we want to implement in the code? Where do we stop? 

An alternative approach would be to implement a a flavour of set my_foo = "script|" that does lazy loading on demand and isn't cached, so users can employ whatever backend they want to store their passwords.

jindraj:
> I'd like to ask the question: how many different backends do we want to implement in the code? Where do we stop?
> 
> An alternative approach would be to implement a a flavour of set my_foo = "script|" that does lazy loading on demand and isn't cached, so users can employ whatever backend they want to store their passwords.

perhaps instead of implementing backends for keychain, 1password, lastpass, bitwarden and other password managers  people could put their scripts utilizing `my_foo="script|"` solutions under contrib/ directory.

gahr:
> perhaps instead of implementing backends for keychain, 1password, lastpass, bitwarden and other password managers people could put their scripts utilizing `my_foo="script|"` solutions under contrib/ directory.

Yep - the only difference is that the secret would be kept in NeoMutt's memory, which is why I sugges a lazy+forget solution.

flatcap:
For passwords, there's one big problem with running external scripts -- special chars.
How many times have we been asked for help, only to find that the password contains `'` `$` or similar.

flatcap:
Closed in favour of a general solution: #3408

flatcap:
@the-x-at
This PR may be closed, but we'd still welcome your help.
We chose to create a single command so that anyone can script a solution for their needs.

- [Account Command](https://neomutt.org/feature/account-cmd)

Would you be interested in writing that Mac script for us?  `bash`, `python`, etc
Then we can distribute it along with NeoMutt.

The script will be passed arguments, e.g. `--host gmail.com` or `--username flatcap`,
then it will lookup the matching details in the Keychain and respond:

```
usename: flatcap
password: secret
```

A quick Google shows a few sites with examples of how to script the Keychain.
- https://blog.koehntopp.info/2017/01/26/command-line-access-to-the-mac-keychain.html
- https://scriptingosx.com/2021/04/get-password-from-keychain-in-shell-scripts/
- https://www.netmeister.org/blog/keychain-passwords.html

If you want to discuss it, you can also find us on IRC:
`#neomutt` on `irc.libera.chat` ([web client](https://kiwiirc.com/nextclient/irc.libera.chat/#neomutt))

Thanks

gahr:
I guess NeoMutt can be started by any other program too

the-x-at:
Sure, but it does not spit out immediately a password like `security` does. This is the reason why I was feeling my solution is not as hardened as it should be.

gahr:
Is it possible to always require the user password to interact with keychain? It might defeat the purpose to a certain extent, but at the end of the day I guess you either trust the account or you don't.. if you don't (i.e., you assume a compromised user account), there are just so many things you can do...

flatcap:
> This sounds exactly like the solution I have already implemented in python

Great.  Would you be willing to share it?

> increases the attack surface

Yes, I understand your point.

Sometimes, we have to turn down some _specific_ features in favour of more general solutions.
NeoMutt is already a LOT of code and we have few developers.
Also, I don't own a Mac :-)

the-x-at:
I am happy to help out here. Currently, I am using `offlineimap` for retrieving emails and NeoMutt internals `smtp_url`/`smtp_pass`, so I will have to align my setup in order to build on @gahr's new code.

the-x-at:
I have a Python 3 script ready for use. What would be the best way to get this to you? Can I open a branch on git and push it? Do I need to fork the project? Which sub-directory should this go to? In addition, I would need to write up some instructions on how to use the whole framework. Again, a hint where to insert this is much appreciated.

flatcap:
Generally, you'd fork, push, PR, but NeoMutt's a very inclusive place, so I've sent you an invitation to join.

I suggest:
- create a branch `devel/keychain`
- create a directory `contrib/account-command`
- create `keychain.py` and `keychain.txt` (we can tweak this later if needed)
- add the new dir to the Makefile.autosetup

https://github.com/neomutt/neomutt/blob/2d2e123c87b014c9e500d87bd130c8286d477e3f/contrib/Makefile.autosetup#L1

When you're done you can create a PR from the branch.
Thanks!

PR #3407, opened 2022-05-15
menu: allow the user to quit generic menus

flatcap:
Recent key-handling changes have left the user unable to quit from the three generic dialogs:
- Verify Certificate Dialog
- History Dialog
- Pattern Dialog

This PR changes those dialogs to accept <kbd>q</kbd> as `<exit>`

The bit I'm not certain about is the change to `retry_generic()`.

Before the change, the `q` keypress is pushed into the queue:
```c
  if (lastkey)
    mutt_unget_ch(lastkey);
```
but then dumped:
```c
  if (mtype != MENU_EDITOR)
  {
    /* probably a good idea to flush input here so we can abort macros */
    mutt_flushinp();
  }
```
meaning that:
```c
  return (struct KeyEvent){ .ch = mutt_getch().ch, .op = OP_NULL };
```
hangs, waiting for the user to hit <kbd>q</kbd> a second time.

### DIFF
--- a/conn/dlg_verifycert.c
+++ b/conn/dlg_verifycert.c
@@ -252,7 +252,11 @@ int dlg_verify_certificate(const char *title, struct CertArray *carr,
     // Try to catch dialog keys before ops
     if (menu_dialog_dokey(menu, &op) != 0)
     {
-      op = km_dokey(MENU_GENERIC);
+      struct KeyEvent event = km_dokey_event(MENU_GENERIC);
+      if (event.ch == 'q')
+        op = OP_EXIT;
+      else
+        op = event.op;
     }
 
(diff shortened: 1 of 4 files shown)


gahr:
Handling `q` in a special way doesn't sound like the right way to do it, unless this is a temporary fix.

flatcap:
> doesn't sound like the right way to do it

I agree, but I'm reluctant to mess with `km_dokey()` and friends.

> unless this is a temporary fix

Let's say yes...
I'm not sure what the correct solution is, but it'd probably involve killing the "generic" menu.
(but I don't have time for _that_)

---

The problem needs to be fixed before the next release -- other suggestions welcome.

gahr (approved):

PR #3411, opened 2022-05-26
Add error when $sendmail has shell metachars

rafork:
Add a check that $sendmail does not contain any shell metacharacters before using it.

The manual entry states that $sendmail is a "program and arguments" but people can mistake this and think that it can be an arbitrary shell command. The presence of shell metacharacters indicates that the user is attempting something that won't work. The existing result is whatever error message comes out of the program in $sendmail. This patch makes neomutt give the user a higher-level error message that indicates the cause of the problem, rather than running a command that won't work, and hoping they figure it out for themselves.

   No documentation needed to be updated.
   There is a new error message, and I've added placeholders in all the .po files for it.

   Before this patch, 9 of 457 unit tests failed on my system.
   After this patch, the same unit tests failed.
   No new unit test failures introduced.

   - Added [doxygen code documentation]
   Not needed. No new functions added.

   - Code follows the [style guide]
   Yes.

   N/A

Note: I just noticed that this also applies to $inews, but the error message only mentions $sendmail. Similarly, the existing check immediately above, with the error message "$sendmail must be set in order to send mail", also applies to $inews but only mentions $sendmail. Should they both be fixed to mention $inews when relevant? I can add that if you like. Perhaps the existing check (not null) can never fail for $inews, but the new check (shell meta) could fail for $inews.

### DIFF
--- a/send/config.c
+++ b/send/config.c
@@ -80,6 +80,27 @@ static int smtp_auth_validator(const struct ConfigSet *cs, const struct ConfigDe
   return CSR_SUCCESS;
 }
 
+/**
+ * simple_command_validator - Validate the "sendmail" and "inews" config variables - Implements ConfigDef::validator() - @ingroup cfg_def_validator
+ */
+static int simple_command_validator(const struct ConfigSet *cs, const struct ConfigDef *cdef,
+                                    intptr_t value, struct Buffer *err)
+{
+  // Check for shell metacharacters that won't do what the user expects
+  const char *valstr = (const char *) value;
+  if (!valstr)
+    return CSR_SUCCESS;
+
+  const char c = valstr[strcspn(valstr, "|&;()<>[]{}$`'~\"\\*?")];
+  if (c == '\0')
+    return CSR_SUCCESS;
+
+  // L10N: This applies to the "$sendmail" and "$inews" config variables.
+  mutt_buffer_printf(err, _("Option %s must not contain shell metacharacters: %c"),
+                     cdef->name, c);
+  return CSR_ERR_INVALID;
+}
+
 static struct ConfigDef SendVars[] = {
   // clang-format off
   { "abort_noattach", DT_QUAD, MUTT_NO, 0, NULL,
@@ -232,7 +253,7 @@ static struct ConfigDef SendVars[] = {
   { "reverse_real_name", DT_BOOL|R_INDEX|R_PAGER, true, 0, NULL,
     "Set the 'From' from the full 'To' address the email was sent to"
   },
-  { "sendmail", DT_STRING|DT_COMMAND, IP SENDMAIL " -oem -oi", 0, NULL,
+  { "sendmail", DT_STRING|DT_COMMAND, IP SENDMAIL " -oem -oi", 0, simple_command_validator,
     "External command to send email"
   },
   { "sendmail_wait", DT_NUMBER, 0, 0, NULL,
@@ -300,7 +321,7 @@ static struct ConfigDef SendVarsNntp[] = {
   { "ask_x_comment_to", DT_BOOL, false, 0, NULL,
     "(nntp) Ask the user for the 'X-Comment-To' field before editing"
   },
-  { "inews", DT_STRING|DT_COMMAND, 0, 0, NULL,
+  { "inews", DT_STRING|DT_COMMAND, 0, 0, simple_command_validator,
     "(nntp) External command to post news articles"
   },
   { "mime_subject", DT_DEPRECATED|DT_BOOL, true, IP "2021-03-24" },


flatcap:
Thanks for the PR, @rafork.
We try to merge all of upstream's fixes, but that can take time, so a PR is really appreciated :heart:

First, I've just backed out the translation changes.
In NeoMutt, the translators have their own branch to reduce churn.
I'll update that when we're done.

> * Before this patch, 9 of 457 unit tests failed on my system

Ah!  Some of the unit tests require external files.
Your code doesn't affect any of the tests, so they all pass.

> https://github.com/neomutt/neomutt-test-files

(I've made a note to update the Pull Request instructions)

> this also applies to $inews, but the error message only mentions $sendmail

good spot

We could do something simple, using `OptNewsSend`
```c
mutt_error(_("$%s must be set in order to send mail"), OptNewsSend ? "inews" : "sendmail");
```

But...  NeoMutt's config system has a better solution: validators.
In Mutt, if you set a variable you might not know there's a problem until you use it.
In NeoMutt, many variables have validators that'll tell you immediately if something's wrong.

For example, this variable:
https://github.com/neomutt/neomutt/blob/2d2e123c87b014c9e500d87bd130c8286d477e3f/send/config.c#L262-L264

Uses this validator: (note: it knows the name of the variable being tested)
https://github.com/neomutt/neomutt/blob/2d2e123c87b014c9e500d87bd130c8286d477e3f/send/config.c#L41-L54

If you're willing, please could you turn your check into a validator function, then add it to both `$sendmail` and `$inews`.

Thanks!

flatcap:
One other thing...
Your commit details are `raf at raf dot org`
Anonymity's fine, but if you'd like your real name in the release notes, you'll need to tell me what it is :-)

rafork:
Thanks. I'm happy to change it to a validator (a much better idea), but I don't know how to get your fixes to the PR back into my fork's branch to work on it. I might need to start again with a new branch and new PR if I can't find helpful instructions on doing that. I'm still looking. And the name raf is fine. Thanks for checking.

rafork:
Thanks! That worked perfectly.

I've changed it to a validator. It checks that sendmail/inews have a non-empty value, and that it doesn't contain shell metacharacters. It doesn't complain for "set sendmail\n", but it does complain for "set sendmail =\n". The message for that is:

  "Option %s must not be empty"

When there is a value, and it contains a shell metacharacter, the message is:

  "Option %s must not contain shell metacharacters: %c"

So it tells the user what the offending character is.

send/config.c
```diff
@@ -80,6 +80,44 @@ static int smtp_auth_validator(const struct ConfigSet *cs, const struct ConfigDe
...
+{
+  const char *valstr = (const char *)value;
+
+  /* Check that there is a value */
+  if (valstr == NULL || *valstr == '\0')
+  {
+    // L10N: This applies to the "$sendmail" and "$inews" config variables.
+    mutt_buffer_printf(err, _("Option %s must not be empty"), cdef->name);
+    return CSR_ERR_INVALID;
+  }
```
flatcap, line 97: We don't need this `if`...

`$inews` is allowed to be empty.
If it is, then NeoMutt will post directly to the server.

For `$sendmail`, we've got a better option: `DT_NOT_EMPTY` (see below)

send/config.c
```diff
@@ -80,6 +80,44 @@ static int smtp_auth_validator(const struct ConfigSet *cs, const struct ConfigDe
...
+#if defined(USE_NNTP)
+/**
+ * inews_validator - Validate the "inews" config * variable - Implements ConfigDef::validator() - @ingroup cfg_def_validator
+ */
+static int inews_validator(const struct ConfigSet *cs, const struct ConfigDef *cdef,
+                           intptr_t value, struct Buffer *err)
+{
+  return sendmail_validator(cs, cdef, value, err);
+}
+#endif
```
flatcap, line 119: We don't need this.
Both variables can refer to the same validator function.

send/config.c
```diff
@@ -80,6 +80,44 @@ static int smtp_auth_validator(const struct ConfigSet *cs, const struct ConfigDe
   return CSR_SUCCESS;
 }
 
+/**
+ * sendmail_validator - Validate the "sendmail" config * variable - Implements ConfigDef::validator() - @ingroup cfg_def_validator
+ */
+static int sendmail_validator(const struct ConfigSet *cs, const struct ConfigDef *cdef,
```
flatcap, line 86: We can rename this to, er... `command_validator`, maybe (see below)
(and adjust the comment)

send/config.c
```diff
@@ -232,7 +270,7 @@ static struct ConfigDef SendVars[] = {
   { "reverse_real_name", DT_BOOL|R_INDEX|R_PAGER, true, 0, NULL,
     "Set the 'From' from the full 'To' address the email was sent to"
   },
-  { "sendmail", DT_STRING|DT_COMMAND, IP SENDMAIL " -oem -oi", 0, NULL,
+  { "sendmail", DT_STRING|DT_COMMAND, IP SENDMAIL " -oem -oi", 0, sendmail_validator,
```
flatcap, line 273: Adding `|DT_NOT_EMPTY` will check for empty values.

flatcap (changes requested):
Nice job!  Nearly there.
There's a lot of words, but the changes will only take a few minutes :-)
Thanks

rafork:
I've named it simple_command_validator because the sh(1) manual entry uses the terms "simple command" and "complex command". It's not quite the same (real simple commands include leading name=value environment variables and I/O redirection), but I think it gives a good idea of the intent of the validator.

Many thanks for the encouragement and guidance. I usually accept patches and then fix them myself, but taking the time to introduce contributors to the ways of the code base is much wiser.

flatcap:
Rebased and squashed.
I made one slight change, because I'm an idiot!
I've dropped the `DT_NOT_EMPTY`, because `$sendmail` is _allowed_ to be empty.
What I should have suggested is adding this check to the validator:
```c
  const char *valstr = (const char *) value;
  if (!valstr)
    return CSR_SUCCESS;
```
If the string is empty, then the validator succeeds.

> I usually accept patches and then fix them myself

Yeah, that's always a temptation :-)

> Many thanks for the encouragement and guidance. 

You're welcome.
I'm always happy to help others help NeoMutt.

flatcap (approved):
Lovely job.
Thanks!

rafork:
Thanks. I thought $sendmail couldn't be empty because mutt_invoke_sendmail() contains a check that it is non-NULL "to avoid a crash".

Issue #181, opened 2016-10-04
Create a new variable $net_folder_format


flatcap:
`$folder_format` is used for both local file/dirs and network mailboxes, such as
IMAP.  If you have multiple mailboxes, some local, some remote, then 

I'd like you to introduce a new variable `$net_folder_format`
If the mailbox is remote and this variable is set, then use it to format the
folder information.  If the variable isn't set, then default to `$folder_format`

Local mailboxes use `$folder_format` as normal.

---

**Easy tasks would be perfect for someone new to NeoMutt.**
You won't need to write lots of code, and you won't need any previous knowledge
of the NeoMutt code.  Obviously, you will need to be able to write C code :-)

**Easy tasks also come with a mentor** -- a NeoMutt project member -- who will help
you learn what you need to complete the work.

If you're interested in having a go, leave a comment or speak to us in our
IRC channel: #neomutt on irc.freenode.net

Thank you :heart:

toogley:
@flatcap Should the new variable also use a global variable like `folder_format`'s [FolderFormat](https://github.com/neomutt/neomutt/blob/master/globals.h#L55) ?

flatcap:
Yes.  This is the simple fix.

The bigger (better) solution would be to refactor the browser to move a lot of the code into the mailboxes.
For each mailbox type, (e.g. maildir, imap, notmuch) work out what they'd want to display, then create a browser that knows nothing about its data.

toogley:
(I'm woring in [devel/net-folder-format](https://github.com/neomutt/neomutt/tree/devel/net-folder-format) on it, but am not finished yet.)

toogley:
On IRC, @flatcap and I discussed about `browser.c` and in particular the `folder_format_str()` method. I was proposing an extra method `net_folder_format_str()` and for handling the shared fields (field for Message of unread messages, etc.) in one method per field. The result would be an enormous list of small functions, so this post tries to collect the amount of those small functions. I think, small functions are more readable and this way we can also cut the very big method folder_format_str() a bit down.


**Note However**: This can only be an estimation. As i've made this list, i've seen multiple fields which seem similar (for instance: %n of pgp_entry_format and %C of e.g. folder_format), but i don't know if the implementation can actually be made shared. Furthermore, i'll try to extend this list in the future.

### multiple format definitions (two or more)

*(by "entry", i mean sth like a mailbox, a newsgroup or similar)*

* entry number  %C  (shared by NNTP, IMAP, local)
[...]

PR #3412, opened 2022-05-26
browser: add mailbox_folder_format config variable

charles2910:
Implementation of #181.

Adds a new format string for viewing non-local mailboxes in the browser.

- browser/browser.c: folder_make_entry: receive BrowserState as data
  and check if it's a mailbox to print it accordingly.
- browser/browser.c: mutt_buffer_select_file: pass BrowserState as data.
- browser/config.c: declare mailbox_folder_format.
- docs/config.c: document the new var.

For the default value, mailboxes should look like this:

Screenshot from 2022-05-24 23-34-34

### DIFF
--- a/browser/browser.c
+++ b/browser/browser.c
@@ -805,7 +805,8 @@ static int select_file_search(struct Menu *menu, regex_t *rx, int line)
  */
 static void folder_make_entry(struct Menu *menu, char *buf, size_t buflen, int line)
 {
-  struct BrowserStateEntry *entry = menu->mdata;
+  struct BrowserState *state = menu->mdata;
+  struct BrowserStateEntry *entry = &state->entry;
   struct Folder folder = {
     .ff = ARRAY_GET(entry, line),
     .num = line,
@@ -821,6 +822,14 @@ static void folder_make_entry(struct Menu *menu, char *buf, size_t buflen, int l
   }
   else
 #endif
+      if (state->is_mailbox_list)
+  {
+    const char *const c_mailbox_folder_format = cs_subset_string(NeoMutt->sub, "mailbox_folder_format");
+    mutt_expando_format(buf, buflen, 0, menu->win->state.cols,
+                        NONULL(c_mailbox_folder_format), folder_format_str,
+                        (intptr_t) &folder, MUTT_FORMAT_ARROWCURSOR);
+  }
+  else
   {
     const char *const c_folder_format = cs_subset_string(NeoMutt->sub, "folder_format");
     mutt_expando_format(buf, buflen, 0, menu->win->state.cols, NONULL(c_folder_format),
@@ -1313,7 +1322,7 @@ void mutt_buffer_select_file(struct Buffer *file, SelectFileFlags flags,
 
   init_menu(&priv->state, priv->menu, m, priv->sbar);
   // only now do we have a valid priv->state to attach
-  priv->menu->mdata = &priv->state.entry;
+  priv->menu->mdata = &priv->state;
 
   // ---------------------------------------------------------------------------
   // Event Loop
--- a/docs/config.c
+++ b/docs/config.c
@@ -2210,6 +2210,43 @@
 ** how often (in seconds) NeoMutt will update message counts.
 */
 
+{ "mailbox_folder_format", DT_STRING, "%2C %t %N %n %m %i" },
+/*
+** .pp
+** This variable allows you to customize the file browser display to your
+** personal taste. It's only used to customize network mailboxes (e.g. imap).
+** This string is similar to $$index_format, but has its own set of
+** \fCprintf(3)\fP-like sequences:
+** .dl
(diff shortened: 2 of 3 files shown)


browser/browser.c
```diff
@@ -821,6 +822,14 @@ static void folder_make_entry(struct Menu *menu, char *buf, size_t buflen, int l
   }
   else
 #endif
+  if (state->is_mailbox_list)
```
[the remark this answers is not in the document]
flatcap: Yeah, I can see why it thinks that.
If the `#ifdef` ... `#endif` weren't there, it'd be clear: the `if` is subordinate to the `else` (L823).

Some day I'll factor our all the places where `#ifdef`s interfere with `if`s.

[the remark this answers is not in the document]
charles2910: should I indent or not it?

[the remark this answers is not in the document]
flatcap: Yes, indent it please.
`clang-format` isn't perfect, but it prevents arguments :-)

flatcap:
squashed, rebased

flatcap (approved):
Very nice!
Thanks

flatcap:
I just sent you an email to accompany the NeoMutt invite.
Unfortunately, Outlook rejected it.  _sigh_

gahr:
Sorry, I'm late: would we be fine with not having a default and falling back to `folder_format` if `mailbox_folder_format` is not set? Just for POLA's sake and not to change everybody's mailbox view without notice.

docs/config.c
```diff
@@ -2210,6 +2210,43 @@
...
+**                   "@" to symbolic links and "*" to executable files)
+** .dt %F  .dd   .dd File permissions
+** .dt %g  .dd   .dd Group name (or numeric gid, if missing)
+** .dt %i  .dd   .dd Description of the folder
+** .dt %l  .dd   .dd Number of hard links
+** .dt %m  .dd * .dd Number of messages in the mailbox
+** .dt %n  .dd * .dd Number of unread messages in the mailbox
+** .dt %N  .dd   .dd "N" if mailbox has new mail, blank otherwise
+** .dt %s  .dd   .dd Size in bytes (see $formatstrings-size)
+** .dt %t  .dd   .dd "*" if the file is tagged, blank otherwise
```
gahr, line 2234: It looks to me like mailboxes can't be tagged.

docs/config.c
```diff
@@ -2210,6 +2210,43 @@
 ** how often (in seconds) NeoMutt will update message counts.
 */
 
+{ "mailbox_folder_format", DT_STRING, "%2C %t %N %n %m %i" },
```
gahr, line 2213: If we go for a default, I'd avoid duplicating the new info with both %N and %n, how about a conditional, e.g. `%2C %?n?%6n&%6 ? %6m %i`?

flatcap: yes, that looks better

docs/config.c
charles2910: Should we remove the options shared with `folder_format_str()` that can't be used in mailboxes from the documentation?

docs/config.c
charles2910: I'm good with `%2C %?n?%6n&%6 ? %6m %i`. I did suggest with `%N` because it's easier *for me* to spot folders with new mail.

charles2910:
>  would we be fine with not having a default and falling back to folder_format if mailbox_folder_format is not set? Just for POLA's sake and not to change everybody's mailbox view without notice.

Is there a way to inform users about the change? From a newbie standpoint (me), I'd rather have the mailboxes shown with the new `mailbox_folder_format`, but I think the change would bother more advanced users.

docs/config.c
flatcap: No, we can remove this line of doc.
NeoMutt will still accept `%t`, but it'll always be empty.

charles2910: Are there other fields that can be removed too (like gid, uid, etc)? Because then I can make the changes altogether

flatcap:
- Remove `%t` from docs
- Change default for `mailbox_folder_format`

[[devel/browser-fix]](https://github.com/neomutt/neomutt/tree/devel/browser-fix)

charles2910:
> * Remove `%t` from docs
> 
>     * Change default for `mailbox_folder_format`
> 
> 
> [[devel/browser-fix]](https://github.com/neomutt/neomutt/tree/devel/browser-fix)

Hum, I don't know how to comment in commits (like you did with mine a long time ago :), but we have to change the dafault value for `mailbox_folder_format` in the docs too.

flatcap:
Good point!  Fixed.

> I don't know how to comment in commits

If you click on the branch link, just above the source, you'll see a line for the last commit:
```
@flatcap browser: change mailbox_folder_format default               83e7499   11 minutes ago
```

If you click on the commit hash, you'll see the diff, which you can comment on (hover and click on the `+`)

PR #3408, opened 2022-05-16
Implement account_command

gahr:
This implements the proposal spec'd in #3405: it introduces a new
`account_command` config variable that can be set to the path of an
executable used to retrieve account credentials.

The following arguments to the script are produced:

* `--hostname`
* `--username`
* `--type` (`imap`, `pop`, `smtp`, `nntp`, with an optional trailing `s` for ssl)

The following keys are parsed from the command's output

* `login`
* `username`
* `password`

As an example, here's an account command that wraps [bwget](https://github.com/gahr/bwget):

```tcl

set entries {
    imap.fastmail.com     { me@fastmail.com   "^FastMail-Me$"
                            you@fastmail.com  "^FastMail-You$" }
    smtp.fastmail.com     { me@fastmail.com   "^FastMail-Me$"
                            you@fastmail.com  "^FastMail-You$" }
    outlook.office365.com { me@hotmail.com    "^Hotmail$"      }
    smtp.office365.com    { me@Hotmail.com    "^Hotmail$"      }
    imap.gmail.com        { me@gmail.com      "^Google$"       }
    smtp.gmail.com        { me@gmail.com      "^Google$"       }}

set host {}
set user {}
foreach {key val} $::argv {
    switch -exact -- $key {
        --hostname { set host $val }
        --username { set user $val }
    }
}

proc pass {entry} {
    set pfd [open "| bwget pass $entry"]
    set pass [read $pfd]
    close $pfd
    set pass
}

if {$host ne {} && $user ne {}} {
    set hostidx [lsearch -exact -stride 2 $entries $host]
    if {$hostidx != -1} {
        set sub [lindex $entries $hostidx+1]
        set useridx [lsearch -exact -stride 2 $sub $user]
        if {$useridx != -1} {
            puts "username: $user"
            puts "login: $user"
            puts "password: [pass [lindex $sub $useridx+1]]"
        }
    }
}
```

To use this, just set `imap_url` and `smtp_url` to `user@host`, and don't set
any of `imap_user`, `imap_pass`, `smtp_user`, or `smtp_pass`.

PR #3414, opened 2022-05-29
Mac OSX Keychain support via account-command

the-x-at:
A small Python 3 wrapper program to access the operating system's Keychain via the `security` command. Details for setting up NeoMutt to use this program are contained in `contrib/account-command/README.md`.

### What does this PR do?

Using the recently implemented [account command](https://neomutt.org/feature/account-cmd)  (#3408), this is a Mac OSX specific implementation to use the internal password manager, Keychain Access. A small Python 3 program called by NeoMutt obtained from the `account_command` variable calls in turn the `security` program to obtain a password from the Keychain that is passed on back to NeoMutt.

For example, if you want to retrieve an IMAP email account `me@example.com` using SSL/TLS from `imap.example.com`, first generate an entry in the Keychain via

```
security add-internet-password -l imap-example -a me@example.com -s imaps://imap.example.com
```

and set the password by adding `-w passphrase` to the above command (not recommended) or using Keychain Access GUI utility (recommended). Setting NeoMutt's variables to 

```
set folder = "imaps://me@example.com@imap.example.com"
set account_command = "/usr/share/doc/neomutt/account-command/keychain.py"
```

will instruct NeoMutt to use the newly implemented `keychain.py` program to retrieve the password when checking mail on `imap.example.com` for user `me@example.com`.

### Documentation

Extensive documentation can be found in `contrib/account-command/README.md`. The document is structured in a way such that future password managers can be added easily.

### Testing

- The program was tested on OSX 10.12.6.
- Retrieved email from an IMAP test account with SSL/TLS.
- Sent email through that test account.
- Installation procedure worked and program was tested with so-installed version of NeoMutt.

### Prerequisites

- Python 3 standard installation.
- `/usr/bin/security` program.

### DIFF
--- a/contrib/Makefile.autosetup
+++ b/contrib/Makefile.autosetup
@@ -1,4 +1,4 @@
-CONTRIB_DIRS=	colorschemes hcache-bench keybase logo lua oauth2 samples vim-keys
+CONTRIB_DIRS=	colorschemes hcache-bench keybase logo lua oauth2 samples vim-keys account-command
 
 all-contrib:
 clean-contrib:
@@ -13,6 +13,7 @@ install-contrib:
 		done \
 	done
 	chmod +x $(DESTDIR)$(docdir)/keybase/*.sh
+	chmod +x $(DESTDIR)$(docdir)/account-command/*.py
 
 uninstall-contrib:
 	for d in $(CONTRIB_DIRS); do \
--- a/contrib/account-command/README.md
+++ b/contrib/account-command/README.md
@@ -1,0 +1,111 @@
+# Introduction
+
+NeoMutt provides a general mechanism to retrieve password information rather
+than storing these sensitive data in the config files using `imap_pass`,
+`smtp_pass`, and other `*_pass` variables.
+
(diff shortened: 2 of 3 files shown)


flatcap (approved):
LGTM
Time to find some Mac users to give it a try

flatcap:
Oh... One oddity.
GitHub doesn't seem to recognise the commit as belonging to @the-x-at 
You can add private email addresses, [here](https://github.com/settings/emails),
which means you'll get extra green squares, [here](https://github.com/the-x-at) :-)

the-x-at:
Ah, thanks for the hint. After committing I have seen that my work address popped up due to an entry in my global git config. Should be fixed when I am performing my next NeoMutt contribution. ;-)

flatcap:
It's easily fixed.
Feel free to `git commit --amend` and `git push --force`

the-x-at:
Thanks, @flatcap, you're a pal. Followed your advice, worked! Sorry for the inconvenience.

Issue #3415, opened 2022-05-29
Duplicated functions: `<refresh>` vs `<redraw-screen>`


darrel-glud:
## Expected Behaviour

Consistent UI and/or proper documentation.

## Actual Behaviour

The generic menu has `<refresh>` (bound to `^L`) with the description "clear and redraw the screen", while in the pager menu has `<redraw-screen>` (bound also to `^L`) with the same description.
A quick test suggests that the pager does not have a `<refresh>` and index does not have `<redraw-screen>`.

Thus: What is the difference between the two?
If there is one, that should be documented.
If there is none, I consider this a bug and one of the two should be deprecated (I suggest deprecating `<refresh>` because of the less descriptive name) and eventually be removed. Until then I suggest introducing an alias for backwards compatibility.

flatcap:
Good spot!
This is one of those legacy problems that are proving difficult to undo.

The first Mutt commit (24 years ago) contains this distinction.
https://github.com/neomutt/upstream-mutt/blob/1a5381e07e97fe482c2b3a7c75f99938f0b105d4/functions.h

The `generic` menu roughly matches `struct Menu` in the code -- it's behind all the lists, like the Index.
Because the Pager doesn't have a list, it doesn't use the `generic` menu.
But, it still needed a `<refresh>` function.
`$deity` knows why they chose a different name for it.

I'd like to get rid of the `generic` menu altogether, but it's tricky.
I recently moved one function out of `generic` into the separate menus that needed it.
That needed quite a few workarounds.

For now, I'll add a `<refresh>` alias for the Pager and kill any references to `<redraw-screen>`.

darrel-glud:
Thanks @flatcap for the archaeological tour (quite a lot of empty lines there...).
I created PR #3418 which is a naive attempt of doing that.

darrel-glud:
@flatcap wrote:
> I'd like to get rid of the generic menu altogether, but it's tricky.
> I recently moved one function out of generic into the separate menus that needed it.
> That needed quite a few workarounds.

Interesting thought.  While searching for the `<refresh>` stuff, I noticed that `<show-version>`, which is in Pager and Index but not in other menus and which I would have moved into the generic menu so it is accessible everywhere.
(I also noticed `<exit>` and `<quit>` all over the place but I better not touch those.)

flatcap:
> I noticed that `<show-version>`, which is in Pager and Index but not in other menus

Yes.
It gets a bit tricky with key-bindings.
[...]

PR #3418, opened 2022-05-31
Deprecate <refresh> function in favour of <redraw-screen>

darrel-glud:
* **What does this PR do?**

Promote <redraw-screen> into the generic menu and make <refresh> an
alias for it.

Disclaimer: This is mostly untested and I have also no clue how all that function-stuff works but it seems reasonable.

   - Manual build fails because I used `<redraw-screen>` in the function description which is interpreted as a tag.  Solutions welcome. (I believe using `&lt` would produce a correct manual but a literal `&lt` at the online help of neomutt.)

#3415

### DIFF
--- a/functions.c
+++ b/functions.c
@@ -323,7 +323,8 @@ const struct MenuFuncOp OpGeneric[] = { /* map: generic */
   { "previous-entry",                OP_PREV_ENTRY },
   { "previous-line",                 OP_PREV_LINE },
   { "previous-page",                 OP_PREV_PAGE },
-  { "refresh",                       OP_REDRAW },
+  { "redraw-screen",                 OP_REDRAW },
+  { "refresh",                       OP_REFRESH_DEPRECATED },
   { "search",                        OP_SEARCH },
   { "search-next",                   OP_SEARCH_NEXT },
   { "search-opposite",               OP_SEARCH_OPPOSITE },
@@ -640,7 +641,6 @@ const struct MenuFuncOp OpPager[] = { /* map: pager */
 #ifdef USE_NNTP
   { "reconstruct-thread",            OP_RECONSTRUCT_THREAD },
 #endif
-  { "redraw-screen",                 OP_REDRAW },
   { "reply",                         OP_REPLY },
   { "resend-message",                OP_RESEND },
   { "root-message",                  OP_MAIN_ROOT_MESSAGE },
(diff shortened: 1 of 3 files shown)


functions.c
```diff
@@ -640,7 +641,6 @@ const struct MenuFuncOp OpPager[] = { /* map: pager */
 #ifdef USE_NNTP
   { "reconstruct-thread",            OP_RECONSTRUCT_THREAD },
 #endif
-  { "redraw-screen",                 OP_REDRAW },
```
flatcap, line 643: I think we need this.
Because the Pager doesn't use the generic menu, it needs its own "redraw-screen".

functions.c
```diff
@@ -1184,7 +1184,6 @@ const struct MenuOpSeq PagerDefaultBindings[] = { /* map: pager */
   { OP_PRINT,                              "p" },
   { OP_QUIT,                               "Q" },
   { OP_RECALL_MESSAGE,                     "R" },
-  { OP_REDRAW,                             "\014" },           // <Ctrl-L>
```
flatcap, line 1187: we need this

functions.c
```diff
@@ -323,7 +323,8 @@ const struct MenuFuncOp OpGeneric[] = { /* map: generic */
   { "previous-entry",                OP_PREV_ENTRY },
   { "previous-line",                 OP_PREV_LINE },
   { "previous-page",                 OP_PREV_PAGE },
-  { "refresh",                       OP_REDRAW },
+  { "redraw-screen",                 OP_REDRAW },
+  { "refresh",                       OP_REFRESH_DEPRECATED },
```
flatcap, line 327: If you move the original, "refresh" to the deprecated section below, that should be enough.

flatcap (changes requested):
Thanks for the PR, @darrel-glud.
I think it goes a little too far, though :-)

NeoMutt has deprecated quite a few things since its split from Mutt, but we don't warn or error for any of them yet.
If a user migrates (upgrades) from Mutt to NeoMutt, we'd like it to be seamless.

I think what we need it probably just a couple of small changes to `OpGeneric`.
I've made a lot of comments.  I hope they make sense.

Thanks

darrel-glud:
Updated, thanks for the comments.
Kind of surprising that Pager doesn't use the GenericOp but well  ¯\_(ツ)_/¯

gahr:
Sorry if this has been discussed already, but wasn’t the plan to move away from generic?

flatcap:
> Kind of surprising that Pager doesn't use the GenericOp

Yeah, just one of those design decisions made long ago

> wasn’t the plan to move away from generic?

Yes, it still is (as soon as I figure out how to do that without breaking everything).
For now, removing confusion will have to do.

PR #3430, opened 2022-06-09
Add GPG+JSON-based account command

gahr:
* **What does this PR do?**

This adds a very simple solution for using the [account_command](https://neomutt.org/feature/account-cmd) feature.

~I'm wondering whether we should reorganize the `contrib/account-command` into per-solution directories.~ I did this.


contrib/account-command/gpg-json/README.md
```diff
@@ -0,0 +1,38 @@
...
+
+```sh
+$ gpg -r me@example.com -e credentials.json && rm credentials.json
+```
+
+This will generate a `credentials.json.gpg` encrypted file.
+
+### Setup your account command
+
+Place the `credentials.sh` shell script alongside your `credentials.json.gpg` file.
```
flatcap, line 37: Hmm...
I'm not sure that we should encourage users to **copy** the script.
If we ever updated/improved the script, they wouldn't benefit.

flatcap: Is `contrib`:
- a suggestion: Here's how you _could_ do it
- an example: Use this working script

flatcap: Could we add some sample config:
(everyone like to cut'n'paste, if it needs tweaking afterwards)

```
set account_command = "/usr/share/doc/neomutt/account-command/gpg-json/credentials.sh"
```

contrib/account-command/macos-keychain/README.md
```diff
@@ -0,0 +1,101 @@
...
+
+The same commands work with the `smtps` account, too.
+
+In your NeoMutt configuration file, be sure to remove any `*_pass` variables
+and set your folder and smtp URLs accordingly, e.g.:
+
+```
+set folder = "imaps://me@example.com@imap.example.com"
+set smtp_url = "smtps://me@example.com@smtp.example.com"
+set account_command = "/usr/share/doc/neomutt/account-command/keychain.py"
```
flatcap, line 88: Needs adjusting for new location

contrib/account-command/gpg-json/credentials.sh
```diff
@@ -0,0 +1,17 @@
...
+        "--hostname") shift; host="$1";;
+        "--username") shift; user="$1";;
+    esac
+    shift
+done
+
+[ "$host" = "" ] || [ "$user" = "" ] && exit 1
+
+query=$(printf '.host == "%s" and .user == "%s"' "$host" "$user")
+gpg -qd credentials.json.gpg | jq -r ".[] | select($query) | .pass"
```
flatcap, line 17: Can we make `credentials.json.gpg` a parameter?
Will that work with the `$account_command`?
(can it handle extra params?)

flatcap (changes requested):

PR #3434, opened 2022-06-12
muttlib/mutt_expando_format: allow for longer expansions in e.g. index_format

proycon:
* **What does this PR do?**

It allows expansion strings not to be limited to 256 characters.

In my index, I like to use a lot of icons and colors to categorize my messages based on the notmuch tags they get, this results in a very long `index_format` string because I'm not using `%g`, but am individually enumerating all the tags as `%Gx`.

My currrent index_format:

``
set index_format='%4C %Z %<[y?%<[m?%<[d?%10[     %H:%M  ]&%10[%a %b %d  ]>&%12[%b %d  ]>&%10[%b %d, %Y]> %-20.20L %?E?%2.e/%2.E&     ? %?Gi?%Gi&?%?GS?%GS&?%?Gr?%Gr&?%?Gt?%Gt&?%?GI?%GI&?%?GL?%GL&?%?Gl?%Gl&?%?GW?%GW&?%?GD?%GD&?%?Gc?%Gc&?%?Gs?%Gs&?%?Ge?%Ge&?%?Gp?%Gp&?%?Gw?%Gw&?%?Gf?%Gf&?%?Gn?%Gn&?%?Gu?%Gu&?%?Gg?%Gg&?%?Gy?%Gy&?%?GP?%GP&?%?Gh?%Gh&?%?GF?%GF&?%?G2?%G2&?%?GT?%GT&?%?GM?%GM&? %s'
``

This didn't render correctly as the index_format was clipped off at 256, I now increased this to 1024 (might not be the most sustainable solution but at least gives some more breathing room)

After the fix: 

image

Before the fix only a few icons showed and %s was cut off.

I haven't seen any issues about this yet

### DIFF
--- a/muttlib.c
+++ b/muttlib.c
@@ -785,7 +785,7 @@ void mutt_expando_format(char *buf, size_t buflen, size_t col, int cols, const c
   FILE *fp_filter = NULL;
   char *recycler = NULL;
 
-  char src2[256];
+  char src2[1024];
   mutt_str_copy(src2, src, mutt_str_len(src) + 1);
   src = src2;
 


flatcap:
> This didn't render correctly as the index_format was clipped off at 256

Oops!  Sorry about that!

> I now increased this to 1024 (might not be the most sustainable solution...

Yeah.
Refactoring (rewriting) `mutt_expando_format()` is on my todo list, but it's a scary proposition.

flatcap (approved):
Thanks for the fix! :heart:

proycon:
This PR might have caused a regression; for some reason I don't understand yet, index navigation is way slower now...

flatcap:
> index navigation is way slower now

Hmm...  
It seems very unlikely that _your_ change caused any problems.

Slower, how?
- Loading the index?
- Moving the cursor from email to email?

What version were you using before your fix?

If you `:reset index_format` are things fast again?

proycon:
> It seems very unlikely that your change caused any problems.

That's what I thought too, being so minor... but it seems I can reproduce it now because when I set it back to 256 things are snappy again (though that clips off my index format).

> Moving the cursor from email to email?

Yes, and the lag seems somewhat proportional to how much vertical screen estate neomutt gets

> What version were you using before your fix?

I've tried this on the main branch, but also on 20220415 (with and without my patch), and it seems my patch somehow ruins it..

I also tried on two different systems (an ubuntu one and an alpine linux one (musl))

> If you :reset index_format are things fast again?

Nope, I had tried that too, that strangely makes no difference at all. 

I'm rather puzzled.

proycon:
Hmm, if I leave it at 1024 but only remove the final `%s`  from my index_format, then the problem doesn't occur..

If I then reset the index_format to the default (which is shorter than 256 chars even), the slow index navigation does occur again!

proycon:
I think I found the culprit now: I have a lot of rules for coloring the subject (`color index_subject`, see  https://github.com/proycon/dotfiles/blob/master/muttcol#L19), some of them fairly complex .. when I disable all of those, the slowed navigation problem no longer occurs. 

Because my subject never rendered at 256 chars, the problem didn't occur then. So this PR is fine.

There might still be a regression elsewhere as I've had those rules for a while but didn't experience that much lag on an older version (20191207).

flatcap:
> So this PR is fine

Great!
Thanks for taking the time to debug this.

Issue #3394, opened 2022-05-02
`debug_file` ignored


balejk:
## Expected Behaviour

Launching `neomutt -d 2` will log to `~/.config/neomutt/debug0` with `set debug_file = "~/.config/neomutt/debug"` in `neomuttrc`.

## Actual Behaviour

`neomutt` logs to `~/.neomuttdebug0`.

## Steps to Reproduce

Specify custom `debug_file` in `neomuttrc` and launch `neomutt -d 2`. Log is written to the default location.

## How often does this happen?

- Always

## NeoMutt Version

```
NeoMutt 20220415
Copyright (C) 1996-2022 Michael R. Elkins and others.
NeoMutt comes with ABSOLUTELY NO WARRANTY; for details type 'neomutt -vv'.
NeoMutt is free software, and you are welcome to redistribute it
under certain conditions; type 'neomutt -vv' for details.

System: Linux 5.15.34_1 (x86_64)
ncurses: ncurses 6.3.20211021 (compiled with 6.3.20211021)
libidn: 1.38 (compiled with 1.38)
GPGME: 1.15.1
OpenSSL: OpenSSL 1.1.1n  15 Mar 2022
libnotmuch: 5.6.0
storage: rocksdb, gdbm, tdb
compression: zlib

Configure options: --ssl --gpgme --notmuch --gdbm --lua --sasl --zlib --tdb --rocksdb --testing

Compilation CFLAGS: -fstack-clash-protection -D_FORTIFY_SOURCE=2 -mtune=generic -O2 -pipe  -D_FILE_OFFSET_BITS=64 -DLARGEFILE_SOURCE=1 -g -fdebug-prefix-map=/builddir/neomutt-20220415=. -std=c99 -fno-delete-null-pointer-checks -D_ALL_SOURCE=1 -D_GNU_SOURCE=1 -D__EXTENSIONS__ -I/usr/include -I/usr/include/lua5.4 -DNCURSES_WIDECHAR

Default options:
  +attach_headers_color +compose_to_sender +compress +cond_date +debug 
  +encrypt_to_self +forgotten_attachments +forwref +ifdef +imap +index_color 
  +initials +limit_current_thread +multiple_fcc +nested_if +new_mail +nntp +pop 
  +progress +quasi_delete +regcomp +reply_with_xorig +sensible_browser +sidebar 
  +skip_quoted +smtp +status_color +timeout +tls_sni +trash 

Compile options:
[...]

PR #3440, opened 2022-06-18
Fix debug_file ignored #3394

Spacefreak18:
* **What does this PR do?**

     for that)

### DIFF
--- a/main.c
+++ b/main.c
@@ -730,11 +730,7 @@ main
   }
 
   mutt_log_prep();
-  if (dlevel)
-    mutt_log_start();
-
   MuttLogger = log_disp_queue;
-
   log_translation();
 
   if (!STAILQ_EMPTY(&cc_list) || !STAILQ_EMPTY(&bcc_list))


flatcap:
:heart: Thanks for PR, @Spacefreak18

That looks about right; I'll have a closer look later.

main.c
```diff
@@ -729,14 +729,6 @@ main
     cs_str_reset(cs, "debug_level", NULL);
   }
 
-  mutt_log_prep();
-  if (dlevel)
-    mutt_log_start();
-
-  MuttLogger = log_disp_queue;
-
-  log_translation();
```
flatcap, line 738: We only want to kill the `mutt_log_start()`

main.c
```diff
@@ -787,7 +779,7 @@ main
     TsSupported = mutt_ts_capability();
     rootwin_set_size(COLS, LINES);
   }
-
+
```
flatcap, line 782: Urgh!
Check your editor settings -- many have an option to kill trailing whitespace

main.c
```diff
@@ -800,7 +792,19 @@ main
...
+  if (!dfile)
+  {
+    dfile = cs_subset_path(NeoMutt->sub, "debug_file");
+    cs_str_initial_set(cs, "debug_file", dfile, NULL);
+  }
+  
+  MuttLogger = log_disp_queue;
+  mutt_log_prep();
+  if (dlevel)
+    mutt_log_start();
```
flatcap, line 804: We don't need to do this, as it's done unconditionally a few lines later.

main.c
```diff
@@ -800,7 +792,19 @@ main
     cs_str_reset(cs, "debug_level", NULL);
   if (dfile)
     cs_str_reset(cs, "debug_file", NULL);
-
+  if (!dfile)
+  {
+    dfile = cs_subset_path(NeoMutt->sub, "debug_file");
+    cs_str_initial_set(cs, "debug_file", dfile, NULL);
+  }
```
flatcap, line 799: Ah!  We don't want this.
The config variables have default values, set in the code.
A special few of them have their defaults set at startup from environment variables or command line options,
e.g.

https://github.com/neomutt/neomutt/blob/34dc388b40666eb4d0db66596455330fc2664512/mutt_config.c#L242

Running `neomutt -l apple` will set the default, so that `reset debug_file` will revert to that command line value.

If the user has the config `set debug_file = "apple"` then `reset debug_file` will revert to `~/.neomuttdebug0`

flatcap:
I've pushed a commit to minimise your fix.
(It still fixes the original problem)

If you have any questions, let me know.
Thanks

PR #3448, opened 2022-06-23
Refactor auto-completion code

flatcap:
This is a small step forward towards an auto-completion API.

- 0ccb31556 complete: kill global data
  Rather than using six global variables, create `struct CompletionData` and pass it to the functions

- 7d023452a complete: unify naming
  Clean up the member naming

- 6d49fd2d9 complete: strip notmuch
  Refactor the Notmuch completion to use the same array as other completions.
  The other completion code fills the array with strings that are owned by other structs.
  In the Notmuch case, we need to free them after use.

### DIFF
--- a/Makefile.autosetup
+++ b/Makefile.autosetup
@@ -198,7 +198,7 @@ $(PWD)/color:
 ###############################################################################
 # libcomplete
 LIBCOMPLETE=	libcomplete.a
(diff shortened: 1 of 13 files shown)


flatcap:
**Key**:
- Green: `op_editor_complete()` - create the `struct CompletionData`
- Red: Functions using the data

Issue #3460, opened 2022-07-04
<next-unread-message> not expanding "on-demand" threaded messages in the index


fbellet:
In version 20211022, there was a feature that disappeared in 20220408, and never been restored since:

When navigating the index with the `<Down>` arrow (`<next-undeleted-message>`) while the pager is open, all messages were displayed in the pager, even those collapsed inside a thread (the index was recomputed, and the related thread was expanded in the index, allowing me to see all messages when progressing with `<next-undeleted-message>`).

Commit 627342bc81e changed that: the pager is closed, when `<next-undeleted-message>` is executed.
Commit 19869c7c8b9 partially reverted this behavior: the pager remains open, but collapsed threads remain collapsed in the index, and it seems to me that there is now no easy way to walk inside a thread.

Would it be possible to come back to pre commit 627342bc81e behavior, that seems the most intuitive to me: to expand the threads in the index while we navigate inside a mbox with the up/down arrow?

flatcap:
Thanks for investigating.
That's definitely a bug.

There used to be a lot of code duplicated in the Index and Pager.
I refactored them so that each NeoMutt `<function>` was handled by one C `function()`,
but clearly something got lost in translation.

flatcap:
> Other functions like `<delete-message>` probably have the same tweak.

Yes.  All the functions that call `resolve_email(priv->menu, shared, RESOLVE_NEXT_UNDELETED);`

- `op_delete()`
- `op_delete_thread()`
- `op_edit_label()`
- `op_flag_message()`
- `op_main_modify_tags()`
- `op_main_set_flag()`
- `op_save()`
- `op_toggle_new()`

flatcap:
Not unusual; in fact there's a config variable for that: (since b0d7420c8)

```sh
# Collapse all threads when entering a folder
set collapse_all = yes
```

adrien-n:
I've had this issue too (and we discussed a bit on IRC a few days ago). This change has greatly improved the situation (thanks again!) but I think there is still an issue. The typical situation for me involves mailing-lists.

I think the issue is basically that uncollapse_jump is not abided by. Full reproducer below in case I misunderstood something:

- Receive one message (call it 'X') in thread A, read it;
- Receive one message in thread B (not sure this one needs to be in a thread);
- Receive one message (call it 'Y') in thread A;
- Now the message order is thread B, then thread A with X (read) and Y (unread);
- Read B and move to A: message 'X' is opened instead of 'Y'.

PR #3462, opened 2022-07-05
pager: fix next/prev behaviour wrt collapsed threads

flatcap:
Some functions when used from the Pager, cause threads to be uncollapsed:

- `<delete-message>`
- `<delete-thread>`
- `<edit-label>`
- `<flag-message>`
- `<modify-labels>`
- `<next-undeleted>`
- `<previous-undeleted>`
- `<save-message>`
- `<set-flag>`
- `<toggle-new>`

This behaviour was broken by the Index/Pager refactoring.

Fixes: #3460 

---

Change functions to optionally uncollapse a thread:

- `ci_next_undeleted()`
- `ci_previous_undeleted()`
- `resolve_email()`

Whether or not to uncollapse is determined by whether the Index is focussed, or not.

### DIFF
--- a/index/dlg_index.c
+++ b/index/dlg_index.c
@@ -217,16 +217,34 @@ void collapse_all(struct MailboxView *mv, struct Menu *menu, int toggle)
 }
 
 /**
- * ci_next_undeleted - Find the next undeleted email
+ * uncollapse_thread - Open a collapsed thread
  * @param m     Mailbox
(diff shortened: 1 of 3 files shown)


gahr:
right, `resolve_email` wouldn't work.. can we please factor out common code at least?

flatcap:
Factor out common code

gahr (approved):

Issue #3464, opened 2022-07-07
neomutt hangs when trying to add email address from help screen


jpalus:
## Expected Behaviour

Nothing happens or error is printed.

## Actual Behaviour

neomutt hangs and eats 100% CPU.

## Steps to Reproduce

```
$ cat test-neomuttrc 
macro pager a     "|abook --add-email\n" 'add sender to abook'
$ neomutt -F test-neomuttrc
```
Hit `?` for help screen and `a`.

## How often does this happen?

- Always

## NeoMutt Version

```
$ neomutt -v
NeoMutt 20220429
Copyright (C) 1996-2022 Michael R. Elkins and others.
NeoMutt comes with ABSOLUTELY NO WARRANTY; for details type 'neomutt -vv'.
NeoMutt is free software, and you are welcome to redistribute it
under certain conditions; type 'neomutt -vv' for details.

System: Linux 5.18.3-1 (aarch64)
ncurses: ncurses 6.3.20220416 (compiled with 6.3.20211120)
libidn: 1.41 (compiled with 1.38)
GPGME: 1.17.1
OpenSSL: OpenSSL 3.0.2 15 Mar 2022
storage: bdb, lmdb
compression: lz4, zlib, zstd

[...]

PR #3465, opened 2022-07-07
help: disable index functions

flatcap:
WIP: This solves the problem, but it may need work...

- 700b91d9f help: disable index functions
  When in the Help Screen (pager) disable the Index and Sidebar functions.

- cb68f49de pager: clear input buffer on error
  If the function isn't recognised, this could be a broken macro.
  Clear the input buffer to prevent accidents.

- c6aafb201 pager: only clear input buffer for help
  Tighten up the check.
  The macro behaviour when viewing emails is unchanged.

Fixes: #3464

### DIFF
--- a/pager/dlg_pager.c
+++ b/pager/dlg_pager.c
@@ -564,6 +564,10 @@ int mutt_pager(struct PagerView *pview)
       priv->rc = op;
       break;
     }
+
+    if ((pview->mode != PAGER_MODE_EMAIL) && (rc == FR_UNKNOWN))
+      mutt_flushinp();
+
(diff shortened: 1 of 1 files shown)


flatcap:
The fundamental problem is that macros are **literal**.
If `<pipe-entry>` fails, then `my_script.sh` will be interpreted as keystrokes.

In Vim, this is irritating, but you can at least undo mistakes.

Some ideas for fixing this were discussed in #69.
They might allow macros to work, or fail, in a predictable manner.

gahr (approved):

PR #3468, opened 2022-07-10
Handle corrupted header caches

stoeckmann:
* **What does this PR do?**

Check local cache files while parsing them to prevent undefined behavior
like out of boundary reads.

No issue created. It can be tested by compiling with asan, gdbm, and zlib
(or lz4) and modifying the relevant store functions to write only 1 byte.

Required configuration: `set header_cache = "~/.mutt/cache"`

### DIFF
--- a/compress/lz4.c
+++ b/compress/lz4.c
@@ -29,6 +29,7 @@
 
 #include "config.h"
 #include <stddef.h>
+#include <limits.h>
 #include <lz4.h>
 #include "private.h"
 #include "mutt/lib.h"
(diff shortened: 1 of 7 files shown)


flatcap:
More checks on the hcache are always welcome.
Thank you very much :heart:

I've pushed some trivial tweaks:
- [clang-format](https://neomutt.org/dev/analysis/clang-format)
- added some `()`s ([code style](https://neomutt.org/dev/code))

flatcap (approved):

PR #3469, opened 2022-07-11
source: fix variable interpretation

flatcap:
Change `source` to look for NeoMutt variables first.

---

The `source` and `set` commands differed in how they interpreted a
$variable name.

    set my_c = `echo $my_dir/test.rc`

`set` would look for a NeoMutt variable called `my_dir` and failing that
look for an extern environment variable with that name.

    source `echo $my_dir/test.rc`

`source` only looked for an extern environment variable.


gahr (approved):
Makes sense!

PR #3470, opened 2022-07-11
Document `source` command more thoroughly

whitney-cumber:
Document that the `source` command accepts multiple filenames.

Document how `source` handles relative file names.


flatcap (commented):
That looks good.  Thanks!

I'm just going to wait before merging because @mjsir911 has created #3471, to fix a different problem with `source`, that might affect the docs.

gahr (approved):
I think this is ready to be merged?

PR #3471, opened 2022-07-12
Allow `source` in hooks to point to relative paths

### LINKED ISSUE #1557: [neomuttrc] Cannot source file from folder-hook using relative path
When try to set up a folder-hook to source a file from `~/.config/neomutt/`, the hook fails with the following error:
```
Error: issue converting path to absolute (No such file or directory)
```
And the debug log shows:
```
[2019-02-10 20:45:15]<E> source_rc() Error: impossible to build path of 'account.mail'.
[2019-02-10 20:45:15]<E> mutt_folder_hook() source: file account.mail could not be sourced.
```
To reproduce, I used the following minimal test case:
**~/.config/neomutt/test.rc**
```
folder-hook account 'source account.mail'
set mbox_type=Maildir
set folder="~/mail"
mailboxes = `find ~/mail -mindepth 1 -type d ! -name cur ! -name tmp ! -name new -printf '"%p"\0' | xargs -0`
source account.mail
```
**~/.config/neomutt/account.mail**
```
set spoolfile = "+account/Inbox"
```
When I run `neomutt -F ~/.config/neomutt/test.rc` the **account.mail** file gets sourced correctly the first time by the bare source call, but then the folder-hook is triggered and fails.
This setup works correctly when I give an absolute path to the folder-hook command.

mjsir911:
Fixes #1557

### DIFF
--- a/command_parse.c
+++ b/command_parse.c
@@ -136,6 +136,48 @@ int parse_grouplist(struct GroupList *gl, struct Buffer *buf, struct Buffer *s,
 }
 
 /**
+ * mutt_parse_rc_line_cwd - Parse and run a muttrc line in a relative directory
+ * @param line   Line to be parsed
+ * @param cwd    File relative where to run the line
+ * @param err    Where to write error messages
+ * @retval #CommandResult Result e.g. #MUTT_CMD_SUCCESS
+ */
+enum CommandResult mutt_parse_rc_line_cwd(const char *line, char *cwd, struct Buffer *err)
+{
+  mutt_list_insert_head(&MuttrcStack, mutt_str_dup(NONULL(cwd)));
+
+  enum CommandResult ret = mutt_parse_rc_line(line, err);
+
+  struct ListNode *np = STAILQ_FIRST(&MuttrcStack);
+  STAILQ_REMOVE_HEAD(&MuttrcStack, entries);
+  FREE(&np->data);
+  FREE(&np);
+
+  return ret;
+}
+
+/**
+ * mutt_get_sourced_cwd - Get the current file path that is being parsed
+ * @retval ptr File path that is being parsed or cwd at runtime
+ *
+ * @note Caller is responsible for freeing returned string
+ */
+char *mutt_get_sourced_cwd(void)
+{
+  struct ListNode *np = STAILQ_FIRST(&MuttrcStack);
+  if (np && np->data)
+    return mutt_str_dup(np->data);
+
+  // stack is empty, return our own dummy file relative to cwd
+  struct Buffer *cwd = mutt_buffer_pool_get();
+  mutt_path_getcwd(cwd);
+  mutt_buffer_addstr(cwd, "/dummy.rc");
+  char *ret = mutt_buffer_strdup(cwd);
+  mutt_buffer_pool_release(&cwd);
+  return ret;
+}
+
+/**
  * source_rc - Read an initialization file
  * @param rcfile_path Path to initialization file
  * @param err         Buffer for error messages
--- a/command_parse.h
+++ b/command_parse.h
@@ -65,6 +65,9 @@ enum CommandResult parse_unsubscribe     (struct Buffer *buf, struct Buffer *s, 
 enum CommandResult parse_unsubscribe_from(struct Buffer *buf, struct Buffer *s, intptr_t data, struct Buffer *err);
 #endif
 
+enum CommandResult mutt_parse_rc_line_cwd(const char *line, char *cwd, struct Buffer *err);
+char *mutt_get_sourced_cwd(void);
+
 int parse_grouplist(struct GroupList *gl, struct Buffer *buf, struct Buffer *s, struct Buffer *err);
 void clear_source_stack(void);
 int source_rc(const char *rcfile_path, struct Buffer *err);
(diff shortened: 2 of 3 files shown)


command_parse.h
```diff
@@ -65,6 +65,9 @@ enum CommandResult parse_unsubscribe     (struct Buffer *buf, struct Buffer *s,
 enum CommandResult parse_unsubscribe_from(struct Buffer *buf, struct Buffer *s, intptr_t data, struct Buffer *err);
 #endif
 
+enum CommandResult mutt_parse_rc_line_cwd(const char *line, struct Buffer *err, char *cwd);
```
mjsir911, line 68: eh should the arg order be re-ordered to put `err` at the end?

command_parse.h
```diff
@@ -65,6 +65,9 @@ enum CommandResult parse_unsubscribe     (struct Buffer *buf, struct Buffer *s,
...
+char *mutt_get_sourced_cwd(void);
```
mjsir911, line 69: is (void) necessary? I think it flags argument-less function rather than unknown argd function?

hook.c
```diff
@@ -281,6 +283,7 @@ enum CommandResult mutt_parse_hook(struct Buffer *buf, struct Buffer *s,
          * based upon some other information. */
         FREE(&hook->command);
         hook->command = mutt_buffer_strdup(cmd);
+        hook->cwd = mutt_get_sourced_cwd(); // ?
```
mjsir911, line 286: not sure what this code path is doing, what kind of hook is it?

mjsir911:
Oh, related. This invalidates some of what #3470 is saying @whitney-cucumber

mjsir911:
Another weird behavior that might come up is what to do about hooks specified at runtime, outlined in https://github.com/neomutt/neomutt/issues/1557#issuecomment-477991396. 

Current behavior if no relative file is found is fallbacking on `pwd`, which happens at hook-calltime rather than hook-setuptime. I can't really imagine a situation where neomutt's `cwd` changes on-the-go, but the fix would be pretty easy:

```diff
-         hook->cwd = mutt_get_sourced_cwd(); // ?
+         hook->cwd = mutt_get_sourced_cwd() || getcwd(); // ?

**Comment by mjsir911** [comment]
Another weird behavior that might come up is what to do about hooks specified at runtime, outlined in https://github.com/neomutt/neomutt/issues/1557#issuecomment-477991396. 

Current behavior if no relative file is found is fallbacking on `pwd`, which happens at hook-calltime rather than hook-setuptime. I can't really imagine a situation where neomutt's `cwd` changes on-the-go, but the fix would be pretty easy:

```diff
-         hook->cwd = mutt_get_sourced_cwd(); // ?
+         hook->cwd = mutt_get_sourced_cwd() || getcwd(); // ?
```

command_parse.h
flatcap: yes please

command_parse.h
flatcap: Yes.  in C++ `function()` specifically means no parameters.
In C, it means unknown, which means the compiler can't issue warnings.

hook.c
```diff
@@ -64,6 +65,7 @@ struct Hook
   HookFlags type;              ///< Hook type
   struct Regex regex;          ///< Regular expression
   char *command;               ///< Filename, command or pattern to execute
+  char *cwd;                   ///< Used for relative-directory source
```
flatcap, line 68: I think a better name might be something like:
```c
source_file; ///< Config file that defined the hook
```

command_parse.c
```diff
@@ -135,6 +135,24 @@ int parse_grouplist(struct GroupList *gl, struct Buffer *buf, struct Buffer *s,
...
+    STAILQ_REMOVE_HEAD(&MuttrcStack, entries);
+    FREE(&np->data);
+    FREE(&np);
+
+    return ret;
+}
+
+char *mutt_get_sourced_cwd(void) { // caller is responsible for freeing returned string
+    struct ListNode *np = STAILQ_FIRST(&MuttrcStack);
+    return np ? np->data ? mutt_str_dup(np->data) : NULL : NULL;
```
flatcap, line 153: If the `MuttrcStack` is empty, then we should add the current working dir.
My quick hack was:
```c
  struct Buffer *cwd = mutt_buffer_pool_get();
  mutt_path_getcwd(cwd);
  mutt_buffer_addstr(cwd, "/dummy.rc");
  char *dir = mutt_buffer_strdup(cwd);
  mutt_buffer_pool_release(&cwd);
  return dir;
```

hook.c
```diff
@@ -322,6 +325,7 @@ enum CommandResult mutt_parse_hook(struct Buffer *buf, struct Buffer *s,
   hook = mutt_mem_calloc(1, sizeof(struct Hook));
   hook->type = data;
   hook->command = mutt_buffer_strdup(cmd);
+  hook->cwd = mutt_get_sourced_cwd();
```
flatcap, line 328: These need to be freed in `delete_hook()`

flatcap (changes requested):
OK, it looks like it's nearly there.

Once you're done with the changes, please can you:
- Add Doxygen comment blocks for your new functions
  https://neomutt.org/dev/doxygen
- Run `clang-format` on your changes
  https://neomutt.org/dev/analysis/clang-format
  https://neomutt.org/dev/code
- Rebase the code onto `main`

Thanks

flatcap (approved):
Great job!  Thanks

Issue #3442, opened 2022-06-20
Update oauth2 docs


flatcap:
The docs for using oauth2 with GMail and Office365 aren't very good.
They need updating and the links need to point to [**our** script](https://github.com/neomutt/neomutt/blob/main/contrib/oauth2/mutt_oauth2.py).

- [NeoMutt guide](https://github.com/neomutt/neomutt/blob/5190a58ba4b6d9d65aa8dc93bdf83d760913d2f1/docs/manual.xml.head#L12245-L12293)
  **Note**: the website is generated from this file
- [oauth script docs](https://github.com/neomutt/neomutt/blob/main/contrib/oauth2/mutt_oauth2.py.README)

I'm looking for volunteers (because I don't understand oauth2),
and I'm going to shamelessly ping people, @rprimus @gerazov @holmanb

gerazov:
:+1:  I agree - there are some comments in the issues, e.g. #1362 by @[knatsakis](https://github.com/knatsakis) and [googlebleh](https://github.com/googlebleh)

On my side I ended up setting up app passwords because of the way [mutt-wizard](https://github.com/LukeSmithxyz/mutt-wizard/issues/849) is setup, so I can't help too much :sweat_smile:

holmanb:
> because I don't understand oauth2

I make no claims of understanding it myself, but since I'm intending to use it ([and have proposed a minor ergonomic improvement](https://github.com/neomutt/neomutt/pull/3443)) I may propose improvements to the docs as well.

the-x-at:
We are being forced to use OAUTH2 soon, so I have to go through the setup and can update the Office365 part in case I find something unclear.

PR #3478, opened 2022-07-15
Store client_{id,secret} in token file #3477

### LINKED ISSUE #3477: mutt_oauth2.py to store client_id and client secret
## Expected Behaviour

When an access token expired, `mutt_oauth2.py` should use the refresh token to obtain a new access token.

## Actual Behaviour

`mutt_oauth2.py` fails because it does not submit the client ID and secret.

## Steps to Reproduce

1. Do not modify `mutt_oauth2.py`,
2. follow the instructions to register a new client (I used Google, but it should not matter),
3. `mutt_oauth2.py --authorize` to obtain an initial set of tokens, eg:
```
mutt_oauth2.py \
	--verbose \
	--debug \
        --provider google \
	--authorize \
	--client-id "${client_id}" \
	--client-secret "${client_secret}" \
        --authflow authcode \
	--encryption-pipe "gpg --encrypt --recipient frank.gruellich@gmail.com" \
	~/.neomutt/tokens/frank.gruellich@gmail.com.gpg
```
4. wait until the access token expire (or modify the code to always flag them as expired),
5. `mutt_oauth2.py --test` the connection results in:
```
...
400 Bad Request
invalid_request
Could not determine client ID from request.
Perhaps refresh token invalid. Try running once with "--authorize"
```

Closer inspections reveal the it submit an empty client_id and client_secret.

Always, whenever the token expires.

I think when the idea surfaced that `mutt_oauth2.py` should be usable without any modifications, eg. by coding client ID and secrets into it.

## NeoMutt Version
Should matter, NeoMutt was installed from homebrew, `mutt_oauth2.py` was taken from current `main` branch.
```
NeoMutt 20220429
Copyright (C) 1996-2022 Michael R. Elkins and others.

System: Darwin 21.5.0 (arm64)
libiconv: 1.11
libidn2: 2.3.2 (compiled with 2.3.2)
GPGME: 1.17.1
OpenSSL: OpenSSL 1.1.1n  15 Mar 2022
libnotmuch: 5.6.0
storage: tokyocabinet, lmdb

Configure options: --prefix=/opt/homebrew/Cellar/neomutt/20220429 --gss --disable-idn --idn2 --lmdb --notmuch --sasl --tokyocabinet --with-gpgme=/opt/homebrew/opt/gpgme --with-lua=/opt/homebrew/opt/lua --with-ssl=/opt/homebrew/opt/openssl@1.1 --with-ui=ncurses

Compilation CFLAGS: -g -O2 -std=c99 -D_ALL_SOURCE=1 -D_GNU_SOURCE=1 -D__EXTENSIONS__ -I/opt/homebrew/opt/gpgme/include -I/opt/homebrew/Cellar/libassuan/2.5.5/include -I/opt/homebrew/opt/libgpg-error/include -I/opt/homebrew/Cellar/neomutt/20220429/include -I/opt/homebrew/opt/lua/include/lua5.4 -DNCURSES_WIDECHAR -I/opt/homebrew/opt/openssl@1.1/include

Default options:

Compile options:
  -autocrypt +fcntl -flock -fmemopen +futimens +getaddrinfo -gnutls +gpgme +gss 
  +hcache -homespool +idn -inotify -locales_hack +lua -mixmaster +nls +notmuch 

MAILPATH="/var/mail"
PKGDATADIR="/opt/homebrew/Cellar/neomutt/20220429/share/neomutt"
SENDMAIL="/usr/sbin/sendmail"
SYSCONFDIR="/opt/homebrew/Cellar/neomutt/20220429/etc"

```

## Extra Info

No relevant information...

FrankGruellich:
* **What does this PR do?**
Makes `mutt_oauth2.py` store / fetch client ID and secret into / from token file.

     for that)

This tries to fix #3477

### DIFF
--- a/contrib/oauth2/mutt_oauth2.py
+++ b/contrib/oauth2/mutt_oauth2.py
@@ -59,8 +59,6 @@ registrations = {
         'smtp_endpoint': 'smtp.gmail.com',
         'sasl_method': 'OAUTHBEARER',
         'scope': 'https://mail.google.com/',
-        'client_id': '',
-        'client_secret': '',
     },
     'microsoft': {
         'authorize_endpoint': 'https://login.microsoftonline.com/common/oauth2/v2.0/authorize',
@@ -75,8 +73,6 @@ registrations = {
         'scope': ('offline_access https://outlook.office.com/IMAP.AccessAsUser.All '
                   'https://outlook.office.com/POP.AccessAsUser.All '
                   'https://outlook.office.com/SMTP.Send'),
-        'client_id': '',
-        'client_secret': '',
     },
 }
 
@@ -107,15 +103,11 @@ ap.add_argument('--client-secret', type=str, default='',
                 help='(optional) Provider secret from registration')
 ap.add_argument('--provider', type=str, choices=registrations.keys(),
                 help='Specify provider to use.')
+ap.add_argument('--email', type=str, help='Your email address.')
 args = ap.parse_args()
 
 ENCRYPTION_PIPE = args.encryption_pipe
 DECRYPTION_PIPE = args.decryption_pipe
-
-provider = registrations.get(args.provider)
-if provider:
-    provider['client_id'] = args.client_id
-    provider['client_secret'] = args.client_secret
 
 token = {}
 path = Path(args.tokenfile)
@@ -148,14 +140,19 @@ if args.debug:
 if not token:
     if not args.authorize:
         sys.exit('You must run script with "--authorize" at least once.')
-    print('Available app and endpoint registrations:', *registrations)
-    token['registration'] = input('OAuth2 registration: ')
-    token['authflow'] = input('Preferred OAuth2 flow ("authcode" or "localhostauthcode" '
-                              'or "devicecode"): ')
-    token['email'] = input('Account e-mail address: ')
+    print('', )
+    token['registration'] = args.provider or input(
+        'Available app and endpoint registrations: {regs}\nOAuth2 registration: '.format(
+            regs=', '.join(registrations.keys())))
+    token['authflow'] = args.authflow or input(
+        'Preferred OAuth2 flow ("authcode" or "localhostauthcode" or "devicecode"): '
+    )
+    token['email'] = args.email or input('Account e-mail address: ')
     token['access_token'] = ''
     token['access_token_expiration'] = ''
     token['refresh_token'] = ''
+    token['client_id'] = args.client_id or input('Client ID: ')
+    token['client_secret'] = args.client_secret or input('Client secret: ')
     writetokenfile()
 
 if token['registration'] not in registrations:
@@ -167,7 +164,7 @@ authflow = token['authflow']
 if args.authflow:
     authflow = args.authflow
 
-baseparams = {'client_id': registration['client_id']}
+baseparams = {'client_id': token['client_id']}
 # Microsoft uses 'tenant' but Google does not
 if 'tenant' in registration:
     baseparams['tenant'] = registration['tenant']
@@ -260,7 +257,7 @@ if args.authorize:
             del p[k]
         p.update({'grant_type': 'authorization_code',
                   'code': authcode,
-                  'client_secret': registration['client_secret'],
+                  'client_secret': token['client_secret'],
                   'code_verifier': verifier})
         try:
             response = urllib.request.urlopen(registration['token_endpoint'],
@@ -343,7 +340,8 @@ if not access_token_valid():
     if not token['refresh_token']:
         sys.exit('ERROR: No refresh token. Run script with "--authorize".')
     p = baseparams.copy()
-    p.update({'client_secret': registration['client_secret'],
+    p.update({'client_id': token['client_id'],
+              'client_secret': token['client_secret'],
               'refresh_token': token['refresh_token'],
               'grant_type': 'refresh_token'})
     try:


flatcap:
Thanks for the improvements @FrankGruellich :heart:

@holmanb @brunnre8 @ryanakca 
As people who understand this better than I...
any thoughts or suggestions?

Thanks

ryanakca:
There's something slightly buggy in the resulting script. The `--provider X` option is not required, and the script interactively prompts users for a provider during authorization if they did not pass a `--provider` option. In this case, however, authorization fails because the script does not send a `client_id`. For example, attempting to get a token from Microsoft using
```
echo -e 'microsoft\nlocalhostauthcode\nneomutt@mcgill.ca\n' | \
   ~/bin/mutt_oauth2.py --authorize \
                        --client-id='08162f7c-0fd2-4200-a84a-f25a4db0b584' \
                        --client-secret='TxRBilcHdC6WGBee]fs?QR:SJ8nI[g82' \
                        --encryption-pipe "cat" \
                        --decryption-pipe "cat" ~/.mutt/xoauth2-mcgill.ca-token
```
results in `AADSTS900144: The request body must contain the following parameter: 'client_id'.` (This does not happen if one passes `--provider microsoft` along with `--authorize`.) 

I suspect the culprit involves lines 115-118 and lines 160-161 (but I haven't carefully traced this through). Lines 115-118 update `registrations[P]['client_id']` (and the corresponding client_secret entry) if users pass `--provider P`, but lines 160-161 do not similarly update `registrations[P]` if they interactively get the provider. Perhaps these lines should be refactored so that we're not storing and fetching client_ids and client_secrets to/from multiple places?

the-x-at:
As far as I understand the script, the original author intended `mutt_oauth2.py` to be taken and modified, as `client_secret` needs to be inserted into the source code. The same for the key store pipeline, where `YOUR_GPG_IDENTITY` needs to be modified in the source. This connects to this issue, and I would suggest to update the command line interface of `mutt_oauth2.py` such that no script modification is required anymore and rethink what should go into the token store, see also issue #3442 where we would like to enhance the documentation for this authentication method.

holmanb:
> As far as I understand the script, the original author intended mutt_oauth2.py to be taken and modified, as client_secret needs to be inserted into the source code. The same for the key store pipeline, where YOUR_GPG_IDENTITY needs to be modified in the source. This connects to this issue, and I would suggest to update the command line interface of mutt_oauth2.py such that no script modification is required anymore and 

@the-x-at for context on recent changes in script behavior: https://github.com/neomutt/neomutt/commit/7decc96c7ccabae4cb940927665783fe6ba81761

the-x-at:
Hi, @holmanb, thanks for pointing me to this. I need to update the repository then. That way I can provide more useful comments in the future. Sorry for that.

FrankGruellich:
> There's something slightly buggy in the resulting script. The `--provider X` option is not required, and the script interactively prompts users for a provider during authorization if they did not pass a `--provider` option. [great description]

Indeed, in that situation the script still tried to fetch information from the `registrations`, sorry for missing that.  I've fixed that, your test case does not throw an error right away and should succeed now.  Please try again and let me know.

contrib/oauth2/mutt_oauth2.py
```diff
@@ -167,7 +170,7 @@ def writetokenfile():
 if args.authflow:
     authflow = args.authflow
 
-baseparams = {'client_id': registration['client_id']}
+baseparams = {'client_id': args.client_id or registration['client_id']}
```
brunnre8, line 173: Getting it from registration is pointless.
Unless the user modified the script by hardcoding those values, it doesn't do anything and if they did they wouldn't pull in your changes.

If we want to swtich to a version where the users doesn't monkeypatch the script file prior to running it we can simply remove that case.

contrib/oauth2/mutt_oauth2.py
```diff
@@ -260,7 +263,7 @@ def do_GET(self):
             del p[k]
         p.update({'grant_type': 'authorization_code',
                   'code': authcode,
-                  'client_secret': registration['client_secret'],
+                  'client_secret': args.client_secret or registration['client_secret'],
```
brunnre8, line 266: same

contrib/oauth2/mutt_oauth2.py
FrankGruellich: > Getting it from registration is pointless. Unless the user modified the script by hardcoding those values, it doesn't do anything and if they did they wouldn't pull in your changes.

Understood.  But I tried to only add functionality and not remove any, so I left this there for compatibility reasons, if users prefer to edit their secrets into the script (for whatever reasons).  What do you think?  I'm not going to argue much, though.

> If we want to swtich to a version where the users doesn't monkeypatch the script file prior to running it we can simply remove that case.

Honestly, I don't think it makes much sense to modify the current script into that direction.  I don't want to offend anyone, but script is quite messy and IMHO difficult to modify.  I don't know if I could do any better, but I could give it a try.

holmanb: > > Getting it from registration is pointless. Unless the user modified the script by hardcoding those values, it doesn't do anything and if they did they wouldn't pull in your changes.
> 
> Understood.  But I tried to only add functionality and not remove any, so I left this there for compatibility reasons, if users prefer to edit their secrets into the script (for whatever reasons).  What do you think?  I'm not going to argue much, though.

I believe backwards compatibility is not required here. Backwards compatibility for editing your config into the script was broken recently, in favor of cli args that can go directly into your invokation in your .muttrc

FrankGruellich: > I believe keeping backwards compatibility is not required.

Okay, fine with me.  Fixed!

contrib/oauth2/mutt_oauth2.py
FrankGruellich: > same

Fixed!

contrib/oauth2/mutt_oauth2.py
```diff
@@ -149,13 +149,16 @@ def writetokenfile():
     if not args.authorize:
         sys.exit('You must run script with "--authorize" at least once.')
     print('Available app and endpoint registrations:', *registrations)
```
holmanb, line 151: If the provider is passed as an argument, this print is not useful. Can we integrate this into the input() call for registration?

contrib/oauth2/mutt_oauth2.py
```diff
@@ -149,13 +149,16 @@ def writetokenfile():
...
-                              'or "devicecode"): ')
+    token['registration'] = args.provider or input('OAuth2 registration: ')
+    token['authflow'] = args.authflow or input(
+        'Preferred OAuth2 flow ("authcode" or "localhostauthcode" or "devicecode"): '
+    )
     token['email'] = input('Account e-mail address: ')
     token['access_token'] = ''
     token['access_token_expiration'] = ''
     token['refresh_token'] = ''
+    token['client_id'] = args.client_id
```
holmanb, line 160: Could we make client_id and client_secret also optionally interactive? If not they are required args but the script will not warn that there is a problem if they aren't included but will create a bogus url - not a great user experience.

contrib/oauth2/mutt_oauth2.py
```diff
@@ -149,13 +149,16 @@ def writetokenfile():
...
         sys.exit('You must run script with "--authorize" at least once.')
     print('Available app and endpoint registrations:', *registrations)
-    token['registration'] = input('OAuth2 registration: ')
-    token['authflow'] = input('Preferred OAuth2 flow ("authcode" or "localhostauthcode" '
-                              'or "devicecode"): ')
+    token['registration'] = args.provider or input('OAuth2 registration: ')
+    token['authflow'] = args.authflow or input(
+        'Preferred OAuth2 flow ("authcode" or "localhostauthcode" or "devicecode"): '
+    )
     token['email'] = input('Account e-mail address: ')
```
holmanb, line 156: If we add an email cli arg in addition to the client_id and client_secret input() below, then auth can be done completely non-interactively, which would be nice.

holmanb (changes requested):
Hi @FrankGruellich,

I added a couple small change requests, but based on some testing and modifications I did in my own branch tonight I think this is looking pretty good. Let me know your thoughts.

My own changes are in a different PR, but I plan on dropping the functional changes from my PR once this lands (while keeping the docs changes).

Thanks for this contribution!

FrankGruellich:
Hi @holmanb,

thanks a lot for the suggestions.  I've added them as good as possible.  Let me know what you think.

Kind regards,
 Frank.

holmanb (approved):
I tested this locally with a gmail account and the `authcode` option.

Looks good to me, thanks!

holmanb:
@brunnre8 @ryanakca Any other concerns? I'm in favor of merging.

brunnre8:
I do not have any objections, but I also don't use the script.

The script itself is a horrifying mess in my view, but that's no fault of the PR author.
It crams to many different things into one interface... Subcommands would help rather than a single thing as that would make the validation of the input args much easier.

Also I'm not exactly fond of interactive scripts, they should just error out when they don't receive what they need. Especially if it's something that's usually run in the background such as this.

But as said, no fault of the PR author and the changes look reasonable when just reading the code, I didn't runtime test it.

ryanakca:
Tested token creation locally with a Microsoft account and it looks good to me. Perhaps mention in the release notes that users will need to delete their old token file after upgrading. Otherwise, the script aborts with:

```
Traceback (most recent call last):
  File "/tmp/neomutt/contrib/oauth2/./mutt_oauth2.py", line 167, in <module>
    baseparams = {'client_id': token['client_id']}
KeyError: 'client_id'
```

FrankGruellich:
> The script itself is a horrifying mess in my view, [...]

I don't want to offend the original contributors, but I admit that the code is indeed quite messy.  For this simple change I had to touch too many areas scattered all other the place of it.  I am thinking about rewriting the whole thing.

> It crams to many different things into one interface... Subcommands would help rather than a single thing as that would make the validation of the input args much easier.
> 
> Also I'm not exactly fond of interactive scripts, they should just error out when they don't receive what they need. Especially if it's something that's usually run in the background such as this.

I generally agree to that and will try to keep these suggestions in mind.

> ```
> Traceback (most recent call last):
>   File "/tmp/neomutt/contrib/oauth2/./mutt_oauth2.py", line 167, in <module>
>     baseparams = {'client_id': token['client_id']}
> KeyError: 'client_id'
> ```

I could try to work around that error.  OTOH, I don't think it's worth the effort.

brunnre8:
nothing specific. Just noticed that a lot of the complexity stems from the overload

holmanb:
Docs coverage for changes here is reflected in #3483, which adds other docs improvements as well

flatcap (approved):

flatcap:
Thanks everyone!  :heart:

> I don't want to offend the original contributors

It's fine, they're Mutt people, not NeoMutt people :-)

> I am thinking about rewriting the whole thing

That sounds great.
I've just started a Discussion #3505 -- all ideas welcome.

PR #3483, opened 2022-07-23
mutt_oauth2.py: Update usage docs

holmanb:
* **What does this PR do?**

Updates docs for changes made in 7decc96c7ccabae4cb940927665783fe6ba81761

This should have been added then, I just noticed they were missed.

### DIFF
--- a/contrib/oauth2/mutt_oauth2.py.README
+++ b/contrib/oauth2/mutt_oauth2.py.README
@@ -174,11 +174,12 @@ matter and could be "mutt registration project".
    - Under scopes, choose Add scope, scroll all the way down, enable the "https://mail.google.com/" scope
    - Fill out additional fields (application logo, etc) if you feel like it (will make the consent screen look nicer)
  - Back at console.developers.google.com, choose Credentials
- - At top, choose Create Credentials / OAuth2 client iD
+ - At top, choose Create Credentials / OAuth client iD
    - Application type is "Desktop app"
+ - On the OAuth consent screen add a test user, (email address)
 
-Edit the client_id (and client_secret if there is one) into the
-mutt_oauth2.py script.
+The client_id (and client_secret if there is one) will need to be passed as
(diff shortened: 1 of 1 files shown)


holmanb:
I have more to add before this is ready for review, converting to a draft for now.

holmanb:
I plan on dropping functional changes here once the other PR lands, but keeping the docs changes.

the-x-at:
I have used the script today (`main` repository at commit 9639bfaf3dfbbe71130b86e4076e2c6ae31ad85b) on a Microsoft Office365 account. I have also dealt with the original script a while ago and know the changes pretty well. I furthermore discovered a couple of pitfalls that are not real bugs but user experience could be improved by some more feedback from the script. I think I understand the (Microsoft Azure Active Directory) OAUTH2 workflow enough to enhance both the script and the documentation. Let me know once the script is finished/if I can contribute already now. (Who is working on which fork on it?)
My setup is even a bit more complicated, as it involves Neomutt for sending email but `offlineimap` for reading email. I would like to share some thoughts how the token can be retrieved there without copy and pasting it.

holmanb:
> I think I understand the (Microsoft Azure Active Directory) OAUTH2 workflow enough to enhance both the script and the documentation.

Great, glad to hear it.

> Let me know once the script is finished/if I can contribute already now.

I don't want to block anyone else, but I'll drop the script changes from this PR now. This PR is "blocked by" 3478, since the docs here reflect the new changes there. Hopefully #3478 lands soon, but since that has no outstanding issues you may want to base your docs changes on what exists there and work in parallel. Its up to you how to do it, of course.

> My setup is even a bit more complicated, as it involves Neomutt for sending email but offlineimap for reading email. I would like to share some thoughts how the token can be retrieved there without copy and pasting it.

Nice :)

holmanb:
Related to https://github.com/neomutt/neomutt/issues/3442

gahr (approved):
LGTM

PR #3490, opened 2022-07-28
nntp: change nntp_context type to DT_LONG

MoSal:
Short values for this option are too limiting. And I came across
 servers/implementations where not only article numbers are not
 serial, but long gaps can exist between successive ones, so you
 would have to set nntp_context to a high enough value (> SHRT_MAX)
 to get enough articles back.

### DIFF
--- a/docs/config.c
+++ b/docs/config.c
@@ -2771,7 +2771,7 @@
 ** authentication fails, NeoMutt will not connect to the IMAP server.
 */
 
-{ "nntp_context", DT_NUMBER, 1000 },
+{ "nntp_context", DT_LONG, 1000 },
 /*
 ** .pp
 ** This variable defines number of articles which will be in index when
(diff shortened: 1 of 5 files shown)


gahr:
Wouldn't this require a change to [anum_t](https://github.com/neomutt/neomutt/blob/8280a6351324722ca327e1edc7006ff75213d49b/nntp/lib.h#L60) too?

MoSal:
> Wouldn't this require a change to [anum_t](https://github.com/neomutt/neomutt/blob/8280a6351324722ca327e1edc7006ff75213d49b/nntp/lib.h#L60) too?

Do you mean in case the article number is larger than `UINT32_MAX`?

gahr:
Mostly to avoid mixing longs and ints

MoSal:
Pushed commits changing `anum_t` define to `long`, and actually using `ANUM` wherever applicable.

gahr (approved):
Thanks, LGTM - I'll let @flatcap have a look when he's back.

flatcap (approved):
Lovely.  Thanks for taking the time to create a fix.

PR #3500, opened 2022-08-08
enter: fix default text

flatcap:
If the entry field has some default text, e.g. when changing folder, fix `<enter>` to select that text.

This was broken by some recent refactoring of the enter code.

### DIFF
--- a/enter/window.c
+++ b/enter/window.c
@@ -232,19 +232,11 @@ int mutt_buffer_get_field(const char *field, struct Buffer *buf, CompletionFlags
     // clang-format on
     win->wdata = &wdata;
 
-    if (wdata.state->wbuf)
-    {
-      wdata.redraw = ENTER_REDRAW_LINE;
-      wdata.first = false;
-    }
-    else
-    {
-      /* Initialise wbuf from buf */
-      wdata.state->wbuflen = 0;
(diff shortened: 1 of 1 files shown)


gahr (approved):
Works, thanks!

the-x-at:
Not sure if we are talking about the issue I commented in commit 8280a63. Compiled the last version and the problem remains. I am using `abook`. When starting to compose an email, entering one email address in e.g. `To:` autocompletes with `tab`, but then adding a comma and a second email address will replace the first email address upon autocompletion instead of adding to it. The same with a tagged lists suggested from an `abook` call. Only one recipient is chosen, not all tagged ones.
If this explanation is not sufficient, I can file a bug report. I am not sure how to write bug reports for daily versions, though. That's why I have chosen to comment here for now.

gahr:
Yep, that looks like a different issue (which I can reproduce).

the-x-at:
I better file a bug report then, no? I there then reference to the last commit to make things clear.

flatcap:
@the-x-at #3503 
Sorry.  I'll stop breaking things.

the-x-at:
C'mon, @flatcap. :-D You guys are doing an awesome job.

the-x-at:
I am waiting till this gets into `main` and then test it all again.

PR #3503, opened 2022-08-09
enter: re-fix lost data

flatcap:
Fix a regression that lost data, e.g. when entering "To:", after using auto-completion on a second address.

---

Working:
- Selecting a default, e.g. a mailbox with new mail
- Adding to existing text, e.g. "To:" prompt + address auto-completion

### DIFF
--- a/enter/state.c
+++ b/enter/state.c
@@ -59,8 +59,11 @@ void enter_state_resize(struct EnterState *es, size_t num)
     return;
 
   num = ROUND_UP(num + 4, 128);
-  es->wbuflen = num;
   mutt_mem_realloc(&es->wbuf, num * sizeof(wchar_t));
+
+  memset(es->wbuf + es->wbuflen, 0, (num - es->wbuflen) * sizeof(wchar_t));
+
+  es->wbuflen = num;
 }
 
 /**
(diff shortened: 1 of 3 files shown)


enter/state.c
```diff
@@ -72,6 +72,7 @@ struct EnterState *enter_state_new(void)
   struct EnterState *es = mutt_mem_calloc(1, sizeof(struct EnterState));
 
   enter_state_resize(es, 1);
+  memset(es->wbuf, 0, es->wbuflen);
```
gahr, line 75: Wouldn't it be best if `enter_state_resize` filled the range `[wbuflen,  num)` with zeroes?

flatcap: Yes, but that would require some careful thought.
Better still would be `mutt_realloc_zero()`; I just need some newbie to volunteer.

flatcap:
`memset()` moved; tests increased; asan happy :-)

the-x-at:
Tested it in `main` after this merge. Works for all kinds of email recipient headers, `To:`, `Cc:`, and '`Bcc:`. The other issue with `abook` multiple recipients selection is still unresolved and is likely unrelated. For this I am going to file a bug report as suggested by @gahr.

PR #3506, opened 2022-08-11
Resolve alternates when subscribing/unsubscribing

### LINKED ISSUE #3461: list-unsubscribe ignores reverse_name
## Expected Behaviour

I am attempting to use the `list-unsubscribe` function to unsubscribe from company spam emails. These emails do have a proper 'To:' header. If I use 'reply' the `From:` header of the reply is set correctly. I would expect that same for `list-unsubscribe`.

## Actual Behaviour

Instead the default `from` set in my configuration is used.

## Steps to Reproduce

* Open an email with a List-Unsubscribe header
* `:exec list-unsubscribe`
* look at resulting email

## NeoMutt Version

20220429

## Extra Info

* Is your email local (maildir) or remote (IMAP)? maildir

gahr:
Fixes #3461

[`mutt_send_message`](https://github.com/neomutt/neomutt/blob/d8fd4643b1a4221163a96974543fcbd4c4a78892/send/send.c#L2303-L2320) needs the current email (first entry in EmailList) to resolve the reverse name.

### DIFF
--- a/send/send.c
+++ b/send/send.c
@@ -2976,15 +2976,15 @@ cleanup:
 /**
  * send_simple_email - Compose an email given a few basic ingredients
  * @param m Mailbox
+ * @param el List of source Emails
  * @param mailto mailto address to parse (can include fields such as subject)
- * @param to Recipient address, if not overridden by mailto
  * @param subj Subject, if not overridden by mailto
  * @param body text/plain body
  * @retval true Success
  * @retval false Failure
  */
-static bool send_simple_email(struct Mailbox *m, const char *mailto,
-                              const char *to, const char *subj, const char *body)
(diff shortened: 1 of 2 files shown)


KenMacD:
This worked perfectly when I tested it using the following Nix overlay. Thank you @gahr!

```nix
self: super:
{
  neomutt = super.neomutt.overrideAttrs (old: {
    patches = (old.patches or []) ++ [
      (super.fetchpatch {
        name = "list-unsubscribe";
        url = "https://patch-diff.githubusercontent.com/raw/neomutt/neomutt/pull/3506.patch";
        sha256 = "sha256-kNzcS+sXEE7+NUrXRi9+os+GClg7/2/+ouBrn9ebrSo=";
      })
    ];
  });
}
```

flatcap (approved):

PR #3513, opened 2022-09-01
imap: fix oob write during debug on 32 bit systems

stoeckmann:
* **What does this PR do?**

This PR fixes a very obscure bug which can be only triggered on a very limited amount of systems.
A malicious IMAP server could crash neomutt with an out of boundary write.

For this, the following criteria must be met:
- 32 bit system
- realloc(NULL, 0) returning NULL or a pointer which may not be written to, e.g. OpenBSD
- neomutt running with debug, e.g. `neomutt -d 5`
- IMAP server replying messages header with {4294967285} as byte amount (no actual message required)

In such a condition, an integer overflow leads to a 0 byte allocation in mutt/buffer.c with a '\0' write into
this 0 byte allocated memory. Although this is not allowed, it still takes systems like OpenBSD which
actually crash a violating program.

An offending IMAP line would be this (important is the last part of the line):
`* 1 FETCH (UID 220705 RFC822.SIZE 6722 INTERNALDATE "01-Sep-2022 00:00:00 +0000      " FLAGS (\Seen NonJunk) BODY[HEADER.FIELDS (DAT    E FROM SENDER SUBJECT TO CC MESSAGE-ID REFERENCES CONTENT-TYPE CONTENT-DESCRIPTION       IN-REPLY-TO REPLY-TO LINES LIST-POST LIST-S    UBSCRIBE LIST-UNSUBSCRIBE X-LABEL X-ORIGINAL-TO)] {4294967285}`


imap/imap.c
```diff
@@ -604,15 +604,15 @@ int imap_read_literal(FILE *fp, struct ImapAccountData *adata,
 
   const short c_debug_level = cs_subset_number(NeoMutt->sub, "debug_level");
   if (c_debug_level >= IMAP_LOG_LTRL)
-    mutt_buffer_alloc(&buf, bytes + 10);
+    mutt_buffer_alloc(&buf, bytes);
```
gahr, line 607: We `'\0'`-terminate the buffer, so we'll need `bytes + 1` at least. I'm not sure about `+ 10`. Perhaps it's a typo in 81e7278ebe3f2c8e5e4c318de5aed090883ee7bf and was supposed to be `+ 1` to begin with?

Anyway, even if we don't overflow here, we'll still have a buffer `SIZE_MAX` long. When we try to add a char to it, we call `mutt_buffer_addch` which calls `mutt_buffer_addstr_n` which checks `((buf->dptr + len + 1) > (buf->data + buf->dsize))`. If everything is 32 bit (size_t and pointers) then probably `buf->data + buf->dsize` will overflow long before `buf->dsize` is `SIZE_MAX` and we'll end up writing out of bounds anyway.

Do you think this change is enough to avoid us writing out of bounds in the `buffer` API?

stoeckmann: Extra space for '\0' is already taken care of in buffer_alloc (https://github.com/neomutt/neomutt/blob/main/mutt/buffer.c#L276). I think the `+ 10` might be a leftover from the past in which a prefix has been added. I didn't find a reason in the commit though. In either way, the buffer logic is supposed to dynamically grow so even if 10 bytes would be needed, eventually it would be allocated.

To overflow these calculations you would need another bug. It is already difficult to get a buffer allocation at the size of around SIZE_MAX/2 on i686 for a real world application. Now you also need a string of at least SIZE_MAX/2-1 in your virtual address space to overflow the calculation. If there's a bug which allows to access the buffer content and concatenate it with itself through mutt_buffer_addstr_n call, then yes. Otherwise it won't be possible since other allocations (and stack space) reside in that 32 bit virtual address space.

I am not arguing that the buffer code is bullet-proof, but the only "interesting" parts from my point of view are places where size_t gets casted to int, e.g. https://github.com/neomutt/neomutt/blob/main/mutt/buffer.c#L387

But since paths are limited within neomutt, I have not found a proof of concept, so no report about it. It was already tricky (and entertaining) to find a system which triggers an issue with this one. :)

gahr: > Extra space for '\0' is already taken care of in buffer_alloc (https://github.com/neomutt/neomutt/blob/main/mutt/buffer.c#L276)

True, but that'd be wasteful: we know we need x bytes + '\0', if we pre-allocate only x, then we'll get x + whatever the buffer code thinks is the best increment step. Do you think it's too bad to keep the +1 there?

> It is already difficult to get a buffer allocation at the size of around SIZE_MAX/2 on i686 for a real world application.

I don't know enough about how operating systems organize their heap addresses to comment on this, but yeah, I get your point.

I think I like the PR but I'd like to keep the +1 there. What do you think?

stoeckmann: I have added `+ 1` again. In this situation it would not be needed and in fact could overflow due to SIZE_MAX as server reply. But yet again, the buffer logic would handle this situation by reallocating until no more memory is left.

stoeckmann: Regarding your pointer arithmetic example: It could theoretically overflow, depending on location of buf->data in virtual memory. I cannot find a system which allows me to create a proof of concept, but I'll create another pull request for it if you like.

gahr: Yeah, the first `mutt_buffer_alloc` is an optimization only: if the imap reply is size_max, it'll try to allocate 0 which will result in the buffer allocating `BufferStepSize`.

gahr: At this point I'm not sure about your change to buffer.c: if it is not true that after the call the buffer can hold at least new_size bytes, then we should turn it into a boolean function and check at the callers.

stoeckmann: In which situation should dsize be smaller than new_size?

gahr: on overflow

gahr: ah nevermind, I mis read `buf->dsize = SIZE_MAX` as `buf->dsize = new_size` here:

```c
  if (new_size > SIZE_MAX - BufferStepSize)
    buf->dsize = SIZE_MAX;
  else
    buf->dsize = ROUND_UP(new_size + 1, BufferStepSize);
```

not sure it really makes sense either way.. you're then going to mutt_mem_realloc with SIZE_MAX, and that'll probably kill it anyway.

stoeckmann: Yeah, instead of trying a SIZE_MAX allocation I could also set errno manually and stop execution. But it is so much easier and consistent to let the system libraries figure this out. :)

gahr: @flatcap thoughts?

flatcap: (after a bit of discussion with @gahr on IRC)
I've made one tweak to `mutt_buffer_alloc()` so that the `Buffer` is never shrunk.
This means we'll never have the `realloc(ptr, 0)` situation.

flatcap (approved):
Thanks, @stoeckmann
It's a small improvement, but welcome.

PR #3524, opened 2022-09-17
Add GNU SASL support for authentication

flatcap:
Add support for the GNU SASL library, using **configure** option **`--gsasl`**

Add multiline response support for SMTP authentication (which is probably not actually needed).  Also add arbitrary line length for the SASL server responses (the RFCs note that for SASL, the protocol line lengths don't apply).

Upstream-commit: https://gitlab.com/muttmua/mutt/commit/68caf9140c8217ecf6c848460c4b4d27996b2922

### DIFF
--- a/.github/workflows/build-and-test.yml
+++ b/.github/workflows/build-and-test.yml
@@ -22,7 +22,7 @@ jobs:
           - name: default
             options:
           - name: disabled
-            options: --disable-idn --disable-nls --disable-pgp --disable-smime --ssl
+            options: --disable-idn --disable-nls --disable-pgp --disable-smime --ssl --gsasl
           - name: everything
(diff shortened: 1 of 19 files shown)


flatcap:
**Note**: some of the builds are failing because the Docker image doesn't have gsasl installed

4nd3r:
First try to send mail failed:
```
[2022-09-18 10:48:11]<2> smtp_auth_gsasl() returned no usable mech
[2022-09-18 10:48:11]<E> smtp_authenticate() No authenticators available
```

But `mutt` in Debian Sid (with gsasl) works. I'll dig into it.

flatcap:
> I'll dig into it.

Thanks!

I'll have another look over the code too
(and compare it to upstream)

4nd3r:
If I add `set smtp_authenticators = PLAIN` to config, then it works.

From manual:
> If it is “unset” (the default) mutt will try all available methods, in order from most-secure to least-secure.

flatcap:
> it works

That's all I saw :-)  Woo!

> "try all available methods"

Yeah, that sounds familiar, but I'm sure we've commonly told people to `set smtp_authenticators = "login:plain"`
Nobody thought to find out why :-)

> If that helps, I can provide you an account for testing.

Yes, please.  That'd be brilliant.

4nd3r:
> Yes, please. That'd be brilliant.

Check your mail.

Issue #2659, opened 2020-09-14
[RFC] Prompt-User command


flatcap:
NeoMutt users are an inventive bunch, coming up with complex macros to solve their daily problems.
One feature they often ask for is a user prompt to set a variable.

I suggest a new command:

- `prompt_user { config_var | my_var } [ "message" ]`

e.g.

```
# Ask the user for a $from value
prompt_user from "Enter from address"
```

The entered value would be put into the variable directly (no escaping or quoting necessary).
Hitting <kbd>Ctrl-G</kbd> (`$abort_key`) would leave the variable unchanged.

We might also want some options to fine-tune the behaviour.
e.g.

- `-hidden` display `*`s to conceal the input
- `-onechar` only read one char
- `-anykey` wait for any key
- `-menu` interpret (k)ey letters

---

If the idea is liked, I'll turned it into a 'good first issue'.

What do you think of the idea?

**Vote now**: :+1: :-1: (and add comments / suggestions)

motorto:
Hey @flatcap  I am interested in taking this "problem" .

flatcap:
The challenge can be broken down into:

- Create a Command
- Parse function
- Prompt the user
- Set the variable
- Return Errors

## Create a Command

There are two parts to this.
In [`mutt_commands.c`](https://github.com/neomutt/neomutt/blob/master/mutt_commands.c) add:

- an entry to the [`mutt_commands[]` array](https://github.com/neomutt/neomutt/blob/master/mutt_commands.c#L49),
- a parsing function, `parse_prompt()`

[`struct Command`](https://github.com/neomutt/neomutt/blob/master/mutt_commands.h#L45-L60) defines both the format of the array and the prototype of the parsing function.

Put the function in [`command_parse.c`](https://github.com/neomutt/neomutt/blob/master/command_parse.c) and a prototype in [`command_parse.h`](https://github.com/neomutt/neomutt/blob/master/command_parse.h)

## Parse function

When the user types in the command, our parse function will be called.
It's the function's job to parse the string and peform any actions required.

We'll start simply, with no options.  The string will have the format:

```
prompt_user variable "message"
```

Use [`mutt_extract_token()`](https://github.com/neomutt/neomutt/blob/master/init.c#L386-L394) to read one "token" at a time, e.g. the command, a prompt string.
The token will be read into a [`struct Buffer`](https://github.com/neomutt/neomutt/blob/master/mutt/buffer.h#L30-L38) (an expanding string type).
[...]

PR #3530, opened 2022-09-22
Implement 'prompt_user' command

danielg0:
* **What does this PR do?**

`prompt_user` is a new command that takes a variable and, optionally, a message, implemented by `parse_prompt`. It prompts the user if a message was given, and stores their response in the variable. The user can abort giving an input, which leaves the variable unchanged.

This command takes a number of flags:

* `-hidden`: hides the user input whilst it is being typed
* `-file`: enables file/directory name completion on the user input
* `-onechar`: takes user input as normal, but then discards all of it except for the first character
* `-anykey`: stop taking user input after the first keypress

Flags can be anywhere in the command, and the target variable is the the first non-flag token. If a second non-flag token is given, it is used as the user prompt. The flags `-hidden` and `-file` conflict, and the one given last will be used instead of the other.

Not yet - documentation still needs to be written for `prompt_user`. There was also discussion in the issue about adding a menus flag that restricts the response to a set of letters, and abort. I have not written any tests for `parse_prompt` yet (not sure how I would given the command requires user input - any pointers would be great)

#2659 - Prompt-User command

### DIFF
--- a/commands.c
+++ b/commands.c
@@ -47,6 +47,7 @@
 #include "commands.h"
 #include "attach/lib.h"
 #include "color/lib.h"
+#include "enter/lib.h"
 #include "imap/lib.h"
 #include "menu/lib.h"
 #include "pager/lib.h"
@@ -57,6 +58,7 @@
 #include "keymap.h"
 #include "muttlib.h"
 #include "mx.h"
+#include "opcodes.h"
 #include "score.h"
 #include "version.h"
 #ifdef USE_INOTIFY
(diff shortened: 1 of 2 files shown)


danielg0:
One issue I am aware of with this PR is when running in batch mode. Having a `prompt_user` command be run on boot (by putting it in neomuttrc), running neomutt in batch mode leads to the program hanging (without displaying the prompt) in `mutt_getch` until enter is pressed, at which point it exits with an error code.

Thinking of getting around this by adding a check for `OptNoCurses` at the beginning?

flatcap:
rebased over `main`

flatcap:
Sorry for taking an age to get to this...

I've pushed a few updates:

- 5b634dc8c rename function to match command
  `parse_prompt()` -> `parse_prompt_user()`

- 1e4ca4697 flip logic
  change an `if (!x) {} else {}` to `if (x) {} else {}`

- 82a05b95f drop onechar option
  The `anykey` works nicely, so I'm not sure the `onechar` adds anything

- 482ee8e51 stub docs
  Please can you write some words :-)

Thanks

flatcap:
> batch mode

> Thinking of getting around this by adding a check for `OptNoCurses` at the beginning?

Yes, that'd be good.

flatcap:

flatcap:
Archived to: https://github.com/neomutt/neomutt-old/tree/prompt_user

PR #3532, opened 2022-09-28
Add preserve_new flag

### LINKED ISSUE #1536: mark_old = no shows all old (O) messages as new (N)
In Mutt, with "mark_old = no" I was able to preserve all message 
states/flags (N, O, !, or read) after closing/reopening folders.

In Neomutt I cannot achieve this.  With "mark_old = no" all 
O messages are automatically converted to N when reopening 
the folder.

This is different from Mutt behavior and prevents keeping 
messages in any of the N and O states indefinitely, as I 
would like.

However, if I open the same folder in Mutt at the same time, 
I see the message states (N, O) as they should be (i.e., the O 
are not converted to N).  So it seems that Neomutt conversion 
affects message states primarily in memory, when loading the 
folder contents.

alkim0:
* **What does this PR do?**
Adds a `preserve_new` flag which differentiates between old and new messages when `mark_old` flag is unset

Fix #1536

* **Other Notes**
I don't really use nntp/pop, so I wasn't sure how to set it for those.

### DIFF
--- a/docs/config.c
+++ b/docs/config.c
@@ -3593,6 +3593,13 @@
 ** .te
 */
 
+{ "preserve_new", DT_BOOL, false },
+/*
+** .pp
+** Controls whether or not NeoMutt marks should distinguish between old and new messages
+** when \fImark_old\fP is unset.
+*/
+
 { "print", DT_QUAD, MUTT_ASKNO },
 /*
 ** .pp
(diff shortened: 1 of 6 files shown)


gahr:
Hi, thanks for the PR.

In my opinion, an additional flag should not be needed. With the IMAP backend and `mark_old` unset, I am able to set emails to N or O, close neomutt, reopen, and have those two different flags preserved.

I did a quick test and I think I noticed that this is indeed not the case for maildir or mbox types of folders, where O resolves to N on reopen. I think this is a bug and I would prefer a solution that fixes this issue on those backends specifically.

alkim0:
Got it. I will make the appropriate changes and open a new pr.

gahr:
Thanks!

PR #3536, opened 2022-10-11
Use pkg-config to locate gpg-error

gahr:
Arch Linux' gpg-error package comes without a `gpg-error-config` binary. Let's keep that as a fallback use pkg-config preferably.

### DIFF
--- a/auto.def
+++ b/auto.def
@@ -528,9 +528,9 @@ if {[get-define want-gpgme]} {
   }
   define-feature gpgme
 
-  msg-checking "Checking for gpg-error..."
-  if {1} {
+  if {![get-define want-pkgconf] || ![pkgconf false gpg-error]} {
     # Locate gpg-error-config
+    msg-checking "Checking for gpg-error..."
     set gpg_error_config_guess [file join $gpgme_prefix bin gpg-error-config]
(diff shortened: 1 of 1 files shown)


flatcap (approved):
LGTM (for what that's worth :-)

vt-alt:
```
Checking for gpg-error...Checking for gpg-error-config...no
Error: Unable to find gpg-error-config
Try: 'configure --help' for options
```

This is a bit strange though. `gpg-error-config` previously used by default when `libgpg-error` is installed. `gpg-error-config` is removed from default installation in [libgpg-error 1.46](https://dev.gnupg.org/source/libgpg-error/browse/master/NEWS) and disappearing from distribution packages:

```
* The gpg-error-config command is not installed by default, because
it is now replaced by use of pkg-config/gpgrt-config with
gpg-error.pc.  Supply --enable-install-gpg-error-config configure
option, if it's really needed.
```
Now we should add `--pkgconf` to neomutt's `./configure` run for this to take effect.

Why is `pkg-config` usage is non-default?

gahr:
Yeah, I think nowadays it might make sense to use pkg-config by default: PR at #3537

PR #3538, opened 2022-10-12
Add support for compiling with undefined behaviour sanitizer

pmatos:
This adds `--ubsan` as a configure option. Adds compilation flag `--fsanitize=undefined`.


pmatos:
Just noticed this for ASAN:
https://neomutt.org/dev/analysis/asan 

Do we want something similar now for ubsan?

gahr (approved):
Works fine, thanks!

flatcap (approved):
Lovely.  Thanks

PR #3541, opened 2022-10-13
Offsetting a NULL, even by 0, is UB

gahr:
See https://reviews.llvm.org/D67122


flatcap (approved):
Hmm... that's something I'd never thought about

PR #3547, opened 2022-10-24
feat(notmuch): allow usage of notmuch profiles

Austin-Ray:
Introduce the `nm_config_profile` option for users to use notmuch profiles. It has a default of an empty string so notmuch will try `NOTMUCH_PROFILE` or the default value (`default` for directories, empty string for config files.)


gahr (approved):
LGTM

flatcap (approved):

PR #3548, opened 2022-10-25
Extend color objects to support pattterns

gahr:
This is a PoC of adding support for patterns to the index_date color object in a backwards compatible way. The idea is to eventually support patterns in all color objects, and to fall back to a match-all pattern (`.*`) if no pattern is given to avoid breaking existing configurations.

I don't know the rationale for supporting patterns in only a subset of color objects - perhaps for performance reasons? - , so I might be missing something here.


flatcap:
Nice!
I haven't tested it, but that looks perfectly reasonable.

I don't see any problems optional patterns.

> I don't know the rationale for supporting patterns in only a subset of color objects

Nor I, but...
It's possible the that author of the index-color feature didn't understand enough of the colour code,
or wanted to keep the patch small (it was maintained out-of-tree).

gahr:
Updated to make all index objects support a pattern.

flatcap:
Rebased, clang-format, sorted alphabetically

flatcap (commented):
I can't review files you haven't altered...,
so here are some places in the docs that will need updating.
Thanks

- https://github.com/neomutt/neomutt/blob/ddc83de16d210ed817606af7c63ea37cdb113ef5/docs/manual.xml.head#L4827
- https://github.com/neomutt/neomutt/blob/ddc83de16d210ed817606af7c63ea37cdb113ef5/docs/manual.xml.head#L5090
- https://github.com/neomutt/neomutt/blob/ddc83de16d210ed817606af7c63ea37cdb113ef5/docs/manual.xml.head#L5451
- https://github.com/neomutt/neomutt/blob/ddc83de16d210ed817606af7c63ea37cdb113ef5/docs/manual.xml.head#L14458
- https://github.com/neomutt/neomutt/blob/ddc83de16d210ed817606af7c63ea37cdb113ef5/docs/manual.xml.head#L19322
- https://github.com/neomutt/neomutt/blob/ddc83de16d210ed817606af7c63ea37cdb113ef5/docs/manual.xml.head#L19570

flatcap (approved):

PR #3549, opened 2022-10-28
Refactor `msg` in `notmuch/db.c`'s `nm_db_do_open()`

Austin-Ray:
Small refactor around the `msg` variable in `notmuch/db.c`.
 - Removes a version check to improve function readability. There's negligible performance impact of an unused null pointer on the stack.
 - Always frees `msg` if it's not null.

### DIFF
--- a/notmuch/db.c
+++ b/notmuch/db.c
@@ -171,7 +171,6 @@ notmuch_database_t *nm_db_do_open(const char *filename, bool writable, bool verb
       if (msg)
       {
         mutt_error(msg);
-        FREE(&msg);
       }
       else
       {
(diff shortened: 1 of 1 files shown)


notmuch/db.c
```diff
@@ -189,6 +184,11 @@ notmuch_database_t *nm_db_do_open(const char *filename, bool writable, bool verb
     }
   }
 
+  if (msg)
+  {
+    FREE(&msg);
```
flatcap, line 189: It's safe to call `FREE()` / `mutt_mem_free()` with a `NULL` pointer, so we don't need the `if (msg)`

Austin-Ray: Neat. Fixed.

flatcap (approved):

PR #3551, opened 2022-10-29
notmuch: document `get_nm_message()` error value

Austin-Ray:
Document that `get_nm_message()` returns `NULL` if an errors occurs, which sets `id` or `db` to `NULL`.

---

I was pretty confused why I was getting `NULL` during debugging when the function didn't mention that as a valid return value.


flatcap (approved):
👍🏻 for docs :-)

PR #3552, opened 2022-11-02
fixed possible fail in mutt_date_parse_imap

ilyakurdyukov:
Obvious bug, this is supposed to be a string, but a struct is passed instead, fails on some uncommon calling conventions.


flatcap (approved):
Nice catch!
I've had a look through the other ~80 uses of `TEST_CASE()` and they look OK.
Thanks

PR #3555, opened 2022-11-07
Use pkgconf to find GPGMe

### LINKED ISSUE #3553: Building on Debian with libgpgme-dev 1.18.0-2 fails
```
./configure --gpgme
...
Checking for GPGME...Checking for gpgme-config...no
Error: Unable to find gpgme-config
Try: 'configure --help' for options
```

Related change in Debian packaging:
https://salsa.debian.org/debian/gpgme/-/commit/fcd36b83bbe828726194e7b39203dcaaaa0af469

Landed in 1.18.0-2:
https://metadata.ftp-master.debian.org/changelogs//main/g/gpgme1.0/gpgme1.0_1.18.0-2_changelog

Remove from `neomutt` too or at least don't fail when there's no `gpgme-config`?

gahr:
The first GPGMe version to ship a pkg-config file is 1.13.0, released in March 2019. See this [upstream commit](https://dev.gnupg.org/rMf3e60521899e6126229b6efedc9f011b84122e11). I think it's ok to assume at least 1.13.0 is present, given that we already use pkg-config for gpg-error.

Fixes #3553

### DIFF
--- a/Makefile.autosetup
+++ b/Makefile.autosetup
@@ -601,7 +601,7 @@ $(PWD)/mutt:
 LIBNCRYPT=	libncrypt.a
 LIBNCRYPTOBJS=	ncrypt/config.o ncrypt/crypt.o ncrypt/cryptglue.o \
 		ncrypt/crypt_mod.o
-@if HAVE_GPGME
+@if HAVE_PKG_GPGME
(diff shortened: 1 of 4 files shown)


flatcap (approved):
Nice!  (and **so** much tidier :-)

PR #3558, opened 2022-11-08
avoid reapplying limit patterns on already-limited emails

### LINKED ISSUE #2763: Limit gets recomputed when returning to index
## Expected Behaviour

When using an expensive limit (e.g. `~b`) and returning from the pager to the index, the index should show immediately.

## Actual Behaviour

Neomutt takes a long time to return to the index, for about as long as it took to initially apply the limit, presumably since it is recomputing it.

- Always

Commits 86afffa524992e2ddda49a1f09c477cf1d4dc88f and d8d765f32030c69f433a82d3622fceab04bda14d.

## NeoMutt Version

```
NeoMutt 20201127

System: Linux 5.4.80-2-lts (x86_64)
GPGME: 1.15.0
GnuTLS: 3.6.15
libnotmuch: 5.3.0
storage: kyotocabinet, gdbm, bdb, lmdb
compression: lz4, zlib, zstd

Configure options: --prefix=/usr --sysconfdir=/etc --libexecdir=/usr/lib --gpgme --sqlite --autocrypt --lua --notmuch --gss --gnutls --sasl --with-ui=ncurses --with-idn2=/usr --disable-idn --idn2 --bdb --lmdb --kyotocabinet --gdbm --lz4 --zlib --zstd

Compilation CFLAGS: -march=x86-64 -mtune=generic -O2 -pipe -fno-plt -std=c99 -fno-delete-null-pointer-checks -D_ALL_SOURCE=1 -D_GNU_SOURCE=1 -D__EXTENSIONS__ -I/usr/include -I/usr/include/lua5.3 -DNCURSES_WIDECHAR -I/include -I/usr/include/

Default options:

Compile options:
  -locales_hack +lua +meta -mixmaster +nls +notmuch -openssl +pgp +regex +sasl 
  +smime +sqlite +start_color +sun_attachment +typeahead 
MAILPATH="/var/mail"
SENDMAIL="/usr/sbin/sendmail"
SYSCONFDIR="/etc"

```

igor47:
* **What does this PR do?**

currently, whenever we update the index, we also re-apply the index pattern on all messages. this is problematic because (a) this can be slow and (b) if a user has been working with a limited index, an incoming new message can suddenly alter that index.

we attempt to avoid re-applying the limit to all messages, and instead only apply it to new messages. to do this, we keep track, in the `Email` struct, of whether the message has already had the limit applied via a boolean `limit_applied` flag. we retain existing visiblity settings for those messages, and only apply new visibility settings when `limit_applied` is not already set.

fixes #2763


gahr:
Looking at the description, it is not clear to me whether `limit_applied` is supposed to mean
1. this email was visited when applying a limit
2. this email was visited when applying a limit + this email is included in the limit

If 1. holds, I would expect to see `limit_applied = true` whenever `visible` is changed.
If 2. holds, I would expect to see `limit_applied = true` whenever `visible = true`.

But I am not seeing it, e.g., here https://github.com/neomutt/neomutt/blob/76683846216fd62b9bcd0ab29cb8214885172ab1/pattern/pattern.c#L227

Can you please elaborate the approach? Thanks!

igor47:
> it is not clear to me whether `limit_applied` is supposed to mean
> 
>     1. this email was visited when applying a limit
>     2. this email was visited when applying a limit + this email is included in the limit

it's (1) -- i renamed the flag to `limit_visited` to make it more clear. the `visible` flag still controls whether the message is visible after the limit is applied.

> If 1. holds, I would expect to see `limit_applied = true` whenever `visible` is changed. If 2. holds, I would expect to see `limit_applied = true` whenever `visible = true`.
> 
> But I am not seeing it, e.g., here
> 
> https://github.com/neomutt/neomutt/blob/76683846216fd62b9bcd0ab29cb8214885172ab1/pattern/pattern.c#L227

that's the one spot where they're not both set together. after thinking about it, this is actually a third spot where we iterate over all messages to apply a limit, the other two are in `update_index_threaded` and `update_index_unthreaded`. but these are kind of different use cases. the other two functions will be reapplied when the mbox changes, like when new messages come in, and the goal of this PR is to prevent those changes from re-scanning the entire inbox.

this second place (`mutt_limit_current_thread`) is only executed on keystroke from `index/functions.c`. i can't actually tell what happens when new mail arrives, because the `limit_current_thread` is not a pattern that's saved in `mv->pattern`. as best i can tell, if you're limited to current thread, and a new message forces an index update, you'll end up just displaying all messages again.

so... i think actually the `limit_visited` flag has no purpose in that function, and it doesn't really matter whether i set it, or what i set it to.

igor47:
i opened a separate issue (#3561) re: the inconsistent behavior or `limit-current-thread`

index/dlg_index.c
```diff
@@ -436,18 +436,21 @@ static void update_index_threaded(struct MailboxView *mv, enum MxStatus check, i
     for (int i = 0; i < m->msg_count; i++)
     {
       struct Email *e = m->emails[i];
-      if ((e->vnum != -1) || mutt_pattern_exec(SLIST_FIRST(mv->limit_pattern),
-                                               MUTT_MATCH_FULL_ADDRESS, m, e, NULL))
+      if (e->limit_visited) continue;
```
gahr, line 439: How about handling `limit_visited` in a single block?

```cpp
if (e->limit_visited)
{
  continue;
}
e->limit_visited = true;
```

pattern/pattern.c
```diff
@@ -407,13 +407,16 @@ int mutt_pattern_func(struct MailboxView *mv, int op, char *prompt)
       /* new limit pattern implicitly uncollapses all threads */
       e->vnum = -1;
       e->visible = false;
+      e->limit_visited = false;
```
gahr, line 410: Don't we want to set this to `true` no matter whether the pattern matches this email?

igor47:
i think i've addressed all feedback and also tested this on my inbox. the limit pattern is no longer re-applied, so that's an improvement! however, i still loose my cursor position in the inbox whenever a new message arrives. let me poke around a little more to see why it's happening, i think it's because of the call to `mutt_sort_headers`. i kinda wish the cursor position was to a specific email and not to a specific index location, because index locations change in threaded mode.

gahr:
> > i still loose my cursor position in the inbox whenever a new message arrives.
> 
> i'm pretty lost on this one.

I think flatcap is working on something similar in #3560

gahr (approved):
Thanks, LGTM!

flatcap:
Lovely job.
Thank you very much :heart:

Issue #3544, opened 2022-10-15
Pager jumps to different message when new messages arrive and pager_index_lines is set


jpalus:
## Expected Behaviour

When reading message in pager I would expect to have it open until I close it.

## Actual Behaviour

With `pager_index_lines` it appears that sometimes `neomutt` jumps to different message, like it wants to keep viewing message with same index on the list even though after new messages arrived message with that index is completely different.

## Steps to Reproduce

Prepare sample maildir with 10 test messages:
```sh
$ mkdir -p mdir/{cur,new,tmp}
$ for i in `seq 11 20 `; do
  cat > mdir/new/test$i  <<EOF
Date: Sat, 15 Oct 2022 12:$i:00 +0000
From: foo@example.com
To: bar@example.com
Subject: test$i

test$i
EOF
done
```
Sample config:
```
set pager_index_lines=6
set spoolfile='mdir'
set mbox_type='Maildir'
set sort=reverse-date
```
Open `neomutt`  with sample config and view ie 3rd message:
![screenshotwl-20221015152755](https://user-images.githubusercontent.com/484155/195989517-4c36550f-bd53-4454-95ea-05a0409ff0be.png)
Generate 10 more messages:
```sh
for i in `seq 21 30 `; do
  cat > mdir/new/test$i  <<EOF
Date: Sat, 15 Oct 2022 12:$i:00 +0000
From: foo@example.com
To: bar@example.com
Subject: test$i

test$i
EOF
done
```
Check that pager shows different message now:
![screenshotwl-20221015152842](https://user-images.githubusercontent.com/484155/195989583-8716ef17-a534-4e79-acb1-9597ed8529ad.png)

## How often does this happen?

- Always

## NeoMutt Version

```
NeoMutt 20220429
Copyright (C) 1996-2022 Michael R. Elkins and others.
NeoMutt comes with ABSOLUTELY NO WARRANTY; for details type 'neomutt -vv'.
NeoMutt is free software, and you are welcome to redistribute it
under certain conditions; type 'neomutt -vv' for details.

System: Linux 6.0.0-1 (aarch64)
ncurses: ncurses 6.3.20220416 (compiled with 6.3.20211120)
libidn: 1.41 (compiled with 1.38)
GPGME: 1.17.1
OpenSSL: OpenSSL 3.0.2 15 Mar 2022
storage: bdb, lmdb
compression: lz4, zlib, zstd

[...]

PR #3560, opened 2022-11-09
index: fix cursor after new mail

flatcap:
When new mail arrives, the backends (maildir/imap) update the Mailbox and send notifications.
In the frontend, the Mailbox view (`MView`) observes these changes (`mview_mailbox_observer()`).
(`MView` was previously known as the `Context`)

Before we do anything that might affect the ordering of the `Email`s, remember the sequence number of the current `Email`.
Afterwards, search for the `Email` by the sequence number.

Fixes: #3544

### DIFF
--- a/email/email.c
+++ b/email/email.c
@@ -77,7 +77,7 @@ void email_free(struct Email **ptr)
  */
 struct Email *email_new(void)
 {
-  static size_t sequence = 0;
+  static size_t sequence = 1;
 
   struct Email *e = mutt_mem_calloc(1, sizeof(struct Email));
 #ifdef MIXMASTER
--- a/mview.c
+++ b/mview.c
@@ -34,10 +34,12 @@
 #include "core/lib.h"
 #include "mview.h"
 #include "imap/lib.h"
+#include "menu/lib.h"
 #include "ncrypt/lib.h"
 #include "pattern/lib.h"
 #include "mutt_header.h"
 #include "mutt_thread.h"
+#include "mview.h"
 #include "mx.h"
 #include "score.h"
 #include "sort.h"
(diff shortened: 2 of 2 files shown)


igor47:
didn't work for me for the specific issue i'm seeing:
1. apply a limit pattern
2. scroll to a middle message in the limited view
3. send myself a new email from a different account

i expect that i would remain where it was. instead, the message "New mail in this mailbox" is displayed, and the cursor position is lost/non-existent. pressing `j` then selects the first message in the mailbox/limited view.

auouymous:
Why do the same unchanged commits keep getting force pushed?

flatcap:
Ah!  That's me `git rebase`ing my work branch over `main` so that I have the latest source.

auouymous:
The latest push appears to fix the issue for me. I'm installing it on my main client and will let you know if it has any problems. Thank you!

igor47:
does not fix it for me. i'm in a limited view, and when i get the notification "new mail in inbox", i still loose my current cursor position. my limited view did not include the new message. my configure (on debian 11.6):

```
/configure --build=x86_64-linux-gnu --prefix=/usr {--includedir=${prefix}/include} {--mandir=${prefix}/share/man} {--infodir=${prefix}/share/info} --sysconfdir=/etc --localstatedir=/var --disable-silent-rules {--libdir=${prefix}/lib/x86_64-linux-gnu} {--libexecdir=${prefix}/lib/x86_64-linux-gnu} --disable-maintainer-mode --disable-dependency-tracking --mandir=/usr/share/man --libexecdir=/usr/libexec --with-mailpath=/var/mail --gpgme --lua --notmuch --with-ui --gnutls --idn --mixmaster --sasl --tokyocabinet --disable-doc --disable-nls
```

flatcap:
Thanks for the feedback :heart:

I spent the day working on this and made _some_ progress, but I'm not done yet.
The trouble is the Pager combined with limit.

There are eight test cases, combinations of:
- Index vs Pager
- Limit vs No limit
- Inotify enabled vs disabled

auouymous:
Oh, using a limit breaks both index and pager by setting cursor to an invalid message, and closes pager. This happens for new messages inserted at top or bottom, matching or not matching the limit.

Using inotify.

auouymous:
New patch fixed the broken cursor after receiving new mail, but pager is still closed when new mail arrives before or after the cursor when a limit is set. It also introduced some new bugs.

Closing pager or changing messages after any mail as arrived does not mark message as old. Switching to another message or opening and closing pager again will mark the message as old.

I think the messages are being marked as old in the UI but not in the backend, and a refresh switches them back to new. So the next two bugs might only happen when messages are new or recently marked as old with no other actions that could sync them to the backend.

Receiving mail while in index or pager causes all old messages to be marked as new, and shows "Mailbox was externally modified".

While in pager, receiving mail before the cursor marks current message as old, updates the index so current message shifts down one, and then marks current message as new.

auouymous:
Why was this pushed to main in 2523464c1d6b293dc5f512944ed84f96823c6c9d?

auouymous:
@flatcap `git revert 2523464c1d6b293dc5f512944ed84f96823c6c9d` cleanly reverts the patch out of main branch (assuming it was accidentally pushed). The only difference is the double `mview.h` include that has since been removed.

PR #3576, opened 2022-12-07
Don't wait after timeout from getaddrinfo_a

gahr:
This is an attempt to speedup recovery from getaddrinfo_a timeouts. I am not familiar with this family of APIs and I don't have a linux box to test this on, so I have no idea if this is correct.

### DIFF
--- a/conn/getdomain.c
+++ b/conn/getdomain.c
@@ -56,16 +56,18 @@ static struct addrinfo *mutt_getaddrinfo_a(const char *node, const struct addrin
    * If it takes longer, the system is mis-configured and the network is not
    * working properly, so...  */
   struct timespec timeout = { 0, 100000000 };
-  struct gaicb *reqs[1];
-  reqs[0] = mutt_mem_calloc(1, sizeof(*reqs[0]));
-  reqs[0]->ar_name = node;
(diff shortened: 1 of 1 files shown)


flatcap (commented):
- add a cast to avoid a warning
- split out the refactoring

conn/getdomain.c
```diff
@@ -56,30 +56,34 @@ static struct addrinfo *mutt_getaddrinfo_a(const char *node, const struct addrin
...
     {
       mutt_debug(LL_DEBUG1, "timeout\n");
       /* request is not finished, cancel it to free it safely */
       if (gai_cancel(reqs[0]) == EAI_NOTCANCELED)
       {
-        while (gai_suspend((const struct gaicb *const *) reqs, 1, NULL) != 0)
-          continue;
+        // try once more for half-a-second, then bail out
+        timeout.tv_nsec = 50000000;
+        gai_suspend((const struct gaicb *const *) reqs, 1, &timeout);
```
flatcap, line 79: Without a timeout, `gai_suspend()` will block, so the original code is quite mystifying.
I've chosen 0.5s as a timeout before we give up.

flatcap (approved):

PR #3592, opened 2023-01-11
Try to create cur/new on access

### LINKED ISSUE #3586: Add (or ignore) missing 'new' and 'tmp' dirs when opening a maildir
I've used git to sync a maildir tree to another machine. The problem is that git doesn't add empty directories, so often you will get a directory with files like ```~/Mail/foo/cur/xxxxx.xxxx.xxx```, but no ```~/Mail/foo/new``` or ```~/Mail/foo/tmp```. Such dirs will provoke ```mx_path_probe() unable to stat =foo/: No such file or directory (errno 2)``` in the log, and the user interface status line will report ```(no mailbox)```. Once I realized the problem, I manually created those two directories, and the maildir does open correctly in NeoMutt.

Perhaps NeoMutt should create empty ```new``` and ```tmp``` directories when opening such a maildir. NeoMutt actually already does this in certain circumstances, for example, saving a message to such a maildir will "repair" it.

I decided to report this issue in case it will help others who happen to have the issue, but I understand that this may be out of scope for the project, so I labeled this issue as a feature request rather than a bug report. Feel free to close the suggestion.

gahr:
Fixes #3586

### DIFF
--- a/bcache/bcache.c
+++ b/bcache/bcache.c
@@ -333,7 +333,7 @@ int mutt_bcache_list(struct BodyCache *bcache, bcache_list_t want_id, void *data
   struct dirent *de = NULL;
   int rc = -1;
 
-  if (!bcache || !(d = opendir(bcache->path)))
+  if (!bcache || !(d = mutt_file_opendir(bcache->path, MUTT_OPENDIR_NONE)))
     goto out;
 
   rc = 0;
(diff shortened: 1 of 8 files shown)


flatcap (approved):
Looks good.
I've backed out some of the `<dirent.h>` changes where the code derefs `struct dirent`.
(otherwise `iwyu` will put them back again)

PR #3596, opened 2023-01-19
Do not truncate "\r\n" in SMTP communication

rayfordshire:
* **What does this PR do?**

Prevent Neomutt from sending an incomplete line (without \r\n) to an SMTP server if the username or password is to long.

### DIFF
--- a/send/smtp.c
+++ b/send/smtp.c
@@ -830,7 +830,7 @@ static int smtp_auth_login(struct SmtpAccountData *adata, const char *method)
   (void) method; // This is LOGIN
 
   char b64[1024] = { 0 };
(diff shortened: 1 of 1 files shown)


flatcap (approved):
Lovely, thanks!

> This also fixes a GCC warning about format-truncation

Yeah, there are quite a lot of these.
We tend to fix all warnings, but these are a bit of a pain.
Most, will never be a problem so it can be hard to spot the ones that might.

In some places, we're using `struct Buffer` rather than a fixed `char[]`.
Unfortunately, upgrading `char[]` often means altering a lot of dependent code.

In some distant future, it'd be nice to jump to `std::string` :-)

PR #3597, opened 2023-01-19
Allow mutt to handle mails dating Wed 31 Dec 22:13:21 1969

rayfordshire:
* **What does this PR do?**

Two things:

1) Make the `mutt_date_gmtime()` and `mutt_date_localtime()` more robust by checking the return value of `gmtime()` and `localtime()` for a failure condition.  Since these function do get passed user supplied data, it is possible that these fail.
In the case of a failure return a default date. I opted for the common 1970-01-01 00:00:00.
Note: There seems to be no *guaranteed* way to produce a valid `tm` structure, all function producing one could fail. However, passing the above mentioned date seems pretty safe so I do not check the return value of `mktime()`.

2) Get rid of `MUTT_DATE_NOW`

`MUTT_DATE_NOW` is the date *Wed 31 Dec 22:13:21 1969* which is used to indicate not some cold day somewhere in the past but "now".  Remove the handling of this magic date and use dedicated functions (`mutt_date_gmtime_now()` and `mutt_date_localtime_now()`) to produce the current date.

The replacement on the caller side was pretty simple:
`mutt_date_gmtime(MUTT_DATE_NOW)` ==> `mutt_date_gmtime_now()`
`mutt_date_localtime(MUTT_DATE_NOW)` ==> `mutt_date_localtime_now()`

The only place which became slightly harder to read was a single call to `mutt_date_localtime_format(..., MUTT_DATE_NOW)` which had to be replaced with `mutt_date_localtime_format(..., mutt_date_epoch())`.

---

If you don't like the second change, I'd at least advise to incorporate the first one.

Have fun
Ray

### DIFF
--- a/autocrypt/autocrypt.c
+++ b/autocrypt/autocrypt.c
@@ -294,7 +294,7 @@ int mutt_autocrypt_process_autocrypt_header(struct Email *e, struct Envelope *en
 
   /* Ignore emails that appear to be more than a week in the future,
    * since they can block all future updates during that time. */
-  if (e->date_sent > (mutt_date_epoch() + (7 * 24 * 60 * 60)))
+  if (e->date_sent > (mutt_date_now() + (7 * 24 * 60 * 60)))
     return 0;
 
   for (struct AutocryptHeader *ac_hdr = env->autocrypt; ac_hdr; ac_hdr = ac_hdr->next)
@@ -437,7 +437,7 @@ int mutt_autocrypt_process_gossip_header(struct Email *e, struct Envelope *prot_
 
   /* Ignore emails that appear to be more than a week in the future,
    * since they can block all future updates during that time. */
-  if (e->date_sent > (mutt_date_epoch() + (7 * 24 * 60 * 60)))
+  if (e->date_sent > (mutt_date_now() + (7 * 24 * 60 * 60)))
     return 0;
 
   struct Buffer *keyid = mutt_buffer_pool_get();
--- a/browser/browser.c
+++ b/browser/browser.c
@@ -240,7 +240,7 @@ static const char *folder_format_str(char *buf, size_t buflen, size_t col, int c
         else
         {
           static const time_t one_year = 31536000;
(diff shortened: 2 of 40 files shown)


flatcap:
I've just:
- Run `clang-format` on the code
- Rebased over `main`

flatcap (changes requested):
OK... I've had a good read, a think and experimented a little.

Adding the error-checking to `mutt_date_localtime()` and `mutt_date_gmtime()` looks good,
though I found it **really** hard to make `localtime_r()` and `gmtime_r()` actually _fail_.
It only happens if `int tm.tm_year` overflows.
That means that the original `time_t` has to be **enormous** -- `LONG_MAX` worked ;-)

Having an error message, means translating it too,
so I suggest demoting the `mutt_error()` to a `mutt_debug(LL_DEBUG1, ...)`
That way, it'll still get logged.

I agree that `MUTT_DATE_NOW` isn't great, but I think that having `mutt_date_localtime()` **and** `mutt_date_localtime_now()`, etc, is overkill.
As the `_now()` versions just make a call to `mutt_date_epoch()`, I think we should pass that instead, e.g.
```diff
- tm = mutt_date_localtime(MUTT_DATE_NOW);
+ tm = mutt_date_localtime(mutt_date_epoch());
```

Perhaps `mutt_date_epoch()` ought to have `now` in its name.

gahr:
> Perhaps `mutt_date_epoch()` ought to have `now` in its name.

Yes

flatcap:
OK, while you're at it, please can you rename:
```sed
s/mutt_date_epoch/mutt_date_now/
s/mutt_date_epoch_ms/mutt_date_now_ms/
```

Thanks!

rayfordshire:
I did

* rebase on latest `main`
* change `mutt_error` into `mutt_debug`
* renamed `mutt_date_epoch()`/`mutt_date_epoch_ms()` to `mutt_date_now()`/mutt_date_now_ms()`. This included renaming the test files of those functions.
* Replaced `MUTT_DATE_NOW` with `mutt_date_now()`

flatcap:
No changes, just `clang-format`

flatcap (approved):
Lovely job.  Thanks!

PR #3600, opened 2023-01-20
Use smime_sign_as when signing S/MIME messages

whitney-cumber:
Do not use pgp_sign_as but smime_sign_as when signing messages.

* **What does this PR do?**

I believe there is a copy and past error as the signing for S/MIME queries the **pgp**_sign_as instead of the **smime**_sign_as config varialble.


flatcap (approved):
Damn!  That was careless.
Thanks!

PR #3601, opened 2023-01-20
Recalc envelope window if smime_sign_as changes

whitney-cumber:
* **What does this PR do?**

Not sure about this but it seems that changes of `$smime_sign_as ` are not responded to.  If this is intentional, feel free to close this PR.


flatcap (approved):
> If this is intentional

Nope.  That looks like it's been overlooked.

Thanks!

PR #3604, opened 2023-01-20
Rename ask_follow_up to ask_followup_to

whitney-cumber:
Rename the config variable `ask_follow_up` to `ask_followup_to` so that its name matches the header name it is guarding (namely `Followup-To:`) and to be consistent with the config variable `followup_to_poster`.

Create synonym `ask_follow_up` to not break existing configs.

### DIFF
--- a/docs/config.c
+++ b/docs/config.c
@@ -189,7 +189,7 @@
 */
 
 #ifdef USE_NNTP
-{ "ask_follow_up", DT_BOOL, false },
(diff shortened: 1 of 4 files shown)


flatcap (approved):
Perfect.  Thanks!

flatcap:
Pushed to the website: https://neomutt.org/guide/reference#ask-followup-to

Issue #3615, opened 2023-01-22
`mono` and `color` command print unhelpful/wrong error message


whitney-cumber:
## Expected Behaviour

A helpful error message pointing to the location of the mistake.

## Actual Behaviour

Misleading/wrong error message.

## Steps to Reproduce

0. Start Neomutt: `neomutt -F /dev/null`
1. Issue the command: `:color iDoNotExist red blue`  (or its `mono` analogue)
3. Observe the error message `red: unknown command`

### Extra Info

I hope we can all agree that that is not a good error message: neither was `red` the command nor was the mistake at `red`
The mistake here was to specify some invalid object to colour.

A better error message would have been:
`color: 'iDoNotExist' is not a valid object to color' or `color: invalid object to color: iDoNotExist`

## How often does this happen?

- Always

## NeoMutt Version

```
20220429
```

PR #3629, opened 2023-01-25
config: skip line rest of line after a warning

flatcap:
- warn about colours not objects
  (we already have a translation)

- config: skip line rest of line after a warning
  Don't try to parse the rest of a line, after a mistake is found

Fixes: #3615


flatcap:
Sample output:
```
Warning in test.rc, line 1: iDoNotExist: no such colour
```

PR #3633, opened 2023-01-25
Rename `tmpdir` to `tmp_dir`

whitney-cumber:
Rename `tmpdir` to `tmp_dir` to bring it in line with other `*_dir` config variables like `attach_save_dir`, `autocrypt_dir`, `news_cache_dir`.

A synonym `tmpdir` is provided for backwards compatibility.

This rename can be seen as controversal:

Pro: uniform naming scheme
Con: $TMPDIR (the environment variable) is a well known, well, environment variable and `tmpdir`/`tmp_dir` is closely related.

If you don't like it, then close it.

### DIFF
--- a/docs/config.c
+++ b/docs/config.c
@@ -5290,7 +5290,7 @@
 ** A value of zero or less will cause NeoMutt to never time out.
 */
 
-{ "tmpdir", DT_PATH, TMPDIR },
+{ "tmp_dir", DT_PATH, TMPDIR },
 /*
 ** .pp
 ** This variable allows you to specify where NeoMutt will place its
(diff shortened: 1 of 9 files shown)


flatcap:
> This rename can be seen as controversial:

hehe, yeah.
I'm not sure how I feel yet.

In an ideal world, we'd eventually deprecate all the old synonyms.
Realistically, while `mutt` exists, we're unlikely to do so.

This might be a case of: rename to `tmp_dir`, but always accept `tmpdir`.

One of my many fantastical ideas for NeoMutt is to copy `readline`s `completion-map-case` feature.
This allows you to auto-complete and have it "fix" any hyphens that should be underscores, or vice versa.

There are plenty of commands with hyphens and plenty with underscores.

https://github.com/neomutt/neomutt/blob/57a9959929f48a41ca1c900450d28af69f67b1a1/mutt_commands.c#L77-L78

PR #3634, opened 2023-01-25
Use the (new) application/pkcs7-* MIME types

whitney-cumber:
Do not generate the legacy MIME types application/x-pkcs7-* (note the "x-") but the standardised types application/pkcs7-*.

The application/pkcs7-* MIME types are standardised in [RFC8551] section 3.2.  The (historic) [RFC2311] states in section C.1

   Some early implementations of S/MIME agents used the following MIME
   types:

   application/x-pkcs7-mime
   application/x-pkcs7-signature
   application/x-pkcs10

   In each case, the "x-" subtypes correspond to the subtypes described
   in this document without the "x-".

[RFC8551] https://datatracker.ietf.org/doc/rfc8551/
[RFC2311] https://datatracker.ietf.org/doc/rfc2311/

### DIFF
--- a/docs/config.c
+++ b/docs/config.c
@@ -4474,7 +4474,7 @@
 /*
 ** .pp
 ** This format string specifies a command which is used to decrypt
-** \fCapplication/x-pkcs7-mime\fP attachments.
+** \fCapplication/pkcs7-mime\fP attachments.
 ** .pp
 ** The OpenSSL command formats have their own set of \fCprintf(3)\fP-like sequences
 ** similar to PGP's:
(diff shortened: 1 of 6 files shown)


flatcap (approved):
Nice!

Checking the logs, `x-pkcs7` was added 17 years ago, then (probably) forgotten about.
Given that the recent RFC is 3 years old, it seems fair to use `pkcs7`.

I've added a missing line to `docs/smime-notes.txt`
The line you added was missing the continuation: `-print_certs...`

PR #3639, opened 2023-01-27
Implement `Smime:` pseudo header for composing

rayfordshire:
When composing messages NeoMutt supports some pseudo headers like "Attach:" or "Pgp:".

Implement the "Smime:" header which allows to select S/MIME encryption/signing just like the "Pgp:" header. The valid values for the header are the same as for "Pgp:", namely "E" (encrypt), "S" (sign), and "S <id>" (sign as).

* **What does this PR do?**

Brings support of S/MIME closer on par with PGP.

* **Rambling Notes**

I noticed that postponed mails have an "X-Mutt-PGP:", "X-Mutt-SMIME:" header.  Maybe it might be worth supporting those too when composing a message? (maybe even saying that those should be used as they avoid ambiguity??)

### DIFF
--- a/docs/manual.xml.head
+++ b/docs/manual.xml.head
@@ -3012,6 +3012,27 @@ color sidebar_divider   color8  default     <emphasis role="comment"># Dark grey
           </para>
         </sect3>
 
+        <sect3 id="smime-header">
+          <title>Smime: Pseudo Header</title>
+          <para>
+            If you want to use S/MIME, you can specify
+          </para>
+          <para>
+            <literal>Smime:</literal> [
+            <literal>E</literal> |
+            <literal>S</literal> |
+            <literal>S</literal>
+            <emphasis>&lt;id&gt;</emphasis> ]
+          </para>
+          <para>
+            <quote>E</quote> selects encryption, <quote>S</quote> selects signing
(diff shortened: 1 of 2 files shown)


flatcap:
> Brings support of S/MIME closer on par with PGP.

Nice!

> postponed mails have an "X-Mutt-PGP:", "X-Mutt-SMIME:" header.
> maybe even saying that those should be used as they avoid ambiguity

Hmm...   They'd certainly be less ambiguity if all our pseudo-headers began: `X-Mutt-*`.
But, the current ones are short and convenient.
I'm not quite sold on the idea of changing the default... yet.

---

> postponed mails have an "X-Mutt-PGP:", "X-Mutt-SMIME:" header.
> Maybe it might be worth supporting those too

Not a bad idea.
Would you mind taking a look? 

The headers are checked here:
https://github.com/neomutt/neomutt/blob/9a6d97bf70c5969e273ecd970345f1f07e155d42/mutt_header.c#L341-L353

We'd need to refactor this code:
https://github.com/neomutt/neomutt/blob/9a6d97bf70c5969e273ecd970345f1f07e155d42/postpone/postpone.c#L764-L772

Thanks :heart:

flatcap (approved):

PR #3640, opened 2023-01-27
Fix `$pager_read_delay`

### LINKED ISSUE #3609: Wrong message is marked read with `$pager_read_delay` = 1
## Expected Behaviour

When moving to the next message in the pager, the origin message should be marked as read not the destination message.

## Actual Behaviour

The destination message is marked as read.

## Steps to Reproduce

Create a mailbox `=testmailbox` (or use an existing one) with two (ideally consecutive) new/unread messages.
We call the messages A and B.

0. Start Neomutt with default config: `neomutt -F /dev/null`
1. Set `$pager_read_delay` to any value > 0
2. Open =testmailbox
3. Move to message A and open it for longer than `$pager_read_delay` seconds in the builtin pager
4. Move or jump to message B (e.g. with `<next-undeleted>`, `<next-entry>`, ... or their `previous-*` counterparts, `<jump>`, ...) and immediately close the pager (more precisely close in under `$page_read_delay` seconds)
5. Observe that message A is still unread while message B is marked as read.

- Always

The moment I enabled `$pager_read_delay`.

## NeoMutt Version

```
20220429
```

## Extra Info

Tested with

* `use_threads=flat`, `$use_threads=threads`
* different move commands (`<jump>`, `<previous-undeleted>`, `<next-entry>`)
* different values to `$pager_read_delay`
* unread messages of new (`N`) and old (`O`) type

in all cases the bug exists.

rayfordshire:
* **What does this PR do?**

I don't understand the event loop (nor the "multi"threading) between pager and index fully.  However, the timer check should be done *after* the statement which takes the idle time, here waiting for user input.  They should also done *before* handling the operations as those can alter the internal state (like the current mail) leading to bugs such as #3609.

Alternatively, we can do the timer check at the end of the loop. To this end, we have to introduce two local variables `cur_mail`, `cur_mailbox` set them at the beginning of the loop and use those cached values at the end of the loop in the timer check (the `mutt_set_flag` call).

Please check if this does not break any other invariant other parts of the code relay upon.

Fixes #3609.

### DIFF
--- a/pager/dlg_pager.c
+++ b/pager/dlg_pager.c
@@ -574,10 +574,6 @@ int mutt_pager(struct PagerView *pview)
   // END OF ACT 3: Read user input loop - while (op != OP_ABORT)
   //-------------------------------------------------------------------------
 
-  if (check_read_delay(&priv->delay_read_timestamp))
-  {
-    mutt_set_flag(shared->mailbox, shared->email, MUTT_READ, true);
-  }
   mutt_file_fclose(&priv->fp);
   if (pview->mode == PAGER_MODE_EMAIL)
   {
--- a/pager/dlg_pager.c
+++ b/pager/dlg_pager.c
@@ -384,11 +384,6 @@ int mutt_pager(struct PagerView *pview)
   priv->loop = PAGER_LOOP_CONTINUE;
   do
   {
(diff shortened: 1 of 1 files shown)


flatcap (approved):
Looks good to me.
Thanks!

PR #3641, opened 2023-01-28
Devel/x mutt pseudo headers

rayfordshire:
### What does this PR do?

1) Consistently use the match length `plen` when comparing for header fields.  See commit message for more implementation details.

2) Accept X-Mutt-Fcc, X-Mutt-Attach, X-Mutt-PGP, X-Mutt-SMIME pseudo headers when composing a message.  These headers are already used by the postpone mechanism.

3) Support X-Mutt-Mix.  This header is already support by the postpone mechanism, I just ported it and replaced `hdr` with `e`.
Important: I have no clue what this does nor how Mixmaster works! So please check before merging.

4) It does not port X-Mutt-References into the compose menu.  If I understood it correctly, it detects whether the postponed message was a reply.  Since we come from the index/pager we know whether we replied to a message or not.  (Also I could not figure out how to set the SEND_REPLY flag from this void-method.)

5) I did not combine/factor out (as hinted in #3639) the header handling of `mutt_get_postponed()` in postpone/postpone.c and `mutt_edit_headers()` in mutt_header.c as their effects for the headers seem to have almost nothing in common except the header names.

### An RFC Reference

During the implementation I came about this wonderful RFC number 6648 [RFC6648].  Since we all know all RFCs by heart I don't need to remind you that this RFC is about *not* using "X-" prefixes in custom headers, we all know its quote:

> Creators of new parameters to be used in the context of application
> protocols:
> [...]
>   3.  SHOULD NOT prefix their parameter names with "X-" or similar
>        constructs.

However the RFC

> [...]
>   4.  Makes no recommendation as to whether existing "X-" parameters
>       ought to remain in use or be migrated to a format without the
>       "X-"; this is a matter for the creators or maintainers of those
>       parameters.

This bears the questions

1. Should we use "Mutt-Fcc" et al instead of "X-Mutt-Fcc"?
2. do 1. only in the compose mechanism or only in the postpone mechanism or both?
3. Should we even strip the prefix "X-Mutt" and so return to the original names?
4. do 3. for the postpone mechanism?
5. and depending on the answers to 1. to 4. what to do with this PR?

These questions heavily affect the second and third commit of this PR.  The first on can be cherry-picked/merged separately if you find its changes compelling.

[RFC6648] https://datatracker.ietf.org/doc/rfc6648/

The idea came up in #3639 .


flatcap:
Wow!  This looks good.
It may take me a little while to unpack it all :-)
(btw, I haven't forgotten about your date fixes)

mutt_header.c
```diff
@@ -289,7 +289,12 @@ void mutt_edit_headers(const char *editor, const char *body, struct Email *e,
     bool keep = true;
     size_t plen;
 
-    if (fcc && (plen = mutt_istr_startswith(np->data, "fcc:")))
+    // Check first for "X-Mutt-Fcc" then for "fcc" as it is the more
+    // unambiguous header name.
+    // Note: `plen` is only assigned once due to the short-circuit behaviour of
+    // `||`
```
flatcap, line 295: Hmm... You've added this comment in a few places
Are you sure?
If the string in `fcc`, then surely `plen` will be assigned a value twice.
Or can the compiler _really_ optimise that away?

postpone/postpone.c
```diff
@@ -761,27 +761,29 @@ int mutt_get_postponed(struct Mailbox *m_cur, struct Email *hdr,
        * see http://dev.mutt.org/trac/ticket/3653 */
       rc |= SEND_POSTPONED_FCC;
     }
+    /* Note: `plen` is only assigned once due to the short-circuit behaviour of
+     * `||` */
     else if (((WithCrypto & APPLICATION_PGP) != 0) &&
              /* this is generated by old neomutt versions */
-             (mutt_str_startswith(np->data, "Pgp:") ||
+             ((plen = mutt_str_startswith(np->data, "Pgp:")) ||
```
flatcap, line 768: In the original code, we're doing case-**sensitive** comparisons.
(these are headers that NeoMutt has written, rather than the user)
I think we should be forgiving (case-insensitive), like we are in `mutt_edit_headers()`.

flatcap (changes requested):
I like the first commit and I've left a couple of comments on the rest (open to debate :-)

After reading the RFC (thanks, I wasn't aware of it), I think we should drop the `X-` prefix.
(Even though our headers should never make it into the wild)

We'd still need to accept `X-*` just in case we encounter and old postponed email, etc.

Thanks!

mutt_header.c
rayfordshire: My bad that's very poorly written, what I actual meant was this:

>    // Note: If the first string comparison succeeds, `plen` is not overridden
>    // by the second one due to the short-circuit behaviour of `||`.

I added this since for less seasoned C programmers this might be hard to read/understand. (Better safe than sorry)

postpone/postpone.c
rayfordshire: I noticed that, too, but didn't want change it.  Will be rectified in the next version.

rayfordshire:
> After reading the RFC (thanks, I wasn't aware of it), I think we should drop the X- prefix.
> (Even though our headers should never make it into the wild)
>
> We'd still need to accept X-* just in case we encounter and old postponed email, etc.

Ok will do.  Just to be clear:

Compose should accept `Foo`, `X-Mutt-Foo` and `Mutt-Foo`.

(currently compose does only accept `Foo`, so we would two new variants.)

flatcap:
> Compose should accept `Foo`, `X-Mutt-Foo` and `Mutt-Foo`.

Yes.
But only mention `Foo`, `Mutt-Foo` in the docs.

rayfordshire:
I did

1) rebase on latest main
2) compose accepts: X-Mutt-Foo, Mutt-Foo, and Foo (only Mutt-Foo and Foo documented)
3) compose accepts: X-Mutt-Mix, Mutt-Mix (only Mutt-Mix documented).  Note: no Mix: header is accepted as this header is new.
4) case insensitive matching in postponed mails
5) postpone accepts Mutt-Foo
6) postpone writes Mutt-Foo (instead of X-Mutt-Foo)

flatcap:
> I did...

Fantastic
I'll have a read later.
Thanks!

flatcap (approved):
Great work.

I made a couple of small changes to the MixMaster parsing
- Add a missing `keep = false;` (so we filter out the fake header)
- Allow commas in string

Thanks!

Issue #3060, opened 2021-10-07
Add "message_id_format" variable for customizing generated Message-IDs


cwfoo:
## What would you like to be able to do?

Add `message_id_format` configuration variable for customizing generated Message-IDs. This feature was implemented in Mutt 2.1.0 (2021-06-12). [Mutt 2.1 Release Notes](http://www.mutt.org/relnotes/2.1/):

> ## Message ID Format
> 
> The Mutt 2.0 release changed the default generated Message-ID format. There are all sorts of opinions about how a Message-ID ought to be generated, though, so this release created a new configuration variable: [$message_id_format](http://www.mutt.org/doc/manual/#message-id-format).
> 
> The pre-2.0 Message-ID style can be generated by setting this to "<%Y%02m%02d%02H%02M%02S.G%c%p@%f>", for instance. There are a few other expandos that may be useful, but also note that format strings can use a [filter](http://www.mutt.org/doc/manual/#formatstrings-filters) and be generated by your own script.
> 
> Mutt doesn't perform validation on the generated value; it's your responsibility to generate legal (and unique) values if you change the format string. Please be a good netizen and customize responsibly.

billchenchina:
#4718 has implemented $message_id_format.

PR #3655, opened 2023-02-02
Use random string as Message-ID

rayfordshire:
* **What does this PR do?**

Implement the $message_id_format config variable which lets the user specify a template for the Message-ID header.  This functionality is already present in Mutt [0].

The patches try to mimic these commits though the last four are combined into a single commit as they are are essentially bugfixes to the first commit.

Mutt seems to be base64 focused while NeoMutt is base32 focused.  To be faithful to the Mutt expandos these preserve their base64 meaning.  However, we also introduce the `%R` expando (which mutt does not have) which expands to base32 encoded random bytes.

Mutt has also a completely different default Message-ID format.  This is also mimicked in this patch set as the `%z` expando.
However, the default of $message_id_format is NeoMutts old message-id format, i.e. nothing is changing for NeoMutt users.

The above expandos required to enhance NeoMutt's Base64 and Random capability, which the first couple of commits do.

Finally, there are many "questions" for the reviewer in `messageid.c`.  They are all tagged with `TODO @review`.  Thus, should be easy to spot.

[0] 
https://gitlab.com/muttmua/mutt/commit/b5a807cadb499d3078aedbdb67e4c22fc1520f88
https://gitlab.com/muttmua/mutt/commit/1f18828754ed781f2562a60414521a5e1e6b93a8
https://gitlab.com/muttmua/mutt/commit/5f5d33f832c82f14280e2c643b2039d4beb68573
https://gitlab.com/muttmua/mutt/commit/dadc0130fc83ae31f9142f238e60c0921bf18da2
https://gitlab.com/muttmua/mutt/commit/8641fb2a69df4ad035cfff3d83127336c7d896cd

#3060

### DIFF
--- a/send/sendlib.c
+++ b/send/sendlib.c
@@ -719,25 +719,42 @@ const char *mutt_fqdn(bool may_hide_host, const struct ConfigSubset *sub)
 }
 
 /**
- * gen_msgid - Generate a unique Message ID
+ * gen_msgid - Generate a random Message ID
  * @retval ptr Message ID
+ *
+ * The length of the message id is chosen such that it is maximal and fits in
+ * the recommended 78 character line length for the headers Message-ID:,
+ * References:, and In-Reply-To:, this leads to 62 available characters
+ * (excluding `@` and `>`).  Since we choose from 32 letters, we have 32^62
+ * = 2^310 different message ids.
+ *
+ * Examples:
+ * ```
+ * Message-ID: <12345678901111111111222222222233333333334444444444@123456789011>
+ * In-Reply-To: <12345678901111111111222222222233333333334444444444@123456789011>
+ * References: <12345678901111111111222222222233333333334444444444@123456789011>
+ * ```
+ *
+ * The distribution of the characters to left-of-@ and right-of-@ was arbitrary.
+ * The choice was made to put more into the left-id and shorten the right-id to
+ * slightly mimic a common length domain name.
  *
  * @note The caller should free the string
  */
-static char *gen_msgid(struct ConfigSubset *sub)
+static char *gen_msgid()
 {
+  const int ID_LEFT_LEN = 50;
+  const int ID_RIGHT_LEN = 12;
+  char rnd_id_left[ID_LEFT_LEN + 1];
+  char rnd_id_right[ID_RIGHT_LEN + 1];
   char buf[128] = { 0 };
-  char rndid[MUTT_RANDTAG_LEN + 1];
 
-  mutt_rand_base32(rndid, sizeof(rndid) - 1);
-  rndid[MUTT_RANDTAG_LEN] = 0;
-  const char *fqdn = mutt_fqdn(false, sub);
-  if (!fqdn)
-    fqdn = NONULL(ShortHostname);
+  mutt_rand_base32(rnd_id_left, sizeof(rnd_id_left) - 1);
+  mutt_rand_base32(rnd_id_right, sizeof(rnd_id_right) - 1);
+  rnd_id_left[ID_LEFT_LEN] = 0;
+  rnd_id_right[ID_RIGHT_LEN] = 0;
 
-  struct tm tm = mutt_date_gmtime(mutt_date_now());
-  snprintf(buf, sizeof(buf), "<%d%02d%02d%02d%02d%02d.%s@%s>", tm.tm_year + 1900,
-           tm.tm_mon + 1, tm.tm_mday, tm.tm_hour, tm.tm_min, tm.tm_sec, rndid, fqdn);
+  snprintf(buf, sizeof(buf), "<%s@%s>", rnd_id_left, rnd_id_right);
   return mutt_str_dup(buf);
 }
 
@@ -775,7 +792,7 @@ void mutt_prepare_envelope(struct Envelope *env, bool final, struct ConfigSubset
     mutt_set_followup_to(env, sub);
 
     if (!env->message_id)
-      env->message_id = gen_msgid(sub);
+      env->message_id = gen_msgid();
   }
 
   /* Take care of 8-bit => 7-bit conversion. */
@@ -848,7 +865,7 @@ static int bounce_message(FILE *fp, struct Mailbox *m, struct Email *e,
     fprintf(fp_tmp, "Resent-Date: %s\n", mutt_buffer_string(date));
     mutt_buffer_pool_release(&date);
 
-    char *msgid_str = gen_msgid(sub);
+    char *msgid_str = gen_msgid();
     fprintf(fp_tmp, "Resent-Message-ID: %s\n", msgid_str);
     FREE(&msgid_str);
     fputs("Resent-To: ", fp_tmp);


flatcap:
Thanks for having a go at this.
It's been on one of my many todo lists for ages.

The list of commits I had was:

| Commit                                                                                       | Description                                                        |
| :------------------------------------------------------------------------------------------- | :----------------------------------------------------------------- |
| [99ba609f](https://gitlab.com/muttmua/mutt/commit/99ba609f09582469360f3a5fd7132ee556cbaeaf)  | Loosen message-id parser to parse ids without angle brackets       |
| [194a60ae](https://gitlab.com/muttmua/mutt/commit/194a60ae0ec6f0f09d5e40023894363288df532e)  | Adjust message-id extraction to retry more permissively on failure |
| [e3e0a0ba](https://gitlab.com/muttmua/mutt/commit/e3e0a0ba0d1fbe242111eae04da6084a821d5aca)  | Change mailto handling to parse in-reply-to loosely                |
| [9da4e6e1](https://gitlab.com/muttmua/mutt/commit/9da4e6e11e7037668d0ca7e8f5d6773d26e379ac)  | Change Message-ID to be more unique and leak less information      |
| [ee2b9d1d](https://gitlab.com/muttmua/mutt/commit/ee2b9d1dd8963a7b5b469ddfd8609f9a40bd5157)  | Remove message-id security leaks section of manual                 |
| [b5a807ca](https://gitlab.com/muttmua/mutt/commit/b5a807cadb499d3078aedbdb67e4c22fc1520f88)  | Move mutt_gen_msgid() over to messageid.c                          |
| [1f188287](https://gitlab.com/muttmua/mutt/commit/1f18828754ed781f2562a60414521a5e1e6b93a8)  | Add $message_id_format for customizing generated Message-IDs       |
| [5f5d33f8](https://gitlab.com/muttmua/mutt/commit/5f5d33f832c82f14280e2c643b2039d4beb68573)  | Change $message_id_format year expando from %y to %Y               |
| [dadc0130](https://gitlab.com/muttmua/mutt/commit/dadc0130fc83ae31f9142f238e60c0921bf18da2)  | Provide %x format for $message_id_format                           |

I see that the bottom 5 overlap with yours.
I'll check if the others are (still) relevant.

flatcap:
### History

Long ago, the Mutt and NeoMutt code bases were fairly closely aligned.
Back then, I merged all the upstream commits.
As we pushed forward with NeoMutt this grew harder and harder.

I'd still like to merge them all, but my focus has been on fixing the major architectural problems (still a work in progress).

I have a repo that contains my todo list and all the outstanding patches,
- https://github.com/neomutt/upstream-patches

but unfortunately, you can't see that because it's private and you haven't accepted my invitation ;-)

Anything in that list is fair game.

gahr:
what’s the use case for wanting to customize message id to these extremes? And where do we stop? 

Most expandos in neomutt are context-sensitive: the date the *email* was sent, *its* sender, a *mailbox*’ unread count.. on the other hand, here, we’re poviding expandos for general date formatting and randomness. The only context-sensitive piece is the pid, but I fail to see the value.

Moreover, as you noted, users must ensure the final string respects an rfc.

if we really think fully customizable message-ids add value to neomutt, the I would suggest to just provide the pipe thingie and let users compose their own scripts.

rayfordshire:
> what’s the use case for wanting to customize message id to these extremes?

This feature has been requested often here [#3060] and over at mutt's mailing list.  My personal use case is to get rid of the date encoded in the message id.

> And where do we stop?

When we reached world domination! ;-)

> Moreover, as you noted, users must ensure the final string respects an rfc.

Sorry, but you can already shoot yourself in the foot by using `my_hdr`, so that's not an argument here.

Obviously, people using this feature should understand it because it is – like `my_hdr` and many other of NeoMutt's features (`$wrap_headers`, ...) – an advanced feature.

The default will work for most people, those who need it can temper with the config variable at their own risk.

> if we really think fully customizable message-ids add value to neomutt, the I would suggest to just provide the pipe thingie and let users compose their own scripts.
> edit: see account_cmd for a precedent of the approach I suggest :)

It looks like this would cover my use case.

---

The question now is:

If Mutt has a config variable like in this case, should Neomutt adapt it?

gahr:
> This feature has been requested often here [#3060] and over at mutt's mailing list.

Where you say "often", I head "once, and the reason was that upstream had it" . Admittedly, I haven't checked the discussion that have happened on the Mutt mailing list, but my intuition would be that people want to disclose as little as possible about their systems in the message-id. Your use case matches my intuition :)

> > Moreover, as you noted, users must ensure the final string respects an rfc.
> 
> Sorry, but you can already shoot yourself in the foot by using `my_hdr`, so that's not an argument here.

Honestly, I don't know. Yes, we do provide ways to customize things to the extremes. But is that a reason for adding complexity on top of complexity?

I think message-id is as implementation-detail-ish as it gets. Perhaps all we should do is stop encoding system information in it, just use uuid, and stash an "@" sign in the middle.

gahr:
> The question now is:
> 
> If Mutt has a config variable like in this case, should Neomutt adapt it?

My answer would be a strong "no", and my main reason is complexity: there are a limited amount of "us" around here (mostly 1), and we're already maintaining things nobody of us uses (mixmaster, lua, and autocrypt are three examples that come to mind). I think we should aim at having something we "own", in terms of knowledge. Adding features is easy, maintaining them and answering to issues and bug requests about them, less so.

I'm sure @flatcap 's take would be more nuanced.

rayfordshire:
> I think message-id is as implementation-detail-ish as it gets. Perhaps all we should do is stop encoding system information in it, just use uuid, and stash an "@" sign in the middle.

Incidentally, I wanted to provide a patch which just did that (because that's similar to how Thunderbird does it).  But then I noticed that Mutt had the option `$message_id_format` and though that might have a better chance of getting merged ... I guess was wrong.


Before I go and provide another patch would it be acceptable to replace the message-id generation by `<random-string@$hostname>`, where `random-string` is either of

a) a random sequence of allowed chars in message-id (length to be discussed)
b) a (valid) UUID version 4 [1, section 4.4], i.e. a random UUID.

Note: Thunderbird does a) but prints it like an UUID confusing some people[2]

[1] https://datatracker.ietf.org/doc/html/rfc4122
[2] https://bugzilla.mozilla.org/show_bug.cgi?id=1532424

gahr:
> Before I go and provide another patch would it be acceptable to replace the message-id generation by `<random-string@$hostname>`

I think I would go further and use a random string after the @ sign, too, but yeah, I would like your approach too. I don’t have a preference for how you get the random bytes, but I don’t think we can generate uuid’s without some additional dependency.

rayfordshire:
> > Before I go and provide another patch would it be acceptable to replace the message-id generation by `<random-string@$hostname>`
> 
> I think I would go further and use a random string after the @ sign, too, but yeah, I would like your approach too.

Since you mentioned it, I also like it better if we also randomize the hostname part.

> I don’t have a preference for how you get the random bytes, but I don’t think we can generate uuid’s without some additional dependency.

Unless I'm mistaken that shouldn't be to hard: generate 16 random octets/bytes, then set the six bits required to be a valid UUID v4, then convert the octets with base16 into two characters, finally concatenate everything. (UUID v4 does not have any checksum mechanism).

That said, this is still a little more complex than just using a random string (which might also be more flexible).

Any preference on the entropy of the message id? Or should I come up with a number.

gahr:
> Any preference on the entropy of the message id? Or should I come up with a number.

My uninformed take: make the output as long as possible but not too long so it doesn’t require wrapping the headers that can contain it (message-id, references, others?) :)

rayfordshire:
> > Any preference on the entropy of the message id? Or should I come up with a number.
> 
> My uninformed take: make the output as long as possible but not too long so it doesn’t require wrapping the headers that can contain it (message-id, references, others?) :)

That was also my approach.  See what I came up with. Please read the commit log for the astronomical odds of winning in the lottery.

I also looked at other Message-IDs, for example Gmail uses a Message-ID length of 64 characters 51 of which seems random, yielding in an total of 80 character for the Message-ID header (without CRLF).

gahr:
LGTM - can you please adjust the PR title to reflect what it currently does?

gahr (approved):

flatcap:
I just moved the comment into the Doxygen header so it shows up in the code docs.

flatcap:
Courtesy of bookworm on IRC (@brunnre8): [Fix your mutt](https://people.kernel.org/monsieuricon/fix-your-mutt)

cyphar:
Sorry to necro-post but making neomutt generate a completely random Message-ID is quite annoying as a kernel developer because we often reference emails by their Message-ID in commit messages and in mailing list discussions. Having a Message-ID of the form `%Y%M%d-[random crap]@domain` means that you can at least see some basic information in the Message-ID without having to actually go to the source email to find out what the hell `<u7lakye7ikvyu6g2ktxbzixt5hnvqtzt5s4g72j74tgg4bwlpu@7pcqd4ah5tah>` is.

Maybe this doesn't matter to most users, but having the ability to customise it would mean that we can continue having a more informative Message-ID while other users can have a completely obfuscated one.

I do wonder whether spam filters would have an issue with emails that have a domain section of a Message-ID that doesn't match the From header. [Someone mentioned this does actually penalise emails in the other thread asking for the `message_id_format` option.](https://github.com/neomutt/neomutt/issues/3060#issuecomment-1506583183)

I can send a patch to add `message_id_format` support, but given that this PR was originally that and then it morphed into this version through review, it seems that this is something the NeoMutt maintainers don't want -- in which case I guess I'll have to switch back to Mutt...

yourealwaysbe:
Posting my solution to including hostname in case it's useful. I needed it to be dynamic (switching between multiple accounts means i have folder-hooks to change $hostname) to avoid spam filter issues at work.

    send-hook . 'my_hdr Message-ID: <`uuidgen -r`@$hostname>'

The single quotes are important else $hostname is expanded only once when the configuration is loaded.

vt-alt:
While a random right part is allowed, using a domain is **recommended** in [rfc5322](https://www.rfc-editor.org/rfc/rfc5322#section-3.6.4):
```
   The message identifier (msg-id) syntax is a limited version of the
   addr-spec construct enclosed in the angle bracket characters, "<" and
   ">".  Unlike addr-spec, this syntax only permits the dot-atom-text
   form on the left-hand side of the "@" and does not have internal CFWS
   anywhere in the message identifier.

      Note: As with addr-spec, a liberal syntax is given for the right-
      hand side of the "@" in a msg-id.  However, later in this section,
      the use of a domain for the right-hand side of the "@" is
      RECOMMENDED.

   [...] Though other algorithms will work, it is RECOMMENDED
   that the right-hand side contain some domain identifier (either of
   the host itself or otherwise) such that the generator of the message
   identifier can guarantee the uniqueness of the left-hand side within
   the scope of that domain.
```

Issue #3654, opened 2023-02-01
Commands `subscribe-to` and `unsubscribe-from` not documented


whitney-cumber:
The commands `subscribe-to` and `unsubscribe-from` are not documented anywhere.

Possible solutions I can think of:
* Write the documentation yourselves
* Tell me what they do and I draft something. Note: I don't use IMAP.

flatcap:
In NeoMutt we configure which mailboxes are "interesting" to us with the `mailboxes` command.
These mailboxes are checked for new mail and will appear in the sidebar.

Two of the mailbox backends, IMAP and NNTP, keep their own lists of "interesting" mailboxes.

We can add to / remove from these lists using functions: `<subscribe>` / `<unsubscribe>` in the mailbox browser,
or using the commands: `:subscribe-to` / `unsubscribe-from`.

---

When NeoMutt connects to an IMAP server, it checks `$imap_check_subscribed`.
If set, then NeoMutt asks the server for its list and internally runs the `mailboxes` command on each of them.

Now, NeoMutt will query the server for all the "interesting" mailboxes (whether manually configured, or subscribed to).

When using the mailbox browser, `$imap_list_subscribed` determines whether just the **subscribed** mailboxes are displayed, or **all** mailboxes. 

---

NNTP servers typically have hundreds of newsgroups.
When NeoMutt first connects to an NNTP server, it gets a list of all the newsgroups.
The user then subscribes to those of interest and the list gets stored locally in a `.newsrc` file.

Users can also specify newsgroups directly, using the `mailboxes` command.

Now, NeoMutt will query the server for all the "interesting" newsgroups (whether manually configured, or subscribed to).

PR #3658, opened 2023-02-03
Document the subscribe-to and unsubscribe-from commands

whitney-cumber:
* **What does this PR do?**

* Document the subscribe-to and unsubscribe-from commands for IMAP
* Document the subscription feature of IMAP and how NeoMutt uses it
* Document that NeoMutt can speak NNTP / interact with newsgroups

Closes: #3654


flatcap (approved):
Lovely job!  Thank you very much!

PR #3666, opened 2023-02-04
Fix bug: hdr_order did not sort last header correctly

rayfordshire:
The algorithm to sort headers by the order given in hdr_order sorted the last mentioned header in hdr_order at the same level as unmentioned headers, i.e. not at all.

The bug was introduced in d9ff90dced82e7328dff4abc0918ee0020af8d0a.

The algorithm puts headers in buckets labeled by their priority (0 = highest, n = lowest).

The problem is that the "priority counter" (which starts at 0) `x` is first incremented and then the header is assigned to priority `x`. Thus, the buckets 0 to n become 1 to n+1 and the header mentioned last in hdr_order (number n) was sorted in the bucket n+1 together with all other headers.

The fix is simple: First sort the header in the bucket, then increase the priority counter.

### DIFF
--- a/copy.c
+++ b/copy.c
@@ -310,7 +310,6 @@ int mutt_copy_hdr(FILE *fp_in, FILE *fp_out, LOFF_T off_start, LOFF_T off_end,
 
         STAILQ_FOREACH(np, &HeaderOrderList, entries)
         {
-          x++;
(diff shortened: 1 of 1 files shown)


flatcap (approved):
Good spot!
Thanks!

PR #3673, opened 2023-02-06
Make exiting via SIGINT more graceful

rayfordshire:
* **What does this PR do?**

1) Clarify to the user that exiting via SIGINT does *not* save. Previously, the question was phrased the same as the for the `quit` function (which dose save the mailbox), now it is the one from the `exit` function (which does not save).
(No new translation entry is needed)

2) Exit with exit code 0.  The user was asked, they replied, we acted, so this should be considered a normal run, not one with a special exit condition (i.e. != 0)

---

When comparing the SIGINT exit routine with the `exit` function one, we see differences (see below).
This leads to the question, whether the SIGINT code path should do a little more than calling `exit(0)` directly (maybe closing the mailbox)?

https://github.com/neomutt/neomutt/blob/ff76d02c585b81daae56d03e86222621dbd86abc/gui/curs_lib.c#L330-L349

https://github.com/neomutt/neomutt/blob/ff76d02c585b81daae56d03e86222621dbd86abc/index/functions.c#L519-L539

### DIFF
--- a/gui/curs_lib.c
+++ b/gui/curs_lib.c
@@ -339,7 +339,7 @@ void mutt_query_exit(void)
   const short c_timeout = cs_subset_number(NeoMutt->sub, "timeout");
   if (c_timeout)
     set_timeout(-1); /* restore blocking operation */
-  if (mutt_yesorno(_("Exit NeoMutt?"), MUTT_YES) == MUTT_YES)
+  if (mutt_yesorno(_("Exit NeoMutt without saving?"), MUTT_YES) == MUTT_YES)
   {
     mutt_exit(1);
   }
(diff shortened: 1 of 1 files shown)


flatcap (approved):
Oh!  That's a good idea!

> No new translation entry is needed

Appreciated :-)

flatcap:
@rayfordshire 
@whitney-cumber 
@Pierre-Colin 

May I ask a favour...

Please could you set your profile picture: https://github.com/settings/profile

I "meet" lots of people on GitHub, but sometimes it's hard to remember who's who.
Having an avatar really helps: something distinctive, something creative, it doesn't matter what.

Cheers, @flatcap

Issue #3645, opened 2023-01-28
Allow specifying fall back header cache backends/compress algorithms


whitney-cumber:
## What would you like to be able to do?

I'd like to give a "priority order" for the header cache backends.  Use case is to have the same configuration file for neomutt even when different compile-time features are present, namely header cache backends.

The current `$header_cache_backend` allows only for one backend to be specified (or to use Neomutt's builtin priority list).  If that backend is not compiled in, then the config fails.

A similar problem arises with `$header_cache_compress_method`.

## What needs to change in NeoMutt?

Two solutions pop into my mind:

1) make `$header_cache_backend`/ `$header_cache_compress_method` a list of strings instead of strings (no clue what a good separator char would be)
2) export the compiled in header cache backends / compressions as compile options

In the later case the priority list can then be achieved by
```
ifdef tdb set header_cache_backend = tdb    # fall back if everything else fails
ifdef qdbm set header_cache_backend = qdbm  # silver medal goes here
ifdef ldbm set header_cache_backend = ldbm  # top choice
```

flatcap:
> `ifdef ldbm set header_cache_backend = ldbm  # top choice`

Hmm... I don't think that'll work.
We could add the stores' names to the existing ifdef tests:
https://github.com/neomutt/neomutt/blob/5fcc6c9e9f606f47291066ad1022fe805545640a/command_parse.c#L504-L509

My first thought was: this might be quite a bit of work.  But, if you're offering :-)

Next, I thought this ought not to be too difficult.
We turns the config options into `DT_SLIST`; we use the validator to check them, then tweak the hcache to select one.
(The config string type supports different separators)
https://neomutt.org/code/config_slist.html

But, if a store isn't compiled-in, then the name simply doesn't exist in the code.
We wouldn't be able to distinguish compiled-out from rubbish:
```
set header_cache_backend = rubbish,lmdb
```
Perhaps that doesn't matter.

Another possibility is to reorder `StoreOps[]` to put the modern databases first, e.g. lmdb, tdb, rocksdb:
https://github.com/neomutt/neomutt/blob/5fcc6c9e9f606f47291066ad1022fe805545640a/store/store.c#L50-L53

whitney-cumber:
> My first thought was: this might be quite a bit of work. But, if you're offering :-)

Sorry but I've never learned a C style language.  So that bit of work might be quite to much for me. :-(

Shouldn't it be enough to include something like
```
ifdef HAVE_LMDB
[...]

PR #3678, opened 2023-02-08
store: promote modern databases

flatcap:
Modern databases, like lmdb, rocksdb and tdb outperform the old default of tokyocabinet.

See: #3645 

---

If `$header_cache_backend` isn't set, then the first compiled-in database is used.

The new priority order is:
- lmdb, rocksdb, tdb, tokyocabinet, kyotocabinet, bdb, gdbm, qdbm

This change will naturally invalidate any existing caches,
but due to our frequent refactoring, almost every release does this anyway.

---

If this change is accepted, we should urge the downstream packagers to compile-in at least one of the modern databases.

### DIFF
--- a/hcache/hcachever.sh
+++ b/hcache/hcachever.sh
@@ -1,6 +1,6 @@
 #!/bin/sh
 
-BASEVERSION=7
+BASEVERSION=8
 STRUCTURES="Address Body Buffer Email Envelope ListNode Parameter"
 
 cleanstruct () {
--- a/store/store.c
+++ b/store/store.c
@@ -48,29 +48,29 @@ STORE_BACKEND(tokyocabinet)
  * StoreOps - Backend implementations
  */
 static const struct StoreOps *StoreOps[] = {
+#ifdef HAVE_LMDB
+  &store_lmdb_ops,
+#endif
+#ifdef HAVE_ROCKSDB
+  &store_rocksdb_ops,
+#endif
+#ifdef HAVE_TDB
+  &store_tdb_ops,
+#endif
 #ifdef HAVE_TC
   &store_tokyocabinet_ops,
 #endif
 #ifdef HAVE_KC
   &store_kyotocabinet_ops,
 #endif
-#ifdef HAVE_QDBM
-  &store_qdbm_ops,
-#endif
-#ifdef HAVE_ROCKSDB
-  &store_rocksdb_ops,
+#ifdef HAVE_BDB
+  &store_bdb_ops,
 #endif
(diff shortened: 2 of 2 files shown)


flatcap:
> Or we’ll be trying to read a tokyocabinet file with lmdb?

(I'll check, but) that shouldn't be possible.
We merge the database name into the filename-hash generation,
so for a given path, each database should create a unique file.

<sub>hopefully :-)</sub>

gahr:
> We merge the database name into the filename-hash generation,
> so for a given path, each database should create a unique file.

This is only true if you hit [this branch](https://github.com/neomutt/neomutt/blob/77c3a73202d7d3460661d2d2d1076acddb7f9691/hcache/hcache.c#L286-L306), i.e., your `header_cache` doesn't point to a file and the backend doesn't have its own namer (imap does).

flatcap:
Oh yeah, the `namer` again.

OK, I'll bump this to be sure:
https://github.com/neomutt/neomutt/blob/77c3a73202d7d3460661d2d2d1076acddb7f9691/hcache/hcachever.sh#L3

whitney-cumber:
I just read up on strength/weakness of different backends and I stumbled upon this entry in the documentation of lmdb[0]:

> [...]
> * Do not use LMDB databases on remote filesystems, even between processes on the same host. This breaks flock() on some OSes, possibly memory map sync, and certainly sync between programs on different hosts.
>
> [...]

I know many companies who mount their home directory via NFS from a server, so this might be a deal breaker?

[0] http://www.lmdb.tech/doc/

flatcap:
I'm not an expert, but...

> so this might be a deal breaker

Possibly...

One of LMDB's reasons for being fast is that it `mmap()`s the database file, rather than making lots of small file reads.
I think the idea is to start a transaction, fetch a record and then use pointers into that memory.
While the transaction is alive, that memory is guaranteed.
When you're done with the record, you end the transaction and the mapped memory is unlocked.
I can see how this might be risky on a network drive.

However, that's not how NeoMutt is using LMDB.
NeoMutt starts transaction, copies the data and closes the transaction immediately.
In practice, I think the risk is pretty low.

flatcap:
@gahr and I discussed this quite a bit on IRC.
There's a potential problem with NeoMutt trying to open a file of one database type with the code of another.

I'd really like to nudge the users towards modern features, but doing so without any potential breakage is tricky.

We don't have a way to communicate with our users.
If we had some internal `:news` feature, we could give them tips.

Anyway, this was just a quick hack for discussion.
A better solution is needed.

PR #3697, opened 2023-02-12
Provide expando to show the fingerprint of a key

rayfordshire:
* **What does this PR do?**

1) The first commit fixes a legacy/mixup of $smime_default_key, I've stumpled while coding the rest.  Since you flatten your history anyway, I refrained from providing a separate PR.

2) Add a new expando, `%i` to `$pgp_entry_format` which expands to the full fingerprint of a key (or its long id, if there is no fingerprint).

3) Discussion worthy:  change the meaning of `%k` and `%i` expandos.  This naming aligns better with the actual names `i` for id and `k` for `key fingerprint`.  Obviously, this is a user visible change, its strictly providing more information as the keyid is a substring of the fingerprint.  I guess almost nobody will notice.  To mitigate the change even further we could change the default `$pgp_entry_format` to use  `%i` instead of `%k`, so only those who changed pgp_entry_format will notice a difference.

*Why this change*: It is well known that short key ids are broken.  Long key ids (double the digits) are a little better but still do not provide the guarantee of no collision. Therefore, the key fingerprint is the only safe way to guarantee that we are dealing with the key we think we are dealing.

### DIFF
--- a/docs/config.c
+++ b/docs/config.c
@@ -3113,6 +3113,7 @@
 ** .dt %a     .dd Algorithm
 ** .dt %c     .dd Capabilities
 ** .dt %f     .dd Flags
+** .dt %i     .dd Key fingerprint (or long key id if non-existent)
 ** .dt %k     .dd Key id
 ** .dt %l     .dd Key length
 ** .dt %n     .dd Number
--- a/ncrypt/crypt_gpgme.c
+++ b/ncrypt/crypt_gpgme.c
@@ -3836,7 +3836,7 @@ static SecurityFlags gpgme_send_menu(struct Email *e, bool is_smime)
           snprintf(input_signas, sizeof(input_signas), "0x%s", crypt_fpr_or_lkeyid(p));
 
           if (is_smime)
-            cs_subset_str_string_set(NeoMutt->sub, "smime_default_key", input_signas, NULL);
+            cs_subset_str_string_set(NeoMutt->sub, "smime_sign_as", input_signas, NULL);
           else
             cs_subset_str_string_set(NeoMutt->sub, "pgp_sign_as", input_signas, NULL);
 
--- a/ncrypt/dlg_gpgme.c
+++ b/ncrypt/dlg_gpgme.c
@@ -349,12 +349,14 @@ static char crypt_flags(KeyFlags flags)
  * | \%a     | Algorithm
  * | \%c     | Capabilities
  * | \%f     | Flags
+ * | \%i     | Key fingerprint (or long key id if non-existent)
  * | \%k     | Key id
  * | \%l     | Length
  * |         |
  * | \%A     | Algorithm of the principal key
  * | \%C     | Capabilities of the principal key
  * | \%F     | Flags of the principal key
+ * | \%I     | Key fingerprint of the principal key (or long key id if non-existent)
  * | \%K     | Key id of the principal key
  * | \%L     | Length of the principal key
  */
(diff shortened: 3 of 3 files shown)


docs/config.c
```diff
@@ -3113,7 +3113,8 @@
 ** .dt %a     .dd Algorithm
 ** .dt %c     .dd Capabilities
 ** .dt %f     .dd Flags
-** .dt %k     .dd Key id
+** .dt %i     .dd Key id
+** .dt %k     .dd Key fingerprint (or long key id if non-existent)
```
gahr, line 3117: I would not change existing expandos

gahr:
How would you feel about keeping `k` as-is and adding `K`? At the end of the day, aren't you adding a "bigger" version of it? :)

rayfordshire:
> An alternative, might be to use a two-letter expando, %kf (key-fingerprint)?

But how would you print then the key id followed by a literal `f`, e.g. `0x12345f` you cant use `%` to escape `f`?

flatcap:
> But how would you print then the key id followed by a literal `f`, e.g. `0x12345f` (you can't use `%` to escape `f`)?

You wouldn't be able to.
Fortunately, this situation probably isn't that likely.
(the same with the index expandos)

rayfordshire:
> >  But how would you print then the key id followed by a literal f, e.g. 0x12345f (you can't use % to escape f)?
>
>You wouldn't be able to.
>Fortunately, this situation probably isn't that likely.
>(the same with the index expandos)

Just because it's not likely, should we actively irreversible block this possibility?
The expandos for index are now history, so that ship has sailed but here we talk about a new one in a different menu which does not yet have this flaw.

---

I can see why you feel uncomfortable with the swap of `%i` and `%k`, so feel free to omit that commit (that's way it is a separate one).  If you don't like the choice of `%i` feel free to pick another letter.

flatcap:
> should we actively irreversible block this possibility?

No :-)

> If you don't like the choice of `%i` feel free to pick another letter.

How about `%p` for finger**p**rint?

---

In the long-term (I have a lot of plans in this category), it'd be nice to eliminate the hard-to-remember single letters.
One possibility is the form `%{fingerprint}`, discussed briefly here:

https://github.com/neomutt/flatcap-brain-dump/blob/main/split-expando.txt#L100-L114

It'd be quite a bit of work to implement, though.

flatcap:
# Welcome to NeoMutt

Welcome aboard, @rayfordshire!
I don't have an email address for you, so here's my usual [Welcome Letter](https://gist.github.com/flatcap/b589998642e8a9b494e0a15af045ad29).

rayfordshire:
>> If you don't like the choice of %i feel free to pick another letter.
>
> How about `%p` for fingerprint?

Sorry, already taken by protocol.

flatcap:
> Sorry, already taken by protocol.

hehe, OK.
Pick at a letter that isn't in use and we'll go with that.
My opinion doesn't matter :-)

rayfordshire:
> Pick at a letter that isn't in use and we'll go with that.

I already did, it's `%i`

You have just to decide whether you this with or without swapping `%i` and `%k`, which would result in slightly longer "ids" by default.

flatcap:
> with or without swapping `%i` and `%k`

Without swapping, please.

We'll do our best to inform the users, when we make a release.

rayfordshire:
dropped the last commit.

flatcap (approved):
That's great.  Thanks.
Sorry for all the to-and-fro.

PR #3698, opened 2023-02-12
Fix printf-format string in debug output

rayfordshire:
The format string used 4 arguments, but only 3 were given.

I made a blind guess what could have been meant.  If that's not it, then please help yourself.


gahr (approved):

flatcap (approved):
Good spot, thanks!
Good job they're turned off by default.

PR #3702, opened 2023-02-15
Rename connect_timeout to socket_timeout, use it for read/write ops too

gahr:
Instead of adding dedicated timeout options for read/write, like upstream did, we're reusing the connect_timeout and renaming it to socket_timeout. At this point, we don't see a reason for wanting to specify different values for the connect/read/write operations.

Upstream-commit: https://gitlab.com/muttmua/mutt/-/commit/185346ad778f85f7889cff662516e2121bb1edbb

### DIFF
--- a/conn/config.c
+++ b/conn/config.c
@@ -36,9 +36,10 @@ static struct ConfigDef ConnVars[] = {
   { "account_command", DT_STRING|DT_COMMAND, 0, 0, NULL,
     "Shell command to retrieve account credentials"
   },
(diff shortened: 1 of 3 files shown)


flatcap:
I've just added 76d86bc60 - optional idea.
We don't need negative values, 0 will do

gahr:
Doesn’t this break existing configs that specify -1?

flatcap:
Urgh!  Yes.  Dropped.

flatcap (approved):

Issue #3690, opened 2023-02-12
Pattern `~A` does not match all messages


whitney-cumber:
## Expected Behaviour

The pattern `~A` should match all messages as per documentation.

## Actual Behaviour

`~A` only matches the root messages of threads

## Steps to Reproduce

0. Find a mailbox with messages building a thread, e.g. these two

```
Date: Sun, 12 Feb 2023 12:00:00 +0000
From: foo@example.org
Subject: Foo
Message-ID: <foo@example.org>


Foo
```

```
Date: Sun, 12 Feb 2023 12:00:00 +0000
From: bar@example.org
Subject: Bar
Message-ID: <bar@example.org>
References: <foo@example.org>

Bar
```

1. Enable threading, e.g. with this `neomuttrc-all-pattern`

```
set use_threads = threads
set collapse_all = yes
```

2. Open NeoMutt: `neomutt -F neomuttrc-all-pattern -f mailbox`
3. The threads should be collapsed.  In this example we've done this with `set collapse_all` but can also done manually.
4. Tag all messages: `<tag-pattern>~A<enter>` (usually bound to `T~A`)
5. Notice that only the "Foo" message (thread root) is tagged.  This can be verified either by uncollapsing the threads (`<collapse-all>` bound to `<esc>V` or by looking at the status line which reads "Tag:1".

## NeoMutt Version

```
20220429-371-e26e9d
```

gahr:
It's not about the pattern, it's about what you do with it: limiting works on all messages, while the other functionalities where you can use a pattern (tagging, untagging, deleting) work on the visible messages only.

This is deliberate, i.e., there are explicitly [two branches](https://github.com/neomutt/neomutt/blob/0344ca405f177fdae983fa6cea178d1553dd0146/pattern/pattern.c#L393) in the code for that, so while this might look surprising at first sight, I think this is a behaviour most long-time users are expecting or at least used to.

whitney-cumber:
Ok, can accept that explanation and I see why that behaviour might be useful in the limiting case.

What would the pattern be to tag all messages in a mailbox like in the above mentioned case? `~(~A)`, `~>(~A)`, `~<(~A)` don't do the trick either.

gahr:
I have a macro to delete all messages, I use it on my spam folder. What it does is it uncollapse-all first, then tag all messages.

whitney-cumber:
Where would be a good place to document this discrepancy between the meaning of patterns in

* index with non-limit operation
[...]

PR #3703, opened 2023-02-15
Document that some pattern operations only apply to visible messages

whitney-cumber:
The patterns given to the functions `<delete-pattern>`, `<undelete-pattern>`, `<tag-pattern>`, and `<untag-pattern>` only match messages which are not hidden, i.e. shown by the limit and not collapsed inside a thread.  This is in contrast to the pattern matched by the `<limit>` function which applies to all messages in the mailbox hidden or not.

Note: This might invalidate a couple of translations.

Closes: #3690


flatcap (approved):
Looks good

flatcap:
> This might invalidate a couple of translations.

Yes, I'll let the translators know.

PR #3709, opened 2023-02-16
send: delete signature when sending fails

### LINKED ISSUE #3705: Corrupted signatures after sending failed
## The Problem

For signed mails, NeoMutt adds a broken signature each time sending fails.

## Steps to Reproduce

0. Need a S/MIME certificate (PGP key might work, too)
1. Use this `neomuttrc`
```
set crypt_use_gpgme = yes
set sendmail = "/bin/false"
```
Enable GPGME backend (classic might work, too).
The sendmail config makes sure sending fails.

2. Send a signed mail to someone
```
~/tmp/neomutt/build_plain/neomutt -F neomuttrc-smime -s 'Signing issue' foo@example.org
```
3. Select S/MIME singing and a valid certificate.
4. Press send (`y`)
5. Sending should have failed
6. Press send (`y`) again
7. Press send (`y`) again
8. Press send (`y`) again
9. ...
10. Notice that each time an additional multipart-signed layer is added to the mail.
11. Set `sendmail` (or `smtp_url`) to a valid command/server
12. Send the mail
13. The receving site now has multiple signed layers, all but one which are broken.  NeoMutt prints for these broken parts
```
[-- Error: Missing or bad-format multipart/signed signature --]
```

## Expected Behaviour

NeoMutt should either

a) not generate additional signatures for successive send attempts (and the first on should be valid)
b) properly remove the signature after a broken send attempt

dcpurton:
When sending fails after a message has been signed (but not encrypted), then the `multipart/signed` container is not removed, leaving a corrupt `multipart/signed` message without the actual signature. Subsequent attempts to send the message add additional corrupt `multipart/signed` containers.

This PR removes the `multipart/signed` container in this case.

Fixes #3705

### DIFF
--- a/send/send.c
+++ b/send/send.c
@@ -2892,7 +2892,9 @@ int mutt_send_message(SendFlags flags, struct Email *e_templ, const char *tempfi
         mutt_body_free(&e_templ->body->parts->next); /* destroy sig */
         if (mutt_istr_equal(e_templ->body->subtype, "mixed") ||
             mutt_istr_equal(e_templ->body->subtype, "signed"))
+        {
           e_templ->body = mutt_remove_multipart(e_templ->body);
+        }
       }
 
       FREE(&pgpkeylist);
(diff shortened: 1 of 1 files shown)


gahr (approved):
I trust you :)

dcpurton:
Actually, I wonder if `multipart/mixed` ever actually occurs? Doesn't Neomutt always produce a `multipart/signed` email for signed emails?

flatcap:
> I wonder if `multipart/mixed` ever actually occurs?

Don't know, so let's leave it for now :-)

Issue #3691, opened 2023-02-12
Yes/no prompt input doesn't work in some locales


Strahinja:
## Expected Behaviour

Pressing the keys `y` or `n` in yes/no prompts, for example `Print message? ([no]/yes):`, or their localized variants, for example `д` and `н` in Serbian and Russian locales, results in taking the appropriate action.

## Actual Behaviour

Instead of the desired actions (in the above example: `y` or `д` leading to printing message or `n` or `н` leading to canceling the prompt), nothing happens: the prompt remains displayed.

## Steps to Reproduce

1. Have `LC_ALL` set to relevant locale ending with `.UTF-8`, for example
```
export LC_ALL=sr_RS.UTF-8
```
or
```
export LC_ALL=ru_RU.UTF-8
```
2. Start Neomutt.
3. Go to a message in the list of messages and press `p`.
4. Try pressing `y`, `n`, `д` or `н`.

If Neomutt is started with
```
LC_ALL=sr_RS neomutt
```
or
```
LC_ALL=ru_RU neomutt
```
(note the lack of `.UTF-8`), the `y` and `n` keys work as expected, while the localized keys `д` and `н` still don't work.

## How often does this happen?

- Always

## When did it start to happen?

- When I switched to Alpine Linux (musl + musl-locales) from Artix Linux (glibc), latest versions of all of the relevant programs.

## NeoMutt Version

```
NeoMutt 20220429-380-0344ca-dirty
Copyright (C) 1996-2022 Michael R. Elkins and others.
NeoMutt comes with ABSOLUTELY NO WARRANTY; for details type 'neomutt -vv'.
NeoMutt is free software, and you are welcome to redistribute it
under certain conditions; type 'neomutt -vv' for details.

System: Linux 5.15.93-0-lts (x86_64)
ncurses: ncurses 6.3.20221119 (compiled with 6.3.20221119)
libidn2: 2.3.4 (compiled with 2.3.4)
GPGME: 1.18.0
GnuTLS: 3.7.8
libnotmuch: 5.6.0
storage: kyotocabinet, gdbm, lmdb
compression: lz4, zlib, zstd

Configure options: --with-notmuch=/usr --with-gss=/usr --sasl --gpgme --sqlite --notmuch --gss --gnutls --sasl --disable-idn --idn2 --lmdb --kyotocabinet --gdbm --lz4 --zlib --zstd

[...]

PR #3712, opened 2023-02-17
question/question.c: Add UTF-8 support to mutt_yesno

Strahinja:
* **What does this PR do?**
_This PR adds UTF-8 support to `mutt_yesno`._

_Hopefully, feel free to suggest corrections or ways to improve the PR._

     for that)

#3691


question/question.c
```diff
@@ -185,6 +185,26 @@ int mutt_multi_choice(const char *prompt, const char *letters)
   return choice;
 }
 
+const int
```
flatcap, line 188: Please can you add a Doxygen header:
```c
/**
 * utf_length_table - DESCRIPTION
 */
```

question/question.c
```diff
@@ -185,6 +185,26 @@ int mutt_multi_choice(const char *prompt, const char *letters)
...
+	1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 
+	1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1,		/* 160 */
+	1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 
+	1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1,		/* 192 */ 
+	2, 2, 2, 2, 2, 2, 2, 2, 2, 2, 2, 2, 2, 2, 2, 2, 
+	2, 2, 2, 2, 2, 2, 2, 2, 2, 2, 2, 2, 2, 2, 2, 2,		/* 224 */
+	3, 3, 3, 3, 3, 3, 3, 3, 3, 3, 3, 3, 3, 3, 3, 3, 
+	4, 4, 4, 4, 4, 4, 4, 4, 5, 5, 5, 5, 6, 6, 1, 1,		/* 256 */ 
+};
+
```
flatcap, line 207: Even though this is only used once, I suggest we:
- move this into `mutt/mbyte.c`
- make it static
- create a helper function to get the value

We use [`clang-format`](https://neomutt.org/dev/analysis/clang-format) on the code.
It's not kind to large tables, so please add these around the data:
```
// clang-format off
...
// clang-format on
```

Also, could we make the table `char[]` -- that would reduce the size by a factor of 4!

flatcap (changes requested):

question/question.c
Strahinja: Looking at `mutt/mbyte.c`, I think that the purpose of the table is equivalent to one of the purposes of the function `mbrtowc` used in `mutt_mb_charlen` to get the value `k` which is returned by `mutt_mb_charlen`, with the difference that only the first character is needed to determine the length.

Strahinja: Based on the above however, I can add a separate function `mutt_mb_charlen_first` (or similarly named...), which would be used for this specific purpose.

question/question.c
Strahinja: Added.

question/question.c
flatcap: Thanks for adding the function.
Yes, it's a bit of overkill, but it's good encapsulation.

flatcap (commented):
Lovely job!
Looks good and it works!

Now, how good's your git-fu?

If you know how, would you mind reorganising the PR into two commits:
- Add `mutt_mb_charlen_first()`
- Add UTF-8 support to `mutt_yesorno()`

Don't worry if you don't.
I can sort it in the morning.

Thanks!

gahr:
I would *love* to have NeoMutt utf-8 only, but I don’t think it currently is the case.

if we assume that input is utf-8, don’t we break a number of other encoding that use high bytes?

Strahinja:
> If you know how, would you mind reorganising the PR into two commits:
> 
> * Add `mutt_mb_charlen_first()`
> * Add UTF-8 support to `mutt_yesorno()`

I was thinking about squashing everything into one commit when it gets approved, but I can also do the above, if the "Add `mutt_mb_charlen_first`" also includes moving the table to another file like [it does now][1].

[1]: https://github.com/neomutt/neomutt/pull/3712/commits/52ac53f944b7add297d041bb74545ed1273ed8f7

> I would _love_ to have NeoMutt utf-8 only, but I don’t think it currently is the case.
> 
> if we assume that input is utf-8, don’t we break a number of other encoding that use high bits?

I can add a check if `LC_ALL` and/or `LC_CTYPE` contain `.UTF-8` and default to single byte input if they don't. The issue was when those environment variables were set to `[something].UTF-8` anyway.

flatcap:
> I can also do the above, if the "Add `mutt_mb_charlen_first`"
> also includes moving the table to another file

Yes, exactly.
Commit 1, create a library function; Commit 2, use the function to great effect :-)

---

> > I would _love_ to have NeoMutt utf-8 only, but I don’t think it currently is the case.
> > if we assume that input is utf-8, don’t we break a number of other encoding that use high bits?

I have no real understanding of how all the language config works, or who we'd alienate if we went pure utf-8.

> I can add a check if `LC_ALL > LC_CTYPE > LANG` contain `.UTF-8` and default to single byte input if they don't. The issue was when those environment variables were set to `[something].UTF-8` anyway.

Feel free to open a [Discussion](https://github.com/neomutt/neomutt/discussions/categories/development) where we can work through any ideas.
Possibly starting with a list of all the points where the users enters text.

Strahinja:
I updated the PR with the environment variable check and tested it with several environment variable setups, please let me know if it is ok now.

Strahinja:
> I have no real understanding of how all the language config works, or who we'd alienate if we went pure utf-8.

Around the year 2000, locales were using various code pages for non-English character input/output. For example, for Serbian under (GNU/)Linux, ISO8859-5 was used instead of UTF-8. The ubiquity of UTF-8 came later, during the 2010s. Today everyone should be using UTF-8, but there is a theoretical possibility someone would not specify `LANG=[somelang].UTF-8`, or specify `LANG=[somelang].ISO8859-1` etc, so the concern is formally justified.

Strahinja:
I made the above changes and squashed some commits, leaving the PR split into two commits.

flatcap:
Lovely, thanks!

gahr:
I don’t think I like the proposed approach, sorry.

To me, this feels like a typical “works for me” hack. It will probably also work for many others, I concede, but it relies on implementation details like the codeset names. I am not comfortable with that.
Systems might use lower case “utf-8”, or omit the dash as in “utf8” in their codesets. I don’t think we should play guess games.
Systems might also use utf-8 by default, so that the locale “fr” really means “fr.utf-8”.

I *do* think we should handle multi-byte characters better, but I think the proper way for doing that is to use getwchar.

flatcap:
Rebased and tidied (lightly :-)

flatcap:
> I don’t think I like the proposed approach, sorry.

OK.  Well, there's no rush on this.

Strahinja:
> To me, this feels like a typical “works for me” hack. It will probably also work for many others, I concede, but it relies on implementation details like the codeset names. I am not comfortable with that. Systems might use lower case “utf-8”, or omit the dash as in “utf8” in their codesets. I don’t think we should play guess games. Systems might also use utf-8 by default, so that the locale “fr” really means “fr.utf-8”.

Well, a "works for me hack" would for me currently be creating a script or defining an alias which would call Neomutt with modified l10n variables which contain `sr_RS` without `.UTF-8` just for Neomutt. It would be much better if this was solved more systematically.

After some searching, I found out the function [`nl_langinfo`][1] which could be used instead of inspecting the environment, and is supposedly the guaranteed approach. I might switch this PR to using that function instead.

[1]: https://linux.die.net/man/3/nl_langinfo

Strahinja:
I replaced the querying of l10n environment variables with `nl_langinfo` call. This call is used by locale(1), and for testing, I executed `locale charmap` in a Gentoo installation environment. For the default setting of `LANG=C.UTF8` (note the lack of dash), it returns `UTF-8`.

gahr:
If you want to detect utf-8 in the configured charset, see the `charset`config, and the functions`mutt_ch_get_langinfo_charset` and `mutt_ch_is_utf8`.

but I don’t think you want to do that: you want to read a (possibly) wide char from stdin and compare it to some known character.

Strahinja:
> See getwchar and wcstombs.

That's what I asked here: https://github.com/neomutt/neomutt/issues/3691#issuecomment-1433566710

So I guess what @flatcap said here:
https://github.com/neomutt/neomutt/issues/3691#issuecomment-1433933266
and I used as a basis for this PR is out of the question?

gahr:
Well, no, it’s not out of the question. I just offered my opinion on it, but flatcap has the final say :)

flatcap:
> flatcap has the final say

Yeah, the person least-qualified to answer :-)

My initial idea/hack, some extra calls to `mutt_getch()`, was probably good enough.
That function also deals with NeoMutt's keyboard / macro buffer, and inotify events.
Things would work, and the hack is localised to one small function.

> `getwchar()`

At the moment, we're using ncurses' `getch()`.
Besides characters, that also receives mouse events (which I'd like to keep).

Will libc's `getwchar()` play well with ncurses?  Who knows?

> utf-8 native NeoMutt

Given how long utf-8 locales have been around, this sounds like something we should do.
Perhaps we introduce a `$use_utf8` temporarily (defaulted to **yes**).
Perhaps we just turn off support for non-utf-8.
 
We can't get rid of ncurses, so perhaps we experiment with moving this "multiple-calls-to-getch()" deep into `mutt_getch()`.

Enough rambling, I'm going for a walk.  Back later :-)

gahr:
We could use ncurses’ get_wch

flatcap:
> `get_wch()`

Ah!  Now, we're getting somewhere :-)

flatcap:
Cut'n'pasting that into `mutt_what_key()` seems to work correctly

Strahinja:
Can you please clarify what is the status of this issue at the moment? From the conversation so far, it seems that I'm expected to replace calls to `mutt_getch_timeout` with `get_wch` inside of `mutt_yesorno`? However, from what I see, `get_wch` isn't Unicode/UTF-8 aware?

flatcap:
> what is the status of this issue at the moment?

I haven't had a chance to really dive into the code.

We currently use `getch()` in three places.
I need to trace back all the callers and work out how much work this will be.

---

There **is** another translation-specific enhancement, however...

Upstream added a flag to 8 config variables to allow translations.
They are:
- `$attribution_intro`
- `$compose_format`
- `$forward_attribution_intro`
- `$forward_attribution_trailer`
- `$reply_regex`
- `$status_format`
- `$ts_icon_format`
- `$ts_status_format`

We just need to flag those config and, on startup, lookup the translations.
If you're interested, I'll write up all the details.

Thanks!

Strahinja:
I read a bit about `wint_t` and tested some code using `get_wch`, and I think I have a sufficient grasp of it now.

As for 

> Upstream added a flag to 8 config variables to allow translations.

I'm not sure if I have the necessary knowledge of Neomutt internals to handle this separate issue. I could try to take a look.

flatcap:
> I'm not sure if I have the necessary knowledge of Neomutt internals to handle this separate issue.

Actually, it should be a lot easier that this one :-)

> I could try to take a look

Thanks.  I've put the details in #3747 
More info about the [Config System](https://neomutt.org/dev/config)

flatcap:
> I read a bit about `wint_t` and tested some code using `get_wch`, and I think I have a sufficient grasp of it now.

That's great!
Please share anything you find -- don't assume **I** know anything :-)

flatcap:

flatcap:
Ping?

Strahinja:
> Ping?

If you were referring to my participation in the issue related to this PR, this is from our last conversation on Feb 25th:

> > what is the status of this issue at the moment?
> 
> I haven't had a chance to really dive into the code.
> 
> We currently use `getch()` in three places. I need to trace back all the callers and work out how much work this will be.

About this another, unrelated issue, 

> There **is** another translation-specific enhancement, however...
> 
> Upstream added a flag to 8 config variables to allow translations. They are:
> 
> * `$attribution_intro`
> * `$compose_format`
> * `$forward_attribution_intro`
> * `$forward_attribution_trailer`
> * `$reply_regex`
> * `$status_format`
> * `$ts_icon_format`
> * `$ts_status_format`
> 
> We just need to flag those config and, on startup, lookup the translations. If you're interested, I'll write up all the details.

this indeed requires a fair bit of knowledge about the internals of Neomutt codebase, unfortunately more than I can currently dedicate time to, sorry.

flatcap:
> requires a fair bit of knowledge

It doesn't, else I wouldn't have suggested it.

> more than I can currently dedicate time to

ok

flatcap:
Archived to: https://github.com/neomutt/neomutt-old/tree/question-utf-8

PR #3717, opened 2023-02-18
Allow %[fmt] in $folder_format

leonphilman:
* **What does this PR do?**

This is a patch I've bin sitting on for a while, I've polished it a little, so maybe it might be useful to others.
The patch adds the `%[fmt]` expand to browser format to allow formatting the date with an `strftime(3)` sequence similar to what `$pgp_entry_format` has.

To my knowledge this is the last format which relies on `$date_format` but had no other way of formatting dates, so maybe you guys can think about deprecating `$date_format`.

### DIFF
--- a/browser/browser.c
+++ b/browser/browser.c
@@ -209,7 +209,8 @@ static const char *folder_format_str(char *buf, size_t buflen, size_t col, int c
                                      const char *if_str, const char *else_str,
                                      intptr_t data, MuttFormatFlags flags)
 {
-  char fn[128], fmt[128];
+  char fn[128] = { 0 };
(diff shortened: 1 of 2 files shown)


flatcap:
Welcome to NeoMutt!
Thank you very much :-)

I shall have a good read of this, later.

flatcap:
> maybe you guys can think about deprecating `$date_format`

Hmm... possibly.
We could start by giving examples of how get the same effect without it.

flatcap (approved):
Lovely job.
Thank you! :heart:

Are you sitting on any other useful patches?
Got any other good ideas?

Issue #3716, opened 2023-02-17
`$indent_string` and `$post_indent_string` do not respect `$attribution_locale`


s-peppler:
## Expected Behaviour

`$indent_string` and `$post_indent_string` should respect `$attribution_locale`

## Actual Behaviour

`$indent_string` and `$post_indent_string` do not respect `$attribution_locale`

## Steps to Reproduce

```
set attribution_locale="C"
set date_format = "%c"
set attribution="%d"
set indent_string="%d"
set post_indent_string="%d"
```

New reply to an email and as long as your system local has a different way of displaying the date as the `C` locale you will see a difference.

## How often does this happen?

- Always

PR #3718, opened 2023-02-19
Document alternatives to $date_format

leonphilman:
* **What does this PR do?**

This is a followup to #3717 following a suggestion of @flatcap.
Describe alternatives to using $date_format, namely `%[fmt]` (or in the case of the index also `%{fmt}`)

PR #3719, opened 2023-02-19
Let $indent_string, $post_indent_string respect $attribution_locale

leonphilman:
* **What does this PR do?**

While tracking down the usage sides of $date_format (cf #3718), I also fixed the issue #3716.

Closes: #3716


flatcap (approved):
That's brililant!
Thanks :heart:

flatcap:
Or it would be, if the build hadn't failed :-)

PR #3720, opened 2023-02-19
Let patterns cover Bcc header

leonphilman:
* **What does this PR do?**

Maybe answer @flatcap 's question:
> Are you sitting on any other useful patches?

Whether this is useful to you or not depends.  It slightly modifies existing patterns (or phrased pessimistically it breaks them) and steals another character from the alphabet for the Bcc field.  Note that lower k is used to find keys, since I do not expect a ~c <-> ~C duality for Bcc, one letter is all we need.  I believe I chose K because its counterpart was already taken, instead of e.g. j where both j, J are still free.

1) patterns ~C, %C, ~L and ~p now also search Bcc
2) add pattern ~K to search Bcc field

### DIFF
--- a/pattern/exec.c
+++ b/pattern/exec.c
@@ -385,7 +385,7 @@ static int match_addrlist(struct Pattern *pat, bool match_personal, int n, ...)
   va_list ap;
 
   va_start(ap, n);
-  for (; n; n--)
+  while (n-- > 0)
   {
     struct AddressList *al = va_arg(ap, struct AddressList *);
     struct Address *a = NULL;
--- a/pattern/exec.c
+++ b/pattern/exec.c
@@ -477,13 +477,13 @@ bool mutt_is_list_recipient(bool all_addr, struct Envelope *env)
 /**
  * match_user - Matches the user's email Address
  * @param all_addr If true, ALL Addresses must refer to the user
- * @param n       number of AddressesLists supplied
- * @param ...     Variable number of AddressesLists
+ * @param n       number of AddressLists supplied
+ * @param ...     Variable number of AddressLists
  * @retval true
  * - One Address refers to the user (all_addr is false)
  * - All the Addresses refer to the user (all_addr is true)
  */
-static int match_user(int all_addr, int n, ...)
+static int match_user(bool all_addr, int n, ...)
 {
   va_list ap;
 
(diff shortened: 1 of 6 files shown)


pattern/exec.c
```diff
@@ -477,32 +477,31 @@ bool mutt_is_list_recipient(bool all_addr, struct Envelope *env)
...
-    {
-      if (all_addr ^ mutt_addr_is_user(a))
-        return !all_addr;
-    }
-  }
+  va_list ap;
 
-  if (al2)
+  va_start(ap, n);
+  for (; n; n--)
```
gahr, line 491:
```suggestion
  while (n--)
```

gahr:
Nice. I am not covinced by the choice of `K`, but I don't have any better suggestion (without scanning through the list),

leonphilman:
> Nice. I am not covinced by the choice of `K`, but I don't have any better suggestion (without scanning through the list),

The list of free is: `aKoqWZjJ`
I would only rule out `j` or `J` since that's the only pair left and we don't have a pair.  From the other's I picked one randomly (K reminded me of black as in YMCK which reminds me of blind).

For me any letter will do.

---

I incorporated your suggestion and streamlined the history.
There are now also two new bonus commits. ;-)

gahr:
> I incorporated your suggestion and streamlined the history

Curious, why `while (n-- > 0)` instead of `while (n--)`?

leonphilman:
> > I incorporated your suggestion and streamlined the history
> 
> Curious, why `while (n-- > 0)` instead of `while (n--)`?

Because we are testing for a boolean condition and n-- is a number not a boolean. If you like more the implicit approach of an int is a bool I can change that.

gahr:
> > > I incorporated your suggestion and streamlined the history
> > 
> > 
> > Curious, why `while (n-- > 0)` instead of `while (n--)`?
> 
> Because we are testing for a boolean condition and n-- is a number not a boolean. If you like more the implicit approach of an int is a bool I can change that.

Why didn't this apply to the original `for` version?

leonphilman:
> Why didn't this apply to the original for version?

That I copied from `match_addrlist()`

gahr:
Oh - my bad, I hadn't realized the code was there already!

gahr:
> (K reminded me of black as in YMCK which reminds me of blind).

I guess this is as good a rationale as it gets :)

flatcap:
Rebased; minor tidy

PR #3721, opened 2023-02-19
Document $resolve a little more

whitney-cumber:
$resolve not only applies to messages but also to attachments and potentially others.


flatcap (approved):
lovely. thanks

PR #3722, opened 2023-02-20
Generate compile_commands.json

gahr:
This adds a `--compile-commands` configure option to generated a clang
compilation database. This only works if the compiler is clang.

### DIFF
--- a/test/Makefile.autosetup
+++ b/test/Makefile.autosetup
@@ -648,6 +648,7 @@ TEST_OBJS	= test/main.o test/common.o \
 		  $(TAGS_OBJS) \
 		  $(THREAD_OBJS) \
 		  $(URL_OBJS)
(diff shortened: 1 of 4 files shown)


test/Makefile.autosetup
```diff
@@ -648,6 +648,7 @@ TEST_OBJS	= test/main.o test/common.o \
 		  $(TAGS_OBJS) \
 		  $(THREAD_OBJS) \
 		  $(URL_OBJS)
+ALLOBJS+=   $(TEST_OBJS)
```
flatcap, line 651: This line seems unnecessary.
It forces building the test code.

Removing it, both builds behave correctly:
```sh
./configure --compile-commands
./configure --compile-commands --testing
```

gahr: Not anymore :)

flatcap (approved):

PR #3723, opened 2023-02-21
Rewording around Regex and Pattern

whitney-cumber:
* **What does this PR do?**

1) Try to use the word "pattern" only for NeoMutt patterns like `~f Alice` otherwise use regex or string.  This applies to the synopsis of commands as well as their description

2) Add examples how to quote inside patterns and inside regexs

3) Fix some quoting mistakes in examples (last commit, please verify)


flatcap (approved):
Beautiful job.
I've added a few extra notes about $default_hook and $simple_search.
Thanks!

Issue #3704, opened 2023-02-15
Allow colouring the postponed menu


whitney-cumber:
## What would you like to be able to do?

The postponed menu (which is shown when answering yes to the question "Recall postponed message?" or by calling `<recall-message>`) is not coloured.  I'd like to colour it.

The entry lines of the messages is the same as the entry line for the index (`$index_format`), so maybe just use those colours?!?

flatcap:
Yes, that should be easy.
A quick scan of the code shows that postpone shares all the same code:

**Setup:**
https://github.com/neomutt/neomutt/blob/20cebbe5dc0ad263d4bdbaac2761435b68107753/postpone/dlg_postpone.c#L174

**Use:**
https://github.com/neomutt/neomutt/blob/20cebbe5dc0ad263d4bdbaac2761435b68107753/postpone/dlg_postpone.c#L101-L108

In fact, just adding `MUTT_FORMAT_INDEX` to the `flags` param of `mutt_make_string()` might be enough.

PR #3726, opened 2023-02-21
colour postponed email

flatcap:
The Postponed Email list is almost identical to the Index, so we can colour it in the same way.

Fixes: #3704 

---

To test:
- Postpone some emails
- Set some index colours (below)
- Open postponed emails `<recall-message>` (<kbd>R</kbd>)

```
color index         blue    yellow
color index_author  red     green
color index_subject magenta cyan
```


gahr (approved):

whitney-cumber:
Tested it and worked on my end. Thanks!

PR #3727, opened 2023-02-21
SMTP: Try all available methods even if SASL is not compiled in

rayfordshire:
I quote from the $smtp_authenticators documentation:

> If it is "unset" (the  default) NeoMutt  will  try all available methods, in order from most-secure to least-secure.

Once this PR is merged that statement becomes actually true.


flatcap (approved):
Argh!  Good spot!
Thanks!

PR #3728, opened 2023-02-22
Update autosetup

gahr:
This makes conditional includes effective:

```
@if ENABLE_UNIT_TESTS
@include @srcdir@/test/Makefile.autosetup
@endif
```


flatcap (approved):
My first reaction to the second commit (the GitHub Actions) was: WHAT?!
Followed by a dawning, Oh!
It seems that I've been unknowingly relying on that autosetup bug for a while :-)
Even after configuring **without** `--testing`, I could still `make test`.
Now things behave more consistently :-)

PR #3730, opened 2023-02-22
Fix decryption issue when postponing S/MIME encrypted mails

### LINKED ISSUE #3711: Postponing encrypted messages corrupts headers
## Expected Behaviour

Proper handling of encrypted postponed messages

## Actual Behaviour

Headers are mangled, see below.

## Steps to Reproduce

Use the following `neomuttrc-enc-postpone`:
```
set crypt_use_gpgme
set crypt_protected_headers_write
set postpone_encrypt
set postponed=/tmp/postponed
```

1. `neomutt -F neomuttrc-enc-postone`
2. Compose new message with
   a) subject "Menu"
   b) body "Mail à la carte" (some non-ASCII, Latin 1 Supplement text, note the grave on the A)
3. After exiting the editor the attachments read:
   ```
   -- Attachments
   - I     1 /tmp/neomutt-1233355          [text/plain, 8bit, iso-8859-1, 0.1K]
   ```
4. Select Sign+Encrypt in S/MIME menu `<smime-menu>`
5. Postpone the message: `<postpone-message>`
6. Recall the message: `<recall-message>`; you should be asked for a passphrase (if not cached)
7. Back in the compose menu, note
   a) Subject is "..."
   b) Attachments changed (charset and name)
   ```
   -- Attachments
   - A     1 smime.p7m            [text/plain, 8bit, utf-8, 0.1K]
   ```
8. Send the mail (to someone/somewhere where you can retrieve it)
9. Retrive said the massage
10. Figure out how to get the headers of the encrypted message part, NeoMutt currently refuses to show them to you (see #3706): They are:
```
Content-Type: text/plain; protected-headers=v1; charset=utf-8;
	name="smime.p7m"; smime-type=enveloped-data
Content-Disposition: attachment; filename="smime.p7m"
Content-Transfer-Encoding: quoted-printable
Subject: ...
MIME-Version: 1.0
```

The expected headers are

```
Content-Type: text/plain; protected-headers=v1; charset=iso-8859-1;
Content-Disposition: inline
Content-Transfer-Encoding: quoted-printable
Subject: Menu
MIME-Version: 1.0
```

Please note:
* wrong charset
* additional parameters to Content-Type
* wrong Content-Disposition
* wrong subject

As one consequence, when viewing the mail with NeoMutt, the `à` is mangled to `\340`.
Other mailers, e.g. Thunderbird, refuse to show the message to the reader and only show an attachment.

Not tested but might also be present when using the classic backends instead of GPGME and when using PGP instead of S/MIME.

## NeoMutt Version

```
20220429-395-a1d277
```

rayfordshire:
## What does this PR do?

I don't claim that I fully understand what the code does, but by symmetry between S/MIME end PGP this should be as presented here (and it fixes #3711 too!)

## What I (believe to) have found out:

The issue is that when postponing a Mail marked for S/MIME encryption and then recalling that encrypted message, then the headers seem to get corrupted.  Actually, I believe now that the headers are not corrupted they are the headers belonging to the outer message + some random(?) modification to set the MIME type to text/plain:

https://github.com/neomutt/neomutt/blob/01733225d4eed2a12b8400d37df881b3d94143c5/postpone/postpone.c#L466-L467

(Quest for the ambitious: Figure out why this is there and if it can be removed)

I believe what actually happened was, that the S/MIME message was *not decrypted at all*! A decryption code path existed for PGP but was simply missing for S/MIME.  One of the commits adds the same code path as for PGP but uses S/MIME.
I guess by sheer coincidence a later level decrypted the message (but that level did not care about any headers...)

## What else

The `dump_envelope()` crashed if given a null pointer, now fixed.  Also changed (IMHO enhanced) the debug methods to print if they got a null pointer. Take it or leave it I don't care.

In the code path the protected headers were also treated differently between PGP and S/MIME.  I don't understand what the code does but I'm a firm believer that there should be no difference between PGP and S/MIME.  (Hopefully this does not destroy some hidden invariance somewhere else...)

And a little treat to reduce duplication (and hopefully not forget one of PGP / S/MIME) the next time something gets updated.

## Conclusion

This fixes the postponing issue (at least I can no longer replicate it).

An honourable mention: during the debugging I noticed the following code:

https://github.com/neomutt/neomutt/blob/01733225d4eed2a12b8400d37df881b3d94143c5/ncrypt/crypt_gpgme.c#L2759-L2819

Here `tattach` contains the decrypted part including Content-Type etc headers.  But only some (e.g. `mime_headers`) are actually preserved, the rest is thrown away. (Incidentally, that was the layer which decrypted the postponed message later and one reason for the charset mismatch). Also fascinating(?) `decrypt_part()` passes its return value to its caller via a file descriptor...

And lastly, **kudos to all the maintainers**.

## What are the relevant issue numbers?

#3711

### DIFF
--- a/postpone/postpone.c
+++ b/postpone/postpone.c
@@ -367,7 +367,6 @@ static int create_tmp_files_for_attachments(FILE *fp_body, struct Buffer *file,
 {
   struct Body *b = NULL;
   struct State s = { 0 };
-  SecurityFlags sec_type;
 
   s.fp_in = fp_body;
 
@@ -418,13 +417,19 @@ static int create_tmp_files_for_attachments(FILE *fp_body, struct Buffer *file,
       if (!s.fp_out)
         return -1;
 
-      if (((WithCrypto & APPLICATION_PGP) != 0) &&
-          ((sec_type = mutt_is_application_pgp(b)) & (SEC_ENCRYPT | SEC_SIGN)))
+      SecurityFlags sec_type = SEC_NO_FLAGS;
+      if (((WithCrypto & APPLICATION_PGP) != 0) && sec_type == SEC_NO_FLAGS)
+        sec_type = mutt_is_application_pgp(b);
+      if (((WithCrypto & APPLICATION_SMIME) != 0) && sec_type == SEC_NO_FLAGS)
+        sec_type = mutt_is_application_smime(b);
+      if (sec_type & (SEC_ENCRYPT | SEC_SIGN))
       {
         if (sec_type & SEC_ENCRYPT)
         {
-          if (!crypt_valid_passphrase(APPLICATION_PGP))
+          if (!crypt_valid_passphrase(sec_type))
             return -1;
+          if (sec_type & APPLICATION_SMIME)
+            crypt_smime_getkeys(e_new->env);
           mutt_message(_("Decrypting message..."));
         }
 
@@ -434,6 +439,7 @@ static int create_tmp_files_for_attachments(FILE *fp_body, struct Buffer *file,
           return -1;
         }
 
+        /* Is this the first body part? Then save the headers. */
         if ((b == body) && !protected_headers)
         {
           protected_headers = b->mime_headers;
@@ -443,34 +449,8 @@ static int create_tmp_files_for_attachments(FILE *fp_body, struct Buffer *file,
         e_new->security |= sec_type;
         b->type = TYPE_TEXT;
         mutt_str_replace(&b->subtype, "plain");
-        mutt_param_delete(&b->parameter, "x-action");
-      }
-      else if (((WithCrypto & APPLICATION_SMIME) != 0) &&
-               ((sec_type = mutt_is_application_smime(b)) & (SEC_ENCRYPT | SEC_SIGN)))
-      {
-        if (sec_type & SEC_ENCRYPT)
-        {
-          if (!crypt_valid_passphrase(APPLICATION_SMIME))
-            return -1;
-          crypt_smime_getkeys(e_new->env);
-          mutt_message(_("Decrypting message..."));
-        }
-
-        if (mutt_body_handler(b, &s) < 0)
-        {
-          mutt_error(_("Decryption failed"));
-          return -1;
-        }
-
-        if ((b == body) && !protected_headers)
-        {
-          protected_headers = b->mime_headers;
-          b->mime_headers = NULL;
-        }
-
-        e_new->security |= sec_type;
-        b->type = TYPE_TEXT;
-        mutt_str_replace(&b->subtype, "plain");
+        if (sec_type & APPLICATION_PGP)
+          mutt_param_delete(&b->parameter, "x-action");
       }
       else
       {
--- a/postpone/postpone.c
+++ b/postpone/postpone.c
@@ -462,6 +462,12 @@ static int create_tmp_files_for_attachments(FILE *fp_body, struct Buffer *file,
           return -1;
         }
 
+        if ((b == body) && !protected_headers)
+        {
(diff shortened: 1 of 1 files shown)


debug/email.c
```diff
@@ -139,13 +145,17 @@ void dump_envelope(const struct Envelope *env)
...
-
   struct Buffer buf = mutt_buffer_make(256);
   char arr[256];
 
   mutt_debug(LL_DEBUG1, "Email\n");
+
+  if (!e)
+  {
+    mutt_debug(LL_DEBUG1, "\tNULL pointer\n");
+    return;
```
gahr, line 156: This leaks `buf`

debug/email.c
```diff
@@ -242,14 +261,17 @@ void dump_param_list(const struct ParameterList *pl)
...
-  if (!body)
-    return;
-
   struct Buffer buf = mutt_buffer_make(256);
   char arr[256];
 
   mutt_debug(LL_DEBUG1, "Body\n");
 
+  if (!body)
+  {
```
gahr, line 270: This leaks `buf`

postpone/postpone.c
```diff
@@ -418,13 +417,19 @@ static int create_tmp_files_for_attachments(FILE *fp_body, struct Buffer *file,
...
         return -1;
 
-      if (((WithCrypto & APPLICATION_PGP) != 0) &&
-          ((sec_type = mutt_is_application_pgp(b)) & (SEC_ENCRYPT | SEC_SIGN)))
+      SecurityFlags sec_type = SEC_NO_FLAGS;
+      if (((WithCrypto & APPLICATION_PGP) != 0) && !sec_type)
+        sec_type = mutt_is_application_pgp(b);
+      if (((WithCrypto & APPLICATION_SMIME) != 0) && !sec_type)
+        sec_type = mutt_is_application_smime(b);
+      if (sec_type & (SEC_ENCRYPT | SEC_SIGN))
```
gahr, line 425: I’m not convinced by the use of `sec_type` as a bool, using the fact that `SEC_NO_FLAGS` is 0. Wouldn’t it be better to test explicitely?

gahr:
Thanks, I’ll have a deeper look in the morning (on mobile now) - but it looks good from a first sight! A couple of minor things are commented inline.

rayfordshire:
Fixed the `buf` leaks, thanks for finding them.

>```c
>      SecurityFlags sec_type = SEC_NO_FLAGS;
>      if (((WithCrypto & APPLICATION_PGP) != 0) && !sec_type)
>        sec_type = mutt_is_application_pgp(b);
>      if (((WithCrypto & APPLICATION_SMIME) != 0) && !sec_type)
>        sec_type = mutt_is_application_smime(b);
>      if (sec_type & (SEC_ENCRYPT | SEC_SIGN))
>      { //...
>```
>
> I’m not convinced by the use of sec_type as a bool, using the fact that SEC_NO_FLAGS is 0. Wouldn’t it be better to test explicitely?

I'm also not a fan of it but the code I rewrote did that, so at least I was sure not to break anything (I still don't know what assumptions can be made about the flag/return value of `mutt_is_application_smime()` et al). If there is a better solution I'm happy to hear.

flatcap:
Rebased; no code changes

gahr:
> I'm also not a fan of it but the code I rewrote did that

Fair enough - I'd test against `SEC_NO_FLAGS` explicitely.

rayfordshire:
>> I'm also not a fan of it but the code I rewrote did that
>
>Fair enough - I'd test against SEC_NO_FLAGS explicitely.

Ok, I replaced `!sec_flag` with `sec_fag == SEC_NO_FLAGS`.

gahr (approved):

Issue #3732, opened 2023-02-22
Linking failed with `./configure --debug-mail` due to missing symbols


rayfordshire:
## Expected Behaviour

`./configure --debug-mail && make` succeeds.

## Actual Behaviour

`./configure --debug-mail && make` fails the linking stage due to missing symbols:

* `add_flag()`
* `get_content_disposition()`

These are defined in `debug/graphviz.c`, which is not linked unless `--debug-graphviz` is given to configure.

And while we are at it: Maybe add a `--debug-all` option to `./configure` which enables all the debug options. (Note: `--debug` does nothing wrt libdebug?)

flatcap:
Oh, sorry.
The `--debug-*` options are all the result of some need at the time.
Even I don't turn them on _that_ often, so they get neglected.

I'll try to make some time, to fix this.

Also, I'll probably create a GitHub action to do a weekly test build of all the `--debug-*` options so they don't get out-of-date again.

flatcap:
Or I can use your hard work, thanks :-)

flatcap:
Please can you give #3733 a try -- it now builds with or without email/graphviz.

I've tweaked the order of your fixes so that it doesn't leak if the pointers are `NULL`.

rayfordshire:
> I've tweaked the order of your fixes so that it doesn't leak if the pointers are NULL.

Funny, I fixed them, too, (but over in #3730) since @gahr pointed them out to me.

> Please can you give https://github.com/neomutt/neomutt/pull/3733 a try -- it now builds with or without email/graphviz.

Compiles and, more importantly, links without errors.

PR #3733, opened 2023-02-22
debug: fix email debugging

flatcap:
Fix possible null pointer dereference in dump_envelope()
Let dump_*() functions print if they received a null pointer
Refactor email & graphviz common code

Fixes: #3732


Makefile.autosetup
```diff
@@ -360,6 +360,9 @@ LIBDEBUGOBJS+=	debug/parse_test.o
 @if USE_DEBUG_WINDOW
 LIBDEBUGOBJS+=	debug/window.o
 @endif
+@if USE_DEBUG_EMAIL || USE_DEBUG_GRAPHVIZ
+LIBDEBUGOBJS+=	debug/common.o
+@endif
 @if HAVE_LIBUNWIND || USE_DEBUG_COLOR || USE_DEBUG_EMAIL || USE_DEBUG_GRAPHVIZ || USE_DEBUG_NOTIFY || USE_DEBUG_PARSE_TEST || USE_DEBUG_WINDOW
 LIBDEBUG=	libdebug.a
```
rayfordshire, line 367: Not sure whether I would guard `common.o` behind the individual `USE_DEBUG_xxx` instead of putting it into the common case (maybe later other debug-foo will use it and then the guard becomes almost as long as the other one).

rayfordshire (approved):
Compiled on my end without linking errors. Thanks!

Makefile.autosetup
flatcap: Yeah, fair point.  Fixed.

Issue #3731, opened 2023-02-22
Neomutt does not return `$my_var` when queried with `neomutt -Q`


rayfordshire:
## Expected Behaviour

Neomutt should return the value of `$my_var`, e.g. something like `set my_var = "foo"`.

## Actual Behaviour

Neomutt says that there is no such variable as `$my_var`.

## Steps to Reproduce

neomuttrc
```
set my_var = 'foo'
```

```
$ neomutt -F neomuttrc -Q my_var
No such variable: my_var 
```

flatcap:
I'm not sure if I was aware this was possible, which probably means I broke it long ago with the config system replacement.

flatcap:
Thanks for the PR.
I started bisecting during "TV Time" and the problem went WAAAAAY back!

PR #3734, opened 2023-02-22
Allow querying user-defined variables ($my_var) via -Q

rayfordshire:
* **What does this PR do?**

Allows querying `$my_var` via `neomutt -Q my_var`

It's not optimal that the output format for the `my_var` case has a different code path than built-in ones, so we must remember to keep this and the syntax for built-in variables coherent.

Fixes: #3731


flatcap (approved):
I've made a couple of minor tweaks.
- I've removed the comments on the `continue`s
- I simplified the `fprintf(stdout, ...)` to just `printf(...)`

Thanks!

PR #3735, opened 2023-02-23
commands.c: fix errors when compiling with clang++

ngortheone:
Preparation for C++
Fix compiler errors in `commands.c` when compiling with clang++

2 types of errors encountered: 
- goto,  jump bypasses variable initialization
- enum assignment `Assigning to [enum type] from incompatible type 'int`

Enum assignment fix is easy - it needs explicit type cast.

goto errors were dealt with using 3 approaches:
- move variable initialization up where possible
- replace jumps with function calls
- replace goto lables with function-local `#define / #undefine` cpp macros. This approach is a bit ugly, but is justified IMHO because allows fixing the issue quickly. When C++ compilation errors are fixed everywhere this tech debt can be quickly reverted (buy using destructors and replacing goto with multiple returns)

But I am open to feedback.

### DIFF
--- a/auto.def
+++ b/auto.def
@@ -310,11 +310,18 @@ if {1} {
     define SENDMAIL /usr/sbin/sendmail
   }
 
+  # Compiler vendor
+  if {[has-define {} __clang__]} {
+    define COMPILER_IS_CLANG
+  } elseif {[has-define {} __GNUC__]} {
+    define COMPILER_IS_GCC
+  } else {
+    define COMPILER_IS_UNKNOWN
+  }
+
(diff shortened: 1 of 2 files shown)


gahr:
I don't think the solution is to compile our C code with a C++ compiler, but to start to move code to proper C++ into dedicated C++ files. The boundary between C and C++ will have to be considered carefully.

I don't think this PR improves the code quality: it is still C, so you don't get any of the benefits of C++, but it's a big mangled so to please a C++ compiler and as a result it looks unC-ish. To me, this is a lose-lose situation.

PR #3737, opened 2023-02-24
Dump user-defined variables in `neomutt -D`

rayfordshire:
* **What does this PR do?**

`neomutt -D` now also lists user-defined variables.

### DIFF
--- a/init.c
+++ b/init.c
@@ -961,13 +961,8 @@ int mutt_query_variables(struct ListHead *queries, bool show_docs)
     const char *myvar_value = myvar_get(np->data);
     if (myvar_value)
     {
-      pretty_var(myvar_value, &value);
-      /* style should match style of dump_config_neo() */
-      if (show_docs)
-        printf("# user-defined variable\n");
-      printf("set %s = %s\n", np->data, value.data);
-      if (show_docs)
-        printf("\n");
+      dump_myvar_neo(np->data, myvar_value,
+                     show_docs ? CS_DUMP_SHOW_DOCS : CS_DUMP_NO_FLAGS, stdout);
       continue;
     }
 
--- a/main.c
+++ b/main.c
@@ -854,6 +854,7 @@ main
     if (one_liner)
       cdflags |= CS_DUMP_SHOW_DOCS;
     dump_config(cs, cdflags, stdout);
+    dump_myvar(cdflags, stdout);
     goto main_ok; // TEST18: neomutt -D
   }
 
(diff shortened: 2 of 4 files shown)


rayfordshire:
@flatcap can you have a look at the failures.  It looks like `test/neomutt-test` must now also link against myvar.o. Thanks.

rayfordshire:
A always wondered why the my_var are not part of the config.
Is it that nobody implemented it or because they are fundamentally different in the way that they are not known at compile time?

From the above proposed I would either take 1 or 3 because If everything under config/ is only the `ConfigSubset` machinery, then adding `myvar.[ch]` doesn't fit.  Which one depends on the answer to the question the previous paragraph.

flatcap:
I created our config system to workaround some problems with upstream Mutt's.

Upstream has 420 config variables.
Half of them are backed by global variables, `static char *HdrFmt` (naming isn't consistent),
and the bools and quads are efficiently packed into `unsigned char Options`.
`my_var`s are stored in a dynamically resized array.

See: [init.c](https://github.com/neomutt/upstream-mutt/blob/master/init.c) and [init.h](https://github.com/neomutt/upstream-mutt/blob/master/init.h)

Their `parse_set()` function needs to understand every type, perform any checks and worst of all, notify other parts of the code.
Mutt's code has direct access to these config variables, and can change them at any time.

In our config system, we define the types behind a [Config API](https://neomutt.org/code/group__cfg__type__api.html) and define rules for the types, such as `DT_NOT_NEGATIVE`.  We also implement `validator`s for specific requirements.
In NeoMutt, all variable changes go through the Config API.

When a config variable is changed, a notification is sent out.
NeoMutt has lots of observer functions that listen for changes and perform updates.
e.g. changing `$sidebar_format` will cause **only** the sidebar window to be repainted.

flatcap:
>  Is it that nobody implemented it

Yes.  At the time, I didn't think to do it.

> or because they are fundamentally different in the way that they are not known at compile time?

They're similar enough to merge.
Our config variables are loaded into a `HashTable` at startup.
There's nothing stopping us adding the my_vars too.

> I would either take 1 or 3

3 is the better, more beautiful, solution.
If you're willing to give it a go, that'd be brilliant.
I'll have a look through the code later and give you some pointers.

Thanks!

rayfordshire:
Meanwhile here the code with the dump-functions moved into `myvar.c`

flatcap (approved):
Lovely, thanks!

flatcap:
Notes added as #3746

Issue #3706, opened 2023-02-15
Does not show complete MIME-structure


rayfordshire:
## Expected Behaviour

NeoMutt should show all MIME-parts in the attachment menu.

## Actual Behaviour

Sometime MIME-parts are suppressed.

## Steps to Reproduce

Consider this sample (fake) encrypted and signed S/MIME message (where the encryption is stripped to show the inner MIME-parts):

```
From: Alice
To: Bob
Content-Type: application/pkcs7-mime; smime-type=enveloped-data;
        name="smime.p7m"
Content-Disposition: attachment; filename="smime.p7m"

[------ Start of encoded data ------]
Content-Type: multipart/signed; micalg=sha256;
	protocol="application/pkcs7-signature"; boundary="foo"
Content-Disposition: inline


--foo
Content-Type: text/plain
Content-Disposition: inline

Foo

--foo
Content-Type: application/pkcs7-signature; name="smime.p7s"
Content-Disposition: attachment; filename="smime.p7s"
Content-Transfer-Encoding: base64

[lots of base64]
--foo--
[------ End of encoded data ------]
```

NeoMutt only shows the inner two mime parts (text and signature) but not the middle layer (multipart-signed):
```
<no description>                   [text/plain, quoted-printable, us-ascii, 0.1K]
S/MIME Cryptographic Signature        [application/pkcs7-signature, base64, 6.4K]
```
This make it currently hard for me to debug an S/MIME problem.  I have to resort back to decrypting the content myself to see the inner multipart/signed structure.

Compare this with this unencrypted message:

```
From: Carol
To: Dave
Content-Type: multipart/related; boundary="bar"

--bar
Content-Type: multipart/mixed; boundary="baz"

--baz
Content-Type: text/plain
Content-Disposition: inline

Baz 1

--baz
Content-Type: text/plain
Content-Disposition: inline

Baz 2
--baz--

--bar
Content-Type: text/plain
Content-Disposition: inline

Bar 2

--bar--

```

Here Neomutt shows the middle MIME-part-layer (miltipart-mixed) in the attachment menu.
```
<no description>             [multipart/mixed, 7bit, 0.1K]
├─><no description>     [text/plain, 7bit, us-ascii, 0.1K]
└─><no description>     [text/plain, 7bit, us-ascii, 0.1K]
<no description>        [text/plain, 7bit, us-ascii, 0.1K]
```

[...]

PR #3738, opened 2023-02-24
Allow NeoMutt to show complete MIME structure

rayfordshire:
The attachment menu hides some MIME parts from the user which are usually uninteresting, e.g. a multipart/mixed MIME part at the top level as it is only the container for the attachments.

However, when debugging email problems, especially when encryption is involved, it is useful to see the complete MIME structure. Example where this would have been useful is https://github.com/neomutt/neomutt/issues/3711 I also had a couple of incidents in the past, where I would have liked to have this feature instead of resorting back to an editor and openssl.

Introduce a new config variable $attach_complete_mime_structure, which defaults to false.  If this variable is set, then NeoMutt does not hide any MIME parts from the user.  If unset, it does what it currently does and hides some MIME parts from the user.

---

@gahr I'm sorry, but yes this is a new config variable.  But I don't see another way if the default should be "hide uninteresting stuff from people".

---

Related: https://github.com/neomutt/neomutt/issues/3711
Fixes: https://github.com/neomutt/neomutt/issues/3706

### DIFF
--- a/compose/dlg_compose.c
+++ b/compose/dlg_compose.c
@@ -205,11 +205,10 @@ static void gen_attach_list(struct AttachCtx *actx, struct Body *m, int parent_t
     m->aptr = ap;
     ap->parent_type = parent_type;
     ap->level = level;
-    if ((m->type == TYPE_MULTIPART) && m->parts &&
-        (!(WithCrypto & APPLICATION_PGP) || !mutt_is_multipart_encrypted(m)))
-    {
-      gen_attach_list(actx, m->parts, m->type, level + 1);
-    }
+    //if ((m->type == TYPE_MULTIPART) || mutt_is_message_type(m->type, m->subtype))
+    //{
+    gen_attach_list(actx, m->parts, m->type, level + 1);
+    //}
   }
 }
 
--- a/compose/functions.c
+++ b/compose/functions.c
@@ -363,7 +363,8 @@ static void compose_attach_swap(struct Email *e, struct AttachCtx *actx, int fir
   idx[first] = saved;
 
   // if moved attachment is a group then move subparts too
-  if ((idx[first]->body->type == TYPE_MULTIPART) && (second < actx->idxlen - 1))
+  //if ((idx[first]->body->type == TYPE_MULTIPART) && (second < actx->idxlen - 1))
+  if (idx[first]->body->parts && (second < actx->idxlen - 1))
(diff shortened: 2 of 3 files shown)


gahr:
I wonder if we shouldn't do it always.

flatcap:
> I wonder if we shouldn't do it always.

I'm in favour of that, too.

Compose isn't consistent with received mail -- it's not always fully-formed.
When the email is sent, `mutt_send_message()` checks and adds a wrapper if necessary.

https://github.com/neomutt/neomutt/blob/fd4e14617f95503a36a6bb3162bbe25e56ac3b77/send/send.c#L2817-L2818

rayfordshire:
Reworked the initial proposal, the commits are now:

1.  removing the check to hide some MIME-parts
2. / 3. Two mini improvements
3.-end Attempt to show MIME part in compose menu

The last part is currently [WIP]; the main problem is to get a fully parsed `Body` (or `Email`).  I've modified `mutt_make_message_attach()` which covers the "add mail as attachment" case.  However, postponing still needs to be handled.

**Thus, the first question** when/where are emails usually parsed, so that a `Body` represents the complete MIME structure, i.e. `body->parts` is only NULL if the message does not contain any MIME-parts.  I could add those parsing step for postpone in `postpone/postpone.c` function `mutt_get_postponed()` somewhere around here:

(A hint which function to call would be greatly appreciated)

https://github.com/neomutt/neomutt/blob/b89d8f3f27045d720b5017f6b8ff8bfb18857392/postpone/postpone.c#L714-L719

Another problem is when viewing those MIME-subparts is that they are not backed by a file.  I image the same behaviour as with the regular attachment menu. I tried setting the filename but that a) didn't work and b) made the attachment list look ugly.

**Second question**: How can the MIME-subparts be shown (with function `view-attach`).

flatcap:

rayfordshire:
How do we proceed here?

The first three commits fix the problem for the attach menu (and can be merged), the others are concerned with the compose menu and have open questions: https://github.com/neomutt/neomutt/pull/3738#issuecomment-1453315196

flatcap:
> The first three commits fix the problem for the attach menu (and can be merged)

Thanks.  Done.

flatcap:
Archived to: https://github.com/neomutt/neomutt-old/tree/show-complete-mime-structure

Issue #3741, opened 2023-02-24
Config: Some resets don't set the default value


rayfordshire:
# The Problem

Some config variables have only a runtime default which is not known at compile time, examples include `$tmp_dir` (default is $TMPDIR with fallback to `/tmp`), `$suspend` (default is `yes` or `no` depending on whether NeoMutt is process group leader), `$editor` (default $EDITOR + fallbacks) and probably some more.

Here is an example which shows that default is not equal to resetted:
`# TMPDIR=foo neomutt -F /dev/null`
`:set tmp_dir?` -> `foo`
`reset tmp_dir`
`:set tmp_dir?` -> `/tmp`

# Idea 1: "store tuples"

Extend the Config system to store an additional default value which can be set like a normal value, e.g. `set_default("tmp_dir", "my/fancy/dir");` with fall back to the compile time default.  When NeoMutt initialises itself, it sets the defaults according to whatever it found out.

# Idea 2: "default functions"

Extend Config system by replacing the default value with a function (pointer) returning the default value.  Whenever the default is requested that function is asked and it returns whatever it currently finds is the default value. This approach allows to dynamically react to changes in the environment and adapt the config value accordingly, e.g. the default value should be the current time.
The current defaults would just be a function which returns set default, e.g. `default_auto_tag() { return false; }`

# Idea 3: "layers"

(sort of othogonal to 1 and 2)
Extend the Config system so that a ConfigSet falls back to another ConfigSet if the key is not found in it.  Think of them as layers:
top layer are all configs which the user set, bottom layer are all the "defaults".  Setting a default variable would be setting the value in the lower layer.  This would also allow to detect if a user actively set a variable or not by the fact whether it is present in the upper level or not. (More layers are easily possible.)

---

@flatcap what do you think of these approaches?  Which one would fit in your vision?

Idea 1 and 3 are more in line with "config system is only a data storage".  Idea 2 seems more flexible.

flatcap:
> Some resets don't set the default value

*sigh*
Fixed in #3742 

All the magic happens here in `mutt_init()`:
https://github.com/neomutt/neomutt/blob/8c3731301728c9f8b276ab50d8d3a1da7653bb3f/init.c#L603-L678

The complete list is: `$charset`, `$editor`, `$from`, `$hostname`, `$mailcap_path`, `$real_name`, `$spool_file`, `$suspend`, `$tmp_dir`.

flatcap:
> Idea 1: "store tuples"

[...]

PR #3742, opened 2023-02-25
config: fix initial values

flatcap:
A few config variables' initial values weren't being set correctly.

Fix the initial values of:
- `$mailcap_path`
- `$tmp_dir`
- `$suspend`

Fixes: #3741


rayfordshire (approved):
LGTM

PR #3748, opened 2023-02-27
sidebar: add background colour

flatcap:
A minor addition to the sidebar:

- Add a `sidebar_background` colour
- Allow the `$sidebar_divider_char` to be empty
- Remove some very old sidebar docs

Give the Sidebar a background colour to allow contrast from the Index.
This now allows the divider character to be turned off (giving the user extra space).

### DIFF
--- a/sidebar/observer.c
+++ b/sidebar/observer.c
@@ -49,10 +49,16 @@ void sb_win_remove_observers(struct MuttWindow *win);
 static bool calc_divider(struct SidebarWindowData *wdata)
 {
   enum DivType type = SB_DIV_USER;
+  bool changed = false;
   const char *const c_sidebar_divider_char = cs_subset_string(NeoMutt->sub, "sidebar_divider_char");
 
   // Calculate the width of the delimiter in screen cells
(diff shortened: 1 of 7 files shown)


sidebar/observer.c
```diff
@@ -84,7 +90,8 @@ static bool calc_divider(struct SidebarWindowData *wdata)
     }
   }
 
-  const bool changed = (width != wdata->divider_width);
+done:
+  changed = (width != wdata->divider_width);
```
gahr, line 94: Can `changed` be initialized here?

gahr (approved):

sidebar/observer.c
flatcap: Oddly, no.
I'm sure `gcc` complained too, but I can't reproduce it today.
However, clang complains:
```
sidebar/observer.c:93:3: error: expected expression
  bool changed = (width != wdata->divider_width);
```

It's not happy to initialise a variable immediately after a label.
Putting `{}`s around the last 4 lines solves this, but looks ugly.

PR #3749, opened 2023-02-27
Fix how group addresses are displayed

### LINKED ISSUE #3745: neomutt violates RFC 5322
## Expected Behaviour
neomutt honors correct RFC5322 group addressing syntax. Relevant quotes from RFC5322:

```
When it is desirable to treat several mailboxes as a single unit
(i.e., in a distribution list), the group construct can be used.  The
group construct allows the sender to indicate a named group of
recipients.  This is done by giving a display name for the group,
followed by a colon, followed by a comma-separated list of any number
of mailboxes (including zero and one), and ending with a semicolon.
```

```
[...]
3.4. Address Specification
[...]
group           =   display-name ":" [group-list] ";" [CFWS]
```

## Actual Behaviour
Neomutt automatically removes the trailing semicolon.

==> No coredump.

## Steps to Reproduce
Start neomutt, send an e-mail with (for instance) the To: field containing `undisclosed-recipients:;" (or a non-trivial group list), and the e-mail will be sent without the mandatory semicolon at the end of the group list.

- Always

Unknown.

## NeoMutt Version

```
NeoMutt 20201127
```
(Cannot easily test newer versions)

## Extra Info

* Devuan Chimaera 4.0
  ==> No
  ==> screen, but that should not matter here.

* Is your email local (maildir) or remote (IMAP)?
Remote (IMAP).

gahr:
Fixes #3745 

Also, refactor the AddressList writing routines to avoid duplicated logic.
In particular, we don't want multiple places in the code to be aware of how we represent address groups:
it's ugly and I would like to change it, eventually.

### DIFF
--- a/alias/dlg_alias.c
+++ b/alias/dlg_alias.c
@@ -126,7 +126,7 @@ static const char *alias_format_str(char *buf, size_t buflen, size_t col, int co
                                     const char *if_str, const char *else_str,
                                     intptr_t data, MuttFormatFlags flags)
 {
-  char fmt[128], addr[1024];
+  char tmp[1024];
   struct AliasView *av = (struct AliasView *) data;
   struct Alias *alias = av->alias;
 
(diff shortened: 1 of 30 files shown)


alias/dlg_alias.c
```diff
@@ -463,14 +467,17 @@ int alias_complete(char *buf, size_t buflen, struct ConfigSubset *sub)
...
   struct AliasView *avp = NULL;
   ARRAY_FOREACH(avp, &mdata.ava)
   {
     if (!avp->is_tagged)
       continue;
 
-    mutt_addrlist_write(&avp->alias->addr, buf, buflen, true);
+    mutt_addrlist_write(&avp->alias->addr, tmpbuf, true);
   }
+  mutt_str_copy(buf, mutt_buffer_string(tmpbuf), sizeof(buf));
```
flatcap, line 479:
```suggestion
  mutt_str_copy(buf, mutt_buffer_string(tmpbuf), buflen);
```

gahr:
@dcpurton please allow me to keep this as a draft for now. I am still testing it.

dcpurton:
@gahr sorry, I think I just pressed a button accidentally in app without even noticing

flatcap (approved):

Issue #3750, opened 2023-02-27
Cannot set $sidebar_divider_char to ""


whitney-cumber:
## Expected Behaviour

Be free to $sidebar_divider_char to "", always regardless of background colour.
(Personally, I do not understand why the requirement to have a background colour set.)

## When did it start to happen?

on Feb 27, 2023, 1:59 PM UTC and 29 seconds to be precise.

flatcap:
History, as always :-)

The user wasn't allowed to set an empty divider char, because of some magic going on behind the scenes.

In combination with the `$ascii_chars` variable, ncurses tries to gracefully degrade to an ASCII `|` (pipe char).

Because most users would want a Unicode `│` (vertical line), but may not have been able to type it, the code interpreted an empty string as that.

PR #3752, opened 2023-03-01
sidebar: allow an empty divider

flatcap:
- Allow an empty value for the divider
- Set the divider's default value to a Unicode vertical line
- Simplify the logic for `$ascii_chars`

If the user sets `$ascii_chars` or the charset isn't utf-8, then the divider character will fallback to an ASCII pipe character `|`.

Fixes: #3750

### DIFF
--- a/docs/config.c
+++ b/docs/config.c
@@ -4193,15 +4193,18 @@
 ** \fBSee also:\fP $$sidebar_short_path, $$sidebar_folder_indent, $$sidebar_indent_string.
 */
 
-{ "sidebar_divider_char", DT_STRING, 0 },
+{ "sidebar_divider_char", DT_STRING, "|" },
 /*
+** .pp
+** The default is a Unicode vertical line.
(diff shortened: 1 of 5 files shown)


whitney-cumber (dismissed):
No clue about the code but `docs/config.h` needs an update in the default value.
I have the feeling that the obvious `"│"` (the Unicode character) might not work so the uses will be shown the cryptic `\342\224\202`. In this case a sentence like "The default value corresponds to the Unicode codepoint Box Drawings Light Vertical (U+2502)." should be added.

flatcap:
> `docs/config.h` needs an update

Oh dear.  `makedoc` really isn't ready for utf-8.

gahr (approved):

flatcap:
@whitney-cumber 

> No clue about the code

Yep, understood.

I appreciate any time you can spare to test
(and point out my failings :-)

Thanks!

Issue #3746, opened 2023-02-25
Integrate myvars into the config system


flatcap:
I'll add more to this later...

---

Existing `myvar` behaviour:

```sh
# create a new variable (value stored as a string)
set my_var = 42
# query variable
set my_var
# empty variable (value stored as NULL)
set my_var = ""
# delete variable
unset my_var
reset my_var
# append to variable (appended as a string)
my_var += apple
```

---

To mimic this behaviour, we have two possible choices:
- Use a `DT_STRING` with a `DT_MYVAR` flag
- Create a new `DT_MYVAR` type

Because we're going to dynamically create/delete config variables,
I think a separate type would be cleaner.
I suggest:

- Add a new type `DT_MYVAR` after `DT_SYNONYM`
  Kill the `DT_MY_CONFIG` comment further down.

- Temporarily, create a couple of new variables in `MainVars[]` 
  say, `my_apple` and `my_banana`

- Cut'n'paste `config/string.c` and replace `string` with `myvar`
  Adjust the `libconfig` build section in `Makefile.autosetup`

- Check that `my_apple` and `my_banana` behave correctly

- Replace any calls to `myvar_get()`, `myvar_set()`, `myvar_append()` with calls to the Config API
  Note, we can't use the `cs_subset_*()` helpers as they assert on failure.
  We can use `cs_subset_lookup()` to find if the myvar exists.

I'll tell you how to create/delete variables dynamically, as soon as I've figured it out myself :-)

rayfordshire:
At which level is the existence of a variable checked?

Background is this: For a normal string variable the following is equivalent:
```
set from = ""
set from?
-> ""

unset from = ""
set from?
-> ""
```

However, for `$my_var` an empty variable is different from an unset one:
```
set my_foo = ""
set my_foo?
-> ""

unset my_foo = ""
set my_foo?
-> my_foo: unknown variable
```

I first thought (since myvariables are essentially strings) that maybe I can use `""` (empty string) as a return value to indicate an empty existing variable and `NULL` to indicate a non-existing variable.  Then I noticed that an empty string is stored as NULL.  What are the reasons for this decision?

Then I wondered where (if at all) there is an existence check for a variable in the config system.

Summary:

* What are the reasons for storing `""` as `NULL`
* What does the value `NULL` mean for the config system?  Are there any assumptions the code relys on, e.g. the returned value (I don't mean the status code here) of the various (all?) `*_get` not being `NULL`.
[...]

PR #3753, opened 2023-03-04
Integrate $my_var into Config system

rayfordshire:
* **What does this PR do?**

Implements the handling of user-defined config variables via the config system.

Fixes: #3746

---

Roadmap:

[ ] fix bugs
[ ] Change all call sites to use the MyVariables system of ConfigSet
[ ] Remove myvar.c, myvar.h


rayfordshire:
The status so far is that the implementation of the new type `DT_MYVARIABLE` is completed.

Some ASAN tests are failing but not at the expected places...

Design decisions:

What should the method `cs_subset_myvariable()` in config/helpers.c be?

1. Currently, it mimics the other `cs_subset_foo()` functions that means
  a. it cannot fail, i.e. the caller is responsible to ensure that the config name exists
  b. returns `NULL` for an empty string

2. We could change that by checking for the existence of the config name, returning `NULL` if it doesn't.  an empty string would be, well, signalled by returning an empty string.

3. Remove that method, the user should use `cs_string_get()` et al.

flatcap:
That's a bit clearer now that I can look at some source :-)
Thanks.

---

There are six `my_var` operations we want to support.

The `my_var` **exists**:

1. Set the value
2. Query the value
3. Reset the variable

The `my_var` **doesn't** exist:

4. Set the value
5. Query the value
6. Reset the variable

---

In `parse_set()`, **1,2,3** are trivial.
We just need to treat them as normal config items.
We don't need any special processing, or checking for `"my_"`.

**5,6** are also simple.
In `main`, if you try to query, or reset, an unset `my_var`, there's no warning or error.
This is a bug.
By removing any `my_var` checks, they'll behave the same as if the user did `:set xyz?`, giving an error.

**4** is a bit trickier.
If we ignore the `(!he)` check, we get to the call to `cs_subset_he_string_set()` (~L1006).
Here, we could instead call `cs_str_string_set()` with the `my_var` name.
This moves any special `my_var` logic into the config library.

In `cs_str_string_set()` we can call a custom `myvariable_create()`.

---

In `myvariable.c` we need to handle creation and destruction of `my_var`s.

`myvariable_create()` can check the name for `"my_"` then create the storage and call `myvariable_string_set()`.
`myvariable_reset()` just needs to call `cs_str_delete()` to delete the `my_var` storage.

---

I hope this makes sense.

rayfordshire:
@flatcap 

Before I refactor the parse code I thought it would be nice to have some test code checking that nothing breaks. (It doesn't seem to have test coverage unless I've overlooked something.)

I've added a test `test_command_set` in `test/command/set.c`, however, ... this brought us back to square one as I depend on `mutt_parse_rc_line()`, which is defined in `init.c` (also applies directly to `parse_set` in `command_parse.c` which is the target function of the tests).  This results in a linker error because the tests do not link against "neomutt-objects" (objects of the executable).

Any clue how to fix that?  (I'd also accept something hacky for my developing purposes to ensure I don't break anything during the refactor).

P.S.: Test code is in https://github.com/neomutt/neomutt/pull/3753/commits/e78bd2a99610524763e357f79d66a7df48306e70 the rest is start of the refactor process.

flatcap:
Rebased, "fixed" test build ;-)

flatcap:
> I thought it would be nice to have some test code checking that nothing breaks

Test-Driven Development -- sounds good

> (It doesn't seem to have test coverage unless I've overlooked something.

Yes.
Our first tests were for the config system.
Next, we refactored all the shared code into `libmutt`  and wrote tests for that.

Besides the tests, you might also want to consider the [code coverage](https://neomutt.org/dev/analysis/coverage).  
You just need `lcov` installed.

> I've added a test `test_command_set`
> this brought us back to square one as I depend on `mutt_parse_rc_line()`,
> This results in a linker error

Yes.  Many of these simple-sounding functions actually have ugly dependencies.
Parsing a config line can rely on config variables and lots more besides.

> Any clue how to fix that?
> (I'd also accept something hacky for my developing purposes).

Your wish is my command :-)

I've moved the dummy `mutt_extract_token()` to where you can use it (`test/common.c`).
And I've copied `mutt_extract_token()` into `test/command/common.c` along with some `Command`s.

It builds, it runs.
Hopefully that'll give you enough to work with.

rayfordshire:
> > I've added a test `test_command_set`
> > this brought us back to square one as I depend on `mutt_parse_rc_line()`,
> > This results in a linker error
> 
> Yes. Many of these simple-sounding functions actually have ugly dependencies. Parsing a config line can rely on config variables and lots more besides.
> 
> > Any clue how to fix that?
> > (I'd also accept something hacky for my developing purposes).
> 
> Your wish is my command :-)
> 
> I've moved the dummy `mutt_extract_token()` to where you can use it (`test/common.c`). And I've copied `mutt_extract_token()` into `test/command/common.c` along with some `Command`s.
> 
> It builds, it runs. Hopefully that'll give you enough to work with.

Thanks for the help but that does not solve the core of the problem: The goal is to test the `parse_set()` function.  This function resides in `command_parse.c` which is not linked in the tests (like `init.c` its a "neomutt" (the executable) object and not a "library" object).  Copying that into `test/command/common.c` seems counter-intuitive as that's the code we want to test.

flatcap:
> that does not solve the core of the problem

Yes, the current `parse_set()` is too high-level for unit-testing -- too many deps.

Not yet, but we will be able to.
With your refactoring, it won't depend on the old `my_var` functions and globals.
If you can refactor the code to only have dependencies on `libmutt` and `libconfig`,
then we can 'promote' the code **to** libconfig.

Hope that makes sense.
Gotta rush, now; back later.

rayfordshire:
So this is essentially it:  All that's left is to replace the remaining call sites (which is mostly removing code as myvar's are no longer treated special).

If you are satisfied with this design I would go ahead and make those changes.  The config code itself will not be touched any more. 

Notable changes:
* Cut down the long function of `parse_set()` into its five functionalities: set, reset, unset, toggle, query.  They are now each their own method with `parse_set()` only doing the parsing of the input string.
* `IS_MAILBOX` and `IS_COMMAND` macros are changed, they test now a type and not a `HashElem`.  This fitted better the design of `command_set_expand_value()` which takes now a type and not a `HashElem`.
* Calling `reset()` on a `DT_MYVARIABLE` does, well, reset the variable to its initial value – just like any other variable.  To delete it, use `cs_*_delete()`. The translation of "reset my_foo" into `cs_delete("my_foo")` is done in `command_parse.c` just like many other distinctions based on the type (e.g. unset does different things for different variables).
* ConfigDef became dynamic: The old config code assumed that all ConfigDef's live indefinitely.  I added a `cs_create_variable()` which is essentially `cs_register_variable()` but copies the `ConfigDef` passed to it (to some extend).  This allows the caller to free its `ConfigDef`, e.g. because it is on the stack.  It also allows us to correctly free the memory of `ConfigDef` when calling `cs_*_delete()`.
* The command "set foo+" is now an error.  Previously this seemed to be silently ignored
* `ConfigSubset` gained a `cs_subset_*_delete()` method mirroring the `cs_*_delete()` method of Config.  Needed a new signal constant `NT_DELETED`.

Remark about the tests:

* In the end I hacked around the test suite (linking against `command_parse.o` but "commenting out" everything unneeded).  The commits which do that are marked with `[REMOVE ME]` and will not be part of the final pull request. I included them here so you are able to run the test suite.
* The `[REVERT ME]` commits will also be purged form the final pull request: They change the test cases from using legacy `myvar_foo()` to modern `cs_str_foo()`
* I see little hope that these tests can be included in the test suite even after my_var is purged.  The main point is that the parse code depends also on `mutt_buffer_expand_path()`, which is in `muttlib.c`.

flatcap:
> So this is essentially it

Fantastic, thank you!
I'll have a good read shortly.

> I see little hope that these tests can be included

I'm doing a bit of function shuffling, so I might have a solution for this.

flatcap:
Rebased; squashed fixup, revert

rayfordshire:
**Addendum:**

It seems like the `ConfigDef->type` is more often treated as `(signed) int` than as `uint32_t`, e.g. `HashElem` uses it, and some regex flags (e.g. `REG_ICASE;`), too.  Before I take a deeper dive.  Any clue how robust NeoMutt's code base is if `int` gets replaced by `uint32_t` or even `uint64_t`?

flatcap:
Rebased over function shuffling and renaming.

These functions have been moved to a new library, `libparse`.
 - `parse_extract_token()`
 - `parse_rc_buffer()`
 - `parse_rc_line()`
 - `parse_set()`

I've also added stub test functions for each of them.

rayfordshire:
Thanks for refactoring the code so that we can now test our parse commands individually.
I'd love to test that, however, the following initial test in `parse_set` lets the tests fail as (it consumes `set` and then fails on `Apple` in the next iteration when parsing the string `set Apple = yes`):

```c
  if (StartupComplete)
  {
    if (set_dump(buf, s, data, err) == MUTT_CMD_SUCCESS)
      return MUTT_CMD_SUCCESS;
    if (!mutt_buffer_is_empty(err))
      return MUTT_CMD_ERROR;
  }
```

`StartupComplete` is true and all what `parse_set` does is maybe dumping some stuff (haven't observed any output) then return with `MUTT_CMD_SUCCESS`.

Why is this the purpose of this code?  It was introduced in  79b6430cf2a09e77e96ac614d6ac5a8dd2f9783f and similar for other parse functions.

flatcap:
> Thanks for refactoring the code so that we can now test our parse commands individually.

No problem.
It's not perfect, but it'll do.

I've also enabled coverage for `libparse`: https://coveralls.io/github/neomutt/neomutt

> I'd love to test that, however, the following initial test in `parse_set` lets the tests fail

Argh, sorry about that.
This will probably help.

```diff
   if (StartupComplete)
   {
+    // Save and restore the offset in `s` because set_dump() might change it
+    char *dptr = s->dptr;
     if (set_dump(buf, s, data, err) == MUTT_CMD_SUCCESS)
       return MUTT_CMD_SUCCESS;
     if (!mutt_buffer_is_empty(err))
       return MUTT_CMD_ERROR;
+    s->dptr = dptr;
   }
```

> Why is this the purpose of this code?

Ages ago, we introduced some 'icommands' (informative) commands.
They give us `:set` to list all the config, `:bind` to show mappings.
They were given their own abstraction, but it was overkill.
It needed to be called before the normal command processing happened.
Also, we don't want `:set` to do anything if it's done during startup (when reading a config file).

There are two commands that need to be supported:
- `:set` show only config that is different from the default
- `:set all` show all config
 
You're welcome to refactor that code.

rayfordshire:
> > I'd love to test that, however, the following initial test in `parse_set` lets the tests fail
> 
> Argh, sorry about that. This will probably help.
> 
> ```diff
>    if (StartupComplete)
>    {
> +    // Save and restore the offset in `s` because set_dump() might change it
> +    char *dptr = s->dptr;
>      if (set_dump(buf, s, data, err) == MUTT_CMD_SUCCESS)
>        return MUTT_CMD_SUCCESS;
>      if (!mutt_buffer_is_empty(err))
>        return MUTT_CMD_ERROR;
> +    s->dptr = dptr;
>    }
> ```

Sorry but that doesn't help.  The problem is that the `if (set_dump(buf, s, data, err) == MUTT_CMD_SUCCESS)` evaluates to `true` and thus the method returns early with `return MUTT_CMD_SUCCESS`. This means the processing of the `set` command is done.  NeoMutt detects that the line still continues with "Apple = yes" and applies again its parser, yielding in a "apple: Unknown command".

> > Why is this the purpose of this code?
> 
> Ages ago, we introduced some 'icommands' (informative) commands. They give us `:set` to list all the config, `:bind` to show mappings. They were given their own abstraction, but it was overkill. It needed to be called before the normal command processing happened. Also, we don't want `:set` to do anything if it's done during startup (when reading a config file).
> 
> There are two commands that need to be supported:
> 
>     * `:set` show only config that is different from the default
> 
>     * `:set all` show all config
> 
> 
> You're welcome to refactor that code.

Ah, the naming suggested me for it to be some kind of "debug" utility/test.  Now this makes a lot more sense.
I'd like to keep "parsing" and "execution" separated: one code does the syntax parsing, the other does the semantic, so to speak.  With this in mind it makes a lot more sense to let the `parse_set()` code handle the identification of `set` (without argument or argument "all") and then call the `dump_set()` functions accordingly, just like we currently do with the whole `command_set_unset()`, `command_set_reset()`, ... functions.

flatcap:
> I'd like to keep "parsing" and "execution" separated

Sounds perfect!

If you ever want to bounce ideas around, more quickly than issue comments, you can always find me on IRC.
`#neomutt` on `irc.libera.chat` ([web client](https://kiwiirc.com/nextclient/irc.libera.chat/#neomutt))

rayfordshire:
**Addendum:**
Just looked into it: the function `set_dump()` has a different implementation depending on whether the test suite is run or not.  The test suites implementation is located in `test/parse/parse_set.c` and just returns `MUTT_CMD_SUCCESS` (no parsing is done). Resulting in the early return.  The production implementation in `commands.c` does check to only accept "set" (no arguments) or "set all".  Thus the different behaviour.

flatcap:
> config/set.c:84:22: runtime error: left shift of 1 by 31 places cannot be represented in type 'int'
> #define DT_FREE_CONFIGDEF (1 << 31)  ///< Config item must have its ConfigDef freed

Using `1U` might help...

> I believe this interpreted as an int even in an operation involving uint

It'll _probably_ be correctly interpreted elsewhere in the code...
Or, you could use `(1 << 26)` instead, we've got the space.

> Question 2: The type bits are now officially used up.
> Should we expand ConfigDef->type to be uint64_t?

Not quite.
There are a few gaps that we could use, plus...

```
typedef uint32_t ConfigRedrawFlags; ///< Flags for redraw/resort, e.g. #R_INDEX
```

These redaw flags are a relic from Mutt.

Mutt has flags on the config to tell some **global code** that certain things, e.g. the Pager, need updating.
In NeoMutt, if the Pager needs updating after $xyz is set, then it must listen for config notifications.

They're still used, but soon(ish), we'll be able to get rid of them.
(freeing up 8 bits :-)

> Question 1: What is the feeling of NeoMutt towards "proper" (i.e. non-macro) global constants:

Positive? :-)

We've come a long way from Mutt -- have you taken a look at their source?
We've bool-ified thousands of 1s and 0s and created typedefs for many types.
But we can do better, including `bool` -> named enums.

Feel free to open a [discussion](https://github.com/neomutt/neomutt/discussions) with suggestions.

> extern const int FOO;
> const int FOO = 42;

Perhaps we start with those that are local to a single file.

> (I'll probably add a typedef for the ConfigDef->type later.)
> (Edit: and clean up the mess of type being sometimes int sometimes uint)

Yes, that'd be useful.
I suppose the ideal situation is to split the type and flags:
```c
struct ConfigDef
{
  ...
  ConfigType  type;
  ConfigFlags flags;
  ...
};
```

But that'd mean changing quite a lot of code (plus all the config definitions).

rayfordshire:
I've decided to not do the `ConfigDef->type` polishing in this PR as this is getting out of hand.  Let's try to get MyVariable into the config system first.

However, please give me your thoughts on this:

> It seems like the ConfigDef->type is more often treated as (signed) int than as uint32_t, e.g. HashElem uses it, and some regex flags (e.g. REG_ICASE;), too. Before I take a deeper dive. Any clue how robust NeoMutt's code base is if int gets replaced by uint32_t or even uint64_t?

If we'll use a typedef for type and the HashElem stays at `int`, then – if we later change the typedef to be `uint64_t` – there will  be no errors/warnings about a conversion from `uint64_t` to `int` and the code might silently fail.  IMHO the correct way is to also update the types everywhere to use the typedef `ConfigType`.  (but that probably means that HashElem now depends on libconfig...)

rayfordshire:
So I've implemented the "set" (no arguments) and "set all" syntax.

Here some follow up questions:

**Question 1:**

The function `parse_set()` in `parse/set.c`, does still contain this check at the beginning:

```c
if (StartupComplete)
{
  if (!mutt_buffer_is_empty(err))
    return MUTT_CMD_ERROR;
}
```

What is this check?  Some sanity check, that the function is called correctly (i.e. should not happen on normal execution)?  If so, an assert with a loud failure is more approriate.
Also, why only do this check only after startup is complete and not always?

**Question 2:**

Is the `if (StartupComplete)` check around `set_dump()` (now part of `command_set_query()`) really necessary?
I've omitted it, and it seems to work: a single `set` in `neomuttrc` will force neomutt to display the config values on startup (until that point).
Not really useful but hey, the user wrote that in its config, so lets do what they wish.


**Question 3:**

What is/was your plan for `test/parse/parse_set.c` (created during your refactoring of the parsing code) and how is it different from what `test/command/set.c` (created in this PR) does?

rayfordshire:
Ok, so up to the questions above, this PR is ready from my side.  Happy reading.

For your convenience here are they again:

1)
> So this is essentially it: All that's left is to replace the remaining call sites (which is mostly removing code as myvar's are no longer treated special).
>
> If you are satisfied with this design I would go ahead and make those changes. The config code itself will not be touched any more.

2)

> **Question 1:**
> 
> The function `parse_set()` in `parse/set.c`, does still contain this check at the beginning:
> 
> ```c
> if (StartupComplete)
> {
>   if (!mutt_buffer_is_empty(err))
>     return MUTT_CMD_ERROR;
> }
> ```
> 
> What is this check? Some sanity check, that the function is called correctly (i.e. should not happen on normal execution)? If so, an assert with a loud failure is more approriate. Also, why only do this check only after startup is complete and not always?
> 
> **Question 2:**
> 
> Is the `if (StartupComplete)` check around `set_dump()` (now part of `command_set_query()`) really necessary? I've omitted it, and it seems to work: a single `set` in `neomuttrc` will force neomutt to display the config values on startup (until that point). Not really useful but hey, the user wrote that in its config, so lets do what they wish.
> 
> **Question 3:**
> 
> What is/was your plan for `test/parse/parse_set.c` (created during your refactoring of the parsing code) and how is it different from what `test/command/set.c` (created in this PR) does?

3) 

> However, please give me your thoughts on this:
> 
> > It seems like the ConfigDef->type is more often treated as (signed) int than as uint32_t, e.g. HashElem uses it, and some regex flags (e.g. REG_ICASE;), too. Before I take a deeper dive. Any clue how robust NeoMutt's code base is if int gets replaced by uint32_t or even uint64_t?
> 
> If we'll use a typedef for type and the HashElem stays at `int`, then – if we later change the typedef to be `uint64_t` – there will be no errors/warnings about a conversion from `uint64_t` to `int` and the code might silently fail. IMHO the correct way is to also update the types everywhere to use the typedef `ConfigType`. (but that probably means that HashElem now depends on libconfig...)

flatcap:
## Promotions

I've promoted a few commits from this PR to `main`

- 06898be2d test: refactor parse tests
- 54bd4dd2a Change macros to use type rather than HashElem
- 109c9b084 Fix value of DT_SUBTYPE_MASK

See also: f700ac08b drop unused config flags

## Small stuff

- Rebased
- clang-format
- Doxygen comments
- Squashed down a few minor fixes
- Moved your test code into `/test/parse` with mine

## Medium stuff

I've regrouped the commits into three sections:

- parse_set() refactoring
  This has been stripped of any `my_var` references.
  **Note:** this isn't all your `parse_set()` changes, just the ones that were easy to split out.

- `my_var` refactoring (largely unchanged)

- Test code
   - Tests for `parse_rc_line()`
   - Tests for MyVariable

I've squashed the tests because only the last commit actually had working tests.
Now, every commit builds and the tests succeed -- important for bisecting, etc.

I'd like to move the `parse_rc_line()` tests into the `parse_set()` group, but some of the tests fail.
If you can find the bits of `parse_set()` that need backporting, I can bundle up this block of work and merge it.

## Big Problem

`set` doesn't seem to work any more.

I've been careful with the rebasing and reordering, frequently comparing my branch to yours.
I'm confident I haven't broken anything.

Test case:

- Config file containing just, `set help = no`
- Run minimally, `neomutt -n -F test.rc`

In `main`s NeoMutt, the help bar is hidden.
In the PR's NeoMutt, the help bar is visible.

flatcap:
```
5720b1457 parse_set() Catch invalid syntax 'set foo+'
eac8fc253 set-command: Add sanity assert checks
d399c3175 set-command: Split parse function into smaller pieces
a67cfdb74 set-command: Check return value of parse_extract_token()
5fd354d97 set-command: Handle "set" and "set all" syntax

4ac2b453c ConfigSet: Add delete method
0deb2ca42 ConfigSet: Add cs_register_variable()
10276c7de Config: Add type MyVariable
9626608e1 ConfigSubSet: Add delete() functions
b8f2b9324 ConfigSet: Add cs_create_variable() and cs_create_variables()
86d12f49c MyVariable: Delete MyVariable on reset if NULL is initial
c309f43f3 MyVariable: reset() do not override the return code
b17b351ed ConfigSet: Do not reset MyVariables on registration
bfa145926 ConfigSet: Do not allow inheritance with MyVariables
7d381e84e MyVariable: Delete MyVariable on reset() unconditionally
46786ff70 MyVariable: Handle decrement operation:
4185f7c15 MyVariable: Handle unset operation
21f9f0171 MyVariable: Handle set operation
7f5943d59 MyVariable: Handle increment operation
2d2b8692c MyVariable: Handle reset operation
067d99757 MyVariable: Do not delete on reset()
ab08bb0c1 MyVariable: "reset all" deletes MyVariables
5ddd61c68 MyVariable: Handle query operation
19662014a MyVariables: parse_ifdef does not need special treatment
977131676 parse/set.c: myvar.h no longer needed
206ef4039 MyVariable: Do not treat MyVariable special on registration

4a1e37c65 test: Add tests for parse_rc_line()
db10a4dff test: Add tests for MyVariable
```

flatcap:
> So this is essentially it

You're getting there :-)

It's a bigger job than I imagined (and we won't mention the feature creep :-)

> All that's left is to replace the remaining call sites
> If you are satisfied with this design I would go ahead and make those changes.

It's looking OK -- It's taking me a lot of time to review.

> ```c
> if (StartupComplete)
> ```
> 
> What is this check?

When NeoMutt starts and it's reading config files, `StartupComplete == false`.
If the parser encounters `set` or `set all` (or `bind`, `version`, etc), it will ignore them.

When all the initialisation is complete and the GUI is about to be created, `StartupComplete == true`.
When `true`, `:set` will open a Pager with the dump of variables.

This isn't a perfect situation.
Ideally, we'd want to distinguish interactive vs non-interactive cases.

If the users does `<enter-command>` + `:set` then **that** should trigger the Pager.
Every other case should be ignored.
That's work, for me, for another day.

> I've omitted it, and it seems to work: a single `set` in `neomuttrc` will
> force neomutt to display the config values on startup

Yes, I saw.
I don't want that to happen, thanks.
It's likely to mess with people's workflow.
We'll just quietly ignore info commands.

> What is/was your plan for `test/parse/parse_set.c`
> (created during your refactoring of the parsing code)

Actually it was created nearly five years ago -- bbd52b528 -- when I refactored the `parse_set()` code.

I'm not sure what my plan it.
It gives fairly good coverage of the code and so does yours.
Between them, there's slightly more coverage than either!

Ideally, we want 100% coverage of the code.
Getting the last few percent usually involves lots of broken config.

There's no real issue if we leave both tests in for now.

> It seems like the ConfigDef->type is more often treated as (signed) int than as uint32_t,

Yeah, I'm not surprised.

> if int gets replaced by uint32_t or even uint64_t?

Feel free to investigate, but I **really** don't think we'll need to do this.
I've just killed four unused flags (f700ac08b), so we have plenty of space.

> IMHO the correct way is to also update the types everywhere to use the typedef `ConfigType`.

Yes, that sounds like a good idea.

However, I'm not so sure about my suggestion of splitting the `type` and `flags`.
It's likely we'd need both at some point, and we'd only have `HashElem.type` to store them in.

> (but that probably means that HashElem now depends on libconfig...)

Urgh! No thanks!
My latest PR, #3767, breaks `libmutt`'s dependency on `libconfig`.

Perhaps we make `HashElem.type` into a `uint32_t`.
That's more work for no real gain, though.

flatcap:
Just a heads up...
When you're done, I'm likely to do a mass rename.
I'm tired of reading `MyVariable` (five syllables), when `MyVar` (two syllables) would do :-)

rayfordshire:
> ## Medium stuff
>
> I'd like to move the parse_rc_line() tests into the parse_set() group, but some of the tests fail.
> If you can find the bits of parse_set() that need backporting, I can bundle up this block of work and merge it.

They ain't broken.  The file `test/pattern/dummy.c` states in `myvar_get` that *every* my_variable exists, even the crazy ones named `&Cherry`, so the query succeed which lets the tests fail.  (Side note: `&` is allowed in a variable name, `set my_fo&o = "bar"` works).  I don't know which code relies on `test/pattern/dummy.c` to always return `"hello"` (the value of every my_variable).

It seems like no code does depend on `myvar_get` which seems only there so that things compile.  I've changed that to always return `NULL` (i.e. variable does not exist).  Let's see if that fixes it also for you.



---

> ## Big Problem
> 
> `set` doesn't seem to work any more.

Fixed.  The special handling of `set foo` for a bool/quad `foo` was also done by `command_set_set`.  Probably forgot to eliminate it when I moved this distinction to `parse_set` (so `command_set_set(name, value)` always means `set name = value` and has no other interpretation).


---

>>>    if (StartupComplete)
>
>>    What is this check?
>
> [general text about interactive vs non-interactive use case]

I think you misunderstood me here, I was not talking about `if (StartupComplete)` in general but about this *specific* test (function `parse_set()` in `parse/set.c`):
```c
if (StartupComplete)
{
  if (!mutt_buffer_is_empty(err))
    return MUTT_CMD_ERROR;
}
```
What does it bring to the table in the interactive case vs non-interactive case? As stated above it looks to me like an *assert* check and I wonder whether that outer `if (StartupComplete)` should be removed so that check is always done, i.e.
```c
  if (!mutt_buffer_is_empty(err))
    return MUTT_CMD_ERROR;
```

rayfordshire:
So, I also purged all references of myvar.h resulting in the glorious commit titled

   *Purge myvar.c and myvar.h*

flatcap:
> Let's see if that fixes it also for you.

Yep!

> I think you misunderstood me here,

Yes, sorry about that.
That code shouldn't be there at all.

Previously, it read:

```c
  if (StartupComplete)
  {
    if (set_dump(buf, s, data, err) == MUTT_CMD_SUCCESS)
      return MUTT_CMD_SUCCESS;
    if (!mutt_buffer_is_empty(err))
      return MUTT_CMD_ERROR;
  }
```

There were three possible outcomes from `set_dump()`.
- Success -> return success
- Failure (`err` is set) ->  return failure
- No action

When you refactored `set_dump()` into your new code, its error-checking got left behind.
I've removed the whole `if (StartupComplete)`.

flatcap:
@whitney-cumber Welcome to the party :-)

I've tweaked your comments slightly to use Doxygen's `@pre` (pre-condition marker).

https://github.com/rayfordshire/neomutt/blob/dfc8f4d7e58bd7046cffb047c315253ab5e3abc5/parse/set.c#L89-L102

It'll look something like:
image

config/set.c
```diff
@@ -285,7 +296,79 @@ bool cs_register_variables(const struct ConfigSet *cs, struct ConfigDef vars[],
...
+  cdef_copy->validator = cdef->validator;
+  cdef_copy->docs = mutt_str_dup(cdef->name);
+  cdef_copy->var = cdef->var;
+
+  struct HashElem *he = cs_register_variable(cs, cdef_copy, err);
+  if (!he)
+  {
+    FREE(&cdef_copy->name);
+    FREE(&cdef_copy->docs);
+    FREE(&cdef);
```
flatcap, line 342: `cdef_copy` perhaps?

flatcap (commented):
this isn't a full review, just something I noticed

flatcap:

flatcap:
More updates:
- `parse_set()` refactoring promoted to `main`!
- Rebased over main
- Renamed MyVariable to MyVar
- Separated ConfigSet/ConfigSubset from MyVar changes

rayfordshire:
* fix typo: `cdef` -> `cdef_copy`
* Combined commits:

  * MyVar: Delete MyVar on reset if NULL is initial
  * ConfigSet: Do not reset MyVars on registration
  * MyVar: Delete MyVar on reset() unconditionally
  * MyVar: Do not delete on reset()
  * MyVar: Do not treat MyVar special on registration

  which results in the empty commit, i.e. drop them.

---

> ```
> /** [...]
>  * This function does not know anything about how the internal representation
>  * of the type must be handled.  Thus, the fields 'initial', 'data', 'var' must
>  * be copyable.  If they need allocation, then the caller must set them after
>  * the variable is created, e.g. with cs_he_initial_set(), cs_he_native_set.
>  */
> struct HashElem *cs_create_variable(const struct ConfigSet *cs,
>                                     struct ConfigDef *cdef, struct Buffer *err)
> ```
>
> Wouldn't it make more sense for the caller to offer an allocated ConfigDef.
> Then this function would take ownership.

Thought about that and I think the problem here is the leaky abstraction and shared/unclear responsibilities:

IMHO ConfigDef is a *private* implementation detail, however, it is also used as an easy way to register config variables.  This was more or less fine as long as all variables were registered with a `ConfigDef` which lives forever.

As for whose responsibility it is to clean up the memory.  The status quo (i.e. without "dynamic" ConfigDefs from this PR) for who owns the memory is ("Config" means the config system, "caller" means the code calling the register function, "shared" means both)

* ConfigDef struct itself: *caller*
* name: *shared*, depending on whether its a synonym (Config) or not (caller)
* type: freeing not needed
* initial: *shared*, depending on whether `set_initial()` was called or not
* data: *caller*
* validator: freeing not needed
* docs: *caller*
* var: *config*

This was fine as long as all the caller instances lived forever, now we have "dynamic" ConfigDefs, complicating things.

---

## An idea

Given the structure of the config system, it almost screams for OOP (sadly in a non-OOP language).

The problem I have with a generic "create ConfigDef" function is, that for a `ConfigSet` it cannot now which fields needs freeing and which don't, e.g. `initial` or `var` depend on the type.  Likewise for `data` which can be anything.

Therefore, freeing a ConfigDef should be the responsibility of the *type*.
Each type gets a new function for deleting an entire `ConfigDef`, call it `destroy_configdef()`, this function is responsible to clean up whatever the type has allocated in the `ConfigDef` struct, e.g. `initial`, `var`, ...
(A type already has a function `destroy()` (better call it `destroy_var()`), which deletes `ConfigDef->var`. It is used in conjunction with `DT_INHERITANCE`)

This leads to the second new function: Each type gets a `create_configdef()` function, which the user (i.e. everything outside of the config system) must use to create a new `ConfigDef`.

If the `create_configdef()` are the only function creating `ConfigDef`, then encoding deletion patterns into type flags is almost unncessary, since the type knows exactly what it had allocated and thus needs deleting.

(NB: `create_configdef()` and `destroy_configdef()` are better known as constructor and destructor of `ConfigDef<T>` where `T` is the type.)

In this scenario the config system is solely responsible for cleaning up of everything.

## How could the forever config variables fit in?

1) They could all go through the same registration process as everything eles.  This requires rewriting the config code to us the functions instead of a generic ConfigDef global array but that's just a minor detail.
Since `create_configdef()` would create a copy of `ConfigDef->name` and `ConfigDef->docs`, we would end up with additional memory usage. A dump of `neomutt -D -O` is currently about ~30KiB.

2) Proceed with the current model having shared responsibilities for different parts of the structure, and (partially) encoding the responsibilities in the type.  Allowing to pass the global forever `ConfigDef` to the register function.  (I don't like that approach.)

3) Some midway between 1) and 2).  Have a special function `register_forever_configdef()` (for each type of course), which does not create a copy of `ConfigDef->name` and `ConfigDef->docs` and encodes that in a flag.  All other properties are handled by the Config System (in particular `ConfigDef->initial`).  This should avoid the majority of of memory consumption at the cost of complexity.

---

Having said the above: We currently need dynamic variables only for the MyVar type, so we could get away with this:
```
// in config/myvar.h
struct ConfigDef *create_new_config_definition(const char *name, const char *initial, struct Buffer *err)
{ ... }
struct HashElem *register_new_config_definition(ConfigSet *cs, const char *name, const char *initial, struct Buffer *err)
{
	struct ConfigDef *cdef = create_new_config_definition(name, initial, err)
	if (!cdef)
		return NULL;
	return cs_register_variable(cs, cdef, err);
}
```

The former function creates a `ConfigDef` suitable for a call to `cs_register_variable()`.  The implementation of the cleanup function stays as currently implemented in this PR, except `docs` is not deleted anymore.

flatcap:
GitHub says it's closed, but it's entirely merged.
Thanks for you patience; I've had a lot of other commitments.

I made a few minor changes:

- I moved the helper `cs_subset_myvar()` into the test code
  It was only used there and it had `assert`s in it.
  The other helpers are all called from places where the config name is fixed.

- I dropped `cs_create_variables()` (plural).
  It's not used and unlikely to be used.

- I dropped the validator code in the my_var type.
  It's not used and unlikely to be used.

- I added some tests to cover a few degenerate cases, so the coverage is 100%

PR #3754, opened 2023-03-05
Document command unbind

whitney-cumber:
* **What does this PR do?**

Documents the command unbind which was only mentioned in a note.


flatcap (approved):
Lovely.
Thanks!

PR #3761, opened 2023-03-09
editorconfig: fix charset entry

brunnre8:
According to the spec, utf-8 must be spelled "utf-8" https://spec.editorconfig.org/#supported-pairs

https://spec.editorconfig.org/#supported-pairs


flatcap (approved):
Thanks!

PR #3762, opened 2023-03-10
Keep mailbox in memory

sassmann:
* **What does this PR do?**
add keep_mailbox option

keep_mailbox prevents MailboxViews from being free'd when switching mailboxes. Once a mailbox has been opened it's content is kept in memory. This increases memory consumption, but makes switching between mailboxes lightning fast.
 The default is off to conserve memory.

This is a RFC, feedback is welcome. Please give it a try.

To enable add 
set keep_mailbox=yes
to neomuttrc

### DIFF
--- a/mview.c
+++ b/mview.c
@@ -250,7 +250,7 @@ static void update_tables(struct MailboxView *mv)
   m->msg_flagged = 0;
   padding = mx_msg_padding_size(m);
   const bool c_maildir_trash = cs_subset_bool(NeoMutt->sub, "maildir_trash");
-  for (i = 0, j = 0; i < m->msg_count; i++)
+  for (i = 0, j = 0; i < m->email_max; i++)
   {
     if (!m->emails[i])
       break;
--- a/mview.c
+++ b/mview.c
@@ -54,10 +54,14 @@ void mview_free(struct MailboxView **ptr)
     return;
 
   struct MailboxView *mv = *ptr, *np, *tmp;
+  const bool c_keep_mailbox = cs_subset_bool(NeoMutt->sub, "keep_mailbox");
 
   struct EventMview ev_m = { mv };
   mutt_debug(LL_NOTIFY, "NT_MVIEW_DELETE: %p\n", (void *) mv);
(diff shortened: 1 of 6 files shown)


mview.c
```diff
@@ -237,7 +247,7 @@ static void update_tables(struct MailboxView *mv)
   m->changed = false;
   m->msg_flagged = 0;
   padding = mx_msg_padding_size(m);
-  for (i = 0, j = 0; i < m->msg_count; i++)
+  for (i = 0, j = 0; i < m->email_max; i++)
```
gahr, line 250: Why this change?

gahr:
Interesting indeed, although I don't quite appreciate the consequences... I'll need to think it through

mview.c
sassmann: I've explained that in the commit message, here
https://github.com/neomutt/neomutt/pull/3762/commits/7eef19ede1a19751840b468cc94181a5ede109c6

sassmann:
> Interesting indeed, although I don't quite appreciate the consequences... I'll need to think it through

That's why I made it a config option.

mview.c
gahr: Thanks - yep this is also what #3765 is about. I'm not sure we ever want msg_count to exceed the number of emails, so I'm not yet sure what the solution for *that* is going to be.

gahr:
> > Interesting indeed, although I don't quite appreciate the consequences... I'll need to think it through
> 
> That's why I made it a config option.

Sure, but we don't want it to have unintended consequences, even if it's configurable and off by default.

vuori:
Happened to notice this and I think the feature would great to have, since at least the Maildir backend has pretty good code for updating an already-open Mailbox too. But having stared at the relevant bits lately, a comprehensive solution to #3765 aka `msg_count` hermeneutics seems needed first.

Just iterating up to `m->email_max` with a `if (!e)` guard seems surprisingly effective though. IIRC there are only a couple of places where you need to go directly to the end of `m->emails`.

flatcap:

flatcap:
Archived to: https://github.com/neomutt/neomutt-old/tree/keep_mailbox

PR #3767, opened 2023-03-17
Eliminate config usage in libmutt

flatcap:
`libmutt` shouldn't have any dependencies, but unfortunately it contains three sets of functions that rely on config variables.

These commits update the functions to take extra parameters, use flags or simply move the functions elsewhere. 

### 574579eb8 drop config usage from mutt/state.c

Factor out config usage from:

- `state_mark_attach()`
- `state_mark_protected_header()`

Add a new flag, STATE_PAGER, instead.

### 7b40b874e drop config mutt/charset.c

Factor out config usage from:

- `mutt_ch_convert_nonmime_string()`
- `mutt_ch_get_default_charset()`

Pass in the required strings instead.

### c04a4bd63 move tmp functions to libcore

Move the three temporary-file functions to libcore:

- `mutt_buffer_mktemp_full()`
- `mutt_file_mkstemp_full()`
- `mutt_mktemp_full()`

This includes their macro wrappers:

- `mutt_buffer_mktemp()`
- `mutt_buffer_mktemp_pfx_sfx()`
- `mutt_file_mkstemp()`
- `mutt_mktemp()`
- `mutt_mktemp_pfx_sfx()`

This makes them available to the unit tests, without adding dependencies to libmutt.

### 855dc85ce test: add $tmp_dir to common NeoMutt init

Automatically register `$tmp_dir` when creating a NeoMutt for the unit tests.

### DIFF
--- a/Makefile.autosetup
+++ b/Makefile.autosetup
@@ -326,7 +326,7 @@ $(PWD)/convert:
 # libcore
 LIBCORE=	libcore.a
 LIBCOREOBJS=	core/account.o core/command.o core/dispatcher.o core/mailbox.o \
-		core/neomutt.o
+		core/neomutt.o core/tmp.o
 CLEANFILES+=	$(LIBCORE) $(LIBCOREOBJS)
 ALLOBJS+=	$(LIBCOREOBJS)
 
--- a/attach/cid.c
+++ b/attach/cid.c
@@ -32,11 +32,11 @@
 #include <stdio.h>
 #include <string.h>
 #include "email/lib.h"
+#include "core/tmp.h"
 #include "cid.h"
 #include "attach.h"
 #include "mailcap.h"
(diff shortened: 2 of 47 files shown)


flatcap:
The first commit is the hardest to trace through.

The two functions `state_mark_attach()` and `state_mark_protected_header()` (**Red**) are passed a `struct State`.
If the flags include `STATE_DISPLAY`, **and** the Pager is not set to some external viewer, then they insert markers into the text.

There are nearly 20 places where a `struct State` is created, but only two of them (**Yellow**) set `STATE_DISPLAY`.
These are the only sources where marker functions could be active.

The **Purple** functions pass the `State`, but do not affect the flags.

k4
(**click on the image to expand**)

gahr (approved):
Wonderful!

Issue #3769, opened 2023-03-18
Add support for IMAP ID extension


timothyqiu:
## What would you like to be able to do?

Use neomutt with [NetEase Mail](https://mail.163.com/) via IMAP. NetEase is one of the biggest email service providers in China.

## What needs to change in NeoMutt?

NetEase [requires](https://help.mail.163.com/faqDetail.do?code=d7a5dc8471cd0c0e8b4b8f4f8e49998b374173cfe9171305fa1ce630d7f67ac211b1978002df8b23) IMAP clients to identify themselves via ID command ([RFC2971](https://datatracker.ietf.org/doc/rfc2971/)).

## Have you tried any alternative methods?

I'm not aware of any ways to send arbitary IMAP commands.

flatcap:
Hmm... another one.

A Mutt user, Jia Qingtong, [proposed a patch](https://gitlab.com/muttmua/mutt/-/merge_requests/155) for this a year ago.
There was some discussion, but nobody seemed actively against the patch.
Unfortunately, Mutt's maintainer has taken a step back so no new features are getting merged.

It's a small feature and it would be off by default.
Let's see what **our** users think :-)

PS  It should be a simple job to port the patch to NeoMutt.

gahr:
I and @flatcap had a brief discussion on IRC about this (btw, you all are welcome to join us on Libera/#neomutt).

First, a side note: I think NetEase is misbehaving here. The RFC [specifically states](https://www.rfc-editor.org/rfc/rfc2971.html#section-3) that "Implementations MUST NOT make operational changes based on the data sent as part of the ID command or response".

Having said that, we agree that it would be nice to be able to tell NeoMutt to send ID info, but we wouldn't go as far as allowing configuring each and every possible field. We'd just expose a `imap_send_id` boolean config (name TBD) and hard-code a couple of fields such as "NeoMutt" and the current version.

First off, though, we'd like to know *what* would make NetEase happy: @timothyqiu do you know what's the minimal set of fields that would turn off that warning for you?

timothyqiu:
I've tried to connect manually in Python. NetEase accepts any fields in the parameter list, as long as the list is not empty.

These works:

```
ID ("name" "Client" "version" "1.2.3")
ID ("name" "Client")
ID ("foo" "bar")
ID ("" "")
```

These won't work:

```
ID NIL
ID ()
```

PR #3779, opened 2023-03-25
Add support for IMAP ID extension

timothyqiu:
* **What does this PR do?**
Adds a `imap_send_id` boolean config, defaults to false. When set, an `ID` command will be sent to the server when logging in. This provides the IMAP client name and version to the server.

    Example log output:

    > [2023-03-25 10:49:45]<2> mutt_socket_write_d() 4> a0004 ID ("name" "NeoMutt" "version" "20230322-18-a8bdba-dirty")

#3769

### DIFF
--- a/imap/config.c
+++ b/imap/config.c
@@ -126,6 +126,9 @@ static struct ConfigDef ImapVars[] = {
   { "imap_qresync", DT_BOOL, false, 0, NULL,
     "(imap) Enable the QRESYNC extension"
   },
+  { "imap_send_id", DT_BOOL, false, 0, NULL,
+    "(imap) Send ID command when logging in"
+  },
   { "imap_user", DT_STRING|DT_SENSITIVE, 0, 0, NULL,
     "(imap) Username for the IMAP server"
   },
--- a/imap/imap.c
+++ b/imap/imap.c
@@ -1902,6 +1902,14 @@ int imap_login(struct ImapAccountData *adata)
     }
 #endif
 
+    /* enable RFC2971, if the server supports that */
+    const bool c_imap_send_id = cs_subset_bool(NeoMutt->sub, "imap_send_id");
+    if (c_imap_send_id && (adata->capabilities & IMAP_CAP_ID))
+    {
+      imap_exec(adata, "ID (\"name\" \"NeoMutt\" \"version\" \"" PACKAGE_VERSION "\")",
+                IMAP_CMD_QUEUE);
+    }
+
     /* enable RFC6855, if the server supports that */
     const bool c_imap_rfc5161 = cs_subset_bool(NeoMutt->sub, "imap_rfc5161");
     if (c_imap_rfc5161 && (adata->capabilities & IMAP_CAP_ENABLE))
--- a/docs/config.c
+++ b/docs/config.c
@@ -1990,6 +1990,14 @@
 ** .pp
 ** This variable defaults to your user name on the local machine.
 */
+
+{ "imap_send_id", DT_BOOL, false },
+/*
+** .pp
+** When \fIset\fP, NeoMutt will send an IMAP ID command (RFC2971) to the
+** server when logging in if advertised by the server. This command provides
+** information about the IMAP client, such as "NeoMutt" and the current version.
+*/
 #endif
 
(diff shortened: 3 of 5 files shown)


rayfordshire:
Having a value hardcoded makes me feel uneasy,  similar to the `User-Agent:` header (which is nowadays discouraged) and the `Message-ID:` header.
I think we should either
* add an option, e.g., `imap_send_id_argument`, to specify the complete argument to "ID" (like `my_hdr` does for User-Agent:) or
* make the bool option into a string one which is is the argument to the "ID" IMAP command

Example:
```
set imap_send_id = 'NIL'
set imap_send_id = '("name" "Foo" "version" "12.34")'
```

these would send the commands

```
ID NIL
ID ("name" "Foo" "version" "123")
```

In the latter case (only one option): If unset (the default), no ID will be sent; if set to `neomutt-name-and-version`, then the current proposed value is used: `("name" "NeoMutt" "version" 12345678")`.

imap/config.c
```diff
@@ -131,6 +131,9 @@ static struct ConfigDef ImapVars[] = {
   },
 
   { "imap_servernoise", DT_SYNONYM, IP "imap_server_noise", IP "2021-02-11" },
+  { "imap_send_id", DT_BOOL, false, 0, NULL,
+    "(imap) Send ID command when logging in"
+  },
```
rayfordshire, line 136: Move from the "synonym" section to the "non-synonym" section above (i.e. ~5 lines up).  They are both alphabetically sorted.

imap/imap.c
```diff
@@ -1902,6 +1903,15 @@ int imap_login(struct ImapAccountData *adata)
     }
 #endif
 
+    /* enable RFC2971, if the server supports that */
+    const bool c_imap_send_id = cs_subset_bool(NeoMutt->sub, "imap_send_id");
+    if (c_imap_send_id && (adata->capabilities & IMAP_CAP_ID)) {
+      char id_cmd_msg_buf[PATH_MAX] = { 0 };
+      snprintf(id_cmd_msg_buf, sizeof(id_cmd_msg_buf), "ID (\"name\" "
+              "\"%s\" \"version\" \"%s%s\")", "NeoMutt", PACKAGE_VERSION, GitVer);
```
rayfordshire, line 1911: Is the git version really necessary? Wouldn't `PACKAGE_VERSION` suffice?

gahr:
> Having a value hardcoded makes me feel uneasy

See https://github.com/neomutt/neomutt/issues/3769#issuecomment-1481403207

flatcap:
Looks good.  Thanks!

As the ID string is known at build time, we _could_ go one step further:

```diff
diff --git i/imap/imap.c w/imap/imap.c
index 5dcc3e687..0a8fd2abc 100644
--- i/imap/imap.c
+++ w/imap/imap.c
@@ -1906,10 +1906,7 @@ int imap_login(struct ImapAccountData *adata)
     const bool c_imap_send_id = cs_subset_bool(NeoMutt->sub, "imap_send_id");
     if (c_imap_send_id && (adata->capabilities & IMAP_CAP_ID))
     {
-      char id_cmd_msg_buf[PATH_MAX] = { 0 };
-      snprintf(id_cmd_msg_buf, sizeof(id_cmd_msg_buf),
-               "ID (\"name\" \"%s\" \"version\" \"%s\")", "NeoMutt", PACKAGE_VERSION);
-      imap_exec(adata, id_cmd_msg_buf, IMAP_CMD_QUEUE);
+      imap_exec(adata, "ID (\"name\" \"NeoMutt\" \"version\" \"" PACKAGE_VERSION "\")", IMAP_CMD_QUEUE);
     }
 
     /* enable RFC6855, if the server supports that */
```

@gahr Any thoughts?

gahr:
LGTM, + what flatcap suggested

flatcap (approved):

flatcap:
Thanks!

If you've got any other ideas, _do_ let us know.
Either [start a discussion](https://github.com/neomutt/neomutt/discussions), or join us on IRC: `#neomutt` on `irc.libera.chat` ([web client](https://kiwiirc.com/nextclient/irc.libera.chat/#neomutt))

PR #3782, opened 2023-03-25
notmuch: #if out get_nm_config_file() where unused

nabijaczleweli:
According to Debian clang version 17.0.0 (++20230318071855+864a2b25beac-1\~exp1~20230318072020.597):
```
notmuch/db.c:84:20: warning: unused function 'get_nm_config_file' [-Wunused-function]
static const char *get_nm_config_file(void)
                   ^
```

There's one user, just below, guarded by the same `#if`


flatcap (approved):
Thanks!

PR #3787, opened 2023-03-26
build: tidy Makefiles, fix race condition

flatcap:
- d3403d150 build: use `$(INSTALL_DATA)`
- 04b61e5e2 build: create `$(INSTALL_DIR)`
- 2c67e8161 build: create data dir separately

Tested:
- A wide variety of builds
- Building out-of-tree
- Installing, uninstalling


gahr (approved):
Not too happy about contrib/ relying on a data/ target, but let’s not be too picky :)

PR #3788, opened 2023-03-26
trivial: replace strcmp() with mutt_str_equal()

flatcap:
Replace uses of `strcmp()` with the more intuitive `mutt_str_equal()`.

The changes are split into many commits.
Each commit has one type of change, making reviewing simpler.
They can all be squashed before merging.

- 1d2559ade Form: `strcmp(a,b) == 0` -- simple parameters
- 5353764e4 Form: `strcmp(a,b) == 0` -- parameters contain expressions
- fe62d3578 Form: `strcmp(a,b) != 0` -- parameter contain expressions
- f70117476 Form: `strcmp(a,b)` -- Five places where `mutt_str_cmp()` was needed
- 50917d57c test: Form: `strcmp(a,b) == 0` -- simple parameters
- 2e0593fba test: Form: `strcmp(a,b) == 0` -- parameters contain expressions
- c619352ea test: Form: `!strcmp(a,b)` -- parameters contain expressions
- 102bf5c1c test: Form: `strcmp(a,b)` -- parameters contain expressions
- 6d44cc716 test: Form: `strcmp(a,b)` -- Two places where `mutt_str_cmp()` was needed

### DIFF
--- a/test/base64/mutt_b64_decode.c
+++ b/test/base64/mutt_b64_decode.c
@@ -51,11 +51,7 @@ void test_mutt_b64_decode(void)
       TEST_MSG("Actual  : %zu", len);
     }
     buffer[len] = '\0';
-    if (!TEST_CHECK(mutt_str_equal(buffer, clear)))
-    {
-      TEST_MSG("Expected: %s", clear);
-      TEST_MSG("Actual  : %s", buffer);
-    }
+    TEST_CHECK_STR_EQ(buffer, clear);
   }
 
   {
--- a/test/base64/mutt_b64_encode.c
+++ b/test/base64/mutt_b64_encode.c
@@ -25,6 +25,7 @@
 #include "acutest.h"
 #include <string.h>
 #include "mutt/lib.h"
+#include "test_common.h"
 
 static const char clear[] = "Hello";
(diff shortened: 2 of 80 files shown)


gahr:
Nice! How about using TEST_CHECK_STR_EQ in test code?

flatcap:
Hmm... it turns out that that's not so simple.
Many of the callers check the return value from `TEST_CHECK()`, but `TEST_CHECK_STR_EQ()` doesn't have one.

I've tweaked things a bit.

Now, `test_check_str_eq()`:
- Is wrapped by a macro, so that we get the right `__FILE__`, `__LINE__`
- Uses the acutest machinery
- Returns the value of the condition

Here's a sample program: https://gist.github.com/flatcap/6ab6cb0acb36555209a9220efddb51cb

I'll work through the others later.

gahr:
I think `TEST_CHECK_STR_EQ` doesn't return anything because it expands to both checking and reporting.. If it fails, you're not supposed to continue

flatcap:
Some of the config tests, were they to fail, print out extra debugging info.
Probably all overkill, but it'll take time to weed it out.

PR #3792, opened 2023-03-28
Make $pager_index_lines use a fixed number of rows

auouymous:
`set pager_index_lines = 5` will use less than five rows when fewer than five messages. `set pager_index_lines = -5` always uses five rows, even if some are blank.

> &lt;flatcap&gt; start with no config, and we can think a bit longer about if we want to change the default behaviour

### DIFF
--- a/docs/config.c
+++ b/docs/config.c
@@ -2895,8 +2895,7 @@
 ** giving the reader the context of a few messages before and after the
 ** message.  This is useful, for example, to determine how many messages
 ** remain to be read in the current thread.  A value of 0 results in no index
-** being shown.  If the number of messages in the current folder is less than
-** $$pager_index_lines, then the index will only use as many lines as it needs.
+** being shown.
 */
 
 { "pager_read_delay", DT_NUMBER, 0 },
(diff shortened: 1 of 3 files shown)


auouymous:
I don't like changing default behavior or removing features someone might depend on. However, I was able to live without a fixed size index because zero inbox is hard but also because neomutt has had a nice bug over the past year where the index would not automatically shrink or grow as messages were deleted or added. It would only resize when I did something that required exiting the pager. That bug will need to be fixed if the dynamic behavior is kept.

As for a separate config vs negative value, I've heard there are a lot of configs and using negative keeps it from being a lot plus one. But that only works if dynamic behavior is never removed, and makes it hard to change the default later on. A `dynamic_pager_index_lines` that defaults to on might be better if the behavior has a chance of being removed later, or its default changed.

flatcap:
"De-optioned" to make it the default behaviour.
If anyone complains, who hasn't seen this PR, I _might_ consider introducing a config variable.

flatcap (approved):

PR #3798, opened 2023-04-02
libidn 1.x deprecation

vuori:
2.x has been the recommended version for >5 years, deprecate 1.x.

* **What does this PR do?**

Switches `configure` defaults such that libidn 2.x is now the default and libidn 1.x is the option. Since [upstream](https://www.gnu.org/software/libidn/#libidn2) states that 2.x is the recommended version and 2.0 appeared in 2017, this change seems due.

If no issues/complaints appear in the near future, I'll submit another PR that drops 1.x.

None.

No user visible changes. Docs updated. No new test failures.

None. Discussed on irc.


gahr (approved):

gahr:
I don't know how the integration with google fuzz works, namely, where does the github action takes the configure parameters.

flatcap:
> I don't know how the integration with google fuzz works

I've made a note to document the process...

Roughly, the [projects/neomutt](https://github.com/google/oss-fuzz/tree/master/projects/neomutt) dir in the [google/oss-fuzz](https://github.com/google/oss-fuzz) repo contains a `Dockerfile` and a `build.sh`

Something's changed behind-the-scenes with the default packages, so we need to add `pkgconf`

I've raise a PR here: https://github.com/google/oss-fuzz/pull/10011

flatcap:
- Rebase
- Move `--idn` in `auto.def` to deprecated
- Strip out idn v1 code

I'll merge this in a few days, after the bug-fix release, just in case.

Thanks!

PR #3799, opened 2023-04-03
stuct Address: replace char * uses with struct Buffer

navi-desu:
This marks the beginning of the adventure to replace the uses of raw C pointers with buffers.

* **What does this PR do?**
Replaces the char * in the struct Address with struct Buffer *


address/address.c
```diff
@@ -727,9 +726,10 @@ struct Address *mutt_addr_copy(const struct Address *addr)
     return NULL;
 
   struct Address *p = mutt_addr_new();
-
-  p->personal = mutt_str_dup(addr->personal);
-  p->mailbox = mutt_str_dup(addr->mailbox);
+  p->personal = addr->personal ? mutt_buffer_new(NULL) : NULL;
+  mutt_buffer_copy(p->personal, addr->personal);
+  p->mailbox = addr->mailbox ? mutt_buffer_new(NULL) : NULL;
+  mutt_buffer_copy(p->mailbox, addr->mailbox);
```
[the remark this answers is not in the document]
navi-desu: Indeed

mutt/buffer.c
```diff
@@ -388,6 +416,25 @@ size_t buf_strcpy_n(struct Buffer *buf, const char *s, size_t len)
...
+{
+  if (!buf)
+    return;
+
+  char *tmp = buf_strdup(buf);
+
+  mutt_str_dequote_comment(tmp);
+
+  buf_strcpy(buf, tmp);
+  FREE(&tmp);
```
flatcap, line 435: I don't think we need to dupe the string, here.
`mutt_str_dequote_comment()` only ever shortens the string, so we should be safe.
After calling it, we should call `buf_fix_dptr()` to adjust the end marker.

email/rfc2047.c
```diff
@@ -762,9 +762,9 @@ void rfc2047_encode_addrlist(struct AddressList *al, const char *tag)
   TAILQ_FOREACH(a, al, entries)
   {
     if (a->personal)
-      rfc2047_encode(&a->personal, AddressSpecials, col, c_send_charset);
+      rfc2047_encode(&a->personal->data, AddressSpecials, col, c_send_charset);
```
flatcap, line 765: this...

email/rfc2047.c
```diff
@@ -762,9 +762,9 @@ void rfc2047_encode_addrlist(struct AddressList *al, const char *tag)
...
     else if (a->group && a->mailbox)
-      rfc2047_encode(&a->mailbox, AddressSpecials, col, c_send_charset);
+      rfc2047_encode(&a->mailbox->data, AddressSpecials, col, c_send_charset);
```
flatcap, line 767: this...

email/rfc2047.c
```diff
@@ -781,13 +781,13 @@ void rfc2047_decode_addrlist(struct AddressList *al)
   struct Address *a = NULL;
   TAILQ_FOREACH(a, al, entries)
   {
-    if (a->personal && ((strstr(a->personal, "=?")) || assumed))
+    if (a->personal && ((buf_find_string(a->personal, "=?")) || assumed))
     {
-      rfc2047_decode(&a->personal);
+      rfc2047_decode(&a->personal->data);
```
flatcap, line 786: this...

email/rfc2047.c
```diff
@@ -781,13 +781,13 @@ void rfc2047_decode_addrlist(struct AddressList *al)
...
     }
-    else if (a->group && a->mailbox && strstr(a->mailbox, "=?"))
+    else if (a->group && a->mailbox && buf_find_string(a->mailbox, "=?"))
     {
-      rfc2047_decode(&a->mailbox);
+      rfc2047_decode(&a->mailbox->data);
```
flatcap, line 790: and this are actually dangerous, but it's subtle.

`rfc2047_encode()` and `rfc2047_decode()` replace the original string with a newly allocated one.
This means that `Buffer.data` is changed, `Buffer.dsize` is wrong and `Buffer.dptr` is now dangling!

It's a bit messy, but I think the only safe thing to do is do dupe the string, encode/decode it, then overwrite the original Buffer.

flatcap (changes requested):
Sorry it's taken a while to work through this.
I've done a bit of light tidying as I reviewed.

There are three things I'd like fixed, please:
- Simplify `buf_dequote_comment()`
- Make the `rfc2047_encode()` and `rfc2047_decode()` calls safe
- Add tests for the 12 new Buffer functions that you've introduced

Thanks!

flatcap:
ping?

flatcap (approved):
Nice work.  Thanks!

PR #3803, opened 2023-04-06
progress: update ncurses only when there is actual progress

sesse:
If we're not changing the progress bar, there's no need to ask ncurses for an update.

This means that we won't immediately resize on SIGWINCH, but wait until the next time the progress bar updates. That should be fine, though.

Speeds up opening a large Maildir by about 3.5%.

* **What does this PR do?**

     for that)

### DIFF
--- a/progress/progress.c
+++ b/progress/progress.c
@@ -75,7 +75,10 @@ bool progress_update(struct Progress *progress, size_t pos, int percent)
   // Decloak an opaque pointer
   struct MuttWindow *win = (struct MuttWindow *) progress;
   const bool updated = progress_window_update(win, pos, percent);
-  window_redraw(win);
+  if (updated)
+  {
+    window_redraw(win);
+  }
   return updated;
 }
 


progress/progress.c
```diff
@@ -75,7 +75,9 @@ bool progress_update(struct Progress *progress, size_t pos, int percent)
   // Decloak an opaque pointer
   struct MuttWindow *win = (struct MuttWindow *) progress;
   const bool updated = progress_window_update(win, pos, percent);
-  window_redraw(win);
+  if (updated) {
+    window_redraw(win);
+  }
```
gahr, line 80:
```suggestion
  if (updated)
  {
    window_redraw(win);
  }
```

sesse: Done.

flatcap:
Yes, that is good.

`window_redraw()` will check whether windows need resizing, recalculating or repainting, then execute the minimum required.
But... if we **know** the window hasn't changed, it makes sense to shortcut all of that.

PR #3804, opened 2023-04-06
email: call ftello() less

sesse:
Every call to ftello() translates into an lseek() system call, because glibc cannot know whether someone moved the underlying file descriptor or not. Keep track of the position ourselves instead of checking anew for every header line that we read.

Speeds up opening a large Maildir by about 19% (!).

* **What does this PR do?**

     for that)

### DIFF
--- a/email/parse.c
+++ b/email/parse.c
@@ -1166,7 +1166,7 @@ struct Envelope *mutt_rfc822_read_header(FILE *fp, struct Email *e, bool user_hd
 
   struct Envelope *env = mutt_env_new();
   char *p = NULL;
-  LOFF_T loc;
+  LOFF_T loc = e ? e->offset : ftello(fp);
   struct Buffer *line = buf_pool_get();
 
   if (e)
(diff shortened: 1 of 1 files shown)


email/parse.h
```diff
@@ -46,6 +46,6 @@ struct Body *    mutt_read_mime_header    (FILE *fp, bool digest);
 int              mutt_rfc822_parse_line   (struct Envelope *env, struct Email *e, const char *name, const char *body, bool user_hdrs, bool weed, bool do_2047);
 struct Body *    mutt_rfc822_parse_message(FILE *fp, struct Body *parent);
 struct Envelope *mutt_rfc822_read_header  (FILE *fp, struct Email *e, bool user_hdrs, bool weed);
-char *           mutt_rfc822_read_line    (FILE *fp, char *line, size_t *linelen);
+char *           mutt_rfc822_read_line    (FILE *fp, char *line, size_t *linelen, LOFF_T *loc);
```
gahr, line 49: What do you think about returning it, instead? A return of 0 would be the same as the current return of NULL, I think?

sesse: It's an in/out parameter, so if so, the caller would need to do the addition themselves. But we could do that if you want to; I don't have strong opinions. (I assumed the return value was in use, but it seems it's mostly a true/false in practice.)

gahr: yeah, I think I'd prefer if the function returned a "pure" value, and the caller does with it whatever it wants. I think it's prettier to see, and it doesn't require the `if (loc)` checks in the function.

sesse: What do we do if the buffer is reallocated? How can we communicate the new one back?

We can get rid of the if (loc) by simply making loc compulsory. This means two call sites would need to send in dummy loc values that are thrown away.

gahr: > What do we do if the buffer is reallocated? How can we communicate the new one back?

good point - I guess what we should do is upgrade this function to take a `struct Buffer*` - I can do that change if you want me to

sesse: Yes, please do :-) I'll rebase after that.

flatcap (approved):
Lovely, much tidier now :-)

PR #3805, opened 2023-04-06
hcache: do less work when not in use

sesse:
If hcache is compiled in, but it's not actually in use, we would still go through a lot of the motions; check for flags, stat() files (if maildir_header_cache_verify is not explicitly turned off), and so on. We can short-circuit a lot of these operations by simply checking early on whether the hcache is NULL or not.

Speeds up opening a large Maildir by about 2.5% (in the specific situation where hcache is compiled in but not enabled).

* **What does this PR do?**

     for that)

### DIFF
--- a/hcache/hcache.c
+++ b/hcache/hcache.c
@@ -340,10 +340,13 @@ static char *get_foldername(const char *folder)
  */
 static void *fetch_raw(struct HeaderCache *hc, const char *key, size_t keylen, size_t *dlen)
 {
+  if (!hc)
+    return NULL;
+
   const char *const c_header_cache_backend = cs_subset_string(NeoMutt->sub, "header_cache_backend");
   const struct StoreOps *ops = store_get_backend_ops(c_header_cache_backend);
 
-  if (!hc || !ops)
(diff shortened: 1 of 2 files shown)


flatcap (approved):
Lovely!
Keep 'em coming :-)

sesse:
It seems we're now getting to the point where a significant fraction of the CPU time spent is fopen + fread + fclose. Which means it's time to make a proof-of-concept for io_uring… And that's going to be a lot more code than these simple fixes.

flatcap:
> it's time to make a proof-of-concept for io_uring

If you want to discuss strategies, feel free to [open a discussion](https://github.com/neomutt/neomutt/discussions/categories/development).
But I get the feeling you want to start hacking :-)

> And that's going to be a lot more code than these simple fixes.

If you've got the time and the enthusiasm for it, I'd love to see it.
The maildir code is quite complex, so ask lots of questions.

You can also find us on IRC: `#neomutt` on `irc.libera.chat` ([web client](https://kiwiirc.com/nextclient/irc.libera.chat/#neomutt))

PR #3806, opened 2023-04-06
Use Buffer in mutt_rfc822_read_line

gahr:
This cleans up the mutt_rfc822_read_line API and implementation, and it makes it return the number of chars read from the file.

### DIFF
--- a/email/parse.c
+++ b/email/parse.c
@@ -1095,13 +1095,17 @@ size_t mutt_rfc822_read_line(FILE *fp, struct Buffer *out)
 
   while (true)
   {
-    if (!fgets(buf, linelen - offset, fp) || /* end of file or */
-        (IS_SPACE(*line) && !offset))        /* end of headers */
+    if (!fgets(buf, linelen - offset, fp))
     {
       return 0;
     }
 
     const size_t len = mutt_str_len(buf);
+    if ((IS_SPACE(*line) && !offset))
+    {
+      return len;
+    }
+
     if (len == 0)
     {
       return read;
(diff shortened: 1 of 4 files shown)


sesse:
This makes my ftello() patch basically trivial, right?

email/parse.c
```diff
@@ -1076,35 +1076,40 @@ int mutt_rfc822_parse_line(struct Envelope *env, struct Email *e, const char *na
...
   char *buf = line;
   int ch;
   size_t offset = 0;
+  size_t read = 0;
 
   while (true)
   {
-    if (!fgets(buf, *linelen - offset, fp) || /* end of file or */
+    if (!fgets(buf, linelen - offset, fp) || /* end of file or */
         (IS_SPACE(*line) && !offset))         /* end of headers */
```
sesse, line 1102: In this case, you don't update read. It needs to be done before checking IS_SPACE() (I do this in my patch).

test/parse/mutt_rfc822_read_line.c
```diff
@@ -31,18 +31,12 @@ void test_mutt_rfc822_read_line(void)
   // char *mutt_rfc822_read_line(FILE *fp, char *line, size_t *linelen);
```
sesse, line 31: This probably needs updating now.

email/parse.c
```diff
@@ -1190,21 +1199,23 @@ struct Envelope *mutt_rfc822_read_header(FILE *fp, struct Email *e, bool user_hd
 
   while ((loc = ftello(fp)) != -1)
   {
-    line = mutt_rfc822_read_line(fp, line, &linelen);
-    if (*line == '\0')
+    if ((mutt_rfc822_read_line(fp, line) == 0) || (mutt_buffer_len(line) == 0))
```
sesse, line 1202: I don't understand when mutt_buffer_len(line) could be 0 but we could have read more bytes; could that ever happen? If we do, don't we need to seek back?

flatcap (approved):

PR #3812, opened 2023-04-08
email: cache assumed_charset in decode_rfc2047()

sesse:
Instead of looking assumed_charset every time we want to decode an address, cache it once, and only clear that cache when the configuration (any variable, for simplicity) changes.

Speeds up opening a large Maildir by about 1%.

* **What does this PR do?**

     for that)


sesse:
I haven't actually tested the cache invalidation, because I have no idea how one actually changes these settings runtime.

gahr:
`assumed_charset`, together with `charset`, is ultimately used in `rfc2047_decode`
https://github.com/neomutt/neomutt/blob/f4917d48d20d2c256a6ff275d9fb08f06536d62d/email/rfc2047.c#L699-L707

If we want to optimize this, wouldn't it be a better option to pass these down as parameters? The caller could read the values once before parsing a whole mailbox, for example. But it's a trade off between a nice API and speed... not sure it's worth it.

sesse:
> If we want to optimize this, wouldn't it be a better option to pass these down as parameters? The caller could read the values once before parsing a whole mailbox, for example. But it's a trade off between a nice API and speed... not sure it's worth it.

The problem is that it “infects” a whole bunch of callers down the chain. I tried it, and wasn't happy with the results, so I ended up on this. Given that almost all of config is done by checking globals, do we want to introduce a different pattern (sending it down as parameters) across the API for performance reasons?

flatcap:
I don't like the fact that we're doing config lookups so deep in the code,
(that sort of thing makes testing a nightmare)
but I agree passing down extra parameters is going to be an ugly mess.

flatcap:
I've pushed a working alternative to [[devel/charset-cache]](https://github.com/neomutt/neomutt/tree/devel/charset-cache).
It sets up a config observer, at a global level, which tracks `$charset` and `$assumed_charset`.

gahr:
> I've pushed a working alternative to [[devel/charset-cache]](https://github.com/neomutt/neomutt/tree/devel/charset-cache). It sets up a config observer, at a global level, which tracks `$charset` and `$assumed_charset`.

Nice, I was thinking of something along these lines, but pushed down to the config layer: what if we had a number of functions like `cs_assumed_charset(NeoMutt->sub)` that do the caching and fall back to `cs_subset_slist(NeoMutt->sub, "assumed_charset")` ? We could provide more of these as we find hot spots and they'd all be managed in a central place.

PR #3813, opened 2023-04-08
browser: fix crash in op_browser_subscribe()

seabass-labrax:
* **What does this PR do?**

Fix a segmentation fault that occurs in `op_browser_subscribe()` when the list of mailboxes is empty. The macro `ARRAY_GET` intentionally returns NULL when an out-of-bounds index is accessed; this value is assigned to `*ff`, causing a null pointer access in `mutt_str_copy()`.

Return `FR_ERROR` from `op_browser_subscribe()` if there are no mailboxes in the browser.


seabass-labrax:
I wasn't quite sure what to make the error message, so I've written a
new one analogous to the "There are no attachments" and "There are no
messages" errors. The new one will of course require translations.

gahr (approved):

flatcap:
You're welcome.  Thanks for the fix :-)

PR #3816, opened 2023-04-09
mutt: cache calls to iconv_open()

sesse:
Calling iconv_open() for every message can be expensive (it seems especially with some source charsets, like iso-2022-jp), and libc does not necessarily cache anything itself. Add a small LRU cache around it.

Speeds up opening a large Maildir by about 8%.

* **What does this PR do?**

     for that)

### DIFF
--- a/convert/convert.c
+++ b/convert/convert.c
@@ -185,11 +185,6 @@ size_t mutt_convert_file_to(FILE *fp, const char *fromcode, struct Slist const *
     }
   }
 
-  for (int i = 0; i < ncodes; i++)
-    if (cd[i] != (iconv_t) (-1))
-      iconv_close(cd[i]);
-
-  iconv_close(cd1);
   FREE(&cd);
   FREE(&infos);
   FREE(&score);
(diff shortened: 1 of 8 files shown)


gahr:
I have two questions:
1. when/where do we close them?
2. are we sure that a descriptor in the cache is always in a consistent state? What if a previous conversion failed?

sesse:
It's a good question; I hadn't thought of the fact that iconv descriptors store state beyond what's in the input and output buffers. If so, it might actually be impossible to cache them.

As for closing, we simply never do, except when they fall out of the cache. So there's a slight amount of permanently increased memory, but it's in the kilobytes and does not increase.

sesse:
I've added a cleanup function to main().

flatcap:
Looks good to me.

I've rebased and done a bit of tidying of your code.
It's in a separate commit, so you can see what I've done.
(I'll squash it, if you're happy)

- Move defs and data to top of file
- Renamed globals to MixedCase
- Added Doxygen comments for functions, structs, members, globals
- Added logging for adding/dropping entries to/from the cache
- Added `void` to function prototypes (this is C, not C++ ;-)

The second commit is some general tidying of our iconv usage.
(stuff that occurred to me whilst reading)

flatcap (approved):

PR #3817, opened 2023-04-10
mutt: do strict date parsing without regex

sesse:
sscanf() is a generic and large function, which is overkill for parsing our simple 2- and 4-byte dates and years. Replace it with a trivial integer parsing routine.

Speeds up opening a large Maildir by about 3%.

* **What does this PR do?**

     for that)

### DIFF
--- a/mutt/date.c
+++ b/mutt/date.c
@@ -27,14 +27,17 @@
  */
 
 #include "config.h"
+#include <ctype.h>
 #include <stdbool.h>
 #include <stdint.h>
 #include <stdio.h>
 #include <stdlib.h>
+#include <string.h>
 #include <sys/time.h>
 #include <time.h>
 #include "date.h"
 #include "buffer.h"
+#include "eqi.h"
 #include "logging2.h"
 #include "memory.h"
 #include "prex.h"
@@ -409,7 +412,7 @@ void mutt_date_make_date(struct Buffer *buf, bool local)
 
 /**
  * mutt_date_check_month - Is the string a valid month name
- * @param s String to check
+ * @param s String to check (must be at least 3 bytes long)
  * @retval num Index into Months array (0-based)
  * @retval -1  Error
  *
@@ -418,9 +421,20 @@ void mutt_date_make_date(struct Buffer *buf, bool local)
  */
 int mutt_date_check_month(const char *s)
 {
+  if (!s)
+    return -1;
+
+  char buf[4] = { 0 };
+  memcpy(buf, s, 3);
+  uint32_t sv;
+  memcpy(&sv, buf, sizeof(sv));
   for (int i = 0; i < mutt_array_size(Months); i++)
-    if (mutt_istr_startswith(s, Months[i]))
+  {
+    uint32_t mv;
+    memcpy(&mv, Months[i], sizeof(mv));
+    if (sv == mv)
       return i;
+  }
 
   return -1; /* error */
 }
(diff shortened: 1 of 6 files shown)


gahr:
The PR doesn't seem to (only) do what the title says. That aside, I don't think I like this. `mutt_date_parse_rfc5322_strict` is exactly the type of functions we're been trying to avoid with regex-based parsing: they are large, difficult to parse (for humans), and error prone. `eqi8` is nice, but where do we stop? How much readability are we prepared to give up for some speedup?

sesse:
> The PR doesn't seem to (only) do what the title says. That aside, I don't think I like this. `mutt_date_parse_rfc5322_strict` is exactly the type of functions we're been trying to avoid with regex-based parsing: they are large, difficult to parse (for humans), and error prone. `eqi8` is nice, but where do we stop? How much readability are we prepared to give up for some speedup?

The PR title is outdated, yes; I originally only changed sscanf(), but figured the full parser was pretty easy to write and much faster. The commit message was been updated; I can update the PR text if you want to (if I can figure out how in GitHub :-) ).

As for whether you want a regex or a parser, that's up to you and the other NeoMutt developers; I have a pretty clear preference, but it's not my call.

mutt/date.c
```diff
@@ -447,6 +460,206 @@ uint64_t mutt_date_now_ms(void)
...
+  /* Second (optional) */
+  if (len > 0 && s[0] == ':')
+  {
+    ++s;
+    --len;
+    if (len < 2 || s[0] < '0' || s[0] > '5' || s[1] < '0' || s[1] > '9')
+      return -1;
+    tm.tm_sec = (s[0] - '0') * 10 + (s[1] - '0');
+    if (tm.tm_sec > 60)
+      return -1;
```
vuori, line 589: For min and sec, is there a way for the second comparison to fail if the first one succeeds? For sec, shouldn't the second one be `> 59` (unless you want to support leap seconds)?

mutt/date.c
```diff
@@ -418,9 +420,20 @@ void mutt_date_make_date(struct Buffer *buf, bool local)
...
+  uint32_t sv;
+  memcpy(&sv, buf, sizeof(sv));
   for (int i = 0; i < mutt_array_size(Months); i++)
-    if (mutt_istr_startswith(s, Months[i]))
+  {
+    uint32_t mv;
+    memcpy(&mv, Months[i], sizeof(mv));
+    if (sv == mv)
       return i;
+  }
```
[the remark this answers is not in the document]
sesse: 1. I'm not really sure; should we really have a parameter whose only point is to be asserted on?
2. Yes. Much, much faster (and much less obscure).
3. I don't understand this comment. The memcpy is a single 32-bit read, it becomes a mov instruction. How can you get faster than that? (OK, if you are on MIPS or something with no unaligned read instruction, perhaps…)

mutt/date.c
sesse: The existing code supports 60 (which I would assume comes from the RFC); I should probably fix the code to s[0] > '6' instead.

mutt/date.c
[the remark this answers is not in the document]
vuori: re 1: Does the existing call in `from.c:is_from()` guarantee the 3-character minimum? I don't think the existing version assumed anything besides a null-terminated string? For a non-static function parsing untrusted input this kind of implicit length assumption feels like a footgun-in-the-making to me.

mutt/date.c
vuori: Leap seconds are rare enough that checking for 59 and dropping to slow path if 60 is encountered seems pretty ok to me. Can't think of another situation where 60 is encountered at least.

flatcap:
## Benchmarks

| Test case | Coverage | Time (Net) | Description                                    |
| :-------- | -------: | ---------: | :--------------------------------------------- |
| empty     | 100%     | n/a        | Test, without call to `mutt_date_parse_date()` |
| speed     | 95%      | 0s         | Fast parser only                               |
| speed2    | 100%     | 1s         | Fast parser + lax fallback                     |
| prex      | 95%      | 9s         | Fast prex only                                 |
| prex2     | 100%     | 10s        | Fast prex + lax fallback                       |
| lax       | 100%     | 19s        | Lax only                                       |

**Coverage**: How many dates could be parsed
**Time (net)**: How much longer than the empty test rig

**Note**: The new fast parser essentially took zero time -- the error bars were bigger than the data.

## Test Data

**Data**: https://flatcap.org/mutt/dates.xz (35MiB)

Source: LKML headers
444MiB uncompressed
14,130,022 dates

## Test Source

```c
#include "config.h"
#include <stdio.h>
#include "mutt/lib.h"

int main()
{
  char buf[256];

  int count = 0;
  int valid = 0;
  while (fgets(buf, sizeof(buf), stdin))
  {
    size_t len = strlen(buf);
    buf[len - 1] = '\0';

    struct Tz tz = { 0 };
    time_t t = mutt_date_parse_date(buf, &tz);
    // if (t == -1)
    //   printf("INVALID: %s\n", buf);
    if (t != -1)
      valid++;

    count++;
  }

  printf("%d lines\n", count);
  if (count != valid)
  {
    printf("%d valid\n", valid);
    printf("%d invalid\n", count - valid);
  }
  return 0;
}
```

## Run

**Build neomutt**:

```sh
./configure --disable-doc --pcre2
make EXTRA_CFLAGS="-O2"
```

**Build test**:

```sh
gcc -O2 -I. -o date-test date-test.c libmutt.a -lpcre2-8
```

**Run**:

```sh
cat dates | date-test
```

gahr:
It's hard to argue against this figures alone.

However, in context, I am still dubious that the added complexity is worth the time gain, given the probably not so common use case of parsing tens of thousands of dates in a row.

To me, it would be a pity to give up the nice declarative regex for a hand-written parser. Especially since the author has stated that it won't be around to maintain it.

flatcap (approved):

PR #3818, opened 2023-04-10
build: refactor testing

flatcap:
Make it more obvious how to run the NeoMutt unit tests.

- e40deede4 build: drop configure --testing option
  Always include the test Makefile, just don't build anything by default

- 2f5080c86 build: add instructions for testing
  If `$NEOMUTT_TEST_DIR` isn't set, or the directory doesn't contain `setup.sh`

---

To run the unit tests, test files must be downloaded from:

- https://github.com/neomutt/neomutt-test-files

and an environment variable must be set to point to the test files:

- `export NEOMUTT_TEST_DIR=/path/to/neomutt-test-files`

### DIFF
--- a/Makefile.autosetup
+++ b/Makefile.autosetup
@@ -955,9 +955,7 @@ coverage: all test
 @include @srcdir@/contrib/Makefile.autosetup
 @include @srcdir@/data/Makefile.autosetup
 @include @srcdir@/docs/Makefile.autosetup
-@if ENABLE_UNIT_TESTS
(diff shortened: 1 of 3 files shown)


gahr:
How about we print a message at the end of `./configure` instead?

gahr:
Do you care that configuring with `--compile-commands` still triggers the build of the test objects? :)

`compile_commands.json` depends on `ALLOBJS` and `TEST_OBJS` are added to that in `test/Makefile.autosetup`. Perhaps test objects shouldn't be part of that?

PR #3821, opened 2023-04-13
create config cache for charsets

flatcap:
Cache the values of `$charset` and `$assumed_charset`.

Some email parsing routines rely on these config variables,
but lookup up their values every time is slowing down NeoMutt.

Create a config observer which sets the value of two new global variables: `CachedAssumedCharset`, `CachedCharset`.

Closes: #3812


config/cache.h
```diff
@@ -0,0 +1,32 @@
...
+ *
+ * You should have received a copy of the GNU General Public License along with
+ * this program.  If not, see <http://www.gnu.org/licenses/>.
+ */
+
+#ifndef MUTT_CONFIG_CACHE_H
+#define MUTT_CONFIG_CACHE_H
+
+extern const struct Slist *CachedAssumedCharset;
+extern const char *CachedCharset;
```
gahr, line 27: How about we expose functions instead of global variables? It'd give us more freedom to implement them one way or the other, e.g., we wouldn't need a setup function: we could init the backing global on demand.

flatcap:
Fully-encapsulated.

We can't use `static inline` functions, because we'd still need to `extern` the variables and there'd be no incentive to use the functions.

The observer is now lazily setup on the first call.
This happens in `main()` -> `mutt_init()` -> `mutt_hist_read_file()` -- almost immediately on startup.

The observer isn't manually shutdown, as we get this for free when the `NeoMutt` object is destroyed.

gahr (approved):
Nice!

PR #3826, opened 2023-04-19
Fix crash with the iconv cache

gahr:
A caching layer for iconv conversion descriptors was introduced in https://github.com/neomutt/neomutt/pull/3816.

There are two issues, which are both triggered when `iconv_open` returns an invalid conversion descriptor (e.g., because iconv doesn't understand the charsets being passed).

1. descriptors are [cached without checking their validity](https://github.com/sesse/neomutt/blob/27cf28e8b6f16af7a011e646cc3e88a53d6185e3/mutt/charset.c#L638-L664).  A subsequent use of an invalid cached descriptor will [trigger a reset](https://github.com/sesse/neomutt/blob/27cf28e8b6f16af7a011e646cc3e88a53d6185e3/mutt/charset.c#L616), which is needed for valid descriptor, but which causes a crash on invalid ones.
2. the code assumes that [an invalid descriptor is a descriptor not found in the cache](https://github.com/sesse/neomutt/blob/27cf28e8b6f16af7a011e646cc3e88a53d6185e3/mutt/charset.c#L629).  As explained in 1., this is not necessarily the case. The result is that even if 1. didn't cause a crash, we'd be pushing mroe and more invalid descriptors at the top of the cache.

@sesse

### DIFF
--- a/mutt/charset.c
+++ b/mutt/charset.c
@@ -603,16 +603,12 @@ iconv_t mutt_ch_iconv_open(const char *tocode, const char *fromcode, uint8_t fla
   }
 
   /* check if we have this pair cached already */
-  iconv_t cd = ICONV_T_INVALID;
   for (int i = 0; i < IconvCacheUsed; ++i)
   {
     if (strcmp(tocode1, IconvCache[i].tocode1) == 0 &&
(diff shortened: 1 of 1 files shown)


flatcap (approved):
lgtm

PR #3829, opened 2023-04-20
Add buf_gets to read a whole line from a file

gahr:
Hopefully this can be used in a few places where we manually loop over fgets


gahr:
This doesn't work - I might try again in the future :)

PR #3839, opened 2023-05-03
test: fix `neomutt-test -E` wrt the config cache

flatcap:
- 306bfa7ad test: create a NeoMutt for every test
  Rather than lots of tests creating / destroying their own NeoMutts,
  the shared code now does this for every test.
  (something I was already considering)

- 3d28070ce config cache: restore manual cleanup
  Manually call `config_cache_free()`.
  Many of the tests needed NeoMutt to be set up so they could use config.
  This meant hundreds of NeoMutts being created and destroyed,
  which in turn made the config cache hard to maintain.

Tests:
- Plain `test/neomutt-test` runs correctly
- ASAN `test/neomutt-test` runs correctly
- Plain `test/neomutt-test -E` runs correctly<sup>†</sup>
- ASAN `test/neomutt-test -E` runs correctly<sup>†</sup>

† When tests **test_filter_wait** and **test_mutt_sig_exit_handler** are excluded.

### DIFF
--- a/config/cache.c
+++ b/config/cache.c
@@ -37,12 +37,8 @@
 static bool CacheActive = false;
 /// Cached value of $assumed_charset
 static const struct Slist *CachedAssumedCharset = NULL;
-/// Is $assumed_charset active?
-static bool CacheActiveAssumedCharset = false;
 /// Cached value of $charset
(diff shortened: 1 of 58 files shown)


gahr:
Thanks!

PR #3841, opened 2023-05-04
Allow percentage i18n

bitigchi:
Some languages, including Turkish, French etc. use a different percentage display format like %100, 100 % respectively. Allowing percentage i18n makes it more coherent with the system locale settings.

### DIFF
--- a/progress/window.c
+++ b/progress/window.c
@@ -162,19 +162,19 @@ static int progress_window_repaint(struct MuttWindow *win)
 
   if (wdata->size == 0)
   {
-    message_bar(wdata->win, wdata->display_percent, "%s %zu (%d%%)", wdata->msg,
+    message_bar(wdata->win, wdata->display_percent, _("%s %zu (%d%%)"), wdata->msg,
                 wdata->display_pos, wdata->display_percent);
   }
   else
   {
     if (wdata->is_bytes)
     {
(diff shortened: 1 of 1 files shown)


progress/window.c
```diff
@@ -162,19 +162,19 @@ static int progress_window_repaint(struct MuttWindow *win)
 
   if (wdata->size == 0)
   {
-    message_bar(wdata->win, wdata->display_percent, "%s %zu (%d%%)", wdata->msg,
+    message_bar(wdata->win, wdata->display_percent, _("%s %zu (%d%%)"), wdata->msg,
```
flatcap, line 165: Please can you add a `L10N` comment, before each line, explaining what each of the fields is.
(same for the two below)

flatcap (changes requested):
Oh!  That looks good :-)

There's also `%P` in `$status_format`
https://github.com/neomutt/neomutt/blob/91f6b704c75dff97b58fdebf2b5022544afa7a5a/status.c#L299

I can't find any others (with a quick grep)

I'd _just_ announced a string freeze for the next release :-)
but if we get this wrapped up quickly, I'll merge it and let the translators know.

Thanks

flatcap:
and `$pager_format`
https://github.com/neomutt/neomutt/blob/91f6b704c75dff97b58fdebf2b5022544afa7a5a/pager/pbar.c#L113

PR #3845, opened 2023-05-08
Refactor MxOps->msg_open() to eliminate msgno

flatcap:
This PR refactors numerous functions to replace `int msgno` with `struct Email *e`.
The result is that the backends no longer reference `Email->msgno`.

---

This was a happy discovery...

`Email->msgno` is a sequential number given to Emails as seen in the Index.
(see: [`%C` in `$index_format`](https://neomutt.org/guide/reference#index-format))

It's also used as an index number, into the `Mailbox->emails[]` array, for `MxOps->msg_open()`.

e.g. `pipe_message() -> mx_msg_open() -> maildir_msg_open()`

In every case, the caller passes `Email->msgno` through the MXAPI and the backend, then looks up that number in `Mailbox->emails[]`.

---
Of course, there's a diagram of the changes :-)

- Each left-hand leaf uses `Email->msgno`
- Each right-hand leaf uses `Mailbox->emails[msgno]`

### DIFF
--- a/compmbox/compress.c
+++ b/compmbox/compress.c
@@ -743,7 +743,7 @@ static enum MxStatus comp_mbox_close(struct Mailbox *m)
 /**
  * comp_msg_open - Open an email message in a Mailbox - Implements MxOps::msg_open() - @ingroup mx_msg_open
  */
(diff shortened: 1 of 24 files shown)


gahr (approved):
Nice!

Issue #3648, opened 2023-01-29
A new mail is not treated as a new mail


whitney-cumber:
A new mail is not treated as a new mail.  Here is a reproducer:

## Steps to Reproduce

0. Setup

Generate three mailboxes
```
cd /tmp/testmail/
mkdir -p m{1,2,3}/{cur,new,tmp}
```
with a single new message each
```
cat >> m1/new/mail
From: foo
To: bar
Subject baz

foo
```
```
cp m1/new/mail m2/new/mail
cp m1/new/mail m3/new/mail
```

Use this neomuttrc
```
cat >>neomuttrc

set folder=/tmp/testmail
mailboxes =m1 =m2 =m3
set spool_file = =m2
set mail_check_stats = yes
set sidebar_visible = yes
set sidebar_sort_method = unsorted
set sidebar_format="%D%*  %n %N %o"
```

1. neomutt -F neomuttrc

2. Note that `=m1` and `=m3` have new mail, which can be checked with `.` (`<mailbox-list>`)

3. Mark mail in `m3` as read: `c=m3`, `N`.
4. Mark mail in `m3` as new: `N`
5. Change to `m2`: `c=m2`
6. `set mark_old=no`
7. quit neomutt: `q`
8. `m2` and `m3` should have the mail now in `cur/` not `new/`
9. Restart: `neomutt -F neomuttrc`
10. note that only `=m1` has new mail. (check with `.` or `<mailbox-list>`)
11.  `set mark_old=no`
12. Change to `=m3`:  `c=m3`
13. Notice that the message is marked as new but not listed as new.
14. Change there and back again `c=m2`, `c=m3`; still marked as new.

## NeoMutt version

```
20220429
```

gahr:
I can't repro this.

8. I *do* get m2 and m3 mails in cur
13. I *don't* get a new mail in m3, I get an *old* mail

gahr:
Can you please try with a newer version of NeoMutt?

whitney-cumber:
Tried with current main (496d2f6d0d49678a326bb72274774141f388e2c9) and I cannot reproduce this anymore.  So this seems to have been fixed along the way.  Thanks for looking into this.

Closing this issue.

PR #3848, opened 2023-05-09
Fix counting new mails in maildir

### LINKED ISSUE #3843: sidebar %Z expando is wrong for maildir
It`s always 1 whether mail_check_recent is no or yes.
Note that the equivalent status line expando (%n) is correct.

## Expected Behaviour

%Z should reflect the value of new mail (unread,unseen)

## Actual Behaviour

%Z is 1 

## Steps to Reproduce
set check_new = yes
set mail_check_stats = yes
set mail_check_stats_interval = 5
set mail_check_recent = no #(or yes)

set sidebar_format = "%B %Z"

- Always

## NeoMutt Version

latest master

gahr:
We only want to change msg_new if we're gathering stats. If we are only checking whether the mailbox has any new messages, we'll stop at the first new message we encounter, so msg_new would always be 1.

We've been touching this code lately, so I'm unsure if I'm breaking something else..

Fixes #3843

### DIFF
--- a/maildir/maildir.c
+++ b/maildir/maildir.c
@@ -149,10 +149,14 @@ static void maildir_check_dir(struct Mailbox *m, const char *dir_name,
           }
         }
         m->has_new = true;
-        check_new = false;
-        m->msg_new++;
-        if (!check_stats)
+        if (check_stats)
+        {
+          m->msg_new++;
+        }
+        else
+        {
           break;
+        }
       }
     }
   }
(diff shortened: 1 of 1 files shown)


lucilanga:
It does work for me in the scope of #3843. Thank you!
I do not see any immediate breakage but I need to test this more.

Possibly related: #3648, although not directly as it mentioned different set of expandos.

flatcap (approved):

Issue #3828, opened 2023-04-20
Message body search fails if search pattern happens across a read buffer boundary


gahr:
This is how we look for a pattern `pat` in the body of a message stored in a file we have a `fp` FILE open on.

https://github.com/neomutt/neomutt/blob/0f7155b9d54af25e46f226b5059ee739446f7fa8/pattern/exec.c#L240-L260

See how we read the file in chunks of 256 bytes, and search for the pattern in each chunk. So what happens is that the search will fail if the pattern appears in the file across a 256 bytes boundary.

Try to search for `foobar` in this email:

```email
From foo@bar.baz Mon Feb 27 12:12:21 2023
Date: Mon, 27 Feb 2023 12:12:21 +0000
From: foo@bar.baz
To: foo@bar.baz
Subject: Search for foobar in the body
MIME-Version: 1.0
Content-Type: text/plain; charset=us-ascii; format=flowed
Content-Disposition: inline

0123456789ABCDEF0123456789ABCDEF0123456789ABCDEF0123456789ABCDEF0123456789ABCDEF0123456789ABCDEF0123456789ABCDEF0123456789ABCDEF0123456789ABCDEF0123456789ABCDEF0123456789ABCDEF0123456789ABCDEF0123456789ABCDEF0123456789ABCDEF0123456789ABCDEF0123456789ABCfoobar
``` 

I am not sure how we want to handle this.

PR #3849, opened 2023-05-10
Read body in bigger chunks

gahr:
This is a mitigation for #3828.


flatcap (approved):
That'll do for now :-)

PR #3851, opened 2023-05-10
nm_windowed_query_from_query: better account for buffer length

marceloleitner:
With fortified sources, glibc will notice that this call to snprintf could potentially write more bytes to the buffer than what it can hold. That's because it is not accounting for the initial offset in the buffer itself.

* **What does this PR do?**
Fixes a crash when selecting a notmuch folder.

### DIFF
--- a/notmuch/query.c
+++ b/notmuch/query.c
@@ -250,7 +250,7 @@ nm_windowed_query_from_query(char *buf, size_t buflen, const bool force_enable,
   }
 
   // Add current search to window query.
-  snprintf(buf + length, buflen, " and %s", cur_search);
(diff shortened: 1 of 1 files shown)


flatcap (approved):
Good catch!
Thanks for the fix.

flatcap:
Hi @marceloleitner,

I've just noticed that GitHub doesn't link your commit back to your account.
It's probably just a case of adding your `redhat` email address to [your profile](https://github.com/settings/profile).

marceloleitner:
I should have used the gmail one. Even the s-o-b line was using it, but okay. I just added the redhat one to my account..
Thanks @flatcap !

PR #3852, opened 2023-05-11
Maildir: Configurable field delimiter

gahr:
We allow overriding `:` as field delimiter for maildir using the option maildir_field_delimiter. This option should be changed only in tandem with changing existing maildir files to have that field delimiter.

### DIFF
--- a/config/cache.c
+++ b/config/cache.c
@@ -39,6 +39,8 @@ static bool CacheActive = false;
 static const struct Slist *CachedAssumedCharset = NULL;
 /// Cached value of $charset
 static const char *CachedCharset = NULL;
+/// Cached value of $maildir_field_delimiter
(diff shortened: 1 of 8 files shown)


asdf8dfafjk:
I just learned that offlineimap  uses `!` on Windows. (search for `maildir-windows-compatible` in https://www.offlineimap.org/doc/offlineimap.html). 

Also adding some notes from the earlier IRC conversation:

1) mbsync uses `;` on windows. search for  `FieldDelimiter delim` in https://isync.sourceforge.io/mbsync.html 

2) https://cr.yp.to/proto/maildir.html

Noting this should someone want to understand the code change better ...

flatcap:
Widespread, but minor changes: clang-format, spelling, sorting, reduce scope of variables, doxygen, iwyu.

flatcap (approved):

PR #3856, opened 2023-05-13
Fix broken $tunnel reconnect with $ssl_force_tls=yes

DanielG:
When $ssl_force_tls=yes imap bails out with "Encrypted connection unavailable" on reconnect when the tunnel process exits. We observed `conn->ssf==0` in this case.

This happens because the tunnel backend only sets `conn->ssf=1` once ever in mutt_tunnel_socket_setup, but mutt_socket_close sets `conn->ssf=0` on process exit.

Fix this by refreshing `conn->ssf` in mutt_socket_tunnel_open.

---

Note: as a workaround `$ssl_force_tls` can just be set to false.

### DIFF
--- a/conn/tunnel.c
+++ b/conn/tunnel.c
@@ -129,6 +129,12 @@ static int tunnel_socket_open(struct Connection *conn)
 
   conn->fd = 42; /* stupid hack */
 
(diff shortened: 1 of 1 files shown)


flatcap (approved):
Brilliant!
Thanks for the fix.

PR #3869, opened 2023-05-20
imap: move $delete_untag out

flatcap:
The "(un-)tagging" of Emails is a GUI thing; it doesn't belong in the backends.

When `imap_copy_messages()` succeeds, it checks config `$delete_untag` and untags the Emails that it worked on.
This is called from `mutt_save_message()` and `mx_mbox_close()`.

Move this code out of Imap into the two callers.

### DIFF
--- a/external.c
+++ b/external.c
@@ -917,6 +917,17 @@ int mutt_save_message(struct Mailbox *m, struct EmailList *el,
       case 0:
         mutt_clear_error();
         rc = 0;
+        if (save_opt == SAVE_MOVE)
+        {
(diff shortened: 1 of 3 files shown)


gahr (approved):
Doesn’t this call for a dedicated function to avoi repetition?

flatcap:
> Doesn’t this call for a dedicated function to avoid repetition?

Yes, but I might leave that for now.
All the tag functions are slowly moving forward (toward the front-end).
When then stop moving, I'll collect them up.

Issue #3855, opened 2023-05-13
'bright', 'alert', 'light' are only allowed for named colors:


misaflo:
## Expected Behaviour

Colors like brightcolor223 sould works.

## Actual Behaviour

I have this error at startup:

    [...]/colors-gruvbox.muttrc, ligne 57 : 'bright', 'alert', 'light' are only allowed for named colors: color223
    [...]/colors-gruvbox.muttrc, ligne 58 : 'bright', 'alert', 'light' are only allowed for named colors: color142
    [...]/colors-gruvbox.muttrc, ligne 59 : 'bright', 'alert', 'light' are only allowed for named colors: color214
    [...]/colors-gruvbox.muttrc, ligne 60 : 'bright', 'alert', 'light' are only allowed for named colors: color142

## Steps to Reproduce

Add this to your color colorscheme:

    color index         brightcolor223 color235 '~N'
    color index_flags   brightcolor142 color235 '~N'
    color index_author  brightcolor214 color235 '~N'
    color index_subject brightcolor142 color235 '~N'

Or [this colorscheme](https://github.com/misaflo/neomutt-gruvbox).

## How often does this happen?

- Always

## When did it start to happen?

- When I upgraded to version 2023-05-12.

## NeoMutt Version

```
NeoMutt 20230512
Copyright (C) 1996-2022 Michael R. Elkins and others.
NeoMutt comes with ABSOLUTELY NO WARRANTY; for details type 'neomutt -vv'.
NeoMutt is free software, and you are welcome to redistribute it
under certain conditions; type 'neomutt -vv' for details.

System: Linux 6.3.1-arch2-1 (x86_64)
ncurses: ncurses 6.4.20221231 (compiled with 6.4.20221231)
libidn2: 2.3.4 (compiled with 2.3.4)
GPGME: 1.20.0
GnuTLS: 3.8.0
libnotmuch: 5.6.0
storage: kyotocabinet, gdbm, lmdb
compression: lz4, zlib, zstd

Configure options: --prefix=/usr --sysconfdir=/etc --libexecdir=/usr/lib --gpgme --sqlite --autocrypt --lua --notmuch --gss --gnutls --sasl --with-ui=ncurses --with-idn2=/usr --disable-idn --idn2 --lmdb --kyotocabinet --gdbm --ubsan --lz4 --zlib --zstd

[...]

PR #3871, opened 2023-05-21
Devel/color colorschemes

rayfordshire:
* **What does this PR do?**

  - Use "bold" attribute and "light" prefix instead of "bright" prefix
  - Split the parsing logic for different colour syntaxes ("#RRGGBB", "colorNNN", named colour) into separate functions.

#3855

* **Remark:**

Currently, these internal parser functions abuse `MUTT_CMD_WARNING` to indicate they failed parsing and a different syntax should be used.  It may, or may not be cleaner to introduce a new command result, e.g. `MUTT_CMD_RETRY` or `MUTT_CMD_CONTINUE`, which indicates that the next parsing method should be tried on the same input data.  Alternatively, those methods could take an additional argument `enum CommandResult *ret` which they write their command result to and use the return value of the function to indicate "I handled it"/"next syntax should get a try".

### DIFF
--- a/color/command.c
+++ b/color/command.c
@@ -116,6 +116,17 @@ const struct Mapping ComposeColorFields[] = {
   // clang-format on
 };
 
+/**
+ * ColorPrefix - Constants for colour prefixes of named colours
+ */
+enum ColorPrefix
+{
+  COLOR_PREFIX_NONE,   ///< no prefix
+  COLOR_PREFIX_ALERT,  ///< "alert" colour prefix
+  COLOR_PREFIX_BRIGHT, ///< "bright" colour prefix
+  COLOR_PREFIX_LIGHT,  ///< "light" colour prefix
+};
+
 #ifdef NEOMUTT_DIRECT_COLORS
 /**
  * color_xterm256_to_24bit - Convert a xterm color to its RGB value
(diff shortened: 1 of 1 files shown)


flatcap (changes requested):
> Split the parsing logic for different colour syntaxes

Good call!

> parser functions abuse MUTT_CMD_WARNING to indicate they failed parsing and a different syntax should be used.

That seems OK.

I've updated the Doxygen header to spell out all the possible retvals.

One other change, I've renamed `parse_color_colorNNN()` to `parse_color_colornnn()` for consistency.
(all functions are currently `[a-z0-9_]+`)

The only thing I'm not _entirely_ happy with is the duplication of the prefix parsing ("bright", ...).
If you could factor _that_ out, it'd be perfect.

Thanks

rayfordshire:
> The only thing I'm not entirely happy with is the duplication of the prefix parsing ("bright", ...).
> If you could factor that out, it'd be perfect.

Not entirely sure what you mean.  Do you mean splitting out a method:
```
void parse_prefix(const char *s, uint32_t *col, int *attrs,
                  bool is_fg, struct Buffer *err,
                  bool *bright_out, bool *light_out, bool *alert_out)
```
?  That seems ugly, due to the three return values.

flatcap:
yeah

Could we pass out the combined result in `int *attr` -- merging `A_BOLD | ...`
Something like:

`int parse_prefix(const char *str, int *attr)`

that returns the prefix length.

Maybe :-)

If it still looks ugly, we'll just leave the code as it is.

flatcap (approved):
That'll do nicely.
Thanks!

PR #3877, opened 2023-05-23
imap: factor out tagged emails

flatcap:
OK, this is going to take some explaining...
(fortunately, I have a diagram :-)

There are three external IMAP functions that work on sets of `Email`s: `imap_fast_trash()`, `imap_sync_mailbox()` and `imap_copy_messages()`.
They work by passing a `Mailbox` to the function along with an attribute, such as `MUTT_FLAG` (denoting flagged messages).

The workhorse of these functions is `imap_exec_msgset()`.
It sends a command and a set of UIDs to the IMAP server.
To prevent the UID list getting to long, it must be abbreviated; this is done by 'make_msg_set()'.
It turns: `1,2,3,4,6,8,9,10` into `1:4,6,8:10`.
If the abbreviated list gets too long, then `imap_exec_msgset()` sends a partial list to the server and calls 'make_msg_set()' again (and again).  

'make_msg_set()' is quite complicated.
It takes a `Mailbox` parameter and, say, `MUTT_FLAG`.
It iterates through `Mailbox.emails[]` looking for `Email.flagged == true`, then it looks up the UID in the `ImapEmailData`.
Having found a match, it adds the UID to a `Buffer`.

This works for several attributes, but for **tagged** `Email`s, **it's a problem**.
Tagging an `Email` is a front-end thing: the backend shouldn't be aware of tagging.

**The solution** is to refactor these functions and **pass in** a list of `Email`s that we want to work with.

In this PR, I've separated the `Email` and UID parts.

`imap_fast_trash()` and `sync_helper()` call a new function:
`select_email_uids()` does a lookup based on the attribute, e.g. `MUTT_FLAG` and stores the UIDs in an array.

`imap_copy_messages()` is passed an `EmailArray` by the Index.
It calls a new function:
`emails_to_uid_array()` extracts the UIDs from the `Email`s and stores them in an array.

`imap_exec_msgset()` has been modified to take a `UidArray`.
'make_msg_set()' has been simplified to take a `UidArray` and now only deals with UIDs.


flatcap:
Updates:
- Tidy code
- Move Message Set code to a separate file
- Add **extensive** tests
- 100% code coverage

Ready for review

imap/imap.c
```diff
@@ -234,17 +303,28 @@ static int sync_helper(struct Mailbox *m, AclFlags right, enum MessageType flag,
   if ((right == MUTT_ACL_WRITE) && !imap_has_flag(&imap_mdata_get(m)->flags, name))
     return 0;
 
+  int count = 0;
+  char buf[1024] = { 0 };
+
+  struct UidArray uida = ARRAY_HEAD_INITIALIZER;
+
+  // Set the flag (+FLAGS) on matching emails
+  select_email_uids(emails, num_emails, flag, true, false, &uida);
```
gahr, line 312: Don't we need to sort the array here? Also, would it make sense to move the sorting inside `select_email_uids` and `emails_to_uid_array`?

flatcap: > Don't we need to sort the array here?

No, because we sorted `emails` in `imap_sync_mailbox()`.

https://github.com/neomutt/neomutt/blob/76ba1c95423b760a120e6c3afc85c1e6db1c8a65/imap/imap.c#L1565-L1568

> would it make sense to move the sorting inside `select_email_uids()`

Ideally, yes, but `imap_sync_mailbox()` leads to **ten** `select_email_uids()` calls in quick succession.
Using the same set of emails, it calls `sync_helper()` for each attribute.

https://github.com/neomutt/neomutt/blob/76ba1c95423b760a120e6c3afc85c1e6db1c8a65/imap/imap.c#L1570-L1578

and `sync_helper()` calls `select_email_uids()` and `imap_exec_msg_set()` twice:

https://github.com/neomutt/neomutt/blob/76ba1c95423b760a120e6c3afc85c1e6db1c8a65/imap/imap.c#L311-L327

> and `emails_to_uid_array()`

Yes, that's a bit tidier. Done.

gahr (approved):
LGTM - let's run it for some time before the next release :)

PR #3886, opened 2023-06-06
Unify the new/free/clear/cleanup functions

flatcap:
WIP

- if a struct/obj needs a new() function it must have a _free() function (and vice versa)
- try to only use _free() for a function that actually **frees** the object (and NULLs the pointer)
- use _clear() for functions that empty/initialise an object, but don't actually free it
- use _cleanup() for functions that only work on (hidden) globals -- all those calls at the end of main()


flatcap:
Done for now

PR #3891, opened 2023-06-12
move Mailbox.mtime into private MboxData

flatcap:
Mailbox.mtime is used internally by several backends, but it's not used outside of them.

Move mtime into the private MboxData (or AccountData for the mbox backend)

---

Not all mailbox types use `mtime`, but those that do use it in the same way.

https://github.com/neomutt/neomutt/blob/d0fdd26246988a0efe98b6158461c60c59de35fd/core/mailbox.h#L104

Should it live, conveniently, in `Mailbox`, or probably more correctly in `MboxData`?


gahr:
Yup - this is probably more correct

flatcap:
ok, thanks.
I'll merge this and be on the lookout for others.

PR #3896, opened 2023-06-18
test for, and fix, logging problems

flatcap:
**Updated**:

- Refactor the log dispatcher to explicitly name the printf format parameter
- Add the _printf_ attribute to the log dispatcher

These functions can now be checked for mismatched parameters:

- `mutt_debug()`
- `mutt_error()`
- `mutt_message()`
- `mutt_perror()`
- `mutt_warning()`

e.g.

`error: format specifies type 'char *' but the argument has type 'struct Buffer *'`

### DIFF
--- a/mutt/logging.c
+++ b/mutt/logging.c
@@ -242,8 +242,8 @@ bool log_file_running(void)
  *
  * If stamp is 0, then the current time will be used.
  */
-int log_disp_file(time_t stamp, const char *file, int line,
-                  const char *function, enum LogLevel level, ...)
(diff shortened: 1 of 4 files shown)


gahr:
Is the idea that we turn this on, fix all the issues, then remove the config-time option and enable the new code by default?

flatcap:
**Update**:

After some discussion on IRC, we've decided to always enable the warnings.
(The `configure` option has been dropped)

gahr (approved):
Very nice!

Issue #3884, opened 2023-06-03
<sync-mailbox> "Unable to write" when email (in a Maildir) was marked for deletion but was already deleted on the file system (by mbsync)


aurelf:
## Expected Behaviour

1) Marked a mail in a Maildir folder for deletion ("d") in mutt. 
2) Also deleted the file on another client, synchronized to the Maildir (with mbsync to synchronize an IMAP server). 
3) When syncing mailbox, if the file is already deleted, mutt should log a warning, but let me continue. 

Note that I have this in my config:
```
set trash="~/Maildir/Trash"
set maildir_trash
```
So email deletion is actually moving the email to trash.
I expect that if another email client deleted the mail it was already moved there and nothing is actually needed to do.

## Actual Behaviour

1) Mutt prints "Unable to write", aborts syncing mailbox, which is also required to change folder and quit. So I have to kill mutt and relaunch. 

Debug logs contain: 
```
[2023-06-02 12:56:29]<M> sync_mailbox() Writing /home/<User>/Maildir/<folder>...
[2023-06-02 12:56:29]<5> msgwin_recalc() recalc done, request WA_REPAINT
[2023-06-02 12:56:29]<5> msgwin_repaint() repaint done
[2023-06-02 12:56:29]<P> maildir_sync_message() rename: No such file or directory (errno = 2): No such file or directory
```

## Steps to Reproduce

1) Mark mail in a Maildir folder for deletion ("d") in mutt. 
2) Delete the file manually (or through synchronization)
3) Attempt syncing mailbox

## Proposed fix 
Looks similar to #3040 but not duplicate as far as I can tell. 

Mutt should check if file still present before attempting to rename it in maildir.c:maildir_sync_message()
 https://github.com/neomutt/neomutt/blob/f5d4edf914e2fed867cb74442b6dbf2b20db8339/maildir/maildir.c#L450

My proposed fix (tested and works for me against current Arch neomutt version)

```diff
--- neomutt-20230517/maildir/maildir.c	2023-05-17 13:46:41.000000000 +0200
+++ neomutt-20230517_af/maildir/maildir.c	2023-06-04 00:06:54.456200149 +0200
@@ -392,6 +392,7 @@
   struct Buffer *partpath = NULL;
   struct Buffer *fullpath = NULL;
   struct Buffer *oldpath = NULL;
+  struct stat st = { 0 };
   char suffix[16] = { 0 };
   int rc = 0;
 
@@ -447,6 +448,11 @@
[...]

PR #3897, opened 2023-06-18
Fix for <sync-mailbox> "Unable to write" when email (in a Maildir) wa…

aurelf:
…s marked for deletion but was already deleted on the file system (by mbsync) #3884

Mutt should check if file still present before attempting to rename it in maildir.c:maildir_sync_message()

* **What does this PR do?**

Fixes issue #3884

### DIFF
--- a/maildir/maildir.c
+++ b/maildir/maildir.c
@@ -465,6 +465,13 @@ static int maildir_sync_message(struct Mailbox *m, struct Email *e)
     /* record that the message is possibly marked as trashed on disk */
     e->trash = e->deleted;
 
+    struct stat st = { 0 };
+    if (stat(buf_string(oldpath), &st) == -1)
+    {
+      mutt_debug(LL_DEBUG1, "File already removed (just continuing)");
+      goto cleanup;
+    }
+
     if (rename(buf_string(oldpath), buf_string(fullpath)) != 0)
     {
       mutt_perror("rename");


maildir/maildir.c
```diff
@@ -465,6 +466,11 @@ static int maildir_sync_message(struct Mailbox *m, struct Email *e)
     /* record that the message is possibly marked as trashed on disk */
     e->trash = e->deleted;
 
+    if (stat(buf_string(oldpath),&st) == -1){
+      mutt_warning("File already removed (just continuing).");
+      goto cleanup;
+    }
```
gahr, line 472:
```suggestion
    if (stat(buf_string(oldpath), &st) == -1)
    {
      mutt_warning("File already removed (just continuing).");
      goto cleanup;
    }
```

flatcap:
Thanks, @aurelf
We've made a few cosmetic changes to the PR...

- Run `clang-format` on the change
  It tidies the code, putting the `{}`s in our preferred places.
  The config file is in the repo, so `clang-format FILE.c` will work.

- Reduce scope of variable: `struct stat st`
  I've moved it close to where it's used.

- I've reduced the warning message to a log message
  1. It's only confirming what we'd expect to happen.
  2. Warning messages need translating and this one may not be needed in the future.

- Shorten commit message
  The commit title should be short.
  Try to cram lots of searchable keywords into, say 60 chars or fewer.
  The main message can be as detailed as you like.

flatcap (approved):

aurelf:
Great, thank you for the feedback and the merge !

flatcap:
@aurelf 
I've just noticed that your commit isn't linked back to your account.
That probably means GitHub doesn't know your `a.f@e.fr` address.
You can add it here: https://github.com/settings/emails

aurelf:
Didn't notice that, will do. Thanks !

PR #3898, opened 2023-06-20
shrink the header cache

flatcap:
**tl;dr** -- Save **352 bytes** per record in the header cache.

Mutt's header cache stores an entire `struct Email` and `struct Body` into the backend Store for every Email.
While this is efficient, it wastes a lot of space.

Back in #1870 (the PR, not the year :-) I worked out how little we actually needed to save.
This PR implements the idea.

**Note**: `hcache/hcachever.sh` has been updated because I've changed the on-disk format of the cache, but the structs haven't changed.

---

Here are the `struct Email` members that we want to save:
https://github.com/neomutt/neomutt/blob/14d6d0607e4b9c6c901eeac9425bde872e7c4e37/email/email.h#L38-L60

To save space, I've packed the `security` and other flags into one `uint32_t` and the timezone info into another.
There are some helper functions to do the (un)packing.

**Saving**: 188 bytes per record

---

Here are the `struct Body` members that we want to save:
https://github.com/neomutt/neomutt/blob/14d6d0607e4b9c6c901eeac9425bde872e7c4e37/email/body.h#L37-L62

To save space, I've packed the content type and flags into a `uint32_t`.
There are some helper functions to do the (un)packing.

**Saving**: 164 bytes per record

---

To store the `time_t` and `LOFF_T`, I created:

- `serial_dump_uint64_t()`
- `serial_restore_uint64_t()`

The (un)packing functions:

- `email_pack_flags()`
- `email_unpack_flags()`
- `email_pack_timezone()`
- `email_unpack_timezone()`
- `body_pack_flags()`
- `body_unpack_flags()`

Though not integrated, I have a test rig that exhaustively tested packing and unpacking **every possible combination of flags**.
- https://github.com/neomutt/test-rigs#header-cache-packing

### DIFF
--- a/hcache/hcache.c
+++ b/hcache/hcache.c
@@ -97,6 +97,90 @@ static size_t header_size(void)
 }
 
 /**
+ * email_pack_flags - Pack the Email flags into a uint32_t
+ * @param e Email to pack
+ * @retval num uint32_t of packed flags
+ *
+ * @note Order of packing must match email_unpack_flags()
+ */
+static inline uint32_t email_pack_flags(const struct Email *e)
+{
+  if (!e)
+    return 0;
+
+  // clang-format off
+  return e->security +
+        (e->expired    << 16) +
+        (e->flagged    << 17) +
+        (e->mime       << 18) +
+        (e->old        << 19) +
+        (e->read       << 20) +
+        (e->replied    << 21) +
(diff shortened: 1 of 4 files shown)


gahr (approved):
Nice!

PR #3903, opened 2023-06-25
replace mutt_mktemp() with buf_mktemp()

roccoblues:
This replaces `mutt_mktemp()` with the Buffer version `buf_mktemp()`, as discussed on the mailing list.

~I've only changed `send/send.c` so far because this is my first contribution and I wanted to get some early feedback. Since `mutt_mktemp()` isn't used in that many places I'd probably do all at once in this PR? Or do you prefer to have it split into smaller chunks?~


flatcap:
> get some early feedback

The code looks good.

> I'd probably do all at once in this PR?

Great

> Or do you prefer to have it split into smaller chunks?

No need for these changes, as they're simple.

attach/recvattach.c
```diff
@@ -553,28 +553,25 @@ static void query_pipe_attachment(const char *command, FILE *fp, struct Body *bo
       msgwin_clear_text();
       return;
     }
-    mutt_mktemp(tfile, sizeof(tfile));
-  }
-  else
-  {
-    tfile[0] = '\0';
```
roccoblues, line 560: I didn't understand why need to NULL terminate here. Please double check.

flatcap: You're right, this isn't needed any more.
It was made redundant by commit 7380cb5fae63481d5d8c10c9a669ebfafcd80ef3 (a year ago), which initialises `tfile`.

nntp/nntp.c
```diff
@@ -2654,6 +2655,7 @@ static bool nntp_msg_open(struct Mailbox *m, struct Message *msg, struct Email *
     }
 
     /* fetch message to cache file */
+    char buf[PATH_MAX] = { 0 };
```
roccoblues, line 2658: `PATH_MAX` is a bit misleading now because the buffer is not used for a path anymore. What do you think?

flatcap: true, but it still needs to be quite large for a message-id
```c
snprintf(buf, sizeof(buf), "ARTICLE %s\r\n", nntp_edata_get(e)->article_num ? article : e->env->message_id);
```
https://github.com/neomutt/neomutt/blob/07c62c30ad149c194440386bd8e43fb506413820/nntp/nntp.c#L2657C1-L2658

flatcap: The RFC doesn't seem to limit the message-id, so 2048 bytes?
(and leave converting the rest of the nntp driver for another day)

roccoblues: 👍🏼 I've used 2048 bytes now.

attach/recvattach.c
```diff
@@ -553,28 +553,25 @@ static void query_pipe_attachment(const char *command, FILE *fp, struct Body *bo
       msgwin_clear_text();
       return;
```
flatcap, line 554: If the user cancels the op, we leak `tfile` here

attach/recvattach.c
```diff
@@ -553,28 +553,25 @@ static void query_pipe_attachment(const char *command, FILE *fp, struct Body *bo
...
-    mutt_mktemp(tfile, sizeof(tfile));
-  }
-  else
-  {
-    tfile[0] = '\0';
+    buf_mktemp(tfile);
   }
 
-  if (mutt_pipe_attachment(fp, body, command, tfile))
+  if (mutt_pipe_attachment(fp, body, command, buf_strdup(tfile)))
```
flatcap, line 559: We want `buf_string()` here, or we'll leak

color/command.c
```diff
@@ -863,17 +863,17 @@ static enum CommandResult color_dump(struct Buffer *buf, struct Buffer *s,
...
-  fp_out = mutt_file_fopen(tempfile, "w");
+  buf_mktemp(tempfile);
+  fp_out = mutt_file_fopen(buf_string(tempfile), "w");
   if (!fp_out)
   {
     // L10N: '%s' is the file name of the temporary file
-    buf_printf(err, _("Could not create temporary file %s"), tempfile);
+    buf_printf(err, _("Could not create temporary file %s"), buf_string(tempfile));
     buf_dealloc(&filebuf);
     return MUTT_CMD_ERROR;
```
flatcap, line 878: We leak `tempfile` here

color/command.c
```diff
@@ -1023,7 +1023,8 @@ static enum CommandResult color_dump(struct Buffer *buf, struct Buffer *s,
   struct PagerData pdata = { 0 };
   struct PagerView pview = { &pdata };
 
-  pdata.fname = tempfile;
+  pdata.fname = buf_strdup(tempfile);
+  buf_pool_release(&tempfile);
```
flatcap, line 1027: This is an odd bit of code...
We've created a couple of temporary structs (`pdata`, `pview`) to pass to `mutt_do_pager()`.
They don't want to take ownership of `tempfile`, they just share it: `buf_string()`.
That means we can't `buf_pool_release()` until after `mutt_do_pager()`.

commands.c
```diff
@@ -735,14 +735,14 @@ enum CommandResult parse_my_hdr(struct Buffer *buf, struct Buffer *s,
...
+  buf_mktemp(tempfile);
 
-  FILE *fp_out = mutt_file_fopen(tempfile, "w");
+  FILE *fp_out = mutt_file_fopen(buf_string(tempfile), "w");
   if (!fp_out)
   {
     // L10N: '%s' is the file name of the temporary file
-    buf_printf(err, _("Could not create temporary file %s"), tempfile);
+    buf_printf(err, _("Could not create temporary file %s"), buf_string(tempfile));
     return MUTT_CMD_ERROR;
```
flatcap, line 746: leak

commands.c
```diff
@@ -753,7 +753,8 @@ enum CommandResult set_dump(ConfigDumpFlags flags, struct Buffer *err)
   struct PagerData pdata = { 0 };
   struct PagerView pview = { &pdata };
 
-  pdata.fname = tempfile;
+  pdata.fname = buf_strdup(tempfile);
+  buf_pool_release(&tempfile);
```
flatcap, line 757: `buf_string()`, then `buf_pool_release()` after `mutt_do_pager()`

gui/global.c
```diff
@@ -93,10 +93,10 @@ static int op_shell_escape(int op)
...
-  char tempfile[PATH_MAX] = { 0 };
-  mutt_mktemp(tempfile, sizeof(tempfile));
+  struct Buffer *tempfile = buf_pool_get();
+  buf_mktemp(tempfile);
 
-  FILE *fp = mutt_file_fopen(tempfile, "a+");
+  FILE *fp = mutt_file_fopen(buf_string(tempfile), "a+");
   if (!fp)
   {
     mutt_perror("fopen");
```
flatcap, line 102: leak

gui/global.c
```diff
@@ -109,7 +109,9 @@ static int op_show_log_messages(int op)
   struct PagerData pdata = { 0 };
   struct PagerView pview = { &pdata };
 
-  pdata.fname = tempfile;
+  pdata.fname = buf_strdup(tempfile);
+
+  buf_pool_release(&tempfile);
```
flatcap, line 114: `buf_string()`, then `buf_pool_release()` after `mutt_do_pager()`

keymap.c
```diff
@@ -1481,12 +1480,13 @@ static enum CommandResult dump_bind_macro(struct Buffer *buf, struct Buffer *s,
...
+  struct Buffer *tempfile = buf_pool_get();
+  buf_mktemp(tempfile);
+  fp_out = mutt_file_fopen(buf_string(tempfile), "w");
   if (!fp_out)
   {
     // L10N: '%s' is the file name of the temporary file
-    buf_printf(err, _("Could not create temporary file %s"), tempfile);
+    buf_printf(err, _("Could not create temporary file %s"), buf_string(tempfile));
     buf_dealloc(&filebuf);
     return MUTT_CMD_ERROR;
```
flatcap, line 1491: leak

keymap.c
```diff
@@ -1498,7 +1498,8 @@ static enum CommandResult dump_bind_macro(struct Buffer *buf, struct Buffer *s,
   struct PagerData pdata = { 0 };
   struct PagerView pview = { &pdata };
 
-  pdata.fname = tempfile;
+  pdata.fname = buf_strdup(tempfile);
+  buf_pool_release(&tempfile);
```
flatcap, line 1502: `buf_string()`, then `buf_pool_release()` after `mutt_do_pager()`

send/send.c
```diff
@@ -3011,11 +3012,11 @@ static bool send_simple_email(struct Mailbox *m, struct EmailArray *ea,
...
-  mutt_mktemp(tempfile, sizeof(tempfile));
+  struct Buffer *tempfile = buf_pool_get();
+  buf_mktemp(tempfile);
   if (body)
   {
-    FILE *fp = mutt_file_fopen(tempfile, "w+");
+    FILE *fp = mutt_file_fopen(buf_string(tempfile), "w+");
     if (!fp)
     {
       email_free(&e);
```
flatcap, line 3022: leak

flatcap (changes requested):
It looks pretty good. so far.
I've made a lot of comments, but it's just a couple of simply-fixed bugs in multiple places.
Thanks!

attach/recvattach.c
roccoblues: Is it ok if I change `mutt_pipe_attachment()`?

```c
-int mutt_pipe_attachment(FILE *fp, struct Body *b, const char *path, char *outfile);
+int mutt_pipe_attachment(FILE *fp, struct Body *b, const char *path, const char *outfile);
```

color/command.c
roccoblues: good catch, thanks! Should be fixed now.

gui/global.c
roccoblues: should be fixed now

keymap.c
roccoblues: should be fixed now

attach/recvattach.c
flatcap: Yep, that's good

roccoblues: fixed now

flatcap:
[I've fixed](https://github.com/neomutt/neomutt/pull/3903/commits/c40a6121f9fab920eedbf88b8b6ede5d292bf7eb) one missing leak and run [iwyu](https://neomutt.org/dev/analysis/iwyu) to tidy the includes.

PR #3904, opened 2023-06-27
Buffer convert: autocomplete

navi-desu:
* **What does this PR do?**
Converts the autocompletion functions to use struct Buffer's

See: #3899

### DIFF
--- a/mutt/buffer.h
+++ b/mutt/buffer.h
@@ -56,6 +56,7 @@ char           buf_at              (struct Buffer *buf, size_t offset);
 bool           buf_str_equal       (struct Buffer *a, struct Buffer *b);
 bool           buf_istr_equal      (struct Buffer *a, struct Buffer *b);
 int            buf_coll            (struct Buffer *a, struct Buffer *b);
+size_t         buf_startswith      (struct Buffer *buf, const char *prefix);
 
 // Functions that APPEND to a Buffer
 size_t         buf_addch           (struct Buffer *buf, char c);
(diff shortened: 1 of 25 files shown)


mutt/buffer.h
```diff
@@ -56,6 +56,8 @@ char           buf_at              (struct Buffer *buf, size_t offset);
 bool           buf_str_equal       (struct Buffer *a, struct Buffer *b);
 bool           buf_istr_equal      (struct Buffer *a, struct Buffer *b);
 int            buf_coll            (struct Buffer *a, struct Buffer *b);
+size_t         buf_startswith      (struct Buffer *buf, const char *prefix);
+size_t         buf_istartswith      (struct Buffer *buf, const char *prefix);
```
gahr, line 60: Do you use these functions anywhere?

flatcap:
Rebased, doxygen, iwyu

flatcap (approved):
Sorry for stepping in but I've finished off this PR.
It turns out that @amudtogal isn't dead :tada: and he's working on his fuzzy matching again,
so I'd like to get this merged so we can integrate his changes.

I haven't actually changed much.
I factored out the helper functions into a preceding commit,
then finished off the remaining `char*,buflen` conversions.

I also made a few idiomatic changes, such as

```diff
- if (buf->data[0] == '\0')
+ if (buf_is_empty(buf))
```

The only notable change was to strip back `buf_mb_wcstombs()`.
Now that it takes a `Buffer`, we can pre-allocate space rather than trying to handle errors when running out of space.

Thanks!

PR #3906, opened 2023-06-29
MxOps path_canon() and path_pretty() take a Buffer 

roccoblues:


flatcap (approved):
That's great.

I was briefly wondering about calling `buf_alloc(buf, PATH_MAX)` to make sure there'd always be enough room, but `_canon()` and `_pretty()` tend to leave the path shorter so I think we'd safe.

Thanks.

PR #3908, opened 2023-06-30
 convert MxOps path_parent() to take a Buffer

roccoblues:
Convert MxOps path_parent() to take a Buffer. Based on the discussion here: #3902.

### DIFF
--- a/compmbox/compress.c
+++ b/compmbox/compress.c
@@ -931,20 +931,14 @@ static int comp_path_pretty(struct Buffer *buf, const char *folder)
  */
 static int comp_path_parent(struct Buffer *path)
 {
-  if (mutt_path_parent(path->data))
-  {
-    buf_fix_dptr(path);
+  if (mutt_path_parent(path))
     return 0;
-  }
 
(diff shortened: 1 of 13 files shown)


compmbox/compress.c
```diff
@@ -929,15 +929,15 @@ static int comp_path_pretty(struct Buffer *buf, const char *folder)
 /**
  * comp_path_parent - Find the parent of a Mailbox path - Implements MxOps::path_parent() - @ingroup mx_path_parent
  */
-static int comp_path_parent(char *buf, size_t buflen)
+static int comp_path_parent(struct Buffer *path)
 {
-  if (mutt_path_parent(buf))
+  if (mutt_path_parent(path->data))
```
roccoblues, line 934: Is this correct? Do we need to call buf_fix_dptr() here?

flatcap: Yes and yes.
As we're likely to be shortening the string, we need to ensure `dptr` isn't left pointing beyond the `NUL`.

roccoblues: ok, that's what I thought

PR #3914, opened 2023-07-03
change mutt_path_canon() to take a Buffer

roccoblues:


flatcap (commented):
You'd renamed some vars, which is fine, but it made the PR harder to read,
so I've factored them out: b8574d205 mxapi: tidy path naming

Historically, `Buffer` params were called `buf` to make their type clear.
Now they're becoming the default, we can name the params more meaningfully.

Every path is now called `path`.

Also, I've rebased and squashed your commits.

mutt/path.c
```diff
@@ -274,50 +274,41 @@ bool mutt_path_tilde(char *buf, size_t buflen, const char *homedir)
...
-                 cwdlen + dirlen + 1, buflen);
-      return false;
-    }
-
-    result[cwdlen] = '/';
-    mutt_str_copy(result + cwdlen + 1, buf, sizeof(result) - cwdlen - 1);
-    mutt_str_copy(buf, result, buflen);
+    size_t cwd_len = mutt_str_len(cwd);
+    cwd[cwd_len] = '/';
+    cwd[cwd_len + 1] = '\0';
```
flatcap, line 304: I've added this to ensure we're still `NUL`-terminated. 
We started with a zeroed buffer, `{ 0 }`, but we can't be sure what `getcwd()` did with the space.

mutt/path.c
```diff
@@ -274,50 +274,41 @@ bool mutt_path_tilde(char *buf, size_t buflen, const char *homedir)
...
+    buf_insert(path, 0, cwd);
   }
 
-  if (!mutt_path_tidy(buf, is_dir))
+  if (!mutt_path_tidy(path->data, is_dir))
     return false;
 
+  buf_fix_dptr(path);
```
flatcap, line 311: Added to fix the dptr, obvs :-)

flatcap (approved):
Nice work.
Thanks!

PR #3918, opened 2023-07-10
change mutt_path_pretty() to take a Buffer

roccoblues:


test/path/mutt_path_pretty.c
```diff
@@ -26,17 +26,40 @@
...
 {
-  // bool mutt_path_pretty(char *buf, size_t buflen, const char *homedir);
+  // bool mutt_path_pretty(struct Buffer *path, const char *homedir);
 
   {
-    TEST_CHECK(!mutt_path_pretty(NULL, 10, "apple", true));
+    TEST_CHECK(!mutt_path_pretty(NULL, "/apple", true));
   }
 
+  // test replacement
```
flatcap, line 39: I've looked at the [code coverage](https://neomutt.org/dev/analysis/coverage) and there are a couple of negative tests we could add:
- Path doesn't match `homedir`
- Path matches `homedir` but differs, e.g. `/home/john` vs `/home/johnsmith`

flatcap (commented):
Looks good.  Thanks!

test/path/mutt_path_pretty.c
roccoblues: Good call. Thanks for pointing me to the coverage docs. Done ✅

flatcap (approved):
Perfect, thanks!

PR #3925, opened 2023-07-11
help: Add message flags.

g0mb4:
Add explanation of message flags to the help screen of index.

### DIFF
--- a/hdrline.c
+++ b/hdrline.c
@@ -72,36 +72,6 @@ struct HdrFormatInfo
 };
 
 /**
- * enum FlagChars - Index into the `$flag_chars` variable ($flag_chars)
- */
-enum FlagChars
-{
-  FLAG_CHAR_TAGGED,         ///< Character denoting a tagged email
-  FLAG_CHAR_IMPORTANT,      ///< Character denoting a important (flagged) email
-  FLAG_CHAR_DELETED,        ///< Character denoting a deleted email
-  FLAG_CHAR_DELETED_ATTACH, ///< Character denoting a deleted attachment
-  FLAG_CHAR_REPLIED, ///< Character denoting an email that has been replied to
-  FLAG_CHAR_OLD,     ///< Character denoting an email that has been read
-  FLAG_CHAR_NEW,     ///< Character denoting an unread email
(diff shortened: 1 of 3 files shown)


g0mb4:
Dear @flatcap ! 

Wow! Thank you for the very detailed help! I'll try to implement it.

g0mb4:
:D thanks for the image

flatcap (changes requested):
The code looks good.
There's only one thing that isn't great and that's not your fault.
It's hard to line up a character on the left with the text on the right.

Please can you insert a label for each of the three blocks, e.g.
```c
format_line(fp, 0, "$to_chars", "", "", wraplen);
```

Thanks!

g0mb4:
Thanks! Yup, I think maybe an alternating background color would help, but that is an other PR. Sure, I'll insert the labels.

flatcap (approved):
Lovely job.  Thanks!

flatcap:
> I think maybe an alternating background color would help

Yes.  It's been discussed before but nobody's quite got around to it.

> but that is an other PR

Yes please :-)

Perhaps a [quick discussion](https://github.com/neomutt/neomutt/discussions/categories/development) first...
Do we want it for just the help, or would other dialogs benefit?
Should it be configurable?
Work out some implemtation details (help available, of course)

The other thing I thought of, if you're interested, was creating an enum for these ugly constants.
They're not used in many places.

https://github.com/g0mb4/neomutt/blob/e42431f4d0563a691e648ccf5862727695637156/help.c#L445-L453

Thanks

asdf8dfafjk:
I came to this pull request searching googling "emojis to_chars neomutt" so for posterity: here's a (totally uncool) to_chars. This takes a bit to get hang of but after a while there is less mental effort than recollecting initials. (this is a guess, I can't actually get this to work yet lol)


set to_chars    = "🍌🥭🍉🍅🍪🍇🥥"

# Mnenomic

Banana has no seeds (=you're not a recepient), 

mango has just one (=you)

water melon has lots (you're one of many)

tomato stands for CC here

cookie means you cooked the mail

Grapes stands for mailing list 

Coconut is you are in reply-to




Also 

set flag_chars  = "🏷️❗🗑️🚮↩🙈🆕🙉🆕📜📜"

PR #3932, opened 2023-07-13
Alternating colors in help

g0mb4:
This PR add alternating colors for the help menu.

image

New config variables: `$help_alternate`
New colors: `pager_even`, `pager_odd`.

To enable this feature add this to `.muttrc`:
```bash
set help_alternate = yes
color pager_odd  blue cyan
color pager_even black white
```

### DIFF
--- a/color/color.h
+++ b/color/color.h
@@ -73,6 +73,8 @@ enum ColorId
 #endif
   MT_COLOR_SIGNATURE,                ///< Pager: signature lines
   MT_COLOR_STATUS,                   ///< Status bar (takes a pattern)
+  MT_COLOR_STRIPE_EVEN,              ///< Stripes: even lines of the Help Page
+  MT_COLOR_STRIPE_ODD,               ///< Stripes: odd lines of the Help Page
   MT_COLOR_TILDE,                    ///< Pager: empty lines after message
   MT_COLOR_TREE,                     ///< Index: tree-drawing characters
   MT_COLOR_UNDERLINE,                ///< Underlined text
--- a/color/command.c
+++ b/color/command.c
@@ -93,6 +93,8 @@ const struct Mapping ColorFields[] = {
 #endif
   { "signature",         MT_COLOR_SIGNATURE },
   { "status",            MT_COLOR_STATUS },
+  { "stripe_even",       MT_COLOR_STRIPE_EVEN},
+  { "stripe_odd",        MT_COLOR_STRIPE_ODD},
   { "tilde",             MT_COLOR_TILDE },
   { "tree",              MT_COLOR_TREE },
   { "underline",         MT_COLOR_UNDERLINE },
--- a/color/simple.c
+++ b/color/simple.c
@@ -50,6 +50,7 @@ void simple_colors_init(void)
   SimpleColors[MT_COLOR_SIDEBAR_HIGHLIGHT].attrs = A_UNDERLINE;
 #endif
   SimpleColors[MT_COLOR_STATUS].attrs = A_REVERSE;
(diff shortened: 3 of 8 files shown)


flatcap (approved):
You've done a great job!
Well done understanding the colours in the Pager!
It works, the code looks good, you've documented the feature.

Thanks!

The only thing I'm still thinking about is the default behaviour.

When we introduce new features, few of our users will actually see the Release Notes, so often they miss out.

I'll put my thoughts below and ask the dev mailing list what they think.

flatcap:
I think this feature is too useful to leave if off-by-default.
Possibly even dropping the `$help_alternate` config variable.

If we enable it, then we need to pick some default colours that won't offend too many users :-)
Because we don't know the users' background colours, that probably limits us to attributes.

Below are some sample attributes and colour schemes.

flatcap:
### plain.rc

```sh
set help_alternate = no
```



---

### reverse.rc

```sh
set help_alternate = yes
```



---

### bold.rc

```sh
set help_alternate = yes
color pager_even bold default default
color pager_odd       default default
```



---

### underline.rc

```sh
set help_alternate = yes
color pager_even underline default default
color pager_odd            default default
```



---

### colour.rc

```sh
set help_alternate = yes
color pager_odd  blue  cyan
color pager_even black white
```



---

### light-grey.rc

```sh
set help_alternate = yes
color pager_even default color253
color pager_odd  default default
```



---

### dark-grey.rc

```sh
set help_alternate = yes
color pager_even default color234
color pager_odd  default default
```

g0mb4:
Wow! These looks really nice, but I'll leave this decision to you or to the community, I'm really bad at visual design :D

g0mb4:
I would not drop the config variable, when I was testing it some color combinations and the fast scrolling felt weird, I can imagine it can be more unpleasant or it could cause seizure for someone.

roccoblues:
My 2ct:

Underline looks horrible and bold is confusing (makes those options look more important). I wouldn't use that as default.

The tricky part with picking a default color is that it has to work with light and dark terminals. Do we have default colors somewhere else? Maybe we can use them here as well?

Regarding the "enabled by default"? I don't have a strong opinion in either direction.

lucilanga:
I'm with rocco on this, underline does not work!
I cannot get bold working, I do not know why.
I'm partial to grey thing.
And yes depending on your alignment dark/light .. the default might not be everyone's cup of tea.

flatcap:
> I'm with rocco on this, underline does not work!

Yeah and underline on every line is a bit overwhelming.

> I cannot get bold working, I do not know why.

Some terminals have setting for boldness: disallow, allow, convert into brightness.
In the [test-ansi](https://github.com/neomutt/test-ansi) repo there's `attrs.sh`
That displays a lot of possible attribute settings.

<details>
<summary>Sample output</summary>

image

</details>

lucilanga:
> Some terminals have setting for boldness: disallow, allow, convert into brightness.

Thank you!
As for the bold .. hmm .. not bad! not bad!
I like it!

urxvt, xterm, alacritty
screenshot-2023-07-23_01-01-14

flatcap:
I've made some small changes.
Your original code is [archived to NeoMutt-old](https://github.com/neomutt/neomutt-old).

- Drop the config option
  I'd really like this to be on by default.
  If we encounter problems, I'll put it back.

- Default colour: **bold**.
  It needed to be an attribute so that it was independent of the colour scheme.
  This is the only reasonable attribute.

- Renamed colours to **`stripe_even`**, **`stripe_odd`**.
  Yes, we're using the Pager, but to the user they're only appearing in the Help.
  With a generic name, we might find a use for them elsewhere.

Thanks!

PR #3933, opened 2023-07-14
test: disable faulty editor tests

flatcap:
These failing tests involve non-ascii characters and emojis.

@roccoblues
Yep, you're right.  The tests are wrong.

I found two problems in the code: combining characters and `iswalnum()`.

A simple case for combining characters is 0xFE0F (Variation Selector-16).
This turns, `0xE2 9C 89` ( `✉` ) into `0xE2 9C 89 EF B8 8F` (`✉️`)
In the wide buffer, the "emoji" envelope is stored as two characters.

If 0xFE0F isn't recognised as a combining character then some functions will not work correctly.

`iswalnum()` is used to determine where "words" start and end.
On Linux, `iswalnum()' on Chinese characters returns true, on macOS it returns false.

At some point we need to:
- Define the behaviour code wider character sets
- Devise lots of tests to verify the behaviour
- Create a clean system to workaround the limitations of certain libc's

('we' does not necessarily mean us, and 'at some point' could be in the far future :-)

### DIFF
--- a/.github/workflows/macos.yml
+++ b/.github/workflows/macos.yml
@@ -50,7 +50,6 @@ jobs:
       run: make -j ${{steps.cpu-cores.outputs.count}} test/neomutt-test
 
     - name: Run Tests
-      continue-on-error: true
       run: |
         export NEOMUTT_TEST_DIR=$GITHUB_WORKSPACE/test-files
         make test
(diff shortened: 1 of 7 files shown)


roccoblues (approved):
That makes sense, thanks for the explanation! I was basically just guessing around without understanding all implications.

PR #3934, opened 2023-07-17
functions: move <show-version> to generic

flatcap:
All of NeoMutt functions are divided into blocks, e.g. [`OpAlias`](https://github.com/neomutt/neomutt/blob/5bbc865600107986f20066c2e4169ad1866c117d/functions.c#L55-L64).
Each of these blocks is handled by a function dispatcher, e.g. [`alias_function_dispatcher()`](https://github.com/neomutt/neomutt/blob/5bbc865600107986f20066c2e4169ad1866c117d/alias/functions.c#L332-L357).

Function `<show-version>` is implemented by the [`global_function_dispatcher()`](https://github.com/neomutt/neomutt/blob/5bbc865600107986f20066c2e4169ad1866c117d/gui/global.c#L165-L185), so it belongs in [`OpGeneric`](https://github.com/neomutt/neomutt/blob/5bbc865600107986f20066c2e4169ad1866c117d/functions.c#L289-L345).

This PR moves the function out of [`OpIndex`](https://github.com/neomutt/neomutt/blob/5bbc865600107986f20066c2e4169ad1866c117d/functions.c#L350-L509) .

**Update**: Just `OpIndex`;  `<show-version>` needs to stay in OpPager, for now, because the Pager doesn't use `OpGeneric`. 

----

See Also: [NeoMutt Keybindings](https://gist.github.com/flatcap/fc84519ad7361f6e093eadda039aa0eb)

### DIFF
--- a/functions.c
+++ b/functions.c
@@ -333,6 +333,7 @@ const struct MenuFuncOp OpGeneric[] = { /* map: generic */
   { "select-entry",                  OP_GENERIC_SELECT_ENTRY },
   { "shell-escape",                  OP_SHELL_ESCAPE },
   { "show-log-messages",             OP_SHOW_LOG_MESSAGES },
+  { "show-version",                  OP_VERSION },
   { "tag-entry",                     OP_TAG },
   { "tag-prefix",                    OP_TAG_PREFIX },
   { "tag-prefix-cond",               OP_TAG_PREFIX_COND },
@@ -466,7 +467,6 @@ const struct MenuFuncOp OpIndex[] = { /* map: index */
   { "save-message",                  OP_SAVE },
   { "set-flag",                      OP_MAIN_SET_FLAG },
   { "show-limit",                    OP_MAIN_SHOW_LIMIT },
-  { "show-version",                  OP_VERSION },
 #ifdef USE_SIDEBAR
   { "sidebar-first",                 OP_SIDEBAR_FIRST },
   { "sidebar-last",                  OP_SIDEBAR_LAST },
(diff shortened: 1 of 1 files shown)


lucilanga (commented):
I believe you missed OpPager.

lucilanga:
The way things are wired in km_dokey_event() I believe we can do this changes without any impact to the user.
There are some duplicated functions which might need cleanup like redraw-screen, that's a generic one isn't it?
Do you plan on doing the whole separation right now? tho I'm not entirely sure it matters.

flatcap:
> I believe you missed OpPager.

I took it out of OpPager, then my build failed, so I put it back :-)
The build of the docs is very strange.
I'll see if I can do a better job tomorrow.

flatcap:
> The way things are wired in km_dokey_event() I believe we can do this changes without any impact to the user.

The PR only **adds** functions to Dialogs, so the user shouldn't notice.
(it's unlikely that they've mapped `<show-version>`)

> There are some duplicated functions which might need cleanup like redraw-screen, that's a generic one isn't it?

Yes, that should be in OpGeneric only.

> Do you plan on doing the whole separation right now? tho I'm not entirely sure it matters.

Not yet as I haven't figured out how `bind pager ...` will work, if we've split up `OpPager`!
More diagrams are needed :-)

lucilanga:
> I took it out of OpPager, then my build failed, so I put it back :-) The build of the docs is very strange. 

Indeed, I was running '--disable-docs' configure never noticing the issue.

PR #3938, opened 2023-07-20
simple sprintf() -> snprintf() fixes

roccoblues:


ossilator:
that doesn't seem right. if you want to be that pedantic, you should also add asserts that the buffer wasn't exceeded.

(in mbsync i have nfsnprintf() and similar functions which include assert-like code; "nf" meaning "no fail".)

roccoblues:
> that doesn't seem right. if you want to be that pedantic

Sorry I don't understand. For example `keymap.c`: can you elaborate what the difference is between the one occurence I changed and [the other ones](https://github.com/neomutt/neomutt/pull/3938/files#diff-cd7e6e8161011fe6598cb583b79ccf5733b4ccd457bfa79ff755aabd19897915L866)?

ossilator:
> can you elaborate what the difference is

there is none. the comment wasn't exclusive.

flatcap:
Git has a policy to avoid `sprintf()` and others,

https://github.com/git/git/blob/master/banned.h

> This header lists functions that have been banned from our code base,
> because they're too easy to misuse (and even if used correctly, complicate audits).

Code gets cut'n'pasted all the time, so even though the old code wasn't unsafe,
the new code is better example for others.

flatcap:
> in mbsync i have nfsnprintf() and similar functions

Do you have link for them?

ossilator:
> Do you have link for them?

https://sourceforge.net/p/isync/isync/ci/master/tree/src/util.c#l609

flatcap:
Thanks for the feedback.
I'm happy to accept more fixes for 'banned' functions.
We already have a few non-fail functions like `mutt_mem_malloc()`, but more would be better.
Suggestions, and PRs, welcome :-)

flatcap (approved):

PR #3940, opened 2023-07-22
gui: refactor resize handling

flatcap:
When the terminal is resized it sends a `SIGWINCH` signal to NeoMutt.
This is caught in `curses_signal_handler()` which sets a global variable: `SigWinch = true`.
The signal's arrival causes `getch()` to return immediately with a timeout.

What happens next depends on which Window has focus.

In some Dialogs, the event loop clears the variable `SigWinch = false` and manually forces a refresh of the entire screen.
Others don't.
While _they_ are waiting for keystrokes, the resized screen won't get repainted.

The signal handler now sends a notification to a new notifier: `NeoMutt.notify_resize`.
The Root Window has set up an observer of this notification.

The Root Window, now:
- Looks up the new Terminal size
- Reflows all the Windows (redistributing the space)
- Recalculates all the Windows
- Repaints all the Windows

### DIFF
--- a/mutt/notify.c
+++ b/mutt/notify.c
@@ -37,11 +37,11 @@
 /// Lookup table for NotifyType
 /// Must be the same size and order as #NotifyType
 static const char *NotifyTypeNames[] = {
-  "NT_ALL",     "NT_ACCOUNT",  "NT_ALIAS",   "NT_ALTERN", "NT_ATTACH",
-  "NT_BINDING", "NT_COLOR",    "NT_COMMAND", "NT_CONFIG", "NT_CONTEXT",
-  "NT_EMAIL",   "NT_ENVELOPE", "NT_GLOBAL",  "NT_HEADER", "NT_INDEX",
-  "NT_MAILBOX", "NT_MENU",     "NT_PAGER",   "NT_SCORE",  "NT_SUBJRX",
-  "NT_WINDOW",
+  "NT_ALL",      "NT_ACCOUNT", "NT_ALIAS",   "NT_ALTERN", "NT_ATTACH",
+  "NT_BINDING",  "NT_COLOR",   "NT_COMMAND", "NT_CONFIG", "NT_EMAIL",
+  "NT_ENVELOPE", "NT_GLOBAL",  "NT_HEADER",  "NT_INDEX",  "NT_MAILBOX",
+  "NT_MVIEW",    "NT_MENU",    "NT_RESIZE",  "NT_PAGER",  "NT_SCORE",
+  "NT_SUBJRX",   "NT_WINDOW",
 };
 
 /**
--- a/mutt_signal.c
+++ b/mutt_signal.c
@@ -76,10 +76,12 @@ static void curses_signal_handler(int sig)
       /* We don't receive SIGWINCH when suspended; however, no harm is done by
        * just assuming we received one, and triggering the 'resize' anyway. */
       SigWinch = true;
+      notify_send(NeoMutt->notify_resize, NT_RESIZE, 0, NULL);
       break;
 
     case SIGWINCH:
       SigWinch = true;
+      notify_send(NeoMutt->notify_resize, NT_RESIZE, 0, NULL);
       break;
 
     case SIGINT:
--- a/core/neomutt.c
+++ b/core/neomutt.c
@@ -54,6 +54,9 @@ struct NeoMutt *neomutt_new(struct ConfigSet *cs)
   n->sub->cs = cs;
   n->sub->scope = SET_SCOPE_NEOMUTT;
 
+  n->notify_resize = notify_new();
+  notify_set_parent(n->notify_resize, n->notify);
+
   return n;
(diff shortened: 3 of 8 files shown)


mutt_signal.c
```diff
@@ -76,10 +76,12 @@ static void curses_signal_handler(int sig)
       /* We don't receive SIGWINCH when suspended; however, no harm is done by
        * just assuming we received one, and triggering the 'resize' anyway. */
       SigWinch = true;
+      notify_send(NeoMutt->notify_resize, NT_RESIZE, 0, NULL);
```
ossilator, line 79: doing that from signal context is a recipe for disaster.

the "easy" way out is to handle SigWinch more consistently, possibly posting the notification from there (you could centralize that in a check_signals() function invoked after each syscall which is a cancellation point).

the proper way is to use a self-pipe, which basically requires a proper event loop (the event loop itself can be had rather trivially, as you can see in https://sourceforge.net/p/isync/isync/ci/master/tree/src/util.c#l1090 - but fully converting the code base to an event driven architecture is anything but trivial).

mutt/notify.c
```diff
@@ -37,11 +37,11 @@
 /// Lookup table for NotifyType
 /// Must be the same size and order as #NotifyType
```
ossilator, line 38: for the size part, you can use static_assert().

mutt/notify.c
```diff
@@ -37,11 +37,11 @@
...
 static const char *NotifyTypeNames[] = {
-  "NT_ALL",     "NT_ACCOUNT",  "NT_ALIAS",   "NT_ALTERN", "NT_ATTACH",
-  "NT_BINDING", "NT_COLOR",    "NT_COMMAND", "NT_CONFIG", "NT_CONTEXT",
-  "NT_EMAIL",   "NT_ENVELOPE", "NT_GLOBAL",  "NT_HEADER", "NT_INDEX",
-  "NT_MAILBOX", "NT_MENU",     "NT_PAGER",   "NT_SCORE",  "NT_SUBJRX",
-  "NT_WINDOW",
+  "NT_ALL",      "NT_ACCOUNT", "NT_ALIAS",   "NT_ALTERN", "NT_ATTACH",
```
ossilator, line 40: this drudgery can be avoided by auto-generating the labels. an example can be found in https://sourceforge.net/p/isync/isync/ci/17db5de0cadf15dd52ebead3e45a0507e8c774c9/ (which requires https://sourceforge.net/p/isync/isync/ci/6a78e2c5f61fdc1ec18a87f05f800d22ca59dfa6/ (which is aimed at bitwise enums, but extending it to support regular sequences (SEQ_ENUM?) would be no big deal)).

mutt/notify.c
flatcap: good idea!

mutt/notify.c
flatcap: Since the split from Mutt, I've worked hard to eliminate lots of custom steps in the build.
We could probably copy the macro magic from opcodes.h:

https://github.com/neomutt/neomutt/blob/2f2741f62db1a7662d0b1684bd2036aab700323f/opcodes.h#L36-L37

It generates an enum and a matching string array.

mutt_signal.c
flatcap: > doing that from signal context is a recipe for disaster.

argh!  I should have realised that.

> ... posting the notification from there ...
> ... after each syscall which is a cancellation point ...

`SIGWINCH` -\> `notify_send()` -\> sets flag on `RootWindow`
`SIGWINCH` -\> interrupts `getch()` -\> timeout notification
`RootWindow` observes timeout does reflow/repaint

That should be more robust.
 
> but fully converting the code base to an event driven architecture is anything but trivial

I know.  I've been trying for seven years now :-)
We _have_ made a lot of progress in that time, but there's still much to do.

mutt/notify.c
ossilator: > I've worked hard to eliminate lots of custom steps in the build.

for which reasons in particular?

> We could probably copy the macro magic from opcodes.h

hmm, yeah, the "macro visitor" pattern. mildly ugly, but quite powerful.

i've meanwhile come up with a solution to automatically enumerate bits, so i could actually adapt this. thanks for the inspiration!

flatcap: > for which reasons in particular?

It was fiendishly complicated and quite fragile.
The docs build is still the worst part.
Here are two (fairly current) diagrams of our build process: https://neomutt.org/dev/build/make

And this is the original build process:

ossilator: i'm a build system guy, so this doesn't scare me. :joy: 

my primary concern about code generation is that tools (e.g., IDE code models) don't work right until the project was built, and even then the results might be suboptimal. but deep macro magic doesn't exactly help with debugging and IDE "follow symbol" functionality, either, so it's really a bit of a "choose your death" situation.

flatcap:
superseded by #4021

PR #3946, opened 2023-07-31
change sprintf() to mutt_str_asprintf() where possible

roccoblues:
This replaces a couple more `sprintf()`s with `mutt_str_asprintf()`.


flatcap (approved):
Goodbye 6 ugly bits of code.
Thanks!

PR #3951, opened 2023-08-02
bring IndexSharedData notifications back

### LINKED ISSUE #3950: toggle-write doesn't update %-char in %r expando of status_format
How to reproduce:
1. open a mailbox
2. set `status_format` to something that includes `%r` 
3. press <kbd>%</kbd> or call `toggle-write`
4. see how the status bar *isn't* updated
5. press <kbd>ctrl+L</kbd> or call `redraw-screen`
6. see how the status bar *is* updated

roccoblues:
fixes #3950

### DIFF
--- a/index/shared_data.c
+++ b/index/shared_data.c
@@ -56,7 +56,7 @@ static int index_shared_mview_observer(struct NotifyCallback *nc)
     shared->mailbox_view = NULL;
 
   mutt_debug(LL_NOTIFY, "relay NT_MVIEW to shared data observers\n");
-  notify_send(shared->notify, nc->event_type, nc->event_subtype, shared);
+  notify_send(shared->notify, nc->event_type, nc->event_subtype, nc->event_data);
   return 0;
 }
 
@@ -81,7 +81,7 @@ static int index_shared_account_observer(struct NotifyCallback *nc)
     shared->account = NULL;
 
   mutt_debug(LL_NOTIFY, "relay NT_ACCOUNT to shared data observers\n");
-  notify_send(shared->notify, nc->event_type, nc->event_subtype, shared);
+  notify_send(shared->notify, nc->event_type, nc->event_subtype, nc->event_data);
   return 0;
 }
 
@@ -106,7 +106,7 @@ static int index_shared_mailbox_observer(struct NotifyCallback *nc)
     shared->mailbox = NULL;
 
   mutt_debug(LL_NOTIFY, "relay NT_MAILBOX to shared data observers\n");
-  notify_send(shared->notify, nc->event_type, nc->event_subtype, ev_m);
+  notify_send(shared->notify, nc->event_type, nc->event_subtype, nc->event_data);
   return 0;
 }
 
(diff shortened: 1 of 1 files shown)


roccoblues:
Here's what I understood so far:

Old version:
The actual event source `mx_toggle_write()` send a notification on the `Mailbox`. Those got picked up by the `index_shared_mailbox_observer()` that is added to the mailbox whenever it changes in the `IndexSharedData`. The relay then forwarded the notification to all observers on the `IndexSharedData`. In this particular case the `ibar_index_observer()`.

Challenges for a fix without the relay:
* `mx_toggle_write()` only knows about the `Mailbox`. It can send notifications on that or on `NeoMutt`
* When the `ibar_index_observer` is attached we have `IndexSharedData` so in theory we could listen ob `shared->mailbox`. However that fails because (*I think*) the code is being called before the `IndexSharedData` is setup.

For now I've simply changed the notifications `NeoMutt` which fixes the issue. I don't fully understand the consequences of that though.

roccoblues:
I've added an additional check to only redraw if the notification is about the currently active mailbox. I'm not sure how much that helps because I'd assume the bigger problem is that we're listing on `NT_ALL` which will trigger (potential) redraws on all kinds of notifications.

flatcap:
I've remembered, or worked out again, why the relaying was dodgy.

With the old relay code in place, here's a sequence of events:

**Setup**:
- Index created and given a `Mailbox`
- Index creates `IndexSharedData` with a copy of the current `Account`, `Mailbox`, `Email`
- `IndexSharedData` observes each object and is itself an observable object
- `IndexBar` (status bar) observes the `IndexSharedData`, listening for `NT_ALL` **any** events

**Action**:
- User hits <kbd>%</kbd> (`<toggle-write>`) translated into `OP_TOGGLE_WRITE`
- Index calls `op_toggle_write()` which alters the current `Mailbox`
- Function sends notification via the `Mailbox`

```c
struct EventMailbox ev_m = { m };
notify_send(NeoMutt->notify, NT_MAILBOX, NT_MAILBOX_CHANGE, &ev_m);
```

**Reaction**:
- `IndexSharedData` receives an event notification, `index_shared_mailbox_observer()`
  - Type: `NT_MAILBOX`
  - Subtype: `NT_MAILBOX_CHANGE`
  - Data: `EventMailbox`
- `IndexSharedData` sends a notification on itself:
  `notify_send(shared->notify, nc->event_type, nc->event_subtype, ev_m);`
- `IndexBar` (status bar) receives an event notification, `ibar_index_observer()`
  - Type: `NT_MAILBOX`
  - Subtype: `NT_MAILBOX_CHANGE`
  - Data: `EventMailbox`
- `IndexBar` updates itself

This is correct and _should_ work.
What's broken are the handlers that do _this_:
(see the last argument)

```c
notify_send(shared->notify, nc->event_type, nc->event_subtype, shared);
```

Here, we're sending event data that doesn't match the event type.
I probably thought that giving the observer an `IndexSharedData` pointer would be useful.

The correct solution is:

```c
notify_send(shared->notify, nc->event_type, nc->event_subtype, nc->event_data);
```

and `index_shared_mailbox_observer()` should be changed for consistency.

roccoblues:
Nice detective work 👏🏼. I brought the relay observers back and fixed the payload.

flatcap (approved):

PR #3954, opened 2023-08-04
add support for multiple keymap/menufunc to km_dokey_event()

lucilanga:
I just wanted to get the ball rolling on this.
On the long run, if this is going to fly I believe it requires a third component.
That is the menu domain we've been talking the last time.
I'm not fully certain how this is going to fall into place.
We were discussing having :

DomainPager = { OpPager, pager_function_dispatcher() }

but perhaps we can do a global domain table where we could find our dispatcher() based on the calling parameters.

    void *hsmap[] = { &Keymaps[MENU_FOLDER], &Keymaps[MENU_GENERIC], 0 };
    void *hsfuncs[] = { OpBrowser, OpGeneric, 0 };
    op = km_dokey2(hsmap, hsfuncs);
 
obviously OpBrowser will trigger browser_function_dispatcher() and OpGeneric will trigger global dispatcher.
There is the matter of menu dispatcher called infrequently or index dispatcher for that matter, I'm not sure what to do about them.

Moving on, we import OpAlias, OpGeneric from functions.h but we have a local functions.h in each dialog's directory, perhaps we can move those to keymap.h.

I was setup to use structs for function prototypes but I'd fallen on void. I was not sure about the final prototype and wanted to avoid: { { param}, {param}, {0} }. I've kept the original prototypes in keymap.h though. Also I've made km_dokey_event2() as per your initial thought on  this, it's easier to switch back.

Finally I've ignored mw_get_field() and the _notify part for that matter as well, I believe the whole PR is going to be in need for a makeover. Though the way I'm dragging things perhaps I should consider doing the sidebar things the way are right now and get it out of the way.


flatcap:
Superseded by #4717 - Key: Break up dependencies

PR #3956, opened 2023-08-06
fix 'from' address when real name isn't set

flatcap:
When `$from` isn't set, NeoMutt tries to construct a `From:` address using, the user's _username_ and the machine's _hostname_.
Ensure that we have a `Buffer` to put them in.

When `$real_name` isn't set `Address.personal` should be `NULL`.
Ensure we don't allocate an empty `Buffer`.

**Steps**:
- Run NeoMutt with zero config:
  `./neomutt -n -F /dev/null`
- Compose an email

**Expected Behaviour**:
- `From:` field contains `username@hostname.com`

**Actual Behaviour**:
- `From:` field contains `<;`

### DIFF
--- a/send/send.c
+++ b/send/send.c
@@ -1467,21 +1467,21 @@ struct Address *mutt_default_from(struct ConfigSubset *sub)
    */
 
   const struct Address *c_from = cs_subset_address(sub, "from");
-  const bool c_use_domain = cs_subset_bool(sub, "use_domain");
   if (c_from)
   {
     return mutt_addr_copy(c_from);
   }
-  else if (c_use_domain)
+
+  char domain[1024] = { 0 };
+  const char *mailbox = Username;
+  const bool c_use_domain = cs_subset_bool(sub, "use_domain");
+  if (c_use_domain)
   {
-    struct Address *addr = mutt_addr_new();
-    buf_printf(addr->mailbox, "%s@%s", NONULL(Username), NONULL(mutt_fqdn(true, sub)));
-    return addr;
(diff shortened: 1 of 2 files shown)


roccoblues (approved):
I reproduced the broken behavior and the fix. 👍🏼

send/send.c
```diff
@@ -1475,6 +1475,7 @@ struct Address *mutt_default_from(struct ConfigSubset *sub)
   else if (c_use_domain)
   {
     struct Address *addr = mutt_addr_new();
+    addr->mailbox = buf_new(NULL);
```
gahr, line 1478: Perhaps we should first come up with a string for the mailbox, wither this or `Username` as in the else branch, then call `mutt_addr_create` outside the branch?

gahr (approved):

send/send.c
flatcap: how about (untested)

```c
struct Address *mutt_default_from(struct ConfigSubset *sub)
{
  const struct Address *c_from = cs_subset_address(sub, "from");
  if (c_from)
  {
    return mutt_addr_copy(c_from);
  }

  char domain[1024] = { 0 };
  const char *mailbox = Username;
  const bool c_use_domain = cs_subset_bool(sub, "use_domain");
  if (c_use_domain)
  {
    snprintf(domain, sizeof(domain), "%s@%s", NONULL(Username), NONULL(mutt_fqdn(true, sub)));
    mailbox = domain;
  }

  return mutt_addr_create(NULL, domain);
}
```

PR #3957, opened 2023-08-06
fix mview leak

flatcap:
Fix a race condition that caused a `MailboxView` to not be freed.

**Steps**:
- Run NeoMutt with zero config
  `./neomutt -n -F /dev/null`
- Exit NeoMutt

---

`op_quit()` calls `mview_free(shared->mailbox_view)`

`mview_free()` sends a notification that's observed by the `IndexSharedData` which then does `shared->mailbox_view = NULL;`

This clears the very pointer that `mview_free()` was about to `free()`.

Make a copy of the pointer and free _that_ after the notification.

### DIFF
--- a/mview.c
+++ b/mview.c
@@ -65,7 +65,8 @@ void mview_free(struct MailboxView **ptr)
   FREE(&mv->pattern);
   mutt_pattern_free(&mv->limit_pattern);
 
-  FREE(ptr);
+  *ptr = NULL;
+  FREE(&mv);
 }
 
 /**


roccoblues:
> Fix a race condition that caused a `MailboxView` to not be freed.
> 
> **Steps**:
> 
>     * Run NeoMutt with zero config
>       `./neomutt -n -F /dev/null`
> 
>     * Exit NeoMutt

Your description makes total sense but I can't seem to reproduce this. I guess this is timing related? Or can you reproduce it every time?

flatcap:
> I guess this is timing related?

I didn't _think_ so.

> Or can you reproduce it every time?

Yes...

Ah!  NeoMutt needs to open a mailbox for it to leak.
Even with zero config, it was finding the empty file `/var/spool/mutt`

It also leaks with: `./neomutt -n -F /dev/null -f /dev/null`

roccoblues (approved):
leak and fix reproduced. 👍🏼

PR #3972, opened 2023-08-14
remove unecessary e->env checks

roccoblues:
I ran this through Coverity and it didn't show any new defects. I also tried to follow the callers to see if we didn't initialize the envelope anywhere but that proved impossible.

Do we want to try this out?


flatcap (approved):
Lovely, thanks!

PR #3974, opened 2023-08-14
remove dead code in nntp_select_server()

roccoblues:
`adata` is only initialized with NULL and never touched again until this check.

I suggest we remove it if we don't want need this code to make merging (upstream) nntp changes easier

### DIFF
--- a/nntp/newsrc.c
+++ b/nntp/newsrc.c
@@ -1080,30 +1080,6 @@ struct NntpAccountData *nntp_select_server(struct Mailbox *m, const char *server
     conn->account.user[0] = '\0';
   }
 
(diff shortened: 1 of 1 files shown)


flatcap:
I found the source of the unusable code.
Four years ago, the refactoring of the `Connection`s led to this change:

> cd3ea66ae drop Connections

```diff
-  adata = conn->data;
+  // adata = conn->data;
```

Then recently, some automatic tidying cleared away the comment:

> e3773a736 drop commented-out code

```diff
-  // adata = conn->data;
```

Given that nobody's complained in 4 years, I think we're safe :-)

flatcap (approved):

PR #3975, opened 2023-08-14
fix wrong printf argument in alias_complete()

roccoblues:
This was probably missed during the auto-completion buffer transition (https://github.com/neomutt/neomutt/discussions/3899).

Fixes: https://github.com/neomutt/neomutt/security/code-scanning/589

### DIFF
--- a/alias/dlg_alias.c
+++ b/alias/dlg_alias.c
@@ -425,7 +425,7 @@ int alias_complete(struct Buffer *buf, struct ConfigSubset *sub)
     {
       /* fake the pattern for menu title */
       char *mtitle = NULL;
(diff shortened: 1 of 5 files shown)


flatcap (changes requested):
Nice!  But we can go further.

Please can you add attributes to all the printf-style-functions, like we do here:

https://github.com/neomutt/neomutt/blob/e90854333be1517577c42a11831e6a554bb115c2/mutt/logging2.h#L69C1-L70

Thanks!

roccoblues:
> Please can you add attributes to all the printf-style-functions

Done. Thanks for the hint - that makes total sense. 👍🏼

flatcap (approved):
That's great (and they all warn correctly)

PR #3976, opened 2023-08-15
mutt_file_sanitize_filename: actually replace characters

nabijaczleweli:
* **What does this PR do?**
The documentation says "characters" but it operates on bytes. This means that Рахунок.pdf and Ziemiańska.pdf get turned into ______________.pdf and Ziemia__ska.pdf, which is, respectively, very useful and offensive to me personally.

Instead, keep the old logic for bytes <=0x7F to get rid of quotes or slashes or whatever, but allow all other characters, since they won't interfere with whatever mangling may happen to the path.

a
b


ossilator:
hmm ... bidi controls, zero-width joiners, etc. can be used for all kinds of mischief. i don't think one can use this to cause much actual harm in the case of file names in particular, but someone™ should double-check.
https://unicode.org/reports/tr36/ (and tr39) have the gory details.

nabijaczleweli:
this is supposed to protect the user from filenames with ><'"(){}*?;$IFS/ in them in case they don't get quoted before being pasted into a shell program, not protect the user from seeing their name. the "bad things" you can do here are "make the text render correctly". (even this approach is over-zealous since filenames are sequences of bytes, but reinterpreting this as "sequences of characters in the current locale" is just-about-reasonable for purposes of display, and gv really hates getting invalid characters in the filename)

nabijaczleweli:
CI failure appears unrelated.

flatcap:
> CI failure appears unrelated.

Yes, that was fixed this morning, in ade46c5ae

nabijaczleweli:
rebased, let's see

flatcap:
`test_mutt_file_sanitize_filename()` is failing on the BSD box

nabijaczleweli:
oh, I wrote the test against glibc's broken locales and forgot they were broken. they should apply my v17 so this doesn't happen in future. fixed now

flatcap (approved):

Issue #3924, opened 2023-07-11
"Random" SIGSEGV in index screen after closing pager


Werkov:
<!--
Sorry you've found a bug, but thanks for reporting it here.

Please try to reproduce the bug with a version of NeoMutt built off a recent
commit: we routinely spot and fix bugs. The issue you're seeing might have
already been solved since the last official release.

Please try to give as much information as you can.
-->

## Expected Behaviour

No segfault.

## Actual Behaviour

neomutt segfaulted and exited.

If NeoMutt crashed, did your OS create a 'coredump' file? Yes, call stack below.

## Steps to Reproduce

Alas, no clear reproducer. It happened with particular folder content upon deleteing (with macros) few messages and switching between index and pager view. I could not reproduce it at will.

## How often does this happen?

- Sometimes

## When did it start to happen?

- When I upgraded
  neomutt wasn't part of the update, system libraries were updated though. (Also not sure if it didn't happen before update or I only didn't notice.)


## NeoMutt Version

```
NeoMutt 20230517
Copyright (C) 1996-2022 Michael R. Elkins and others.
NeoMutt comes with ABSOLUTELY NO WARRANTY; for details type 'neomutt -vv'.
NeoMutt is free software, and you are welcome to redistribute it
under certain conditions; type 'neomutt -vv' for details.

System: Linux 6.3.9-1-default (x86_64)
ncurses: ncurses 6.4.20230617 (compiled with 6.4.20230520)
libidn2: 2.3.4 (compiled with 2.3.4)
GPGME: 1.20.0
GnuTLS: 3.8.0
libnotmuch: 5.6.0
PCRE2: 10.42 2022-12-11
storage: kyotocabinet, lmdb
compression: zlib

Configure options: --prefix=/usr --docdir=/usr/share/doc/packages/neomutt --with-mailpath=/var/mail --kyotocabinet --lua --lmdb --gnutls --gpgme --notmuch --sasl --gss --idn2 --mixmaster --pcre2 --zlib

[...]

PR #3978, opened 2023-08-15
Add a check in `index_function_dispatcher`

Pierre-Colin:
* **What does this PR do?**

It makes `index_function_dispatcher` check whether `win` has a parent, fixing a segfault.

A trivial change in structure made the code shorter despite doing more checks.

#3924

### DIFF
--- a/index/functions.c
+++ b/index/functions.c
@@ -3020,14 +3020,11 @@ int index_function_dispatcher(struct MuttWindow *win, int op)
     return FR_ERROR;
   }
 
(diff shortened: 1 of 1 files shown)


roccoblues (approved):
Code change looks fine to me. I'd still leave it to @flatcap to ack that it's the-right-thing-to-do[tm].

flatcap (approved):
Thanks for jumping in on #3924!
I don't understand what could have gone wrong in that issue.

I don't _think_ this PR will stop the problem, but it pays to be cautious :-)
Thanks!

PR #3979, opened 2023-08-16
Objectify NeoMutt

nabijaczleweli:
* **What does this PR do?**
struct NeoMutt *NeoMutt; -> struct NeoMutt NeoMutt;

There's no reason to have four pointer chases when you can have three.

First three commits automatic, as described in the messages.

Hoping CI has an allyesconfig configuration and will catch anything my minimal ./configure has missed.

### DIFF
--- a/core/neomutt.c
+++ b/core/neomutt.c
@@ -56,10 +56,10 @@ void neomutt_init(struct NeoMutt *n, struct ConfigSet *cs)
 }
 
 /**
- * neomutt_free - Free a NeoMutt
- * @param[out] ptr NeoMutt to free
+ * neomutt_clear - Free a NeoMutt's contents
+ * @param[out] ptr NeoMutt to clear
  */
-void neomutt_free(struct NeoMutt *n)
+void neomutt_clear(struct NeoMutt *n)
 {
   if (!n || !n->initialised)
     return;
--- a/core/neomutt.c
+++ b/core/neomutt.c
@@ -34,45 +34,41 @@
 #include "account.h"
 #include "mailbox.h"
 
-struct NeoMutt *NeoMutt = NULL; ///< Global NeoMutt object
+struct NeoMutt NeoMutt; ///< Global NeoMutt object
 
 /**
  * neomutt_new - Create the main NeoMutt object
  * @param cs Config Set
  * @retval ptr New NeoMutt
  */
-struct NeoMutt *neomutt_new(struct ConfigSet *cs)
+void neomutt_init(struct NeoMutt *n, struct ConfigSet *cs)
 {
+  memset(n, 0, sizeof(*n));
   if (!cs)
-    return NULL;
-
-  struct NeoMutt *n = mutt_mem_calloc(1, sizeof(*NeoMutt));
+    return;
 
   TAILQ_INIT(&n->accounts);
   n->notify = notify_new();
   n->sub = cs_subset_new(NULL, NULL, n->notify);
   n->sub->cs = cs;
   n->sub->scope = SET_SCOPE_NEOMUTT;
-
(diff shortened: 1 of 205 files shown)


flatcap:
> Hoping CI has an allyesconfig configuration

Yes
"Build and Test" does default, min and max builds.
 "ASAN" does a max build and test under the [Address Sanitizer](https://neomutt.org/dev/analysis/asan)

flatcap:
Files changed 201
```diff
 diff
+1,279
-1,289
```

Hmm... this is going to make rebasing the other PRs interesting.

flatcap:
Yeah, given that it's all `sed`-able, it shouldn't be a problem.

A technique I often use, is to turn the PR branch into patches (`git format-patch`),
then manipulate the patches (vim-fu, sed, etc), then reapply them.

flatcap:
I've just noticed your branch name!  `NeoMutt->NeoMutt->`
I'm amazed all the tooling between us works!

flatcap:
Time to tick the box!

gahr:
You are now using `NeoMutt.sub` as an indication that `NeoMutt` was init'd. In some places, you use `NeoMutt.notify` for that, because the code uses that member. But the code happens to also use `NeoMutt.sub`. I think it's a bit implicit. How about an explicit `bool initialized` member?

nabijaczleweli:
yeah, that's definitely sensible

core/neomutt.c
```diff
@@ -34,25 +34,24 @@
...
-struct NeoMutt *NeoMutt = NULL; ///< Global NeoMutt object
+struct NeoMutt NeoMutt; ///< Global NeoMutt object
 
 /**
  * neomutt_new - Create the main NeoMutt object
  * @param cs Config Set
  * @retval ptr New NeoMutt
  */
-struct NeoMutt *neomutt_new(struct ConfigSet *cs)
+struct NeoMutt neomutt_new(struct ConfigSet *cs)
```
gahr, line 44: s/neomutt_new/neomutt_init/ and take a `struct Neomutt*` as argument

nabijaczleweli: applied

nabijaczleweli:
aaand CI passes again

ossilator:
you can avoid that enormous churn by declaring the global as
`struct NeoMutt NeoMutt[1];`. that would also keep the code more uniform, and avoid repeated churn if somebody changed it again (e.g., to lib-ify it).

nabijaczleweli:
which also hides `if(NeoMutt)` being broken

flatcap:
@ossilator
That's cute, but the idea was to save time on all the dereferences.

I'm not too worried about churn.
The current users of `NeoMutt` are:

 - 10 `NeoMutt.accounts`
 - 56 `NeoMutt.notify`
 - 1017 `NeoMutt.sub`
 
The config system supports Account- and Mailbox-specific variables.
When we implement these, all the `NeoMutt.sub` will disappear.

flatcap:
> the generated code is the same as with nab's patch

Ah!  Fair enough

flatcap:
Thanks for all the feedback.
I'm happy to consider enormous code changes, but in this case I'm going to say, "**No, thanks**".

- nabijaczleweli's changes reduced the visual clutter and saved one asm instruction each
- ossilator's suggestion cleverly won the same asm saving, without the churn

94% of the `NeoMutt` dereferences are for config lookups.
I'd rather focus on eliminating them.
Doing so would actually restore the pointer derefence, albeit on a different object,
e.g. `account->sub` or `mailbox->sub`.

flatcap:
Archived to https://github.com/neomutt/neomutt-old/tree/neomutt-object

Issue #3980, opened 2023-08-16
failed redraw (aka half-drawn) after op_shell_escape


lucilanga:
Neomutt fails to properly redraw main window when returning from a shell command. It only draws half of the screen.

This bug only occurs if certain cases are met.
Firstly the behavior  seems related to the famous half screen draw bug that has been around and debugged for quite a while.
Secondly it seems to occur only for tilling window managers.
It doesn't matter if it's under tmux/screen, I'm running neomut under tmux most of the time, but the debug was done in single terminal only. See below for alternate reproducer.

While my primary intention was to fix the behavior of shell escape alone, it is possible this fixes other half-drawn related issues like #1208.

How to reproduce:

Tilling window manager. Terminal doesn't matter (tried with foot, alacritty, xterm) the behavior is the same.
1st pass:
1. Open one terminal run neomutt - it should take all the desktop space.
2. Run another application say a pdf reader  - it should split the desktop space with neomutt terminal window.
3. close it - mutt terminal win will take all the desktop space. neomutt redraws correctly as it should be.

2nd pass:
Now:
1. open one term run neomutt
2 run again the same application taking half of the screen etc.
3. open a shell  (defkey "!")  -> we're thrown into a half screen shell
4. close the pdf reader
5. close the shell
6. mutt window half redrawn.

The main problem is SIGWINCH is not triggered when coming back after curses shutdown.
It does however work on the "1st pass", so I'm unsure if this is a curses bug.
I wasn't able to reproduce this in a floating window mode, I don't know that is can be either.
However coming across bug #1208 I was able to reproduce this in a tmux env regardless of WM.
The steps are similar: run neomutt inside tmux, make a new vertical or horizontal pane, run a shell, close pane, close shell -> same half drawn issue.

PS. I tested this on sway and i3 and lastly with tmux, I couldn't find any half draw issues except for #1208. I'm interested to see if there are other similar reports to this issue.

roccoblues:
Hey, I noticed that the corresponding PR is merged. Can this issue be closed?

Issue #1208, opened 2018-05-15
NeoMutt doesn't redraw as i3wm nor tmux pane


NWuensche:
## Expected Behaviour

Neomutt uses the size of the window that it gets. It should show everything inside the pane by resizing. When I open a new pane in tmux or i3, neomutt should react to whether it is displayed in fullscreen or only has half of the screen size

## Actual Behaviour

Neomutt only resizes on the y-axis but not on the x-axis. The Blue Lines at the top and bottom of the Neomutt window are always too long and I can't read everything that stands on these lines. It is the same for email titles too.

## Steps to Reproduce

Used i3wm with and without tmux, and NeoMutt 20171215 on Ubuntu 18.04 and Ubuntu 17.10.

## How often does this happen?

- Always

## When did it start to happen?

Since I use Neomutt with Ubuntu, 1-2 Months

## NeoMutt Version

```
NeoMutt 20171215
Copyright (C) 1996-2016 Michael R. Elkins und andere.
NeoMutt übernimmt KEINERLEI GEWÄHRLEISTUNG. Starten Sie `neomutt -vv', um
weitere Details darüber zu erfahren. NeoMutt ist freie Software. 
Sie können es unter bestimmten Bedingungen weitergeben; starten Sie
`neomutt -vv' für weitere Details.

System: Linux 4.15.0-20-generic (x86_64)
ncurses: ncurses 6.1.20180127 (compiled with 6.0)
libidn: 1.33 (compiled with 1.33)
hcache backends: tokyocabinet

Compiler:
Using built-in specs.
COLLECT_GCC=cc
COLLECT_LTO_WRAPPER=/usr/lib/gcc/x86_64-linux-gnu/7/lto-wrapper
OFFLOAD_TARGET_NAMES=nvptx-none
OFFLOAD_TARGET_DEFAULT=1
Target: x86_64-linux-gnu
[...]

PR #3981, opened 2023-08-16
force SigWinch after endwin()

lucilanga:
* **What does this PR do?**

I was about to add mutt_resize_screen() when returning from mutt_shell_escape() and call it a day.
But looking at bug #1208, flatcap suggested comparing mutt and neomutt behavior. And to my own shock mutt was
working fine for issues described on #3980.

Debugging mutt I could reproduce the same behavior: SIGWINCH does not trigger when coming back from curses shutdown (eg. mutt_endwin()), but it was working there so I narrowed down to commit 619db54f5133cc788513737ca354f91ab54130fc in mutt tree.
Kevin writes the following:
"Since mutt_reflow_window() needs to be called on a resize, and it's
    possible for programs to block SIGWINCH being sent to Mutt, this is a
    fail-safe to ensure it's run."

I'm not entirely sure about programs blocking SIGWINCH signal, I've tried with several terminals or different WMs but it's obviously a known thing and probably a better fix than mine.

#3980

### DIFF
--- a/gui/curs_lib.c
+++ b/gui/curs_lib.c
@@ -366,6 +366,7 @@ void mutt_endwin(void)
    * doesn't properly flush the screen without an explicit call.  */
   mutt_refresh();
   endwin();
+  SigWinch = true;
 
   errno = e;
 }


ossilator:
the "blocking" of SIGWINCH is a complete misunderstanding of what is happening. that signal is simply delivered only to processes in the terminal's current process group. unless mutt spawns the shell incorrectly, that will be a different process group than mutt itself. also, the shell will launch its jobs in separate process groups in turn.

so the patch is eminently reasonable. a more verbose commit message would be in order, though.

flatcap (approved):

Issue #3987, opened 2023-08-18
SIGSEGV upon searching in an e-mail


Werkov:
## Expected Behaviour

A searched phrase is located and screen rolled to it.

## Actual Behaviour

```
'~/projects/neomutt/neomutt' terminated by signal SIGSEGV (Address boundary error)
```

If NeoMutt crashed, did your OS create a 'coredump' file? -- yes

## Steps to Reproduce

 * start neomutt
 * `C+/folder.name<Enter>` // change folder
 * `<Enter>` // open first message
 * `<Ctrl-K>` // [toggle wordwrap off](https://github.com/Werkov/dotfiles/blob/40a60739c8b0fc9e0012b0711bb5cb4ff45b8020/.mutt/inc/base#L35)
 * `/word<Enter>` // search a 'word', it's in the message
 * crash

## How often does this happen?

- Always

- Sometimes

- On a particular email
  - I can't attach it
  - mail contains the word and long lines that are subject to (un)wrapping

## When did it start to happen?

- at least since neomutt-20230517-1.3.x86_64 on openSUSE TW
- happens also with upstream 186c63f8f

## NeoMutt Version

```
NeoMutt 20230517-300-186c63
Copyright (C) 1996-2022 Michael R. Elkins and others.
NeoMutt comes with ABSOLUTELY NO WARRANTY; for details type 'neomutt -vv'.
NeoMutt is free software, and you are welcome to redistribute it
under certain conditions; type 'neomutt -vv' for details.

System: Linux 6.4.9-1-default (x86_64)
ncurses: ncurses 6.4.20230715 (compiled with 6.4.20230715)
libidn2: 2.3.4 (compiled with 2.3.4)

Configure options: --disable-doc

Compilation CFLAGS: -fno-delete-null-pointer-checks -D_ALL_SOURCE=1 -D_GNU_SOURCE=1 -D__EXTENSIONS__ -D_XOPEN_SOURCE_EXTENDED -DNCURSES_WIDECHAR -O2

Default options:
  +attach_headers_color +compose_to_sender +compress +cond_date +debug 
  +encrypt_to_self +forgotten_attachments +forwref +ifdef +imap +index_color 
  +initials +limit_current_thread +multiple_fcc +nested_if +new_mail +nntp +pop 
  +progress +quasi_delete +regcomp +reply_with_xorig +sensible_browser +sidebar 
  +skip_quoted +smtp +status_color +timeout +tls_sni +trash 

Compile options:
[...]

PR #3988, opened 2023-08-18
pager: fix search crash

flatcap:
Don't search empty/missing arrays.

Fixes: #3987

### DIFF
--- a/pager/display.c
+++ b/pager/display.c
@@ -178,7 +178,7 @@ static void resolve_color(struct MuttWindow *win, struct Line *lines, int line_n
   }
 
   color = def_color;
-  if ((flags & MUTT_SHOWCOLOR) && lines[m].syntax_arr_size)
+  if ((flags & MUTT_SHOWCOLOR) && lines[m].syntax)
   {
     matching_chunk = bsearch(&cnt, lines[m].syntax, lines[m].syntax_arr_size,
                              sizeof(struct TextSyntax), comp_syntax_t);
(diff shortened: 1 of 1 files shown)


pager/display.c
```diff
@@ -178,7 +178,7 @@ static void resolve_color(struct MuttWindow *win, struct Line *lines, int line_n
   }
 
   color = def_color;
-  if ((flags & MUTT_SHOWCOLOR) && lines[m].syntax_arr_size)
+  if ((flags & MUTT_SHOWCOLOR) && (lines[m].syntax_arr_size > 0))
```
roccoblues, line 181: I read into `bsearch()` a bit because I was suprised that it would fail with an empty array. And if I understand [the standard](https://pubs.opengroup.org/onlinepubs/9699919799/functions/bsearch.html) correctly it shouldn't:
> If the nel argument has the value zero, the comparison function pointed to by compar shall not be called and no match shall be found.

So the problem isn't an empty array but `lines[m].syntax` being `NULL`?

Long story short: while I understand how this prevents the crash wouldn't this be a more "correct" fix?

```suggestion
  if ((flags & MUTT_SHOWCOLOR) && lines[m].syntax)
```

flatcap: Yeah.  Updated.

I don't know why, but the number of elements is initialised to `-1`.
This makes all the `if (num)` checks a bit iffy.

The code needs to die, but refactoring the pager is a huge block of work

PR #3989, opened 2023-08-18
move config/cache to core library

roccoblues:
See this discussion: https://github.com/neomutt/neomutt/discussions/3968


flatcap (approved):
Good attention to detail :-)
Thanks!

PR #3991, opened 2023-08-21
add missing include xlocale.h for macOS

roccoblues:
On macOS we need to include xlocale.h to define 'locale_t'.


gahr:
Mind to remove it from mx.c?

roccoblues:
> Mind to remove it from mx.c?

sure, done.

PR #3992, opened 2023-08-23
move config address type to address library

roccoblues:
see this discussion: https://github.com/neomutt/neomutt/discussions/3969

I removed `address_free()` since it was just a wrapper around `mutt_addr_free()`.


roccoblues (commented):
I'm not sure about the filenames `address/config.c` and `test/address/config.c`. Might be a misleading - open for suggestions.

flatcap (approved):
Perfect!  Thanks!

PR #3995, opened 2023-08-25
move mutt_history.c to history lib

roccoblues:
This moves the two remaining functions from `mutt_history.c` to the `history` lib:
* `mutt_hist_complete()`
* `main_hist_observer()`


flatcap:
Ah!  Good idea.

I had to rewind 5 years through the commit history to understand why I didn't do this before.
When I first started creating libraries, I often ended up with `X/*.c` (the bits with few deps) and `/mutt_X.c` (the ugly bits).
At the time of libhistory, neither libenter nor libcomplete existed.  Now they do :-)

roccoblues:
Yeah, I thought that it would be something like this. :-) Now the dependencies look fine.

flatcap (approved):
lovely, thanks!

PR #4017, opened 2023-08-29
remove search globals

roccoblues:
**Goals**

* get rid of `OptSearchReverse` from `globals.c`
* don't use op codes (`OP_SEARCH_BACKWARDS`, ...) in `mutt_search_command()` and `mutt_search_alias_command()`
* remove `SearchPattern`, `LastSearch` and `LastSearchExpn` globals from `pattern/pattern.c`

**Idea**

I split the required search data into two parts:
1. State that needs to be kept across operations. This is what the globals `OptSearchReverse` and `SearchPattern`, `LastSearch`, `LastSearchExpn` currrently hold.
2. "one-time" search flags:
  a. `OP_SEARCH` and `OP_SEARCH_BACKWARDS` trigger the search functions to prompt the user for input.
  b. `OP_SEARCH_OPPOSITE` tells the search function to search in the opposite direction of what is currently the global state. 

For the state data I've added a new struct `SearchState` that holds the data previously in the globals. That struct itself is kept on `IndexSharedData`, `AliasMenuData` and `PostponeData`.

For the flags I added `SearchFlags` (`SEARCH_PROMPT`, `SEARCH_OPPOSITE`) that get computed in `op_search()` based on the operation.

-> Related discussion: https://github.com/neomutt/neomutt/discussions/3997

### DIFF
--- a/pattern/pattern.c
+++ b/pattern/pattern.c
@@ -434,16 +434,16 @@ bail:
 
 /**
  * mutt_search_command - Perform a search
- * @param mv     Mailbox view to search through
- * @param menu   Current Menu
- * @param search Current search state
- * @param cur    Index number of current email
- * @param flags  Search flags, e.g. SEARCH_PROMPT
- * @retval >=0   Index of matching email
- * @retval -1    No match, or error
+ * @param mv    Mailbox view to search through
+ * @param menu  Current Menu
+ * @param state Current search state
+ * @param cur   Index number of current email
+ * @param flags Search flags, e.g. SEARCH_PROMPT
+ * @retval >=0  Index of matching email
+ * @retval -1   No match, or error
  */
 int mutt_search_command(struct MailboxView *mv, struct Menu *menu,
-                        struct SearchState *search, int cur, SearchFlags flags)
+                        struct SearchState *state, int cur, SearchFlags flags)
 {
   struct Progress *progress = NULL;
   int rc = -1;
@@ -451,12 +451,12 @@ int mutt_search_command(struct MailboxView *mv, struct Menu *menu,
   if (!m)
     return -1;
 
-  if (buf_is_empty(search->string) || (flags & SEARCH_PROMPT))
+  if (buf_is_empty(state->string) || (flags & SEARCH_PROMPT))
   {
-    if ((mw_get_field((search->reverse) ? _("Reverse search for: ") : _("Search for: "),
-                      search->string, MUTT_COMP_CLEAR | MUTT_COMP_PATTERN,
-                      false, NULL, NULL, NULL) != 0) ||
-        buf_is_empty(search->string))
+    if ((mw_get_field((state->reverse) ? _("Reverse search for: ") : _("Search for: "),
+                      state->string, MUTT_COMP_CLEAR | MUTT_COMP_PATTERN, false,
+                      NULL, NULL, NULL) != 0) ||
+        buf_is_empty(state->string))
     {
       goto done;
     }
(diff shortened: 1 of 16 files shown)


flatcap:
That looks good!

> remove `SearchPattern`, `LastSearch` and `LastSearchExpn` globals from `pattern/pattern.c`

I was worried that this might prevent the user continuing a search in a different dialog.
I thought that you could:
- Search in Index `/john`
- Open Alias dialog
- Hit `<search-next>` to find `john`

But apparently that's not the case, so we needn't worry :-)

> The search state manipulation and flags computation clutter all `op_search()` functions.

I don't see that as a problem.
Each `op_search()` is self-contained and bridges the gap between GUI and library.

> Where/when to initialize the Search? On demand like here or once when IndexSharedData is created.

Hmm...
The Alias and Postponed dialogs are fairly short-lived.
The Index is very long-lived.

Given that the search object is near-trivial in size, I'd say create it once on startup of each Dialog.

> Extract search state/flags manipulation to separate method(s)?.

Is there _that_ much that could be extracted?

> Naming all the things ... struct Search, mutt_search_new(), mutt_search_free(), ...

I think I'd go for `struct SearchState`, `search_state_new()`, `_free()` and vars called `state` (in a new file).

flatcap:
One larger problem for us to think about, is whether the Index and Alias search code should be in `libpattern`

Should `mutt_search_command()` live in `libindex` and rely on `libpattern`? (probably in a dedicated file)
Likewise `mutt_search_alias_command()` in `libalias`.

Or would it be possible to abstract the pattern-searchiness to work on arbitrary objects/data?
(this is probably a **lot** of work)

roccoblues:
> Should `mutt_search_command()` live in `libindex` and rely on `libpattern`? (probably in a dedicated file)
> Likewise `mutt_search_alias_command()` in `libalias`.

Interesting. `mutt_search_command()`is also used in `libpostpone` are we fine with that?
On that note: does `SearchState` really belong in `libpattern`? Or should we move everything together into a `libsearch`?

pattern/pattern.c
```diff
@@ -434,65 +430,58 @@ int mutt_pattern_func(struct MailboxView *mv, int op, char *prompt)
...
       buf_init(&err);
       OptSearchInvalid = true;
-      mutt_str_copy(LastSearch, buf_string(buf), sizeof(LastSearch));
-      mutt_str_copy(LastSearchExpn, buf_string(tmp), sizeof(LastSearchExpn));
+      buf_copy(state->string_expn, tmp);
       mutt_message(_("Compiling search pattern..."));
-      mutt_pattern_free(&SearchPattern);
+      mutt_pattern_free(&state->pattern);
       err.dsize = 256;
       err.data = mutt_mem_malloc(err.dsize);
```
roccoblues, line 476: Unrelated to my changes but do you know why we manually allocate the buffer data here? Shouldn't the `buf_*` functions simply allocate/grow the buffer if necessary?

flatcap: In the early days, `buf_new()` didn't exist, nor did the Buffer Pool.
In some code where the Buffer might not be needed (and to save an alloc) it was allocated on the stack.

With modern processors and lots of memory, we shouldn't be worrying about `malloc()`ing a few KB.
An interesting question is: Is the Buffer Pool really a saving?

flatcap (changes requested):
Idea great.  Code great. Execution: **crash** :-)
`AliasMenuData.search_state` is `NULL` so `<search>` or `<search-next>` crashes.

roccoblues:
> Execution: **crash** :-) `AliasMenuData.search_state` is `NULL` so `<search>` or `<search-next>` crashes.

Wow, thanks for checking! For me it works. 😕 Maybe we're triggering different code paths. Here's what I do:
* hit 'm' to compose a new mail
* hit 'tab' to open alias window
* hit '/' to trigger a search

-> this works

I find 4 different methods where `AliasMenuData` is initialized: `query_complete()`, `query_index()`, `alias_complete()` and `alias_dialog()`.

pattern/pattern.c
roccoblues: > An interesting question is: Is the Buffer Pool really a saving?

For the `string`, `string_expn` in the `SearchState` or in general?

roccoblues: > With modern processors and lots of memory, we shouldn't be worrying about malloc()ing a few KB.

agreed, I've changed it

flatcap: > > Is the Buffer Pool really a saving?
> 
> For the `string`, `string_expn` in the `SearchState` or in general?

In general.
The _idea_ of a pool sounds good, but there aren't any _time-critical_ loops where we alloc/free Buffers.

> Shouldn't the `buf_*` functions simply allocate/grow the buffer if necessary?

Many of them don't: `if (!buf || !buf->data) FAIL;`
We just need a plan.

flatcap:
> Here's what I do:
> 
>     * hit 'm' to compose a new mail
> 
>     * hit 'tab' to open alias window
> 
>     * hit '/' to trigger a search

Yeah, that's exactly what I was doing, I thought.
It still crashes.
I'll run it through the debugger, PC (post-cornflakes :-)

pattern/pattern.c
roccoblues: > > > Is the Buffer Pool really a saving? 
> > 
> > For the `string`, `string_expn` in the `SearchState` or in general?
> 
> In general. The _idea_ of a pool sounds good, but there aren't any _time-critical_ loops where we alloc/free Buffers.

I can't really judge this. But I'm happy to rip it out if we decide to do it. ;-)

flatcap (approved):

PR #4019, opened 2023-09-01
remove OptSearchInvalid global

roccoblues:
Instead of setting `OptSearchInvalid = true` when an action (switch mailbox, new mail, sort, ...) "invalidates" the current search, we free the current search pattern. This triggers a recompile in the search methods.

This means the search methods behavior changes a bit: before we only "forgot" which mails we already searched and if they matched. Now we recompile the search pattern on invalidation.

I actually like this change because it makes the search methods `mutt_search_command()` and `mutt_search_alias_command()` a bit clearer. But if we decide that this is too expensive we could simply move the `invalid` bool to the `SearchState` and keep the current logic.


flatcap (approved):
very nice, thanks!

PR #4023, opened 2023-09-04
fix observer event types

roccoblues:
* `compose_email_observer` is only interested in `NT_ENVELOPE`
* `pager_index_observer` is only interested in `NT_INDEX`
* `sb_index_observer`is only interested in `NT_INDEX`


gahr:
Are there any behavioural changes, or is this an optimization?

roccoblues:
> Are there any behavioural changes, or is this an optimization?

No behaviroural changes. But I also don't think it'll have any measurable performance impact. 

Just something I noticed a long time ago when I was working with the observers. tbh, I forgot all about it but the new github startpage just presented it to me. ;-)

flatcap:
@roccoblues Nice spot!  Thanks

PR #4025, opened 2023-09-06
editor: move completion functions

flatcap:
**NOTE**: this _is_ working, but I need to do some more testing.

Deep within libenter there are lots of dependencies caused by auto-completion and history.

Users call `mw_get_field()` with arguments like `COMP_ALIAS` which determine which completion function to use and where to save the history.
Worse still is the fact that the alias code is in libenter.
and browser code and notmuch code and pattern code...

This PR creates a basic Completion API that passes in a function to do the work.
It should help with: #3955  Neomutt completion API

All the functions on the right are now in more appropriate places.
The abstraction isn't perfect, but it's a step forward.

Source: [graphviz](https://flatcap.org/mutt/keys/op_editor_complete.gv), [svg](https://flatcap.org/mutt/keys/op_editor_complete.svg)

### DIFF
--- a/complete/lib.h
+++ b/complete/lib.h
@@ -35,6 +35,7 @@
 #ifndef MUTT_COMPLETE_LIB_H
 #define MUTT_COMPLETE_LIB_H
 
+#include <stddef.h>
 #include <stdbool.h>
 // IWYU pragma: begin_keep
 #include "data.h"
@@ -48,5 +49,7 @@ int  mutt_label_complete    (struct CompletionData *cd, struct Buffer *buf, int 
 bool mutt_nm_query_complete (struct CompletionData *cd, struct Buffer *buf, int pos, int numtabs);
 bool mutt_nm_tag_complete   (struct CompletionData *cd, struct Buffer *buf, int numtabs);
 int  mutt_var_value_complete(struct CompletionData *cd, struct Buffer *buf, int pos);
+void matches_ensure_morespace(struct CompletionData *cd, int new_size);
+bool candidate               (struct CompletionData *cd, char *user, const char *src, char *dest, size_t dlen);
 
 #endif /* MUTT_COMPLETE_LIB_H */
--- a/complete/lib.h
+++ b/complete/lib.h
@@ -38,10 +38,14 @@
 #include <stddef.h>
 #include <stdbool.h>
 // IWYU pragma: begin_keep
+#include "compapi.h"
 #include "data.h"
 // IWYU pragma: end_keep
 
 struct Buffer;
+
+extern const struct CompleteOps CompleteCommandOps;
+extern const struct CompleteOps CompleteLabelOps;
 
 int  mutt_command_complete  (struct CompletionData *cd, struct Buffer *buf, int pos, int numtabs);
 int  mutt_complete          (struct CompletionData *cd, struct Buffer *buf);
--- a/Makefile.autosetup
+++ b/Makefile.autosetup
@@ -106,9 +106,9 @@ $(PWD)/address:
 ###############################################################################
 # libalias
 LIBALIAS=	libalias.a
-LIBALIASOBJS=	alias/alias.o alias/array.o alias/commands.o alias/config.o \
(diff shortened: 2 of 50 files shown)


Amudtogal:
Hey flatcap,
this is amazing, thank you for the great refactor.
I will rebase my autocompletion branch on this, and this should speed things up.

complete/lib.h
```diff
@@ -35,18 +35,25 @@
...
+#include <stddef.h>
 #include <stdbool.h>
 // IWYU pragma: begin_keep
+#include "compapi.h"
 #include "data.h"
 // IWYU pragma: end_keep
 
 struct Buffer;
 
+extern const struct CompleteOps CompleteCommandOps;
```
Amudtogal, line 47: why are these needed here?
if we use the `#include "compapi.h", are they not defined already (or is this an alias)?

flatcap: `compapi.h` defines the _interface_,
`CompleteCommandOps`, `CompleteLabelOps`, define _implementations_ which are used outside of the library.

We could do with better homes for these.

Amudtogal (approved):
Excellent, this is now so much clearer (at least in my head).

It's great that each user (alias, browser etc) manages their own completion function, without needing too much logic.

Amudtogal (commented):
Excellent, this is now so much clearer (at least in my head).

It's great that each user (alias, browser etc) manages their own completion function, without needing too much logic.

Amudtogal:
just one thing I spotted when compiling:
```
browser/complete.c:122:20: warning: implicit declaration of function ‘mutt_mb_is_shell_char’; did you mean ‘mutt_mb_is_pathname_char’? [-Wimplicit-function-declaration]
  122 |        (i > 0) && !mutt_mb_is_shell_char(wdata->state->wbuf[i - 1]); i--)
      |                    ^~~~~~~~~~~~~~~~~~~~~
      |                    mutt_mb_is_pathname_char

```

how would i best submit tiny patches like this?
is it best in a pull request, or assemble them along with the completion api branch?

complete/lib.h
Amudtogal: I see. I wasn't aware that C implicitly declares functions as `extern`, and thus got confused by the mix in interface & implementation.

flatcap:
> browser/complete.c:122:20: warning: implicit declaration of function ‘mutt_mb_is_shell_char’; did you mean ‘mutt_mb_is_pathname_char’? [-Wimplicit-function-declaration]

Hmm... I don't understand.

https://github.com/neomutt/neomutt/blob/b99bce75547070bfa43e491e80744ca1b409ccdb/browser/complete.c#L121-L124

should be fine, because the file includes `mutt/lib.h`

https://github.com/neomutt/neomutt/blob/b99bce75547070bfa43e491e80744ca1b409ccdb/browser/complete.c#L33

which includes `mbyte.h`

https://github.com/neomutt/neomutt/blob/b99bce75547070bfa43e491e80744ca1b409ccdb/mutt/lib.h#L87

which declares `mutt_mb_is_shell_char()`

https://github.com/neomutt/neomutt/blob/b99bce75547070bfa43e491e80744ca1b409ccdb/mutt/mbyte.h#L49

I can't repeat the error myself with gcc or clang.

> is it best in a pull request, or assemble them along with the completion api branch?

if you can figure out what's happened, push a commit to this branch.

> how would i best submit tiny patches like this?

Generally, feel free to create/add to a branch `[trivial]` in the neomutt repo.
I'll merge it into my next bundle of tidying.

Amudtogal:
Fair enough, it is gone on my side as well, after `make clean && make`.
Maybe it was a lingering object file...
If it comes back up, I will contribute to `[trivial]`

Amudtogal:
Ah found it:
further down in my dev branch, I am renaming `mutt_mb_is_shell_char` to `mutt_mb_is_pathname_char` (more verbose, and it was only used in one spot anyway).
This was causing the compilation problem.

I have rebased my dev branch on this pull request to get started.

PR #4031, opened 2023-09-10
email/rfc2047: avoid reading past `len` in QP decode.

vuori:
* **What does this PR do?**

When `decode_word` handles quoted-printable encoded text, it will look past the current point when it encounters a `=` to check whether it looks like a hex-encoded character. If passed a malformed QP string with a `=` at the end, this may read past the end of the buffer as given by `len`.

In practice this shouldn't make a difference, because AFAICT the buffer passed from `rfc2047_decode` will have `?=` at the end. But to avoid surprises, do a bounds check anyway.

No issue opened.

### DIFF
--- a/email/rfc2047.c
+++ b/email/rfc2047.c
@@ -358,12 +358,15 @@ static void finalize_chunk(struct Buffer *res, struct Buffer *buf, char *charset
  * @param enc Encoding type
  * @retval ptr Decoded string
  *
- * @note The caller must free the returned string
+ * @note The input string must be null-terminated; the len parameter is
+ *       an optimization. The caller must free the returned string.
  */
 static char *decode_word(const char *s, size_t len, enum ContentEncoding enc)
 {
   const char *it = s;
   const char *end = s + len;
+
+  assert(*end == '\0');
 
   if (enc == ENC_QUOTED_PRINTABLE)
   {


gahr:
> it will look past the current point when it encounters a =

Yes, but not the way I think you mean it :) it will look into `it[1]`, which is a fine thing to do. If `it+1 == end`, then `it[1]` is `\0` -> `hexval(it[1]) != -1)` is `0` -> the conditional is short-circuit and it doesn't look into `it[2]`, which I think is what you're worried about.

vuori:
Ah, I see the meaning of `text[textlen] = '\0';` in the caller now. Though I would still argue that a function that takes array+length AND still opaquely requires that the array be null-terminated at `len` is terrible even by the terrible standards of C string handling.

gahr:
I don't necessarily disagree, but i) `decode_word` is an internal API (it's static within `rfc2047.c`), ii) we need the string to be null-terminated anyway to be able to pass it to mutt_b64_decode, iii) it's nice to pass the length since we have it, because we'd need to compute it anyway to figure the size of the base64-decoded string.
So we're left with:
1. we don't null-terminate it, we add the checks you suggest, and we change `mutt_b64_decode` to take pointer+length
2. we don't pass the length, we compute it internally, but it's throwing away information we already have at caller site
3. we keep null-terminating it *and* passing the length, as inelegant as it might look 

I think in this particular case I'm fine with 3.

vuori:
How about updating the docstring + an assert that `s[end] == '\0'` to prevent footguns in the future then? If someone really cares about micro-optimizations here, they can build with `NDEBUG`.

While new production usages are unlikely to appear soon, there was talk on irc about making some of these statics not-static for testing/fuzzing.

flatcap (approved):
Thanks!

PR #4034, opened 2023-09-11
complete: relax complete/query checks

flatcap:
Only the Alias completion needs to distinguish between `<complete-query>` and `<complete>`.
In other places, treat `<complete-query>` as `<complete>`.

This restores the behaviour that we had before the completion refactor.

Fixes: #4032

### DIFF
--- a/alias/complete.c
+++ b/alias/complete.c
@@ -36,7 +36,7 @@
 #include "opcodes.h"
 
 /**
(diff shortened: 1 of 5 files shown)


gahr:
my fault? sorry :)

flatcap:
> my fault?

No, my fault.
I didn't convert your fix that allowed `<complete-query>` to fallback to `<complete>`

Amudtogal (approved):
Looks good on my side.

PR #4035, opened 2023-09-12
remove global EnvList from mutt/filter.c

roccoblues:
Change `filter_create_fd()` and `filter_create()` to accept the environment variables as parameter. This removes `globals.h` from `mutt/filter.c`. 

Discussion: https://github.com/neomutt/neomutt/discussions/3993


flatcap (approved):
Lovely, `libmutt` is dep-free!
Thanks!

PR #4036, opened 2023-09-13
eliminate global OptAttachMsg

roccoblues:
* move `attach_msg` flag from `IndexPrivateData` to `IndexSharedData`
* add `attach_msg` flag to `AttachPrivateData`
* replace uses of `OptAttachMsg` with this flags
* remove `reply_validator()` and drop global `OptAttachMsg`

Triggered by this discussion: https://github.com/neomutt/neomutt/discussions/3996

### DIFF
--- a/attach/dlg_attach.c
+++ b/attach/dlg_attach.c
@@ -459,17 +459,18 @@ static int attach_window_observer(struct NotifyCallback *nc)
 
 /**
  * dlg_attachment - Show the attachments in a Menu - @ingroup gui_dlg
- * @param sub Config Subset
- * @param mv  Mailbox view
- * @param e   Email
- * @param fp File with the content of the email, or NULL
+ * @param sub        Config Subset
+ * @param mv         Mailbox view
+ * @param e          Email
+ * @param fp         File with the content of the email, or NULL
+ * @param attach_msg Are we in "attach message" mode?
  *
  * The Select Attachment dialog shows an Email's attachments.
  * They can be viewed using the Pager or Mailcap programs.
  * They can also be saved, printed, deleted, etc.
  */
 void dlg_attachment(struct ConfigSubset *sub, struct MailboxView *mv,
-                    struct Email *e, FILE *fp)
+                    struct Email *e, FILE *fp, bool attach_msg)
 {
   if (!mv || !mv->mailbox || !e || !fp)
     return;
@@ -495,6 +496,7 @@ void dlg_attachment(struct ConfigSubset *sub, struct MailboxView *mv,
   priv->actx = actx;
   priv->sub = sub;
   priv->mailbox = m;
+  priv->attach_msg = attach_msg;
   menu->mdata = priv;
   menu->mdata_free = attach_private_data_free;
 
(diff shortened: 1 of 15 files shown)


roccoblues:
The only place left is [reply_validator](https://github.com/neomutt/neomutt/blob/main/mutt_config.c#L119).

flatcap:
> you introduced `reply_validator()` ...

The _validator_, yes, but not the check.
The check goes back to b4aaf6835 Add caching for a message's flags. (25 years ago)

roccoblues:
> > you introduced `reply_validator()` ...
> 
> The _validator_, yes, but not the check. The check goes back to [b4aaf68](https://github.com/neomutt/neomutt/commit/b4aaf6835e78da352a163801a2cb287a7b42222e) Add caching for a message's flags. (25 years ago)

oh yeah, you're right ... wow

roccoblues:
I think I found something: https://github.com/neomutt/neomutt/blob/main/index/index.c#L194. `config_reply_regex()` can change the envelope of the emails and triggers a redraw of the index. Maybe this doesn't play nicely with the nested indexes in attache message mode?

flatcap:
Conversation from IRC: `#neomutt` on `irc.libera.chat` ([web client](https://kiwiirc.com/nextclient/irc.libera.chat/#neomutt))

roccoblues | I found `config_reply_regex()` that might explain the check
flatcap | yes, that's our addition too
flatcap | that might be the sign that we're good to drop the restriction
flatcap | a long time ago, in a MUA not that far away...
flatcap | In Mutt there's a global Context which is the **current** Mailbox (other mailboxes may exist)
flatcap | it holds the top of the thread tree for the emails
flatcap | (in neomutt we have a MailboxView for each Mailbox)
roccoblues | so attach msg mode *shouldn't* mess around with the other
flatcap | if `$reply_regex(p)` is changed we need to rethread all the emails
roccoblues | but in the good old mutt days it would have
flatcap | yes
roccoblues | makes sense
flatcap | in compose, when you try to attach an email, compose makes a new Context and backs up all the sort global config variables
flatcap | the *new* index has its own Context and thread tree, but messing around with the new index too much could invalidate the original thread tree
flatcap | in neomutt, when the user changes `$reply_regex,` we get a notification leading to `config_reply_regex()` being called
flatcap | there'll be one call for **each** index.  and each index will have its own Mailbox/MailboxView with its own thread tree
flatcap | so... they should be (untested) independent and safe
roccoblues | but ...
roccoblues | isn't changing **both** the same as in the old mutt days
roccoblues | otoh that is probably what you expect
roccoblues | or is the config variable not shared?
flatcap | _probably_ not.  in mutt, Index1 doesn't get notified by the events of Index2, things are just changed
flatcap | when index1 gets back control, it doesn't know (it can't tell) if anything's changed
flatcap | in neomutt, the notification system, means that index1 updates itself immediately
flatcap | in mutt the config is a global variable, in neomutt it's still global because we're only using one config subset. but...
flatcap | in the future, what I have envisioned is...
flatcap | Index1 has a MailboxView1 of Mailbox1, using the global config subset...
flatcap | compose creates MailboxView2 as a clone of MV1, and it creates a new config subset Sub2 on top of `NeoMutt->sub`
flatcap | Index2 is now identical to the one the user left earlier, but
flatcap | the view info (all the gui parts) are independent of index1
flatcap | AND any config changes are indepdenent of index1/`NeoMutt->sub`
flatcap | a lot of the pieces are in place to make this happen

flatcap (approved):

PR #4038, opened 2023-09-15
Add Help to Quad/Yes/No Questions

flatcap:
This PR adds an upstream feature and extends it.

Many of the questions in NeoMutt are associated with config variables.
(All quad-options and many bools)

When the user is presented with a 

**Before**:
```
Append messages to work? ([yes]/no): 
```

**After**:  (note the '/?' )
```
Append messages to work? ([yes]/no/?): 
```

**Press <kbd>?</kbd>** to get:

```
$confirm_append - Confirm before appending emails to a mailbox
Append messages to work? ([yes]/no): 
```

**Additionally**, if `$help` is set (as beginners might):

```
$confirm_append - Confirm before appending emails to a mailbox
https://neomutt.org/guide/reference#confirm-append
Append messages to work? ([yes]/no): 
```

---

- dd32bddae query help y/n/?
  Add help to quad option and some y/n questions

- b4fd3a4a9 query help web link
  Display a web link if `$help` is set

---

For comparision, Mutt displays:

```
See $confirmappend for more information.
Append message(s) to work? ([yes]/no): 
```

### DIFF
--- a/index/functions.c
+++ b/index/functions.c
@@ -532,8 +532,7 @@ static int op_exit(struct IndexSharedData *shared, struct IndexPrivateData *priv
   if (priv->attach_msg)
     return FR_DONE;
 
-  const enum QuadOption c_quit = cs_subset_quad(shared->sub, "quit");
-  if (query_quadoption(c_quit, _("Exit NeoMutt without saving?")) == MUTT_YES)
+  if (query_quadoption(_("Exit NeoMutt without saving?"), shared->sub, "quit") == MUTT_YES)
   {
     if (shared->mailbox_view)
     {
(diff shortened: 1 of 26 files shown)


index/functions.c
```diff
@@ -357,9 +357,15 @@ static int op_delete_thread(struct IndexSharedData *shared,
 static int op_display_address(struct IndexSharedData *shared,
                               struct IndexPrivateData *priv, int op)
 {
-  if (!shared->email)
-    return FR_NO_ACTION;
-  mutt_display_address(shared->email->env);
+  // enum QuadOption ans = query_quadoption("How many roads must a man walk down, before they call him a man?", NeoMutt->sub, "quit");
```
roccoblues, line 360: What is going on here? 😆

roccoblues (commented):
> Additionally, if $help is set (as beginners might):

What do you think of showing this always when the users presses `?` to get more info?

index/functions.c
flatcap: hehe, oops!
Some of my debugging escaped.
I needed to display lots of long strings that I could wrap.

flatcap:
> What do you think of always showing this when the users presses `?` to get more info?

I'm OK with that.
I wasn't quite sure if it overstepped the balance between useful and space-wasting.

roccoblues:
> > What do you think of always showing this when the users presses `?` to get more info?
> 
> I'm OK with that. I wasn't quite sure if it overstepped the balance between useful and space-wasting.

My take is: if I press `?` to get help I want all the information available. And we're showing it only on demand - so the normal experience is still fine.

flatcap:
## Screenshots

Old style, New style, Basic help, Full help

flatcap:
Updated to always display the web link.

PR #4039, opened 2023-09-16
refactor cursor positioning

flatcap:
Painting the screen inevitably means moving the cursor.
After the painting is done, an interactive Window, e.g. `mw_multi_choice()` needs an opportunity to restore the cursor position.
This can be achieved by implementing `Window.recursor()`.

---

- 529c04bc2 window: add recursor()
  After all the repainting is done, the focussed window will be given an
  opportunity to set the position and visibility of the cursor.

- adb3c45f9 msgwin: implement recursor()
  - Drop WT_MSG_FOCUS Window type
  - Drop MsgWinWindowData.interactive
  - Move cursor setting to msgwin_recursor()

- ef419f630 upgrade simple MessageWindow users
  Let the MessageWindow do the cursor setting.

- 0bc3cce1b mw_get_field: implement recursor()
  - Drop EnterWindowData.col
    A local variable in enter_repaint() was sufficient
  - Add EnterWindowData.row, col to store the cursor position
  - Move the cursor setting to enter_recursor()

### DIFF
--- a/enter/wdata.h
+++ b/enter/wdata.h
@@ -46,7 +46,6 @@ struct EnterWindowData
 {
   // Function parameters
   struct Buffer *buffer;          ///< struct Buffer for the result
-  int col;                        ///< Initial cursor positions
   CompletionFlags flags;          ///< Flags, see #CompletionFlags
   struct EnterState *state;       ///< Current state of text entry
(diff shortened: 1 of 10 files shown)


roccoblues (commented):
Gave it a quick test-run and didn't notice anything fishy.

PR #4040, opened 2023-09-16
explicitly set the focus

flatcap:
Explicitly set and restore the Window focus.
Before, this used to happen in helper functions.

### DIFF
--- a/alias/dlg_alias.c
+++ b/alias/dlg_alias.c
@@ -337,6 +337,7 @@ static bool dlg_alias(struct Buffer *buf, struct AliasMenuData *mdata)
     avp->num = ARRAY_FOREACH_IDX;
   }
 
+  struct MuttWindow *old_focus = window_set_focus(menu->win);
   // ---------------------------------------------------------------------------
   // Event Loop
   int rc = 0;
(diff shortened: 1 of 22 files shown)


roccoblues:
Can you explain the reason for this change? To my untrained eye `dialog_push()` seems to be the right place to set the focus?

flatcap:
Partly, it's about visibility in the code.
The focus is never on the Dialog Window itself, but one of its children.
(or grandchildren in the case of the Index)

`dialog_push()` ensured that the chain of focus went: `RootWindow` -> `AllDialogs` -> `AliasDialog`
Someone still perform the final link: `AliasDialog` -> `MenuWindow`

Only the Dialog knows this final step, so **it** calls `window_set_focus(MenuWindow)`

flatcap:
In the Index, the **Dialog** creates a **Panel** and sets `Dialog.focus = Panel`.
The **Panel** creates a **Menu** and sets `Panel.focus = Menu`.
Then `dialog_push()` calls `window_set_focus(Dialog)`.

We want to send a notification for the **focussed** Window.
This means that `window_set_focus()` had to descend the focus chain to find the right Window.

roccoblues:
> Before this PR...
> 
> In the Index, the **Dialog** creates a **Panel** and sets `Dialog.focus = Panel`. The **Panel** creates a **Menu** and sets `Panel.focus = Menu`. Then `dialog_push()` calls `window_set_focus(Dialog)`.
> 
> We want to send a notification for the **focussed** Window. This means that `window_set_focus()` had to descend the focus chain to find the right Window.

Thanks! Now I understand why we want to make setting the focus explicit.

Issue #4041, opened 2023-09-19
Segmentation fault when recalling postponed message


shasta-bierly:
## Expected Behaviour

No crash

## Actual Behaviour

Segmentation fault

## Steps to Reproduce

I condensed my neomuttrc down to this:

```
set mbox_type   = Maildir
set folder = "../testinbox/"
set postponed = "../testdrafts/"
set spool_file = "../testinbox/"

color status bold green blue 'foo'
color index        red   default '~(~F) ~v'

set use_threads = threads
```

Make sure `testdrafts/` has two postponed messages, e.g. create them.  With one alone it did not crash.
Not sure if relevant, but the inbox was empty and all mailboxes have maildir format.

1. `./neomutt -F ../neomuttrc`
2. You should see an empty inbox
3. Press `R` to recall message
4. Select the second (bottom) message  (first one sometimes also crashed)
5. Hit `Enter`
6. Crash

Note: Sometimes (while condensing down the neomuttrc) I had to recall/postpone the same message multiple times (2-3 times) until the crash happened.

## How often does this happen?

- Always

## When did it start to happen?

- When I upgraded

  Crashing versions: 20230517, and commit 568ff1cf11b31f495a92a3a3ef406428fc8951ff
  Not Crashing: 20220429

## Extra Info

Maybe related to https://github.com/neomutt/neomutt/issues/3790.

roccoblues:
Thanks for the report, so far I wasn't able to reproduce the crash. Can you build the latest `main` with [debugging symbols](https://neomutt.org/dev/build/debug#compile-with-debugging-information-) and see if you get a [core dump](https://neomutt.org/dev/build/debug#debugging-a-coredump-)?

shasta-bierly:
I build https://github.com/neomutt/neomutt/commit/568ff1cf11b31f495a92a3a3ef406428fc8951ff with debugging symbols and got the following backtrace:
<pre>
#0  0x000055555563dcb3 in match_threadcomplete
    (pat=0x5555558302b0, flags=1 '\001', m=0x55555583f6e0, t=0xda130712f29e1f8e, left=1, up=1, right=1, down=0) at pattern/exec.c:546
#1  0x000055555563dd23 in match_threadcomplete
    (pat=0x5555558302b0, flags=1 '\001', m=0x55555583f6e0, t=0x555555843530, left=1, up=1, right=1, down=1) at pattern/exec.c:551
#2  0x000055555563e816 in pattern_exec
[...]

PR #4044, opened 2023-09-22
disconnect emails from threads

flatcap:
When freeing the MailboxView, reset the Emails' Thread pointer before freeing the ThreadsContext.

Fixes: #4041


roccoblues:
Fixes the crash for me. 👍🏼

roccoblues (approved):

Issue #4055, opened 2023-09-30
notmuch query


JonJFineman:
## Expected Behaviour

Open a notmuch virtual mailbox

## Actual Behaviour
Illegal instruction (core dumped)

Didn't work.

If NeoMutt crashed, did your OS create a 'coredump' file? yes

## Steps to Reproduce
Open below mailbox:
```
virtual-mailboxes "UNREAD" "notmuch://?query=tag:unread \
  and not tag:deleted \
  and not tag:move:trash \
  and not folder:Sent \
  and not folder:Trash \
  and not folder:Feeds.yt_pending \
  and not folder:Feeds.ted_pending \
  and not folder:Feeds.z_rejected_pending \
  and not folder:Spam \
  and not folder:sales"
```

## How often does this happen?

- Always yes

- On a particular email
  If the email isn't private, please attach it to this issue.

Other virtual mailboxes work, for ex:
virtual-mailboxes "All-Mbox" "notmuch://?query=* (not folder:Trash and not folder:Sent and not tag:move:trash)"

## When did it start to happen?

- When I upgraded yes
  Which version did you use to use?
commit 8c90f5e2cde587ad4742be813737da4bd7f3957a

- When I changed my config
  Can you narrow down **what** you changed?

## NeoMutt Version

```
+ /home/jjf/bin/x_neomutt -v
NeoMutt 20230517-448-8c90f5-dirty
Copyright (C) 1996-2022 Michael R. Elkins and others.
NeoMutt comes with ABSOLUTELY NO WARRANTY; for details type 'neomutt -vv'.
NeoMutt is free software, and you are welcome to redistribute it
under certain conditions; type 'neomutt -vv' for details.

System: OpenBSD 7.4 (amd64)
ncurses: ncurses 5.7.20081102 (compiled with 5.7.20081102)
libiconv: 1.17
libidn2: 2.3.0 (compiled with 2.3.0)
GPGME: 1.22.0
LibreSSL: LibreSSL 3.8.2
OpenSSL: LibreSSL 3.8.2
libnotmuch: 5.6.0
storage: tdb

Configure options: --autocrypt --with-lock=flock --fmemopen --gpgme --notmuch --ssl --sasl --sqlite --tdb --debug --disable-doc

[...]

PR #4056, opened 2023-10-02
msgwin: check for dupe text

flatcap:
Return early from `msgwin_set_text()` if the text hasn't changed.

Fixes: #4055 

`msgwin_set_text()` is called to set or clear a message.

Displaying a long message would require the `MsgWin` to resize itself.
Later, calling `msgwin_set_text(NULL)` to clear the message would cause the `MsgWin` to shrink back to 1 line.

However, if the `MsgWin` wasn't visible, e.g. a progress bar was in use, then it would get into an infinite loop trying to resize itself.

Fix: If the text isn't changing, then don't try to resize yourself.

### DIFF
--- a/gui/msgwin.c
+++ b/gui/msgwin.c
@@ -490,6 +490,9 @@ void msgwin_set_text(struct MuttWindow *win, const char *text, enum ColorId colo
 
   struct MsgWinWindowData *wdata = win->wdata;
 
+  if (mutt_str_equal(buf_string(wdata->text), text))
+    return;
+
   buf_strcpy(wdata->text, text);
   ARRAY_FREE(&wdata->chars);
   if (wdata->text)


flatcap:
Test case (notmuch)

**test.rc**
```
set nm_default_url = "notmuch:///home/mutt/mail/linode"
set nm_query_type = "threads"

virtual-mailboxes "test" "notmuch://?query=wibble \
        and tag:Apple \
        and tag:Banana \
        and tag:Cherry \
        and tag:Damson \
        and tag:Elderberry \
        and tag:Fig \
        and tag:Guava \
        and tag:Hawthorn \
        and tag:Ilama \
        and tag:Jackfruit \
        and tag:Kumquat \
        and tag:Lemon \
        and tag:Mango \
        and tag:Nectarine \
        and tag:Olive \
        and tag:Papaya \
        and tag:Quince \
        and tag:Raspberry \
        and tag:Strawberry \
        and tag:Tangerine \
        and tag:Ugli \
        and tag:Vanilla \
        and tag:Wolfberry \
        and tag:Xigua \
        and tag:Yew \
        and tag:Ziziphus"

set spool_file = "test"
```

**Run**:
```
neomutt -n -F test.rc
```

Issue #4070, opened 2023-10-16
Message-ID doesn't have the hostname in it


nabijaczleweli:
Message-ID doesn't have the hostname in it, inline documentation says it should, out-of-line documentation doesn't say anything, broken since 20230407

## Expected Behaviour

/etc/mailname is in the message-ids. Like it was (according to my inbox) in NeoMutt/20220429 (`<20230322205351.zqytwcf2eqbtnhnj@tarta.nabijaczleweli.xyz>`) but is no longer since at least NeoMutt/20230407 (`<qlk7v62xob23ezhsu2ocqtjxzh3crahl7rxs6wboriyulwpzcx@b347xeja2pe5>`).

```
224   { "hidden_host", DT_BOOL, false, 0, NULL,
225     "Don't use the hostname, just the domain, when generating the message id"
226   },
```

```
       hidden_host
              Type: boolean
              Default: no

              When set, NeoMutt will skip the host name part of $hostname variable when adding the domain part to addresses.
```

How the message-id is generated is not mentioned in the manual, and, thus, clearly hasn't changed. It's also not documented in

> [neomutt-devel] NeoMutt 2023-04-07

or any of the later release mails, and, thus, again, mustn't've changed.

## Actual Behaviour

It clearly has, since since like april I've been sending broken mail. Very cool. Thanks.

## Steps to Reproduce

Send a mail.

## How often does this happen?

- Always

## When did it start to happen?

- When I upgraded
  Which version did you use to use? 20220429 and earlier

## NeoMutt Version

```
NeoMutt 20231006
Copyright (C) 1996-2022 Michael R. Elkins and others.
NeoMutt comes with ABSOLUTELY NO WARRANTY; for details type 'neomutt -vv'.
NeoMutt is free software, and you are welcome to redistribute it
under certain conditions; type 'neomutt -vv' for details.

System: Linux 6.1.0-9-amd64 (x86_64)
ncurses: ncurses 6.4.20221231 (compiled with 6.4.20221231)
libidn2: 2.3.3 (compiled with 2.3.3)
GPGME: 1.18.0
GnuTLS: 3.7.9
libnotmuch: 5.6.0
storage: lmdb
compression: zlib, zstd

[...]

PR #4071, opened 2023-10-16
Add $message_id_format=hostname|random

nabijaczleweli:
`random` does the new 20230407 format after the @, `hostname` uses get_fqdn()

The bit before the @ is unchanged

The `$message_id_format` name mimicks the name it has in mutt (cf. #3060); this can be extended, if desired, to handle its expando-based parser with just two compat streq()s. Or not. Personally I don't see the point.

message_id_format=hostname:
`<ewj2p7jxpvpfcekjogvu3xgwzufpj5mcygn3osyn3vueemuksn@tarta.nabijaczleweli.xyz>`

message_id_format=hostname, hidden_host:
`<c2skand7visl5eme2tj3pz2btanfnbz4tgaohocpvfgmk6synp@nabijaczleweli.xyz>`

message_id_format=random:
`<pyplqe6fbwxsf2eatqzsttkt5hj72odwzeor6veptdbpnzouj5@wkcl3frwzgix>`

#4070

### DIFF
--- a/docs/config.c
+++ b/docs/config.c
@@ -1640,6 +1640,15 @@
 ** when adding the domain part to addresses.
 */
 
+{ "message_id_format", DT_STRING, "hostname" },
+/*
+** .pp
+** If "hostname", $$hostname, subject to $$hidden_host,
+** is used after the \fB@\fP in generated Message-Ids.
+** .pp
+** If "random", a random 12-byte alphanumeric string is used.
+*/
+
 { "hidden_tags", DT_SLIST, "unread,draft,flagged,passed,replied,attachment,signed,encrypted" },
 /*
 ** .pp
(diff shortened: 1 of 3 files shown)


nabijaczleweli:
Ah, yes. Sure it's a valid mail, but breaking a behaviour unchanged since 1997 (probably earlier, but mutt 0.88 in debian bo) that people clearly depend on and then burying it, is not to me, an "actually you can use 
```
send-hook . "my_hdr Message-ID: <\`tr -cd '[:alnum:]' < /dev/urandom | head -c 50\`@$hostname>"`
```
so it's actually your fault you're gonna have to re-tag months of mail by hand :)" moment. Ah well. I'll just carry the patch, thanks.

PR #4077, opened 2023-10-19
[wip] new-mail notification observer

roccoblues:
_[This is just a proof-of-concept to see if the general connection works]_

Trigger the new-mail logic in a notification observer.

* add a hash to `IndexPrivateData` to keep track of the notification status per mailbox
* add a new mailbox notification type `NT_MAILBOX_NEW_MAIL`
* `index_mailbox_newmail_observer()` listens to `NT_MAILBOX` events and updates the hash
* `handle_new_mail()` is called in the index loop and triggers the new-mail logic

We need the new notification because `mx_mbox_check()` doesn't update `Mailbox->has_new` like `mx_mbox_check_stats()`. I still think we should only have one method that does the full check. It would make a lot of things simpler.

I intentionally made this minimally invasive, to just show the concept. I think this can be a first step to simplifying things. For example `mutt_mailbox_list()` and `mutt_mailbox_notify()` could be removed.

I didn't polish the message yet. So for multiple mailboxes where one is the current one you get something like `New mail =INBOX, =neomutt-devel`. But this can easily be changed to `New mail in this mailbox and =neomutt-devel`.

Also, is a hash the correct data structure? Is the mailbox path guaranteed to be unique?


roccoblues:
closing, stale

PR #4081, opened 2023-10-24
logging: minor fixes / improvements

flatcap:
- 1c69ae782 fix logging on error
  Some debugging messages were lost if there was an error in the config file

- 9c8613ba6 boolify: need_pause

- 49967ec49 log: vim-style
  Change the form of the error message to: `file:line_num: message`


roccoblues:
Looks good. 👍🏼

PR #4085, opened 2023-10-26
build: configure --with-iconv

flatcap:
Add a configure option, `--with-iconv` to override the location of libiconv.


auto.def
```diff
@@ -694,8 +696,13 @@ apply {{ncurses_prefix} {
 
 ###############################################################################
 # Iconv - try to mimic AM_ICONV by preferring an installed libiconv
-if {![check-iconv $prefix]} {
-  user-error "Unable to find iconv()"
+if {[opt-val with-iconv] ne {}} {
+  set iconv_include [opt-val with-iconv $prefix]
+  define-append CFLAGS -I$iconv_include/include
```
gahr, line 701: I think you also want to link from the same location

flatcap: I thought about that, but the @roccoblues's homebrew PR didn't include it
https://github.com/Homebrew/homebrew-core/pull/152445/files

roccoblues:
Doesn't seem to work for me:
```
./configure --disable-doc --disable-nls --testing --compile-commands --lmdb --zlib --debug --notmuch --with-notmuch=/opt/homebrew/Cellar/notmuch/0.38 --with-ncurses=/opt/homebrew/Cellar/ncurses/6.4 --with-iconv=
/opt/homebrew/Cellar/libiconv/1.17

[...]

Checking for iconv() in /opt/homebrew/Cellar/libiconv/1.17...no
Checking for iconv() in libiconv...no
Checking for iconv() in libc...no
Error: Unable to find iconv()
```

```
❯ ll /opt/homebrew/Cellar/libiconv/1.17
Permissions  Size User   Group Date Modified Name
drwxr-xr-x@     - dennis admin 26 Jun 15:39  .
drwxr-xr-x@     - dennis admin 26 Jun 15:39  ..
drwxr-xr-x@     - dennis admin 15 May  2022  .brew
drwxr-xr-x@     - dennis admin 26 Jun 15:39  bin
drwxr-xr-x@     - dennis admin 15 May  2022  include
drwxr-xr-x@     - dennis admin 26 Jun 15:39  lib
drwxr-xr-x@     - dennis admin 15 May  2022  share
```
```

roccoblues:
Ignore what I said above, seems to be something weird with my setup. The error was caused by `--with-ncurses=`. If I don't specify that it works.

roccoblues (approved):

gahr:
I agree. I want to link liba.so from /path/1 and libb.so from /path/2. With LDFLAGS and LIBS as separated values I end up with a link line containg -L/path/1 -L/path/2 -la -lb. Now, if libb.so also happens to be in /path/1, that’s the one that will be linked in.

ossilator:
i know - that wasn't the objection.

btw, cmake avoids this problem by simply resolving the libraries and passing absolute paths. this wouldn't have worked with old linkers that would embed incorrect SONAMEs, but it's generally supported nowadays.

gahr:
Turns out, I got totally sidetracked by thinking that it was picking up the library from the wrong place, while instead we just weren't keeping around the CFLAGS/LDFLAGS for ncurses, when it was found in a custom location.

roccoblues:
`--with-idn=/opt/homebrew/opt/libidn2 --with-iconv=/opt/homebrew/opt/libiconv --with-ncurses=/opt/homebrew/opt/ncurses` now works for me on macOS 14. 👍🏼

PR #4086, opened 2023-10-26
Simplify the API of smtp_code()

alejandro-colomar:
* **What does this PR do?**

There's only 1 call to this function, so we can simplify by passing the entire buffer, instead of extracting the string before the call.

$ grep -rn '\<smtp_code\>' *
send/smtp.c:440: * smtp_code - Extract an SMTP return code from a string send/smtp.c:445:static int smtp_code(const struct Buffer *buf, int *n)
send/smtp.c:484:    if (!smtp_code(input_buf, smtp_rc))

     for that)

N/A

I can't build from source successfully without the patch, so didn't try with the patch.

N/A

Except for a small thing: alignment of local variable declaractions.  It's a small suggestion I'm trying (a habit from nginx).

N/A

### DIFF
--- a/send/smtp.c
+++ b/send/smtp.c
@@ -438,22 +438,18 @@ static int smtp_helo(struct SmtpAccountData *adata, bool esmtp)
 #ifdef USE_SASL_GNU
 /**
  * smtp_code - Extract an SMTP return code from a string
- * @param[in]  str String to parse
- * @param[in]  len Length of string
+ * @param[in]  buf Buffer containing the string to parse
  * @param[out] n   SMTP return code result
  * @retval true Success
- *
- * Note: the 'len' parameter is actually the number of bytes, as
- * returned by mutt_socket_readln().  If all callers are converted to
- * mutt_socket_buffer_readln() we can pass in the actual len, or
- * perhaps the buffer itself.
  */
-static int smtp_code(const char *str, size_t len, int *n)
+static int smtp_code(const struct Buffer *buf, int *n)
 {
-  char code[4];
-
-  if (len < 4)
+  if (buf_len(buf) < 3)
     return false;
+
+  char code[4];
+  const char *str = buf_string(buf);
+
   code[0] = str[0];
   code[1] = str[1];
   code[2] = str[2];
@@ -484,7 +480,7 @@ static int smtp_get_auth_response(struct Connection *conn, struct Buffer *input_
   {
     if (mutt_socket_buffer_readln(input_buf, conn) < 0)
       return -1;
-    if (!smtp_code(buf_string(input_buf), buf_len(input_buf) + 1 /* number of bytes */, smtp_rc))
+    if (!smtp_code(input_buf, smtp_rc))
     {
       return -1;
     }


gahr (approved):

flatcap:
Lovely, thanks!

> alignment of local variable declaractions.

I _do_ like a bit of whitespace for alignment (see the header files),
but if we do it in the source files, it needs to be automatic.

We decided very early in NeoMutt's history that using an automatic formatter was the only way to avoid arguments :-)
We picked `clang-format` and a set of options that closely matched Mutt's source.

The `clang-format` option for that is: `AlignConsecutiveDeclarations: true`
It _does_ look nice most of the time, but there are quite a few places where it looks dreadful.
Here's one:

```c
int mutt_view_attachment(FILE *fp, struct Body *a, enum ViewAttachMode mode,
                         struct Email *e, struct AttachCtx *actx, struct MuttWindow *win)
{  
  bool                 use_mailcap = false;
  bool                 use_pipe = false;
  bool                 use_pager = true;
  char                 type[256] = { 0 };  
  char                 desc[256] = { 0 };
  char                *fname = NULL;    
  struct MailcapEntry *entry = NULL;       
  int                  rc = -1;
  bool                 has_tempfile = false;
  bool                 unlink_pagerfile = false;
```

Diffstat: 403 files changed, 3419 insertions(+), 3418 deletions(-)

Adding `AlignConsecutiveAssignments: true` makes it a bit more palatable:

```c
int mutt_view_attachment(FILE *fp, struct Body *a, enum ViewAttachMode mode,
                         struct Email *e, struct AttachCtx *actx, struct MuttWindow *win)
{  
  bool                 use_mailcap      = false;
  bool                 use_pipe         = false;
  bool                 use_pager        = true;
  char                 type[256]        = { 0 };
  char                 desc[256]        = { 0 };
  char                *fname            = NULL;
  struct MailcapEntry *entry            = NULL; 
  int                  rc               = -1;   
  bool                 has_tempfile     = false;
  bool                 unlink_pagerfile = false;
```

Diffstat: 476 files changed, 7621 insertions(+), 7619 deletions(-)

I need a bit more convincing before I enable them :-)

alejandro-colomar:
> Lovely, thanks!
> 
> > alignment of local variable declaractions.
> 
> I _do_ like a bit of whitespace for alignment (see the header files), but if we do it in the source files, it needs to be automatic.
> 
> We decided very early in NeoMutt's history that using an automatic formatter was the only way to avoid arguments :-) We picked `clang-format` and a set of options that closely matched Mutt's source.
> 
> The `clang-format` option for that is: `AlignConsecutiveDeclarations: true` It _does_ look nice most of the time, but there are quite a few places where it looks dreadful. Here's one:
> 
> ```c
> int mutt_view_attachment(FILE *fp, struct Body *a, enum ViewAttachMode mode,
>                          struct Email *e, struct AttachCtx *actx, struct MuttWindow *win)
> {  
>   bool                 use_mailcap = false;
>   bool                 use_pipe = false;
>   bool                 use_pager = true;
>   char                 type[256] = { 0 };  
>   char                 desc[256] = { 0 };
>   char                *fname = NULL;    
>   struct MailcapEntry *entry = NULL;       
>   int                  rc = -1;
>   bool                 has_tempfile = false;
>   bool                 unlink_pagerfile = false;
> ```
> 
> Diffstat: 403 files changed, 3419 insertions(+), 3418 deletions(-)
> 
> Adding `AlignConsecutiveAssignments: true` makes it a bit more palatable:
> 
> ```c
> int mutt_view_attachment(FILE *fp, struct Body *a, enum ViewAttachMode mode,
>                          struct Email *e, struct AttachCtx *actx, struct MuttWindow *win)
> {  
>   bool                 use_mailcap      = false;
>   bool                 use_pipe         = false;
>   bool                 use_pager        = true;
>   char                 type[256]        = { 0 };
>   char                 desc[256]        = { 0 };
>   char                *fname            = NULL;
>   struct MailcapEntry *entry            = NULL; 
>   int                  rc               = -1;   
>   bool                 has_tempfile     = false;
>   bool                 unlink_pagerfile = false;
> ```

I usually avoid initialization with declarations.  Regarding my prefered style, you can see an example in this code from my previous job at nginx:

https://www.alejandro-colomar.es/src/alx/nginx/unit.git/tree/src/nxt_http_compress_gzip.c?h=gzip-v37#n174

Our rule was:

-  sort the types from shortest to longest.
-  when two types are equal, sort alphabetically.
-  leave two spaces between the longest type and the variable.

Then, in the NGINX project, the start of the pointer went to the left of the aligned variable names, but in the newer Unit project, it was aligned as if it was part of the variable name.  Both of them look well to me (but the latter slightly better).

I'm not sure if clang-format can do that; maybe it can.

> 
> Diffstat: 476 files changed, 7621 insertions(+), 7619 deletions(-)
> 
> I need a bit more convincing before I enable them :-)

alejandro-colomar:
> Lovely, thanks!

Cheers!

flatcap:
> I usually avoid initialization with declarations

I _do_ insist on pointers being initialised immediately.
Here's a [quick guide to code style](https://neomutt.org/dev/code).
(Some day, I'll finish it)

> the start of the pointer went to the left of the aligned variable names,
> but in the newer Unit project, it was aligned as if it was part of the variable name

I quite like the first one, where the "human" bits (the names) are aligned.
The header files are far from consistent, but sometime I experiment with whitespace :-)

https://github.com/neomutt/neomutt/blob/6675b175aed5f960536f67159a8bc37096a73b5e/alias/gui.h#L51-L62

alejandro-colomar:
On Mon, Oct 30, 2023 at 06:57:26AM -0700, Richard Russon wrote:
> > I usually avoid initialization with declarations
> 
> I _do_ insist on pointers being initialised immediately.
> Here's a [quick guide to code style](https://neomutt.org/dev/code).
> (Some day, I'll finish it)

A minor disadvantage of this (IMO) is the following:

-  If you don't initialize, and use some value uninitialized in some
   rare path, the compiler (especially now with gcc's -fanalyzer option)
   can warn you that you forgot to initialize the variable in that path.
   So, you can fix the bug by inspecting the code that triggers the
   warning, and find out a good value for it.

-  However, if you initalize it at the declaration, the compiler can't
   know if that value is invalid or not, and bugs don't trigger any
   warning (as they're second-order bugs), so you're on your own.  You
   don't have Undefined Behavior, but the behavior is still unintended.

-- 
<https://www.alejandro-colomar.es/>

PR #4089, opened 2023-10-29
fix logic for growing mailbox memory

jpalus:
* **What does this PR do?**

Plugs memory leak when growing memory space for mailbox.

* **What are the relevant issue numbers?**

Fixes #4087

Issue #4087, opened 2023-10-28
Memory leak after opening maildir mailbox


jpalus:
## Expected Behaviour

Memory use remains constant when reopening over and over same, unmodified maildir mailbox.

## Actual Behaviour

Memory use grows with every open.

## Steps to Reproduce

1. Open neomutt which by default goes to small maildir mailbox
2. Go to maildir mailbox that has bigger number of mails (~100K in my case)
3. Open big mailbox a couple of times
4. Go back to small maildir mailbox

Observe difference in memory between first and last step.

It appears to be caused by memory not being freed:
https://github.com/neomutt/neomutt/blob/fa83255d2e5cee1a34dcaf108a06107942b59865/maildir/shared.c#L95

as indicated by `massif` after step 4:
```
99.88% (165,580,230B) (heap allocation functions) malloc/new/new[], --alloc-fns, etc.
->98.03% (162,515,648B) 0x518404: mutt_mem_realloc (memory.c:131)
| ->65.31% (108,265,600B) 0x43033F: mx_alloc_memory (mx.c:1260)
| | ->65.31% (108,265,600B) 0x4B6E46: maildir_move_to_mailbox (shared.c:95)
| |   ->65.31% (108,265,600B) 0x4B1E40: maildir_read_dir (maildir.c:731)
| |     ->65.31% (108,265,600B) 0x4B1FEA: maildir_mbox_open (maildir.c:1127)
| |       ->65.31% (108,265,600B) 0x431563: mx_mbox_open (mx.c:388)
| |         ->65.31% (108,265,600B) 0x43BCAC: change_folder_mailbox (dlg_index.c:696)
| |         | ->65.31% (108,265,600B) 0x44035F: op_main_change_folder (functions.c:1156)
| |         |   ->65.31% (108,265,600B) 0x4428F3: index_function_dispatcher (functions.c:3323)
| |         |     ->65.31% (108,265,600B) 0x43C782: dlg_index (dlg_index.c:1343)
| |         |       ->65.31% (108,265,600B) 0x40C7B6: main (main.c:1392)
| |         |         
| |         ->00.00% (0B) in 1+ places, all below ms_print's threshold (01.00%)
| |         
| ->32.65% (54,132,800B) 0x430350: mx_alloc_memory (mx.c:1261)
| | ->32.65% (54,132,800B) 0x4B6E46: maildir_move_to_mailbox (shared.c:95)
[...]

PR #4090, opened 2023-10-29
Mx alloc overhead reduction

jpalus:
* **What does this PR do?**

Improves performance when increasing memory space required for mailbox.

Note that this PR is made on top of #4089. By no means this PR should be merged before #4087 is fixed as increasing `grow` to `1000` would absolutely blow memory requirements.

Related to #4087.


flatcap (approved):
What a disaster area `mx_alloc_memory()` is! :-)

Very nice fix.  Thanks :heart:

PR #4094, opened 2023-10-30
remove MxOps::path_pretty() and mutt_path_pretty()

roccoblues:
I'm probably missing something but this seems to be unused?


flatcap:
This was a step towards a Path API.

The `Path` would encapsulate the Mailbox's path in several forms, e.g.
- As entered by the user
- Canonical
- Pretty

This would make future lookups and comparisons efficient.

`struct Path` and its flags:
https://github.com/neomutt/neomutt-old/blob/f0d61bfb8f560bc7b49700476fb0c74c9e70d463/core/path.h#L29-L47

Path API:
https://github.com/neomutt/neomutt-old/blob/f0d61bfb8f560bc7b49700476fb0c74c9e70d463/mx.h#L368-L511

flatcap (approved):
Path2 will return! :-)
But for now, this is just clutter.

PR #4096, opened 2023-11-05
hcache: fix key handling

flatcap:
Fix a couple of bugs in the header cache key handling.

- maildir: factor out key location
- realkey(): honour key length

Given an Email with path, **`cur/1462463620.M306996P28816.flatcap.org,S=4057,W=4147:2,RS`**
Maildir was adding 3 to the start, which meant that the **`/`** was included.
This has been changed to 4, so that the **filename** is the base of the key.

In the Header Cache, `realkey()` performs a transform on the key depending on whether compression is enabled.
Unfortunately, both branches were producing the wrong output.
When compression was enabled, the entire filename was used -- this includes the flags: **`:2,RS`**
When compression wasn't enabled, the `memcpy()` assumed the string was `NUL`-terminated at keylen.
This lead to a stray **`:`**  being included.

The fixes mean that for path, **`cur/1462463620.M306996P28816.flatcap.org,S=4057,W=4147:2,RS`**
the key is either:
- `1462463620.M306996P28816.flatcap.org,S=4057,W=4147`
- `1462463620.M306996P28816.flatcap.org,S=4057,W=4147-zstd`

---

**Note**:  This will caused the invalidation of all cached maildir headers.

### DIFF
--- a/hcache/hcache.c
+++ b/hcache/hcache.c
@@ -292,12 +292,13 @@ static struct RealKey *realkey(struct HeaderCache *hc, const char *key, size_t k
 #ifdef USE_HCACHE_COMPRESSION
   if (hc->compr_ops)
   {
-    rk.len = snprintf(rk.key, sizeof(rk.key), "%s-%s", key, hc->compr_ops->name);
+    rk.len = snprintf(rk.key, sizeof(rk.key), "%.*s-%s", (int) keylen, key,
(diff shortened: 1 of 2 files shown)


roccoblues (approved):
Looks good. 👍🏼

flatcap:
**Update**: renamed `hcache_key()` to `maildir_hcache_key()`

PR #4097, opened 2023-11-05
config: add DT_ON_STARTUP

flatcap:
Some config variables may only be set at startup.
This can be either in a system/user config file or using the `-e` command line. 

- 95ff7c418 config: add more sameness checks
- 466a3fac0 config: create a new flag, `DT_ON_STARTUP`
- 6f95fdb79 config: add tests for DT_ON_STARTUP
- 4b66a50ba config: apply DT_ON_STARTUP
  - `$abort_key`
  - `$color_directcolor`
  - `$crypt_use_gpgme`
  - `$maildir_field_delimiter`

### DIFF
--- a/config/bool.c
+++ b/config/bool.c
@@ -81,6 +81,9 @@ static int bool_string_set(const struct ConfigSet *cs, void *var, struct ConfigD
     if (num == (*(bool *) var))
       return CSR_SUCCESS | CSR_SUC_NO_CHANGE;
 
+    if (startup_only(cdef, err))
+      return CSR_ERR_INVALID | CSR_INV_VALIDATOR;
+
     if (cdef->validator)
     {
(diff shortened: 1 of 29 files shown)


gahr:
Nice! Would it be best to only error out if we're *changing* the current value? I can think of having a config file that's sourced at different times, possibly, and that should be fine as long as I'm always setting the same value.

flatcap:
Yeah, good idea.  It'll take a little work though.

atm, I'm trapping the commands at the `ConfigSet` level.
I'll need to move the tests down into each individual config type.

flatcap:
Done.
Tests passing, Coverage 100% :-)

gahr (approved):
Nice - I haven't read all the code, but I get the idea :)

PR #4098, opened 2023-11-05
build: use fallthrough attribute

flatcap:
Use the statement attribute: `__attribute__((fallthrough));` to tell the compiler where fallthrough is permitted.

Previously, `case` statements that we allowed to fallthrough were marked with `/* fallthrough */`
The comment was used as a hint for Coverity.

The attribute is supported by gcc, clang and coverity.


ossilator:
it's still a gcc-ism.
the convention is to wrap it into a macro, which can easily cover msvc and the c23 standard [[fallthrough]].

flatcap:
- e921517e3 macro
- b3f48afe9 auto.def
  Protect the build from crazies using msvc

gahr (approved):

PR #4105, opened 2023-11-12
use buffer pool in tests

roccoblues:


flatcap (approved):
Lovely!
I backdated the fixes.

PR #4106, opened 2023-11-13
switch some buffers to use the buffer pool

roccoblues:
This should be slightly better because we do less small allocations by reusing the buffers.


flatcap (approved):
lovely, thanks!

PR #4108, opened 2023-11-15
strip non-conditionals

flatcap:
Remove nearly 700 lines of clutter.

---

Strip out over 300 conditionals (`#if`) for symbols that are always defined.

- `USE_COMP_MBOX`
- `USE_IMAP`
- `USE_NNTP`
- `USE_POP`
- `USE_SIDEBAR`
- `USE_SOCKET`
- `USE_SMTP`
- `SUN_ATTACHMENT`

These were permanent set to true, long ago, because they didn't add any dependencies to the build.

### DIFF
--- a/Makefile.autosetup
+++ b/Makefile.autosetup
@@ -803,7 +803,6 @@ $(PWD)/send:
 
 ###############################################################################
 # libsidebar
-@if USE_SIDEBAR
 LIBSIDEBAR=	libsidebar.a
 LIBSIDEBAROBJS=	sidebar/commands.o sidebar/config.o sidebar/functions.o \
 		sidebar/observer.o sidebar/sidebar.o sidebar/sort.o \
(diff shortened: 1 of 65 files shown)


gahr:
YES!

gahr (approved):

roccoblues:
👍🏼 I'm just wondering about `SUN_ATTACHMENT`. Do we really want/need to support that in future versions? Especially given comments like [/* this is deep and utter perversion */](https://github.com/neomutt/neomutt/blob/main/email/parse.c#L451) 😆

roccoblues:
> 👍🏼 I'm just wondering about `SUN_ATTACHMENT`. Do we really want/need to support that in future versions? Especially given comments like [/* this is deep and utter perversion */](https://github.com/neomutt/neomutt/blob/main/email/parse.c#L451) 😆

As @gahr pointed out on IRC, people will use new Neomutt versions to read old emails. So it makes sense to still support this.

PR #4111, opened 2023-11-21
Educate Coverity about ARRAYs

flatcap:
Use `__coverity_escape__()` to tell Coverity that memory isn't being leaked.
This only affects Coverity's build.

---

Coverity has trouble tracking memory allocations in our `ARRAY` and `TAILQ` macros.
This leads to false-positive defects.

Redefine `ARRAY_ADD_NORESERVE()` to call `__coverity_escape__()`.

"Fixes":
- [Defect 215478](https://scan9.scan.coverity.com/reports.htm#v34346/p12145/fileInstanceId=124135703&defectInstanceId=9495496&mergedDefectId=215478)
- [Defect 215480](https://scan9.scan.coverity.com/reports.htm#v34346/p12145/fileInstanceId=124135699&defectInstanceId=9456300&mergedDefectId=215480)
- [Defect 257775](https://scan9.scan.coverity.com/reports.htm#v34346/p12145/fileInstanceId=124135455&defectInstanceId=9475297&mergedDefectId=257775)
- [Defect 333337](https://scan9.scan.coverity.com/reports.htm#v34346/p12145/fileInstanceId=124135703&defectInstanceId=9495495&mergedDefectId=333337)

It might be possible to do something similar for the `TAILQ` macros.

### DIFF
--- a/mutt/array.h
+++ b/mutt/array.h
@@ -289,15 +289,13 @@
       true)                                                                    \
    : false)
 
+#ifndef __COVERITY__
+#define __coverity_escape__(x) 0
+#endif
 #define ARRAY_ADD_NORESERVE(head, elem)                                        \
   ((head)->capacity > (head)->size                                             \
    ? (((head)->entries[(head)->size++] = (elem)),                              \
-      true)                                                                    \
-   : false)
-
-#ifdef __COVERITY__
-#undef ARRAY_ADD_NORESERVE
-#define ARRAY_ADD_NORESERVE(head,elem) __coverity_escape__(elem);
-#endif
+      __coverity_escape__(elem), true)                                         \
+   : __coverity_escape__(elem), false)
 
 #endif /* MUTT_MUTT_ARRAY_H */


flatcap:
Yep, coverity's happy with the new version

gahr (approved):

flatcap:
Hmm... bad news.
I applied your patch, built it, pushed it to coverity and everything was good.
Merged.

However, I built again later and it threw lots of warnings.
clang's happy, but gcc isn't.
```
mutt/array.h:297:49: warning: right-hand operand of comma expression has no effect [-Wunused-value]
mutt/array.h:299:31: warning: right-hand operand of comma expression has no effect [-Wunused-value]
```

Looking again, we can kill the `__coverity_escape__()` in the `false` case.
The object **hasn't** been added to the array, so we shouldn't tell coverity that it **has** :-)

I'm not sure what the best solution is for the case:
```c
entries[size] = elem, __coverity_escape__(elem), true
```

gahr:
Nice catch about us not wanting to mute coverity in the false case. I suggest
```diff
diff --git a/mutt/array.h b/mutt/array.h
index 9ea8dadc7..878a78650 100644
--- a/mutt/array.h
+++ b/mutt/array.h
@@ -295,7 +295,7 @@
 #define ARRAY_ADD_NORESERVE(head, elem)                                        \
   ((head)->capacity > (head)->size                                             \
    ? (((head)->entries[(head)->size++] = (elem)),                              \
-      __coverity_escape__(elem), true)                                         \
-   : __coverity_escape__(elem), false)
+     ((void)__coverity_escape__(elem), true))                                  \
+   : false)
 
 #endif /* MUTT_MUTT_ARRAY_H */
```

flatcap:
Thanks, that sorts out the warnings.

Unfortunately, removing the `__coverity_escape__()` from the `false` case, now causes defects again.
I'm not going to investigate exactly _why_ that's the case.

Fortunately, **we** know that `ARRAY_ADD_NORESERVE()` is only called after `ARRAY_RESERVE()`,
and `ARRAY_RESERVE()` will either succeed or `panic()`.

So, the final version [[devel/coverity]](https://github.com/neomutt/neomutt/tree/devel/coverity) is:
```c
#ifndef __COVERITY__
#define __coverity_escape__(x) 0
#endif
#define ARRAY_ADD_NORESERVE(head, elem)                                        \
  ((head)->capacity > (head)->size                                             \
   ? (((head)->entries[(head)->size++] = (elem)),                              \
     ((void)__coverity_escape__(elem), true))                                  \
   : ((void)__coverity_escape__(elem), false))
```

PR #4115, opened 2023-11-25
create HelpBar after colours

flatcap:
The HelpBar has a colour observer, so it needs to be created after the colours are initialised.

**Test**, in NeoMutt:

- Enable the Help Bar, `set help = yes`
- Create a hook: `folder-hook . 'color status black yellow'`
- Change folder using `<sidebar-open>`

Both the Status Bar (Index) and the Help Bar should change colour.

### DIFF
--- a/main.c
+++ b/main.c
@@ -779,10 +779,6 @@ main
   if (!OptNoCurses && !isatty(1))
     goto main_curses;
 
-  /* Always create the mutt_windows because batch mode has some shared code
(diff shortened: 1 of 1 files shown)


roccoblues (approved):
Tested and works. 👍🏼 

Context from IRC: the colors changed correctly when switching folders with `<change-folder>` because that causes a refresh. Only `<sidebar-open>` didn't work.

PR #4117, opened 2023-11-26
remove MxOps::path_parent() and mutt_path_parent()

roccoblues:
This is unused and probably a leftover similar to: https://github.com/neomutt/neomutt/pull/4094#issuecomment-1790581457.


flatcap (approved):
thanks!

PR #4121, opened 2023-12-01
mutt_oauth2.py: error out if ENCRYPTION_PIPE was not supplied

mgerstner:
Currently when this is not set on the command line then a complex error backtrace is the result, since the script tries to run `None` via subprocess.

Thus require the command line argument, but only if we are actually going to need encryption.

* **What does this PR do?**

     for that)


flatcap (approved):
Thank you very much!

PR #4122, opened 2023-12-01
add github action to check for unused functions (xunused)

roccoblues:
As discussed here: https://github.com/neomutt/neomutt/discussions/4118

Sample job run: https://github.com/neomutt/neomutt/actions/runs/7223904797/job/19683982035?pr=4122

### DIFF
--- a/.github/unused.sh
+++ b/.github/unused.sh
@@ -1,0 +1,22 @@
+#!/usr/bin/env bash
+
+set -euo pipefail
+
+EXIT_CODE=0
+IGNORELIST=("progress_set_size" "mutt_socket_buffer_readln_d")
+
+while IFS= read -r line; do
+    if ! [[ $line =~ "is unused" ]]; then
+        continue
+    fi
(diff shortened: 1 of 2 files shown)


roccoblues:
Improvements discussed on IRC:
- [ ] output only filename:line function
- [ ] add "ignore list" to ignore false positives
- [ ] fail job if we have unused functions not on the ignore list

flatcap (commented):
I've made a few minor tweaks:
- I added the packages to the [Docker](https://github.com/neomutt/docker) to save installing them each time
- I swapped `--full-doc` for `--disable-doc` to save CPU
  (I don't _think_ we need the docs)
- I re-titled the Action to "Unused Functions" to keep the name short for the [Actions List](https://github.com/neomutt/neomutt/actions)

flatcap (approved):
Very nice!
I'm sure we can think of more things to tweak,
but let's merge it and start using it first.

PR #4124, opened 2023-12-06
config: fix flag overlaps

flatcap:
Adding the `DT_ON_STARTUP` config flag, in 9a904e578, exposed a bug.
Some of the config flags overlapped causing a bug in the browser.

Trying to set the file mask in the browser <kbd>m</kbd> (`<enter-mask>`)
fails with the message: "Option mask may only be set at startup".

---

Each config variable has a type, e.g. `DT_STRING`.
Additionally, they have flags `OR`d with the type, e.g. `DT_NOT_EMPTY`.

`$mask` has the flag `DT_REGEX_MATCH_CASE` which had the same value as `DT_ON_STARTUP`.
This caused the bug.

Of the 471 config variables, just 8 have flags from multiple sets of flags:
 
```c
{ "charset",            DT_NOT_EMPTY|DT_CHARSET_SINGLE },
{ "group_index_format", DT_NOT_EMPTY|R_INDEX },
{ "index_format",       DT_NOT_EMPTY|R_INDEX },
{ "mbox",               DT_MAILBOX|R_INDEX },
{ "postponed",          DT_MAILBOX|R_INDEX },
{ "send_charset",       SLIST_SEP_COLON|SLIST_ALLOW_EMPTY|DT_CHARSET_STRICT },
{ "sort",               DT_SORT_REVERSE|DT_SORT_LAST|R_INDEX|R_RESORT },
{ "sort_aux",           DT_SORT_REVERSE|DT_SORT_LAST|R_INDEX|R_RESORT|R_RESORT_SUB },
```

They can be confirmed to be non-overlapping.

https://github.com/neomutt/neomutt/blob/349c9824a0efcb348547e38e0568d69f3bbe4c72/config/types.h#L49-L58
https://github.com/neomutt/neomutt/blob/349c9824a0efcb348547e38e0568d69f3bbe4c72/config/types.h#L69-L72
https://github.com/neomutt/neomutt/blob/349c9824a0efcb348547e38e0568d69f3bbe4c72/config/sort2.h#L31-L32
https://github.com/neomutt/neomutt/blob/349c9824a0efcb348547e38e0568d69f3bbe4c72/mutt/slist.h#L33-L41
https://github.com/neomutt/neomutt/blob/349c9824a0efcb348547e38e0568d69f3bbe4c72/config/charset.h#L28-L29
https://github.com/neomutt/neomutt/blob/349c9824a0efcb348547e38e0568d69f3bbe4c72/mutt/regex3.h#L35-L37

### DIFF
--- a/config/charset.h
+++ b/config/charset.h
@@ -25,6 +25,8 @@
 
 #include <stdint.h>
 
+/* Note: To save space, sets of config variable flags are packed into a uint32_t.
+ * When adding flags, check all config variables to ensure there are no overlaps of values */
 #define DT_CHARSET_SINGLE  (1 << 24) ///< Flag for charset_validator to allow only one charset
 #define DT_CHARSET_STRICT  (1 << 25) ///< Flag for charset_validator to use strict char check
 
--- a/config/charset.h
+++ b/config/charset.h
@@ -25,8 +25,8 @@
 
 #include <stdint.h>
 
-#define DT_CHARSET_SINGLE  (1 << 11) ///< Flag for charset_validator to allow only one charset
-#define DT_CHARSET_STRICT  (1 << 12) ///< Flag for charset_validator to use strict char check
+#define DT_CHARSET_SINGLE  (1 << 24) ///< Flag for charset_validator to allow only one charset
+#define DT_CHARSET_STRICT  (1 << 25) ///< Flag for charset_validator to use strict char check
 
 struct Buffer;
 struct ConfigDef;
--- a/config/sort2.h
+++ b/config/sort2.h
@@ -28,6 +28,8 @@
 extern const struct Mapping SortMethods[];
 
 /* ... DT_SORT */
+/* Note: To save space, sets of config variable flags are packed into a uint32_t.
+ * When adding flags, check all config variables to ensure there are no overlaps of values */
 #define DT_SORT_LAST    (1 << 21) ///< Sort flag for -last prefix
 #define DT_SORT_REVERSE (1 << 22) ///< Sort flag for -reverse prefix
 
--- a/config/sort2.h
+++ b/config/sort2.h
@@ -28,8 +28,8 @@
 extern const struct Mapping SortMethods[];
 
 /* ... DT_SORT */
-#define DT_SORT_LAST    (1 << 11) ///< Sort flag for -last prefix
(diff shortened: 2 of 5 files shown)


roccoblues:
Maybe add comments to the define blocks to check for overlaps when we add new values?

Or, not sure if it's possible, can we add a test for overlaps?

flatcap:
> Maybe add comments

Good call, done.

> not sure if it's possible, can we add a test for overlaps?

I don't think so, because all the flags are `#define`d -- they're replaced by the preprocessor.
Any solution would have to parse the code.

I did it manually.  I 'grep'd out all the config definitions, then tidied them with vim-fu.
Next, I used vim highlighing to give each group a different colour.
I eliminated all the config with just a single colour, leaving the 8 above to check.

**vim patterns** :-)
```vim
\v<(DT_ADDRESS|DT_BOOL|DT_ENUM|DT_HCACHE|DT_LONG|DT_MBTABLE|DT_MYVAR|DT_NUMBER|DT_PATH|DT_QUAD|DT_REGEX|DT_SLIST|DT_SORT|DT_STRING|DT_SYNONYM)>
\v<(DT_ON_STARTUP|DT_NOT_EMPTY|DT_NOT_NEGATIVE|DT_MAILBOX|DT_SENSITIVE|DT_COMMAND|DT_INHERIT_ACC|DT_INHERIT_MBOX|DT_PATH_DIR|DT_PATH_FILE)>
\v<(R_INDEX|R_RESORT|R_RESORT_SUB|R_RESORT_INIT)>
\v<(DT_SORT_LAST|DT_SORT_REVERSE)>
\v<(SLIST_SEP_SPACE|SLIST_SEP_COMMA|SLIST_SEP_COLON|SLIST_ALLOW_DUPES|SLIST_ALLOW_EMPTY|SLIST_CASE_SENSITIVE)>
\v<(DT_CHARSET_SINGLE|DT_CHARSET_STRICT)>
\v<(DT_REGEX_MATCH_CASE|DT_REGEX_ALLOW_NOT|DT_REGEX_NOSUB)>
\v<(DT_FREE_CONFIGDEF|DT_DEPRECATED|DT_INHERITED|DT_INITIAL_SET|DT_DISABLED)>
```

roccoblues (approved):

PR #4133, opened 2023-12-16
move log_disp_null() into test folder

roccoblues:
This function is only used by the tests and the fuzzer.

As discussed here: https://github.com/neomutt/neomutt/discussions/4118#discussioncomment-7686223


flatcap (approved):
Lovely, thanks!
Keep 'em coming :-)

PR #4135, opened 2023-12-17
remove unused functions

roccoblues:
Removes the following unused functions:

* `buf_upper()`
* `slist_add_list()`
* `slist_empty()`
* `mutt_file_chmod()`
* `mutt_file_chmod_rm()`
* `mutt_perror_debug()`
* `quad_str_toggle()`
* `cs_subset_str_native_get()`
* `cs_subset_str_string_plus_equals()`
* `cs_subset_str_string_plus_equals()`
* `cs_subset_str_string_minus_equals()`
* `cs_subset_str_delete()`

As discussed here: https://github.com/neomutt/neomutt/discussions/4118#discussioncomment-7686223


roccoblues:
@flatcap I'm not sure about:

```
cs_str_delete (config/set.c:1083)
cs_str_native_get (config/set.c:885)
cs_str_string_get (config/set.c:706)
cs_str_string_minus_equals (config/set.c:1044)
cs_str_string_plus_equals (config/set.c:961)
```

They are all used in multiple tests and I'm not sure if I can simply remove the tests or if they test something other than the functions themselves.

flatcap:
> I'm not sure if I can simply remove the tests or if they test something other than the functions themselves.

Ah, yes.
The config tests pre-date the unit tests and they're structured a little differently.
They tend to be centred around each **config type**, rather than individual functions.
I went a bit overboard to ensure 100% code coverage :-)

The solution is probably as simple as deleting every reference to the functions. 
I'll try to have a dig around, later this evening.

flatcap:
I've moved the comments into the `@page` so they show up in the Doxygen docs..

flatcap (approved):
Lovely, thanks!

flatcap:
> The solution is probably as simple as deleting every reference to the functions.

Nope :-)

In the test code we tend to use the **string** name for config variables, so it turns out we need these functions.
We _could_ eliminate them, but that'd mean quite a bit of refactoring (of test code).

I think the best solution is to demote them to `test/config/common.c`
Thanks

PR #4137, opened 2023-12-18
move unused config string name functions into tests

roccoblues:
The following functions are only used by the tests. I've moved them into `test/config/common.[ch]`.

* `cs_str_string_get()`
* `cs_str_native_get()`
* `cs_str_string_minus_equals()`
* `cs_str_string_plus_equals()`
* `cs_str_delete()`


flatcap (approved):
Great.  Coverage still at 100% :-)

PR #4142, opened 2023-12-21
add mutt_time_now()

roccoblues:
Abstracts away the `#ifdef` and also uses nanosecond precision for the mbox `stats_last_checked` field if available.

Nanosecond precision was introduced in commit 42b436392846f06ebbc01dba7642361f163a8aad to fix high resolution issues with the inotify interface.


flatcap (approved):
Nice!

PR #4146, opened 2023-12-23
config: factor out R_ flags

flatcap:
I should have done this long ago...

---

Kill the R_ flags that tell the Index what to do.

- R_INDEX
- R_RESORT
- R_RESORT_INIT
- R_RESORT_SUB

Instead, check for specific config options.

---

#4124 (config: fix flag overlaps) fixed one problem, but introduced another.
`$reply_regex` was left with conflicting config flags.

### DIFF
--- a/browser/config.c
+++ b/browser/config.c
@@ -61,7 +61,7 @@ static struct ConfigDef BrowserVars[] = {
   { "folder_format", DT_STRING|DT_NOT_EMPTY, IP "%2C %t %N %F %2l %-8.8u %-8.8g %8s %d %i", 0, NULL,
     "printf-like format string for the browser's display of folders"
   },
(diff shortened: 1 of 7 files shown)


gahr:
I'm curious: why listing a config in multiple places in the code is better than "tagging" the config with flags, in a declarative way and in a single place, in your opinion?

flatcap:
It's the scope of the `R_` flags that annoys me -- they're everywhere.
Except, ironically: `index/config.c`
https://github.com/neomutt/neomutt/blob/8e3ea66819fce92a334947009ed89351dbf1fab3/index/config.c#L39-L53

The Index is complex and is affected by a lot of config, but _some_ of that config shouldn't be aware of the Index.
e.g.
- `$imap_headers`
- `$postponed`
- `$reverse_alias`

However, it's OK if a component wants to mark its config with flags that **aren't visible** outside of _that_ library.

PR #4147, opened 2023-12-23
envelope: manage subject/real_subj together

flatcap:
As it stands, subject is assigned at will throughout every interested piece of code, and real_subj is parsed with $reply_regex and set in three unaffiliated places.

This means that, for example, IMAP mailboxes simply do not have $reply_regex processing. Unless you change it, then they refresh to have it. Until you reopen and then it's lost.

Mark both fields as const and introduce a mutt_env_set_subject() setter that sets subject and parses real_subj. This interior mutability is bypassed only in hcache serialisation, by necessity. rfc2047 (un)parsing needs to abuse it slightly to match the calling convention.

This ensures that real_subj always matches subject, and so $reply_regex works as documented instead of not at all.

Fixes: https://mailman.neomutt.org/pipermail/neomutt-devel-neomutt.org/2023-December/000897.html

### DIFF
--- a/email/envelope.c
+++ b/email/envelope.c
@@ -33,6 +33,8 @@
 #include <string.h>
 #include "mutt/lib.h"
 #include "address/lib.h"
+#include "config/lib.h"
+#include "core/lib.h"
 #include "envelope.h"
 #include "email.h"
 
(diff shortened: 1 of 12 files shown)


flatcap (commented):
Very nice!  Thanks!
(I can't _approve_, because **I** created the Pull Request :-)

gahr:
I'm not a fan of the fake `const` as a way of documenting that subject shouldn't be set manually, then doing it in a few places...  but of course, a fixed bug is better than an unfixed one.

gahr:
I guess an alternative would be to compute `real_subj` on demand... would that be too slow?

roccoblues:
> I guess an alternative would be to compute `real_subj` on demand... would that be too slow?

It's used in a couple of places. The most expensive might be "sorting by subject":
https://github.com/neomutt/neomutt/blob/8e3ea66819fce92a334947009ed89351dbf1fab3/sort.c#L100-L120

PR #4149, opened 2023-12-27
Respect umask when saving attachments

### LINKED ISSUE #4148: Respect `umask` when saving attachments

I use [Syncthing](https://syncthing.net) to sync documents, that sometimes I get on email. I run `syncthing` as a system service, with it's own user and group both named `syncthing`. My unix user is part of the `syncthing` group, and my most of my files at `~/documents` belong to the `syncthing` group - thus enabling the `syncthing.service` sync those files.

I aliased `neomutt` to `sg syncthing -c neomutt`, but still, it seems that attachments are saved with the permissions `500` by default, and there is no way to change that, and it doesn't matter what is my `umask`.

It would have been great if `neomutt` would have honored `umask`, or if it would have been possible to configure it to honor `umask` settings.

- Macros: I was thinking of using a macro for that, but I don't know how the macro will know the path of the file I just saved with the `<save-entry>` command - seems impossible to me.
- Hooks: I didn't find a `<save-entry>` hook in the docs.
- Scripts: I could write a script that would completely override the interface with which I save attachments, but I'm not sure how to do that.

doronbehar:
* **What does this PR do?**
Adds a new option that allows for umask to be respected when attachments are saved, or when the attachments are viewed in an external program. Fixes #4148.

* **Does this PR meet the acceptance criteria?** Not yet, that's why I added a TODO list. I'm currently failing one of the tests, and probably righteously, since an assertion fails. Something is wrong with how I use the `NeoMutt` "global" variable.

 - [ ] Documentation created/updated (you have to edit [docs/manual.xml.head](https://www.github.com/neomutt/neomutt/blob/main/docs/manual.xml.head) for that).
 - [x] All builds and tests are passing.
 - [ ] Added [doxygen code documentation](https://neomutt.org/dev/doxygen) ([syntax](http://www.stack.nl/~dimitri/doxygen/manual/docblocks.html)).
 - [x] Code follows the [style guide](https://neomutt.org/dev/code)

#4148

### DIFF
--- a/core/neomutt.c
+++ b/core/neomutt.c
@@ -229,11 +229,13 @@ FILE *mutt_file_fopen_masked_full(const char *path, const char *mode,
 {
   // Set the user's umask (saved on startup)
   mode_t old_umask = umask(NeoMutt->user_default_umask);
+  mutt_debug(LL_DEBUG3, "umask set to %03o\n", NeoMutt->user_default_umask);
 
   // The permissions will be limited by the umask
   FILE *fp = mutt_file_fopen_full(path, mode, 0666, file, line, func);
 
   umask(old_umask); // Immediately restore the umask
+  mutt_debug(LL_DEBUG3, "umask set to %03o\n", old_umask);
 
   return fp;
 }
--- a/core/neomutt.c
+++ b/core/neomutt.c
@@ -30,6 +30,7 @@
 #include "config.h"
 #include <stddef.h>
 #include <errno.h>
+#include <sys/stat.h>
 #include "mutt/lib.h"
 #include "config/lib.h"
 #include "neomutt.h"
@@ -209,4 +210,30 @@ size_t neomutt_mailboxlist_get_all(struct MailboxList *head, struct NeoMutt *n,
   }
 
   return count;
+}
+
+/**
+ * mutt_file_fopen_masked_full - Wrapper around mutt_file_fopen_full()
+ * @param path  Filename
+ * @param mode  Mode e.g. "r" readonly; "w" read-write
+ * @param file  Source file
+ * @param line  Source line number
+ * @param func  Source function
+ * @retval ptr  FILE handle
+ * @retval NULL Error, see errno
+ *
+ * Apply the user's umask, then call mutt_file_fopen_full().
+ */
+FILE *mutt_file_fopen_masked_full(const char *path, const char *mode,
+                                  const char *file, int line, const char *func)
+{
+  // Set the user's umask (saved on startup)
+  mode_t old_umask = umask(NeoMutt->user_default_umask);
+
+  // The permissions will be limited by the umask
+  FILE *fp = mutt_file_fopen_full(path, mode, 0666, file, line, func);
+
+  umask(old_umask); // Immediately restore the umask
+
+  return fp;
 }
--- a/core/neomutt.h
+++ b/core/neomutt.h
@@ -45,6 +45,7 @@ struct NeoMutt
   struct ConfigSubset *sub;      ///< Inherited config items
   struct AccountList accounts;   ///< List of all Accounts
   locale_t time_c_locale;        ///< Current locale but LC_TIME=C
+  mode_t user_default_umask;     ///< User's default file writing permissions (inferred from umask)
 };
 
 extern struct NeoMutt *NeoMutt;
@@ -68,5 +69,8 @@ struct NeoMutt *neomutt_new           (struct ConfigSet *cs);
 
 void   neomutt_mailboxlist_clear  (struct MailboxList *ml);
 size_t neomutt_mailboxlist_get_all(struct MailboxList *head, struct NeoMutt *n, enum MailboxType type);
+
+// Similar to mutt_file_fopen, but with the proper permissions inferred from
+#define mutt_file_fopen_masked(PATH, MODE) mutt_file_fopen_masked_full(PATH, MODE, __FILE__, __LINE__, __func__)
 
 #endif /* MUTT_CORE_NEOMUTT_H */
--- a/main.c
+++ b/main.c
@@ -595,8 +595,6 @@ main
 
   init_locale();
 
-  umask(077);
-
   EnvList = envlist_init(envp);
   for (optind = 1; optind < double_dash;)
   {
@@ -754,6 +752,9 @@ main
 
   NeoMutt = neomutt_new(cs);
   init_config(cs);
+
+  // Change the current umask, and save the original one
+  NeoMutt->user_default_umask = umask(077);
   subjrx_init();
   attach_init();
   alternates_init();
--- a/main.c
+++ b/main.c
@@ -792,6 +792,8 @@ main
   mutt_log_prep();
   MuttLogger = log_disp_queue;
   log_translation();
+  mutt_debug(LL_DEBUG1, "user's umask %03o\n", NeoMutt->user_default_umask);
+  mutt_debug(LL_DEBUG3, "umask set to 077\n");
 
   if (!STAILQ_EMPTY(&cc_list) || !STAILQ_EMPTY(&bcc_list))
   {
--- a/mutt/file.c
+++ b/mutt/file.c
@@ -569,10 +569,11 @@ const char *mutt_file_rotate(const char *path, int count)
  * mutt_file_open - Open a file
  * @param path  Pathname to open
  * @param flags Flags, e.g. O_EXCL
+ * @param mode  Permissions of the file (Relevant only when writing or appending)
  * @retval >0 Success, file handle
  * @retval -1 Error
  */
-int mutt_file_open(const char *path, uint32_t flags)
+int mutt_file_open(const char *path, uint32_t flags, mode_t mode)
 {
   if (!path)
     return -1;
@@ -592,7 +593,7 @@ int mutt_file_open(const char *path, uint32_t flags)
       goto cleanup;
     }
 
-    fd = open(buf_string(safe_file), flags, 0600);
+    fd = open(buf_string(safe_file), flags, mode);
     if (fd < 0)
     {
       rmdir(buf_string(safe_dir));
@@ -651,6 +652,7 @@ DIR *mutt_file_opendir(const char *path, enum MuttOpenDirMode mode)
  * mutt_file_fopen_full - Call fopen() safely
  * @param path  Filename
  * @param mode  Mode e.g. "r" readonly; "w" read-write
+ * @param perms Permissions of the file (Relevant only when writing or appending)
  * @param file  Source file
  * @param line  Source line number
  * @param func  Source function
@@ -660,8 +662,8 @@ DIR *mutt_file_opendir(const char *path, enum MuttOpenDirMode mode)
  * When opening files for writing, make sure the file doesn't already exist to
  * avoid race conditions.
  */
-FILE *mutt_file_fopen_full(const char *path, const char *mode, const char *file,
-                           int line, const char *func)
+FILE *mutt_file_fopen_full(const char *path, const char *mode, const mode_t perms,
+                           const char *file, int line, const char *func)
 {
   if (!path || !mode)
     return NULL;
@@ -676,7 +678,7 @@ FILE *mutt_file_fopen_full(const char *path, const char *mode, const char *file,
     else
       flags |= O_WRONLY;
 
-    int fd = mutt_file_open(path, flags);
+    int fd = mutt_file_open(path, flags, perms);
     if (fd >= 0)
     {
       fp = fdopen(fd, mode);
--- a/mutt/file.h
+++ b/mutt/file.h
@@ -108,7 +108,8 @@ time_t      mutt_file_decrease_mtime(const char *fp, struct stat *st);
 void        mutt_file_expand_fmt(struct Buffer *dest, const char *fmt, const char *src);
 void        mutt_file_expand_fmt_quote(char *dest, size_t destlen, const char *fmt, const char *src);
 int         mutt_file_fclose_full(FILE **fp, const char *file, int line, const char *func);
-FILE *      mutt_file_fopen_full(const char *path, const char *mode, const char *file, int line, const char *func);
+FILE *      mutt_file_fopen_full(const char *path, const char *mode, const mode_t perms, const char *file, int line, const char *func);
+FILE *      mutt_file_fopen_masked_full(const char *path, const char *mode, const char *file, int line, const char *func);
 int         mutt_file_fsync_close(FILE **fp);
 long        mutt_file_get_size(const char *path);
 long        mutt_file_get_size_fp(FILE* fp);
@@ -117,7 +118,7 @@ bool        mutt_file_iter_line(struct MuttFileIter *iter, FILE *fp, ReadLineFla
 int         mutt_file_lock(int fd, bool excl, bool timeout);
 bool        mutt_file_map_lines(mutt_file_map_t func, void *user_data, FILE *fp, ReadLineFlags flags);
 int         mutt_file_mkdir(const char *path, mode_t mode);
-int         mutt_file_open(const char *path, uint32_t flags);
+int         mutt_file_open(const char *path, uint32_t flags, mode_t mode);
 DIR *       mutt_file_opendir(const char *path, enum MuttOpenDirMode mode);
 char *      mutt_file_read_keyword(const char *file, char *buf, size_t buflen);
 char *      mutt_file_read_line(char *line, size_t *size, FILE *fp, int *line_num, ReadLineFlags flags);
@@ -143,7 +144,8 @@ void        mutt_file_resolve_symlink(struct Buffer *buf);
 void        buf_quote_filename(struct Buffer *buf, const char *filename, bool add_outer);
 void        buf_file_expand_fmt_quote(struct Buffer *dest, const char *fmt, const char *src);
 
-#define mutt_file_fopen(PATH, MODE) mutt_file_fopen_full(PATH, MODE, __FILE__, __LINE__, __func__)
+// Safest default permissions should be 0600
+#define mutt_file_fopen(PATH, MODE) mutt_file_fopen_full(PATH, MODE, 0600, __FILE__, __LINE__, __func__)
 #define mutt_file_fclose(FP)        mutt_file_fclose_full(FP, __FILE__, __LINE__, __func__)
 
 #endif /* MUTT_MUTT_FILE_H */
--- a/attach/mutt_attach.c
+++ b/attach/mutt_attach.c
@@ -542,8 +542,9 @@ int mutt_view_attachment(FILE *fp, struct Body *b, enum ViewAttachMode mode,
     const bool c_wait_key = cs_subset_bool(NeoMutt->sub, "wait_key");
     if (use_pager || use_pipe)
     {
-      if (use_pager && ((fd_pager = mutt_file_open(buf_string(pagerfile),
-                                                   O_CREAT | O_EXCL | O_WRONLY)) == -1))
+      if (use_pager &&
+          ((fd_pager = mutt_file_open(buf_string(pagerfile),
+                                      O_CREAT | O_EXCL | O_WRONLY, 0600)) == -1))
       {
         mutt_perror("open");
         goto return_error;
@@ -745,7 +746,7 @@ int mutt_pipe_attachment(FILE *fp, struct Body *b, const char *path, const char 
 
   if (outfile && *outfile)
   {
-    out = mutt_file_open(outfile, O_CREAT | O_EXCL | O_WRONLY);
+    out = mutt_file_open(outfile, O_CREAT | O_EXCL | O_WRONLY, 0600);
     if (out < 0)
     {
       mutt_perror("open");
--- a/attach/mutt_attach.c
+++ b/attach/mutt_attach.c
@@ -886,11 +886,11 @@ bail:
 static FILE *save_attachment_open(const char *path, enum SaveAttach opt)
 {
   if (opt == MUTT_SAVE_APPEND)
-    return mutt_file_fopen(path, "a");
+    return mutt_file_fopen_masked(path, "a");
   if (opt == MUTT_SAVE_OVERWRITE)
-    return mutt_file_fopen(path, "w");
+    return mutt_file_fopen_masked(path, "w");
 
-  return mutt_file_fopen(path, "w");
+  return mutt_file_fopen_masked(path, "w");
 }
 
 /**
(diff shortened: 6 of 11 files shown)


main.c
```diff
@@ -554,8 +554,6 @@ main
 
   init_locale();
 
-  umask(077);
```
roccoblues, line 557: We need to be careful, currently this ensures that all files created by NeoMutt are only accessible for the current user. The default umask on unix systems is 022. When we remove this we could suddenly create temp files, mbox/maildir files, saved mails and so on that are readable by everyone.

mutt/file.c
```diff
@@ -561,6 +562,17 @@ int mutt_file_open(const char *path, uint32_t flags)
   struct Buffer safe_file = buf_make(0);
   struct Buffer safe_dir = buf_make(0);
 
+  mode_t current_mask;
+  if (cs_subset_bool(NeoMutt->sub, "attachments_respect_umask"))
```
roccoblues, line 566: `mutt_file_open` is used for everything - not just attachments. So we either need to make the config option "global" or we need to make this change only for attachments.

roccoblues (commented):
I left some thoughts already even though this is still a draft.

test/color/color_dump.c
```diff
@@ -34,6 +34,7 @@
 
 static struct ConfigDef Vars[] = {
   // clang-format off
+  { "attachments_respect_umask", DT_BOOL, false, 0, NULL, },
```
doronbehar, line 37: Wow, I never would have guessed that, thanks!

flatcap:
Thanks for the PR @doronbehar!

The file perms are certainly an annoyance, but we'll have to be careful fixing the problem.
That `umask(077)` in `main()` goes back to the initial revision of Mutt, 26 years ago.

I've only had a quick look through the code, yours and Neomutt's.
When the user saves emails/files/etc, they should probably honour the shell's umask.
However, **every** temporary file we create (and there are a lot), should be `rw-------`

There are 167 calls to `mutt_file_fopen()` (though 71 are read-only).
Of the 96 read-write opens, about 30 explicitly mention a "tmpfile".

We also need to consider ~80 `buf_mktemp()` and 40 `mutt_file_mkstemp()` calls.

I'll have another dig around tomorrow.

---

If we come up with a good plan...
How much free time can you spare?  :-)

test/color/color_dump.c
flatcap: hehe, it's a temporary hack.

You added a config lookup in `mutt_file_open()`.
 
```c
if (cs_subset_bool(NeoMutt->sub, "attachments_respect_umask"))
```

The `mutt/` library shouldn't have any dependencies on other libraries.
We can fix that later.

doronbehar: > hehe, it's a temporary hack.
> 
> You added a config lookup in `mutt_file_open()`.
> 
> ```c
> if (cs_subset_bool(NeoMutt->sub, "attachments_respect_umask"))
> ```
> 
> The `mutt/` library shouldn't have any dependencies on other libraries. We can fix that later.

I don't see the full picture yet, so I'll let it go :).

doronbehar:
> If we come up with a good plan...
> How much free time can you spare? :-)

Haha I think I have enough spare time :).

> here are 167 calls to `mutt_file_fopen()` (though 71 are read-only). Of the 96 read-write opens, about 30 explicitly mention a "tmpfile".
> 
> We also need to consider ~80 `buf_mktemp()` and 40 `mutt_file_mkstemp()` calls.

What do you think about adding adding a boolean parameter to `mutt_file_fopen` that will tell it to whether to respect the config `attachments_respect_umask`, and then use that parameter only for attachment savings.

If via a simple `git grep mutt_file_fopen(` I can reach all of the usages of this function, and you could point me to where `mutt_file_fopen` is used for attachments, it shouldn't be too hard to implement this.

roccoblues:
> What do you think about adding adding a boolean parameter to `mutt_file_fopen` that will tell it to whether to respect the config `attachments_respect_umask`, and then use that parameter only for attachment savings.

Wouldn't it be better to have an option `respect_umask` that works for everything that isn't automatically created by NeoMutt? I think it makes sense also for saved emails.

If we make the boolean parameter off by default, we can enable it where we're sure it's safe without much risk.

doronbehar:
I agree that perhaps not even adding a config option for this would be good. As a start, we could use the original user's umask only for attachments, and add more locations where umask will be considered in follow up commits.

flatcap:
@doronbehar 
Should you have time to waste :-)
you can also find us on IRC: `#neomutt` on `irc.libera.chat` ([web client](https://kiwiirc.com/nextclient/irc.libera.chat/#neomutt))

flatcap:
Edited highlights from IRC today:

```
roccoblues | how do you feel about not adding a config option for #4149
   flatcap | backwards-compatibility is good, but programs behaving as you'd expect is possibly a better idea
```
```
   flatcap | how many different ways do we save?
   flatcap | temp files, attachments, an email / multiple emails (in mbox or maildir format), history file, alias file
roccoblues | we could start with just the attachment?
   flatcap | yeah
```
```
   flatcap | I'm just trying to work out if we need a config option
   flatcap | or if we can be smart and break the old behaviour without upsetting too many people
   flatcap | whichever we announce will annoy somebody...
roccoblues | imho everything the user saves manually (attachment, emails, ...) should honor the umask
   flatcap | yes, that sounds correct
```
```
   flatcap | that'll have to extend to directories too, for maildir
   flatcap | iirc, there's some logic in maildir that considers the parent dir's perms
```
```
roccoblues | but why maildir?
roccoblues | copy or move message should not change
```
```
   flatcap | are there any times when we automatically create a new mailbox?
   flatcap | unfortunately, move (<save-message>) isn't smart so it creates a new copy
roccoblues | I think we shouldn't touch messages
roccoblues | privacy wise that could be very sensitive
   flatcap | yeah
```
```
   flatcap | where is the copying/saving actually happening?
   flatcap | who's driving it?
   flatcap | who knows whether the files/dirs are local?
```
```
   flatcap | if the user copies some emails to a new mailbox, what are the perms of that new mailbox?
   flatcap | for maildir: if there's a parent maildir we _could_ copy its perms
   flatcap | if not?  what?
   flatcap | for mbox: there's no parent, so we've no guide
   flatcap | if I copy a "safe" (rw-------) email into a permissive (rwxr-xr-x) maildir, what are its perms?
```
```
roccoblues | oh I just realized that it probably isn't as easy as I thought
roccoblues | because we don't want to remove the 'umask' from main to be safe
roccoblues | but how do we know it then in the one case where we want to honor it
roccoblues | umask() returns the old mask so we could save it somewhere?
   flatcap | yes
roccoblues | ok, and then we use it only for attachments for now
   flatcap | yeah, I think so
```
```
   flatcap | and keep thinking if we need: `set attach_umask = 022` (or $save_umask), `set mailbox_umask = 077`
roccoblues | at least we wouldn't annoy anyone with a new config option
   flatcap | yeah
roccoblues | and `set attach_umask = 022` is nice and explicit
         * | roccoblues likes it
```

flatcap:
@doronbehar 
After a bit of discussion, @roccoblues and I have concluded that saving things is tricky :-)

Please can you:

- **Restore** the `umask(077)` to `main()`
  This is for safety -- everything else will be secure. 
  You can store the return value in the `NeoMutt` global object.

- **Change** the attachment-save to honour the user's umask
  Attachment Dialog, `<save-entry>` -- Reached from the Index or Pager
  Compose Dialog, `<copy-file>` -- When creating a new email

- **Remove** the config option `$attachments_respect_umask`
  Let's keep things simple to start with.
  We might add `$attach_umask` in the future.

Thanks :heart:

flatcap:
As you'd already dived in, I forgot to say:  Ask lots of questions (here, or on irc).

doronbehar:
Thank you for the support and for laying out a plan, it sounds good to me! I will keep you updated when I reach this .

doronbehar:
Hey all, so I found time to work on this a bit :) I managed to not change all of the usages of `mutt_file_fopen` by declaring an intermediate function. In the 2nd commit I iterated the functions `mutt_file_fopen` usages I found that are related to attachments. Tell me if you think I missed something.

Unfortunately, it won't build for me - I get no warning during the build related to the new stuff I added, and I get a many `ld` `undefined reference to` errors at the end of the build. I'd thank you for your review and/or your help.

flatcap:
- Rebased over `main`
- Added prototype for `mutt_file_fopen_full_masked()`
- clang-format

It seems to build cleanly.
I'll have a play with it tomorrow.

mutt/file.c
```diff
@@ -700,6 +703,28 @@ FILE *mutt_file_fopen_full(const char *path, const char *mode, const char *file,
...
+
+/**
+ * mutt_file_fopen_full - A wrapper around mutt_file_fopen_full_with_mask with
+ * the mask parameter set to the user's default mode inferred from the umask
+ */
+FILE *mutt_file_fopen_full_masked(const char *path, const char *mode,
+                                  const char *file, int line, const char *func)
+{
+  struct ConfigSet *cs = cs_new(16);
+  NeoMutt = neomutt_new(cs);
```
flatcap, line 724: Hmm... That's a surprise!
There can be only one NeoMutt and this isn't it :-)

The `mutt/` library is intended to be self-contained (dependency free).
Any config or global variables need to be passed in as parameters.

core/neomutt.h
```diff
@@ -45,6 +45,7 @@ struct NeoMutt
   struct ConfigSubset *sub;      ///< Inherited config items
   struct AccountList accounts;   ///< List of all Accounts
   locale_t time_c_locale;        ///< Current locale but LC_TIME=C
+  mode_t userDefaultMode;        ///< User's default file writing mode (inferred from umask)
```
flatcap, line 48: Mind if we rename this to a C-style: `user_default_mode`

mutt/file.h
```diff
@@ -144,6 +145,7 @@ void        buf_quote_filename(struct Buffer *buf, const char *filename, bool ad
 void        buf_file_expand_fmt_quote(struct Buffer *dest, const char *fmt, const char *src);
 
 #define mutt_file_fopen(PATH, MODE) mutt_file_fopen_full(PATH, MODE, __FILE__, __LINE__, __func__)
+#define mutt_file_fopen_masked(PATH, MODE) mutt_file_fopen_full_masked(PATH, MODE, __FILE__, __LINE__, __func__)
```
flatcap, line 148: If we really need to get the mode from `NeoMutt`,
then we might need to move the wrapper to `core/`

core/neomutt.h
doronbehar: > Mind if we rename this to a C-style: `user_default_mode`

Done!

doronbehar:
I'm also still puzzled how come on my PC there is no compilation error in `file.c`, but there is a link error, that doesn't appear in CI here.

flatcap:
> there is a link error, that doesn't appear in CI here.

Hmm... what errors are you getting?
It's always best to start from a completely clean repo.
This'll do it:
```
git clean -xdf
```

mutt/file.c
flatcap: > I didn't understand the meaning of `fuzz`

Ah, that's a standalone [fuzzing](https://en.wikipedia.org/wiki/Fuzzing) tester.
It creates a fake `NeoMutt` to keep the linked libraries happy.

> I don't understand how to access `NeoMutt`

The global `NeoMutt` object is declared in `core/neomutt.h`, so `#include "core/lib.h"` will get you access.

doronbehar: > The global `NeoMutt` object is declared in `core/neomutt.h`, so `#include "core/lib.h"` will get you access.

It is already included in `file.c`...

doronbehar:
`git clean -xdf` helped! Thanks. It fully compiles and all tests pass.

I also had to fix the behavior of what to save in `NeoMutt`, and what to do in `mutt_file_fopen_masked_full`. So overall, everything works as expected and desired :tada:.

mutt/file.c
doronbehar: It works :)

alejandro-colomar:
> Thanks for the PR @doronbehar!
> 
> The file perms are certainly an annoyance, but we'll have to be careful fixing the problem. That `umask(077)` in `main()` goes back to the initial revision of Mutt, 26 years ago.
> 
> I've only had a quick look through the code, yours and Neomutt's. When the user saves emails/files/etc, they should probably honour the shell's umask. However, **every** temporary file we create (and there are a lot), should be `rw-------`
> 
> There are 167 calls to `mutt_file_fopen()` (though 71 are read-only). Of the 96 read-write opens, about 30 explicitly mention a "tmpfile".
> 
> We also need to consider ~80 `buf_mktemp()` and 40 `mutt_file_mkstemp()` calls.

Setting a umask doesn't mean you should allow all permissions that it allows.  It means that you should not give permissions that it doesn't allow.

```
alx@debian:~/tmp/mutt$ ls
alx@debian:~/tmp/mutt$ umask
0002
alx@debian:~/tmp/mutt$ mktemp -p .
./tmp.pCD272Jg9q
alx@debian:~/tmp/mutt$ ls -l
total 0
-rw------- 1 alx alx 0 Apr 18 11:23 tmp.pCD272Jg9q
alx@debian:~/tmp/mutt$ umask 477
alx@debian:~/tmp/mutt$ mktemp -p .
./tmp.WVw6zC1ZKM
alx@debian:~/tmp/mutt$ ls -l
total 0
--w------- 1 alx alx 0 Apr 18 11:23 tmp.WVw6zC1ZKM
-rw------- 1 alx alx 0 Apr 18 11:23 tmp.pCD272Jg9q
alx@debian:~/tmp/mutt$ umask 0277
alx@debian:~/tmp/mutt$ mktemp -p .
./tmp.UMFadP0Tew
alx@debian:~/tmp/mutt$ ls -l
total 0
-r-------- 1 alx alx 0 Apr 18 11:24 tmp.UMFadP0Tew
--w------- 1 alx alx 0 Apr 18 11:23 tmp.WVw6zC1ZKM
-rw------- 1 alx alx 0 Apr 18 11:23 tmp.pCD272Jg9q
alx@debian:~/tmp/mutt$ umask 0777
alx@debian:~/tmp/mutt$ mktemp -p .
./tmp.1yPgqymIuD
alx@debian:~/tmp/mutt$ ls -l
total 0
---------- 1 alx alx 0 Apr 18 11:24 tmp.1yPgqymIuD
-r-------- 1 alx alx 0 Apr 18 11:24 tmp.UMFadP0Tew
--w------- 1 alx alx 0 Apr 18 11:23 tmp.WVw6zC1ZKM
-rw------- 1 alx alx 0 Apr 18 11:23 tmp.pCD272Jg9q
```

mktemp(3) probably already uses 0600 in _open_(3) (although you should double-check).

doronbehar:
@alejandro-colomar Thanks for your comment. Your observation fits my understanding of umask too - you can perform the same tests with `touch` instead of `mktemp`. I also added [a testing script in my `open-test` branch](https://github.com/doronbehar/neomutt/blob/open-test/open-test.c).

The PR is ready for you to test these different `umask`s with Neomutt as well.

alejandro-colomar:
> ```
>    flatcap | and keep thinking if we need: `set attach_umask = 022` (or $save_umask), `set mailbox_umask = 077`
> roccoblues | at least we wouldn't annoy anyone with a new config option
>    flatcap | yeah
> roccoblues | and `set attach_umask = 022` is nice and explicit
>          * | roccoblues likes it
> ```

I'm yet neutral to having specific umasks, but I think all should be masked by the actual umask(2).  If I specify a umask of 0177, which is more restrictive than anything neomutt(1) has, the program should respect it, and do `umask | $attach_umask`.

flatcap:
OK, here's the layout you need.

### `mutt/file.h`

Macro: `mutt_file_fopen()` that calls `mutt_file_fopen_full(..., 0600)`

We need a macro because we want to preserve the `__FILE__`, etc, of the **caller**.

Hard-coding the perms is ok because we want the same defaults as old NeoMutt.
We never create any executable files, so the `x` flag isn't relevant.
(We could define those perms as a constant)

### `mutt/file.c`

Function: `mutt_file_fopen_full()` needs a perms parameter.

### `core/tmp.h`

Macro:  `mutt_file_fopen_masked()` that calls `mutt_file_fopen_masked_full()`

A macro again to preserve `__FILE__`, etc

### `core/tmp.c`

Function: `mutt_file_fopen_masked_full()`

This has access to `NeoMutt`.
If `NeoMutt` isn't set, default to a umask of `077`
Merge the perms `0666` with the umask.
Then call `mutt_file_fopen_full()`

flatcap:
## Tests

Things to check:

- Temp files when viewing an Email in the Pager
- Temp files when composing an Email
- Log file
- History file
- Saved Certificates
- Cached headers
- Cached messages (body)

doronbehar:
Thanks @flatcap for summarizing the discussion from IRC, I'm not able to join that. I will also perform the checks you requested once I'm sure I got the layout correct.

> OK, here's the layout you need.
> ### `mutt/file.h`
> 
> Macro: `mutt_file_fopen()` that calls `mutt_file_fopen_full(..., 0600)`
> 
> We need a macro because we want to preserve the `__FILE__`, etc, of the **caller**.
> 
> Hard-coding the perms is ok because we want the same defaults as old NeoMutt. We never create any executable files, so the `x` flag isn't relevant. (We could define those perms as a constant)
> ### `mutt/file.c`
> 
> Function: `mutt_file_fopen_full()` needs a perms parameter.

Hmm OK it took me a few minutes to figure out how different is it from the current layout - basically only the `mutt_file_fopen_with_perms` function is not needed - the amount of `mutt_file_fopen*` functions can be reduced from 3 to 2. Now I implemented this part of the layout, hopefully aligning with your request, and performed the same tests as before and it works as desired.

> ### `core/tmp.h`
> 
> Macro: `mutt_file_fopen_masked()` that calls `mutt_file_fopen_masked_full()`
> 
> A macro again to preserve `__FILE__`, etc
> ### `core/tmp.c`
> 
> Function: `mutt_file_fopen_masked_full()`
> 
> This has access to `NeoMutt`. If `NeoMutt` isn't set, default to a umask of `077` Merge the perms `0666` with the umask. Then call `mutt_file_fopen_full()`

I don't understand, you want the `mutt_file_fopen_masked` macro to be defined in `core/tmp.h`? What is the problem in putting it in `mutt/file.h`? In `mutt/file.h` it makes more sense to me, as this function will be used exactly for files which are not `tmp` files, and it is near the `mutt_file_fopen` macro.

flatcap:
NeoMutt's built up from libraries.
They're roughly divided up by their dependencies.
`libgui` has the most dependencies, relying on ncurses, etc.

The files/functions in `mutt/` are the simplest -- plain C.
This means we can build and test `libmutt` independently.

`NeoMutt` in `libcore` is a high-level object with a lot of dependencies.

This (out-of-date) diagram shows some of the dependencies in NeoMutt: https://github.com/neomutt/gfx/tree/main/arch#components

---

On startup you store the user's umask in `NeoMutt`.
If you want the `fopen_masked()` function to use this umask, you need access to `NeoMutt`.

When the Attachment Dialog wants to save a file, there are two possibilities:

- Explicit: `fopen_masked(file, NeoMutt->umask)`
  Every caller needs to specify the umask.

- Helper: `fopen_masked(file)`
  A helper function does the `NeoMutt` lookup for us

If we want the tidier-looking version, then that helper function cannot be in `mutt/`
because it would need to first `#include "core/lib.h"`, then link against `libcore`.

doronbehar:
Hmm OK thanks for the excellent explanation :). Yea so I was looking at the header files listed in `core/lib.h`, and only `core/tmp.{c,h}` defines stuff related to files. I still find that a bit weird to to put it there, perhaps `core/neomutt.{c,h}` would be more appropriate? Because in the end the `_masked` function and macro relate deeply to the `NeoMutt` object.

doronbehar:
> > core/neomutt.{c,h}
> 
> Yeah, OK

OK great:) Done.

Don't have time for the other tests now, noted myself to do them.

core/neomutt.c
```diff
@@ -209,3 +209,22 @@
...
+ * mask parameter set to the user's default mode_t permissions, inferred from
+ * the umask
+ */
+FILE *mutt_file_fopen_masked_full(const char *path, const char *mode, const char *file,
+                                  int line, const char *func)
+{
+  // We need to change the current umask to the original one we retrieved from
+  // main.c, _and_ also to use the original umask to calculate the permissions
+  // of the file we create.
+  umask(NeoMutt->user_default_umask);
```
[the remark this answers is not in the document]
doronbehar: Tried to fix this github security warning on `umask` in the last force push.

flatcap:
> **Saved Certificates**:
> How to test this?

Passed.
Remove `~/.mutt_certificates` and connect to an IMAP server using TLS.

> **Cached messages (body)**:

Passed.
The message cache (body), keeps a local copy of any emails you read.
It's used by remote mailboxes: IMAP, POP, etc, but not Mailbox, mbox.

core/neomutt.c
```diff
@@ -209,3 +210,22 @@ size_t neomutt_mailboxlist_get_all(struct MailboxList *head, struct NeoMutt *n,
...
+
+/**
+ * mutt_file_fopen_masked_full - A wrapper around mutt_file_fopen_full with the
+ * mask parameter set to the user's default mode_t permissions, inferred from
+ * the umask
+ */
+FILE *mutt_file_fopen_masked_full(const char *path, const char *mode, const char *file,
+                                  int line, const char *func)
+{
+  // We need to change the current umask to the original one we retrieved from
```
flatcap, line 222: **Do** we?
If we're applying the umask to the perms, below, then the file will have the right permissions.

If we don't, we can drop both `umask()` calls.

core/neomutt.c
```diff
@@ -209,3 +210,22 @@ size_t neomutt_mailboxlist_get_all(struct MailboxList *head, struct NeoMutt *n,
...
+  // main.c, _and_ also to use the original umask to calculate the permissions
+  // of the file we create.
+  umask(NeoMutt->user_default_umask);
+  // This 0666 & ~default_umask argument implements the same behavior every unix
+  // utility operates with respect to the shell's umask
+  return mutt_file_fopen_full(path, mode, 0666 & ~NeoMutt->user_default_umask, file, line, func);
+  // Change it quickly back to our default, safe umask of 0700
+  umask(0700);
```
flatcap, line 230: If we **do** need them (and I'm missing the point),
then it's important for this `umask()` **NOT** to be after the `return` statement! :-)

core/neomutt.c
doronbehar: > **Do** we?

I was a bit surprised by this too :) Basically, the `mode_t` that `open` gets, is unmasked unconditionally by the current umask. This is also observed in this [QA](https://stackoverflow.com/questions/9380671/how-to-neglect-the-umask-so-as-to-create-the-file-with-given-permission).

You are correct though, that we don't have to implement the `0666 & ~NeoMutt->user_default_umask` binary operation ourselves, as `open` does it by itself as well, since we set `umask` before calling `open`.

flatcap (approved):

doronbehar:
Thanks!

PR #4155, opened 2024-01-02
fix parsing of $REPLYTO

flatcap:
The environment variable REPLYTO should be parsed into a user header of "Reply-To:"

---

Some past refactoring broke this.
The fix is to reset the `Buffer` pointer so that `parse_extract_token()` reads from the beginning of the string.

---

To test:
```sh
REPLYTO=john.doe@example.com neomutt
```

then compose an email.
In Compose, you'll see: `Reply-To: john.doe@example.com`


roccoblues (approved):
Fix confirmed.

But I have to say this is suprising behavior, at least for me. I would have expected the underlying code to start reading the buffer string from the beginning.

PR #4162, opened 2024-01-18
remove mutt_str_dequote_comment()

roccoblues:
It's only used in `buf_dequote_comment()`. Moved the code there.


roccoblues:
> Rather than remove the test, please can you move/rename it into `test/buffer`

We already `test/buffer/buf_dequote_comment.c`, the tests were duplicate.

flatcap (approved):

PR #4165, opened 2024-01-18
bufferize mutt_str_inline_replace() as buf_inline_replace()

roccoblues:
The API of the method slightly changed:

* The previous implementation always replaced at the beginning of the string. The new version takes a position argument to specify the starting location of the string to replace.

* The previous implementation never increased the underlying buffer. The new, buffer backed, version increases it if necessary. The diff can be seen in the test cases:

```
  -    { "XXXXbanana", 4, "OOOOOOOO",  "OOOOOOOObanan"   },
  -    { "XXXXbanana", 4, "OOOOOOOOO", "OOOOOOOOObana"   },
  +    { "XXXXbanana", 4, "OOOOOOOO",  "OOOOOOOObanana"  },
  +    { "XXXXbanana", 4, "OOOOOOOOO", "OOOOOOOOObanana" },
```

  I think this is the better behavior based on the existing use cases.


flatcap (approved):
very nice, thanks!

Issue #4033, opened 2023-09-10
Leak if I try a notmuch search on a mail folder that doesn't exist


scottkosty:
## Expected Behaviour

I accidentally opened a folder that doesn't exist. I made a notmuch query and then exited. 

## Actual Behaviour

There is a memory leak reported by ASAN.

## Steps to Reproduce

Set 'folder' to a non-existent directory and start NeoMutt. Then do a notmuch query. Then exit NeoMutt.

I get the following ASAN report after exiting NeoMutt:

<details>

```
=================================================================
==1807926==ERROR: LeakSanitizer: detected memory leaks

Direct leak of 240 byte(s) in 1 object(s) allocated from:
    #0 0x7f7c910de9a7 in __interceptor_calloc ../../../../src/libsanitizer/asan/asan_malloc_linux.cpp:77
    #1 0x561f8187d4e6 in mutt_mem_calloc mutt/memory.c:55
    #2 0x561f81810203 in mailbox_new core/mailbox.c:71
    #3 0x561f815b3a73 in mx_path_resolve /home/scott/IndependentInstalls/neomutt/mx.c:1709
    #4 0x561f815ce399 in change_folder_notmuch index/dlg_index.c:745
    #5 0x561f815e6470 in op_main_vfolder_from_query index/functions.c:2748
    #6 0x561f815e71b6 in index_function_dispatcher index/functions.c:3061
    #7 0x561f815d2971 in dlg_index index/dlg_index.c:1340
    #8 0x561f8158581d in main /home/scott/IndependentInstalls/neomutt/main.c:1392
    #9 0x7f7c90623a8f in __libc_start_call_main ../sysdeps/nptl/libc_start_call_main.h:58

Indirect leak of 256 byte(s) in 1 object(s) allocated from:
    #0 0x7f7c910dddf0 in __interceptor_realloc ../../../../src/libsanitizer/asan/asan_malloc_linux.cpp:85
    #1 0x561f8187d785 in mutt_mem_realloc mutt/memory.c:131
    #2 0x561f8185af30 in buf_alloc mutt/buffer.c:375
    #3 0x561f81859e62 in buf_addstr_n mutt/buffer.c:122
    #4 0x561f8185a796 in buf_addstr mutt/buffer.c:242
    #5 0x561f8185b0cc in buf_strcpy mutt/buffer.c:410
    #6 0x561f815b3a8a in mx_path_resolve /home/scott/IndependentInstalls/neomutt/mx.c:1710
[...]

PR #4166, opened 2024-01-18
Fix two memory leaks in notmuch support

Pierre-Colin:
This addresses issue #4033 as well as another leak when opening a notmuch database.

* **What does this PR do?**

It fixes two memory leaks that have to do with notmuch support.

The first has to do with `change_folder_mailbox` returning a nonnull pointer to a mailbox on which the call to `change_folder_mailbox` failed. Since `change_folder_mailbox` doesn’t return any error code, the checking is done by checking whether it stored a mailbox view into the index shared data.

The second has to do with the case of `config_file` being a null pointer instead of an empty string in `nm_db_do_open`. This lead to `msg` not being freed.

With this patch, Valgrind no longer reports any definitely lost block in @scottkosty’s example.

#4033

### DIFF
--- a/index/dlg_index.c
+++ b/index/dlg_index.c
@@ -737,6 +737,8 @@ struct Mailbox *change_folder_notmuch(struct Menu *menu, char *buf, int buflen, 
 
   struct Mailbox *m_query = mx_path_resolve(buf);
   change_folder_mailbox(menu, m_query, oldcount, shared, read_only);
+  if (!shared->mailbox_view)
+    mailbox_free(&m_query);
   return m_query;
 }
 #endif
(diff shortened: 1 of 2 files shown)


flatcap (commented):
Brilliant!
Thanks for the fixes.

The `notmuch/db.c` fix is good.

The `index/dlg_index.c` fix, I need to have a think about.
(the mailbox code here is complicated)
I _did_ have a look at this bug before, but couldn't find the right spot fix it.
Clearly I didn't think hard enough :-)

Thanks!

flatcap (approved):
Brilliant!  Thanks!

PR #4167, opened 2024-01-19
bufferize mutt_strn_rfind() as buf_rfind()

roccoblues:
Bufferize `mutt_strn_rfind()`. The callers already have a `struct Buffer`.

I've also simplified and fixed the Notmuch code. `mutt_nm_query_complete()` is only called from one place and the "position" parameter was always set to the buffer length. I've removed it. Also it used to pass the Buffer pointer instead of `buf->data` for the string to search into `mutt_strn_rfind()`.


flatcap (approved):
I made one small tweak
```diff
-  if (!buf || buf_len(buf) == 0 || !str)
+  if (buf_is_empty(buf) || !str)
```

Thanks! Keep 'em coming.

PR #4173, opened 2024-01-26
turn MoreArgs/MoreArgsf macros into functions

roccoblues:
### TODO:

- [ ] naming is hard, what does "args" even mean? And ending the name with "f" usually means format - can we come up with something better?
- [ ] should the functions be inlined?
- [ ] more tests for flags function

### DIFF
--- a/alias/commands.c
+++ b/alias/commands.c
@@ -139,7 +139,7 @@ enum CommandResult parse_alias(struct Buffer *buf, struct Buffer *s,
   struct GroupList gl = STAILQ_HEAD_INITIALIZER(gl);
   enum NotifyAlias event;
 
-  if (!MoreArgs(s))
+  if (!has_more_args(s))
   {
     buf_strcpy(err, _("alias: no address"));
     return MUTT_CMD_WARNING;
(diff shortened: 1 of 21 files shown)


alias/commands.c
```diff
@@ -220,7 +220,7 @@ enum CommandResult parse_alias(struct Buffer *buf, struct Buffer *s,
     }
   }
   mutt_grouplist_destroy(&gl);
-  if (!MoreArgs(s) && (s->dptr[0] == '#'))
+  if (!has_more_args(s) && (s->dptr[0] == '#'))
```
roccoblues, line 223: Can we use the flags function and get rid of the extra check?

flatcap:
> naming is hard, what does "args" even mean?

Chunks?  Symbols?

The function is splitting a string into chunks whitespace-separated blocks of text.

It's done in steps because we need to know the command to know how to parse the rest of the string.
The grammar of the commands is quite complicated.
It gets _more_ complicated when we consider multiple; commands; chained; together.

> And ending the name with "f" usually means format - can we come up with something better?

The 'f' version was added to allow parsing of colours of the form `#123456`
(otherwise `#` is a comment)

> should the functions be inlined?

It may not matter too much.
The command processing isn't time-critical.

> more tests for flags function

yeah, let's go overboard :-)

With reliable tests in place, it _might_ be possible to consider using regexes (Prex) to work on entire commands.

PR #4176, opened 2024-01-31
add parse_extract_color_name() to extract color name tokens

roccoblues:
Color names are simple to parse. Adding a dedicated function allows us to get rid of the `MoreArgsF()` macro and maybe later simplify `parse_extract_token()`.

Related: https://github.com/neomutt/neomutt/pull/4173, https://github.com/neomutt/neomutt/discussions/4174

PR #4183, opened 2024-02-06
refactor token parsing

roccoblues:
Add a dedicated function `parse_next_word()` that handles a simple case of `parse_next_token()` - extracting the next set of consecutive non-whitespace characters from a string.

This allows us to get rid of the `MoreArgsF()` macro and maybe later simplify `parse_extract_token()`.

I've also move the `MoreArgs()` macro into a function `has_more_tokens()` and added tests for it. I renamed `args` to `tokens` to match the extract function name.

Related discussion: https://github.com/neomutt/neomutt/discussions/4174

Supersedes: https://github.com/neomutt/neomutt/pull/4173, https://github.com/neomutt/neomutt/pull/4176

### DIFF
--- a/alias/commands.c
+++ b/alias/commands.c
@@ -139,7 +139,7 @@ enum CommandResult parse_alias(struct Buffer *buf, struct Buffer *s,
   struct GroupList gl = STAILQ_HEAD_INITIALIZER(gl);
   enum NotifyAlias event;
 
-  if (!MoreArgs(s))
+  if (!has_more_tokens(s))
   {
     buf_strcpy(err, _("alias: no address"));
     return MUTT_CMD_WARNING;
@@ -221,7 +221,7 @@ enum CommandResult parse_alias(struct Buffer *buf, struct Buffer *s,
     }
   }
   mutt_grouplist_destroy(&gl);
-  if (!MoreArgs(s) && (s->dptr[0] == '#'))
+  if (!has_more_tokens(s) && (s->dptr[0] == '#'))
   {
     s->dptr++; // skip over the "# "
     if (*s->dptr == ' ')
@@ -277,6 +277,6 @@ enum CommandResult parse_unalias(struct Buffer *buf, struct Buffer *s,
       alias_free(&np);
       break;
     }
-  } while (MoreArgs(s));
+  } while (has_more_tokens(s));
   return MUTT_CMD_SUCCESS;
 }
--- a/alias/commands.c
+++ b/alias/commands.c
@@ -221,7 +221,8 @@ enum CommandResult parse_alias(struct Buffer *buf, struct Buffer *s,
     }
   }
   mutt_grouplist_destroy(&gl);
-  if (!has_more_tokens(s) && (s->dptr[0] == '#'))
+
+  if (s->dptr[0] == '#')
   {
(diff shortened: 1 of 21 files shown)


roccoblues:
`parse_next_word()` can be used to parse more commands. I'll go through the list later but first wanted to get your opinion on it.

flatcap:
That looks good.
Perhaps `parse_next_word()` could be a bool -- it either extracts something, or it doesn't.

> `parse_next_word()` can be used to parse more commands

Hmm... possibly.
We need `pnw()` because the RGB colours contain `#`.
Everywhere else, expects `#` to be the start of a comment.

<table>
<tr>
 <td><code>echo hello</code>
 <td>"hello"
<tr>
 <td><code>echo hello # world</code>
 <td>"hello"
<tr>
 <td><code>echo hello#world</code>
 <td>"hello"
</table>

 Perhaps if we had a single flag to allow _leading_ `#`s.

roccoblues:
> Perhaps `parse_next_word()` could be a bool -- it either extracts something, or it doesn't.

good point, I think I even had that initially but then changed it to match `parse_next_token()`.

> > `parse_next_word()` can be used to parse more commands
> 
> Hmm... possibly. We need `pnw()` because the RGB colours contain `#`. Everywhere else, expects `#` to be the start of a comment.
> `echo hello` 	"hello"
> `echo hello # world` 	"hello"
> `echo hello#world` 	"hello"
> 
> Perhaps if we had a single flag to allow _leading_ `#`s.

Yep, let me try that.

roccoblues:
> > Perhaps if we had a single flag to allow _leading_ `#`s.
> 
> Yep, let me try that.

I tried it but didnt' really like it. Then we have to add a `false` to all cases except one.

I think we should just accept the fact that parsing color names is a special case and add a dedicated function for it. Which is what I did now.

I've also changed `parse_echo()` and `parse_cd()` to use `parse_next_word()` now. wdyt?

flatcap:
> I tried it but didnt' really like it. Then we have to add a `false` to all cases except one.

Fair enough

> I've also changed `parse_echo()` and `parse_cd()` to use `parse_next_word()` now. wdyt?

Unfortunately, they don't work if there's any whitespace, e.g.
```
cd jim\ bob
echo "jim bob"
```

roccoblues:
> Unfortunately, they don't work if there's any whitespace, e.g.
> 
> ```
> cd jim\ bob
> echo "jim bob"
> ```

🙈 Yeah, also `echo "jim \" # bob"` needs to work. I'll add a bunch of failing tests now.

roccoblues:
Ok, I now realize that this is much more complicated than I initialy thought. I don't see a clear path forward at the moment. 

My suggestion would be to start with https://github.com/neomutt/neomutt/commit/91ba8ccc5f540785383311c5be5990e63d88f2af that at least gets rid of the complicated `MorgArgsF` macro and is simple and well tested.

roccoblues:
Closed. We went with a first step in https://github.com/neomutt/neomutt/commit/91ba8ccc5f540785383311c5be5990e63d88f2af.

The bigger change will be tracked by the discussion: https://github.com/neomutt/neomutt/discussions/4174

PR #4185, opened 2024-02-11
Fix memory leak in mx.c

ashamedbit:
* **What does this PR do?**
* The PR fixes a memory leak involving `mx_path_resolve`. It addresses some leaks in `trash_append` where `m_trash` is allocated by `mx_path_resolve` but not freed on a given path when appending to trash fails when syncing or closing a mailbox.
In fact there is a similar error path above where `m_trash` is freed and returns -1. Additionally I was able to verify that all tests succeeded with the patch

### DIFF
--- a/mx.c
+++ b/mx.c
@@ -566,6 +566,7 @@ static int trash_append(struct Mailbox *m)
         // L10N: Displayed if appending to $trash fails when syncing or closing a mailbox
         mutt_error(_("Unable to append to trash folder"));
         m_trash->append = old_append;
+        mailbox_free(&m_trash);
         return -1;
       }
     }


mx.c
```diff
@@ -566,6 +566,7 @@ static int trash_append(struct Mailbox *m)
         // L10N: Displayed if appending to $trash fails when syncing or closing a mailbox
         mutt_error(_("Unable to append to trash folder"));
         m_trash->append = old_append;
```
roccoblues, line 568: this looks suspicious, why do we set `m_trash->append` before freeing the mailbox?

ashamedbit: Hi! Thank you for reviewing the PR :). I would like to point out that same happens on line 575-576 on the non error path.

flatcap: We set `Mailbox.append` because we want to change the behaviour for writing,
but we restore it afterwards, because we may not be the only user of _that_ Mailbox.

`Mailbox.append` is an ugly hack that I haven't figured out how to kill yet.

roccoblues: > because we may not be the only user of that Mailbox.

oh I see, `mailbox_free()` doesn't unconditionally free the mailbox.

flatcap (approved):
Good spot, @ashamedbit!
Thanks for the fix.

PR #4189, opened 2024-02-28
Fix off-by-one error in %b with notmuch

tim-clifford:
When the mailbox path is fetched with `nm_email_get_folder_rel_db`, the first character should not be removed, but was. This bug would lead to the first character of %b being chopped off in notmuch folders. This PR fixes that.

### DIFF
--- a/hdrline.c
+++ b/hdrline.c
@@ -486,20 +486,22 @@ static const char *index_format_str(char *buf, size_t buflen, size_t col, int co
     case 'b':
       if (m)
       {
-        p = strrchr(mailbox_path(m), '/');
+        p = NULL;
 #ifdef USE_NOTMUCH
         if (m->type == MUTT_NOTMUCH)
         {
-          char *rel_path = nm_email_get_folder_rel_db(m, e);
-          if (rel_path)
-            p = rel_path;
+          p = nm_email_get_folder_rel_db(m, e);
         }
 #endif
-
-        if (p)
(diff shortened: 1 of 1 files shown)


hdrline.c
```diff
@@ -497,7 +498,7 @@ static const char *index_format_str(char *buf, size_t buflen, size_t col, int co
 #endif
```
gahr, line 498: Good catch - do we perhaps want to rewrite it as follows?
```c
if (m)
{
  p = NULL;
#ifdef USE_NOTMUCH
  if (m->type == MUTT_NOTMUCH)
  {
    p = nm_email_get_folder_rel_db(m, e);
  }
#endif
  if (!p)
  {
    p = strrchr(mailbox_path(m), '/');
    if (p)
    {
      p++;
    }
  }
  if (p)
  {
    mutt_str_copy(buf, p, buflen);
  }
}
```

tim-clifford: Sure, I've updated the PR to match.

gahr (approved):

flatcap:
Hi @tim-clifford, thanks for the fix!
I've just noticed that GitHub doesn't recognise the email address on the commit (`g@c.l`).
You can add the address, here: https://github.com/settings/emails
(Then you'll get a green dot on your activity chart :-)

tim-clifford:
Thanks for the tip!

PR #4193, opened 2024-03-03
Fix oauth2 example on README.md

Yutsuten:
* **What does this PR do?**

Fixes oauth2 example. The script returns an error if the command is copy-pasted as is.


flatcap (approved):
Thanks!

PR #4195, opened 2024-03-04
Oauth2 script improvement: use gpg default recipient

Yutsuten:
I have just finished setting up my gmail account with neomutt, and would like to make two small improvements on the script and README.

### 1. Add default `--encryption-pipe` value

In the `mutt_oauth2.py` script, `ENCRYPTION_PIPE` required a recipient (GPG identity) to be specified as argument.

But gpg have a very convenient option to get a default recipient. Quoting the man page:

> `--default-recipient-self`
>        Use the default key as default recipient if option `--recipient` is not used and don't ask if this is a valid one. The default key is the first one from the secret keyring or the one set with `--default-key`.

I've set it to be used by default, as I think most people only have one secret key on their gpg. If someone have several secret keys, they're probably used specifying the recipient anyway.

I've also updated the README to reflect this.

### 2. Update README `set imap_oauth_refresh_command`

After all the initial setup is done, we use `imap_oauth_refresh_command` to retrieve new tokens as necessary.

I took a look in the script and noticed that specifying `--provider` was unnecessary, so I deleted it. The provider information is saved inside the encrypted file `${imap_user}.tokens`.

My configuration don't have the `--provider` option and it still works as expected.

---

That's it. Sorry if there are any English mistakes, I'm not English native. So please tell me if there is anything wrong!

Thanks!


Yutsuten:
Actually, I noticed that **after** the initial setup (`--authorize`), getting a new access token does not require any option at all.

The file generated by the script contains quite a lot of data in it.

```shell
gpg -d myaccount@gmail.com.tokens | jq .
```

```json
{
  "registration": "google",
  "authflow": "localhostauthcode",
  "email": "myaccount@gmail.com",
  "access_token": "secret",
  "access_token_expiration": "2024-03-25T20:41:04.439294",
  "refresh_token": "secret",
  "client_id": "secret",
  "client_secret": "secret"
}
```

- `--provider` is saved at `registration`
- `--client-id` is saved at `client_id`
- `--client-secret` is saved at `client_secret`

My setting `imap_oauth_refresh_command` now don't include any options, and still can retrieve new access tokens.

I'll update the README to reflect this.

flatcap (approved):
That looks really good, thanks!
I've tweaked the first commit to fix a typo (you didn't cause it).

In other news, [thank you very much](https://neomutt.org/sponsor) :-)

PR #4196, opened 2024-03-04
use FreeBSD 14.0 in Cirrus CI

roccoblues:
This fixes the recent build failures.

### DIFF
--- a/.cirrus.yml
+++ b/.cirrus.yml
@@ -1,5 +1,5 @@
 freebsd_instance:
-  image_family: freebsd-13-2
+  image_family: freebsd-14-0
 
 task:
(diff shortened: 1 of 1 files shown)


flatcap:
Thanks!  I tested the same in [[flatcap/bsd]](https://github.com/neomutt/neomutt/tree/flatcap/bsd).
Oddly, [their docs still list](https://cirrus-ci.org/guide/FreeBSD/) 13.2.
I don't know if it's a deprecation or a temporary misconfiguration.

@gahr Any objections to using FreeBSD 14.0?

gahr (approved):
Let's jump to 14 - thanks!

Issue #4184, opened 2024-02-09
Run multiple commands with single hook


rafaelrc7:
Hello, I'm trying to run multiple commands with a single hook. This sounds simple enough, but I was not able to do it nor find any information about online.

My ideia is to use `shutdown-hook` to first save changes with `<sync-mail>` and then run a shell script that synchronises the maildir itself. I am able to easily do each alone, writing either `shutdown-hook 'exec sync-mail'` or ``shutdown-hook 'echo `sync-script.sh`'``. However, I was not able to run both, first `sync-mail` and then the script.

Is there a way that I could do that? Thanks.

rafaelrc7:
Ok. I think I did manage to run both commands, by either doing:

``shutdown-hook 'exec sync-mailbox `sync-script.sh`'``

or

``shutdown-hook 'exec sync-mailbox ; echo `sync-script.sh`'``

However it still seems to not be working as expected... The sync-script (that is just a call to mbsync) detects no changes, so read emails are not updated, as if `sync-mailbox` had not been executed. However, if I run the hook two times, by closing, opening and closing it works, so the command is working but the mailbox is only being saved after the program exits. 

How could I fix it?

dorsiflexion:
In fact it seems that `shutdown-hook 'exec sync-mailbox'` is doing exactly nothing.

I'm trying to do something very similar. It doesn't even seem to be a race condition, because when doing
```
shutdown-hook 'exec sync-mailbox; echo `sleep 5`; echo `script.sh`'
```
it is still *not* updated when the script runs.

I'd also greatly appreciate some advice on what's wrong here.

rafaelrc7:
> In fact it seems that `shutdown-hook 'exec sync-mailbox'` is doing exactly nothing.
> 
> I'm trying to do something very similar. It doesn't even seem to be a race condition, because when doing
> ```
> shutdown-hook 'exec sync-mailbox; echo `sleep 5`; echo `script.sh`'
> ```
> it is still *not* updated when the script runs.
> 
> Btw. neomutt updates mailboxes when it quits. So if you are closing, opening, closing and it works, that is no indication that the hook actually did something.
> 
> I'd also greatly appreciate some advice on what's wrong here.

Yes, I also tried using `sleep` in my hook to test if it was some race condition, and had the same results.

I ended up rebinding the exit key (`q`) to run the sync commands and then quit. It works. But it still seems to be something wrong with the `shutdown-hook`.

roccoblues:
[...]

PR #4200, opened 2024-03-18
allow shutdown-hook to run functions

roccoblues:
This adds support for function execution (`exec sync-mailbox`) to the `shutdown-hook`.

Notes:
* only `Index` and "Global" functions are supported
* the functions are only exectured after the whole hook is parsed. So in this example:

    ```
    shutdown-hook "echo 'before'; exec sync-mailbox; echo 'after'"
    ```

    the order of execution will be:
    1. before
    2. after
    3. `sync-mailbox`

* ~~`run_events()` is a terrible name and the method is probably in the wrong place~~

Related to: https://github.com/neomutt/neomutt/issues/4184

### DIFF
--- a/hook.c
+++ b/hook.c
@@ -45,6 +45,7 @@
 #include "email/lib.h"
 #include "core/lib.h"
 #include "alias/lib.h"
+#include "gui/lib.h"
 #include "hook.h"
 #include "attach/lib.h"
 #include "compmbox/lib.h"
@@ -56,6 +57,7 @@
 #include "pattern/lib.h"
 #include "commands.h"
 #include "globals.h"
+#include "mutt_logging.h"
 #include "muttlib.h"
 #include "mx.h"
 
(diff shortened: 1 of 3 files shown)


gahr:
> the functions are only exectured after the whole hook is parsed. So in this example:
>       ```
>       shutdown-hook "echo 'before'; exec sync-mailbox; echo 'after'"
>       ```
> the order of execution will be:
>
>    before
>    after
>    sync-mailbox


Not a fan of this "feature" :) Can we do better?

roccoblues:
> > the functions are only exectured after the whole hook is parsed. So in this example:
> > `shutdown-hook "echo 'before'; exec sync-mailbox; echo 'after'"`
> > the order of execution will be:
> > before
> > after
> > sync-mailbox
> 
> Not a fan of this "feature" :) Can we do better?

yeah, same here. But fixing this will require bigger changes. Because the code that parses the hooks and `exec` statements is pretty low level. I'll have another look. If you have any ideas - shoot. ;-)

Issue #4163, opened 2024-01-18
Show information about the encryption of a message


alejandro-colomar:
Thunderbird shows to which recipient PGP (sub)keys a mail is encrypted.  That's quite useful information, but I didn't find any way to find it using mutt(1) or neomutt(1).

I needed that info while testing <https://github.com/neomutt/neomutt/pull/4134>, so I've had to use Thunderbird to open some emails, to check that they're encrypted to whom I expected.

Similarly to how [neo]mutt(1) show the `[-- Begin signature information --]`/`[-- End signature information --]` block, it could also show a `[-- Begin encryption information --]`/`[-- End encryption information --]` block, showing details of the encryption.

flatcap:
Good idea!
If Thunderbird can get the info, then I'm sure we can.

---

I don't know how NeoMutt encrypts, but `gpg` has two options:

> --recipient name
>  -r     Encrypt for user id name.
>
> --hidden-recipient name
> -R     Encrypt for user ID name, but hide the key ID of this user's key.

The hidden option means that key is only known at decryption time,
and then only by trying our private keys until we get a match.

alejandro-colomar:
I'll need to test gpg(1) to see what info is available, and what info can be faked.  Thanks for the pointers!

alejandro-colomar:
I've found this:

<https://www.gnupg.org/documentation/manuals/gpgme.pdf#Decrypt>

```c
$ grepc gpgme_decrypt_result_t /usr/include/
/usr/include/gpgme.h:typedef struct _gpgme_op_decrypt_result *gpgme_decrypt_result_t;
```
```c
$ grepc _gpgme_op_decrypt_result /usr/include/
/usr/include/gpgme.h:struct _gpgme_op_decrypt_result
{
  char *unsupported_algorithm;

  /* Key should not have been used for encryption.  */
  unsigned int wrong_key_usage : 1;

  /* True if the message was encrypted in compliance to the de-vs
   * mode.  */
  unsigned int is_de_vs : 1;

  /* The message claims that the content is a MIME object.  */
  unsigned int is_mime : 1;

  /* The message was made by a legacy algorithm without any integrity
   * protection.  This might be an old but legitimate message. */
  unsigned int legacy_cipher_nomdc : 1;

  /* Internal to GPGME, do not use.  */
  int _unused : 28;

  gpgme_recipient_t recipients;

  /* The original file name of the plaintext message, if
   * available.  */
  char *file_name;

[...]

PR #4221, opened 2024-03-29
ncrypt/crypt_gpgme.c: Show encryption information

alejandro-colomar:
Show a block of information about the encryption of a message, which includes the recipients that can decrypt the message.

This info isn't necessarily complete: the sender could hide some recipients, so that they are able to decrypt the message, but aren't listed.  See gpg(1) options --recipient and --hidden-recipient.  I'm not sure, but I suspect a malicious sender could also add fake recipients, which are listed, but cannot really decrypt the message.

Therefore, this should be taken only as an informative block, and with a pinch of salt.

Closes: <https://github.com/neomutt/neomutt/issues/4163>
Cc: @flatcap
Cc: @dd9jn

* **What does this PR do?**

   Show encryption information about a message.

   Unencrypted mail:

   ```
   Message-ID: <ZgU0xhvp82jUPsy1@debian>

   [-- Begin signature information --]
   Good signature from: Alejandro Colomar <alx@alejandro-colomar.es>
                   aka: Alejandro Colomar <alx@kernel.org>
                   aka: Alejandro Colomar Andres <alx.manpages@gmail.com>
               created: Thu Mar 28 10:13:42 2024
   [-- End signature information --]

   [-- The following data is signed --]
   ```

   Encrypted mail:

   ```
   Message-ID: <ZgA4K7hxzbGPTTGY@debian>

   [-- Begin encryption information --]
   Recipient: (RSA)      C527609C7E1875BD
   Recipient: (RSA)      00C14A7DBBDD521C
   [-- End encryption information --]

   [-- Begin signature information --]
   Good signature from: Alejandro Colomar <alx@alejandro-colomar.es>
                   aka: Alejandro Colomar <alx@kernel.org>
                   aka: Alejandro Colomar Andres <alx.manpages@gmail.com>
               created: Sun Mar 24 15:26:50 2024
   [-- End signature information --]

   [-- The following data is PGP/MIME signed and encrypted --]
   ```

      I need to test a few things still:

      *  Try to fake a recipient.
      *  Hide a recipient.
      *  Test the application handler.

     for that)

      Not yet.

      It builds (and runs) on my PC.

      Small, but yeah.  I may need to expand those a little bit.

      I think so.

   <https://github.com/neomutt/neomutt/issues/4163>

### DIFF
--- a/ncrypt/crypt_gpgme.c
+++ b/ncrypt/crypt_gpgme.c
@@ -1410,6 +1410,37 @@ static void print_smime_keyinfo(const char *msg, gpgme_signature_t sig,
 }
 
 /**
+ * show_one_recipient - Show information about one encryption recipient
+ * @param state State to write to
+ * @param r     Encryption recipient
+ */
+static void show_one_recipient(struct State *state, gpgme_recipient_t r)
+{
+  const char *algo = gpgme_pubkey_algo_name(r->pubkey_algo);
+  if (!algo)
+    algo = "?";
+
+  // L10N: Show the algorithm and key ID of the encryption recipients, e.g
+  //       Recipient: RSA key, ID 1111111111111111
+  state_printf(state, _("Recipient: %s key, ID %s\n"), algo, r->keyid);
+}
+
+/**
+ * show_encryption_info - Show encryption information
+ * @param state  State to write to
+ * @param result Decryption result
+ */
+static void show_encryption_info(struct State *state, gpgme_decrypt_result_t result)
+{
+  state_attach_puts(state, _("[-- Begin encryption information --]\n"));
+
+  for (gpgme_recipient_t r = result->recipients; r; r = r->next)
+    show_one_recipient(state, r);
+
+  state_attach_puts(state, _("[-- End encryption information --]\n\n"));
+}
+
+/**
  * show_one_sig_status - Show information about one signature
  * @param ctx   GPGME handle of a successful verification
  * @param idx   Index
@@ -1704,6 +1735,7 @@ static struct Body *decrypt_part(struct Body *b, struct State *state,
   struct Body *tattach = NULL;
   gpgme_error_t err = GPG_ERR_NO_ERROR;
   gpgme_data_t ciphertext = NULL, plaintext = NULL;
+  gpgme_decrypt_result_t result = NULL;
   bool maybe_signed = false;
   bool anywarn = false;
   int sig_stat = 0;
@@ -1752,6 +1784,10 @@ restart:
     goto cleanup;
 #endif
 
+  result = gpgme_op_decrypt_result(ctx);
+  if (result && (state->flags & STATE_DISPLAY))
+    show_encryption_info(state, result);
+
   if (err != GPG_ERR_NO_ERROR)
   {
     if (is_smime && !maybe_signed && (gpg_err_code(err) == GPG_ERR_NO_DATA))
@@ -1760,9 +1796,6 @@ restart:
        * header told us.  Retry then.  gpgsm returns the error information
        * "unsupported Algorithm '?'" but GPGME will not store this unknown
        * algorithm, thus we test that it has not been set. */
-      gpgme_decrypt_result_t result = NULL;
-
-      result = gpgme_op_decrypt_result(ctx);
       if (result && !result->unsupported_algorithm)
       {
         maybe_signed = true;
@@ -2584,6 +2617,10 @@ int pgp_gpgme_application_handler(struct Body *b, struct State *state)
           }
         }
         redraw_if_needed(ctx);
+
+        gpgme_decrypt_result_t result = gpgme_op_decrypt_result(ctx);
+        if (result && (state->flags & STATE_DISPLAY))
+          show_encryption_info(state, result);
 
         if (err != GPG_ERR_NO_ERROR)
         {
--- a/ncrypt/crypt_gpgme.c
+++ b/ncrypt/crypt_gpgme.c
@@ -1745,14 +1745,15 @@ restart:
   }
   gpgme_data_release(ciphertext);
   ciphertext = NULL;
+
+#ifdef USE_AUTOCRYPT
+  // Abort right away and silently.  Autocrypt will retry on the normal keyring.
+  if (OptAutocryptGpgme && (err != GPG_ERR_NO_ERROR))
+    goto cleanup;
+#endif
+
   if (err != GPG_ERR_NO_ERROR)
   {
-#ifdef USE_AUTOCRYPT
-    /* Abort right away and silently.
-     * Autocrypt will retry on the normal keyring. */
-    if (OptAutocryptGpgme)
-      goto cleanup;
-#endif
     if (is_smime && !maybe_signed && (gpg_err_code(err) == GPG_ERR_NO_DATA))
     {
       /* Check whether this might be a signed message despite what the mime
--- a/ncrypt/crypt_gpgme.c
+++ b/ncrypt/crypt_gpgme.c
@@ -1432,6 +1432,9 @@ static void show_one_recipient(struct State *state, gpgme_recipient_t r)
  */
 static void show_encryption_info(struct State *state, gpgme_decrypt_result_t result)
 {
+  if (!cs_subset_bool(NeoMutt->sub, "crypt_encryption_info"))
+    return;
+
   state_attach_puts(state, _("[-- Begin encryption information --]\n"));
 
   for (gpgme_recipient_t r = result->recipients; r; r = r->next)
--- a/docs/config.c
+++ b/docs/config.c
@@ -918,6 +918,15 @@
 ** (Crypto only)
 */
 
+{ "crypt_encryption_info", DT_BOOL, true },
+/*
+** .pp
+** If \fIset\fP, NeoMutt will include an informative block
+** before an encrypted part,
+** with details about the encryption.
+** (Crypto only)
+*/
+
 { "crypt_timestamp", DT_BOOL, true },
 /*
 ** .pp
--- a/docs/manual.xml.head
+++ b/docs/manual.xml.head
@@ -14603,14 +14603,16 @@ Recipient: RSA key, ID 0000000000000000
       <sect2 id="encryption-info-usage">
         <title>Usage</title>
         <para>
-          It's always enabled.
+          It's enabled by default.
+          The variable to control it is
+          <literal>crypt_encryption_info</literal>.
         </para>
       </sect2>
 
       <sect2 id="encryption-info-credits">
         <title>Credits</title>
         <para>
-          Alejandro Colomar, Richard Russon
+          Alejandro Colomar, Richard Russon, наб
         </para>
       </sect2>
     </sect1>
--- a/docs/manual.xml.head
+++ b/docs/manual.xml.head
@@ -14569,6 +14569,52 @@ set smime_default_key = "SMIME-KEY"
       </sect2>
     </sect1>
 
+    <sect1 id="encryption-info">
+      <title>Encryption information block</title>
+      <subtitle>Show a block of encryption information about a message
+      </subtitle>
+
+      <sect2 id="encryption-info-support">
+        <title>Support</title>
+        <para>
+          <emphasis role="bold">Since:</emphasis> NeoMutt 2024-04-10
+        </para>
+        <para>
+          <emphasis role="bold">Dependencies:</emphasis> Gpgme
+        </para>
+      </sect2>
+
+      <sect2 id="encryption-info-intro">
+        <title>Introduction</title>
+        <para>
+          This feature shows a block of information that provides
+          information about an encrypted message;
+          mainly about its recipients.
+          Here is an example of such a block:
+        </para>
+<screen>
+[-- Begin encryption information --]
+Recipient: RSA key, ID 00C14A7DBBDD521C
+Recipient: RSA key, ID 0000000000000000
+[-- End encryption information --]
+</screen>
+      </sect2>
+
+      <sect2 id="encryption-info-usage">
+        <title>Usage</title>
+        <para>
+          It's always enabled.
+        </para>
+      </sect2>
+
+      <sect2 id="encryption-info-credits">
+        <title>Credits</title>
+        <para>
+          Alejandro Colomar, Richard Russon
+        </para>
+      </sect2>
+    </sect1>
+
     <sect1 id="fmemopen">
       <title>Fmemopen Feature</title>
       <subtitle>Replace some temporary files with memory buffers</subtitle>
--- a/ncrypt/config.c
+++ b/ncrypt/config.c
@@ -126,6 +126,9 @@ static struct ConfigDef NcryptVars[] = {
   { "crypt_protected_headers_write", DT_BOOL, false, 0, NULL,
     "Generate protected header (Memory Hole) for signed and encrypted emails"
   },
+  { "crypt_encryption_info", DT_BOOL, true, 0, NULL,
+    "Add an informative block with details about the encryption"
+  },
   { "crypt_timestamp", DT_BOOL, true, 0, NULL,
     "Add a timestamp to PGP or SMIME output to prevent spoofing"
   },


alejandro-colomar:
v1b (by @flatcap) changes:

-  Rebase on main

```
$ git range-diff 2dde10ea0^..2dde10ea0 neomutt/main..gh/gpg 
1:  2dde10ea0 = 1:  f09907fe9 ncrypt/crypt_gpgme.c: Show encryption information
```

alejandro-colomar:
v1c changes:

-  Rebase on main
-  Add some links

```
$ git range-diff gh/gpg^..gh/gpg neomutt/main..gpg
1:  f09907fe9 ! 1:  7ed3c312c ncrypt/crypt_gpgme.c: Show encryption information
    @@ Commit message
         pinch of salt.
     
         Closes: <https://github.com/neomutt/neomutt/issues/4163>
    +    Link: <https://www.gnupg.org/documentation/manuals/gpgme.pdf#Public%20Key%20Algorithms>
    +    Link: <https://www.gnupg.org/documentation/manuals/gpgme.pdf#Decrypt>
    +    Link: <https://linux.codidact.com/posts/291205>
         Cc: Richard Russon <rich@flatcap.org>
         Cc: Werner Koch <wk@gnupg.org>
```

alejandro-colomar:
(BTW, @flatcap , the feature page for the -C flag hasn't been updated with the recent fix (s/neomuttrc/gitconfig/).)

flatcap:
>  the feature page for the -C flag hasn't been updated with the recent fix (s/neomuttrc/gitconfig/).)

I pushed a docs update an hour ago.
Updating the docs is a fiddly manual job, so it doesn't happen that often.

Do this look correct, now?  https://neomutt.org/feature/cli-crypto

alejandro-colomar:
> > the feature page for the -C flag hasn't been updated with the recent fix (s/neomuttrc/gitconfig/).)
> 
> I pushed a docs update an hour ago. Updating the docs is a fiddly manual job, so it doesn't happen that often.
> 
> Do this look correct, now? https://neomutt.org/feature/cli-crypto

Ahh, thanks!  Yep, it looks good now.

alejandro-colomar:
v2 changes:

-  Call a gpgme function to get the algo string.
-  Use a format similar to gpg-agent(1): `"%s key, ID %s\n", algo, keyid`

```
$ git range-diff main gh/gpg gpg 
1:  7ed3c312c ! 1:  28d469fb7 ncrypt/crypt_gpgme.c: Show encryption information
    @@ ncrypt/crypt_gpgme.c: static void print_smime_keyinfo(const char *msg, gpgme_sig
     +{
     +  const char  *algo;
     +
    ++  algo = gpgme_pubkey_algo_name(r->pubkey_algo);
    ++  if (algo == NULL)
    ++    algo = "?";
    ++
     +  state_puts(state, "Recipient: ");
    -+  switch (r->pubkey_algo) {
    -+  case GPGME_PK_RSA:
    -+  case GPGME_PK_RSA_E:
    -+  case GPGME_PK_RSA_S:
    -+    algo = "(RSA)      ";
    -+    break;
    -+  case GPGME_PK_DSA:
    -+    algo = "(DSA)      ";
    -+    break;
    -+  case GPGME_PK_ELG:
    -+  case GPGME_PK_ELG_E:
    -+    algo = "(ElGamal)  ";
    -+    break;
    -+  case GPGME_PK_ECC:
    -+    algo = "(Elliptic) ";
    -+    break;
    -+  case GPGME_PK_ECDSA:
    -+    algo = "(ECDSA)    ";
    -+    break;
    -+  case GPGME_PK_ECDH:
    -+    algo = "(ECDH)     ";
    -+    break;
    -+  case GPGME_PK_EDDSA:
    -+    algo = "(EdDSA)    ";
    -+    break;
    -+  default:
    -+    algo = "           ";
    -+    break;
    -+  }
     +  state_puts(state, algo);
    ++  state_puts(state, " key, ID ");
     +  state_puts(state, r->keyid);
     +  state_puts(state, "\n");
     +}
```

alejandro-colomar:
Hmmm, interesting thing:  If the recipient is hidden (which I've tested by manually crafting a message with gpg(1)), gpgme reports a key ID consisting of all 0s.  The following is an experiment encrypted to me (non-hidden), and two other recipients (both with `--hidden-recipient`).

```
[-- Begin encryption information --]
Recipient: RSA key, ID 00C14A7DBBDD521C
Recipient: RSA key, ID 0000000000000000
Recipient: RSA key, ID 0000000000000000
[-- End encryption information --]
````

**Comment by flatcap** [comment]
> I like that behavior for now.

Yeah, it looks good.

A possible alternative is
```
Recipient: RSA key, ID [hidden]
```

File: ncrypt/crypt_gpgme.c [hunk 1]
```diff
@@ -1404,6 +1407,39 @@ static void print_smime_keyinfo(const char *msg, gpgme_signature_t sig,
...
+ */
+static void show_one_recipient(struct State *state, gpgme_recipient_t r)
+{
+  const char  *algo;
+
+  algo = gpgme_pubkey_algo_name(r->pubkey_algo);
+  if (algo == NULL)
+    algo = "?";
+
+  state_puts(state, "Recipient: ");
```
alejandro-colomar, line 1436: @flatcap 

When I press `v` from the pager (to view attachments), this line provokes a segmentation fault.

If you apply the following debugging diff on top of this patch, you'll see that the segfault is precisely there (the first one prints, but it crashes before the second one):

```
diff --git a/ncrypt/crypt_gpgme.c b/ncrypt/crypt_gpgme.c
index 247166179..d2b05dec7 100644
--- a/ncrypt/crypt_gpgme.c
+++ b/ncrypt/crypt_gpgme.c
@@ -1432,8 +1432,12 @@ static void show_one_recipient(struct State *state, gpgme_recipient_t r)
   algo = gpgme_pubkey_algo_name(r->pubkey_algo);
   if (algo == NULL)
     algo = "?";
+fprintf(stderr, "ALX  %s(): %d\n", __func__, __LINE__);
+getchar();
 
   state_puts(state, "Recipient: ");
+fprintf(stderr, "ALX  %s(): %d\n", __func__, __LINE__);
+getchar();
   state_puts(state, algo);
   state_puts(state, " key, ID ");
   state_puts(state, r->keyid);
@@ -1779,6 +1783,7 @@ restart:
   ciphertext = NULL;
 
   result = gpgme_op_decrypt_result(ctx);
+fprintf(stderr, "ALX  %s(): %d\n", __func__, __LINE__);
   if (result)
     show_encryption_info(state, result);
 
@@ -2583,6 +2588,7 @@ int pgp_gpgme_application_handler(struct Body *b, struct State *state)
           gpgme_decrypt_result_t result = NULL;
 
           result = gpgme_op_decrypt_result(ctx);
+fprintf(stderr, "ALX  %s(): %d\n", __func__, __LINE__);
           if (result)
             show_encryption_info(state, result);
 
```

(You'll also notice that the call is done from *decrypt_part*().)

I guess the screen is in a state that doesn't want to print this stuff, but do you know/suspect why?  What can we do about it?

alejandro-colomar:
> > I like that behavior for now.
> 
> Yeah, it looks good.
> 
> A possible alternative is
> 
> ```
> Recipient: RSA key, ID [hidden]
> ```

Yup, that's a good alternative.  For now, I'll be lazy and not code that.  :)
But we can definitely do that if 0s confuse anyone.

ncrypt/crypt_gpgme.c
alejandro-colomar: Maybe I should *state_attach_puts*() there too?  I don't know what each of those do.

alejandro-colomar: Or maybe check `(state->flags & STATE_DISPLAY)`?  Or maybe both?  I don't really understand those flags too much yet.

flatcap:
A quick look at the crash (it's late here)...
I see:
```
(gdb) p	*state
$2 = {fp_in = 0x615000016600, fp_out = 0x0, prefix = 0x0, flags = 0, wraplen = 0}
```

It's writing to `state->fp_out == NULL`

alejandro-colomar:
> A quick look at the crash (it's late here)... I see:
> 
> ```
> (gdb) p	*state
> $2 = {fp_in = 0x615000016600, fp_out = 0x0, prefix = 0x0, flags = 0, wraplen = 0}
> ```
> 
> It's writing to `state->fp_out == NULL`
> 
> Update: Though `decrypt_part(..., FILE *fp_out, ...)` is valid.

Those two seem to always have different values.  Without triggering the crash, I see that
`state->fp_out=0x559e77af3c30`
but
`       fp_out=0x559e77afe2f0`

I don't understand what those two pointers are use for.  :|

alejandro-colomar:
> > A quick look at the crash (it's late here)... I see:
> > ```
> > (gdb) p	*state
> > $2 = {fp_in = 0x615000016600, fp_out = 0x0, prefix = 0x0, flags = 0, wraplen = 0}
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > It's writing to `state->fp_out == NULL`
> > Update: Though `decrypt_part(..., FILE *fp_out, ...)` is valid.
> 
> Those two seem to always have different values. Without triggering the crash, I see that `state->fp_out=0x559e77af3c30` but ` fp_out=0x559e77afe2f0`
> 
> I don't understand what those two pointers are used for. :|

Hmmm, I checked that the time that `state->fp_out` is NULL, `state->flags` is 0, so a check `(state->flags & STATE_DISPLAY)` would prevent the crash, I suspect.

I've also made sense in my head of the difference between `state->fp_out` and `fp_out`: the latter seems to be (just) the message, while the former is the entire screen, which includes the informative blocks (crypto, sig) and of course the message.  When viewing attachments, the messages exist, but there's no wrapper screen, so a NULL might make sense.  This is all supposition, though; don't trust me much.

I still have no clue what `_attach_` is.

flatcap:
My best guess of how it works, is...
- An email might need several steps of processing to display it
- Each `State` is one of those steps
- The `State`s are _maybe_ chained together

These "handlers" originate in `mutt_body_handler()`
`mutt_body_handler()` may be called recursively.

Where does the extra `fp_out` fit in?  No idea, yet :-)

alejandro-colomar:
v3 changes:

-  Fix crash (make the print conditional)
-  Explicitly test `!= NULL`  (@flatcap , is that OK?  I see mixed style in the project)
-  Add feature page

```
$ git range-diff main gh/gpg gpg 
1:  28d469fb7 ! 1:  cb4c830dc ncrypt/crypt_gpgme.c: Show encryption information
    @@ ncrypt/crypt_gpgme.c: static void print_smime_keyinfo(const char *msg, gpgme_sig
     + */
     +static void show_encryption_info(struct State *state, gpgme_decrypt_result_t result)
     +{
    -+    if (state->flags & STATE_DISPLAY)
    -+      state_attach_puts(state, _("[-- Begin encryption information --]\n"));
    ++    state_attach_puts(state, _("[-- Begin encryption information --]\n"));
     +
     +    for (gpgme_recipient_t r = result->recipients; r; r = r->next)
     +      show_one_recipient(state, r);
     +
    -+    if (state->flags & STATE_DISPLAY)
    -+      state_attach_puts(state, _("[-- End encryption information --]\n\n"));
    ++    state_attach_puts(state, _("[-- End encryption information --]\n\n"));
     +}
     +
     +/**
    @@ ncrypt/crypt_gpgme.c: restart:
        ciphertext = NULL;
     +
     +  result = gpgme_op_decrypt_result(ctx);
    -+  if (result)
    ++  if (result != NULL && (state->flags & STATE_DISPLAY))
     +    show_encryption_info(state, result);
     +
        if (err != 0)
    @@ ncrypt/crypt_gpgme.c: int pgp_gpgme_application_handler(struct Body *b, struct S
     +          gpgme_decrypt_result_t result = NULL;
     +
     +          result = gpgme_op_decrypt_result(ctx);
    -+          if (result)
    ++          if (result != NULL && (state->flags & STATE_DISPLAY))
     +            show_encryption_info(state, result);
      
                /* Check whether signatures have been verified.  */
-:  --------- > 2:  869884246 docs: Add feature page for the encryption information block
```

alejandro-colomar:
v3b changes:

-  Fix indentation.
-  Explicitly test `!= NULL` in loop controlling expression

```
$ git range-diff main gh/gpg gpg 
1:  cb4c830dc ! 1:  a3274d3bf ncrypt/crypt_gpgme.c: Show encryption information
    @@ ncrypt/crypt_gpgme.c: static void print_smime_keyinfo(const char *msg, gpgme_sig
     + */
     +static void show_encryption_info(struct State *state, gpgme_decrypt_result_t result)
     +{
    -+    state_attach_puts(state, _("[-- Begin encryption information --]\n"));
    ++  state_attach_puts(state, _("[-- Begin encryption information --]\n"));
     +
    -+    for (gpgme_recipient_t r = result->recipients; r; r = r->next)
    -+      show_one_recipient(state, r);
    ++  for (gpgme_recipient_t r = result->recipients; r != NULL; r = r->next)
    ++    show_one_recipient(state, r);
     +
    -+    state_attach_puts(state, _("[-- End encryption information --]\n\n"));
    ++  state_attach_puts(state, _("[-- End encryption information --]\n\n"));
     +}
     +
     +/**
2:  869884246 = 2:  adeebfdbb docs: Add feature page for the encryption information block
```

alejandro-colomar:
When does the *pgp_gpgme_application_handler*() handler run?  I don't know how to test it.  :|

flatcap:
(This is from a quick trawl of the code, not from understanding :-)

An email with an attachment of `application/pgp` (or something like that) will trigger it.

In [`mutt_body_handler()`](https://github.com/neomutt/neomutt/blob/main/handler.c#L1631),

https://github.com/neomutt/neomutt/blob/37d021f7257c98e6e82ce64ec5129790a8e5578a/handler.c#L1666-L1670

Then we set the handler to `crypt_pgp_application_handler`.
This calls the Crypto API function `application_handler()`.
GPGME implements the API, here

https://github.com/neomutt/neomutt/blob/37d021f7257c98e6e82ce64ec5129790a8e5578a/ncrypt/crypt_mod_pgp_gpgme.c#L58-L67

Leading us to

https://github.com/neomutt/neomutt/blob/37d021f7257c98e6e82ce64ec5129790a8e5578a/ncrypt/crypt_gpgme.c#L2427

alejandro-colomar:
> application/pgp

Would you mind sending me an email with such an attachment?  I don't know how to write one.  ;)

alejandro-colomar:
> > application/pgp
> 
> Would you mind sending me an email with such an attachment? I don't know how to write one. ;)

Ahh, this worked: attaching an encrypted file.

```
[-- Attachment #2: bar.asc --]
[-- Type: application/pgp-keys, Encoding: 7bit, Size: 1.5K --]

[-- Begin encryption information --]
Recipient: RSA key, ID 4A83A79DA571C41C
Recipient: RSA key, ID 0000000000000000
[-- End encryption information --]

[-- BEGIN PGP MESSAGE --]

foo

[-- END PGP MESSAGE --]
```

alejandro-colomar:
I'm done with testing, I think.

There are a couple of things that I have doubts:

-  Is it possible to fake a recipient?  This wouldn't be our problem, as we're just showing gpg(1) data.  However, we may contribute to give some appearance of correctness to some attacks, as the one reported here: <https://github.com/neomutt/neomutt/issues/4223>.  This is more of a doubt, but doesn't affect the outcome of this patch.  Maybe I'd reword the commit message, and the feature page, depending on the answer.

-  Why are there *state_attach_puts*() and *state_puts*()?  I don't see why one is used in some cases and the other in others.  This makes me uncomfortable with those calls in the patch.  I can't explain why I used both.

Other than that, I'll polish the commit message and the feature page, and ping you for reviewing the patch.

alejandro-colomar:
v4 changes:

-  Also show the encryption information block in the application/pgp handler when there are decryption errors.

```
$ git range-diff main gh/gpg gpg 
1:  a3274d3bf ! 1:  ba8b0219f ncrypt/crypt_gpgme.c: Show encryption information
    @@ ncrypt/crypt_gpgme.c: restart:
              maybe_signed = true;
     @@ ncrypt/crypt_gpgme.c: int pgp_gpgme_application_handler(struct Body *b, struct State *state)
      
    -           bool sig_stat = false;
    -           char *tmpfname = NULL;
    -+          gpgme_decrypt_result_t result = NULL;
    -+
    -+          result = gpgme_op_decrypt_result(ctx);
    -+          if (result != NULL && (state->flags & STATE_DISPLAY))
    -+            show_encryption_info(state, result);
    +   gpgme_error_t err = 0;
    +   gpgme_data_t armored_data = NULL;
    ++  gpgme_decrypt_result_t result = NULL;
      
    -           /* Check whether signatures have been verified.  */
    -           gpgme_verify_result_t verify_result = gpgme_op_verify_result(ctx);
    +   bool maybe_goodsig = true;
    +   bool have_any_sigs = false;
    +@@ ncrypt/crypt_gpgme.c: int pgp_gpgme_application_handler(struct Body *b, struct State *state)
    +         }
    +         redraw_if_needed(ctx);
    + 
    ++        result = gpgme_op_decrypt_result(ctx);
    ++        if (result != NULL && (state->flags & STATE_DISPLAY))
    ++          show_encryption_info(state, result);
    ++
    +         if (err != 0)
    +         {
    +           char errbuf[200] = { 0 };
2:  adeebfdbb = 2:  ca258ec8c docs: Add feature page for the encryption information block
```

I've tested it with a mail containing an attachment that I can't decrypt, and now shows the following:

```
[-- Attachment #2: qwe.asc --]
[-- Type: application/pgp-keys, Encoding: 7bit, Size: 0.8K --]

[-- Begin encryption information --]
Recipient: RSA key, ID 4A83A79DA571C41C
[-- End encryption information --]

Error: decryption/verification failed: No secret key
[-- BEGIN PGP MESSAGE --]


[-- END PGP MESSAGE --]



[-- End of signed data --]
```

alejandro-colomar:
v4b changes:

-  Reduce scope of variable

```
$ git range-diff main gh/gpg gpg 
1:  ba8b0219f ! 1:  0a875fb63 ncrypt/crypt_gpgme.c: Show encryption information
    @@ ncrypt/crypt_gpgme.c: restart:
            {
              maybe_signed = true;
     @@ ncrypt/crypt_gpgme.c: int pgp_gpgme_application_handler(struct Body *b, struct State *state)
    +       {
    +         gpgme_data_t plaintext = create_gpgme_data();
    +         gpgme_ctx_t ctx = create_gpgme_context(false);
    ++        gpgme_decrypt_result_t result = NULL;
      
    -   gpgme_error_t err = 0;
    -   gpgme_data_t armored_data = NULL;
    -+  gpgme_decrypt_result_t result = NULL;
    - 
    -   bool maybe_goodsig = true;
    -   bool have_any_sigs = false;
    +         if (clearsign)
    +         {
     @@ ncrypt/crypt_gpgme.c: int pgp_gpgme_application_handler(struct Body *b, struct State *state)
              }
              redraw_if_needed(ctx);
2:  ca258ec8c = 2:  f7650a686 docs: Add feature page for the encryption information block
```

alejandro-colomar:
v4c changes:

-  Rebase on main.
-  Document @flatcap 's help during development of the patch.
-  Clarify the relation with bug #4223 
-  Hide current details in the feature page (that hidden recipients are 0s) that we may want to change in the future.

```
$ git range-diff main..gh/gpg neomutt/main..gpg 
1:  0a875fb63 ! 1:  c0fe74125 ncrypt/crypt_gpgme.c: Show encryption information
    @@ Commit message
         Therefore, this should be taken only as an informative block, and with a
         pinch of salt.
     
    +    It would be good to protect the To and Cc header fields, to provide a
    +    guarantee to the recipient that the listed recipients match the original
    +    recipients as intended by the sender.  See the bug #4223 linked below.
    +
         Closes: <https://github.com/neomutt/neomutt/issues/4163>
         Link: <https://www.gnupg.org/documentation/manuals/gpgme.pdf#Public%20Key%20Algorithms>
         Link: <https://www.gnupg.org/documentation/manuals/gpgme.pdf#Decrypt>
         Link: <https://linux.codidact.com/posts/291205>
    -    Cc: Richard Russon <rich@flatcap.org>
    +    Link: <https://github.com/neomutt/neomutt/issues/4223>
    +    Co-developed-by: Richard Russon <rich@flatcap.org>
         Cc: Werner Koch <wk@gnupg.org>
     
2:  f7650a686 ! 2:  e23019dfc docs: Add feature page for the encryption information block
    @@ docs/manual.xml.head: set smime_default_key = "SMIME-KEY"
     +      <sect2 id="encryption-info-support">
     +        <title>Support</title>
     +        <para>
    -+          <emphasis role="bold">Since:</emphasis> NeoMutt 2024-??-??
    ++          <emphasis role="bold">Since:</emphasis> NeoMutt 2024-03-30
     +        </para>
     +        <para>
     +          <emphasis role="bold">Dependencies:</emphasis> Gpgme
    @@ docs/manual.xml.head: set smime_default_key = "SMIME-KEY"
     +        <title>Introduction</title>
     +        <para>
     +          This feature shows a block of information that provides
    -+          information about an encrypted message,
    ++          information about an encrypted message;
     +          mainly about its recipients.
     +          Here is an example of such a block:
     +        </para>
    @@ docs/manual.xml.head: set smime_default_key = "SMIME-KEY"
     +Recipient: RSA key, ID 0000000000000000
     +[-- End encryption information --]
     +</screen>
    -+        <para>
    -+          Hidden recipients (see gpg(1)'s --hidden-recipient)
    -+          are shown with a key ID consisting of all 0s.
    -+        </para>
     +      </sect2>
     +
     +      <sect2 id="encryption-info-usage">
    @@ docs/manual.xml.head: set smime_default_key = "SMIME-KEY"
     +      <sect2 id="encryption-info-credits">
     +        <title>Credits</title>
     +        <para>
    -+          Alejandro Colomar
    ++          Alejandro Colomar, Richard Russon
     +        </para>
     +      </sect2>
     +    </sect1>
```

alejandro-colomar:
@flatcap It's all yours for review.  ;-)

ncrypt/crypt_gpgme.c
alejandro-colomar: The crash has been fixed.  I'll resolve this thread.

alejandro-colomar:
v4d changes:

-  Rebase on main

```
$ git range-diff main..gh/gpg neomutt/main..gpg 
1:  c0fe74125 = 1:  3beb2e015 ncrypt/crypt_gpgme.c: Show encryption information
2:  e23019dfc = 2:  3af7dc9e4 docs: Add feature page for the encryption information block
```

flatcap:
A light sprinkling of tidy :-)
Squash when you've read them.

- Reorder functions
- Tidy code

My preference for pointers is `if (ptr)` and `if (!ptr)`.
There are some exceptions, most notably in the tests where the macro needs an explicit bool.

ncrypt/crypt_gpgme.c
```diff
@@ -1424,12 +1424,14 @@ static void show_one_recipient(struct State *state, gpgme_recipient_t r)
 
 /**
  * show_encryption_info - Show encryption information
+ * @param state  State to write to
+ * @param result Encryption recipients
```
alejandro-colomar, line 1428: Actually, `result` contains slightly more than the recipients.  We're not using the extra info, but there is (and we may want to expand this block in the future to show it).

```c
$ grepc gpgme_decrypt_result_t /usr/include/
/usr/include/gpgme.h:typedef struct _gpgme_op_decrypt_result *gpgme_decrypt_result_t;
```
```c
$ grepc _gpgme_op_decrypt_result /usr/include/
/usr/include/gpgme.h:struct _gpgme_op_decrypt_result
{
  char *unsupported_algorithm;

  /* Key should not have been used for encryption.  */
  unsigned int wrong_key_usage : 1;

  /* True if the message was encrypted in compliance to the de-vs
   * mode.  */
  unsigned int is_de_vs : 1;

  /* The message claims that the content is a MIME object.  */
  unsigned int is_mime : 1;

  /* The message was made by a legacy algorithm without any integrity
   * protection.  This might be an old but legitimate message. */
  unsigned int legacy_cipher_nomdc : 1;

  /* Internal to GPGME, do not use.  */
  int _unused : 28;

  gpgme_recipient_t recipients;

  /* The original file name of the plaintext message, if
   * available.  */
  char *file_name;

  /* A textual representation of the session key used to decrypt the
   * message, if available */
  char *session_key;

  /* A string with the symmetric encryption algorithm and mode using
   * the format "<algo>.<mode>".  */
  char *symkey_algo;
};
```

I would say "Decryption result" in the comment.

alejandro-colomar: @flatcap

alejandro-colomar:
v4e changes:

-  Squash @flatcap 's fixups
-  Reword doxygen imprecise `@param` comment

```
$ git range-diff main gh/gpg gpg 
1:  3beb2e015 ! 1:  d1e26ee34 ncrypt/crypt_gpgme.c: Show encryption information
    @@ Commit message
     
      ## ncrypt/crypt_gpgme.c ##
    -@@ ncrypt/crypt_gpgme.c: static char *CurrentSender = NULL;
    - #define BEGIN_PGP_SIGNATURE(_y)                                                \
    -   _LINE_COMPARE("-----BEGIN PGP SIGNATURE-----", _y)
    - 
    -+static void show_encryption_info(struct State *state, gpgme_decrypt_result_t result);
    -+static void show_one_recipient(struct State *state, gpgme_recipient_t r);
    -+
    - /**
    -  * is_pka_notation - Is this the standard pka email address
    -  * @param notation GPGME notation
     @@ ncrypt/crypt_gpgme.c: static void print_smime_keyinfo(const char *msg, gpgme_signature_t sig,
        }
      }
      
    -+/**
    -+ * show_encryption_info - Show encryption information
    -+ */
    -+static void show_encryption_info(struct State *state, gpgme_decrypt_result_t result)
    -+{
    -+  state_attach_puts(state, _("[-- Begin encryption information --]\n"));
    -+
    -+  for (gpgme_recipient_t r = result->recipients; r != NULL; r = r->next)
    -+    show_one_recipient(state, r);
    -+
    -+  state_attach_puts(state, _("[-- End encryption information --]\n\n"));
    -+}
    -+
     +/**
     + * show_one_recipient - Show information about one encryption recipient
    ++ * @param state State to write to
    ++ * @param r     Encryption recipient
     + */
     +static void show_one_recipient(struct State *state, gpgme_recipient_t r)
     +{
    -+  const char  *algo;
    -+
    -+  algo = gpgme_pubkey_algo_name(r->pubkey_algo);
    -+  if (algo == NULL)
    ++  const char *algo = gpgme_pubkey_algo_name(r->pubkey_algo);
    ++  if (!algo)
     +    algo = "?";
     +
     +  state_puts(state, "Recipient: ");
    @@ ncrypt/crypt_gpgme.c: static void print_smime_keyinfo(const char *msg, gpgme_sig
     +  state_puts(state, r->keyid);
     +  state_puts(state, "\n");
     +}
    ++
    ++/**
    ++ * show_encryption_info - Show encryption information
    ++ * @param state  State to write to
    ++ * @param result Decryption result
    ++ */
    ++static void show_encryption_info(struct State *state, gpgme_decrypt_result_t result)
    ++{
    ++  state_attach_puts(state, _("[-- Begin encryption information --]\n"));
    ++
    ++  for (gpgme_recipient_t r = result->recipients; r; r = r->next)
    ++    show_one_recipient(state, r);
    ++
    ++  state_attach_puts(state, _("[-- End encryption information --]\n\n"));
    ++}
     +
      /**
       * show_one_sig_status - Show information about one signature
    @@ ncrypt/crypt_gpgme.c: restart:
        ciphertext = NULL;
     +
     +  result = gpgme_op_decrypt_result(ctx);
    -+  if (result != NULL && (state->flags & STATE_DISPLAY))
    ++  if (result && (state->flags & STATE_DISPLAY))
     +    show_encryption_info(state, result);
     +
        if (err != 0)
    @@ ncrypt/crypt_gpgme.c: restart:
            if (result && !result->unsupported_algorithm)
            {
              maybe_signed = true;
    -@@ ncrypt/crypt_gpgme.c: int pgp_gpgme_application_handler(struct Body *b, struct State *state)
    -       {
    -         gpgme_data_t plaintext = create_gpgme_data();
    -         gpgme_ctx_t ctx = create_gpgme_context(false);
    -+        gpgme_decrypt_result_t result = NULL;
    - 
    -         if (clearsign)
    -         {
     @@ ncrypt/crypt_gpgme.c: int pgp_gpgme_application_handler(struct Body *b, struct State *state)
              }
              redraw_if_needed(ctx);
      
    -+        result = gpgme_op_decrypt_result(ctx);
    -+        if (result != NULL && (state->flags & STATE_DISPLAY))
    ++        gpgme_decrypt_result_t result = gpgme_op_decrypt_result(ctx);
    ++        if (result && (state->flags & STATE_DISPLAY))
     +          show_encryption_info(state, result);
     +
              if (err != 0)
2:  3af7dc9e4 = 2:  c3d9dc7ae docs: Add feature page for the encryption information block
3:  6b20c50ac < -:  --------- fixup! ncrypt/crypt_gpgme.c: Show encryption information
4:  ec14331aa < -:  --------- fixup! ncrypt/crypt_gpgme.c: Show encryption information
```

alejandro-colomar:
v4f changes:

-  Rebase on main

```
$ git range-diff main..gh/gpg neomutt/main..gpg 
1:  d1e26ee34 = 1:  5fac04450 ncrypt/crypt_gpgme.c: Show encryption information
2:  c3d9dc7ae = 2:  5b102b2ad docs: Add feature page for the encryption information block
```

alejandro-colomar:
v4g changes:

-  Rebase on main

```
$ git range-diff main..gh/gpg neomutt/main..gpg 
1:  d1e26ee34 = 1:  2bcfbe9bc ncrypt/crypt_gpgme.c: Show encryption information
2:  c3d9dc7ae = 2:  bc36b9f2b docs: Add feature page for the encryption information block
```

flatcap:
It looks good, but the info seems to be duplicated:
```
Subject: ...
User-Agent: NeoMutt/20240329-149-6243fc

[-- Begin encryption information --]
Recipient: RSA key, ID 19D4D91AF95EE3D1
Recipient: RSA key, ID 934F123BFA86E1E9
Recipient: RSA key, ID 7D276576DECDF4F8
Recipient: RSA key, ID 544F271C223CE792
[-- End encryption information --]

[-- Begin encryption information --]
Recipient: RSA key, ID 19D4D91AF95EE3D1
Recipient: RSA key, ID 934F123BFA86E1E9
Recipient: RSA key, ID 7D276576DECDF4F8
Recipient: RSA key, ID 544F271C223CE792
[-- End encryption information --]

[-- The following data is PGP/MIME encrypted --]
```

alejandro-colomar:
> It looks good, but the info seems to be duplicated:
> 
> ```
> Subject: ...
> User-Agent: NeoMutt/20240329-149-6243fc
> 
> [-- Begin encryption information --]
> Recipient: RSA key, ID 19D4D91AF95EE3D1
> Recipient: RSA key, ID 934F123BFA86E1E9
> Recipient: RSA key, ID 7D276576DECDF4F8
> Recipient: RSA key, ID 544F271C223CE792
> [-- End encryption information --]
> 
> [-- Begin encryption information --]
> Recipient: RSA key, ID 19D4D91AF95EE3D1
> Recipient: RSA key, ID 934F123BFA86E1E9
> Recipient: RSA key, ID 7D276576DECDF4F8
> Recipient: RSA key, ID 544F271C223CE792
> [-- End encryption information --]
> 
> [-- The following data is PGP/MIME encrypted --]
> ```

It doesn't show duplicated in the messages I send.  Maybe it has to do with the difference with how you encrypt mail (I already asked you about it in private last week)?  Please send me an email that will reproduce that.

alejandro-colomar:
v4i changes:

-  Rebase on main

```
$ git range-diff main..gh/gpg neomutt/main..gpg 
1:  85e49d91e = 1:  9f093defd ncrypt/crypt_gpgme.c: Show encryption information
2:  7004a67e5 = 2:  efe9a692e docs: Add feature page for the encryption information block
```

alejandro-colomar:
My suspicion regarding your duplicated info is that you're encrypting twice.

```
@@ -1699,6 +1732,7 @@ static struct Body *decrypt_part(struct Body *b, struct State *state,
   struct Body *tattach = NULL;
   int err = 0;
   gpgme_data_t ciphertext = NULL, plaintext = NULL;
+  gpgme_decrypt_result_t result = NULL;
   bool maybe_signed = false;
   bool anywarn = false;
   int sig_stat = 0;
@@ -1740,6 +1774,11 @@ restart:
   }
   gpgme_data_release(ciphertext);
   ciphertext = NULL;
+
+  result = gpgme_op_decrypt_result(ctx);
+  if (result && (state->flags & STATE_DISPLAY))
+    show_encryption_info(state, result);
+
   if (err != 0)
   {
 #ifdef USE_AUTOCRYPT
@@ -1754,9 +1793,6 @@ restart:
        * header told us.  Retry then.  gpgsm returns the error information
        * "unsupported Algorithm '?'" but GPGME will not store this unknown
        * algorithm, thus we test that it has not been set. */
-      gpgme_decrypt_result_t result = NULL;
-
-      result = gpgme_op_decrypt_result(ctx);
       if (result && !result->unsupported_algorithm)
       {
         maybe_signed = true;
@@ -2580,6 +2616,10 @@ int pgp_gpgme_application_handler(struct Body *b, struct State *state)
         }
         redraw_if_needed(ctx);
 
+        gpgme_decrypt_result_t result = gpgme_op_decrypt_result(ctx);
+        if (result && (state->flags & STATE_DISPLAY))
+          show_encryption_info(state, result);
+
         if (err != 0)
         {
           char errbuf[200] = { 0 };
```

So, the second info should be from the application handler.  Add an extra fprintf(stderr) around there to check that the second is an application handler stuff.  Probably handling a part that is found inside the encrypted message, and which is likely to be also encrypted.

flatcap:
I've sent you an email that on arrival show this block twice:

```
[-- Begin encryption information --]
Recipient: RSA key, ID 00C14A7DBBDD521C
Recipient: RSA key, ID 6B3E7AA5D6E67ABC
Recipient: RSA key, ID 68FBF7AE33355DDF
Recipient: RSA key, ID 753B8A1DAE72518C
Recipient: RSA key, ID 5E79FFE0751EF20E
Recipient: RSA key, ID 544F271C223CE792
[-- End encryption information --]
```

> you're encrypting twice

Hmm...  `set pgp_self_encrypt = yes` perhaps?

alejandro-colomar:
> I've sent you an email that on arrival show this block twice:
> 
> ```
> [-- Begin encryption information --]
> Recipient: RSA key, ID 00C14A7DBBDD521C
> Recipient: RSA key, ID 6B3E7AA5D6E67ABC
> Recipient: RSA key, ID 68FBF7AE33355DDF
> Recipient: RSA key, ID 753B8A1DAE72518C
> Recipient: RSA key, ID 5E79FFE0751EF20E
> Recipient: RSA key, ID 544F271C223CE792
> [-- End encryption information --]
> ```
> 
> > you're encrypting twice
> 
> Hmm... `set pgp_self_encrypt = yes` perhaps?

I only see it once.  However I'll test on another computer, and with a fresh compilation, just in case.

alejandro-colomar:
> > you're encrypting twice
> 
> Hmm... `set pgp_self_encrypt = yes` perhaps?

I also use it; it shouldn't be that.

alejandro-colomar:
> It looks good, but the info seems to be duplicated:
> 
> ```
> Subject: ...
> User-Agent: NeoMutt/20240329-149-6243fc
> 
> [-- Begin encryption information --]
> Recipient: RSA key, ID 19D4D91AF95EE3D1
> Recipient: RSA key, ID 934F123BFA86E1E9
> Recipient: RSA key, ID 7D276576DECDF4F8
> Recipient: RSA key, ID 544F271C223CE792
> [-- End encryption information --]
> 
> [-- Begin encryption information --]
> Recipient: RSA key, ID 19D4D91AF95EE3D1
> Recipient: RSA key, ID 934F123BFA86E1E9
> Recipient: RSA key, ID 7D276576DECDF4F8
> Recipient: RSA key, ID 544F271C223CE792
> [-- End encryption information --]
> 
> [-- The following data is PGP/MIME encrypted --]
> ```

Maybe if you bounce me a message where you currently see the dup?  Without creating a new message?

flatcap:
Weird...
Switching between my real and dev accounts, I see a difference on the same emails.

Running with zero config, I see **one** block.
Running with autocrypt enabled, I see **two** blocks.

This is enough, **test.rc**
```
set autocrypt = yes
set autocrypt_dir = "~/.config/neomutt/autocrypt"
```

```sh
./neomutt -n -F test.rc -f test.mbox
```

alejandro-colomar:
> Weird... Switching between my real and dev accounts, I see a difference on the same emails.
> 
> Running with zero config, I see **one** block. Running with autocrypt enabled, I see **two** blocks.
> 
> This is enough, **test.rc**
> 
> ```
> set autocrypt = yes
> set autocrypt_dir = "~/.config/neomutt/autocrypt"
> ```
> 
> ```shell
> ./neomutt -n -F test.rc -f test.mbox
> ```

I don't use autocrypt, so don't have an autocrypt database, which seems to be necessary for this:

```
$ ./neomutt -n -F ~/tmp/test.rc -f ~/tmp/foss/
Unable to open autocrypt database /home/alx/.config/neomutt/autocrypt/autocrypt.db
Unable to open mailbox /home/alx/tmp/foss/
```

I can't help testing unless I create an autocrypt account, but I'll send you a diff that might help me understand why this happens:

```diff
diff --git a/ncrypt/crypt_gpgme.c b/ncrypt/crypt_gpgme.c
index a2cce2c69..bb5bbff70 100644
--- a/ncrypt/crypt_gpgme.c
+++ b/ncrypt/crypt_gpgme.c
@@ -1777,7 +1777,11 @@ restart:
 
   result = gpgme_op_decrypt_result(ctx);
   if (result && (state->flags & STATE_DISPLAY))
+{
+state_puts(state, "ALX: part\n");
     show_encryption_info(state, result);
+state_puts(state, "ALX: part end\n");
+}
 
   if (err != 0)
   {
@@ -2618,7 +2622,11 @@ int pgp_gpgme_application_handler(struct Body *b, struct State *state)
 
         gpgme_decrypt_result_t result = gpgme_op_decrypt_result(ctx);
         if (result && (state->flags & STATE_DISPLAY))
+{
+state_puts(state, "ALX: app\n");
           show_encryption_info(state, result);
+state_puts(state, "ALX: app end\n");
+}
 
         if (err != 0)
         {
```

Can you please show the encryption information blocks after applying that patch, when using autocrypt?

alejandro-colomar:
Thanks!

v5 changes:

-  Squash @flatcap 's fix for autocrypt
-  Use 2 spaces after period (in comment).

```
$ git range-diff main gh/gpg gpg 
1:  9f093defd ! 1:  007c2e6d1 ncrypt/crypt_gpgme.c: Show encryption information
    @@ ncrypt/crypt_gpgme.c: restart:
        }
        gpgme_data_release(ciphertext);
        ciphertext = NULL;
    +-  if (err != 0)
    +-  {
    ++
    + #ifdef USE_AUTOCRYPT
    +-    /* Abort right away and silently.
    +-     * Autocrypt will retry on the normal keyring. */
    +-    if (OptAutocryptGpgme)
    ++    // Abort right away and silently.  Autocrypt will retry on the normal keyring.
    ++    if (OptAutocryptGpgme && (err != 0))
    +       goto cleanup;
    + #endif
     +
     +  result = gpgme_op_decrypt_result(ctx);
     +  if (result && (state->flags & STATE_DISPLAY))
     +    show_encryption_info(state, result);
     +
    -   if (err != 0)
    -   {
    - #ifdef USE_AUTOCRYPT
    -@@ ncrypt/crypt_gpgme.c: restart:
    ++  if (err != 0)
    ++  {
    +     if (is_smime && !maybe_signed && (gpg_err_code(err) == GPG_ERR_NO_DATA))
    +     {
    +       /* Check whether this might be a signed message despite what the mime
             * header told us.  Retry then.  gpgsm returns the error information
             * "unsupported Algorithm '?'" but GPGME will not store this unknown
             * algorithm, thus we test that it has not been set. */
2:  efe9a692e = 2:  55a9af7c2 docs: Add feature page for the encryption information block
3:  c3e979f3a < -:  --------- autocrypt: drop out before show_encryption_info()
```

nabijaczleweli:
Looks fine, but this introduces another tall header that I don't really care about into the gpg mail view; there's probably some scenarios where this is interesting data, but I'd like to be able to disable this by default and just run `set gpg_mail_show_key_info` (or w/e) and reopen.

alejandro-colomar:
> Looks fine, but this introduces another tall header that I don't really care about into the gpg mail view; there's probably some scenarios where this is interesting data, but I'd like to be able to disable this by default and just run `set gpg_mail_show_key_info` (or w/e) and reopen.

This info is not critical, since a sender can add hidden recipients by simply sending separate emails.  It's just informative.  Also, a malicious party might be able to modify the metadata without needing to decrypt it.  So, I think making this optional makes sense.  Thanks for the feedback!

alejandro-colomar:
To: @flatcap 
Cc: @nabijaczleweli 

For such an option, I propose the name `$crypt_show_encryption_info`.  Any opinions?  Anyway, I'll use that name for now and you can comment on the patches when I publish them.

nabijaczleweli:
The current UI is good as-is, and needs to not be messed with. And certainly not in the form of emojis (or boxes). Maybe squeeze the
```
[-- End signature information --]

[-- The following data is PGP/MIME signed and encrypted --]
```
lines together, but meh.

alejandro-colomar:
v6 changes:

-  Add a variable to toggle the feature.  [@nabijaczleweli ]

```
$ git range-diff main gh/gpg gpg 
1:  007c2e6d1 = 1:  007c2e6d1 ncrypt/crypt_gpgme.c: Show encryption information
2:  55a9af7c2 = 2:  55a9af7c2 docs: Add feature page for the encryption information block
-:  --------- > 3:  764a31104 $crypt_encryption_info: Add variable
```

alejandro-colomar:
v6b changes:

-  Rebase on main

```
$ git range-diff main..gh/gpg neomutt/main..gpg 
1:  007c2e6d1 = 1:  007c2e6d1 ncrypt/crypt_gpgme.c: Show encryption information
2:  55a9af7c2 = 2:  55a9af7c2 docs: Add feature page for the encryption information block
3:  764a31104 = 3:  764a31104 $crypt_encryption_info: Add variable
```

alejandro-colomar:
v6c changes:

-  Rebase on main

```
$ git range-diff main..gh/gpg neomutt/main..gpg 
1:  007c2e6d1 = 1:  2841b625e ncrypt/crypt_gpgme.c: Show encryption information
2:  55a9af7c2 = 2:  d8ef632ed docs: Add feature page for the encryption information block
3:  764a31104 = 3:  7292aac39 $crypt_encryption_info: Add variable
```

alejandro-colomar:
v6d changes:

-  Update timestamp

```
$ git range-diff main gh/gpg gpg 
1:  2841b625e = 1:  2841b625e ncrypt/crypt_gpgme.c: Show encryption information
2:  d8ef632ed ! 2:  b25db2d69 docs: Add feature page for the encryption information block
    @@ docs/manual.xml.head: set smime_default_key = "SMIME-KEY"
     +      <sect2 id="encryption-info-support">
     +        <title>Support</title>
     +        <para>
    -+          <emphasis role="bold">Since:</emphasis> NeoMutt 2024-03-30
    ++          <emphasis role="bold">Since:</emphasis> NeoMutt 2024-04-10
     +        </para>
     +        <para>
     +          <emphasis role="bold">Dependencies:</emphasis> Gpgme
3:  7292aac39 = 3:  efd27e017 $crypt_encryption_info: Add variable
```

alejandro-colomar:
> > Looks fine, but this introduces another tall header
> 
> I had a bit of a brainstorming session about the UI...
> 
>     * Most emails **won't** be forgeries
> 
>     * We need a concise way to show that an email is valid
>       (in a way that can't be faked)
> 
>     * All the crypto info must be accessible
> 
> 
> My idea was a fake header (in memory only):
> 
> **Security**: ✅ 🔒 From Richard Russon <[rich@flatcap.org](mailto:rich@flatcap.org)>

I guess this is related to the phishing discussion we had (have?).  Maybe you want to open a new issue about it.  For the purpose of showing headers that don't match the protected ones (i.e., they have been tampered), maybe we could use colors.  Like @nabijaczleweli , I don't like the idea of emojis or boxes, but a color (red?) wouldn't be invasive, and would be clear enough.

My idea would be:

Non-protected email:  Stay the same as always.
Protected email; protected header fields match:  Stay the same as always
Protected email, and the protected fields don't match the unprotected ones:  Print a warning line of the form `[-- Header field Xxx has been tampered --]`.

For encrypted emails, maybe we could check that the encryption recipient list matches the actual recipient list.  We could maybe add a warning if a recipient can't decrypt the message, or if a non-recipient can decrypt it, or if there are hidden recipients.  This would be also interesting, because a common pattern is that the sender doesn't encrypt its own copy, which is rather common from what I've seen (with the people I've talked to, they didn't know they could do it; and they're programmers that know about security), and could be an important leak of secret information.  Such a warning could help find and address those leaks.

Let's open an issue for that, and I'll think about it while working on the current PRs.  We need another one for Cyrillic evilness.  Do you have any answer to the apples I sent you?  (I was starting to think that maybe you responded to the address; I forgot to provide you a Reply-To header; ouch!).  ;)

> (some icons for "signed" and/or "encrypted")

This reminds me of HTTPS, which I dislike so much.  :)

> If the security fails, we display the full bank of crypto info. We'd also have some function like "weed" that toggles the full crypto info.

I tend to dislike this magic too.

alejandro-colomar:
v6e changes:

-  Rebase on devel/security

(Omg, this range-diff is eye-bleeding.  But it seems to be all right.)

```
$ git range-diff devel/security..gpg  neomutt/devel/security..gh/gpg
1:  2841b625e ! 1:  4e46bcbf4 ncrypt/crypt_gpgme.c: Show encryption information
    @@ ncrypt/crypt_gpgme.c: static void print_smime_keyinfo(const char *msg, gpgme_sig
       * @param ctx   GPGME handle of a successful verification
     @@ ncrypt/crypt_gpgme.c: static struct Body *decrypt_part(struct Body *b, struct State *state,
        struct Body *tattach = NULL;
    -   int err = 0;
    +   gpgme_error_t err = GPG_ERR_NO_ERROR;
        gpgme_data_t ciphertext = NULL, plaintext = NULL;
     +  gpgme_decrypt_result_t result = NULL;
        bool maybe_signed = false;
    @@ ncrypt/crypt_gpgme.c: restart:
        }
        gpgme_data_release(ciphertext);
        ciphertext = NULL;
    --  if (err != 0)
    --  {
     +
    - #ifdef USE_AUTOCRYPT
    --    /* Abort right away and silently.
    --     * Autocrypt will retry on the normal keyring. */
    --    if (OptAutocryptGpgme)
    -+    // Abort right away and silently.  Autocrypt will retry on the normal keyring.
    -+    if (OptAutocryptGpgme && (err != 0))
    -       goto cleanup;
    - #endif
    ++#ifdef USE_AUTOCRYPT
    ++  // Abort right away and silently.  Autocrypt will retry on the normal keyring.
    ++  if (OptAutocryptGpgme && (err != GPG_ERR_NO_ERROR))
    ++    goto cleanup;
    ++#endif
     +
     +  result = gpgme_op_decrypt_result(ctx);
     +  if (result && (state->flags & STATE_DISPLAY))
     +    show_encryption_info(state, result);
     +
    -+  if (err != 0)
    -+  {
    +   if (err != GPG_ERR_NO_ERROR)
    +   {
    +-#ifdef USE_AUTOCRYPT
    +-    /* Abort right away and silently.
    +-     * Autocrypt will retry on the normal keyring. */
    +-    if (OptAutocryptGpgme)
    +-      goto cleanup;
    +-#endif
          if (is_smime && !maybe_signed && (gpg_err_code(err) == GPG_ERR_NO_DATA))
          {
            /* Check whether this might be a signed message despite what the mime
    @@ ncrypt/crypt_gpgme.c: int pgp_gpgme_application_handler(struct Body *b, struct S
     +        if (result && (state->flags & STATE_DISPLAY))
     +          show_encryption_info(state, result);
     +
    -         if (err != 0)
    +         if (err != GPG_ERR_NO_ERROR)
              {
                char errbuf[200] = { 0 };
2:  b25db2d69 = 2:  a63a6fe17 docs: Add feature page for the encryption information block
3:  efd27e017 = 3:  a5fd31aaf $crypt_encryption_info: Add variable
```

alejandro-colomar:
v6f changes:

-  Move refactor to separate commit

```
$ git range-diff devel/security gh/gpg gpg 
-:  --------- > 1:  1d010928b ncrypt/crypt_gpgme.c: Split conditional
1:  4e46bcbf4 ! 2:  cb99ec8ab ncrypt/crypt_gpgme.c: Show encryption information
    @@ ncrypt/crypt_gpgme.c: static struct Body *decrypt_part(struct Body *b, struct St
        bool anywarn = false;
        int sig_stat = 0;
     @@ ncrypt/crypt_gpgme.c: restart:
    -   }
    -   gpgme_data_release(ciphertext);
    -   ciphertext = NULL;
    -+
    -+#ifdef USE_AUTOCRYPT
    -+  // Abort right away and silently.  Autocrypt will retry on the normal keyring.
    -+  if (OptAutocryptGpgme && (err != GPG_ERR_NO_ERROR))
    -+    goto cleanup;
    -+#endif
    -+
    +     goto cleanup;
    + #endif
    + 
     +  result = gpgme_op_decrypt_result(ctx);
     +  if (result && (state->flags & STATE_DISPLAY))
     +    show_encryption_info(state, result);
     +
        if (err != GPG_ERR_NO_ERROR)
        {
    --#ifdef USE_AUTOCRYPT
    --    /* Abort right away and silently.
    --     * Autocrypt will retry on the normal keyring. */
    --    if (OptAutocryptGpgme)
    --      goto cleanup;
    --#endif
          if (is_smime && !maybe_signed && (gpg_err_code(err) == GPG_ERR_NO_DATA))
    -     {
    -       /* Check whether this might be a signed message despite what the mime
    +@@ ncrypt/crypt_gpgme.c: restart:
             * header told us.  Retry then.  gpgsm returns the error information
             * "unsupported Algorithm '?'" but GPGME will not store this unknown
             * algorithm, thus we test that it has not been set. */
2:  a63a6fe17 = 3:  4884b978d docs: Add feature page for the encryption information block
3:  a5fd31aaf = 4:  67ff7ec3b $crypt_encryption_info: Add variable
```

alejandro-colomar:
v6g (by @flatcap) changes:

-  Rebase

```
$ git range-diff devel/security..gpg 3c399e9^^^^..3c399e9
1:  1d010928b = 1:  1a3a5f676 ncrypt/crypt_gpgme.c: Split conditional
2:  cb99ec8ab = 2:  59eaf2ddb ncrypt/crypt_gpgme.c: Show encryption information
3:  4884b978d = 3:  6f159be02 docs: Add feature page for the encryption information block
4:  67ff7ec3b = 4:  3c399e919 $crypt_encryption_info: Add variable
```

alejandro-colomar:
v6h changes:

-  Rebase on devel/security

```
$ git range-diff 3c399e9^^^^..3c399e9 neomutt/devel/security..gpg 
1:  1a3a5f676 = 1:  a8ef98b77 ncrypt/crypt_gpgme.c: Split conditional
2:  59eaf2ddb = 2:  4969f8910 ncrypt/crypt_gpgme.c: Show encryption information
3:  6f159be02 = 3:  e00d3e56b docs: Add feature page for the encryption information block
4:  3c399e919 = 4:  c7d7d1502 $crypt_encryption_info: Add variable
```

alejandro-colomar:
v6i changes:

-  Rebase

```
$ git range-diff devel/security..gh/gpg neomutt/devel/security..gpg
1:  a8ef98b77 = 1:  55be3699b ncrypt/crypt_gpgme.c: Split conditional
2:  4969f8910 = 2:  76180da27 ncrypt/crypt_gpgme.c: Show encryption information
3:  e00d3e56b = 3:  582879f10 docs: Add feature page for the encryption information block
4:  c7d7d1502 = 4:  7b29638ef $crypt_encryption_info: Add variable
```

gahr:
Two things (I'm unsure if they have been discussed already, all these gpg-related issues/PRs/discussions are a bit too much for me to get familiar with at this time):
1. Can we do without the extra config? I can't think of a reason why one would want to hide it.
2. Can we please make sure all messages  i18n'd?

flatcap:
> Can we please make sure all messages  are i18n'd?

Don't worry, that'll definitely happen.
We're still hacking and deciding on things, but it'll happen before it gets merged to `main`.

alejandro-colomar:
> Two things (I'm unsure if they have been discussed already, all these gpg-related issues/PRs/discussions are a bit too much for me to get familiar with at this time):
> 
>     1. Can we do without the extra config? I can't think of a reason why one would want to hide it.

I would love that.  I use mutt(1) for a reason: I don't like configuring things, and really don't need any of neomutt(1)'s configuration.

I offered the configurability because @nabijaczleweli complained about verbosity, and wanted a way to keep old behavior.

But if there's consensus on not giving a choice, I'm all in.  In the end, encrypted messages will just have 2 recipients, so 4 lines + 1 blank, which shouldn't really get in the way of reading the mail, IMO.

alejandro-colomar:
This one is ready by my side.  Feel free to merge when you're happy with it, @flatcap.

Feel free to drop the last patch (the config variable, if you don't want it).

ncrypt/crypt_gpgme.c
```diff
@@ -1409,6 +1409,42 @@ static void print_smime_keyinfo(const char *msg, gpgme_signature_t sig,
...
+{
+  const char *algo = gpgme_pubkey_algo_name(r->pubkey_algo);
+  if (!algo)
+    algo = "?";
+
+  state_puts(state, "Recipient: ");
+  state_puts(state, algo);
+  state_puts(state, " key, ID ");
+  state_puts(state, r->keyid);
+  state_puts(state, "\n");
```
nabijaczleweli, line 1427: this needs to be a `state_putf(_("Recipient %s ...` or whatever it's spelled as but otherwise sure

nabijaczleweli (approved):

ncrypt/crypt_gpgme.c
alejandro-colomar: Thanks!  Will do.

alejandro-colomar:
v7 changes:

-  _("")  (except for the newline, which I guess doesn't need to be i18n'd.

```
$ git range-diff devel/security gh/gpg gpg 
1:  55be3699b = 1:  55be3699b ncrypt/crypt_gpgme.c: Split conditional
2:  76180da27 ! 2:  f0a0767ca ncrypt/crypt_gpgme.c: Show encryption information
    @@ Commit message
         Link: <https://github.com/neomutt/neomutt/issues/4223>
         Co-developed-by: Richard Russon <rich@flatcap.org>
         Cc: Werner Koch <wk@gnupg.org>
    +    Cc: наб <nabijaczleweli@nabijaczleweli.xyz>
     
      ## ncrypt/crypt_gpgme.c ##
    @@ ncrypt/crypt_gpgme.c: static void print_smime_keyinfo(const char *msg, gpgme_sig
     +  if (!algo)
     +    algo = "?";
     +
    -+  state_puts(state, "Recipient: ");
    ++  state_puts(state, _("Recipient: "));
     +  state_puts(state, algo);
    -+  state_puts(state, " key, ID ");
    ++  state_puts(state, _(" key, ID "));
     +  state_puts(state, r->keyid);
     +  state_puts(state, "\n");
     +}
3:  582879f10 = 3:  4ba79c5a0 docs: Add feature page for the encryption information block
4:  7b29638ef = 4:  a60ecaa89 $crypt_encryption_info: Add variable
```

nabijaczleweli:
This is still wrong, it needs to be `state_printf(_("Recipient: %s key, ID %s\n"), algo, r->keyid)` and probably annotated with `// L10N: %s=RSA/ECDSA/..., %s=4A83A79DA571C41C`

alejandro-colomar:
> This is still wrong, it needs to be (if git grep isn't leading me astray) `state_printf(_("Recipient: %s key, ID %s\n"), algo, r->keyid)` and probably annotated with `// L10N: %s=RSA/ECDSA/..., %s=4A83A79DA571C41C`

Hmmm!  I didn't know of that function.  Yep, that looks much nicer.

alejandro-colomar:
v8 changes:

-  Compact into a single stat_printf() call, and comment L10N.  [@nabijaczleweli ]

```
$ git range-diff devel/security gh/gpg gpg 
1:  55be3699b = 1:  55be3699b ncrypt/crypt_gpgme.c: Split conditional
2:  f0a0767ca ! 2:  c4738ad68 ncrypt/crypt_gpgme.c: Show encryption information
    @@ ncrypt/crypt_gpgme.c: static void print_smime_keyinfo(const char *msg, gpgme_sig
     +  if (!algo)
     +    algo = "?";
     +
    -+  state_puts(state, _("Recipient: "));
    -+  state_puts(state, algo);
    -+  state_puts(state, _(" key, ID "));
    -+  state_puts(state, r->keyid);
    -+  state_puts(state, "\n");
    ++  // L10N: %s (algo) = RSA/ECDSA/..., %s (r->keyid) = 1111111111111111
    ++  state_printf(state, _("Recipient: %s key, ID %s\n"), algo, r->keyid);
     +}
     +
     +/**
3:  4ba79c5a0 = 3:  d9f1eecb3 docs: Add feature page for the encryption information block
4:  a60ecaa89 = 4:  c268086e6 $crypt_encryption_info: Add variable
```

alejandro-colomar:
Sorry @nabijaczleweli , I forgot to remove your `Reviewed-by` after I changed the code (even if after your suggestion; I might have messed it accidentally).  Can you please confirm the tag?

alejandro-colomar:
v8b (by @flatcap ) changes:

-  Rebase

```
$ git range-diff devel/security..gpg neomutt/devel/security..gh/gpg 
1:  55be3699b = 1:  db58c6335 ncrypt/crypt_gpgme.c: Split conditional
2:  c4738ad68 = 2:  4df67334a ncrypt/crypt_gpgme.c: Show encryption information
3:  d9f1eecb3 = 3:  4d4644abd docs: Add feature page for the encryption information block
4:  c268086e6 = 4:  3066f1f63 $crypt_encryption_info: Add variable
```

ncrypt/crypt_gpgme.c
```diff
@@ -1409,6 +1409,39 @@ static void print_smime_keyinfo(const char *msg, gpgme_signature_t sig,
...
+ * @param state State to write to
+ * @param r     Encryption recipient
+ */
+static void show_one_recipient(struct State *state, gpgme_recipient_t r)
+{
+  const char *algo = gpgme_pubkey_algo_name(r->pubkey_algo);
+  if (!algo)
+    algo = "?";
+
+  // L10N: %s (algo) = RSA/ECDSA/..., %s (r->keyid) = 1111111111111111
```
flatcap, line 1423:
```suggestion
  // L10N: Show the algorithm and key ID of the encryption recipients, e.g
  //       Recipient: RSA key, ID 1111111111111111
```

flatcap (changes requested):

alejandro-colomar:
v9 changes:

-  Fixup with @flatcap 's suggested patch for L10N comment.

```
$ git range-diff devel/security gh/gpg gpg 
1:  db58c6335 = 1:  db58c6335 ncrypt/crypt_gpgme.c: Split conditional
2:  4df67334a ! 2:  640f64465 ncrypt/crypt_gpgme.c: Show encryption information
    @@ ncrypt/crypt_gpgme.c: static void print_smime_keyinfo(const char *msg, gpgme_sig
     +  if (!algo)
     +    algo = "?";
     +
    -+  // L10N: %s (algo) = RSA/ECDSA/..., %s (r->keyid) = 1111111111111111
    ++  // L10N: Show the algorithm and key ID of the encryption recipients, e.g
    ++  //       Recipient: RSA key, ID 1111111111111111
     +  state_printf(state, _("Recipient: %s key, ID %s\n"), algo, r->keyid);
     +}
     +
3:  4d4644abd = 3:  6fc8c64fd docs: Add feature page for the encryption information block
4:  3066f1f63 = 4:  c63186d66 $crypt_encryption_info: Add variable
5:  4cdf7c3d0 < -:  --------- ncrypt/crypt_gpgme.c: @flatcap's suggested comment
```

alejandro-colomar:
v9b changes:

-  Rebase

```
$ git range-diff main..gh/gpg devel/security..gpg 
1:  db58c6335 = 1:  149652003 ncrypt/crypt_gpgme.c: Split conditional
2:  640f64465 = 2:  18fff6841 ncrypt/crypt_gpgme.c: Show encryption information
3:  6fc8c64fd = 3:  9e50d8358 docs: Add feature page for the encryption information block
4:  c63186d66 = 4:  c95200e83 $crypt_encryption_info: Add variable
```

flatcap:
- rebased over `[devel/security]`

flatcap (approved):

alejandro-colomar:
Thanks!  :-)

PR #4225, opened 2024-03-31
Show '@' character instead of number of attachments.

auouymous:
Adds an `%X@` expando to show '@' when one or more attachments, and a space when no attachments.

> All builds and tests are passing

No, but `test_mutt_file_stat_compare` fails without the patch.

Should this have a test, if so, where should I put it?


flatcap:
Thanks for PR, @auouymous.
Thanks for taking a dive into the exciting new expando code :-)

However...

What you're describing is already possible with **conditional expandos** (even before the expando rewrite).

```sh
# (less-than percent X question-mark at ampersand space larger-than)
set index_format = "%<X?@& >"
```

If `X` is non-zero, then the `@` is displayed, if `X` is zero, the ` ` (space) is shown.

You can even be fancy, if you have Nerd Fonts installed.
This example displays a paperclip:

```sh
set index_format = "%<X?📎& >"
```

**Note**: Some emoji-like characters take up two columns, so you'd need two spaces for alignment.

**See**:
- Conditionals: https://neomutt.org/guide/configuration.html#formatstrings-conditionals
- Nerd Fonts: https://www.nerdfonts.com/

auouymous:
Thank you, that is much simpler.

PR #4250, opened 2024-04-13
imap: add mailboxes more directly

flatcap:
Shortcut IMAP's workflow when adding a Mailbox.

When `$imap_check_subscribed` is set, NeoMutt sends an `LSUB` command to the server.
The server responds with a list of "subscribed" Mailboxes.

The IMAP code used to create a fake `mailboxes NAME` command.
Now it skips that step and calls `mailbox_add()` (via a wrapper).

This removes the need to quote the Mailbox name, only for `parse_rc_line()` to dequote it.

---

- 35dd92fca create `mailbox_add_simple()` wrapper
- de6354027 imap: add mailboxes more directly

### DIFF
--- a/commands.c
+++ b/commands.c
@@ -718,6 +718,19 @@ static enum CommandResult mailbox_add(const char *folder, const char *mailbox,
 }
 
 /**
+ * mailbox_add_simple - Add a new Mailbox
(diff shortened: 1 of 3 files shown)


roccoblues (approved):
nice one 👏🏼

Issue #4259, opened 2024-04-18
index_format: conditional right justification does not work


ernix:
@flatcap @cyphar `index_format` looked broke again.  See: https://github.com/neomutt/neomutt/issues/1074

## Expected Behaviour

~~~
N F  4/18 13:42 To kojima@inter (0.1K) hello
~~~

## Actual Behaviour

~~~
N F  4/18 13:42 To kojima@inter (0.1K) hello- [#00 ]>
~~~

## Steps to Reproduce

~~~
set index_format = "%Z %[%b/%d %H:%M] %-15.15L %?y?[%4.4y]&(%4c)? %s%<M? %>- [#%03M]>"
~~~

## How often does this happen?

- Always

## When did it start to happen?

- When I upgraded from version `20231006`

## NeoMutt Version

```
NeoMutt 20240416-525-b01e1f
Copyright (C) 2015-2024 Richard Russon and friends
NeoMutt comes with ABSOLUTELY NO WARRANTY; for details type 'neomutt -vv'.
NeoMutt is free software, and you are welcome to redistribute it
under certain conditions; type 'neomutt -vv' for details.

System: Darwin 22.6.0 (arm64)
ncurses: ncurses 6.4.20221231 (compiled with 6.4.20221231)
libiconv: 1.17
libidn2: 2.3.7 (compiled with 2.3.7)
GPGME: 1.23.2
OpenSSL: OpenSSL 3.2.1 30 Jan 2024
libnotmuch: 5.6.0
PCRE2: 10.43 2024-02-16
storage: tokyocabinet, lmdb
compression: zlib

Configure options: --sysconfdir=/opt/homebrew/etc --autocrypt --gss --disable-idn --idn2 --lmdb --nls --notmuch --pcre2 --sasl --sqlite --tokyocabinet --zlib --with-idn2=/opt/homebrew/opt/libidn2 --with-lua=/opt/homebrew/opt/lua --with-ncurses=/opt/homebrew/opt/ncurses --with-ssl=/opt/homebrew/opt/openssl@3 --with-sqlite=/opt/homebrew/opt/sqlite --with-iconv=/opt/homebrew/opt/libiconv --disable-debug --disable-dependency-tracking --prefix=/opt/homebrew/Cellar/neomutt/HEAD-b01e1fd --libdir=/opt/homebrew/Cellar/neomutt/HEAD-b01e1fd/lib

[...]

Issue #4258, opened 2024-04-17
save-hook only works once?


martonbalazs:
## Expected Behaviour

Did
`:save-hook .* =foo`
which changes the default mailbox to save to "foo" as expected.

Then did
`:save-hook .* =bar`
hoping that the default mailbox now becomes "bar".

## Actual Behaviour
Instead, it stays "foo". I.e., save-hook only seems to work once.

## How often does this happen?

- Always

## When did it start to happen?

- When I upgraded from 20240329-2 to 20240416-1 (on Arch)
## NeoMutt Version

```
NeoMutt 20240416
Copyright (C) 2015-2024 Richard Russon and friends
NeoMutt comes with ABSOLUTELY NO WARRANTY; for details type 'neomutt -vv'.
NeoMutt is free software, and you are welcome to redistribute it
under certain conditions; type 'neomutt -vv' for details.

System: Linux 6.6.27-1-lts (x86_64)
ncurses: ncurses 6.4.20230520 (compiled with 6.4.20230520)
libidn2: 2.3.7 (compiled with 2.3.7)
GPGME: 1.23.2
GnuTLS: 3.8.5
libnotmuch: 5.6.0
storage: kyotocabinet, gdbm, lmdb
compression: lz4, zlib, zstd

Configure options: --prefix=/usr --sysconfdir=/etc --libexecdir=/usr/lib --gpgme --sqlite --autocrypt --lua --notmuch --gss --gnutls --sasl --with-ui=ncurses --with-idn2=/usr --disable-idn --idn2 --lmdb --kyotocabinet --gdbm --ubsan --lz4 --zlib --zstd

[...]

PR #4261, opened 2024-04-20
expando: fixes

flatcap:
This PR fixes a few problems with the Expandos in the recent release.

- df6ee63d5 add lower-case operator
  Adding underscore before an expando letter, lowercases the field,
  e.g. `%30_s`

- b52ac1d27 fix conditional padding
  Using hard-padding `%>` in a conditional expando `%<...>` caused problems

- a6a503e8c fix container
  Some formats weren't handled correctly.
  The container node is used by the combined tree-subject in the Index

- df3c01124 add external filter
  Adding a pipe character (`|`) causes the field to be executed.
  e.g. `set sidebar_format = "sidebar.sh '%B'|"`

- e7b8e36e7 test: expando filter
  Unit tests

- 9bc780b69 enable external filter
  Enable the expando filter for lots of format strings.
 
Fixes: #4258 
Fixes: #4259

### DIFF
--- a/Makefile.autosetup
+++ b/Makefile.autosetup
@@ -430,12 +430,12 @@ $(PWD)/envelope:
 ###############################################################################
 # libexpando
 LIBEXPANDO=	libexpando.a
-LIBEXPANDOOBJS=	expando/config_type.o expando/expando.o expando/format.o \
-		expando/helpers.o expando/node.o expando/node_condbool.o \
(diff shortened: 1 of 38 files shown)


flatcap:
I should point out that both the "lower-case" and "filter" features existed in NeoMutt before the Expando upgrade.

PR #4268, opened 2024-04-23
Offer a way to filter out CFLAGS with paths from the output of `-v`

### LINKED ISSUE #4265: Embedding CFLAGS make neomutt unreproducible
## Expected Behaviour
I'd love to see neomutt [reproducible](https://reproducible-builds.org/) across different environments.

## Actual Behaviour
Right now, neomutt isn't reproducible because CFLAGS contains -ffile-prefix-map as shown below.

```diff
--g -O2 -Werror=implicit-function-declaration -ffile-prefix-map=/tmp/reprotest.qf9VDY/const_build_path/const_build_path=. -fstack-protector-strong -fstack-clash-protection -Wformat -Werror=format-security -fcf-protection -fno-delete-null-pointer-checks -D_ALL_SOURCE=1 -D_GNU_SOURCE=1 -D__EXTENSIONS__ -D_XOPEN_SOURCE_EXTENDED -I/usr/include/lua5.4 -I/usr/include -I/usr/include -I/usr/include -DNCURSES_WIDECHAR -I/usr/include -I/usr/include/p11-kit-1 -I/usr/include -isystem /usr/include/mit-krb5 -O2
+-g -O2 -Werror=implicit-function-declaration -ffile-prefix-map=/tmp/reprotest.qf9VDY/build-experiment-1/build-experiment-1=. -fstack-protector-strong -fstack-clash-protection -Wformat -Werror=format-security -fcf-protection -fno-delete-null-pointer-checks -D_ALL_SOURCE=1 -D_GNU_SOURCE=1 -D__EXTENSIONS__ -D_XOPEN_SOURCE_EXTENDED -I/usr/include/lua5.4 -I/usr/include -I/usr/include -I/usr/include -DNCURSES_WIDECHAR -I/usr/include -I/usr/include/p11-kit-1 -I/usr/include -isystem /usr/include/mit-krb5 -O2
```

This was caught in our (Debian) CI which tests two consecutive builds with different environments.

## Steps to Reproduce
Two consecutive builds from different directories should do the trick. Although there are more differences in neomutt's binary, though I think all of them might be a consequence of the changed CFLAGS. I'll attach the diff here, but it's really overwhelming.

## Possible Solution
I think a possible solution is to filter `-ffile-prefix-map` from the `cflags` before setting `cc_cflags` variable. The problem is I'm absolutely unversed in TCL so I can't send a PR to fix it. My instinct tells me something like `[lkill [get-define CFLAGS] {{x} {string equal -length 17 $x {-ffile-prefix-map}}}]` would do in `auto.def` `set conststrings`. Though I'm not sure if it should be default or have a variable to turn it on and off like `want-include-path-in-cflags`.
[reprotest_log.txt](https://github.com/neomutt/neomutt/files/15056455/reprotest_log.txt)

## NeoMutt Version

```
NeoMutt 20240416
Copyright (C) 2015-2024 Richard Russon and friends

System: Linux 6.6.15-amd64 (x86_64)
ncurses: ncurses 6.4.20240414 (compiled with 6.4.20240414)
libidn2: 2.3.4 (compiled with 2.3.7)
GPGME: 1.18.0
GnuTLS: 3.8.5
libnotmuch: 5.6.0
PCRE2: 10.42 2022-12-11
storage: tokyocabinet, lmdb
compression: lz4, zlib, zstd

Configure options: --build=x86_64-linux-gnu --prefix=/usr {--includedir=${prefix}/include} {--mandir=${prefix}/share/man} {--infodir=${prefix}/share/info} --sysconfdir=/etc --localstatedir=/var --disable-option-checking --disable-silent-rules {--libdir=${prefix}/lib/x86_64-linux-gnu} --runstatedir=/run --disable-maintainer-mode --disable-dependency-tracking --mandir=/usr/share/man --libexecdir=/usr/libexec --with-mailpath=/var/mail --autocrypt --full-doc --gnutls --gpgme --gsasl --gss --lmdb --lua --lz4 --notmuch --mixmaster --pcre2 --sqlite --tokyocabinet --zlib --zstd

Compilation CFLAGS: -g -O2 -Werror=implicit-function-declaration -ffile-prefix-map=/build/neomutt-6CRW6J/neomutt-20240416+dfsg=. -fstack-protector-strong -fstack-clash-protection -Wformat -Werror=format-security -fcf-protection -fno-delete-null-pointer-checks -D_ALL_SOURCE=1 -D_GNU_SOURCE=1 -D__EXTENSIONS__ -D_XOPEN_SOURCE_EXTENDED -I/usr/include/lua5.4 -I/usr/include -I/usr/include -I/usr/include -DNCURSES_WIDECHAR -I/usr/include -I/usr/include/p11-kit-1 -I/usr/include -isystem /usr/include/mit-krb5 -O2

Compile options:
  +gsasl +gss +hcache -homespool +idn +inotify -locales_hack +lua +mixmaster
  +nls +notmuch -openssl +pcre2 +pgp -sasl +smime +sqlite +truecolor

MAILPATH="/var/mail"
MIXMASTER="mixmaster"
SENDMAIL="/usr/sbin/sendmail"
SYSCONFDIR="/etc"

```

## Extra Info

* Operating System and its version: Debian unstable
* Were you using 'screen' or 'tmux'? Yes, tmux.
* Is your email local (maildir) or remote (IMAP)? Not important for this one (but IMAP :- )

gahr:
This adds --disable-paths-in-cflags, and makes --include-path-in-cflags deprecated. The new option will allow us to add more cflags to be filtered out.

Fixes #4265

### DIFF
--- a/auto.def
+++ b/auto.def
@@ -109,7 +109,7 @@ set valid_options {
   with-sysroot:path         => "Target system root"
   with-tmpdir:=/tmp         => "location of the tmp directory"
 # Misc
-  include-path-in-cflags=1  => "Remove include paths from CFLAGS in the output of neomutt -v"
+  paths-in-cflags=1         => "Remove paths from CFLAGS in the output of neomutt -v"
 # Testing
   asan=0                    => "Enable the Address Sanitizer"
   ubsan=0                   => "Enable the Undefined Behaviour Sanitizer"
@@ -142,6 +142,7 @@ set deprecated_options  {
   with-gpgme:
   with-lz4:
   with-zstd:
+  include-path-in-cflags:=
 }
 
 foreach dep $deprecated_options {
@@ -175,9 +176,10 @@ if {1} {
     asan autocrypt bdb compile-commands coverage debug-backtrace debug-color
     debug-email debug-graphviz debug-keymap debug-logging debug-names
     debug-notify debug-queue debug-window doc everything fmemopen full-doc
-    fuzzing gdbm gnutls gpgme gsasl gss homespool idn2 include-path-in-cflags
-    inotify kyotocabinet lmdb locales-fix lua lz4 mixmaster nls notmuch pcre2
-    pgp qdbm rocksdb sasl smime sqlite ssl tdb tokyocabinet ubsan zlib zstd
+    fuzzing gdbm gnutls gpgme gsasl gss homespool idn2 inotify kyotocabinet
+    lmdb locales-fix lua lz4 mixmaster nls notmuch paths-in-cflags pcre2 pgp
+    qdbm rocksdb sasl smime sqlite ssl tdb
+    tokyocabinet ubsan zlib zstd
   } {
     define want-$opt [opt-bool $opt]
   }
@@ -1150,11 +1152,22 @@ if {[get-define debug_build]} {
 
 ###############################################################################
 # Generate conststrings.c
+proc remove-paths-from-cflags {} {
+  lkill [get-define CFLAGS] {{x} {
+    foreach cflag {{-I} {-ffile-prefix-map=}} {
+      if {[string equal -length [string length $cflag] $x $cflag]} {
+        return 1
+      }
+    }
+    return 0
+  }}
+}
+
 set conststrings "\
   unsigned char cc_cflags\[\] = {[text2c [expr {
-    [get-define want-include-path-in-cflags]
+    [get-define want-paths-in-cflags]
     ? [get-define CFLAGS]
-    : [lkill [get-define CFLAGS] {{x} {string equal -length 2 $x {-I}}}]
+    : [remove-paths-from-cflags]
   }]]};\n\
   unsigned char configure_options\[\] = {[text2c $conf_options]};\n"
 if {[catch {set fd [open conststrings.c w]


ossilator:
these really should not be conflated into one option, as they serve entirely different purposes.
in fact, stripping -ffile-prefix-map doesn't even need to be an option, as it's about build meta, not build configuration.

gahr:
> these really should not be conflated into one option, as they serve entirely different purposes.

Can you please list the two different purposes?

ossilator:
-ffile-prefix-map is for making relocatable reproducible builds.
i have no clue what --disable-paths-in-cflags is for (check history?), but not _that_.

flatcap:
Original option: `--disable-include-path-in-cflags`:
- Issue: #2366
- PR: #2367
- Commit: 0dae5a39

gahr:
Yeah - would any distro want one but not the other?

ossilator:
ehm, how about properly answering the question instead of just proceeding?

did you notice that the requests are from two fundamentally different distros?

there is no reason why _anyone_ wouldn't want to filter -ffile-prefix-map, because recording it in the binary defeats the very purpose of this option.

filtering include paths is a completely different thing, and i don't think distros other than the requesting ones will be particularly happy about it (if they notice at all, given that this is a somewhat minor issue).

Issue #4271, opened 2024-04-24
Newline in attribution no longer properly handled


jkorb:
## Expected Behaviour

I have a newline character "\n" in my string for ```attribution```. This used to be printed as a newline in reply messages, but since the latest upgrade it prints as a "?".

## Actual Behaviour

When including a "\n" in the string for ```attribution```, a "?" is printed instead.

## Steps to Reproduce

1. Put the following into muttrc

```
set   attribution = "On %d, %n wrote:\n"
```

2. Create a reply message. The relevant line is printed as something like:

```
On 24.04.2024 16:38, Jon Doe wrote:?
```

## How often does this happen?

- Always

## When did it start to happen?

- When I upgraded
  Which version did you use to use?

neomutt-20240329-2

Downgrading fixes the issue.

I tried the neomutt-git AUR package, which is on 250e6180f342db1d95700c641c8433af7c0ca80c and the bug persists.

## NeoMutt Version

```
NeoMutt 20240329
Copyright (C) 2015-2024 Richard Russon and friends
NeoMutt comes with ABSOLUTELY NO WARRANTY; for details type 'neomutt -vv'.
NeoMutt is free software, and you are welcome to redistribute it
under certain conditions; type 'neomutt -vv' for details.

System: Linux 6.8.7-arch1-1 (x86_64)
ncurses: ncurses 6.4.20230520 (compiled with 6.4.20230520)
libidn2: 2.3.7 (compiled with 2.3.7)
GPGME: 1.23.2
GnuTLS: 3.8.4
libnotmuch: 5.6.0
storage: kyotocabinet, gdbm, lmdb
compression: lz4, zlib, zstd

Configure options: --prefix=/usr --sysconfdir=/etc --libexecdir=/usr/lib --gpgme --sqlite --autocrypt --lua --notmuch --gss --gnutls --sasl --with-ui=ncurses --with-idn2=/usr --disable-idn --idn2 --lmdb --kyotocabinet --gdbm --ubsan --lz4 --zlib --zstd

[...]

PR #4272, opened 2024-04-24
expando: allow newline

flatcap:
Fix expandos to allow newlines `\n`.
Newlines were being treated as invalid and replaced with `?`s.

Some format strings, e.g. `$attribution_intro` need to allow `\n`.

e.g.
```sh
# Intro with blank line after
set attribution_intro = "On %d, %n wrote:\n"
```

Fixes: #4271

### DIFF
--- a/expando/format.c
+++ b/expando/format.c
@@ -143,6 +143,10 @@ int format_string(struct Buffer *buf, int min_cols, int max_cols, enum FormatJus
     {
       w = 1; /* hack */
     }
+    else if (iswspace(wc))
+    {
+      w = 1;
+    }
     else
     {
 #ifdef HAVE_ISWBLANK


gahr (approved):

Yutsuten:
> Fix expandos to allow newlines \n.
> Newlines were being treated as invalid and replaced with `?`s.

Something that I remembered when seeing this, in my e-mail listing (index), some subject names shows those `?`.

```
Google Payments              | ???      ? ? ?    ?  ?  ?  ?   ??Google Cloud Platform & APIs: 001112-001112-001112 の請求書の用意ができました
```

It is the expand of

```
set index_format = '%-28.28L | %s'
```

This is the raw data when seeing in the browser (copy-pasted from gmail):

```
Subject: 

 

Google Cloud Platform & APIs: 001112-001112-001112 の請求書の用意ができました
From: Google Payments <payments-noreply@google.com>
```

gmail-subject-bug

(some data hidden for privacy)

`%s` (and Google) is the culprit. I wonder if this PR fixed this issue.

Sorry for reporting it here. If after the update the problem persists, I'll create a new issue. (Don't know if it was reported already though, I'll take a look)

flatcap:
> `???      ? ? ?    ?  ?  ?  ?   ??` ...

That's probably what we want to display.
Those whitespace formatting characters really shouldn't be in the subject.
Perhaps we could use a replacement char of ` ` (space) instead, though.

> I wonder if this PR fixed this issue.

Not sure.
Hopefully, I haven't made things worse for spam emails.

Please can you send me _that_ subject -- cut'n'paste it exactly into a new email.

- `rich@flatcap.org`

Thanks

PR #4282, opened 2024-04-30
mutt/string.c: Report an error and exit on ENOMEM

alejandro-colomar:
* **What does this PR do?**

Report an error and exit on ENOMEM, instead of silently using a NULL.

### DIFF
--- a/mutt/signal.c
+++ b/mutt/signal.c
@@ -32,6 +32,7 @@
 #include <stdbool.h>
 #include <stdio.h>
 #include <stdlib.h>
+#include <unistd.h>
 #include "message.h"
 #include "signal2.h"
 
@@ -63,6 +64,41 @@ static sig_handler_t SegvHandler = mutt_sig_exit_handler;
 volatile sig_atomic_t SigInt;   ///< true after SIGINT is received
 volatile sig_atomic_t SigWinch; ///< true after SIGWINCH is received
 
+static void exit_print_int_recursive(int n)
+{
+  char digit;
+
+  if (n > 9)
+    exit_print_int_recursive(n / 10);
+
+  digit = '0' + (n % 10);
+  write(1, &digit, 1);
+}
+
+static void exit_print_int(int n)
+{
+  if (n < 0)
+  {
+    write(1, "-", 1);
+    n = -n;
+  }
+  exit_print_int_recursive(n);
+}
+
+static void exit_print_string(const char *str)
+{
+  size_t len = 0;
+
+  if (!str)
+    return;
+
+  while (str[len])
+    len++;
+
+  if (len > 0)
+    write(1, str, len);
+}
+
 /**
  * mutt_sig_empty_handler - Dummy signal handler
  * @param sig Signal number, e.g. SIGINT
@@ -80,15 +116,17 @@ void mutt_sig_empty_handler(int sig)
  */
 void mutt_sig_exit_handler(int sig)
 {
+  exit_print_string("Caught signal ");
+  exit_print_int(sig);
+  exit_print_string(" ");
 #ifdef HAVE_DECL_SYS_SIGLIST
-  printf(_("Caught signal %d (%s) ...  Exiting\n"), sig, sys_siglist[sig]);
+  exit_print_string(sys_siglist[sig]);
 #elif (defined(__sun__) && defined(__svr4__))
-  printf(_("Caught signal %d (%s) ...  Exiting\n"), sig, _sys_siglist[sig]);
+  exit_print_string(_sys_siglist[sig]);
 #elif (defined(__alpha) && defined(__osf__))
-  printf(_("Caught signal %d (%s) ...  Exiting\n"), sig, __sys_siglist[sig]);
-#else
-  printf(_("Caught signal %d ...  Exiting\n"), sig);
+  exit_print_string(__sys_siglist[sig]);
 #endif
+  exit_print_string("...  Exiting\n");
   exit(0);
 }
 
--- a/mutt/signal.c
+++ b/mutt/signal.c
@@ -32,6 +32,7 @@
 #include <stdbool.h>
 #include <stdio.h>
 #include <stdlib.h>
+#include <string.h>
 #include <unistd.h>
 #include "message.h"
 #include "signal2.h"
@@ -85,18 +86,16 @@ static void exit_print_int(int n)
   exit_print_int_recursive(n);
 }
 
+/**
+ * exit_print_string - AS-safe version of printf("%s", str)
+ * @param str String to be printed
+ */
 static void exit_print_string(const char *str)
 {
-  size_t len = 0;
-
   if (!str)
     return;
 
-  while (str[len])
-    len++;
-
-  if (len > 0)
-    write(1, str, len);
+  write(STDOUT_FILENO, str, strlen(str));
 }
 
 /**
--- a/mutt/signal.c
+++ b/mutt/signal.c
@@ -65,25 +65,33 @@ static sig_handler_t SegvHandler = mutt_sig_exit_handler;
 volatile sig_atomic_t SigInt;   ///< true after SIGINT is received
 volatile sig_atomic_t SigWinch; ///< true after SIGWINCH is received
 
-static void exit_print_int_recursive(int n)
+/**
+ * exit_print_uint - AS-safe version of printf("%u", n)
+ * @param n Number to be printed
+ */
+static void exit_print_uint(unsigned int n)
 {
   char digit;
 
   if (n > 9)
-    exit_print_int_recursive(n / 10);
+    exit_print_uint(n / 10);
 
   digit = '0' + (n % 10);
-  write(1, &digit, 1);
+  write(STDOUT_FILENO, &digit, 1);
 }
 
+/**
+ * exit_print_int - AS-safe version of printf("%d", n)
+ * @param n Number to be printed
+ */
 static void exit_print_int(int n)
 {
   if (n < 0)
   {
-    write(1, "-", 1);
+    write(STDOUT_FILENO, "-", 1);
     n = -n;
   }
-  exit_print_int_recursive(n);
+  exit_print_uint(n);
 }
 
 /**
--- a/mutt/string.c
+++ b/mutt/string.c
@@ -31,6 +31,7 @@
 
 #include "config.h"
 #include <ctype.h>
+#include <errno.h>
 #include <stdarg.h>
 #include <stdbool.h>
 #include <stdio.h>
@@ -258,8 +259,8 @@ char *mutt_str_dup(const char *str)
   char *p = strdup(str);
   if (!p)
   {
-    mutt_error(_("Out of memory")); // LCOV_EXCL_LINE
-    mutt_exit(1);                   // LCOV_EXCL_LINE
+    mutt_error("%s", strerror(errno)); // LCOV_EXCL_LINE
+    mutt_exit(1);                      // LCOV_EXCL_LINE
   }
   return p;
 }
@@ -784,8 +785,8 @@ int mutt_str_asprintf(char **strp, const char *fmt, ...)
    * is undefined when the return code is -1.  */
   if (n < 0)
   {
-    mutt_error(_("Out of memory")); /* LCOV_EXCL_LINE */
-    mutt_exit(1);                   /* LCOV_EXCL_LINE */
+    mutt_error("%s", strerror(errno)); /* LCOV_EXCL_LINE */
+    mutt_exit(1);                      /* LCOV_EXCL_LINE */
   }
 
   if (n == 0)
--- a/mutt/string.c
+++ b/mutt/string.c
@@ -255,7 +255,13 @@ char *mutt_str_dup(const char *str)
   if (!str || (*str == '\0'))
     return NULL;
 
-  return strdup(str);
+  char *p = strdup(str);
+  if (!p)
+  {
+    mutt_error(_("Out of memory")); // LCOV_EXCL_LINE
+    mutt_exit(1);                   // LCOV_EXCL_LINE
+  }
+  return p;
 }
 
 /**
--- a/core/neomutt.c
+++ b/core/neomutt.c
@@ -29,6 +29,7 @@
 
 #include "config.h"
 #include <stddef.h>
+#include <errno.h>
 #include "mutt/lib.h"
 #include "config/lib.h"
 #include "neomutt.h"
@@ -61,8 +62,8 @@ struct NeoMutt *neomutt_new(struct ConfigSet *cs)
 
   if (!n->time_c_locale)
   {
-    mutt_error(_("Out of memory")); // LCOV_EXCL_LINE
-    mutt_exit(1);                   // LCOV_EXCL_LINE
+    mutt_error("%s", strerror(errno)); // LCOV_EXCL_LINE
+    mutt_exit(1);                      // LCOV_EXCL_LINE
   }
 
   n->notify_timeout = notify_new();
--- a/imap/msn.c
+++ b/imap/msn.c
@@ -28,6 +28,7 @@
  */
 
 #include "config.h"
+#include <errno.h>
 #include <limits.h>
 #include <stdlib.h>
 #include "mutt/lib.h"
@@ -46,7 +47,7 @@ void imap_msn_reserve(struct MSNArray *msn, size_t num)
    * if msn_count is this big, we have a serious problem. */
   if (num >= (UINT_MAX / sizeof(struct Email *)))
   {
-    mutt_error(_("Out of memory"));
+    mutt_error("%s", strerror(ENOMEM));
     mutt_exit(1);
   }
 
--- a/mutt/buffer.c
+++ b/mutt/buffer.c
@@ -35,6 +35,7 @@
  */
 
 #include "config.h"
+#include <errno.h>
 #include <stdarg.h>
 #include <stdbool.h>
 #include <stdint.h>
@@ -100,7 +101,7 @@ size_t buf_addstr_n(struct Buffer *buf, const char *s, size_t len)
   if (len > (SIZE_MAX - BufferStepSize))
   {
     // LCOV_EXCL_START
-    mutt_error(_("Out of memory"));
+    mutt_error("%s", strerror(ENOMEM));
     mutt_exit(1);
     // LCOV_EXCL_STOP
   }
--- a/mutt/memory.c
+++ b/mutt/memory.c
@@ -30,7 +30,9 @@
  */
 
 #include "config.h"
+#include <errno.h>
 #include <stdlib.h>
+#include <string.h>
 #include "memory.h"
 #include "exit.h"
 #include "logging2.h"
@@ -55,8 +57,8 @@ void *mutt_mem_calloc(size_t nmemb, size_t size)
   void *p = calloc(nmemb, size);
   if (!p)
   {
-    mutt_error(_("Out of memory")); // LCOV_EXCL_LINE
-    mutt_exit(1);                   // LCOV_EXCL_LINE
+    mutt_error("%s", strerror(errno)); // LCOV_EXCL_LINE
+    mutt_exit(1);                      // LCOV_EXCL_LINE
   }
   return p;
 }
@@ -95,8 +97,8 @@ void *mutt_mem_malloc(size_t size)
   void *p = malloc(size);
   if (!p)
   {
-    mutt_error(_("Out of memory")); // LCOV_EXCL_LINE
-    mutt_exit(1);                   // LCOV_EXCL_LINE
+    mutt_error("%s", strerror(errno)); // LCOV_EXCL_LINE
+    mutt_exit(1);                      // LCOV_EXCL_LINE
   }
   return p;
 }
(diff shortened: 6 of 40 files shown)


gahr (approved):

alejandro-colomar:
```
$ git range-diff neomutt/main gh/enomem enomem 
1:  a7af434e4 ! 1:  abaed74c0 mutt/string.c: Report an error and exit on ENOMEM
    @@ Metadata
      ## Commit message ##
         mutt/string.c: Report an error and exit on ENOMEM
     
    +    Reviewed-by: Pietro Cerutti <gahr@gahr.ch>
     
      ## mutt/string.c ##
```

mutt/string.c
```diff
@@ -255,7 +255,13 @@ char *mutt_str_dup(const char *str)
   if (!str || (*str == '\0'))
     return NULL;
 
-  return strdup(str);
+  char *p = strdup(str);
+  if (!p)
+  {
+    mutt_error(_("Out of memory")); // LCOV_EXCL_LINE
+    mutt_exit(1);                   // LCOV_EXCL_LINE
```
alejandro-colomar, line 262: Maybe we should use

```
mutt_error("%s", strerror(errno));
```

instead of having our own string.

alejandro-colomar: @flatcap , I copied that from

```c
$ grepc mutt_mem_calloc .
./mutt/memory.c:void *mutt_mem_calloc(size_t nmemb, size_t size)
{
  if ((nmemb == 0) || (size == 0))
    return NULL;

  void *p = calloc(nmemb, size);
  if (!p)
  {
    mutt_error(_("Out of memory")); // LCOV_EXCL_LINE
    mutt_exit(1);                   // LCOV_EXCL_LINE
  }
  return p;
}
./mutt/memory.h:void *mutt_mem_calloc(size_t nmemb, size_t size);
```

Should I fix both to use *strerror*(3)?  That would reduce localization effort.

alejandro-colomar:
v2 changes:

-  Use strerror(3) instead of `"Out of memory"`.
-  Update translations.

```
$ git range-diff neomutt/main abaed74c0 enomem 
1:  abaed74c0 = 1:  abaed74c0 mutt/string.c: Report an error and exit on ENOMEM
-:  --------- > 2:  18e14035e Use strerror(3) instead of a custom error message
-:  --------- > 3:  cd1abf1bf po/: Remove translations of "Out of memory"
```

alejandro-colomar:
v2b changes:

-  Add missing include

```
$ git range-diff neomutt/main gh/enomem enomem 
1:  abaed74c0 = 1:  abaed74c0 mutt/string.c: Report an error and exit on ENOMEM
2:  18e14035e ! 2:  7c80c23be Use strerror(3) instead of a custom error message
    @@ core/neomutt.c: struct NeoMutt *neomutt_new(struct ConfigSet *cs)
        n->notify_timeout = notify_new();
     
      ## imap/msn.c ##
    +@@
    +  */
    + 
    + #include "config.h"
    ++#include <errno.h>
    + #include <limits.h>
    + #include <stdlib.h>
    + #include "mutt/lib.h"
     @@ imap/msn.c: void imap_msn_reserve(struct MSNArray *msn, size_t num)
         * if msn_count is this big, we have a serious problem. */
        if (num >= (UINT_MAX / sizeof(struct Email *)))
3:  cd1abf1bf = 3:  e46f4ece6 po/: Remove translations of "Out of memory"
```

alejandro-colomar:
v2c changes:

-  Add missing includes

```
$ git range-diff neomutt/main gh/enomem enomem 
1:  abaed74c0 = 1:  abaed74c0 mutt/string.c: Report an error and exit on ENOMEM
2:  7c80c23be ! 2:  69b40abe5 Use strerror(3) instead of a custom error message
    @@ Commit message
     
      ## core/neomutt.c ##
    +@@
    +  */
    + 
    + #include "config.h"
    ++#include <errno.h>
    + #include <stddef.h>
    + #include "mutt/lib.h"
    + #include "config/lib.h"
     @@ core/neomutt.c: struct NeoMutt *neomutt_new(struct ConfigSet *cs)
      
        if (!n->time_c_locale)
    @@ imap/msn.c: void imap_msn_reserve(struct MSNArray *msn, size_t num)
      
     
      ## mutt/buffer.c ##
    +@@
    +  */
    + 
    + #include "config.h"
    ++#include <errno.h>
    + #include <stdarg.h>
    + #include <stdbool.h>
    + #include <stdint.h>
     @@ mutt/buffer.c: size_t buf_addstr_n(struct Buffer *buf, const char *s, size_t len)
        if (len > (SIZE_MAX - BufferStepSize))
        {
    @@ mutt/buffer.c: size_t buf_addstr_n(struct Buffer *buf, const char *s, size_t len
        }
     
      ## mutt/memory.c ##
    +@@
    +  */
    + 
    + #include "config.h"
    ++#include <errno.h>
    + #include <stdlib.h>
    + #include "memory.h"
    + #include "exit.h"
     @@ mutt/memory.c: void *mutt_mem_calloc(size_t nmemb, size_t size)
        void *p = calloc(nmemb, size);
        if (!p)
    @@ mutt/memory.c: void mutt_mem_realloc(void *ptr, size_t size)
        *p = r;
     
      ## mutt/string.c ##
    +@@
    + 
    + #include "config.h"
    + #include <ctype.h>
    ++#include <errno.h>
    + #include <stdarg.h>
    + #include <stdbool.h>
    + #include <stdio.h>
     @@ mutt/string.c: char *mutt_str_dup(const char *str)
        char *p = strdup(str);
        if (!p)
3:  e46f4ece6 = 3:  e78a695a6 po/: Remove translations of "Out of memory"
```

mutt/string.c
flatcap: > Should I fix both to use _strerror_(3)?

Good idea.  Yes, please.

FYI.
Upstream have taken an interesting route.
Worried about re-entrancy during the apocalyse, they avoid `gettext()` and `printf()`.
They save the translation on startup and use `write(2)` to display it.

alejandro-colomar: Could you please point to specific commits of upstream?  I'd like to see exactly what they did.

flatcap: A quick grep shows:

alejandro-colomar: Thanks!  It looks a bit complex... I'll check some day.  Maybe we could track it in a bug report.

alejandro-colomar:
v3 changes:

-  Cherry-pick a patch from mutt(1)  [@flatcap ]

```
$ git range-diff neomutt/main gh/enomem enomem 
1:  abaed74c0 = 1:  abaed74c0 mutt/string.c: Report an error and exit on ENOMEM
2:  69b40abe5 = 2:  69b40abe5 Use strerror(3) instead of a custom error message
3:  e78a695a6 = 3:  e78a695a6 po/: Remove translations of "Out of memory"
-:  --------- > 4:  7c358694d Remove nonreentrant calls from mutt_sig_exit_handler()
```

It may be a bit off-topic for this PR, but since you mentioned it...

I haven't yet tried cherry-picking the others.

flatcap:
> I haven't yet tried cherry-picking the others.

FYI.  The complete list of outstanding upstream patches is here:
- https://github.com/neomutt/upstream-patches/blob/main/todo.txt

There are about 200 of them of the >2000 commits since the fork.
- ~30 are to do with background editing
  The \*\*\*\* that Mutt does to support this feature discourages me :-)
- ~30 to do with colours
  Our colour support is already better, but there are some things we could improve.

I'd love to merge more of them, but they're quite time-consuming to work on.

test/acutest.h
```diff
@@ -1830,7 +1830,7 @@ main(int argc, char** argv)
 
     acutest_test_data_ = (struct acutest_test_data_*)calloc(acutest_list_size_, sizeof(struct acutest_test_data_));
     if(acutest_test_data_ == NULL) {
-        fprintf(stderr, "Out of memory.\n");
+        fprintf(stderr, "%s\n", strerror(errno));
```
flatcap, line 1833: This is 3rd-party source -- https://github.com/mity/acutest
Please revert.

mutt/signal.c
```diff
@@ -63,6 +64,41 @@ static sig_handler_t SegvHandler = mutt_sig_exit_handler;
...
+
+  if (!str)
+    return;
+
+  while (str[len])
+    len++;
+
+  if (len > 0)
+    write(1, str, len);
+}
```
flatcap, line 100: This all seems like overkill to me :-)
If you'd like to keep it, please can you add Doxygen comments.
Thanks.

alejandro-colomar: If I would have written that, I'd have simply called strlen(3).  I just cherry-picked the upstream commit as is.

By overkill, do you mean the entire commit?  Or the way it's implemented?

test/acutest.h
alejandro-colomar: Done.

alejandro-colomar:
v4 changes:

-  Drop change to 3rd party code  [@flatcap ]

```
$ git range-diff neomutt/main gh/enomem enomem 
1:  abaed74c0 = 1:  abaed74c0 mutt/string.c: Report an error and exit on ENOMEM
2:  69b40abe5 ! 2:  f13a1b3da Use strerror(3) instead of a custom error message
    @@ mx.c: void mx_alloc_memory(struct Mailbox *m, int req_size)
          mutt_exit(1);
        }
      
    -
    - ## test/acutest.h ##
    -@@ test/acutest.h: main(int argc, char** argv)
    - 
    -     acutest_test_data_ = (struct acutest_test_data_*)calloc(acutest_list_size_, sizeof(struct acutest_test_data_));
    -     if(acutest_test_data_ == NULL) {
    --        fprintf(stderr, "Out of memory.\n");
    -+        fprintf(stderr, "%s\n", strerror(errno));
    -         acutest_exit_(2);
    -     }
    - 
3:  e78a695a6 = 3:  1c52d1efe po/: Remove translations of "Out of memory"
4:  7c358694d = 4:  66fd8fae7 Remove nonreentrant calls from mutt_sig_exit_handler()
```

mutt/signal.c
alejandro-colomar: BTW, it would be interesting if gcvt(3) was wide spread; we could just call it.  :(

flatcap: > overkill

A lot of extra code for a situation that "shouldn't happen" :-)

It's the correct thing to do; we should do it.
It's just a shame C can't do it more concisely.

alejandro-colomar: Yup.  If gcvt(3) was present in the BSDs, we could use that...  :/

alejandro-colomar:
v5 changes

-  Remove period (upstream has slightly different strings).
-  Simplify the two AS-safe functions and add Doxygen comments.  [@flatcap ]

```
$ git range-diff neomutt/main gh/enomem enomem 
1:  abaed74c0 = 1:  abaed74c0 mutt/string.c: Report an error and exit on ENOMEM
2:  f13a1b3da = 2:  f13a1b3da Use strerror(3) instead of a custom error message
3:  1c52d1efe = 3:  1c52d1efe po/: Remove translations of "Out of memory"
4:  66fd8fae7 ! 4:  5ee509805 Remove nonreentrant calls from mutt_sig_exit_handler()
    @@ mutt/signal.c: void mutt_sig_empty_handler(int sig)
     -  printf(_("Caught signal %d ...  Exiting\n"), sig);
     +  exit_print_string(__sys_siglist[sig]);
      #endif
    -+  exit_print_string("...  Exiting.\n");
    ++  exit_print_string("...  Exiting\n");
        exit(0);
      }
      
-:  --------- > 5:  de4225547 mutt/signal.c: Simplify exit_print_string()
-:  --------- > 6:  8aae2aeb9 mutt/signal.c: Simplify exit_print_int()
```

alejandro-colomar:
@flatcap 

Please double-check the simplifications.  I think I did it correctly, but these trivial things I always get them twisted.  :-)

mutt/signal.c
```diff
@@ -63,6 +64,40 @@ static sig_handler_t SegvHandler = mutt_sig_exit_handler;
...
 
+/**
+ * exit_print_int - AS-safe version of printf("%d", n)
+ * @param n Number to be printed
+ */
+static void exit_print_int(int n)
+{
+  if (n < 0)
+  {
+    write(1, "-", 1);
```
flatcap, line 75:
```suggestion
    write(STDOUT_FILENO, "-", 1);
```

mutt/signal.c
```diff
@@ -63,6 +64,40 @@ static sig_handler_t SegvHandler = mutt_sig_exit_handler;
...
+    n = -n;
+  }
+
+  for (/*void*/; n > 9; n / 10)
```
flatcap, line 79:
```suggestion
  for (; n > 9; n = n / 10)
```

We already have lots of `for (; ...`
`n / 10` didn't _do_ anything.

flatcap (changes requested):
Looks good...
hehe, except it prints the numbers backwards 🤦🏻

mutt/signal.c
alejandro-colomar: D'oh!  Thanks!

alejandro-colomar:
> Looks good... hehe, except it prints the numbers backwards 🤦🏻

Hehe; maybe it's not worth writing a loop.  The recursive implementation seems simpler.  To reverse it I'd need a buffer.

flatcap:
> The recursive implementation seems simpler

Agreed.  Nice try, though :-)

alejandro-colomar:
v6b changes:

-  Add missing include

```
$ git range-diff neomutt/main gh/enomem enomem 
1:  abaed74c0 = 1:  abaed74c0 mutt/string.c: Report an error and exit on ENOMEM
2:  f13a1b3da = 2:  f13a1b3da Use strerror(3) instead of a custom error message
3:  1c52d1efe = 3:  1c52d1efe po/: Remove translations of "Out of memory"
4:  5ee509805 = 4:  5ee509805 Remove nonreentrant calls from mutt_sig_exit_handler()
5:  de4225547 ! 5:  492a79627 mutt/signal.c: Simplify exit_print_string()
    @@ Commit message
     
      ## mutt/signal.c ##
    +@@
    + #include <stdbool.h>
    + #include <stdio.h>
    + #include <stdlib.h>
    ++#include <string.h>
    + #include <unistd.h>
    + #include "message.h"
    + #include "signal2.h"
     @@ mutt/signal.c: static void exit_print_int(int n)
        exit_print_int_recursive(n);
      }
6:  56dbb8497 = 6:  8fdd60f42 mutt/signal.c: Implement exit_print_int() in terms of exit_print_uint()
```

alejandro-colomar:
dprintf(3) seems to be AS-safe by design, according to <https://www.openwall.com/lists/musl/2024/05/01/6>.  I'll confirm this and maybe we can just use that, which would be a lot simpler.

flatcap:
> dprintf(3) seems to be AS-safe by design

It's not mentioned in [signal-safety(7)](https://man7.org/linux/man-pages/man7/signal-safety.7.html) yet,
so let's not take up any more time and stick with what we've got.

alejandro-colomar:
> > dprintf(3) seems to be AS-safe by design
> 
> It's not mentioned in [signal-safety(7)](https://man7.org/linux/man-pages/man7/signal-safety.7.html) yet, so let's not take up any more time and stick with what we've got.

Yep, let's stick with what we've got.  At most, I'll patch it in the future.

For reference, here's the discussion the Rich Felker talked about:
<https://inbox.sourceware.org/libc-alpha/20130925180327.0351F2C097@topped-with-meat.com/>
and
<https://inbox.sourceware.org/libc-alpha/20130925212954.GQ20515@brightrain.aerifal.cx/>

So, originally, the function was designed to be AS-safe, but then it changed (at least in 2013 it wasn't any more).  I've opened a mail thread with glibc to find the current status, so that I can document it, and to see if they can make it AS-safe again.

flatcap (approved):

alejandro-colomar:
Thanks!

alejandro-colomar:
@flatcap 

I see that you reordered an include during the merge.  I used alphabetic order; is there a different order scheme in this project?

BTW, thanks for adding the missing <string.h>!

```
$ git range-diff main..enomem 74d60ade8..gh/enomem 
1:  abaed74c0 = 1:  1e1611297 mutt/string.c: Report an error and exit on ENOMEM
2:  f13a1b3da ! 2:  3f5b54abe Use strerror(3) instead of a custom error message
    @@ Commit message
     
      ## core/neomutt.c ##
     @@
    -  */
      
      #include "config.h"
    -+#include <errno.h>
      #include <stddef.h>
    ++#include <errno.h>
      #include "mutt/lib.h"
      #include "config/lib.h"
    + #include "neomutt.h"
     @@ core/neomutt.c: struct NeoMutt *neomutt_new(struct ConfigSet *cs)
      
        if (!n->time_c_locale)
    @@ mutt/memory.c
      #include "config.h"
     +#include <errno.h>
      #include <stdlib.h>
    ++#include <string.h>
      #include "memory.h"
      #include "exit.h"
    + #include "logging2.h"
     @@ mutt/memory.c: void *mutt_mem_calloc(size_t nmemb, size_t size)
        void *p = calloc(nmemb, size);
        if (!p)
3:  1c52d1efe = 3:  fb2a4ee66 po/: Remove translations of "Out of memory"
4:  5ee509805 = 4:  5107a6c8e Remove nonreentrant calls from mutt_sig_exit_handler()
5:  492a79627 = 5:  0dfc3fd66 mutt/signal.c: Simplify exit_print_string()
6:  8fdd60f42 = 6:  ac5e8e65c mutt/signal.c: Implement exit_print_int() in terms of exit_print_uint()
```


BTW, I see that a few other files are missing it:
```sh
$ grep -rl strerror | grep '\.c$' | xargs grep -L '<string\.h>'
conn/gsasl.c
imap/auth_gsasl.c
imap/msn.c
send/smtp.c
store/lmdb.c
core/neomutt.c
autocrypt/gpgme.c
```

Of which one seems to be my fault: `imap/msn.c`.

```
$ git show --stat 3f5b54abe
commit 3f5b54abe07843702ebcc3f2c14f774f4dd4c1a3
Author: Alejandro Colomar (@alejandro-colomar) <alx@kernel.org>
Date:   Tue Apr 30 17:55:10 2024 +0200

    Use strerror(3) instead of a custom error message
    
 core/neomutt.c |  5 +++--
 imap/msn.c     |  3 ++-
 mutt/buffer.c  |  3 ++-
 mutt/memory.c  | 14 ++++++++------
 mutt/string.c  |  9 +++++----
 mx.c           |  2 +-
 6 files changed, 21 insertions(+), 15 deletions(-)
```

```diff
$ git show 3f5b54abe -- imap/msn.c
commit 3f5b54abe07843702ebcc3f2c14f774f4dd4c1a3
Author: Alejandro Colomar (@alejandro-colomar) <alx@kernel.org>
Date:   Tue Apr 30 17:55:10 2024 +0200

    Use strerror(3) instead of a custom error message
    
    Signed-off-by: Alejandro Colomar <alx@kernel.org>

diff --git a/imap/msn.c b/imap/msn.c
index 11223981d..5b3d3b040 100644
--- a/imap/msn.c
+++ b/imap/msn.c
@@ -28,6 +28,7 @@
  */
 
 #include "config.h"
+#include <errno.h>
 #include <limits.h>
 #include <stdlib.h>
 #include "mutt/lib.h"
@@ -46,7 +47,7 @@ void imap_msn_reserve(struct MSNArray *msn, size_t num)
    * if msn_count is this big, we have a serious problem. */
   if (num >= (UINT_MAX / sizeof(struct Email *)))
   {
-    mutt_error(_("Out of memory"));
+    mutt_error("%s", strerror(ENOMEM));
     mutt_exit(1);
   }
 
```

I'll prepare a patch for those missing includes.

alejandro-colomar:
Hmmm, I'm running iwyu (0.22) on the entire project, and will prepare corrections for libc includes.  I won't tackle internal includes in that.

flatcap:
> I see that you reordered an include during the merge

Yes, that was done by `clang-format`.
It dates back to ac6ef83f3 2017-04-15 fix: build of imap/auth_sasl.c
`<stddef.h>` was promoted to fix a build bug.

---

There are quite a few "magic" rules governing `#include`s.

First, all libraries should be including using their `"foo/lib.h"`
`iwyu` has `.imp` files defining these transforms: https://github.com/neomutt/iwyu/

Then `clang-format` has rules to fix the ordering.

- `"config.h"` and `"acutest.h"` always comes first
- `<header.h>` always come before `"header.h"`
- There are rules to put our headers in some dependency order,
  e.g. `"mutt/lib.h"` before others

Full ruleset:

https://github.com/neomutt/neomutt/blob/585158a5542cbe4c81d1ce43d0eec142bc76b9c8/.clang-format#L10-L61

---

Most of these rules date back a long way -- often to fix specific problems.
I'm open to suggestions, though.

flatcap:
> I'm running iwyu (0.22)

Hmm.. me too.

I frequently run `clang-format` and `iwyu` across the entire codebase,
so there are rarely many changes to apply...

But, recently I upgraded my distro and got new versions of all the tools.
Now, I see a lot of `iwyu` suggestions, but I haven't had a chance to review them.

> will prepare corrections for libc includes

Thanks!

alejandro-colomar:
> > I'm running iwyu (0.22)
> 
> Hmm.. me too.
> 
> I frequently run `clang-format` and `iwyu` across the entire codebase, so there are rarely many changes to apply...
> 
> But, recently I upgraded my distro and got new versions of all the tools. Now, I see a lot of `iwyu` suggestions, but I haven't had a chance to review them.

Most of them are probably my fault.  I turned iwyu(1) a little bit more strict with libc headers in the last versions.  :-)

Link: <https://github.com/include-what-you-use/include-what-you-use/pull/1283>
Link: <https://github.com/include-what-you-use/include-what-you-use/pull/1344>

> 
> > will prepare corrections for libc includes
> 
> Thanks!

:-)

alejandro-colomar:
> Most of these rules date back a long way -- often to fix specific problems. I'm open to suggestions, though.

I would drop the forward declarations in favor of libc includes.

Also, personally I'd put a blank separating include blocks, as suggested by google <https://google.github.io/styleguide/cppguide.html#Names_and_Order_of_Includes>.  However, given the lack of blank lines in the project in general, I guess that's the taste of maintainers here, so it's your preference.

PR #4287, opened 2024-05-05
hcache directory pollution

lucilanga:
Hcache directory get polluted by imap completion data.

To reproduce: have an imap dir with hcache enabled, type s to save mail to another mailbox (eg. mailbox)
type mailbox name but type it wrong eg. =maiz pres TAB for completion,
see maiz.hcache  and maiz.hache-lock created (8k each file)
same thing goes for each letter TAB press while completion is not complete.

* **What does this PR do?**

Add imap_adata_find_nocache() just for imap_completion() that basically is the same thing as imap_adata_find() without
any hcache calls.
The better approach would probably be to have a specific imap completion function that searches for a path.
However I'm unsure it's worth it.
There is another issue folded into this behavior, say when you want to save mail to an nonexistent mailbox type its name and say no to create mailbox dialog, you will end up with .hcache and .hcache-lock files once again however the impact is less visible.

### DIFF
--- a/imap/imap.c
+++ b/imap/imap.c
@@ -1283,7 +1283,7 @@ int imap_complete(struct Buffer *buf, const char *path)
   int completions = 0;
   int rc;
 
-  if (imap_adata_find(path, &adata, &mdata) < 0)
+  if (imap_adata_find_nocache(path, &adata, &mdata) < 0)
   {
     buf_strcpy(buf, path);
     return complete_hosts(buf);
--- a/imap/mdata.c
+++ b/imap/mdata.c
@@ -70,7 +70,7 @@ struct ImapMboxData *imap_mdata_get(struct Mailbox *m)
  * @param name  Name for Mailbox
  * @retval ptr New ImapMboxData
  */
-struct ImapMboxData *imap_mdata_new(struct ImapAccountData *adata, const char *name)
+struct ImapMboxData *imap_mdata_nocache_new(struct ImapAccountData *adata, const char *name)
 {
   char buf[1024] = { 0 };
   struct ImapMboxData *mdata = mutt_mem_calloc(1, sizeof(struct ImapMboxData));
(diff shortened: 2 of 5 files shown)


gahr:
Thanks, I'll take a closer look asap

gahr:
I suggest we enhance [`hcache_open`](https://github.com/neomutt/neomutt/blob/d7060ffc7268a47cc6c5fdeb4fa81806c56ca9dd/hcache/lib.h#L115-L124) and the storage [`open` API](https://github.com/neomutt/neomutt/blob/d7060ffc7268a47cc6c5fdeb4fa81806c56ca9dd/store/lib.h#L72-L85) that's [called from there](https://github.com/neomutt/neomutt/blob/d7060ffc7268a47cc6c5fdeb4fa81806c56ca9dd/hcache/hcache.c#L515) to take a flag to specify whether we want to create the file if it's not there. Here's how we could check that in the different implementations:

* bdb has an `O_CREAT` flag, https://github.com/neomutt/neomutt/blob/d7060ffc7268a47cc6c5fdeb4fa81806c56ca9dd/store/bdb.c#L126
* gdb has a `GDBM_WRCREAT` flag, https://github.com/neomutt/neomutt/blob/d7060ffc7268a47cc6c5fdeb4fa81806c56ca9dd/store/gdbm.c#L48
* kc has a `KCOCREATE` flag, https://github.com/neomutt/neomutt/blob/d7060ffc7268a47cc6c5fdeb4fa81806c56ca9dd/store/kc.c#L53
* lmdb uses `MDB_NOSUBDIR`, which means `path` is really the file path. We need to check if it exists ourselves, https://github.com/neomutt/neomutt/blob/d7060ffc7268a47cc6c5fdeb4fa81806c56ca9dd/store/lmdb.c#L165
* qdbm has a `VL_OCREAT` flag, https://github.com/neomutt/neomutt/blob/d7060ffc7268a47cc6c5fdeb4fa81806c56ca9dd/store/qdbm.c#L47
* rocksdb has an explicit option we call, https://github.com/neomutt/neomutt/blob/d7060ffc7268a47cc6c5fdeb4fa81806c56ca9dd/store/rocksdb.c#L88
* tc has a `BDBOCREAT` flag, https://github.com/neomutt/neomutt/blob/d7060ffc7268a47cc6c5fdeb4fa81806c56ca9dd/store/tc.c#L49
* tdb has a `O_CREAT` flag, https://github.com/neomutt/neomutt/blob/d7060ffc7268a47cc6c5fdeb4fa81806c56ca9dd/store/tdb.c#L53C61-L53C68

gahr:
I have started to work on an implementation at #4312. @lucilanga could you please check that branch and see if it fixes the issue for you?

lucilanga:
> @lucilanga could you please check that branch and see if it fixes the issue for you?

It does work fine. Thank you!

gahr:
> > @lucilanga could you please check that branch and see if it fixes the issue for you?
> 
> It does work fine. Thank you!

Thanks, I'm going to close this PR and continue working on that other one.

PR #4292, opened 2024-05-09
free(NULL) is valid C; it's a no-op

alejandro-colomar:
Simplify.

### DIFF
--- a/mutt/memory.c
+++ b/mutt/memory.c
@@ -71,11 +71,8 @@ void mutt_mem_free(void *ptr)
   if (!ptr)
     return;
   void **p = (void **) ptr;
-  if (*p)
-  {
-    free(*p);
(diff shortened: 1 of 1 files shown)


flatcap (approved):
Thanks!
That dates back to `safe_free()` in commit ONE :-)
https://github.com/neomutt/upstream-mutt/blob/1a5381e07e97fe482c2b3a7c75f99938f0b105d4/lib.c#L302-L309

alejandro-colomar:
> Thanks! That dates back to `safe_free()` in commit ONE :-) https://github.com/neomutt/upstream-mutt/blob/1a5381e07e97fe482c2b3a7c75f99938f0b105d4/lib.c#L302-L309

I thought of adding a Fixes tag for commit one, but I forgot  :P

PR #4296, opened 2024-05-09
Use wmem*() functions with wide-character strings

alejandro-colomar:
This avoids an explicit size multiplication, which can overflow the calculation.


flatcap (approved):
Lovely, thanks!

PR #4294, opened 2024-05-09
Add mutt_mem_reallocarray() and mutt_mem_mallocarray(), and some wrapper macros

alejandro-colomar:
These functions add safety to the nmemb * size multiplication.

I used them even in cases where the nmemb is 1, and used a sizeof(char) in that case, because I think that's more consistent, and also prepares the ground for using magic macros similar to the ones used in the shadow project.

Also add the magic macros and use them where appropriate.

<https://github.com/shadow-maint/shadow/blob/9b3889696b42d799dbd1b11db50e87b180c98c31/lib/alloc.h#L29>

Only in allocations of void, I kept the old _realloc() and _malloc(), since there's no underlying type, and thus, no size of an element.

Where the macros wouldn't work for some reason, I kept calls to the functions, but the macros are used wherever we can.

This requires C11 for _Generic(3).

---

Revisions:

<details>
<summary>v2</summary>

-  Fix a few accidents
-  Add some magic macros

```
$ git range-diff neomutt/main neomutt/alx/reallocarray alx/reallocarray 
 1:  bd2e52eb1 =  1:  bd2e52eb1 mutt/memory.[ch]: Add mutt_mem_reallocarray()
 2:  7ee0da20a !  2:  4f9013c2b Use mutt_mem_reallocarray()
    @@ ncrypt/crypt.c: static void crypt_fetch_signatures(struct Body ***b_sigs, struct
          {
            if ((*n % 5) == 0)
     -        mutt_mem_realloc(b_sigs, (*n + 6) * sizeof(struct Body **));
    -+        mutt_mem_reallocarray(b_sigs, *n + 6, sizeof(struct Body **));
    ++        mutt_mem_reallocarray(b_sigs, *n + 6, sizeof(struct Body *));
      
            (*b_sigs)[(*n)++] = b;
          }
 3:  5d6f2382f =  3:  de12f0ed7 mutt/memory.[ch]: Add mutt_mem_mallocarray()
 4:  be0492cbd !  4:  f1e069e8b Use mutt_mem_mallocarray()
    @@ hcache/serialize.c: void serial_restore_char(char **c, const unsigned char *d, i
        }
      
     -  *c = mutt_mem_malloc(size);
    -+  *c = mutt_mem_mallocarray(size, sizeof(char *));
    ++  *c = mutt_mem_mallocarray(size, sizeof(char));
        memcpy(*c, d + *off, size);
        if (convert && !mutt_str_is_ascii(*c, size))
        {
[...]

PR #4297, opened 2024-05-10
ncrypt/crypt.c: Fix allocation size calculation

alejandro-colomar:
Fixes: b423ebbfa9d2 ("Make the experimental branch the main trunk.")
Cc: @ossilator 
Cc: @gahr 
Cc: @flatcap 
Link: <https://github.com/neomutt/neomutt/pull/4294>

This was found thanks to the MUTT_MEM_REALLOC() macro, which caused a compiler error.

### DIFF
--- a/ncrypt/crypt.c
+++ b/ncrypt/crypt.c
@@ -1085,7 +1085,7 @@ static void crypt_fetch_signatures(struct Body ***b_sigs, struct Body *b, int *n
     else
     {
       if ((*n % 5) == 0)
-        mutt_mem_realloc(b_sigs, (*n + 6) * sizeof(struct Body **));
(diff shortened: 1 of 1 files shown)


flatcap (approved):
Wow, subtle.

> Fixes: b423ebbfa9d2679e64174fd58e775bc5b8124d3d

hehe

ossilator:
pedantically, you fixed only the optics, as a pointer to a pointer is still just a pointer size-wise. i.e., the commit message slightly over-sold the otherwise fine patch.

alejandro-colomar:
> pedantically, you fixed only the optics, as a pointer to a pointer is still just a pointer size-wise. i.e., the commit message slightly over-sold the otherwise fine patch.

The result happened to be correct, but the calculation that led to it was wrong conceptually, IMO.  :)

Issue #4276, opened 2024-04-25
Subject with non-printable characters or new line is shown with `?`


Yutsuten:
Continuation of the discussion in #4272

> [...] in my e-mail listing (index), some subject names shows those `?`.
> 
> ```
> Google Payments              | ???      ? ? ?    ?  ?  ?  ?   ??Google Cloud Platform & APIs: 001112-001112-001112 の請求書の用意ができました
> ```
> 
> It is the expand of
> 
> ```
> set index_format = '%-28.28L | %s'
> ```
> 
> This is the raw data when seeing in the browser (copy-pasted from gmail):
> 
> ```
> Subject: 
> 
>  
> 
> Google Cloud Platform & APIs: 001112-001112-001112 の請求書の用意ができました
> From: Google Payments <payments-noreply@google.com>
> ```
> 
> ![gmail-subject-bug](https://github.com/neomutt/neomutt/assets/7322925/526c55c4-daa8-4b43-b11d-0295b602c690)
> 
> (some data hidden for privacy)
> 
> `%s` (and Google) is the culprit. I wonder if this PR fixed this issue.

## Expected Behaviour

Subject text without `?`, like when we check the emails in the browser.

![gmail](https://github.com/neomutt/neomutt/assets/7322925/b7b99036-0fe0-4b27-b669-8a112ccf1c24)

## Actual Behaviour

Subject is shown as `???      ? ? ?    ?  ?  ?  ?   ??Google Cloud Platform & APIs: 001112-001112-001112 の請求書の用意ができました`.

> That's probably what we want to display.

Really? I'm confused by this answer in the PR.

## Steps to Reproduce

Not sure how to send an email like that, maybe only programmatically. I'll look into it.

## How often does this happen?

- On a particular email
  If the email isn't private, please attach it to this issue.

## When did it start to happen?

- Since I started using it, I guess.

## NeoMutt Version

```
NeoMutt 20240416
```

## Extra Info

* Operating System and its version: Linux 6.8.7-arch1-1 (x86_64)
* Were you using multiple copies of NeoMutt at once? No
* Were you using 'screen' or 'tmux'? No
* Is your email local (maildir) or remote (IMAP)? IMAP

I also have something to ask @flatcap 
I'd like to try fixing this issue myself. My main programming language is python, and only used C while in the university, about 6 years ago. Do you think I would be able to do it?

flatcap:
> Do you think I would be able to do it?

Sure :-)

[...]

PR #4298, opened 2024-05-11
Fix newlines in subject being shown with `?`

Yutsuten:
* **What does this PR do?**

1. Keeps `env->subject` with the original content, the previous code was replacing `LF` characters with `?`.
    1. Other fields may be affected, but I think that if any additional fixes become needed, it should be done in `format_string()` or similar
2. Sanitize `env->subject` into `env->disp_subj` by replacing whitespace characters (`iswspace()`) with a space ` ` character.

I also tested with and without a `subjectrx` configuration, seems to be working as expected.

There are also some side changes as discussed in #4276 

1. `#ifdef` not needed anymore in `expando/format.c`
> As `iswspace()` is a bigger set than `iswblank()`, we can drop `iswblank()`.
2. Return value of `subjrx_apply_mods` from `bool` to `void`. Return value is not used anywhere

Subject with newlines before the change:

```
???      ? ? ?    ?  ?  ?  ?   ??Google Cloud Platform & APIs...
```

Subject with newlines after the change:

```
                                 Google Cloud Platform & APIs...
```

- **What are the relevant issue numbers?** 
#4276 

---

This is my first PR with C language related changes. Please tell me if there are any security or performance issues in my code.

Thank you @flatcap for giving me so much information in the issue.

### DIFF
--- a/email/rfc2047.c
+++ b/email/rfc2047.c
@@ -330,7 +330,7 @@ static size_t choose_block(char *d, size_t dlen, int col, const char *fromcode,
 }
 
 /**
- * finalize_chunk - Perform charset conversion and filtering
+ * finalize_chunk - Perform charset conversion
  * @param[out] res        Buffer where the resulting string is appended
  * @param[in]  buf        Buffer with the input string
  * @param[in]  charset    Charset to use for the conversion
@@ -346,7 +346,6 @@ static void finalize_chunk(struct Buffer *res, struct Buffer *buf, char *charset
   charset[charsetlen] = '\0';
   mutt_ch_convert_string(&buf->data, charset, cc_charset(), MUTT_ICONV_HOOK_FROM);
   charset[charsetlen] = end;
-  mutt_mb_filter_unprintable(&buf->data);
   buf_addstr(res, buf->data);
   FREE(&buf->data);
   buf_init(buf);
--- a/expando/format.c
+++ b/expando/format.c
@@ -149,13 +149,8 @@ int format_string(struct Buffer *buf, int min_cols, int max_cols, enum FormatJus
     }
     else
     {
-#ifdef HAVE_ISWBLANK
-      if (iswblank(wc))
-        wc = ' ';
-      else
-#endif
-          if (!IsWPrint(wc))
-        wc = '?';
+      if (!IsWPrint(wc))
+        wc = ReplacementChar;
       w = wcwidth(wc);
     }
 
--- a/subjectrx.c
+++ b/subjectrx.c
@@ -27,6 +27,8 @@
  */
 
 #include "config.h"
+#include <errno.h>
+#include <limits.h>
 #include <stddef.h>
 #include <stdint.h>
 #include "mutt/lib.h"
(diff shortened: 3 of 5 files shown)


flatcap:
Thanks!
I shall have a read and a play as soon as I can.

Yutsuten:
> > I'll remove the `mutt_mb_filter_unprintable()` call from `finalize_chunk()`, hopefully this won't break anything. I think it's ok because filtering unprintable characters is the job of functions like `format_string()`.
>
> Be careful -- this has security implications because you don't want to send untrusted data (which email headers clearly are) to the terminal.

OK so I checked what `mutt_mb_filter_unprintable()` ([source](https://github.com/neomutt/neomutt/blob/devel/subject-bad-characters/mutt/mbyte.c#L423)) that I removed is doing. As far as I can understand, 3 things:

1. This check: `if ((k == ICONV_ILLEGAL_SEQ) || (k == ICONV_BUF_TOO_SMALL))`
2. Replacement of non white space characters: `if (!IsWPrint(wc)) wc = '?';`
3. And further check for bad utf8 characters: `else if (CharsetIsUtf8 && mutt_mb_is_display_corrupting_utf8(wc))`

The remaining of the code is just rebuilding the string as far as I can understand.

@flatcap have told me [here](https://github.com/neomutt/neomutt/issues/4276#issuecomment-2077149652) that:

> For the Index Subject Expando `%s` we finally arrive in `node_expando_render()`, where it calls `format_string()`.
> This function fits the data into the space available.

So if I understand correctly, before the subject can be printed to the terminal, it must pass `format_string()`. It is already doing (1) [at line 122](https://github.com/neomutt/neomutt/blob/devel/subject-bad-characters/expando/format.c#L122) and (2) [at line 152](https://github.com/neomutt/neomutt/blob/devel/subject-bad-characters/expando/format.c#L152).

But we are missing (3) `mutt_mb_is_display_corrupting_utf8()`. I think a change like this will work.

```diff
--- a/expando/format.c
+++ b/expando/format.c
@@ -129,7 +129,7 @@ int format_string(struct Buffer *buf, int min_cols, int max_cols, enum FormatJus
     }
 
     // How many screen cells will the character require?
-    if (escaped)
+    if (escaped || (CharsetIsUtf8 && mutt_mb_is_display_corrupting_utf8(wc)))
     {
       escaped = false;
       w = 0;
```

I'm adding this change to the PR.

With this I believe we'll have solved the possible security issue. Please someone tell me if I'm missing anything.

Yutsuten:
By the way, would we be able to fix the issue with the `U+3000 : IDEOGRAPHIC SPACE` character [reported here](https://github.com/neomutt/neomutt/issues/4276#issuecomment-2080446412) by updating `mutt_mb_is_display_corrupting_utf8()`?

```diff
--- a/mutt/mbyte.c
+++ b/mutt/mbyte.c
@@ -387,6 +387,7 @@ bool mutt_mb_is_display_corrupting_utf8(wchar_t wc)
   if ((wc == (wchar_t) 0x00ad) || /* soft hyphen */
       (wc == (wchar_t) 0x200e) || /* left-to-right mark */
       (wc == (wchar_t) 0x200f) || /* right-to-left mark */
+      (wc == (wchar_t) 0x3000) || /* ideographic space */
       (wc == (wchar_t) 0xfeff))   /* zero width no-break space */
   {
     return true;
```

For this one I'll wait for someone more experienced to tell me if it is ok.

Yutsuten:
Hope this PR gets reviewed eventually...

Yutsuten:
@flatcap if you think this change is too dangerous just say it, no need to pressure yourself to merge this and possibly have trouble later.

We could start with the `IDEOGRAPHIC SPACE` fix only or whatever, its okay. :+1:

PR #4305, opened 2024-05-13
remove mutt_expand_path() and mutt_expand_path_regex()

roccoblues:
This replaces the remaining uses of `mutt_expand_path()` with `buf_expand_path()` and removes the unused functions.

### DIFF
--- a/commands.c
+++ b/commands.c
@@ -1071,25 +1071,29 @@ static enum CommandResult parse_setenv(struct Buffer *buf, struct Buffer *s,
 static enum CommandResult parse_source(struct Buffer *buf, struct Buffer *s,
                                        intptr_t data, struct Buffer *err)
 {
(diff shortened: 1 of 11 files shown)


commands.c
```diff
@@ -1080,17 +1080,20 @@ static enum CommandResult parse_source(struct Buffer *buf, struct Buffer *s,
       buf_printf(err, _("source: error at %s"), s->dptr);
       return MUTT_CMD_ERROR;
```
flatcap, line 1081: We leak `path`, here

flatcap (changes requested):

commands.c
roccoblues: 🙈 fixed, thanks!

flatcap (approved):

PR #4307, opened 2024-05-13
build: require C11

flatcap:
Update NeoMutt's build requirements to C11.

- https://en.wikipedia.org/wiki/C11_(C_standard_revision)

This will allow us to use `_Generic(3)` in macros.
See also: #4294


gahr (approved):

alejandro-colomar:
Thanks!

PR #4308, opened 2024-05-13
fix -Wdouble-promotion warnings

roccoblues:
With this two fixes NeoMutt compiles for me with -Werror -Wdouble-promotion (`Apple clang version 15.0.0 (clang-1500.3.9.4)`).

### DIFF
--- a/conn/zstrm.c
+++ b/conn/zstrm.c
@@ -105,9 +105,9 @@ static int zstrm_close(struct Connection *conn)
 
   mutt_debug(LL_DEBUG5, "read %lu->%lu (%.1fx) wrote %lu<-%lu (%.1fx)\n",
              zctx->read.z.total_in, zctx->read.z.total_out,
-             (float) zctx->read.z.total_out / (float) zctx->read.z.total_in,
+             (double) zctx->read.z.total_out / (double) zctx->read.z.total_in,
              zctx->write.z.total_in, zctx->write.z.total_out,
-             (float) zctx->write.z.total_in / (float) zctx->write.z.total_out);
+             (double) zctx->write.z.total_in / (double) zctx->write.z.total_out);
 
   // Restore the Connection's original functions
   conn->sockdata = zctx->next_conn.sockdata;
--- a/send/sendlib.c
+++ b/send/sendlib.c
@@ -391,8 +391,8 @@ static void set_encoding(struct Body *b, struct Content *info, struct ConfigSubs
   else
   {
     /* Determine which encoding is smaller  */
-    if (1.33 * (float) (info->lobin + info->hibin + info->ascii) <
-        3.0 * (float) (info->lobin + info->hibin) + (float) info->ascii)
+    if (1.33f * (float) (info->lobin + info->hibin + info->ascii) <
+        3.0f * (float) (info->lobin + info->hibin) + (float) info->ascii)
     {
       b->encoding = ENC_BASE64;
     }


conn/zstrm.c
```diff
@@ -105,9 +105,9 @@ static int zstrm_close(struct Connection *conn)
 
   mutt_debug(LL_DEBUG5, "read %lu->%lu (%.1fx) wrote %lu<-%lu (%.1fx)\n",
              zctx->read.z.total_in, zctx->read.z.total_out,
-             (float) zctx->read.z.total_out / (float) zctx->read.z.total_in,
+             zctx->read.z.total_out / zctx->read.z.total_in,
              zctx->write.z.total_in, zctx->write.z.total_out,
-             (float) zctx->write.z.total_in / (float) zctx->write.z.total_out);
+             zctx->write.z.total_in / zctx->write.z.total_out);
```
flatcap, line 110: I get warnings about this now:
```
conn/zstrm.c:106:25: warning: format ‘%f’ expects argument of type ‘double’, but argument 5 has type ‘long unsigned int’
conn/zstrm.c:106:25: warning: format ‘%f’ expects argument of type ‘double’, but argument 8 has type ‘long unsigned int’
```

The fields: `zctx->read.z.total_in` and `zctx->read.z.total_out` are **`unsigned long`**.
(see `struct z_stream` in `zlib.h`)

flatcap: Casting to `(double)` seems to keep the compiler happy:
```c
mutt_debug(LL_DEBUG5, "read %lu->%lu (%.1fx) wrote %lu<-%lu (%.1fx)\n",
           zctx->read.z.total_in, zctx->read.z.total_out,
           (double) zctx->read.z.total_out / (double) zctx->read.z.total_in,
           zctx->write.z.total_in, zctx->write.z.total_out,
           (double) zctx->write.z.total_in / (double) zctx->write.z.total_out);
```

roccoblues: yeah makes sense, sorry. I wonder why this didn't show up for me. 🤔

I changed the cast to `double` now. But there's much more important stuff for you to look at than this cosmetic fixes. Feel free to just close this if it doesn't work.

flatcap: I probably have some `-Wformat-???` options enabled

flatcap (approved):

conn/zstrm.c
roccoblues: I have `-Wformat=2` which should enable all of them. 🤷🏼

conn/zstrm.c
```diff
@@ -105,9 +105,9 @@ static int zstrm_close(struct Connection *conn)
 
   mutt_debug(LL_DEBUG5, "read %lu->%lu (%.1fx) wrote %lu<-%lu (%.1fx)\n",
              zctx->read.z.total_in, zctx->read.z.total_out,
-             (float) zctx->read.z.total_out / (float) zctx->read.z.total_in,
+             (double) zctx->read.z.total_out / (double) zctx->read.z.total_in,
```
ossilator, line 108: i think the second cast can be omitted, as the second operand will be promoted automatically to match.

send/sendlib.c
```diff
@@ -391,8 +391,8 @@ static void set_encoding(struct Body *b, struct Content *info, struct ConfigSubs
   else
   {
     /* Determine which encoding is smaller  */
-    if (1.33 * (float) (info->lobin + info->hibin + info->ascii) <
-        3.0 * (float) (info->lobin + info->hibin) + (float) info->ascii)
+    if (1.33f * (float) (info->lobin + info->hibin + info->ascii) <
```
ossilator, line 394: same here.

PR #4310, opened 2024-05-14
scanf: initialise out-vars

flatcap:
`scanf()` returns values via pointers.
Ensure that all of them are initialised.

It seems to be the only difference between whether CodeQL warns us, or not.

### DIFF
--- a/history/history.c
+++ b/history/history.c
@@ -206,7 +206,7 @@ static void shrink_histfile(void)
 {
   FILE *fp_tmp = NULL;
   int n[HC_MAX] = { 0 };
-  int line, hclass, read;
+  int line, hclass = 0, read = 0;
   char *linebuf = NULL, *p = NULL;
   size_t buflen;
   bool regen_file = false;
@@ -607,7 +607,7 @@ void mutt_hist_read_file(void)
   if (!fp)
     return;
 
-  int line = 0, hclass, read;
+  int line = 0, hclass = 0, read = 0;
   char *linebuf = NULL, *p = NULL;
   size_t buflen;
 
--- a/imap/message.c
+++ b/imap/message.c
@@ -168,8 +168,8 @@ static int msg_cache_commit(struct Mailbox *m, struct Email *e)
  */
 static int imap_bcache_delete(const char *id, struct BodyCache *bcache, void *data)
 {
-  uint32_t uv;
-  unsigned int uid;
+  uint32_t uv = 0;
+  unsigned int uid = 0;
   struct ImapMboxData *mdata = data;
 
   if (sscanf(id, "%u-%u", &uv, &uid) != 2)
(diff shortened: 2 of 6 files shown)


ossilator:
i deeply dislike make-tool-happy changes, as they often prevent the detection of legitimate bugs by better tools. in this case, if there is a branch that genuinely forgets to set a variable, it would no longer be found.

flatcap:
I'm initialising variables -- it's hardly controversial.

There's no guarantee that `some_function(&x);` will set `x`.

gahr (approved):

ossilator:
> I'm initialising variables -- it's hardly controversial.

i don't.

> There's no guarantee that some_function(&x); will set x.

yes, and it's up to static and dynamic checkers to find if they aren't. this is a lot more reliable than initializing it to _something_. this is particularly relevant when variables are re-used.

alejandro-colomar:
Here's another example:

GCC doesn't warn, but other tools do warn:

```c
alx@debian:~/tmp/analyzer$ cat bug2.c 
#include <stdlib.h>

void
g(int *x)
{
	if (arc4random() % 2)
		*x = 42;
}

int
main(void)
{
	int x;

	g(&x);
	return x;  // maybe uninitialized use
}
```
```sh
alx@debian:~/tmp/analyzer$ clang-tidy ./bug2.c 
Error while trying to load a compilation database:
Could not auto-detect compilation database for file "./bug2.c"
No compilation database found in /home/alx/tmp/analyzer or any parent directory
fixed-compilation-database: Error while opening fixed database: No such file or directory
json-compilation-database: Error while opening JSON database: No such file or directory
Running without flags.
1 warning generated.
/home/alx/tmp/analyzer/bug2.c:16:2: warning: Undefined or garbage value returned to caller [clang-analyzer-core.uninitialized.UndefReturn]
        return x;  // maybe uninitialized use
        ^      ~
/home/alx/tmp/analyzer/bug2.c:13:2: note: 'x' declared without an initial value
        int x;
        ^~~~~
/home/alx/tmp/analyzer/bug2.c:15:2: note: Calling 'g'
        g(&x);
        ^~~~~
/home/alx/tmp/analyzer/bug2.c:6:6: note: Assuming the condition is false
        if (arc4random() % 2)
            ^~~~~~~~~~~~~~~~
/home/alx/tmp/analyzer/bug2.c:6:2: note: Taking false branch
        if (arc4random() % 2)
        ^
/home/alx/tmp/analyzer/bug2.c:8:1: note: Returning without writing to '*x'
}
^
/home/alx/tmp/analyzer/bug2.c:15:2: note: Returning from 'g'
        g(&x);
        ^~~~~
/home/alx/tmp/analyzer/bug2.c:16:2: note: Undefined or garbage value returned to caller
        return x;  // maybe uninitialized use
        ^      ~
```

alejandro-colomar:
> Here's another example:
> 
> GCC doesn't warn in this one, but other tools (clang analyzer) do warn:

It seems GCC can warn on that one too, although it requires an obscure flag:

<https://gcc.gnu.org/bugzilla/show_bug.cgi?id=115089#c3>

They'll probably improve that.

PR #4312, opened 2024-05-15
Allow opening the header cache in non-O_CREAT mode

gahr:
This enhances the header cache API so we can specify whether we want to create the cache file if it doesn't exist at `hcache_open` time. So far, I have only tried to come up with meaningful variations for IMAP. All other uses of the header cache always specify to create the file if it doesn't exist (current behaviour).

Fixes #4287

### DIFF
--- a/hcache/hcache.c
+++ b/hcache/hcache.c
@@ -468,7 +468,7 @@ static unsigned int generate_hcachever(void)
 /**
  * hcache_open - Multiplexor for StoreOps::open
  */
-struct HeaderCache *hcache_open(const char *path, const char *folder, hcache_namer_t namer)
+struct HeaderCache *hcache_open(const char *path, const char *folder, hcache_namer_t namer, bool create)
 {
   if (!path || (path[0] == '\0'))
     return NULL;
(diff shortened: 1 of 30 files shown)


flatcap:
That looks good (and nice and simple :-)
It's not going to throw an error, higher up, is it? -- if `create=false` and the file doesn't exist

gahr:
> That looks good (and nice and simple :-) It's not going to throw an error, higher up, is it? -- if `create=false` and the file doesn't exist

I have checked that, and I think we check for `mdata->hcache` not being `NULL` before using it. We'll see :)

gahr:
@flatcap are you ok with leaving the current behaviour for maildir/pop/nntp, until that becomes a problem for someone?

flatcap:
Oh!
I'm not too worried about POP/NNTP,
but if you can repeat the problem in Maildir, please can you fix it.
Thanks

gahr:
> but if you can repeat the problem in Maildir, please can you fix it.

I don't think you can: maildir (and other backends) use readdir and such to check if a folder exist.

flatcap:
OK, good.

PR #4326, opened 2024-06-11
ignore macro events in mw_yesorno()

### LINKED ISSUE #4316: autocrypt breaks command line argument `-e "push foo"`
## Expected Behaviour

I’m trying to run `neomutt -e "push foo"` (example). I expect to have  that in the status bar:

which corresponds to the "f" key and oo typed.

## Actual Behaviour

Nothing happens, and the "push foo" is completely ignored

## Steps to Reproduce

The config `set autocrypt = yes` is the one line that needs to be enabled in the neomutt config to reproduce the issue

- Always

When I upgraded: the issue appeared somewhere between 20230517 and 20231103. I’m still working on trying to figure out the exact release / commit

## Extra Info

* Operating System and its version: NixOS
* Were you using multiple copies of NeoMutt at once? no
* Were you using 'screen' or 'tmux'? yes (tmux)
* Is your email local (maildir) or remote (IMAP)? IMAP

roccoblues:
`mw_yesorno()` is used for asking the user a yes/no question and hopefully doesn't need to read macro events.

If we ignore them we can remove the flushing of the buffer from `mutt_autocrypt_init()` and with that fix #4316.

### DIFF
--- a/autocrypt/autocrypt.c
+++ b/autocrypt/autocrypt.c
@@ -106,8 +106,6 @@ int mutt_autocrypt_init(bool can_create)
   if (!c_autocrypt || !c_autocrypt_dir)
     return -1;
 
-  mutt_flushinp();
-
   if (autocrypt_dir_init(can_create))
     goto bail;
 
--- a/question/question.c
+++ b/question/question.c
@@ -249,7 +249,7 @@ static enum QuadOption mw_yesorno(const char *prompt, enum QuadOption def,
   window_redraw(NULL);
   while (true)
   {
-    event = mutt_getch(GETCH_NO_FLAGS);
+    event = mutt_getch(GETCH_IGNORE_MACRO);
     if ((event.op == OP_TIMEOUT) || (event.op == OP_REPAINT))
     {
       window_redraw(NULL);


immae:
tested locally, it fixes the said issue. Thanks!

gahr (approved):
I think I'm prepared to accept this w/o a 100% understanding of the consequences :)

roccoblues:
> I think I'm prepared to accept this w/o a 100% understanding of the consequences :)

I checked the callers once again and they all look interactive. I think we should be fine. And if not - its a small change that we can easily revert.

gahr:
@roccoblues : I'm sorry, I read the code, accepted it, but didn't test it. This breaks this macro of mine:

```
macro index S '<save-message>=Junk<enter>y<sync-mailbox>y' 'Move to Junk'
```

`<save-message>` asks for a confirmation because `confirm_append="yes"`.

I think I misunderstood this PR as saying: don't expand the responses to `mw_yesorno()` as macros, i.e., if you bind a macro to `y`, `y` still means yes in `mw_yesorno()` context. Now, I really don't know what it's doing.

roccoblues:
Oh damn, sorry about that. Feel free to revert it. 

> Now, I really don't know what it's doing.

What I understood:

`mutt_getch()` reads from multiple buffers:
```
 * The priority for reading events is:
 * 1. UngetKeyEvents buffer
 * 2. MacroEvents buffer
 * 3. Keyboard
 ```

This PR changes the read in `mw_yesorno()` to **not** read from the `MacroEvents` buffer. I assumed this was safe because why would you answer that question from a macro. Now we now why. :D 

I did the change because https://github.com/neomutt/neomutt/issues/4316#issuecomment-2161264183

> I think I understand what is happening:
autocrypt initialization can ask for user response ("/tmp/foo does not exist. Create it?"). To not use the provided command (-e "push foo") as input for the question the old code set/unset the global OptIgnoreMacroEvents.

Back to the drawing board.

**Comment by gahr** [comment]
> Back to the drawing board.

Guess so.. I'll try to put some more brain cycles into this, too. I'll revert your commit for now. Thanks!

**Comment by roccoblues** [comment]
Braindump from IRC:
```
      gahr  roccoblues: ouch: https://github.com/neomutt/neomutt/pull/4326#issuecomment-2212537146
roccoblues  gahr: sorry about that. I think the equivalent to the previous global would be to pass down GETCH_IGNORE_MACRO through       
            query_yesorno() -> mw_yesorno() ->  mutt_getch() and only set it in the autocrypt init.                                      
roccoblues  But that looks ugly.                                                                                                         
      [...]
      gahr  so autocrypt init contains code that triggers a prompt ("/tmp/foo does not exist. Create it?")                               
      gahr  but we don't want the prompt?                                                                                                
      gahr  or we don't want the prompt only in the case where there are characters in the input buffer already?                         
roccoblues  we always want the prompt                                                                                                    
roccoblues  but we don't want to use what is passed as cli arg (-e 'push foo') as input for the prompt                                   
      [...]
      gahr  so what if I wanted to answer y to the autocrypt prompt by ./neomutt -e 'push y' ?                                           
      gahr  I think this is as valid a usecase as any                                                                                    
      gahr  i.e., expecting a question and providing an answer to it on the cli                                                         
roccoblues  well then we can just remove the flushing of the input buffer                                                                
roccoblues  but that changes the existing behavior                                                                                       
roccoblues  when we still had the global we explicitly turned it off                                                                     
      [...]
      gahr  originally, with the global, we'd turn out reading from the input buffer just for that particular prompt?                    
      gahr  then we would resume reading from it?                                                                                        
roccoblues  yes, turn macros off, prompt, turn on again                                                                                  
      gahr  then I guess I agree with you                                                                                                
      gahr  we need to pass the option down                                                                                              
      [...]
      gahr  I think I don't agree with the design of autocrypt being special                                                             
      gahr  if you run NeoMutt on a fresh system it'll ask about creating ~/Mail or something                                            
roccoblues  fair, also the prompt only happens during the first init                                                                     
      [...]
      gahr  it's a change in behaviour for sure, probably a breaking one, but I think the current code is more coherent                  
      [...]
      gahr  I have the feeling that #3998 changed the thing at the wrong layer                                                           
   neo-bot  PR by @flatcap "eliminate global OptIgnoreMacroEvents": https://github.com/neomutt/neomutt/pull/3998                         
      gahr  i.e., it should have passed the option down as you were suggesting                                                           
roccoblues  yeah, it definitely changed the behavior                                                                                     
roccoblues  I'm wondering about this comment: "The init process can display menus at various points (e.g. browser, pgp key selection)."  
roccoblues  not sure if we want to read macro events for that ...                                                                        
roccoblues  https://github.com/neomutt/neomutt/commit/3b2790cb5c808e71ed7978f54b4fbb15de77a289                                           
      gahr  The muttrc can push events into the macro buffer, with the assumption                                                        
      gahr  that mailbox will be opened next.                                                                                            
      gahr  I think this assumption is wrong                                                                                             
      gahr  *I* can push events with -e w/o any assumption, but with knowledge :)                                                        
      gahr  of course, if you push events expecting one thing and another thing happens, you're screwed                                  
roccoblues  true                                                                                                                         
      gahr  perhaps there are two types of prompts: i) those initiated by a user event, e.g., change-folder, and ii) those initiated by  
            the system, e.g., the autocrypt thing and the "create ~/Mail?" thing                                                         
      gahr  although I'm pretty sure that the                                                                                            
      gahr  although I'm pretty sure that the "create ~/Mail" comes from changing folder into the initial $folder                        
      gahr  so that's nuanced, too 
```

PR #4329, opened 2024-06-12
remove mixmaster

roccoblues:
The project seems more or less dead:

* @vuori on IRC: "Their sourceforge project's last activity was in 2015, there's a github fork that was active in 2017 and that's about it."
* Jörg Sommer to [neomutt-users](https://mailman.neomutt.org/pipermail/neomutt-users-neomutt.org/2024-May/001394.html): "I've tried mixmaster with the last nodes I have in my list, but the e-mails didn't reach any recipients."


gahr (approved):

gahr:
Let's wait for @flatcap on this one, perhaps?

flatcap (approved):

PR #4330, opened 2024-06-13
Dont' mix Content-ID with parameters

### LINKED ISSUE #4327: RFC2231-style MIME part header continuations erroneously used for long Content-ID fields
On NeoMutt 20240425 (Debian unstable), so relatively recent, I just encountered the following corner-case bug:

When composing a MIME message and the Content-ID of a MIME part I enter exceeds a certain length, NeoMutt splits the Content ID across multiple lines, according to RFC2231, but RFC2231 only applies to parameters, not fields.

This is what a normal MIME part header with Content-ID looks like:

```
--_004_b5b53e4f78ff4f1d88c062752ba7d24falarmkittelde_
Content-Type: image/jpeg; name="image001.jpg"
Content-Disposition: inline; filename="image001.jpg"
Content-ID: <image001.jpg@01DABB52.AA1ADAD0>
```

If the Content-ID value exceeds a certain length, it gets broken across lines:

```
--_004_b5b53e4f78ff4f1d88c062752ba7d24falarmkittelde_
Content-Type: image/jpeg; name="image001.jpg"
Content-Disposition: inline; filename="image001.jpg"
Content-ID: =?utf-8?q?=3Cimage001=2Ejpg=4001DasvasdvABasdvasBasdv52=2EAAasdv?=
 =?utf-8?q?asdvas1asdvasdvaAsdvaDsdvasAasdvasdvasdvdvasdvDas0dv=3E?=
```

Note how the angle brackets are preserved in the encoding (`=3C` and `=3E`).

When I assign such a long content ID to a part using NeoMutt, however, it seems to think it needs to invoke RFC2231, and makes the Content-ID field a parameter of Content-Type, which is wrong:

```
--_004_b5b53e4f78ff4f1d88c062752ba7d24falarmkittelde_
Content-Type: image/jpeg; name="image001.jpg";
        content-id*0="image001.jpg@01DasvasdvABasdvasBasdv52.AAasdvasdvas1asdvasd";
        content-id*1="vaAsdvaDsdvasAasdvasdvasdvdvasdvDas0dv"
Content-Disposition: inline; filename="image001.jpg"
```

In doing so, it also loses the angle brackets, but that doesn't really matter anymore, as the MIME part is now misformed.

gahr:
Fixes #4327

This moves `Content-ID` into its own dedicated place instead of reusing the container for `Content-Type` parameters, so we stop applying to it (`Content-ID`) what we should apply to them (`Content-Type` parameters), e.g., RF2231 splitting. See https://github.com/neomutt/neomutt/issues/4327#issuecomment-2165520188.

### DIFF
--- a/attach/cid.c
+++ b/attach/cid.c
@@ -105,7 +105,7 @@ static void cid_save_attachment(struct Body *b, struct CidMapList *cid_map_list)
   if (!b || !cid_map_list)
     return;
 
-  char *id = mutt_param_get(&b->parameter, "content-id");
+  char *id = b->content_id;
   if (!id)
     return;
 
--- a/compose/functions.c
+++ b/compose/functions.c
@@ -1024,7 +1024,7 @@ static int op_attachment_edit_content_id(struct ComposeSharedData *shared, int o
   struct AttachPtr *cur_att = current_attachment(shared->adata->actx,
                                                  shared->adata->menu);
 
-  char *id = mutt_param_get(&cur_att->body->parameter, "content-id");
+  char *id = cur_att->body->content_id;
   if (id)
   {
     buf_strcpy(buf, id);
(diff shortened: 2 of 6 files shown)


dcpurton (approved):
Looks good. Thanks for fixing.

roccoblues (approved):

PR #4336, opened 2024-06-27
Allow toggling numeric configs

### LINKED ISSUE #4333: Toggle `pager_index_lines`
I find `pager_index_lines` useful, but sometimes I want to just view the message and hide the `pager_index_lines`. I've come up with the following macro:
```ini
set pager_index_lines = 10
macro pager \Cf '<enter-command>set pager_index_lines=`echo $((10 - $pager_index_lines))`<enter>' "toggle pager_index_lines"
```
While this works alright, it could be desirable to have dedicated function to toggle `pager_index_lines` (cleaner config code, easier to setup for users, ...).

gahr:
This allows the `toggle` command to work with numeric configs, by switching their value from whatever they are to 0 and back.

This can be useful for temporarily turning "off" features like `pager_index_lines` or `toggle`.

Fixes #4333


parse/set.c
```diff
@@ -554,15 +561,15 @@ enum CommandResult parse_set(struct Buffer *buf, struct Buffer *s,
       s->dptr++;
     }
 
-    if (!bool_or_quad && (inv || (unset && prefix)))
+    if (!invertible && (inv || (unset && prefix)))
     {
       if (data == MUTT_SET_SET)
       {
-        buf_printf(err, _("Prefixes 'no' and 'inv' may only be used with bool/quad variables"));
+        buf_printf(err, _("Prefixes 'no' and 'inv' may only be used with bool/quad/number variables"));
```
roccoblues, line 568: Hm, I'm probably doing something stupid but how can I use `set no` with a number config?

roccoblues (commented):
👍🏼 I like it. `toggle` also works for me.

I just stumbled upon the `no`/`inv` prefix error message. Maybe you can shed some light on it?

config/number.c
```diff
@@ -265,7 +293,31 @@ static int number_reset(const struct ConfigSet *cs, void *var,
...
+}
+
+/**
+ * number_he_toggle - Toggle the value of a number (value <-> 0)
+ * @param sub Config Subset
+ * @param he  HashElem representing config item
+ * @param err Buffer for error messages
+ * @retval num Result, e.g. #CSR_SUCCESS
+ */
+int number_he_toggle(struct ConfigSubset *sub, struct HashElem *he, struct Buffer *err)
```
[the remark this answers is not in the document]
gahr: yep

parse/set.c
gahr: Oh right, I haven't looked into `no`. `inv` is how `toggle` is implemented:
https://github.com/neomutt/neomutt/blob/f10c0f0dd82d97f88fe54efc1e9aad2c1066d407/commands.c#L1700

gahr: A `no` prefix is the same as `unset`:
https://github.com/neomutt/neomutt/blob/f10c0f0dd82d97f88fe54efc1e9aad2c1066d407/commands.c#L1719

We can probably make `unset wrap` be equivalent to `set wrap 0`.

gahr: Now you can unset a numeric value (or setting it with a `no` prefix) to turn it into a 0 :)

gahr:
> 👍🏼 I like it. `toggle` also works for me.
> 
> I just stumbled upon the `no`/`inv` prefix error message. Maybe you can shed some light on it?

* `inv`: equivalent to `toggle`
* `no`: I made it work like `set <foo>=0`

roccoblues (approved):

gahr:
Thanks @roccoblues . Let's wait and see what @flatcap thinks. If I recall correctly, we have discussed `set nowrap` at some poiny and he had arguments against it.

flatcap (commented):
I like the concept, but for the implementation I'd probably have created a new flag for [`struct ConfigDef`](https://github.com/neomutt/neomutt/blob/771158d10511f3f81fd41a2a5440db9af7767afb/config/set.h#L63-L86).
This would keep the config's value as just a number.

https://github.com/neomutt/neomutt/blob/771158d10511f3f81fd41a2a5440db9af7767afb/config/types.h#L52-L66

The flags are a little complicated as they mix global and type-specific values.
 (наб's refactor helped)

 ```c
  D_B_NUMBER_TOGGLED,              ///< Flag to indicate that a number is in a toggled state

#define D_NUMBER_TOGGLED           (1 << D_B_NUMBER_TOGGLED)           ///< Flag to indicate that a number is in a toggled state
```

flatcap:
I had a go at implementing a _flag_ solution.
Unfortunately, `struct ConfigDef` (where I'd be storing the flag) is passed `const` to the variable-changing functions.

flatcap (approved):

Issue #3410, opened 2022-05-23
Neomutt taking more than 7 minutes to open


vikramrajsitpal:
Whenever I try to open neomutt, it takes more than 5 minutes to open and after opening (if I patiently wait), does not work. [Infinite hang]

## Expected Behaviour

Just opening neomutt

## Actual Behaviour
Opening neomutt takes way too long (upwards of 7 minutes) to open up, and show the inbox. Caching of headers is not a problem either. Even when it does open up (sometimes) with the correct emails in inbox, it hangs forever. It never works and I have to terminate eventually.

https://user-images.githubusercontent.com/7771452/169736866-d2adb38a-8dec-40f7-b9ed-f95973431070.mov

## Steps to Reproduce

Just open neomutt. (screen recording above)

## How often does this happen?

- Always

## When did it start to happen?

Since I started using neotmutt after a month of not using it. It never opens up, and even when it does sometimes (after waiting patiently for too long), it immediately hangs. This never used to happen when I initially started using it in March 2022.
`top` command shows no high CPU or memory consumption.

I used brew to install it on my Mac.

I see the error sometimes which shows up in the screen recording.

## NeoMutt Version

```
NeoMutt 20220429
Copyright (C) 1996-2022 Michael R. Elkins and others.
NeoMutt comes with ABSOLUTELY NO WARRANTY; for details type 'neomutt -vv'.
NeoMutt is free software, and you are welcome to redistribute it
under certain conditions; type 'neomutt -vv' for details.

System: Darwin 21.5.0 (x86_64)
ncurses: ncurses 6.3.20211021 (compiled with 6.3.20211021)
libiconv: 1.11
libidn2: 2.3.2 (compiled with 2.3.2)
GPGME: 1.17.1
OpenSSL: OpenSSL 1.1.1n  15 Mar 2022
libnotmuch: 5.6.0
storage: tokyocabinet, lmdb

Configure options: --prefix=/usr/local/Cellar/neomutt/20220429 --gss --disable-idn --idn2 --lmdb --notmuch --sasl --tokyocabinet --with-gpgme=/usr/local/opt/gpgme --with-lua=/usr/local/opt/lua --with-ssl=/usr/local/opt/openssl@1.1 --with-ui=ncurses

[...]

PR #4346, opened 2024-07-07
Do not force mailbox stats calculations on startup.

vchimishuk:
* **What does this PR do?**
A few years ago after update I've noticed that new version of `neomutt` becomes very slow at startup. It takes tens of seconds on my configuration to start, showing `Sorting mailbox...` during this time. I have a few `mbox`es of 100-200Mb size each.
 In [6bad373](https://github.com/neomutt/neomutt/commit/6bad3733ee2d59d04db522d396e4b999b28a2811?diff=unified&w=0) behavior of mutt_mailbox_check() was changed. Looks like incidentally. Original mutt_mailbox_check()'s code calls mailbox_check() with force flag only if needed. For instance, passing force flag to mutt_mailbox_check() doesn't mean that it will be passed directly to mailbox_check(). Commit 6bad373 changed that behavior by passing flags directly to mailbox_check().

 Probably the same bug as https://github.com/neomutt/neomutt/issues/3410

\---

     for that)
     \---

     `SUCCESS: No unit tests have failed.`

     \---

     Yes

https://github.com/neomutt/neomutt/issues/3410

### DIFF
--- a/mutt_mailbox.c
+++ b/mutt_mailbox.c
@@ -213,12 +213,12 @@ int mutt_mailbox_check(struct Mailbox *m_cur, CheckStatsFlags flags)
     if (!m->visible || !m->poll_new_mail)
       continue;
 
-    CheckStatsFlags mb_flags = MUTT_MAILBOX_CHECK_NO_FLAGS;
+    CheckStatsFlags m_flags = MUTT_MAILBOX_CHECK_NO_FLAGS;
     if (force_stats || (!m->first_check_stats_done && c_mail_check_stats))
     {
-      mb_flags |= MUTT_MAILBOX_CHECK_FORCE_STATS;
+      m_flags |= MUTT_MAILBOX_CHECK_FORCE_STATS;
     }
-    mailbox_check(m_cur, m, &st_cur, mb_flags);
+    mailbox_check(m_cur, m, &st_cur, m_flags);
     if (m->has_new)
       MailboxCount++;
     m->first_check_stats_done = true;
--- a/mutt_mailbox.c
+++ b/mutt_mailbox.c
@@ -182,10 +182,11 @@ int mutt_mailbox_check(struct Mailbox *m_cur, CheckStatsFlags flags)
   if ((flags == MUTT_MAILBOX_CHECK_NO_FLAGS) && ((t - MailboxTime) < c_mail_check))
     return MailboxCount;
 
+  bool force_stats = false;
   if ((flags & MUTT_MAILBOX_CHECK_FORCE_STATS) ||
       (c_mail_check_stats && ((t - MailboxStatsTime) >= c_mail_check_stats_interval)))
   {
-    flags |= MUTT_MAILBOX_CHECK_FORCE_STATS;
+    force_stats = true;
     MailboxStatsTime = t;
   }
 
@@ -212,12 +213,12 @@ int mutt_mailbox_check(struct Mailbox *m_cur, CheckStatsFlags flags)
     if (!m->visible || !m->poll_new_mail)
       continue;
 
-    CheckStatsFlags m_flags = flags;
-    if (!m->first_check_stats_done && c_mail_check_stats)
+    CheckStatsFlags mb_flags = MUTT_MAILBOX_CHECK_NO_FLAGS;
+    if (force_stats || (!m->first_check_stats_done && c_mail_check_stats))
     {
-      m_flags |= MUTT_MAILBOX_CHECK_FORCE_STATS;
+      mb_flags |= MUTT_MAILBOX_CHECK_FORCE_STATS;
     }
-    mailbox_check(m_cur, m, &st_cur, m_flags);
+    mailbox_check(m_cur, m, &st_cur, mb_flags);
     if (m->has_new)
       MailboxCount++;
     m->first_check_stats_done = true;


gahr:
Thanks @vchimishuk, I think you nailed it! I'll just need to re-understand it all to convince me that this doesn't introduce some subtle unwanted behaviour. I'll be on it!

mutt_mailbox.c
```diff
@@ -212,8 +213,8 @@ int mutt_mailbox_check(struct Mailbox *m_cur, CheckStatsFlags flags)
     if (!m->visible || !m->poll_new_mail)
       continue;
 
-    CheckStatsFlags m_flags = flags;
-    if (!m->first_check_stats_done && c_mail_check_stats)
+    CheckStatsFlags m_flags = MUTT_MAILBOX_CHECK_NO_FLAGS;
```
roccoblues, line 216: This means that we don't pass the existing flags through anymore. Won't that mean that `MUTT_MAILBOX_CHECK_IMMEDIATE` no longer gets passed through to the imap code?

gahr: Interesting.. perhaps we only want to remove the `MUTT_MAILBOX_CHECK_FORCE_STATS` bit from the flags here?

gahr: Would this work? (diff against the vchimishuk:slow-mbox branch)

```diff
diff --git a/mutt_mailbox.c b/mutt_mailbox.c
index 48f7d6003..81fd0db4a 100644
--- a/mutt_mailbox.c
+++ b/mutt_mailbox.c
@@ -213,10 +213,10 @@ int mutt_mailbox_check(struct Mailbox *m_cur, CheckStatsFlags flags)
     if (!m->visible || !m->poll_new_mail)
       continue;
 
-    CheckStatsFlags m_flags = MUTT_MAILBOX_CHECK_NO_FLAGS;
-    if (force_stats || (!m->first_check_stats_done && c_mail_check_stats))
+    CheckStatsFlags m_flags = flags;
+    if (!(force_stats || (!m->first_check_stats_done && c_mail_check_stats)))
     {
-      m_flags |= MUTT_MAILBOX_CHECK_FORCE_STATS;
+      m_flags &= ~MUTT_MAILBOX_CHECK_FORCE_STATS;
     }
     mailbox_check(m_cur, m, &st_cur, m_flags);
     if (m->has_new)
```

roccoblues: This function is a mess. Now it feels weird that we set `force_stats` based on some conditions and then later check it *again* plus some other condition to remove the flag. Can we set `force_stats` in a way that later we just need:
```
if (!force_stats)
  m_flags &= ~MUTT_MAILBOX_CHECK_FORCE_STATS;
```

gahr: `m_flags` is per mailbox

roccoblues: ah right, the `if` statement is still giving me a headache but I think it should work.

vchimishuk: @gahr, `m_flags &= ~MUTT_MAILBOX_CHECK_FORCE_STATS;` doesn't fix the issue because `flags` has `MUTT_MAILBOX_CHECK_FORCE` flag set not `MUTT_MAILBOX_CHECK_FORCE_STATS`.

I've verified and `m_flags &= ~MUTT_MAILBOX_CHECK_FORCE` does fix the issue.

gahr: Now I'm confused. So here's the logic in this PR:

1. if flag has `MUTT_MAILBOX_CHECK_FORCE`, we call `mutt_update_num_postponed`
https://github.com/neomutt/neomutt/blob/4309b939c88f410aa1cddf6c04abd04adf89f128/mutt_mailbox.c#L174-L175
ok, I think this is fine and not under debate
1. if no flag is set and `c_mail_check` hasn't elapsed, don't do anything
https://github.com/neomutt/neomutt/blob/4309b939c88f410aa1cddf6c04abd04adf89f128/mutt_mailbox.c#L181-L183
ok, I think this is also fine and not under debate
3. then, we *always* disable `MUTT_MAILBOX_CHECK_FORCE` altogether, and only set `MUTT_MAILBOX_CHECK_FORCE_STATS` if needed
https://github.com/neomutt/neomutt/blob/4309b939c88f410aa1cddf6c04abd04adf89f128/mutt_mailbox.c#L216-L220
where *if needed* is driven by whether `MUTT_MAILBOX_CHECK_FORCE_STATS` was originally part of the flags or whether the `c_mail_check_stats_interval` time has elapsed
https://github.com/neomutt/neomutt/blob/4309b939c88f410aa1cddf6c04abd04adf89f128/mutt_mailbox.c#L185-L191

My confusion is about `MUTT_MAILBOX_CHECK_FORCE`. It looks like now we only use it to call `mutt_update_num_postponed`. It also looks like no backend explicitly checks for `MUTT_MAILBOX_CHECK_FORCE`. There's only mbox but it treats CHECK and CHECK_STATS just the same:
https://github.com/neomutt/neomutt/blob/19653add8a9b372fd078aa98662b4d2c989445a7/mbox/mbox.c#L1691

So, two questions:

1. why removing `MUTT_MAILBOX_CHECK_FORCE` from the flags help?
2. so, should we rename `MUTT_MAILBOX_CHECK_FORCE` to `MUTT_MAILBOX_CHECK_NUM_POSTPONED`?

I'm sure I'm missing something, but I can't pinpoint what it is.

vchimishuk: > why removing MUTT_MAILBOX_CHECK_FORCE from the flags help?

Because `MUTT_MAILBOX_CHECK_FORCE` is passed into `mutt_mailbox_check()` from `dlg_index()`
https://github.com/neomutt/neomutt/blob/4309b939c88f410aa1cddf6c04abd04adf89f128/index/dlg_index.c#L1103 So, it is `MUTT_MAILBOX_CHECK_FORCE` who triggers the check.

In the original code (changed by https://github.com/neomutt/neomutt/commit/6bad3733ee2d59d04db522d396e4b999b28a2811) `check_stats` flag (passed to `mailbox_check()`) were set only if `force & MUTT_MAILBOX_CHECK_FORCE_STATS`. Now it passes `flags`  as is because of the next line.
```
CheckStatsFlags m_flags = flags;
```
Which makes `MUTT_MAILBOX_CHECK_FORCE` proxied. Which was not true originally. 


> so, should we rename MUTT_MAILBOX_CHECK_FORCE to MUTT_MAILBOX_CHECK_NUM_POSTPONED?

Sorry, I'm not sure I can answer this question. I still has a limited understanding of those constants. However, for me it looks like `MUTT_MAILBOX_CHECK_FORCE` has much wider usage than simply check number of postponed.

vchimishuk: What do you mean "the wider usage"?

gahr: Ah! Maybe I *do* get it and this is specifically an issue for `mbox`.

Do we achieve the same result if change this
https://github.com/neomutt/neomutt/blob/19653add8a9b372fd078aa98662b4d2c989445a7/mbox/mbox.c#L1691

into `if (flags & MUTT_MAILBOX_CHECK_FORCE_STATS)`?

gahr: Also, please check if #4357 makes sense to you

vchimishuk: > Ah! Maybe I do get it and this is specifically an issue for mbox.

Yes. 

> into if (flags & MUTT_MAILBOX_CHECK_FORCE_STATS)?

Yes. This trick also works.

> Also, please check if https://github.com/neomutt/neomutt/pull/4357 makes sense to you

I checked the code and it does make sense to me. Also I've compiled and tested it on my setup and it works just fine.

gahr: Thanks a lot! Then, if you're ok with that, I'd close this PR and proceed with mine.

vchimishuk: Sure. Thank you for getting the fix moving into mainline.

gahr: Thank you for your patience while I tried to understand the issue :)

PR #4347, opened 2024-07-07
Support BerkeleyDB 18.1

### LINKED ISSUE #4345: Current stable of  BerkeleyDB apparently not supported "ouf-of-the-box"
## Expected Behaviour

Compile successfully with version 18.1.40 released in 2020

## Actual Behaviour

I get following error message

~~~
Checking for BerkeleyDB in /usr/include/...no (expecting 5.3, got 18.1)
Checking for BerkeleyDB in /usr/include/db-5-3...no
Checking for BerkeleyDB in /usr/include/db53...no
Checking for BerkeleyDB in /usr/include/db5.3...no
Checking for BerkeleyDB in /usr/include/db-5...no
Checking for BerkeleyDB in /usr/include/db5...no
Checking for BerkeleyDB in /usr/include/...no (expecting 6.2, got 18.1)
Checking for BerkeleyDB in /usr/include/db-6-2...no
Checking for BerkeleyDB in /usr/include/db62...no
Checking for BerkeleyDB in /usr/include/db6.2...no
Checking for BerkeleyDB in /usr/include/db-6...no
Checking for BerkeleyDB in /usr/include/db6...no
Checking for BerkeleyDB in /usr/include/...no (expecting 4.8, got 18.1)
Checking for BerkeleyDB in /usr/include/db-4-8...no
Checking for BerkeleyDB in /usr/include/db48...no
Checking for BerkeleyDB in /usr/include/db4.8...no
Checking for BerkeleyDB in /usr/include/db-4...no
Checking for BerkeleyDB in /usr/include/db4...no
~~~

## Steps to Reproduce

Install version 18.1 of BerkeleyDB, and configure with

> ./configure --sasl --bdb  --gss --pgp --sqlite --disable-idn --idn2 --zlib --prefix=/usr --sysconfdir=/etc --libexecdir=/usr/lib --localstatedir=/var --datadir=/usr/share --mandir=/usr/share/man --infodir=/usr/share/info

(other dependencies might be required)

- Always

I guess whe my environment updated the default BerkeleyDB to version 18.1

## NeoMutt Version

> 20240425

## Extra Info

* Operating System and its version -> Windows 10, cygwin
* Were you using multiple copies of NeoMutt at once? -> no
* Were you using 'screen' or 'tmux'? -> irrelevant
* Is your email local (maildir) or remote (IMAP)? -> irrelevant

gahr:
Fixes #4345

### DIFF
--- a/auto.def
+++ b/auto.def
@@ -831,7 +831,7 @@ if {[get-define want-bdb]} {
     return 0
   }
 
-  set bdb_versions [opt-val with-bdb-version { 5.3 6.2 4.8 }] ;# Will be checked in order
+  set bdb_versions [opt-val with-bdb-version { 18.1 5.3 6.2 4.8 }] ;# Will be checked in order
   set bdb_prefix [opt-val with-bdb $prefix]
 
   foreach ver $bdb_versions {


roccoblues (approved):
Crazy version jump. 🤷🏼‍♂️

auto.def
```diff
@@ -831,7 +831,7 @@ if {[get-define want-bdb]} {
     return 0
   }
 
-  set bdb_versions [opt-val with-bdb-version { 5.3 6.2 4.8 }] ;# Will be checked in order
+  set bdb_versions [opt-val with-bdb-version { 18.1 5.3 6.2 4.8 }] ;# Will be checked in order
```
ossilator, line 834: this order is kinda wrong.
the preference for 5.3 is due to it being the last gpl version, and thus likely the only one to be found on most linux systems. the preference for 6.2 over 4.8 is probably feature-based (check git history). so it probably should be `5.3 18.1 6.2 4.8`.
not that it would be terribly relevant for a client app, but linking to bdb > 5.3 puts the build under agpl terms (which are compatible only with gpl3). so a more comprehensive build config would allow selecting acceptable licenses.

speaking of license: unlike mutt, neomutt's top-level docu isn't clear about whether it's v2-only or v2+: README.md contradicts itself, and LICENSE.md doesn't say anything, but kind of implies v2-only. the c/h files i checked say v2+.

on a tanget, you may want to change the license headers to SPDX to cut down the boilerplate.

gahr: > the preference for 5.3 is due to it being the last gpl version

I initially introduced support for 6.2 in #613. I can't find references to this rationale and I don't remember doing this reasoning myself. What do you think that was the case, as opposed to, say, 5.3 still being the most deployed version at that time?

ossilator: i don't really understand your question, as the grammar is botched.
but if nobody thought about it, then i presume it was kinda random and equally wrong-ish. then the order should be `5.3 4.8 18.1 6.2`, i guess. but i wonder about all the omitted versions in between.

gahr: s/What/Why/

HTH.

gahr: So, first it was 5.3, 4.8.

Then I introduced 6.2 and made it 5.3, 6.2, 4.8. This was in 2017. It's been working fine and nobody complained about the ordering.

Then I introduced the possibility to specify the version in #1064. This was in 2018.

Now, I'm adding 18.1 in the list of versions checked by default.

You argued about some rationale 5.3 being preferred to 6.2. I said *I* did the change and I don't remember using that rationale. Now you're arguing that, then, the whole ordering is wrong.

What are you after?

ossilator: i'm pointing out the preference _of the code_. you clearly have none, because you didn't realize the intent of the original authors.

it's unsurprising that nobody complained, because being affected by this is a total corner case, it being an actual problem is unlikely, and somebody noticing it without looking at the code/licensing is basically impossible. but it's legalese, so you should be damn careful about it, lest you lead your downstreams right into a trap.

given the legal minefield and the fact that the version can be selected manually, i'd argue that you should just drop versions > 5.3 from the auto search list.
4.8's presence seems kinda arbitrary and it is way outdated, so maybe drop it as well.

gahr: > i'm pointing out the preference _of the code_. you clearly have none, because you didn't realize the intent of the original authors.

Which intent?

gahr: > given the legal minefield and the fact that the version can be selected manually, i'd argue that you should just drop versions > 5.3 from the auto search list.

Are you suggesting that if a user has bdb 18 installed, we shouldn't pick it up because we know better?

ossilator: > Which intent?

to prefer versions with an unproblematic license.

> Are you suggesting that if a user has bdb 18 installed, we shouldn't pick it up because we know better?

yes. it seems self-evident to me that incompatibly licensed deps shouldn't be automatically picked up without at least a big fat warning.

gahr: >>> i'm pointing out the preference _of the code_. you clearly have none, because you didn't realize the intent of the original authors.
>> 
>> Which intent?
> to prefer versions with an unproblematic license.

Did I mention that *I* am the one who introduced support for 6.2 and I didn't intend to prefer versions with unproblematic licenses?

gahr: > yes. it seems self-evident to me that incompatibly licensed deps shouldn't be automatically picked up without at least a big fat warning.

Also, for the benefits of whoever might read this: BerkleyDB is dually licensed under a proprietary/AGPL license, if I'm reading this page correctly:

https://www.oracle.com/database/technologies/related/berkeleydb/berkeleydb-licensing.html

From the first paragraph: "Our open source license is OSI-certified and permits use of Berkeley DB in open source projects".

So what exactly are you calling an incompatible license?

ossilator: > Did I mention that I am the one who introduced support for 6.2 and I didn't intend to prefer versions with unproblematic licenses?

right, but without _any_ intent it doesn't make sense that the order is jumbled, so presumably _someone_ suggested that you should put 5.3 first.

> BerkleyDB is dually licensed under a proprietary/AGPL license, if I'm reading this page correctly

yes, you are.

> So what exactly are you calling an incompatible license?

agpl is incompatible with gpl2.

gahr: > > Did I mention that I am the one who introduced support for 6.2 and I didn't intend to prefer versions with unproblematic licenses?
> 
> right, but without _any_ intent it doesn't make sense that the order is jumbled, so presumably _someone_ suggested that you should put 5.3 first.

At this point I think you're just trolling me. 



> > So what exactly are you calling an incompatible license?
> > 
> agpl is incompatible with gpl2.

Grep for 'or (at your option) any later' in our source code.

ossilator: > At this point I think you're just trolling me.

you are free to believe whatever you like.

> > agpl is incompatible with gpl2.
> 
> Grep for 'or (at your option) any later' in our source code.

it's not "at your option" if it's actually the only legal option. this is an actual legal problem for people distributing binaries.

also, see my first comment.

PR #4348, opened 2024-07-07
Check mutt_date_parse_date's return value

### LINKED ISSUE #4344: Behavior with unparsable values of the "Expires" header
This issue is to report a minor problem that confused me about how neomutt seems to handle the `Expires` header.

Considering the following mbox file, which is adapted from a real email I received:

```
From test@example.com Sat Jul  6 16:20:16 2024
From: test@example.com
To: test2@example.com
Subject: 
Expires: Thursday, 02 January 2025 16:15:00 +0000
Status: O
Content-Length: 5
Lines: 1

BLAH
```

Using a neomutt built from master, this email is considered to be expired (as can be seen, e.g., with the `color index black red ~E` directive and adding another mail to the mbox file). This puzzled me because the `Expires` value is in the future.

Looking more into it, what seems to happen is that `mutt_date_parse_date` on the value of `Expires` returns -1, which indicates that the parsing failed. It appears that this is the email sender's fault: the value of `Expires` should have been an RFC-5322 datetime value (see https://www.ietf.org/archive/id/draft-billon-expires-08.html). It looks like `mutt_date_parse_date` has a fallback (with `PREX_RFC5322_DATE_LAX`) but which also doesn't manage to parse this value.

It further appears that, in `email/parse.c`, the value of `-1` which is returned when failing to parse the date is always considered to be less than `mutt_date_now()`, so messages with an unparseable `Expires` value are always considered as expired.

This is not very improvement but I would think that it's better to not consider/show messages as expired if the `Expires` field contains an unparseable date. In `email/parse.c`, I guess this would mean testing for the return value of `-1` for `mutt_date_parse_date`.

(Also, it would be better if `mutt_date_parse_date` managed to parse the incorrect date format used by this sender. It doesn't look like the current heuristics are bulletproof: trying to use the defaut format of `date -d '2 months'` on my system, namely "Fri Sep  6 21:58:57 CEST 2024", also returns `-1`. I'd understand if it's not considered feasible for neomutt to understand every possible broken date format, however.)

I hope this report is useful. Thanks in any case for developing neomutt!

gahr:
While at it, extend PREX_RFC5322_DATE_LAX to accept full month names.

Fixes #4344

### DIFF
--- a/email/parse.c
+++ b/email/parse.c
@@ -791,6 +791,7 @@ int mutt_rfc822_parse_line(struct Envelope *env, struct Email *e,
       if (e)
       {
         struct Tz tz = { 0 };
+        // the caller will check e->date_sent for -1
         e->date_sent = mutt_date_parse_date(body, &tz);
         if (e->date_sent > 0)
         {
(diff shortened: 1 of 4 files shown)


roccoblues (approved):
nit: I would have preferred two separate commits.

PR #4351, opened 2024-07-13
 mutt_path_to_absolute: Fail when realpath fails.

falsifian:
## What does this PR do?

 mutt_path_to_absolute was assuming realpath leaves something sensible in `*path` when it fails with `ENOENT`. Also, the last test case for `mutt_path_to_absolute` required it to put the right answer in `*path` in a situation where `realpath` fails with `EACCES`.
 
 In fact, the contents of `*path` are undefined when `realpath` fails. In particular, the last test case in `mutt_path_to_absolute` (which involves exactly this situation) was failing on OpenBSD, like this:

```
Test test_mutt_path_to_absolute...              [ FAILED ]
  Case Common setup:
    mutt_path_to_absolute.c:102: test_check_str_eq... failed
      Expected : '/home/falsifian/co/neomutt-test-files/maildir/damson/tmp'
      Actual   : 'tmp'
```

So:
 - Remove that last test case, which can't be expected to pass when `mutt_path_to_absolute` relies on `realpath`; and
 - Make `mutt_path_to_absolute` always fail when `realpath` fails, even with `ENOENT`.

Also:
- Correct the doc comment for `mutt_path_to_absolute`.
- In the test, make sure the path argument is always at least `PATH_MAX` long. (It doesn't matter under the current implementation, but IMO tests should be written to match the documentation, not the implementation.)
- Simplify cleanup by calling `buf_pool_release` earlier.

### DIFF
--- a/mutt/path.c
+++ b/mutt/path.c
@@ -322,11 +322,12 @@ char *mutt_path_dirname(const char *path)
 /**
  * mutt_path_to_absolute - Convert relative filepath to an absolute path
  * @param path      Relative path
- * @param reference Absolute path that \a path is relative to
+ * @param reference \a path is relative to the dirname of this path
  * @retval true  Success
  * @retval false Failure
  *
- * Use POSIX functions to convert a path to absolute, relatively to another path
+ * Use POSIX functions to convert a path to absolute, relative to the dirname
+ * of another path.
  *
  * @note \a path should be at least of PATH_MAX length
  */
@@ -348,14 +349,13 @@ bool mutt_path_to_absolute(char *path, const char *reference)
   FREE(&dirpath);
 
   path = realpath(buf_string(abs_path), path);
-  if (!path && (errno != ENOENT))
+  buf_pool_release(&abs_path);
+  if (!path)
   {
     mutt_perror(_("Error: converting path to absolute"));
-    buf_pool_release(&abs_path);
     return false;
   }
 
-  buf_pool_release(&abs_path);
   return true;
 }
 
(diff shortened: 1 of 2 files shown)


falsifian:
NB I only briefly tried actually using neomutt after this change. As far as I can tell the function in question is just used when resolving paths to config files; I guess I could try setting up some complicated situation where my config file(s) have symlinks in their paths if more testing is desired.

falsifian:
I just realized the `ENOENT` exception in `mutt_path_to_absolute` was unrelated to the error seen in the last test case, which was `EACCES`. I pushed a change to the commit message accordingly. (No change to the diff.)

falsifian:
@flatcap You introduced the `ENOENT` special case in commit 92ca49b16d23c6d30a7354792d1e89bff5792793, which seems related to https://github.com/neomutt/neomutt/pull/1293. Do you remember why you added it?

gahr:
I think the idea is that if `realpath` fails, we want to retain the original `path`. As you said, the assumption that `realpath` would not touch the output variable on failure was misplaced. I'd rather fix the code so it behaves the way it was intended to behave, i.e., return the original `path` and success if ENOENT.

gahr:
> I should say: shouldn't require resolving symlinks and `/./` and `/../`. Realpath makes sure the output path doesn't have any of that, but shouldn't they be harmless in this case?

I think the reason why we want to normalize the path is for cycle prevention:
https://github.com/neomutt/neomutt/blob/cb368416dab3aafa9332fc9f1be8eb37e65fc600/commands.c#L234-L248

We keep a list of paths we've traversed, and we need a way to compare those paths. I'll need to think more about what to do in the ENOENT case :)

flatcap:
Superseded by #4391.

Thanks for the PR, @falsifian.

PR #4357, opened 2024-07-20
Don't force mbox stats calculations on startup

gahr:
This supersedes #4346.

Here's my current best understanding of the issue:

On startup, `main()` used the flags `MUTT_MAILBOX_CHECK_FORCE | MUTT_MAILBOX_CHECK_IMMEDIATE` in its call to `mutt_mailbox_check`. The flag `MUTT_MAILBOX_CHECK_FORCE` was checked only in two places, inconsistently:

1. in `mutt_mailbox_check`, to call `mutt_update_num_postponed()` https://github.com/neomutt/neomutt/blob/19653add8a9b372fd078aa98662b4d2c989445a7/mutt_mailbox.c#L174-L175

2. by the mbox backend, same as it used `MUTT_MAILBOX_CHECK_FORCE_STATS` https://github.com/neomutt/neomutt/blob/19653add8a9b372fd078aa98662b4d2c989445a7/mbox/mbox.c#L1691

The result is that main is forcing a stats check on mbox mailboxes, even though that was never the intention.

As correctly pointed out in #4346, in
6bad3733ee2d59d04db522d396e4b999b28a2811 we inadvertently started propagating `MUTT_MAILBOX_CHECK_FORCE` from `mutt_mailbox_check` to `mailbox_check`, which initially took a boolean flag, aptly named `check_stats`.

So, because the mbox backend had a `MUTT_MAILBOX_CHECK_FORCE | MUTT_MAILBOX_CHECK_FORCE_STATS` logic, we ended up checking stats unnecessarily.

This PR clarifies the flags a bit.

* `MUTT_MAILBOX_CHECK_FORCE` becomes `MUTT_MAILBOX_CHECK_POSTPONED`, since this is what it does.
* it clarifies that `MUTT_MAILBOX_CHECK_STATS` is only used in the context of the `<check-stats>` function

### DIFF
--- a/main.c
+++ b/main.c
@@ -1322,7 +1322,7 @@ main
     {
       const bool c_imap_passive = cs_subset_bool(NeoMutt->sub, "imap_passive");
       cs_subset_str_native_set(NeoMutt->sub, "imap_passive", false, NULL);
-      const CheckStatsFlags csflags = MUTT_MAILBOX_CHECK_FORCE | MUTT_MAILBOX_CHECK_IMMEDIATE;
+      const CheckStatsFlags csflags = MUTT_MAILBOX_CHECK_IMMEDIATE;
       if (mutt_mailbox_check(NULL, csflags) == 0)
       {
         mutt_message(_("No mailbox with new mail"));
--- a/mutt_mailbox.c
+++ b/mutt_mailbox.c
@@ -78,7 +78,7 @@ static bool is_same_mailbox(struct Mailbox *m1, struct Mailbox *m2,
  * @param m_cur   Current Mailbox
  * @param m_check Mailbox to check
  * @param st_cur  stat() info for the current Mailbox
- * @param flags   Flags, e.g. #MUTT_MAILBOX_CHECK_FORCE
+ * @param flags   Flags, e.g. #MUTT_MAILBOX_CHECK_POSTPONED
  */
 static void mailbox_check(struct Mailbox *m_cur, struct Mailbox *m_check,
                           struct stat *st_cur, CheckStatsFlags flags)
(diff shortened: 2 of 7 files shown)


main.c
```diff
@@ -1322,7 +1322,7 @@ main
     {
       const bool c_imap_passive = cs_subset_bool(NeoMutt->sub, "imap_passive");
       cs_subset_str_native_set(NeoMutt->sub, "imap_passive", false, NULL);
-      const CheckStatsFlags csflags = MUTT_MAILBOX_CHECK_FORCE | MUTT_MAILBOX_CHECK_IMMEDIATE;
```
roccoblues, line 1325: Why do we no longer want to update the postponed number on startup?

mutt_mailbox.c
```diff
@@ -368,7 +368,7 @@ struct Mailbox *mutt_mailbox_next(struct Mailbox *m_cur, struct Buffer *s)
     if (m_res)
       return m_res;
 
-    mutt_mailbox_check(m_cur, MUTT_MAILBOX_CHECK_FORCE); /* mailbox was wrong - resync things */
+    mutt_mailbox_check(m_cur, MUTT_MAILBOX_CHECK_POSTPONED); /* mailbox was wrong - resync things */
```
roccoblues, line 371: `/* mailbox was wrong - resync things */` I don't really understand this comment. How does checking the postponded number help here?

roccoblues (commented):
Thanks, I think that clarifies the situation. I'm not sure about the change in main though.

main.c
gahr: I don't think that was ever needed. The branch here is to handle `-Z`, which is meant to open the first mailbox with new mail. We're only interested in new mail checks here.

mutt_mailbox.c
gahr: I have no idea - I just changed the enumerator name here.

roccoblues: Screenshot 2024-08-12 at 09 38 50

I'd probably remove it. 😆

main.c
roccoblues: Ah, I didn't check which main branch we where in. Yep, makes sense.

roccoblues (approved):
Thanks for digging into this!

PR #4362, opened 2024-07-31
Assume 'struct timespec' exists

alejandro-colomar:
It is guaranteed by ISO C11, which is the minimum C version we support.


roccoblues (approved):
Nice, thanks!

gahr (approved):

PR #4365, opened 2024-08-05
Fix sorting INBOX and its subfolders

### LINKED ISSUE #4364: INBOX-first is broken for children of INBOX
In NeoMutt, we try to put INBOX at the top of the folder list in the browser and sidebar. This has a long story, see for example #356.

What I didn't think of, at the time, was that INBOX can have children. So if you have a list of mailboxes like this:

```
INBOX
INBOX.Archive
FooBar
FooBar.Subdir
```

What you see in the browser is this (INBOX.Subdir is alpha-sorted without any special treatment):

```
+INBOX
+FooBar
+FooBar.SubDir
+INBOX.Subdir
```

Worse still, in the sidebar you see this (`sidebar_short_path=false`):

```
INBOX
FooBar
..FooBar.SubDir
..INBOX.Archive
```

Worse-worse still, in the sidebar you see this (`sidebar_short_path=true`):

```
INBOX
FooBar
..SubDir
..Archive
```

gahr:
Fixes #4364

The bottom line is:

```c
    return mutt_istr_equal(a + 1, "inbox") ? -1 :
           mutt_istr_equal(b + 1, "inbox") ? 1 :
                                             0;
```

vs.

```c
#define IS_INBOX(s) (mutt_istrn_equal(s, "inbox", 5) && !isalnum((s)[5]))
#define CMP_INBOX(a, b) (IS_INBOX(b) - IS_INBOX(a))

...

return CMP_INBOX(a + 1, b + 1);
```

so,

1. check both `a` and `b` in all cases: before, we weren't contemplating the possibility that both folders might be Inbox-ish
2. treat `Inbox.<Anything>` as `Inbox`

### DIFF
--- a/mutt/string.c
+++ b/mutt/string.c
@@ -858,3 +858,66 @@ void mutt_str_hyphenate(char *buf, size_t buflen, const char *str)
       *buf = '-';
   }
 }
+
+/**
+ * mutt_str_inbox_cmp - Do two folders share the same path and one is an inbox - @ingroup sort_api
+ * @param a First path
+ * @param b Second path
+ * @retval -1 a is INBOX of b
+ * @retval  0 None is INBOX
+ * @retval  1 b is INBOX for a
+ *
+ * This function compares two folder paths. It first looks for the position of
+ * the last common '/' character. If a valid position is found and it's not the
+ * last character in any of the two paths, the remaining parts of the paths are
+ * compared (case insensitively) with the string "INBOX" followed by a non
+ * alpha character, e.g., '.' or '/'. If only one of the two paths matches,
+ * it's reported as being less than the other and the function returns -1 (a <
+ * b) or 1 (a > b).  If both or no paths match the requirements, the two paths
+ * are considered equivalent and this function returns 0.
+ *
+ * Examples:
+ * * mutt_str_inbox_cmp("/foo/bar",      "/foo/baz") --> 0
+ * * mutt_str_inbox_cmp("/foo/bar/",     "/foo/bar/inbox") --> 0
+ * * mutt_str_inbox_cmp("/foo/bar/sent", "/foo/bar/inbox") --> 1
+ * * mutt_str_inbox_cmp("=INBOX",        "=Drafts") --> -1
+ * * mutt_str_inbox_cmp("=INBOX",        "=INBOX.Foo") --> 0
+ * * mutt_str_inbox_cmp("=INBOX.Foo",    "=Drafts") --> -1 
+ */
+int mutt_str_inbox_cmp(const char *a, const char *b)
+{
+#define IS_INBOX(s) (mutt_istrn_equal(s, "inbox", 5) && !isalnum((s)[5]))
+#define CMP_INBOX(a, b) (IS_INBOX(b) - IS_INBOX(a))
+
+  /* fast-track in case the paths have been mutt_pretty_mailbox'ified */
+  if ((a[0] == '+') && (b[0] == '+'))
+  {
+    return CMP_INBOX(a + 1, b + 1);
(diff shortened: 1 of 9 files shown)


mutt/string.c
```diff
@@ -858,3 +858,66 @@ void mutt_str_hyphenate(char *buf, size_t buflen, const char *str)
...
+
+  const char *a_end = strrchr(a, '/');
+  const char *b_end = strrchr(b, '/');
+
+  /* If one path contains a '/', but not the other */
+  if ((!a_end) ^ (!b_end))
+    return 0;
+
+  /* If neither path contains a '/' */
+  if (!a_end)
```
roccoblues, line 906: The comment doesn't match the code?

test/string/mutt_str_inbox_cmp.c
```diff
@@ -0,0 +1,52 @@
...
+  if (rc != 0)
+  {
+    return rc;
+  }
+  return mutt_str_coll(fa, fb);
+}
+
+void test_mutt_str_inbox_cmp(void)
+{
+  char *folders[] = { "+Inbox", "+Inbox.Archive", "+FooBar", "+FooBar.Baz" };
```
roccoblues, line 46: I tried to break this by not using exactly what we expect as input:

```
  char *folders[] = { "+Inbox", "+Foo", "+Inbox.Archive", "+Bar", "+FooBar", "+FooBar.Baz" };
```

This lead to this result?
```
  TEST_CHECK_STR_EQ(folders[0], "+Inbox");
  TEST_CHECK_STR_EQ(folders[1], "+Inbox.Archive");
  TEST_CHECK_STR_EQ(folders[2], "+Bar");
  TEST_CHECK_STR_EQ(folders[2], "+Bar");
  TEST_CHECK_STR_EQ(folders[2], "+Bar");
  TEST_CHECK_STR_EQ(folders[3], "+Foo");
```

Am I doing something stupid?

roccoblues: Also adding one entry with `+` in the list breaks it in interesting ways.

mutt/string.c
gahr: Well it does, in a way. Just above, I check `(!a_end) ^ (!b_end)`. So here, `!a_end` ↔ `!b_end`.

test/string/mutt_str_inbox_cmp.c
gahr: > I tried to break this by not using exactly what we expect as input:
> 
> ```
>   char *folders[] = { "+Inbox", "+Foo", "+Inbox.Archive", "+Bar", "+FooBar", "+FooBar.Baz" };
> ```
> 
> This lead to this result?
> 
> ```
>   TEST_CHECK_STR_EQ(folders[0], "+Inbox");
>   TEST_CHECK_STR_EQ(folders[1], "+Inbox.Archive");
>   TEST_CHECK_STR_EQ(folders[2], "+Bar");
>   TEST_CHECK_STR_EQ(folders[2], "+Bar");
>   TEST_CHECK_STR_EQ(folders[2], "+Bar");
>   TEST_CHECK_STR_EQ(folders[3], "+Foo");
> ```
> 
> Am I doing something stupid?

Yes, you're checking `folders[2]` multiple times. I've pushed an update to the test including your entries.

gahr: > Also adding one entry with `+` in the list breaks it in interesting ways.

The `+` will be alpha-sorted before `+<anything-else>` but `+Inbox.<anything-else>`, which we always want to sort first. This is what I would expect. What would *you* expect instead?

roccoblues (approved):

PR #4367, opened 2024-08-06
Let ~Y match each tag individually

### LINKED ISSUE #4360: `=Y`, `~Y` should match individual keywords, not the space-separated list of keywords
I've been diving into the neomutt source code recently so may try writing a patch if it's likely to be accepted.

I'd like to be able to limit to messages with the "s" IMAP keyword using the limit pattern `~Y^s$`.

Unfortunately that doesn't work right now because the pattern is matched against the whole string. So if my email has keywords "r" and "s" the pattern gets matched against the string "s r". (Presumably "r s" is possible too, but that's not what neomutt ended up with in my case.)

I suggest the meaning of `=Y` and `~Y` change to: at least one of the keywords matches the pattern.

This might annoy people who were relying on the old behaviour. But the old behaviour seems undefined since the order in which the keywords appear in the string is unspecified.

I am guessing (hoping) this is a simple change to the `MUTT_PAT_DRIVER_TAGS` case in `pattern_exec` in `pattern/exec.c`, to break the tag list into words and try each separately. However I don't understand why `MUTT_PAT_DRIVER_TAGS` appears in the `mutt_pattern_alias_exec` function (I don't really understand what that function does) so there's a piece of the picture I'm missing.

I can instead limit to `~Y"(^| )s($| )"` but that's annoying.

gahr:
Fixes #4360

### DIFF
--- a/pattern/exec.c
+++ b/pattern/exec.c
@@ -1078,14 +1078,12 @@ static bool pattern_exec(struct Pattern *pat, PatternExecFlags flags,
     case MUTT_PAT_DRIVER_TAGS:
     {
       struct Tag *tag = NULL;
+      bool matched = false;
       STAILQ_FOREACH(tag, &e->tags, entries)
       {
-        if (pat->pat_not ^ patmatch(pat, tag->name))
-        {
-          return true;
-        }
+        matched |= patmatch(pat, tag->name);
       }
-      return false;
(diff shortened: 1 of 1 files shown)


roccoblues (approved):
Looks good. 👍🏼 

I was about to ask for a test but then I saw how big the method is. 😄

PR #4378, opened 2024-09-03
Fix padding with an empty string

### LINKED ISSUE #4377: Neomutt freezes when `pager_format` incorrectly ends with `%>`
## Expected Behaviour

Neomutt should give an error when I specify a `pager_format` that is invalid (because it does not have a character to pad with after `%>`).

## Actual Behaviour

With the invalid `pager_format` set, when I press Enter on an email in the index to open the pager, Neomutt completely freezes up and stops responding to all user input. Even pressing Ctrl+C has no effect. (Sending SIGQUIT does work to terminate Neomutt.)

## Steps to Reproduce

Minimal neomuttrc:

```
# Paths --------------------
set mbox_type = Maildir  # mailbox type
set folder = ~/.mail/sorenbjornstad.com/
set spool_file = "~/.mail/sorenbjornstad.com/inbox"
set pager_format = "%s%>"
```

Launch neomutt and open a message.

Earlier it seemed to only happen on *most* emails; now it is happening on every email I try. I'm guessing the partial reproducibility was due to something in my RC that got removed by the minimal version above, but if you have trouble reproducing let me know and I can attach an email where it happens.

When I cleaned up my RC and changed `pager_format` to an invalid value by accident, as explained above.

## NeoMutt Version

```
NeoMutt 20240425
Copyright (C) 2015-2024 Richard Russon and friends

System: Linux 5.10.223-3-MANJARO (x86_64)
ncurses: ncurses 6.5.20240427 (compiled with 6.4.20230520)
libidn2: 2.3.7 (compiled with 2.3.7)
GPGME: 1.23.2
GnuTLS: 3.8.5
libnotmuch: 5.6.0
compression: lz4, zlib, zstd

Configure options: --prefix=/usr --sysconfdir=/etc --libexecdir=/usr/lib --gpgme --sqlite --autocrypt --lua --notmuch --gss --gnutls --sasl --with-ui=ncurses --with-idn2=/usr --disable-idn --idn2 --lmdb --kyotocabinet --gdbm --ubsan --lz4 --zlib --zstd

Compilation CFLAGS: -march=x86-64 -mtune=generic -O2 -pipe -fno-plt -fexceptions         -Wp,-D_FORTIFY_SOURCE=2 -Wformat -Werror=format-security         -fstack-clash-protection -fcf-protection -g -ffile-prefix-map=/build/neomutt/src=/usr/src/debug/neomutt -flto=auto -fno-delete-null-pointer-checks -D_ALL_SOURCE=1 -D_GNU_SOURCE=1 -D__EXTENSIONS__ -D_XOPEN_SOURCE_EXTENDED -I/usr/include/lua5.3 -I/usr/include -I/usr/include -I/usr/include -DNCURSES_WIDECHAR -I/usr/include -I/usr/include/p11-kit-1 -I/usr/include -I/usr/include -fsanitize=undefined -fno-omit-frame-pointer -fno-common -g -O0

Compile options:
  -gsasl +gss +hcache -homespool +idn +inotify -locales_hack +lua -mixmaster 
  +nls +notmuch -openssl +pgp +regex +sasl +smime +sqlite +truecolor 

Devel options:
  ubsan 

MAILPATH="/var/mail"
SENDMAIL="/usr/sbin/sendmail"
SYSCONFDIR="/etc"
```

## Extra Info

* Operating System and its version – Manjaro Linux, latest rolling release (just updated this morning)
* Were you using multiple copies of NeoMutt at once? – no
* Were you using 'screen' or 'tmux'? – no
* Is your email local (maildir) or remote (IMAP)? – Maildir, see RC above

gahr:
Fixes #4377

There were actually two bugs here:

1. the first loop went on infinitely if `pad_len` is 0
2. the second loop incremented the `buf` pointer, probably a left-over of thinking that `buf` is a `char *` instead of a `struct Buffer`

### DIFF
--- a/expando/node_padding.c
+++ b/expando/node_padding.c
@@ -79,18 +79,21 @@ int pad_string(const struct ExpandoNode *node, struct Buffer *buf, int max_cols)
   const int pad_cols = mutt_strnwidth(node->start, pad_len);
   int total_cols = 0;
 
-  while (pad_cols <= max_cols)
+  if (pad_len != 0)
   {
-    buf_addstr_n(buf, node->start, pad_len);
+    while (pad_cols <= max_cols)
+    {
+      buf_addstr_n(buf, node->start, pad_len);
 
-    max_cols -= pad_cols;
-    total_cols += pad_cols;
+      max_cols -= pad_cols;
+      total_cols += pad_cols;
+    }
   }
 
-  for (; (max_cols > 0); buf++, max_cols--)
+  if (max_cols > 0)
   {
-    buf_addch(buf, ' ');
-    total_cols++;
+    buf_add_printf(buf, "%*s", max_cols, "");
+    total_cols += max_cols;
   }
 
   return total_cols;


roccoblues (approved):
nice

expando/node_padding.c
```diff
@@ -79,18 +79,21 @@ int pad_string(const struct ExpandoNode *node, struct Buffer *buf, int max_cols)
...
+      total_cols += pad_cols;
+    }
   }
 
-  for (; (max_cols > 0); buf++, max_cols--)
+  if (max_cols > 0)
   {
-    buf_addch(buf, ' ');
-    total_cols++;
+    buf_add_printf(buf, "%*s", max_cols, "");
```
g0mb4, line 95: I think the padding with spaces was @flatcap s addition and padding with "empty string" is a bit stange (for me), why dont we use `buf_add_printf(buf, "%*s", max_cols, " ");`  instead?

g0mb4 (approved):

expando/node_padding.c
gahr: I think you are misunderstanding the format string. The "padding with spaces" comes from the fact that the minimum field width will add spaces if the provided string doesn't fill the width. It doesn't repeat the provided string to fill in the width. So, to me, providing an empty string is less confusing than printing one space then filling in the remaining width with spaces, because the two uses of a space aren't related:
* don't print anything, but fill in with spaces (because that's what the width specifier does)
* print a space, then fill in with spaces (not related to the fact that I printed a space anyway)

not my clearer answer ever, I realize.. I hope it clarifies anyway

g0mb4: Sure, I understood your intent. It it perfectly fine solution, I just wanted to know the reasoning behind it :)

PR #4386, opened 2024-09-16
notmuch: Fix nm_query_window_available boolean return value

rbmarliere:
Currently, if the user simply set nm_query_window_duration=1 or higher in his configuration will be enough to enable windowed queries. But he might want to use folder hooks that disable conditionally for some patterns and for that he should be able to only set nm_query_window_enable=no.

### DIFF
--- a/notmuch/notmuch.c
+++ b/notmuch/notmuch.c
@@ -1640,7 +1640,7 @@ bool nm_query_window_available(void)
   const short c_nm_query_window_duration = cs_subset_number(NeoMutt->sub, "nm_query_window_duration");
   const bool c_nm_query_window_enable = cs_subset_bool(NeoMutt->sub, "nm_query_window_enable");
 
-  return c_nm_query_window_enable || (c_nm_query_window_duration > 0);
+  return c_nm_query_window_enable && (c_nm_query_window_duration > 0);
 }
 
 /**


roccoblues (approved):
This seems sensible to me. But I don't use notmuch myself so I'd prefer to wait for someone else to also have a look.

roccoblues:
Oh wait. I just realized that this doesn't match the [documentation](https://neomutt.org/feature/notmuch):

> Duration between start and end dates for windowed notmuch query. This corresponds to a bounded notmuch date: query. See notmuch-search-terms manual page for more info. Value of 0 disables windowed queries unless nm_query_window_enable=yes

gahr:
I don't use `notmuch` myself, and I have very little idea how it works.. but this got me asking myself: why do we need both params? And apparently in some places we say that windowing is enabled even when duration is 0:
https://github.com/neomutt/neomutt/blob/508d2216b4e5994b53c031a5800afa6c0333b87e/notmuch/query.c#L164

This is called from here, where we pass c_nm_query_window_enable:

https://github.com/neomutt/neomutt/blob/508d2216b4e5994b53c031a5800afa6c0333b87e/notmuch/notmuch.c#L299-L302

So, I have a slight hunch that this change might not be correct?

gahr:
@austin-ray perhaps you can shed some light on the rationale behind having both configs, and on the use cases for wanting windowing with a duration of 0?

rbmarliere:
My confusion was thinking that nm_query_window_enable would simply enable windowed queries, not only for the case where nm_query_window_duration is 0. Thank you all for the feedback.

PR #4391, opened 2024-09-24
Don't return the result of `realpath` on failure

gahr:
According to POSIX, the output argument of `realpath(3)` is undefined on failure. The Linux man page agrees, but apparently if the failure is caused by EACCESS or ENOENT, the resolved path is still provided as output. This is also the behaviour on FreeBSD, which specifically says that on failure the path that caused the issue is left in the output.

Anyway, we should not rely on anything being in the output argument on failure, and we should leave the path argument alone.

Follow up of #4351.

### DIFF
--- a/mutt/path.c
+++ b/mutt/path.c
@@ -320,15 +320,15 @@ char *mutt_path_dirname(const char *path)
 }
 
 /**
- * mutt_path_to_absolute - Convert relative filepath to an absolute path
- * @param path      Relative path
- * @param reference Absolute path that \a path is relative to
- * @retval true  Success
- * @retval false Failure
+ * mutt_path_to_absolute - Convert a relative path to its absolute form
+ * @param[in,out] path      Relative path
+ * @param[in]     reference Absolute path that \a path is relative to
+ * @retval true  Success, path was changed into its absolute form
+ * @retval false Failure, path is untouched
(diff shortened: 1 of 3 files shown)


mutt/path.c
```diff
@@ -347,15 +347,22 @@ bool mutt_path_to_absolute(char *path, const char *reference)
...
-  path = realpath(buf_string(abs_path), path);
-  if (!path && (errno != ENOENT))
+  char rpath[PATH_MAX + 1] = { 0 };
+  const char *result = realpath(buf_string(abs_path), rpath);
+  buf_pool_release(&abs_path);
+  if (result == NULL)
   {
-    mutt_perror(_("Error: converting path to absolute"));
-    buf_pool_release(&abs_path);
+    if (errno != ENOENT)
```
roccoblues, line 355: I wonder if we should exclude this case at all. Is it common that `mutt_path_to_absolute` gets called with a non-existing path?

roccoblues (approved):
Definitely makes sense. 👍🏼

mutt/path.c
gahr: This was introduced in #1293. I have no idea why. @flatcap ?

flatcap (approved):

PR #4399, opened 2024-10-08
fix duplicate save-hook

flatcap:
Fixes: #4258
(and fixes a memory leak)

**Test config**:
```
save-hook . +wrong
save-hook . +right
```

The second `save-hook` should overwrite the first, so `<save-message>` would show the prompt `+right`

---

When a duplicate pattern is found, the `hook->command` was being replaced, but **not** the `hook->expando`.


roccoblues (approved):
Looks good. Also verified that it works as expected.

PR #4427, opened 2024-11-06
Alias/Query: Upgrade tagging

flatcap:
Two minor upgrades:

**Query Dialog**:
Allow tagging aliases using <kbd>\<space\></kbd> as well as <kbd>t</kbd>
This brings it in line with the Alias dialog

**Alias/Query Dialogs**:
Add two functions: `<tag-pattern>` and `<untag-pattern>`.
This copies the behaviour of the Index dialog.

### DIFF
--- a/alias/dlg_alias.c
+++ b/alias/dlg_alias.c
@@ -273,7 +273,7 @@ static int alias_alias_observer(struct NotifyCallback *nc)
 
     if (alias_array_count_visible(&mdata->ava) != ARRAY_SIZE(&mdata->ava))
     {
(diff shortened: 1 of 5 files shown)


gahr:
Nice! I'll test it in the morning. I was looking at the alias/query code and was wondering: shouldn't they be the same dialog with different data sources? Almost..

flatcap:
> shouldn't they be the same dialog with different data sources?

That should be possible with the modern dialogs.
Almost all the working is templated, `simple_dialog_new()`, or abstracted, `alias_function_dispatcher()`.

gahr (approved):

PR #4429, opened 2024-11-08
buffer: Refactor the string comparison functions

flatcap:
**NOTE**: Post-Release -- This is quite simple, but I'll queue it up and merge it after the next release.

It's a lot of diff, but it's quite simple.
It eliminates quite a lot of `buf_string()`s from string comparisons.

Simplify the name of the **Buffer**-only comparision: `buf_str_equal()` to `buf_equal()`
Then use the old name, `buf_str_equal()`, for a `Buffer` vs string (`char *`) comparison.

Then repeat the same for the case-insensitive versions.

---

1. cb0e61b39 buffer: rename `buf_str_equal()` to `buf_equal()`
1. 7e286b902 buffer: create `buf_str_equal()`
1. 59c9298db buffer: rename `buf_istr_equal()` to `buf_iequal()`
1. 00c280b2c buffer: create `buf_istr_equal()`

### DIFF
--- a/address/address.c
+++ b/address/address.c
@@ -849,8 +849,7 @@ bool mutt_addrlist_equal(const struct AddressList *ala, const struct AddressList
 
   while (ana && anb)
   {
-    if (!buf_str_equal(ana->mailbox, anb->mailbox) ||
-        !buf_str_equal(ana->personal, anb->personal))
+    if (!buf_equal(ana->mailbox, anb->mailbox) || !buf_equal(ana->personal, anb->personal))
     {
       break;
     }
--- a/address/address.c
+++ b/address/address.c
@@ -1085,7 +1085,7 @@ size_t mutt_addr_write(struct Buffer *buf, struct Address *addr, bool display)
 
   if (addr->mailbox)
   {
-    if (!mutt_str_equal(buf_string(addr->mailbox), "@"))
+    if (!buf_str_equal(addr->mailbox, "@"))
     {
       const char *a = display ? mutt_addr_for_display(addr) : buf_string(addr->mailbox);
       buf_addstr(buf, a);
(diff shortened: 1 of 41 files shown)


roccoblues (approved):
Looks good. 👍🏼

If I had a wish I'd like to have better names to more easily distinguish the buffer and string versions. But I also don't have a good idea.

flatcap:
Thanks!

Yeah, the naming isn't perfect.
I'd like the utility functions to have the shortest intelligible names possible.

Perhaps `buf_buf_equal()`?

roccoblues:
> Yeah, the naming isn't perfect. I'd like the utility functions to have the shortest intelligible names possible.

yeah makes sense. I'm actually quite happy with the names in this MR. The "problem" (it really isn't a problem and maybe we should just get on with our lifes 😄 ) is more when you look in buffer.h you'll notice that most of the time `str` in the name signifies a `char *` but not always. `buf_addstr` vs. `buf_insert`. Or `buf_startswith` should be `buf_startswith_str` because we could also have a `buf_startwith(Buffer, Buffer)`?
 
> Perhaps `buf_buf_equal()`?

please not :-)

gahr:
I think I'd just get rid of `buf_str_equal` and `buf_istr_equal` and convert the callers to use `mutt_str_equal`, `mutt_istr_equal`, wrapping their buffers into `buf_string`. Let's do separation of concerns and leave string functions into the string module.

flatcap:
Thanks for the feedback.
I _do_ like the tidier callers, but we _do_ already have a **lot** of string comparison functions.

This branch has been archived to https://github.com/neomutt/neomutt-old

Branch: [buffer-string](https://github.com/neomutt/neomutt-old/tree/buffer-string)

PR #4437, opened 2024-11-12
show message preview in compose view

### LINKED ISSUE #4415: Show message preview in compose view to verify attachment list matches message content
## Current Behavior

When composing a message in compose view, we can only see:
- Headers (To, From, Subject, etc)
- List of attachments 
- No preview of the actual message content

## Use Case

A very common email workflow is:
1. Write a message listing attachments, e.g.:
   "Please find attached:
   - Report Q1 2024
   - Financial summary
   - Team metrics"
2. Attach the files
3. Need to verify that all mentioned attachments are actually attached

Currently this requires:

- Press 'v' to check message content
- Switch back to compose view to see attachment list 
- Mentally compare the two lists
- Risk missing attachments because the lists are never visible simultaneously

This context switching:
- Makes it easy to miss attachments
- Increases cognitive load of comparing two separate views
- Slows down the composition process
- Could lead to embarrassing emails with missing attachments

## Desired Behavior
Show message preview below headers/attachments in compose view so users can:
- See their list of promised attachments
- Compare against actual attached files
- Verify everything matches before sending
- All in a single view without switching contexts

## Implementation Notes
This could be implemented as:
1. A new config option like `compose_show_preview` (boolean)
2. A preview pane below headers/attachments
3. Optional toggling via keybinding

## Current Workarounds
- Press 'v' to view content then switch back to check attachments
- Use external editor with $edit_headers
- No way to see both simultaneously in native compose view

## Environment
- NeoMutt version: 20240425
- Terminal: xfce4-terminal 0.8.10 (Xfce 4.16)

roccoblues:
This adds a message preview to the compose dialog. The preview is enabled by default but can be disabled with `compose_show_preview=false`. Two additional config variables control the behavior:

*   `compose_preview_min_rows`: The minimum number of rows that must be available for the message preview window to be shown.
*   `compose_preview_above_attachments`: Show the message preview above the attachments list. By default, it is shown below it.

If the message is larger than the number of available rows in the window, the content can be scrolled using:

* `preview-page-down` (default `<pagedown>`)
* `preview-page-up` (default `<pageup>`)

Example screenshot (with `compose_preview_above_attachments=true`):
Screenshot 2024-11-25 at 07 32 43

Initially I wanted to add text wrapping to the window but I decided to skip that for the first iteration because it wasn't non-trivial.

fixes #4415

### DIFF
--- a/compose/attach.c
+++ b/compose/attach.c
@@ -5,6 +5,7 @@
  * @authors
  * Copyright (C) 2021-2024 Richard Russon <rich@flatcap.org>
  * Copyright (C) 2023-2024 Tóth János <gomba007@gmail.com>
+ * Copyright (C) 2024 Dennis Schön <mail@dennis-schoen.de>
  *
  * @copyright
  * This program is free software: you can redistribute it and/or modify it under
@@ -236,9 +237,32 @@ static int compose_make_entry(struct Menu *menu, int line, int max_cols, struct 
 }
 
 /**
+ * attach_recalc - Recalculate the Window data - Implements MuttWindow::recalc() - @ingroup window_recalc
+ */
+static int attach_recalc(struct MuttWindow *win)
+{
+  struct Menu *menu = win->wdata;
+  struct ComposeAttachData *adata = menu->mdata;
+
+  const int cur_rows = win->state.rows;
+  const int new_rows = adata->actx->idxlen;
+
+  if (new_rows != cur_rows)
+  {
+    win->req_rows = new_rows;
+    mutt_window_reflow(win->parent);
+    menu_adjust(menu);
+  }
+
+  win->actions |= WA_REPAINT;
+  mutt_debug(LL_DEBUG5, "recalc done, request WA_REPAINT\n");
+  return 0;
+}
+
+/**
  * attach_new - Create the Attachments Menu
- * @param parent Parent Window
- * @param shared Shared compose data
+ * @param parent     Parent Window
+ * @param shared     Shared compose data
  */
 struct MuttWindow *attach_new(struct MuttWindow *parent, struct ComposeSharedData *shared)
 {
@@ -264,4 +288,30 @@ struct MuttWindow *attach_new(struct MuttWindow *parent, struct ComposeSharedDat
   adata->menu = menu;
 
   return win_attach;
+}
+
+/**
+ * attachment_size_fixed - Make the Attachment Window fixed-size
+ * @param win Attachment Window
+ */
+void attachment_size_fixed(struct MuttWindow *win)
+{
+  if (!win || (win->size == MUTT_WIN_SIZE_FIXED))
+    return;
+
+  win->size = MUTT_WIN_SIZE_FIXED;
+  win->recalc = attach_recalc;
+}
+
+/**
+ * attachment_size_max - Make the Attachment Window maximised
+ * @param win Attachment Window
+ */
+void attachment_size_max(struct MuttWindow *win)
+{
+  if (!win || (win->size == MUTT_WIN_SIZE_MAXIMISE))
+    return;
+
+  win->size = MUTT_WIN_SIZE_MAXIMISE;
+  win->recalc = NULL;
 }
--- a/compose/dlg_compose.c
+++ b/compose/dlg_compose.c
@@ -5,6 +5,7 @@
  * @authors
  * Copyright (C) 2017-2023 Richard Russon <rich@flatcap.org>
  * Copyright (C) 2021 Pietro Cerutti <gahr@gahr.ch>
+ * Copyright (C) 2024 Dennis Schön <mail@dennis-schoen.de>
  *
  * @copyright
  * This program is free software: you can redistribute it and/or modify it under
@@ -40,6 +41,7 @@
  * - @ref envelope_window
  * - @ref gui_sbar
  * - @ref compose_attach
+ * - @ref compose_preview
  * - @ref compose_cbar
  *
  * ## Data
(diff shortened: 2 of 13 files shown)


flatcap:
**From IRC**

@roccoblues
> can we make the attachments menu smaller? I can't find the code that sets that.
> Ideally I'd want it to only take up the space needed to show all attachments
> right now there's lots for whitespace below it

@flatcap
> currently, the Envelope Window is MUTT_WIN_SIZE_FIXED.  we calculate how much space we need and *set* that size...
> the Bars are also fixed, at one char high...
> the Attachment Window is MUTT_WIN_SIZE_MAXIMISE, so when mutt_window_reflow() gets called, it gets all the remaining space
> the Envelope Window is created in env_window_new().  it sets Window.recalc = env_recalc
> whenever we change something in the Envelope, we trigger a notification that's observed by, say, env_email_observer()
> the observers can set WA_RECALC and or WA_REPAINT
> Window redraws happen in two stages, recalc() then repaint()
> env_recalc() works out how much space is needed, sets this new *fixed* size, then forcibly *reflow()s* the layout
> as a MAXIMISE Window, the Attachment Window gets what's leftover
> corollary :-)  Make the Attachment Window FIXED and copy'n'paste all the layout code

compose/attach.c
```diff
@@ -234,15 +234,47 @@ static int compose_make_entry(struct Menu *menu, int line, int max_cols, struct
...
-struct MuttWindow *attach_new(struct MuttWindow *parent, struct ComposeSharedData *shared)
+struct MuttWindow *attach_new(struct MuttWindow *parent,
+                              struct ComposeSharedData *shared, bool fixed_size)
 {
   struct MuttWindow *win_attach = menu_window_new(MENU_COMPOSE, NeoMutt->sub);
 
+  if (fixed_size)
+  {
+    // menu_window_new creates a maximised window, change it to fixed size
+    // XXX(ds): this feels a bit hacky
```
flatcap, line 273: We _could_ parameterise `menu_window_new()`, but this'll do for now.

compose/preview.c
```diff
@@ -0,0 +1,358 @@
...
+  {
+    mutt_perror("%s", e->body->filename);
+    return;
+  }
+
+  mutt_window_clear(win);
+
+  // XXX(ds): I'm not particularly happy with this flag but so far I couldn't
+  //          think of a better way.
+  wdata->more_content = false;
```
flatcap, line 130: We _could_ parse the file once and keep track of rows vs byte offsets, but...

compose/preview.c
```diff
@@ -0,0 +1,358 @@
...
+
+  // XXX(ds): We read the message from disk on every repaint. I _think_ this
+  //          is fine as a repaint should only happen when a new attachment is
+  //          added or the number of headers in the envelope change. Both not
+  //          something that happens on every email and even if, not many times.
+  //
+  //          The alternative would be to cache the content in PreviewWindowData
+  //          but that would be much more involved and could also get ugly with
+  //          large messages.
```
flatcap, line 139: it's generally going to be a small file that doesn't get reviewed.

**Emails** may get large when files are attached, but the message text is seldom large.
We can afford a little inefficiency here.

flatcap:
I've added one more todo, ["Create Feature Page"](https://neomutt.org/feature).
They're part of the main docs and heavily templated, so you can copy'n'paste quite easily.
They run between `<sect1 id="account-cmd">` and `<sect1 id="autocryptdoc">`

Thanks!

compose/dlg_compose.c
```diff
@@ -265,23 +266,37 @@ static struct MuttWindow *compose_dlg_init(struct ConfigSubset *sub,
...
   struct MuttWindow *win_cbar = cbar_new(shared);
   struct MuttWindow *win_abar = sbar_new();
   sbar_set_title(win_abar, _("-- Attachments"));
 
+  struct MuttWindow *win_preview = NULL;
+  struct MuttWindow *win_preview_bar = NULL;
+  if (c_show_preview)
+  {
+    win_preview = preview_window_new(e, shared);
+    win_preview_bar = sbar_new();
```
gahr, line 282:
```c
sbar_set_title(win_preview_bar, _("-- Preview"));\
```

compose/dlg_compose.c
```diff
@@ -265,23 +266,37 @@ static struct MuttWindow *compose_dlg_init(struct ConfigSubset *sub,
...
+
   const bool c_status_on_top = cs_subset_bool(sub, "status_on_top");
   if (c_status_on_top)
   {
     mutt_window_add_child(dlg, win_cbar);
     mutt_window_add_child(dlg, win_env);
+    mutt_window_add_child(dlg, win_preview_bar);
+    mutt_window_add_child(dlg, win_preview);
     mutt_window_add_child(dlg, win_abar);
     mutt_window_add_child(dlg, win_attach);
```
gahr, line 293:
```suggestion
    mutt_window_add_child(dlg, win_abar);
    mutt_window_add_child(dlg, win_attach);
    mutt_window_add_child(dlg, win_preview_bar);
    mutt_window_add_child(dlg, win_preview);
```

compose/dlg_compose.c
```diff
@@ -265,23 +266,37 @@ static struct MuttWindow *compose_dlg_init(struct ConfigSubset *sub,
...
   }
   else
   {
     mutt_window_add_child(dlg, win_env);
+    mutt_window_add_child(dlg, win_preview_bar);
+    mutt_window_add_child(dlg, win_preview);
     mutt_window_add_child(dlg, win_abar);
     mutt_window_add_child(dlg, win_attach);
```
gahr, line 301:
```suggestion
    mutt_window_add_child(dlg, win_abar);
    mutt_window_add_child(dlg, win_attach);
    mutt_window_add_child(dlg, win_preview_bar);
    mutt_window_add_child(dlg, win_preview);
```

gahr (commented):
For POLA's sake, I'd add the preview *under* the attachments

roccoblues:
> For POLA's sake, I'd add the preview _under_ the attachments

I actually prefer it above because it feels much more natural. But I understand the argument so I'm going to make the position configurable.

gahr:
I noticed that changing `compose_show_preview` while in the compose window doesn't cause a redraw.

roccoblues:
> I noticed that changing `compose_show_preview` while in the compose window doesn't cause a redraw.

yeah, there's no config observer yet. It's on my list.

gahr:
> > I noticed that changing `compose_show_preview` while in the compose window doesn't cause a redraw.
> 
> yeah, there's no config observer yet. It's on my list.

I should have read the list :)

compose/dlg_compose.c
roccoblues: done

flatcap:
For rearranging Windows, I forgot about `$sidebar_on_right`:
https://github.com/neomutt/neomutt/blob/c391140ef109a8fdbbc53869cae5a40464a21013/sidebar/observer.c#L308-L324

flatcap:
Here are some potential ways to test window swapping.

- `mutt_window_swap(parent, win1, win2)`

Create a parent with 8 children
- apple, banana, cherry, damson, endive, fig, guava, hawthorn

We could store the name in `MuttWindow.wdata`.
That would allow us to easily check the results.

**Note**: the window layout will need regenerating for every test

### degenerate tests

- `mutt_window_swap(NULL, NULL, NULL)`
- `mutt_window_swap(NULL, apple, NULL)`
- `mutt_window_swap(NULL, NULL, banana)`
- `mutt_window_swap(NULL, apple, banana)`
- `mutt_window_swap(parent, NULL, NULL)`
- `mutt_window_swap(parent, apple, NULL)`
- `mutt_window_swap(parent, NULL, banana)`

### spread out tests

- `mutt_window_swap(parent, cherry, fig)`
- `mutt_window_swap(parent, fig, cherry)`

### neighbouring tests

- `mutt_window_swap(parent, damson, endive)`
- `mutt_window_swap(parent, endive, damson)`

### edge tests, spread out

- `mutt_window_swap(parent, apple, damson)`
- `mutt_window_swap(parent, damson, apple)`

- `mutt_window_swap(parent, endive, hawthorn)`
- `mutt_window_swap(parent, hawthorn, endive)`

### edge tests, neighbouring

- `mutt_window_swap(parent, apple, banana)`
- `mutt_window_swap(parent, banana, apple)`

- `mutt_window_swap(parent, guava, hawthorn)`
- `mutt_window_swap(parent, hawthorn, guava)`

### spread out tests, sharing a neighbour

- `mutt_window_swap(parent, banana, damson)`
- `mutt_window_swap(parent, damson, banana)`

<details>
<summary>Gratuitous Vim Syntax Highlighting</summary>

image

</details>

flatcap:
Rebased, added "Limitations" section:

https://github.com/neomutt/neomutt/blob/3a4736ca7c4c94a27ae895fb8581bb0c14658a25/docs/manual.xml.head#L13562-L13580

flatcap (approved):

PR #4438, opened 2024-11-13
autocrypt: fix copy_normalize_addr()

flatcap:
`copy_normalize_addr()` creates a new `Address` for the result.
`mutt_addr_new()` doesn't create any `Buffer`s in the `Address` so create one manually.

Only `Address.mailbox` is needed.

### DIFF
--- a/autocrypt/db.c
+++ b/autocrypt/db.c
@@ -210,8 +210,7 @@ static struct Address *copy_normalize_addr(struct Address *addr)
    * The normalize_addrlist above is extended to work on a list
    * because of requirements in autocrypt.c */
 
-  struct Address *norm_addr = mutt_addr_new();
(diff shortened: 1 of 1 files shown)


autocrypt/db.c
```diff
@@ -211,7 +211,7 @@ static struct Address *copy_normalize_addr(struct Address *addr)
    * because of requirements in autocrypt.c */
 
   struct Address *norm_addr = mutt_addr_new();
-  buf_copy(norm_addr->mailbox, addr->mailbox);
+  norm_addr->mailbox = buf_new(buf_string(addr->mailbox));
```
gahr, line 214: How about

```
struct Address *norm_addr = 
‎mutt_addr_create(NULL, addr->mailbox);
```

flatcap: Ah, much better.  Done.

PR #4440, opened 2024-11-14
remove unused mailbox param from mutt_count_body_parts

roccoblues:
I'd like to use the `mutt_count_body_parts()` in the compose window where I don't have a `Mailbox`.


flatcap:
Good spot!

Back in 2021...

`mutt_count_body_parts()` called `mutt_parse_mime_message()`, which called `mx_msg_open()` which required a `Mailbox`.
`mx_msg_open()` got factored out, but the `Mailbox` got left behind.

flatcap (approved):

PR #4442, opened 2024-11-17
remove unused fields from ComposeSharedData

roccoblues:
Both `ComposeEnvelopeData` and `news` are unused.


flatcap (approved):
Good spot!

PR #4445, opened 2024-11-18
Unify Alias/Query

flatcap:
For two Dialogs that show pretty much the same things, there were a **lot** of differences in the code.

This PR unifies the parsing, display and behaviour of the Alias and Query Dialogs.

**Note**: The defaults for `$alias_format` and `$query_format` have changed (see [comment below](https://github.com/neomutt/neomutt/pull/4445#issuecomment-2483464979))

## alias: unify query parsing

When parsing an Alias, NeoMutt stores the "alias" in `Alias.name` and the "name" and "email" in an `Address`.

Change the Query parsing to store the "name" and "email" the same way.

## alias: add expandos for split address

The Aliases are displayed as combined Name and Email:

- `Adam Apple <aa@example.com>`
- `Bill Banana <bb@example.com>`
- `Caroline Cherry <cc@example.com>`

but the Queries are displayed as separate columns:

- `Adam Apple       <aa@example.com>`
- `Bill Banana      <bb@example.com>`
- `Caroline Cherry  <cc@example.com>`

Add extra Expandos so the user can choose.

## alias: rename expandos

Aliases and Queries use lots of different Expandos for the same things.

Rename lots of Expandos and create new ones to unify both.

**Aliases**:

  SAME:
  - `%E` email
  - `%f` flags
  - `%N` name
  - `%t` tagged
  - `%Y` tags

  NEW:
  - `%a` alias
  - `%A` address
  - `%C` comment
  - `%i` number

  DEPRECATED:
  - `%c` Use `%C` instead
  - `%n` Use `%i` instead
  - `%r` Use `%A` instead

**Query**:

  SAME:
  - `%E` email
  - `%t` tagged
  - `%Y` tags

  NEW:
  - `%A` address
  - `%C` comment
  - `%i` number
  - `%N` name

  DEPRECATED:
  - `%a` Use `%E` instead
  - `%c` Use `%i` instead
  - `%e` Use `%C` instead
  - `%n` Use `%N` instead

## alias: add sort methods

Allow the use to sort Aliases/Queries by Real Name or Email Address.

- "Sort (a)lias, (n)ame, (e)mail or (u)nsorted?"

## alias: merge expando callbacks

Now that the Alias and Query expandos have been unified, we don't need
the `query_*()` Expando callbacks.

## alias: rename expando callbacks

Rename the Expando callbacks to match their purpose.

The old names were closely tied to the Expando letters, which aren't particularly memorable.

The new functions are named by Domain and Member, e.g.

- `alias_comment()`
- `alias_view_tagged()`

## docs: alias tags

Add some docs for the Alias `tags:` feature.

### DIFF
--- a/docs/manual.xml.head
+++ b/docs/manual.xml.head
@@ -4100,9 +4100,21 @@ alternates -group me -group work address3
         <arg choice="opt" rep="repeat">
           <replaceable class="parameter">, address</replaceable>
         </arg>
-        <arg choice="opt">
-          <replaceable class="parameter"># comment</replaceable>
-        </arg>
+        <group choice="opt">
+          <arg choice="plain">
+            <replaceable class="parameter"># </replaceable>
+          </arg>
+          <group choice="opt">
+            <arg choice="plain">
+              <replaceable class="parameter">comments</replaceable>
+            </arg>
+          </group>
+          <group choice="opt">
+            <arg choice="plain">
+              <replaceable class="parameter">tags:...</replaceable>
+            </arg>
+          </group>
+        </group>
         <command>unalias</command>
         <arg choice="opt" rep="repeat">
           <option>-group</option>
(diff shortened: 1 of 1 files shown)


gahr (approved):
Nice! I have smoke-tested it and it seems fine. I can run a bit on this branch

gahr:
I noticed that changing `compose_show_preview` while in the compose window doesn't cause a redraw.

flatcap:
Thanks.
If you don't immediately see any changes, then it's probably OK :-)

PR #4449, opened 2024-11-21
add mutt_window_swap()

roccoblues:
This adds `mutt_window_swap()` to swap the location of two windows. 

I need this for https://github.com/neomutt/neomutt/pull/4437 but I figured we could add it separately.

Thanks to @flatcap for coming up with all the test cases. This version differs a bit from the initial one. 😏

### DIFF
--- a/gui/mutt_window.c
+++ b/gui/mutt_window.c
@@ -4,6 +4,7 @@
  *
  * @authors
  * Copyright (C) 2018-2023 Richard Russon <rich@flatcap.org>
+ * Copyright (C) 2024 Dennis Schön <mail@dennis-schoen.de>
  *
  * @copyright
  * This program is free software: you can redistribute it and/or modify it under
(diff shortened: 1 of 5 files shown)


roccoblues:
Ok, I should free some stuff in the tests. 😆

test/gui/swap.c
```diff
@@ -0,0 +1,416 @@
...
+  "cherry",
+  "damson",
+  "endive",
+  "fig",
+  "guava",
+  "hawthorn"
+};
+// clang-format on
+
+static const int num_children = sizeof(initial_order) / sizeof(initial_order[0]);
```
gahr, line 44: https://github.com/neomutt/neomutt/blob/7605466f4ab5b361e4903d6b4744f83041c3a0e1/mutt/memory.h#L38

gui/mutt_window.c
```diff
@@ -808,3 +809,73 @@ bool window_status_on_top(struct MuttWindow *panel, struct ConfigSubset *sub)
...
+                      struct MuttWindow *win2)
+{
+  if (!parent || !win1 || !win2)
+    return false;
+
+  struct MuttWindow *win1_next = TAILQ_NEXT(win1, entries);
+  if (win1_next == win2)
+  {
+    // win1 is directly in front of win2, move it behind
+    TAILQ_REMOVE(&parent->children, win1, entries);
```
gahr, line 849: This is a good optimization, but what if win1 and/or win2 aren't children of the parent?

roccoblues: If only one window is not a child of the parent we won't reach this. Because then they can't, or at least really shouldn't, be connected.

But you're right, if both are next to each other but not children of the parent this will fail. I'll fix it tomorrow.

gui/mutt_window.c
```diff
@@ -808,3 +809,73 @@ bool window_status_on_top(struct MuttWindow *panel, struct ConfigSubset *sub)
...
 }
+
+/**
+ * mutt_window_is_child - Check whether a window is a child of a parent
+ * @param parent Parent Window
+ * @param win    Window
+ * @retval true  Window is a child of the parent
+ * @retval false Window is not a child of the parent
+ */
+static bool mutt_window_is_child(struct MuttWindow *parent, struct MuttWindow *win)
```
flatcap, line 820: We could probably get away without this.
All the children get added by `mutt_window_add_child()` which sets `MuttWindow.parent`.
Checking `win1.parent == win2.parent` should be enough.

test/gui/swap.c
```diff
@@ -0,0 +1,431 @@
...
+  TAILQ_FOREACH(win, &parent->children, entries)
+  {
+    if (mutt_str_equal(win->wdata, name))
+      return win;
+  }
+
+  ASSERT(false);
+}
+
+static void check_order(struct MuttWindow *parent, const char **expected)
```
flatcap, line 87: All the removing/inserting is happening through the `TAILQ_` functions, so it ought to be safe,
but (if we were being paranoid) this function only checks that the **forward** links are correct
(`TAILQ` is doubly linked)

flatcap (approved):
Good job!
I'm happy with the code as it is :-)
We can obsess over the details some other day.

gui/mutt_window.c
roccoblues: Ah, I didn't even think of `win.parent`. I've remove `mutt_is_child` now.

test/gui/swap.c
roccoblues: I also think that would be pretty paranoid. If you really mess with the internal structure like that, you're on your own. 😆

PR #4452, opened 2024-11-25
Smartos fixes

pekdon:
* **What does this PR do?**

Fix two small compilation issues on SmartOS, use of tm_gmtoff (not available) and index_t already defined in sys/types.h.

### DIFF
--- a/auto.def
+++ b/auto.def
@@ -430,6 +430,10 @@ if {1} {
   cc-with {-includes sys/stat.h} {
     cc-check-members "struct stat.st_atim.tv_nsec"
   }
+
+  cc-with {-includes time.h} {
+    cc-check-members "struct tm.tm_gmtoff"
+  }
 }
 ###############################################################################
 
--- a/docs/config.c
+++ b/docs/config.c
@@ -2258,6 +2258,8 @@
 **                is a bang ("!"), the date is formatted ignoring any locale
 **                settings.  Note that the sender's time zone might only be
 **                available as a numerical offset, so "%Z" behaves like "%z".
+**                %{fmt} behaves like %[fmt] on systems where struct tm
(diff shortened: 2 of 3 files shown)


hdrline.c
```diff
@@ -348,7 +348,9 @@ static void index_email_date(const struct ExpandoNode *node, const struct Email
       int offset = (e->zhours * 3600 + e->zminutes * 60) * (e->zoccident ? -1 : 1);
       const time_t now = e->date_sent + offset;
       tm = mutt_date_gmtime(now);
+#ifdef HAVE_STRUCT_TM_TM_GMTOFF
       tm.tm_gmtoff = offset;
+#endif /* HAVE_STRUCT_TM_TM_GMTOFF */
```
gahr, line 353: Just skipping setting `tm_gmtoff` sounds confusing to me. As an alternative, I suggest to fallback to using `SENT_LOCAL` if an offset cannot be set. This can be documented (`%{fmt}` behaves like `%[fmt]` on systems where `struct tm` doesn't have a `tm_gmtoff` member.

pekdon: Updated, is that close to what you had in mind @gahr?

gahr: Yep, that works. You could simplify it even further to

```c
    case SENT_SENDER:
    {
#ifdef HAVE_STRUCT_TM_TM_GMTOFF
      int offset = (e->zhours * 3600 + e->zminutes * 60) * (e->zoccident ? -1 : 1);
      const time_t now = e->date_sent + offset;
      tm = mutt_date_gmtime(now);
      tm.tm_gmtoff = offset;
      break;
#else
      FALLTHROUGH;
#endif
    }

flatcap:
I got around to renaming all the Expando callbacks and rebased the PR over them.

PR #4464, opened 2024-12-03
Remove BOM from UTF-8 text

gahr:
It is [possible](https://en.wikipedia.org/wiki/Byte_order_mark#cite_ref-5) for UTF-8 to have a BOM. We currently display it as a question mark.

### DIFF
--- a/mutt/mbyte.c
+++ b/mutt/mbyte.c
@@ -441,6 +441,10 @@ int mutt_mb_filter_unprintable(char **s)
       memset(&mbstate1, 0, sizeof(mbstate1));
       wc = ReplacementChar;
     }
+    if (CharsetIsUtf8 && IsBOM(wc))
+    {
+      continue;
+    }
     if (!IsWPrint(wc))
       wc = '?';
     else if (CharsetIsUtf8 && mutt_mb_is_display_corrupting_utf8(wc))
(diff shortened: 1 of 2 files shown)


flatcap:
From IRC:

@gahr
> do people think this is the proper place to handle this?
  
@flatcap
> probably :-) ...
> I'm trying to think where a BOM could get into NeoMutt
> composing an email, obviously
> reading config?
> `convert_string()` _does_ seem like a sensible place to ignore it
> history: commands cut'n'pasted into the `:` prompt
> hmm... it might be an idea to check what happens to BOM text pasted into `:`

gahr:
> I'm trying to think where a BOM could get into NeoMutt

In a `Subject: =?utf-8?Q?=EF=BB=BFFooBar?=` :)

gahr:
minimal reproducer

```
Date: Tue, 3 Dec 2024 08:01:27 +0100 (CET)
From: <foo@bar.baz>
Subject: =?utf-8?Q?=EF=BB=BFFoo?=
To: <foo@bar.baz>

foo
```

```
q:Quit  d:Del  u:Undel  s:Save  m:Mail  r:Reply  g:Group  ?:Help                                                                                                                   
   1 O   Dec 03 foo@bar.baz     (0.1K) ?Foo
```

gahr:
The issue is only present on systems where `iswprint(L'\xfeff')` is `0`. This seems to always be the case with FreeBSD's libc, while on glibc the return is `1` when locale is properly set.

flatcap (approved):

Issue #4466, opened 2024-12-05
"Accept always" for self-signed certificates doesn't work


frei-0xff:
When connecting to server with self-signed certificate and selecting "Accept always"
prompt option it doesn't work as intended.
On subsequent runs neomutt shows certificate check warning with "Reject" and "Accept once" options once again.

## NeoMutt Version

```
NeoMutt 20241114
Copyright (C) 2015-2024 Richard Russon and friends
NeoMutt comes with ABSOLUTELY NO WARRANTY; for details type 'neomutt -vv'.
NeoMutt is free software, and you are welcome to redistribute it
under certain conditions; type 'neomutt -vv' for details.

System: Linux 6.6.63-1-lts (x86_64)
ncurses: ncurses 6.5.20240427 (compiled with 6.5.20240427)
libidn2: 2.3.7 (compiled with 2.3.7)
GPGME: 1.24.0
GnuTLS: 3.8.8
libnotmuch: 5.6.0
storage: kyotocabinet, gdbm, lmdb
compression: lz4, zlib, zstd

Configure options: --prefix=/usr --sysconfdir=/etc --libexecdir=/usr/lib --gpgme --sqlite --autocrypt --lua --notmuch --gss --gnutls --sasl --with-ui=ncurses --with-idn2=/usr --disable-idn --idn2 --lmdb --kyotocabinet --gdbm --ubsan --lz4 --zlib --zstd

Compilation CFLAGS: -march=x86-64 -mtune=generic -O2 -pipe -fno-plt -fexceptions         -Wp,-D_FORTIFY_SOURCE=3 -Wformat -Werror=format-security         -fstack-clash-protection -fcf-protection         -fno-omit-frame-pointer -mno-omit-leaf-frame-pointer -g -ffile-prefix-map=/build/neomutt/src=/usr/src/debug/neomutt -flto=auto -std=c11 -fno-delete-null-pointer-checks -D_ALL_SOURCE=1 -D_GNU_SOURCE=1 -D__EXTENSIONS__ -D_XOPEN_SOURCE_EXTENDED -I/usr/include/lua5.3 -I/usr/include -I/usr/include -I/usr/include -DNCURSES_WIDECHAR -I/usr/include -I/usr/include/p11-kit-1 -I/usr/include -I/usr/include -fsanitize=undefined -fno-omit-frame-pointer -fno-common -g -O0

Compile options:
[...]

PR #4467, opened 2024-12-05
fix: Bug with wrong fingerprints in certificate_file

frei-0xff:
Fixes bug with wrong fingerprints been written to
certificate_file when using "Accept always" option.

The issue is that `fpbuf` has to be cleared before calling `tls_fingerprint` because `tls_fingerprint` uses `buf_add_printf` and only adds data to the buffer on every call.

* **What does this PR do?**
Fixes the issue https://github.com/neomutt/neomutt/issues/4466

https://github.com/neomutt/neomutt/issues/4466

### DIFF
--- a/conn/gnutls.c
+++ b/conn/gnutls.c
@@ -578,7 +578,7 @@ static int tls_check_one_certificate(const gnutls_datum_t *certdata,
     {
       if (certerr & CERTERR_HOSTNAME) // Save hostname if necessary
       {
-        buf_strcpy(fpbuf, "");
+        buf_reset(fpbuf);
         tls_fingerprint(GNUTLS_DIG_MD5, fpbuf, certdata);
         fprintf(fp, "#H %s %s\n", hostname, buf_string(fpbuf));
         saved = true;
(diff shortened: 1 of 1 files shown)


frei-0xff:
Bug as far as I can see was introduced by https://github.com/neomutt/neomutt/commit/eb5cb9ccdcea86792913bfb15c6aff13fbe835b8
In earlier code buffer was explicitly cleared before calling `tls_fingerprint` as it should.

conn/gnutls.c
```diff
@@ -578,6 +578,7 @@ static int tls_check_one_certificate(const gnutls_datum_t *certdata,
     {
       if (certerr & CERTERR_HOSTNAME) // Save hostname if necessary
       {
+        buf_strcpy(fpbuf, "");
```
gahr, line 581:
```suggestion
        buf_reset(fpbuf);
```

gahr:
Good catch!

flatcap:
Thanks for the bug-fix, @frei-0xff.

I see that your commit doesn't have a your real name on it.
Anonymity's fine, but if you'd like your name in the credits, you'll need to tell me what it is :-)

frei-0xff:
@flatcap My name is Dmitry Polunin.

flatcap:
Thanks, Dmitry :-)

flatcap (approved):

PR #4481, opened 2024-12-17
Simplify mutt_file_fopen

gahr:
Historically, mutt_file_fopen was trying to be smart when files were opening in write or read/write mode:

"When opening files for writing, make sure the file doesn't already exist to avoid race conditions."

We think this is over-zealous and just makes the code difficult. Here, I'm turning mutt_file_fopen into a simple wrapper around fopen + logging.

### DIFF
--- a/alias/alias.c
+++ b/alias/alias.c
@@ -529,10 +529,6 @@ retry_name:
   }
 
   /* terminate existing file with \n if necessary */
-  if (!mutt_file_seek(fp_alias, 0, SEEK_END))
-  {
-    goto done;
-  }
   if (ftell(fp_alias) > 0)
   {
     if (!mutt_file_seek(fp_alias, -1, SEEK_CUR))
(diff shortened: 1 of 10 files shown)


ossilator:
> We think this is over-zealous and just makes the code difficult.

this might well be correct, but given the gravity of the matter at hand (race conditions often relate to security issues or data loss), i think some more elaboration and possibly even a case-by-case analysis (and therefore splitting of the commit) would seem appropriate.

gahr:
> > We think this is over-zealous and just makes the code difficult.
> 
> this might well be correct, but given the gravity of the matter at hand (race conditions often relate to security issues or data loss), i think some more elaboration and possibly even a case-by-case analysis (and therefore splitting of the commit) would seem appropriate.

I went through the calls one-by-one. I am not going to have a 75-commits PR where I explain why every single mutt_file_fopen with a mode of "w" is safe.

flatcap (approved):

ossilator:
> I went through the calls one-by-one.

that's good to know. why doesn't the commit message mention this, rather than just implying that the original authors had no clue? why aren't you explaining by what criteria you judged? that would inspire a helluva lot more confidence, because it would show that you considered symlink attacks and such stuff.

> I am not going to have a 75-commits PR

well, you could group them by what (type of) file is being accessed.

PR #4485, opened 2024-12-17
lua: fix lua_mutt_call()

flatcap:
Fix a faulty `Buffer `refactor.

When we pass a `Buffer` to the command parsing routines,
we need to `seek(0)` so that they start at the beginning of the string.

**Test NeoMutt config file**: `test.rc`
```
echo "hello"
```

**Test steps:**
- Build with Lua `--lua`
- Run command  `:lua mutt.call("source", "/tmp/foo.muttrc")`

**Expected Result**:
- Message window shows "hello"

### DIFF
--- a/mutt_lua.c
+++ b/mutt_lua.c
@@ -124,6 +124,7 @@ static int lua_mutt_call(lua_State *l)
     buf_addstr(buf, lua_tostring(l, i));
     buf_addch(buf, ' ');
   }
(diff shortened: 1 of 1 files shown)


roccoblues (approved):

roccoblues:
So this was broken for more than half a year without anyone noticing. 🤔

PR #4489, opened 2024-12-28
docs: use roff's CR font instead of C (non-portable) for monospace fonts

charles2910:
C is a non-portable font. Use CR instead which is a "Regular constant width font".

This was spotted by `lintian` tool in the Debian packaging and can be reproduced with `LC_ALL=C.UTF-8 MANROFFSEQ='' MANWIDTH=80 man --warnings -E UTF-8 -l -Tutf8 -Z  /usr/share/man/man5/neomuttrc.5  >/dev/null`. Example:

```
[...]
W: neomutt: groff-message troff:<standard input>:9012: warning: cannot select font 'C' [usr/share/man/man5/neomuttrc.5.gz:217]
W: neomutt: groff-message troff:<standard input>:9014: warning: cannot select font 'C' [usr/share/man/man5/neomuttrc.5.gz:218]
W: neomutt: groff-message troff:<standard input>:9017: warning: cannot select font 'C' [usr/share/man/man5/neomuttrc.5.gz:219]
[...]
```

As far as I was able to [read](https://lists.nongnu.org/archive/html/emacs-orgmode/2024-12/msg00245.html), the only difference it could cause is when rendering HTML/PDF/richer media. For `man` which prints to the terminal - which already uses monospace fonts - it's just a no-op. 

More info available at:

- https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=1043256

- https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=1041809

- https://github.com/cpuguy83/go-md2man/issues/99


flatcap:
@alejandro-colomar 
As the man-page guru, would you like to comment on this?

docs/config.c
```diff
@@ -97,7 +97,7 @@
 /*
 ** .pp
 ** The default file in which to save aliases created by the
-** \fC$<create-alias>\fP function. Entries added to this file are
+** \f[CR]$<create-alias>\fP function. Entries added to this file are
```
alejandro-colomar, line 100: In the Linux man-pages project we completely avoid both \fC and \f[CR].  We use either italics or bold.

Rationale: most of the time, the manual page is read most of the time in the terminal, where everything is monospaced, so marking some identifier as Courier (monospaced) doesn't help the reader at all.  It helps on the PDF, but few people read manual pages as PDF these days.

charles2910: @flatcap do you have any opinions on whether to keep the `CR` or change to something else? Maybe bold **and** italic? What do you use on kernel's documentation in this case @alejandro-colomar? 

I have no strong opinions on this one and used `CR` to keep the same behaviour as before but in a compatible way (both troff and groff support it). After all, I think this predates neomutt by a lot (6cd98f2bb8fbe0b75bf2de3338e034db1de9fded) so I think it was mainly not touched for a while and just existed on the codebase.

flatcap: > most of the time, the manual page is read in the terminal, where everything is monospaced

Oh yeah :-)

> do you have any opinions on whether to keep the `CR` or change to something else?

I think bold would be a reasonable choice for highlighting keywords, functions, etc.
(but I don't feel  too strongly about it).

ossilator: there are actually rules for that: https://man7.org/linux/man-pages/man7/groff_man_style.7.html, section "Font style macros"

alejandro-colomar: Indeed, we have style rules for this in groff_man_style(7).  Here's the rules that @ossilator mentions:

```
       .B [text]
              Set text in bold.  If no argument is given, a one‐line input trap
              is planted; text on the next line, which can be further formatted
              with a macro, is set in bold.

              Use  bold  for  literal portions of syntax synopses, for command‐
              line options in running text, and for  literals  that  are  major
              topics  of  the  subject under discussion; for example, this page
              uses bold for macro, string, and register names.  In  an  .EX/.EE
              example  of  interactive  I/O (such as a shell session), set only
              user input in bold.

       .I [text]
              Set text in an italic or oblique face.  If no argument is  given,
              a  one‐line  input  trap is planted; text on the next line, which
              can be further formatted with a macro, is set  in  an  italic  or
              oblique face.

              Use  italics  for file and path names, for environment variables,
              for C data types, for enumeration or preprocessor constants in C,
              for variant (user‐replaceable) portions of syntax  synopses,  for
              the  first  occurrence (only) of a technical concept being intro‐
              duced, for names of journals and of literary works longer than an
              article, and anywhere a parameter requiring  replacement  by  the
              user  is  encountered.   An  exception involves variant text in a
              context already typeset in italics, such as file  or  path  names
              with replaceable components; in such cases, follow the convention
              of  mathematical typography: set the file or path name in italics
              as usual but use roman for the variant part (see .IR and .RI  be‐
              low),  and  italics again in running roman text when referring to
              the variant material.
```

A rough TL;DR would be: variable names and placeholders in italics, and literals and constants in bold.  Examples of commands, expressions, etc., in italics.  Use of bold should be minimal (think of it like uppercase; abusing it is considered rude).

flatcap: OK, italics it is.

docs/config.c
```diff
@@ -497,8 +497,8 @@
 ** names in the browser mailbox list, using '~' and '='
 ** shortcuts.
 ** .pp
-** The default \fC"alpha"\fP setting of $$browser_sort uses
-** locale-based sorting (using \fCstrcoll(3)\fP), which ignores some
+** The default \fI"alpha"\fP setting of $$browser_sort uses
```
alejandro-colomar, line 500: Since "alpha" is a literal, I'd make it bold.

docs/config.c
```diff
@@ -1099,14 +1099,14 @@
...
-** See also: \fC$$debug_file\fP
+** See also: \fI$$debug_file\fP
 */
 
 { "default_hook", DT_STRING, "~f %s !~P | (~P ~C %s)" },
 /*
 ** .pp
 ** This variable controls how some hooks are interpreted if their pattern is a
-** plain string or a regex. i.e. they don't contain a pattern, like \fC~f\fP
+** plain string or a regex. i.e. they don't contain a pattern, like \fI~f\fP
```
alejandro-colomar, line 1109: This is a literal, so I'd use bold, I think.  It's not a clear case, I think, though.

docs/config.c
```diff
@@ -1187,7 +1187,7 @@
 ** .pp
 ** \fBNote:\fP when using $$sendmail for delivery, you should not enable
 ** this unless you are either using Sendmail 8.8.x or greater or a MTA
-** providing a \fCsendmail(1)\fP-compatible interface supporting the \fC-N\fP option
+** providing a \fIsendmail(1)\fP-compatible interface supporting the \fI-N\fP option
```
alejandro-colomar, line 1190: For command-line flags, you should use bold.

docs/config.c
```diff
@@ -1206,7 +1206,7 @@
 ** .pp
 ** \fBNote:\fP when using $$sendmail for delivery, you should not enable
 ** this unless you are either using Sendmail 8.8.x or greater or a MTA
-** providing a \fCsendmail(1)\fP-compatible interface supporting the \fC-R\fP option
+** providing a \fIsendmail(1)\fP-compatible interface supporting the \fI-R\fP option
```
alejandro-colomar, line 1209: cmd flag bold

docs/config.c
```diff
@@ -1519,14 +1520,14 @@
...
-** For a full listing of defined \fCprintf(3)\fP-like sequences see
+** For a full listing of defined \fIprintf(3)\fP-like sequences see
 ** the section on $$index_format.  See also $$attribution_locale.
 */
 
 { "forward_decode", DT_BOOL, true },
 /*
 ** .pp
-** Controls the decoding of complex MIME messages into \fCtext/plain\fP when
+** Controls the decoding of complex MIME messages into \fItext/plain\fP when
```
alejandro-colomar, line 1530: @g-branden-robinson  What would you use here?  Bold, italics, or roman?  I don't know.

docs/config.c
```diff
@@ -1615,7 +1616,7 @@
 ** entry when expanding the alias.  The default value
 ** will return the string up to the first "," encountered.
 ** If the GECOS field contains a string like "lastname, firstname" then you
-** should set it to "\fC.*\fP".
+** should set it to "\fI.*\fP".
```
alejandro-colomar, line 1619: I would consider this a literal, so bold, I think.  Also, since the config should include the quotes, I'd also make them bold.

docs/config.c
```diff
@@ -2526,7 +2527,7 @@
 ** The default mailbox type used when creating new folders. May be any of
 ** "mbox", "MMDF", "MH" or "Maildir".
 ** .pp
-** This can also be set using the \fC-m\fP command-line option.
+** This can also be set using the \fI-m\fP command-line option.
```
alejandro-colomar, line 2530: cmd flag bold

docs/config.c
```diff
@@ -2587,8 +2588,8 @@
 /*
 ** .pp
 ** This is the string displayed in the "attachment" menu for
-** attachments of type \fCmessage/rfc822\fP.  For a full listing of defined
-** \fCprintf(3)\fP-like sequences see the section on $$index_format.
+** attachments of type \fImessage/rfc822\fP.  For a full listing of defined
```
alejandro-colomar, line 2591: Similar question, @g-branden-robinson.

docs/config.c
```diff
@@ -2597,9 +2598,9 @@
 ** If \fIset\fP, forces NeoMutt to interpret keystrokes with the high bit (bit 8)
 ** set as if the user had pressed the Esc key and whatever key remains
 ** after having the high bit removed.  For example, if the key pressed
-** has an ASCII value of \fC0xf8\fP, then this is treated as if the user had
+** has an ASCII value of \fI0xf8\fP, then this is treated as if the user had
 ** pressed Esc then "x".  This is because the result of removing the
-** high bit from \fC0xf8\fP is \fC0x78\fP, which is the ASCII character
+** high bit from \fI0xf8\fP is \fI0x78\fP, which is the ASCII character
```
alejandro-colomar, line 2603: These are clearly literals, and should be bold.

docs/config.c
```diff
@@ -2637,7 +2638,7 @@
 /*
 ** .pp
 ** When \fIset\fP, the message you are forwarding will be attached as a
-** separate \fCmessage/rfc822\fP MIME part instead of included in the main body of the
+** separate \fImessage/rfc822\fP MIME part instead of included in the main body of the
```
alejandro-colomar, line 2641: @g-branden-robinson

docs/config.c
```diff
@@ -2649,7 +2650,7 @@
 { "mime_forward_decode", DT_BOOL, false },
 /*
 ** .pp
-** Controls the decoding of complex MIME messages into \fCtext/plain\fP when
+** Controls the decoding of complex MIME messages into \fItext/plain\fP when
```
alejandro-colomar, line 2653: gbr

docs/config.c
charles2910: I was about to type that I've changed everything to italics but looking at the diff, it didn't see right. So I was going to go through everything to see were bold should be used. @alejandro-colomar was faster than me. I'll add his suggestions later - not sure if today because I need to finish updating my last year thesis with the reviewer notes and submit before 2025 ;-)

docs/config.c
```diff
@@ -2739,10 +2740,10 @@
 ** This variable specifies domain name or address of NNTP server.
 ** .pp
 ** You can also specify username and an alternative port for each news server,
-** e.g. \fC[[s]news://][username[:password]@]server[:port]\fP
+** e.g. \fI[[s]news://][username[:password]@]server[:port]\fP
```
alejandro-colomar, line 2743: This is an interesting one.

The [] should be roman, since they are not variable.  They convey meanining by saying that the enclosed text is optional.

Also, some parts are literal, and they should be bold.  This is one of the most complex cases you'll find in a man(7) document, and involves 3 fonts in a single input line.

So, I'd use

```man
.RB [[ s ] news:// ][ \fIusername\fP [ :\fIpassword\fP ] @ ] \fIserver\fP [ :\fIport\fP ]
```

using man(7) macros.  Or the following equivalent with escapes:

```
[[\fBs\fP]\fBnews://\fP][\fIusername\fP[\fB:\fP\fIpassword\fP]\fB@\fP]\fIserver\fP[\fB:\fP\fIport\fP]
```

docs/config.c
```diff
@@ -3154,7 +3155,7 @@
 **            string otherwise. Note: This may be used with a %<...> construct.
 ** .dt %r .dd One or more key IDs (or fingerprints if available).
 ** .dt %s .dd Expands to the name of a file containing the signature part
-**            of a \fCmultipart/signed\fP attachment when verifying it.
+**            of a \fImultipart/signed\fP attachment when verifying it.
```
alejandro-colomar, line 3158: gbr

docs/config.c
```diff
@@ -3327,12 +3328,12 @@
 ** gpg --list-keys --with-colons --with-fingerprint
 ** .te
 ** .pp
-** Note: gpg's \fCfixed-list-mode\fP option should not be used.  It
+** Note: gpg's \fIfixed-list-mode\fP option should not be used.  It
```
alejandro-colomar, line 3331: This is a command line option, and should be bold.

BTW, it is missing the leading `--`.

docs/config.c
```diff
@@ -3345,12 +3346,12 @@
...
```
alejandro-colomar, line 3349: Leading `--` is missing.

As a cmd flag, this should be bold.

docs/config.c
```diff
@@ -3421,11 +3422,11 @@
...
 ** If \fIset\fP, signed and encrypted messages will consist of nested
-** \fCmultipart/signed\fP and \fCmultipart/encrypted\fP body parts.
+** \fImultipart/signed\fP and \fImultipart/encrypted\fP body parts.
 ** .pp
 ** This is useful for applications like encrypted and signed mailing
-** lists, where the outer layer (\fCmultipart/encrypted\fP) can be easily
-** removed, while the inner \fCmultipart/signed\fP part is retained.
+** lists, where the outer layer (\fImultipart/encrypted\fP) can be easily
+** removed, while the inner \fImultipart/signed\fP part is retained.
 ** (PGP only)
```
alejandro-colomar, line 3430: gbr

docs/config.c
```diff
@@ -3461,10 +3462,10 @@
 /*
 ** .pp
 ** This command is used to create the detached PGP signature for a
-** \fCmultipart/signed\fP PGP/MIME body part.
+** \fImultipart/signed\fP PGP/MIME body part.
```
alejandro-colomar, line 3465: gbr

docs/config.c
```diff
@@ -3623,9 +3624,9 @@
 { "pop_last", DT_BOOL, false },
 /*
 ** .pp
-** If this variable is \fIset\fP, NeoMutt will try to use the "\fCLAST\fP" POP command
+** If this variable is \fIset\fP, NeoMutt will try to use the "\fILAST\fP" POP command
```
alejandro-colomar, line 3627: @g-branden-robinson what do you think of this one?  I would make it bold, but I'm not entirely sure.

docs/config.c
```diff
@@ -3717,7 +3718,7 @@
 ** .pp
 ** If \fIset\fP, a shell command to be executed if NeoMutt fails to establish
 ** a connection to the server. This is useful for setting up secure
-** connections, e.g. with \fCssh(1)\fP. If the command returns a  nonzero
+** connections, e.g. with \fIssh(1)\fP. If the command returns a  nonzero
```
alejandro-colomar, line 3721: Hmmm, I haven't checked the others.  The `(1)` part should be roman.  The \fI\fP should enclose only the name of the page, and not the parenthesized number.  Please check all such occurences.

docs/config.c
```diff
@@ -3774,21 +3775,21 @@
...
-** Used in connection with the \fC<print-message>\fP function.  If this option
+** Used in connection with the \fI<print-message>\fP function.  If this option
 ** is \fIset\fP, the command specified by $$print_command is executed once for
 ** each message which is to be printed.  If this option is \fIunset\fP,
 ** the command specified by $$print_command is executed only once, and
 ** all the messages are concatenated, with a form feed as the message
 ** separator.
 ** .pp
-** Those who use the \fCenscript\fP(1) program's mail-printing mode will
+** Those who use the \fIenscript\fP(1) program's mail-printing mode will
```
alejandro-colomar, line 3792: Hmmm, it seems there's a mix of style.  This one is good.

docs/config.c
```diff
@@ -3981,20 +3982,20 @@
...
-** The initial \fC"\\["\fP means a literal left-bracket character.
+** bracketed numbers following the prefix, such as \fI"Re[1]: "\fP.
+** The initial \fI"\\["\fP means a literal left-bracket character.
 ** Note the backslash must be doubled when used inside a double
-** quoted string in the neomuttrc.  \fC"[0-9]+"\fP means one or more
-** numbers.  \fC"\\]"\fP means a literal right-bracket.  Finally the
-** whole parenthesized expression has a \fC"*"\fP suffix, meaning it
+** quoted string in the neomuttrc.  \fI"[0-9]+"\fP means one or more
+** numbers.  \fI"\\]"\fP means a literal right-bracket.  Finally the
+** whole parenthesized expression has a \fI"*"\fP suffix, meaning it
```
alejandro-colomar, line 3994: @g-branden-robinson Any opinion on these?

docs/config.c
```diff
@@ -3981,20 +3982,20 @@
...
 ** can occur zero or more times.
 ** .pp
 ** The last part matches a colon followed by an optional space or
-** tab.  Note \fC"\t"\fP is converted to a literal tab inside a
+** tab.  Note \fI"\t"\fP is converted to a literal tab inside a
```
alejandro-colomar, line 3998: This is a literal.  Bold.

docs/config.c
```diff
@@ -4054,7 +4055,7 @@
 { "resume_draft_files", DT_BOOL, false },
 /*
 ** .pp
-** If \fIset\fP, draft files (specified by \fC-H\fP on the command
+** If \fIset\fP, draft files (specified by \fI-H\fP on the command
```
alejandro-colomar, line 4058: cmd flag bold

docs/config.c
```diff
@@ -4064,7 +4065,7 @@
 { "resume_edited_draft_files", DT_BOOL, true },
 /*
 ** .pp
-** If \fIset\fP, draft files previously edited (via \fC-E -H\fP on
+** If \fIset\fP, draft files previously edited (via \fI-E -H\fP on
```
alejandro-colomar, line 4068: cmd flag bold

docs/config.c
```diff
@@ -4265,13 +4266,13 @@
 ** Specifies the program and arguments used to deliver mail sent by NeoMutt.
 ** NeoMutt expects that the specified program interprets additional
 ** arguments as recipient addresses.  NeoMutt appends all recipients after
-** adding a \fC--\fP delimiter (if not already present).  Additional
+** adding a \fI--\fP delimiter (if not already present).  Additional
```
alejandro-colomar, line 4269: literal bold

docs/config.c
```diff
@@ -4654,9 +4655,9 @@
 /*
 ** .pp
 ** This format string specifies a command which is used to decrypt
-** \fCapplication/pkcs7-mime\fP attachments.
+** \fIapplication/pkcs7-mime\fP attachments.
```
alejandro-colomar, line 4658: gbr

docs/config.c
```diff
@@ -4669,11 +4670,11 @@
 ** .dt %i .dd Intermediate certificates
 ** .dt %k .dd The key-pair specified with $$smime_default_key
 ** .dt %s .dd Expands to the name of a file containing the signature part
-**            of a \fCmultipart/signed\fP attachment when verifying it.
+**            of a \fImultipart/signed\fP attachment when verifying it.
```
alejandro-colomar, line 4673: gbr

docs/config.c
```diff
@@ -4835,10 +4836,10 @@
 /*
 ** .pp
 ** This command is used to created S/MIME signatures of type
-** \fCmultipart/signed\fP, which can be read by all mail clients.
+** \fImultipart/signed\fP, which can be read by all mail clients.
```
alejandro-colomar, line 4839: gbr

docs/config.c
```diff
@@ -4861,21 +4862,21 @@
 { "smime_verify_command", D_STRING_COMMAND, 0 },
 /*
 ** .pp
-** This command is used to verify S/MIME signatures of type \fCmultipart/signed\fP.
+** This command is used to verify S/MIME signatures of type \fImultipart/signed\fP.
```
alejandro-colomar, line 4865: gbr

docs/config.c
```diff
@@ -4861,21 +4862,21 @@
...
 ** (S/MIME only)
 */
 
 { "smime_verify_opaque_command", D_STRING_COMMAND, 0 },
 /*
 ** .pp
 ** This command is used to verify S/MIME signatures of type
-** \fCapplication/pkcs7-mime\fP.
+** \fIapplication/pkcs7-mime\fP.
 ** .pp
```
alejandro-colomar, line 4877: gbr

docs/config.c
```diff
@@ -5120,14 +5121,14 @@
...
 ** advertising the capability. When \fIunset\fP, NeoMutt will not attempt to
-** use \fCSTARTTLS\fP regardless of the server's capabilities.
+** use \fISTARTTLS\fP regardless of the server's capabilities.
 ** .pp
-** \fBNote\fP that \fCSTARTTLS\fP is subject to many kinds of
+** \fBNote\fP that \fISTARTTLS\fP is subject to many kinds of
 ** attacks, including the ability of a machine-in-the-middle to
 ** suppress the advertising of support.  Setting $$ssl_force_tls is
-** recommended if you rely on \fCSTARTTLS\fP.
+** recommended if you rely on \fISTARTTLS\fP.
```
alejandro-colomar, line 5131: @g-branden-robinson  What do you think?  Bold?  Italics?

docs/config.c
```diff
@@ -5355,18 +5356,18 @@
...
-** the above mentioned use case of "\fCxterm -e neomutt\fP".
+** the above mentioned use case of "\fIxterm -e neomutt\fP".
 */
 
 { "text_flowed", DT_BOOL, false },
 /*
 ** .pp
 ** When \fIset\fP, NeoMutt will generate "format=flowed" bodies with a content type
-** of "\fCtext/plain; format=flowed\fP".
+** of "\fItext/plain; format=flowed\fP".
```
alejandro-colomar, line 5370: gbr

docs/config.c
```diff
@@ -5381,7 +5382,7 @@
 { "thorough_search", DT_BOOL, true },
 /*
 ** .pp
-** Affects the \fC~b\fP, \fC~B\fP, and \fC~h\fP search operations described in
+** Affects the \fI~b\fP, \fI~B\fP, and \fI~h\fP search operations described in
```
alejandro-colomar, line 5385: These are literals.  Bold.

docs/config.c
```diff
@@ -5567,10 +5568,10 @@
 /*
 ** .pp
 ** \fBWarning:\fP do not set this variable unless you are using a version
-** of sendmail which supports the \fC-B8BITMIME\fP flag (such as sendmail
+** of sendmail which supports the \fI-B8BITMIME\fP flag (such as sendmail
 ** 8.8.x) or you may not be able to send mail.
 ** .pp
-** When \fIset\fP, NeoMutt will invoke $$sendmail with the \fC-B8BITMIME\fP
+** When \fIset\fP, NeoMutt will invoke $$sendmail with the \fI-B8BITMIME\fP
```
alejandro-colomar, line 5574: cmd flags bold

docs/config.c
```diff
@@ -5591,9 +5592,9 @@
 ** "From:" header.
 ** .pp
 ** Note that this information is passed to sendmail command using the
-** \fC-f\fP command line switch. Therefore setting this option is not useful
-** if the $$sendmail variable already contains \fC-f\fP or if the
-** executable pointed to by $$sendmail doesn't support the \fC-f\fP switch.
+** \fI-f\fP command line switch. Therefore setting this option is not useful
+** if the $$sendmail variable already contains \fI-f\fP or if the
+** executable pointed to by $$sendmail doesn't support the \fI-f\fP switch.
```
alejandro-colomar, line 5597: cmd flags bold

docs/config.c
alejandro-colomar: Heh!  Have a lovely new year!  If your thesis allows...  :-)

docs/config.c
g-branden-robinson: Holy blap, I've been blitzed.

Uh, okay, I'll work through these callouts one by one.

That, I would quote.  We really need to stop fearing quotation in our _man_(7) documents.

```
** Controls the decoding of complex MIME messages into \[lq]text/plain\[rq] when
```

docs/config.c
g-branden-robinson: Quotation marks again.

docs/config.c
g-branden-robinson: Quotation marks again.

docs/config.c
g-branden-robinson: Quotation marks again.

docs/config.c
g-branden-robinson: Quotation marks again.

docs/config.c
g-branden-robinson: Quotation marks again.

docs/config.c
g-branden-robinson: I certainly wouldn't use straight double quotes unless those are part of the literal.

I'd probably use both quotation marks _and_ bold.

```
If this variable is
.IR set , \" is emphasis necessary here?
NeoMutt will try to use the
.RB \[lq] LAST \[rq]
POP command
```

docs/config.c
g-branden-robinson: 3985-3986 look like they should be bold, not italic.

I would spell `^` as `\[ha]` for good output on UTF-8 terminals and PDF.

3989-3990 offers an example and a "mention" instead of a "use" of a literal.  I'd lose the italics and use only quotation, but proper quotation special characters as shown above, not neutral double quotes.

3992-3994 offers the same sort of specimens as 3989-3990.  So roman quotation again, I'd say.

docs/config.c
g-branden-robinson: Quotation marks again.

docs/config.c
g-branden-robinson: Quotation marks again.

docs/config.c
g-branden-robinson: Quotation marks again.

docs/config.c
g-branden-robinson: Quotation marks again.

docs/config.c
g-branden-robinson: Quotation marks again.

docs/config.c
g-branden-robinson: I'd probably quote these since they are literals but not squarely within the topical domain of NeoMutt.  They're relevant to the discussion of course, but since NeoMutt is theoretically independent of the network transport, `STARTTLS` is not a central matter.

docs/config.c
g-branden-robinson: Quotation marks again.

docs/config.c
alejandro-colomar: > Holy blap, I've been blitzed.

:-)

> Uh, okay, I'll work through these callouts one by one.
> 
> That, I would quote. We really need to stop fearing quotation in our _man_(7) documents.
> 
> ```
> ** Controls the decoding of complex MIME messages into \[lq]text/plain\[rq] when
> ```

Hmmm, true.  I now think I have seen some of those in the Linux man-pages project.

However, since this string is usually quoted in code, I might use vertical quotes, as if it was a string.  What do you think?

docs/config.c
alejandro-colomar: Why lq and rq?  I think bold should be enough, no?

docs/config.c
alejandro-colomar: > 3985-3986 look like they should be bold, not italic.

I was having doubts because of:

-  It's an example, and we use italics for examples.

-  Or maybe we should consider that as a literal with placeholders inside, in which case it would be a mixed bold italics (italics for the letters).

What do you think?

> 
> I would spell `^` as `\[ha]` for good output on UTF-8 terminals and PDF.

+1

> 3989-3990 offers an example and a "mention" instead of a "use" of a literal. I'd lose the italics and use only quotation, but proper quotation special characters as shown above, not neutral double quotes.

In the Linux man-pages project we use italics for examples.  Do you have different advice?

> 3992-3994 offers the same sort of specimens as 3989-3990. So roman quotation again, I'd say.

docs/config.c
alejandro-colomar: +1

docs/config.c
ossilator: on a related note, all literal dashes need to be \-escaped.

docs/config.c
ossilator: well, actually, this should use the .mr macro.

alejandro-colomar: That's still not portable enough for neomutt(1).

docs/config.c
g-branden-robinson: > Inline examples of commands, expressions, etc., in italics.

I don't think that's part of the advice in _groff_man_style_(7).

```
     Use italics for emphasis rarely, and bold almost never.  Brief
     specimens of literal text, such as article titles, mentions of
     individual characters or short strings, and (sub)section headings
     in man pages, are suitable objects for quotation; see the \(lq,
     \(rq, \(oq, and \(cq escape sequences in subsection “Portability”
     below.
```

docs/config.c
g-branden-robinson: I'd only use neutral quotes if they were a necessary part of the example.  I would not use them to perform quotation itself in the man page itself.

YES: `Consider the declaration \[lq]const char foo[] = "bar";\[rq] next.`
NO: `We wanted "foobar" to be null-terminated.`

docs/config.c
g-branden-robinson: Quotation marks again.

docs/config.c
alejandro-colomar: > > Inline examples of commands, expressions, etc., in italics.
> 
> I don't think that's part of the advice in _groff_man_style_(7).
> 
> ```
>      Use italics for emphasis rarely, and bold almost never.  Brief
>      specimens of literal text, such as article titles, mentions of
>      individual characters or short strings, and (sub)section headings
>      in man pages, are suitable objects for quotation; see the \(lq,
>      \(rq, \(oq, and \(cq escape sequences in subsection “Portability”
>      below.
> ```

Sorry, that part is from man-pages(7).

docs/config.c
g-branden-robinson: Strictly speaking, it's not necessary to select the previous font with `\fP` immediately prior to changing it again, as with `\fP\fI`.  But it can make a lengthy run of font-dancing easier to follow.  There's a tradeoff between length comprehensibility.

This is also a situation where `\c` can ride to your rescue.

```
.RB [[ s ] news:// ]\c
.RI [ username\c
.RB [ :\c
.IR password ]\c
.BR @ ]\c
.IR server [\c
.BI : port\c
]
```

docs/config.c
g-branden-robinson: @ossilator: Agreed.  That argues for bold or quotation marks, in my view.

docs/config.c
alejandro-colomar: > I'd only use neutral quotes if they were a necessary part of the example. I would not use them to perform quotation itself in the man page itself.
> 
> YES: `Consider the declaration \[lq]const char foo[] = "bar";\[rq] next.` NO: `We wanted "foobar" to be null-terminated.`

Hmmm, I disagree here. `"foobar"` is a C string, and I think that writing it as it would be written in a C program, with vertical quotes, aids readability.

docs/config.c
g-branden-robinson: I often add quotes (in roman) when a boldfaced literal comprises multiple words, so that when copying and pasting man page content, the boundaries of the literal are still visible.  But yes, that's not necessary here.  I don't think it's wrong in this scenario to omit or retain the quotation marks; that's more a question of style for the page author (or project style guide).

docs/config.c
alejandro-colomar: > Strictly speaking, it's not necessary to select the previous font with `\fP` immediately prior to changing it again, as with `\fP\fI`. But it can make a lengthy run of font-dancing easier to follow. There's a tradeoff between length and comprehensibility, in the sense of how much state you have to track in your brain.
> 
> This is also a situation where `\c` can ride to your rescue.
> 
> ```
> .RB [[ s ] news:// ]\c
> .RI [ username\c
> .RB [ :\c
> .IR password ]\c
> .BR @ ]\c
> .IR server [\c
> .BI : port\c
> ]
> ```

I'm never sure if `\fI\fB\fP` returns to the `I` or to the font prior to it, which is why I prefer keeping a stack of 1.  That way I'm sure of the behavior.

docs/config.c
alejandro-colomar: I wasn't going to mention them for now, to not confuse too much.  :)
But yeah, we need to fix that afterwards.

docs/config.c
g-branden-robinson: @alejandro-colomar : Hey, don't wave him off!  If _neomutt_ wants to help usher in the `MR` revolution, I'm all for it!

But a frank word about its portability is in order.

_mandoc_(1) [added support for `MR` in its CVS over a year ago](https://cvsweb.bsd.lv/mandoc/man_macro.c?rev=1.149&content-type=text/x-cvsweb-markup&sortby=date) but [has not done a release in 3 years](https://mandoc.bsd.lv/).  I'd expect Heirloom Doctools to update [their copy of our extensions](https://github.com/n-t-roff/heirloom-doctools/blob/master/troff/troff.d/tmac.d/an-ext) at some point, but at present [its pace of development is pretty slow](https://github.com/n-t-roff/heirloom-doctools/commits/master/).

docs/config.c
g-branden-robinson: Yes, I have different advice.  I've never advised the use of italics for examples, inline or displayed.

I would use quotation for inline examples and `EX`/`EE` for displayed examples.

alejandro-colomar: > Because inline examples are so often multiple words, quoting them protects you from losing perceptibility of their boundaries when copy-and-pasting them.

A problem is that inline examples quite often include (vertical) quotes (think of string literals, mainly; sometimes character literals too).

docs/config.c
g-branden-robinson: `\t` isn't going to format like that:

```
$ nroff -man <<EOF
##> .TH neomuttrc 5 2025-01-01 "groff test suite"
.SH Description
Note \fI"\t"\fP is converted to a literal tab inside a
EOF
neomuttrc(5)                   File Formats Manual                  neomuttrc(5)

Description
     Note "" is converted to a literal tab inside a

groff test suite                   2025‐01‐01                       neomuttrc(5)
```

You need `\[rs]t`.

_groff_man_style_(7):
```
     \(rs   Reverse solidus (backslash).  The backslash is the default
            escape character in the roff language, so it does not
            represent itself in output.  Also see \e above.
```

docs/config.c
g-branden-robinson: As a bit of off-topic English style advice...

1.  The Latin abbreviation "i.e." should be followed by a comma, since _id est_ roughly means "that is".
2.  The Latin abbreviation remains subject to the capitalization rules of English sentences, thus "I.e.,".
3.  I would avoid the use of Latin abbreviations except in a conventional scholarly context anyway.  In man pages I favor English words over Latin abbreviations.  This also avoids the common embarrassment of using "cf." incorrectly to mean "see", when it abbreviates Latin _confere_, which roughly means "compare".

docs/config.c
g-branden-robinson: In my opinion, you are then implying the C string literal `\"foobar\"`.

My view is that C syntax and English syntax differ greatly and should not be intermixed.

docs/config.c
g-branden-robinson: And if someone said `\fC`, the identity of the previous font may be [implementation-dependent](https://lists.gnu.org/archive/html/groff/2022-11/msg00162.html).  😮

docs/config.c
g-branden-robinson: I don't think that's much of a problem, and not one at all on UTF-8 or typesetting devices.

```
$ nroff -man <<EOF
.TH foobar 1 2025-01-01 "groff test suite"
We should have declared \[lq]char foo[]="bar";\[rq] as
.BR const .
EOF
foobar(1)                    General Commands Manual                   foobar(1)

We should have declared “char foo[]="bar";” as const.

groff test suite                   2025‐01‐01                          foobar(1)
```

g-branden-robinson:
@alejandro-colomar: I'll wager the NeoMutt developers didn't count on one of their MRs being the site of a summit between us on open questions in man page styling. ;-)

Sorry, guys.  If it helps, I'm a happy _neomutt_ user.  Albeit a little nonplussed when an IMAP connection gets rudely reset by GMail and it drops core.  😋

ossilator:
> Albeit a little nonplussed when an IMAP connection gets rudely reset [...]

given the context, i'll have a slight chuckle at the common abuse of "nonplussed" here. 😋

g-branden-robinson:
> > Albeit a little nonplussed when an IMAP connection gets rudely reset [...]
> 
> given the context, i'll have a slight chuckle at the common abuse of "nonplussed" here. 😋

I wouldn't say I _abused_ it...  Ordinarily when I experience a core dump, I get angry. 😈

In this case I am merely bewildered.  No idea who to blame, and if it's Google, there's no point getting mad.  The mindless juggernaut cares not what chaos it creates or RFCs it violates...

docs/config.c
ossilator: i'm not sure what you're arguing here at this point. :}

i'd write:

> Consider the declaration \\[lq]const char foo[] = \\[dq]bar\\[dq];\\[rq] next.
> We wanted \\[lq]bar\\[rq] to be null-terminated.

it is my understanding that literal quotes should never be used, as their rendering varies in undesirable ways.

quotes aside, inlining a whole line of c code is a bad idea; i'd write it like that instead:

> .RS 8
> .nf
> const char foo[] = \\[dq]bar\\[dq];
> .fi
> .RE
> .IP

ossilator:
i mean that the word means literally the opposite of what it is now used for. better avoid it altogether.

ossilator:
as for the crash, that would be obviously a bug even if it's triggered by google being funny. so please follow the usual reporting procedure next time you repro it.

docs/config.c
g-branden-robinson: > i'm not sure what you're arguing here at this point. :}
> 
> i'd write:
> 
> > Consider the declaration \[lq]const char foo[] = \[dq]bar\[dq];\[rq] next.
> > We wanted \[lq]bar\[rq] to be null-terminated.
> 
> it is my understanding that literal quotes should never be used,
> as their rendering varies in undesirable ways.

"Never" is, technically, too strong, but pretty close.  There are two different problems.

- `"` is interpreted as delimiting a macro argument, but only when used in a macro call.  It's "safe" on text lines, but if someone inattentively grabs a phrase from a text line and stuffs it into a macro call, they likely will not get the formatted output they expect.  Unfortunately, `\[dq]` (or even `\(dq`) isn't perfectly portable; see below.

- Single quotes have gotten progressively more problematic.
- - The grave accent (I can't get this Markdown to format it inline correctly) and apostrophe/neutral single quote `'` traditionally were perfectly reliable ways of getting directional single quotes, _if_ they existed at all on the output device.  When the Bell Labs CSRC developed _nroff_ and _troff_, these characters looked on their Teletype Corporation Model 37s much as depicted in [ANSI X3.4-1977](https://nvlpubs.nist.gov/nistpubs/Legacy/FIPS/fipspub1-2-1977.pdf).  
- - In the 1980s, ISO 8859 disambiguated the apostrophe from the acute accent, rendering the former neutral (vertical) in orientation, which also removed it from consideration as a candidate directional single quotation mark.
- - Starting about 15 years ago, it became fashionable to have one's _groff_ _man_ package remap the grave accent and apostrophe to "themselves".  That created a problem.

As a forthcoming commit of mine to _groff_man_style_(7) puts it:

```
       Historically, man pages used ` and ' exclusively for directional
       single quotation marks.  However, in recent years, some
       distributors of groff have chosen to override the meanings of
       these characters in man pages, remapping them to their Unicode
       Basic Latin code points.  Unfortunately, ` and ' are the only
       reliable means of obtaining directional single quotation marks in
       AT&T troff; in that implementation, often no special character
       escape sequences exist to obtain them.  Further, in AT&T troff,
       the glyph repertoire, like the set of supported font names, is
       device‐specific.  To achieve quotation portably in man pages
       rendered both by AT&T and more modern troffs, consider adding a
       preamble to your page after the .TH call as follows.

              .ie \n(.g \{\
              .  ds oq \(oq
              .  ds cq \(cq
              .\}
              .el \{\
              .  ds oq `
              .  ds cq '
              .\}

       You must then use the \* escape sequence to interpolate the
       quotation mark strings.

              The command
              .RB \*(oq "while ! git pull; do sleep 10; done" \*(cq
              retries an update from our repository until it succeeds.

       If this procedure seems complex, petition your distributor to
       revert their remapping of the ` and ' characters.
```

Since old _troffs_ don't reliably support all of the special character escape sequences for obtaining quotation marks that _groff_ supports, if you need portability to those formatters, you can't say "never".

> quotes aside, inlining a whole line of c code is a bad idea; i'd write it like that instead:
> 
> > .RS 8
> > .nf
> > const char foo[] = \[dq]bar\[dq];
> > .fi
> > .RE
> > .IP

I won't argue with your suggestion to display such an example instead of inlining it.  But I would:

1. ...not supply arguments to the `RS` request unless necessary.  The relative inset amount is configurable at formatting time via the `IN` register, and the user's selection should be respected.  Where necessary, multiple `RS` insets can be nested.
2. ...use `EX` and `EE` inboard of the `nf` and `fi` requests if you need portability to formatters that don't support `EX` and `EE`, and instead of the requests if you don't.  This gets you a monospaced font family (Courier) on typesetters, which in turn makes ASCII art possible on those devices.  Traditional `troff`s silently ignore calls of macro names they don't recognize, _groff_ supports them, and so does _mandoc_ since its 1.12.2 release over 10 years ago.

g-branden-robinson:
> i mean that the word means literally the opposite of what it is now used for. better avoid it altogether.

It's its own antonym, like "inflammable", or "French spacing"!  Is this a good time to say "I could care less"? 😂

docs/config.c
alejandro-colomar: > In my opinion, you are then implying the C string literal `\"foobar\"` (as I would spell it in source code; the object in memory of course has no backslashes).
> 
> My view is that C syntax and English syntax differ greatly and should not be intermixed.
> 
> Moreover, in man pages we often have occasion to discuss programming or configuration language syntaxes that are not C. Aggressively using C-correct utterances in those contexts can cause confusion, or outright error on the part of the user.

Hmmmm, maybe it depends on the context.  I don't fully agree, but don't strongly disagree either.

docs/config.c
alejandro-colomar: Hmmmm, okay.  I'll keep that in mind, and maybe do some global reform some day.  Not now, though.

alejandro-colomar:
> @alejandro-colomar: I'll wager the NeoMutt developers didn't count on one of their MRs being the site of a summit between us on open questions in man page styling. 😂

:D

> Sorry, guys. If it helps, I'm a happy _neomutt_ user.

:-)

> Albeit a little nonplussed when an IMAP connection gets rudely reset by GMail and it drops core. 😋

I personally don't use neomutt(1)'s IMAP.  I have mbsync(1) sync the mail, and then neomutt(1) read from a local maildir.  This has the benefit of keeping a local backup of my mail server.  My mind is calmed knowing neomutt(1)'s bugs (which sometimes happen), can't affect my mail server at all.  Since mbsync(1) is a much simpler program, I don't expect it to be too dangerous.

BTW, I also use gmail (but am stopping that, I mainly use migadu now), so if you're interested in this, my config files might help.

mbsync(1):
<https://www.alejandro-colomar.es/src/alx/alx/config.git/tree/home/.mbsyncrc>

mutt(1):
<https://www.alejandro-colomar.es/src/alx/alx/config.git/tree/home/.config/mutt/>

docs/config.c
charles2910: >  OK, italics it is.

@flatcap I've been looking at this right now and realized a small problem. We generate all html documentation from these using `docs/makedoc.c`. If we change `\fC` to `\fI`, `\fB` or else, we are loosing the code blocks on html. I think you generate the website documentation the same way right? Like the [reference](https://neomutt.org/guide/reference.html), all code blocks would become italics or bold.

So I'm a bit in doubt about loosing the code blocks on html in order to make the manpages compliant to `groff_man_style(7)`. Using `\fCR` would keep things as is.

Right now I see the two options above, maybe you guys can give a better and different one.

charles2910: Uhm... I have an idea starting to form on my head - to be continued tomorrow

(apparently all I needed was to stop typing and go to shower)

charles2910:
Hi @gahr. Yes! I'm tackling this MR this next weekend. Sorry for the delay, I attended Fosdem and took a week of vacation afterwards.

gahr:
Hi @charles2910 , not at all, thanks a lot!

flatcap:
Educational as this journey has been, it hasn't arrived anywhere.

I've just commited: https://github.com/neomutt/neomutt/commit/96aa94c514b0d86676c2e45e7eae4b5f72f4160c which fixes a bunch of linting problems,
and it also fixes the original point of this PR: Use `\f[CR]` rather than `\fC`.

Until someone volunteers to **take responsibility** for the Man Pages, they won't get fixed.

Sorry.

alejandro-colomar:
> Educational as this journey has been, it hasn't arrived anywhere.
> 
> I've just commited: [96aa94c](https://github.com/neomutt/neomutt/commit/96aa94c514b0d86676c2e45e7eae4b5f72f4160c) which fixes a bunch of linting problems, and it also fixes the original point of this PR: Use `\f[CR]` rather than `\fC`.
> 
> Until someone volunteers to **take responsibility** for the Man Pages, they won't get fixed.
> 
> Sorry.

I'll have a look at them eventually.  I just can't promise when.  I promise to help with review of patches promptly, but not with creation of patches.  :)

PR #4494, opened 2025-01-06
Help Page refactor

flatcap:
**post-release**

This PR removes some Pager complexity by simplifying the Help Page.

The Pager code is a mess.
It's called from many different places in NeoMutt and each of the callers has different requirements.
The Pager has custom code for each of them.

---

### Help Page

When you hit <kbd>?</kbd> in NeoMutt, it opens the Help Page.

The Help Page displays three blocks of information:

- Bindings: All the current keybindings for functions and macros
- Unbound: A list of all the available functions that have no keybinding
- Flag Chars: (Index only) A cheat sheet for `$crypt_chars`, `$flag_chars`, `$to_chars`

### Current Help Page

The text is displayed in three columns.
The width of these columns is calculated from the width of the Terminal.

- Bindings: Keystroke | Function | Description
- Macros: Keystroke | Macro text | Description

This layout has its problems.

It doesn't look great for macros.
They're either squashed into column 2, or badly wrapped in column 3.

The list of unbound functions doesn't use the second columns, so it wastes a lot of space.

If you resize the Terminal, the Help Page is regenerated from scratch.
The column sizes are recalculated and the text spread out to fill the space available.

This means a lot of added complexity in the Help Page and custom code in the Pager.

### New Help Page

The new Help Page creates the help text once and lets the Pager do any wrapping.

To avoid wasted space, it calculates the mimimum required width for each column.
To make the macros legible it introduces a two-line format.


flatcap:
## Screenshots

### Bound keys

Old

image

New - Note the two-line display of macros (with a description)

image

---

### Unbound functions

Old

image

New

image

---

### Flag Chars

Old

image

New

image

gahr:
The screenshots look nice :) I'll take a look at the code tomorrow. Anything in particular you need a pair of eyes on?

flatcap:
Thanks.
There's nothing special to look at.
The code is a little repetitive, atm, but it's likely to get refactored again.

I've been working on a Pager 2.0 (well a SimplePager, at least).
This will offload a lot of the Pager burden onto the caller.
The caller will be able to markup the text and pass this info to the SimplePager
(rather than the Pager having to understand every possible use-case).

help.c
```diff
@@ -77,24 +89,32 @@ static const struct MenuFuncOp *help_lookup_function(int op, enum MenuType menu)
...
- * @param[out] macro    Macro string
- * @retval num Number of screen columns used
+ * help_sort_alpha - Compare two Help Lines by their first entry - Implements ::sort_t - @ingroup sort_api
+ */
+static int help_sort_alpha(const void *a, const void *b, void *sdata)
+{
+  struct HelpLine x = *(const struct HelpLine *) a;
+  struct HelpLine y = *(const struct HelpLine *) b;
+
+  return mutt_str_cmp(x.first, y.first);
```
gahr, line 99: How about avoiding a copy:

```suggestion
  const struct HelpLine *x = (const struct HelpLine *) a;
  const struct HelpLine *y = (const struct HelpLine *) b;

  return mutt_str_cmp(x->first, y->first);
```

flatcap:
Here are the macros in a separate section:

### Wide screen

image

### Narrow screen, `smart_wrap = no` 

image

### Narrow screen, `smart_wrap = yes` 

image

PR #4536, opened 2025-01-10
expand tabs to spaces in compose preview (fixes #4534)

### LINKED ISSUE #4534: compose_show_preview renders <TAB> character as questionmark
## NeoMutt Version

```
neomutt -v              
NeoMutt 20250109-dirty
Copyright (C) 2015-2025 Richard Russon and friends

System: Linux 6.13.0-rc6-1-mainline (x86_64)
ncurses: ncurses 6.5.20240427 (compiled with 6.5.20240427)
libidn2: 2.3.7 (compiled with 2.3.7)
GPGME: 1.24.1
GnuTLS: 3.8.8
libnotmuch: 5.6.0
compression: lz4, zlib, zstd

Configure options: --prefix=/usr --sysconfdir=/etc --libexecdir=/usr/lib --gpgme --sqlite --autocrypt --lua --notmuch --gss --gnutls --sasl --with-ui=ncurses --with-idn2=/usr --disable-idn --idn2 --lmdb --kyotocabinet --gdbm --ubsan --lz4 --zlib --zstd

Compilation CFLAGS: -march=x86-64 -mtune=generic -O2 -pipe -fno-plt -fexceptions         -Wp,-D_FORTIFY_SOURCE=3 -Wformat -Werror=format-security         -fstack-clash-protection -fcf-protection         -fno-omit-frame-pointer -mno-omit-leaf-frame-pointer -g -ffile-prefix-map=/build/neomutt/src=/usr/src/debug/neomutt -flto=auto -std=c11 -fno-delete-null-pointer-checks -D_ALL_SOURCE=1 -D_GNU_SOURCE=1 -D__EXTENSIONS__ -D_XOPEN_SOURCE_EXTENDED -I/usr/include -I/usr/include -I/usr/include -DNCURSES_WIDECHAR -I/usr/include -I/usr/include/p11-kit-1 -I/usr/include -I/usr/include -fsanitize=undefined -fno-omit-frame-pointer -fno-common -g -O0

Compile options:
  -gsasl +gss +hcache -homespool +idn +inotify -locales_hack +lua +nls +notmuch 
  -openssl +pgp +regex +sasl +smime +sqlite +truecolor 

Devel options:
  ubsan 

MAILPATH="/var/mail"
SENDMAIL="/usr/sbin/sendmail"
SYSCONFDIR="/etc"

```

## Expected Behaviour

I was sending a mail which contained a patch with code that was indented by tab characters. I was expecting the send-preview to render these characters properly like the full windowed preview does.

Screenshots2025-01-10_10:04:05

## Actual Behaviour

All <TAB> characters were rendered as "?"

Screenshots2025-01-10_10:04:00

## Steps to Reproduce

have `set compose_show_preview = yes` in your config
preview a mail that contains tabs

Always 

I don't know when it started to happen, if it is of interest I can compare the behavior with the previous version of neomutt.

## Extra Info

* Operating System and its version: Arch Linux (rolling)
* Were you using multiple copies of NeoMutt at once? no
* Were you using 'screen' or 'tmux'? no
* Is your email local (maildir) or remote (IMAP)? local

roccoblues:
As reported in (#4534), the message preview in the compose window incorrectly rendered tabs (`\t`) as `?`.

The reason was that `mutt_paddstr()` replaces everything that is not `IsWPrint()` with `?`. To fix this we expand the tabs in each line before we measure the width and print it.

### DIFF
--- a/compose/preview.c
+++ b/compose/preview.c
@@ -181,6 +181,8 @@ static void draw_preview(struct MuttWindow *win, struct PreviewWindowData *wdata
 
       content_lines++;
 
+      line = mutt_str_expand_tabs(line, &line_len, 8);
+
       // Check how much of the string fits into the window width.
       size_t width = 0;
       size_t bytes = mutt_wstr_trunc(&line[pos], line_len - pos, win->state.cols, &width);
--- a/test/string/mutt_str_expand_tabs.c
+++ b/test/string/mutt_str_expand_tabs.c
@@ -1,0 +1,80 @@
+/**
+ * @file
+ * Test code for mutt_str_expand_tabs()
+ *
+ * @authors
+ * Copyright (C) 2025 Dennis Schön <mail@dennis-schoen.de>
+ *
+ * @copyright
+ * This program is free software: you can redistribute it and/or modify it under
+ * the terms of the GNU General Public License as published by the Free Software
+ * Foundation, either version 2 of the License, or (at your option) any later
+ * version.
+ *
+ * This program is distributed in the hope that it will be useful, but WITHOUT
+ * ANY WARRANTY; without even the implied warranty of MERCHANTABILITY or FITNESS
+ * FOR A PARTICULAR PURPOSE.  See the GNU General Public License for more
+ * details.
+ *
+ * You should have received a copy of the GNU General Public License along with
+ * this program.  If not, see <http://www.gnu.org/licenses/>.
+ */
+
+#include "mutt/memory.h"
+#define TEST_NO_MAIN
+#include "config.h"
+#include "acutest.h"
+#include <string.h>
+#include "mutt/lib.h"
+#include "test_common.h"
+
+struct TestCase
+{
+  char *in;
+  char *result;
+  int tabwidth;
+};
+
+void test_mutt_str_expand_tabs(void)
+{
+  // char *mutt_str_expand_tabs(char *str, size_t *len, int tabwidth);
+
+  {
+    TEST_CHECK(mutt_str_expand_tabs(NULL, 0, 4) == NULL);
+    TEST_CHECK(mutt_str_expand_tabs("", 0, 4) == NULL);
+  }
+
+  static struct TestCase tests[] = {
+    // clang-format off
+    { "\tapple",      "    apple",     4 },
+    { "X\tapple",     "X   apple",     4 },
+    { "XX\tapple",    "XX  apple",     4 },
+    { "XXX\tapple",   "XXX apple",     4 },
+    { "XXXX\tapple",  "XXXX    apple", 4 },
+    { "XXXXX\tapple", "XXXXX   apple", 4 },
+    { "\tapple\t",    "    apple   ",  4 },
+    { "🐛\tapple",    "🐛  apple",     4 },
+    { "\t🐛\tapple",  "    🐛  apple", 4 },
+    { "\t\tapple",    "        apple", 4 },
+    { "X\t\tapple",   "X       apple", 4 },
+    { "XX\t\tapple",  "XX      apple", 4 },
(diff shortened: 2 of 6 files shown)


test/string/mutt_str_expand_tabs.c
```diff
@@ -0,0 +1,64 @@
...
+    char *str = mutt_str_dup("apple");
+    char *result = mutt_str_expand_tabs(str, &len, 4);
+    TEST_CHECK_STR_EQ(result, str);
+    FREE(&result);
+  }
+  {
+    size_t len = 5;
+    char *str = mutt_str_dup("ap\tle");
+    char *result = mutt_str_expand_tabs(str, &len, 4);
+    TEST_CHECK_STR_EQ(result, "ap    le");
```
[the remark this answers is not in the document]
roccoblues: Hey this is still a draft - no peeking. 😄 

> Before 	After
> `\tapple` 	`....apple`
> `X\tapple` 	`X...apple`
> `XX\tapple` 	`XX..apple`
> `XXX\tapple` 	`XXX.apple`
> `XXXX\tapple` 	`XXXX....apple`
> `XXXXX\tapple` 	`XXXXX...apple`

Thanks! I wanted to add way more tests next and those will come handy.

compose/preview.c
```diff
@@ -181,6 +181,8 @@ static void draw_preview(struct MuttWindow *win, struct PreviewWindowData *wdata
 
       content_lines++;
 
+      line = mutt_str_expand_tabs(line, &line_len, 4);
```
roccoblues, line 184: Do we want to constant the 4 here? Or even use another value (please not 😄).

christian-heusel:
This MR addresses the issue from #4534, so it works for me:
image

Thanks for coming up with a fix so quickly! 🎉

roccoblues:
Thanks for giving it a try to quickly!

gahr:
How about doing that in mutt_file_read_line using a new ReadLineFlags?

roccoblues:
> How about doing that in mutt_file_read_line using a new ReadLineFlags?

Interesting idea. I guess it depends on wether we ever want to change the tab width value?

gahr:
> > How about doing that in mutt_file_read_line using a new ReadLineFlags?
> 
> Interesting idea. I guess it depends on wether we ever want to change the tab width value?

I guess if we do, we'd make it into a config, and `mutt_file_read_line` could read from it

compose/preview.c
flatcap: Sorry, it really needs to be 8.
That's the default in terminals and that's how it's hard-coded in the Pager.

flatcap:
> > > How about doing that in mutt_file_read_line using a new ReadLineFlags?
> >
> > Interesting idea. I guess it depends on wether we ever want to change the tab width value?
> 
> I guess if we do, we'd make it into a config, and `mutt_file_read_line` could read from it

Hmm...
If the tab-size is hard-coded, perhaps,
but I don't want to see config creeping into library functions.

gahr:
> > > > How about doing that in mutt_file_read_line using a new ReadLineFlags?
> > > 
> > > 
> > > Interesting idea. I guess it depends on wether we ever want to change the tab width value?
> > 
> > 
> > I guess if we do, we'd make it into a config, and `mutt_file_read_line` could read from it
> 
> Hmm... If the tab-size is hard-coded, perhaps, but I don't want to see config creeping into library functions.

Sure, let's go with a fixed 8-chars expansion

roccoblues:
Sorry I lost track here. Is this fine now or should I move the tab expansion into `mutt_file_read_line()`?

flatcap (approved):
All good, thanks! (Sorry for the slow response (lost in my own world :-)

PR #4545, opened 2025-01-15
Allow nested ARRAY_FOREACH()

flatcap:
## Overview

`ARRAY_FOREACH_FROM_TO()` uses a loop variable, `ARRAY_FOREACH_IDX`.

If you compile with `-Wshadow` you get warnings if `ARRAY_FOREACH()` are nested.
(There aren't any examples in the code yet, but I'd like to be able to).

The problem can be worked around by manually creating a `for` loop.

## How it works

The change works by replacing `ARRAY_FOREACH_IDX` with a new "templated" loop variable.
The variable is a macro concatenation of literal `idx_` and the `elem` param.

e.g.
```c
struct Object *obj = NULL;
ARRAY_FOREACH(obj, object_array)
{
    printf("%ld\n", idx_obj);
}
```

**Note**: In a couple of the `FOREACH` macros, I've removed the `()` from the `elem` param, so that it can be concatenated.

## Alternative Naming

One possible problem with the naming is that it's not immediately obvious where the variable comes from.

I've chosen (arbitrarily), a prefix of `idx_`.
Given an `elem` name of **`obj`**, some possible alternatives are:

- `obj_idx`
- `ARRAY_IDX_obj`
- `obj_ARRAY_IDX`
- `ARRAY_FOREACH_IDX_obj`
- `obj_ARRAY_FOREACH_IDX`

## Commits

- 842f3e831 array: allow nested `ARRAY_FOREACH()`
  Small changes to `mutt/array.h`

- 39faffbb2 array: use new index variable
  Code changes to match

- 6b739c2c3 nested array sample code
  Sample code [nested-array.c](https://github.com/neomutt/neomutt/pull/4545/files#diff-8fd0b694390466db9dbb488eb34bb4260e83198a791ce4ec362dcb66b5ea1801) for test compiling

### DIFF
--- a/mutt/array.h
+++ b/mutt/array.h
@@ -210,7 +210,7 @@
  * @param head Pointer to a struct defined using ARRAY_HEAD()
  */
 #define ARRAY_FOREACH(elem, head)                                              \
-  ARRAY_FOREACH_FROM_TO((elem), (head), 0, (head)->size)
+  ARRAY_FOREACH_FROM_TO(elem, (head), 0, (head)->size)
 
 /**
  * ARRAY_FOREACH_FROM - Iterate from an index to the end
@@ -221,7 +221,7 @@
  * @note The from index must be between 0 and ARRAY_SIZE(head)
  */
 #define ARRAY_FOREACH_FROM(elem, head, from)                                   \
-  ARRAY_FOREACH_FROM_TO((elem), (head), (from), (head)->size)
+  ARRAY_FOREACH_FROM_TO(elem, (head), (from), (head)->size)
 
 /**
  * ARRAY_FOREACH_TO - Iterate from the beginning to an index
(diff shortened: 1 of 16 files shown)


gahr:
Nice! My colour is `ARRAY_FOREACH_IDX_obj`

mutt/array.h
```diff
@@ -245,10 +245,9 @@
  *       the from index must not be bigger than to index.
  */
 #define ARRAY_FOREACH_FROM_TO(elem, head, from, to)                            \
-  for (size_t ARRAY_FOREACH_IDX = (from);                                      \
-       (ARRAY_FOREACH_IDX < (to)) &&                                           \
-       ((elem) = ARRAY_GET((head), ARRAY_FOREACH_IDX));                        \
-       ARRAY_FOREACH_IDX++)
+  for (size_t idx_##elem = (from);                                             \
+       (idx_##elem < (to)) && ((elem) = ARRAY_GET((head), idx_##elem));        \
```
gahr, line 249: I agree: `elem` must be a variable name so it's going to be a single token anyway

```suggestion
       (idx_##elem < (to)) && (elem = ARRAY_GET((head), idx_##elem));        \
```

flatcap:
- Variable prefix renamed to `ARRAY_FOREACH_IDX_`
- Sample code dropped
- Squashed

gahr (approved):

PR #4546, opened 2025-01-15
Respect Ignore when modifying an email's headers

### LINKED ISSUE #4535: Linking threads and header weeding
This is very minor, but it should be noted for the brave explorer.

Let's say you have this in `neomuttrc` (as you should IMO)

```
ignore references
```

Now suppose you're in a mailbox with this structure:

```
1 Foo
2 |-> Re: Foo
3     |-> Re: Foo
4 Foo
```

Clearly, article 4 was sent as an OP by mistake, and should instead be part of the thread. You decide to fix it for yourself, so you tag 4, move to 3 and link them by hitting `&`. Now you open 4 - and bingo! a brand new References header is staring at you in the garish font you reserve for ignored headers.

As I said, not a huge deal -- but probably not that hard to fix either.

-- 
Ian

gahr:
Fixes #4535

### DIFF
--- a/copy.c
+++ b/copy.c
@@ -425,6 +425,7 @@ int mutt_copy_header(FILE *fp_in, struct Email *e, FILE *fp_out,
                      CopyHeaderFlags chflags, const char *prefix, int wraplen)
 {
   char *temp_hdr = NULL;
+  const bool c_weed = cs_subset_bool(NeoMutt->sub, "weed");
 
   if (e->env)
   {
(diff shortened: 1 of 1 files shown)


flatcap:
clang-format, rebase

flatcap (approved):
Nice!
8 years ago, kwantam created a fix for notmuch headers (de6e7e1fc7),
but it didn't occur to us to check the other headers :-)

PR #4553, opened 2025-01-22
config: API `has_been_set()`

flatcap:
This adds a new function to the Config API: `has_been_set()`.
This allows us to efficiently determine which config has been changed.

This builds on several recent commits:

- e87c36489 fix dumping of initial values, #4537
- fa841ed1e array: upgrade `get_elem_list()`
- 8d2ea311d query: unify `neomutt -D` and `-Q`

---

### Testing

- Dump all config: `neomutt -D`
- Show changed config in NeoMutt: `:set`
- Show all config in NeoMutt: `:set all`

These can be run with either zero config, or this test config file which sets one option for each type.

**test.rc**
```sh
set assumed_charset = "ascii:utf-8"                        # slist
set attach_format = "attach"                               # expando
set attach_save_dir = "save-dir"                           # path
set editor = "nano"                                        # string
set envelope_from_address = "John Smith <js@example.com>"  # address
set flag_chars = "abcdefghijk"                             # mbtable
set mbox_type = maildir                                    # enum
set pgp_timeout = 1234                                     # long
set print = yes                                            # quad
set quote_regex = "|"                                      # regex
set sidebar_visible = yes                                  # bool
set sleep_time = 99                                        # number
set sort_aux = "label"                                     # sort
```

### DIFF
--- a/config/regex.c
+++ b/config/regex.c
@@ -3,7 +3,7 @@
  * Type representing a regular expression
  *
  * @authors
- * Copyright (C) 2017-2023 Richard Russon <rich@flatcap.org>
+ * Copyright (C) 2017-2025 Richard Russon <rich@flatcap.org>
  * Copyright (C) 2020 Pietro Cerutti <gahr@gahr.ch>
  * Copyright (C) 2023 наб <nabijaczleweli@nabijaczleweli.xyz>
  *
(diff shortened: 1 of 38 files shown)


config/regex.c
```diff
@@ -280,6 +280,20 @@ static intptr_t regex_native_get(const struct ConfigSet *cs, void *var,
...
+ */
+static bool regex_has_been_set(const struct ConfigSet *cs, void *var,
+                               const struct ConfigDef *cdef, struct Buffer *err)
+{
+  const char *initial = (const char *) cdef->initial;
+
+  struct Regex *currx = *(struct Regex **) var;
+  const char *curval = currx ? currx->pattern : NULL;
+
+  return (!mutt_str_equal(initial, curval));
```
gahr, line 294:
```suggestion
  return !mutt_str_equal(initial, curval);
```

flatcap:
squashed

PR #4554, opened 2025-01-22
CLI: `neomutt -DD` -- Dump Different

flatcap:
Add a new command line sub-option `-D -D` or `-DD`

This simple addition to the command line only shows the **changed** config.
It can be used with the other dump options, e.g. `-O` (one-liner).

---

### Testing

- Dump all config: `neomutt -D`
- Dump changed config: `neomutt -DD`

This is particularly effective when using zero config, or a simple test config, below:

- `neomutt -n -F /dev/null -DD`
  Shows nothing

- `neomutt -n -F test.rc -DD`
  Shows just the config frrom `test.rc`
  (plus `$alias_file`, which is automatic)

**test.rc**
```sh
set assumed_charset = "ascii:utf-8"                        # slist
set attach_format = "attach"                               # expando
set attach_save_dir = "save-dir"                           # path
set editor = "nano"                                        # string
set envelope_from_address = "John Smith <js@example.com>"  # address
set flag_chars = "abcdefghijk"                             # mbtable
set mbox_type = maildir                                    # enum
set pgp_timeout = 1234                                     # long
set print = yes                                            # quad
set quote_regex = "|"                                      # regex
set sidebar_visible = yes                                  # bool
set sleep_time = 99                                        # number
set sort_aux = "label"                                     # sort
```

### DIFF
--- a/main.c
+++ b/main.c
@@ -776,7 +776,7 @@ static bool usage(void)
   puts(_("  neomutt [-nRy] [-e <command>] [-F <config>] [-f <mailbox>] [-m <type>]"));
   puts(_("  neomutt [-n] [-e <command>] [-F <config>] -A <alias>"));
   puts(_("  neomutt [-n] [-e <command>] [-F <config>] -B"));
-  puts(_("  neomutt [-n] [-e <command>] [-F <config>] -D [-S] [-O]"));
+  puts(_("  neomutt [-n] [-e <command>] [-F <config>] -D [-D] [-O] [-S]"));
   puts(_("  neomutt [-n] [-e <command>] [-F <config>] -d <level> -l <file>"));
   puts(_("  neomutt [-n] [-e <command>] [-F <config>] -G"));
   puts(_("  neomutt [-n] [-e <command>] [-F <config>] -g <server>"));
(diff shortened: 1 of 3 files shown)


main.c
```diff
@@ -798,6 +798,7 @@ static bool usage(void)
   puts(_("  -c <address>  Specify a carbon copy (Cc) recipient"));
   puts(_("  -C            Enable Command-line Crypto (signing/encryption)"));
   puts(_("  -D            Dump all config variables as 'name=value' pairs to stdout"));
+  puts(_("  -D -D         (or -DD) Like -D, but only show changed config"));
```
roccoblues, line 801: What do you think about using `-D -C` thinking `DumpChanged`?

main.c
```diff
@@ -1124,7 +1125,7 @@ main
           mutt_list_insert_tail(&cc_list, mutt_str_dup(optarg));
           break;
         case 'D':
-          dump_variables = true;
+          dump_variables++;
```
roccoblues, line 1128: Not a big fan of this tbh. A separate variable would be clearer.

roccoblues (approved):
```
~/projects/neomutt flatcap/neomutt-dd
λ ./neomutt -DD | wc -l
47
```

I thought I'd be worse. 😆

I really like the feature! I'd prefer another bool but not a blocker.

main.c
flatcap: **bzzzt** Sorry, `C` already used -- you miss out on tonight's star prize :-)

A lot of the useful alphabet is already taken:

| Flags | Upper | Lower |
|--------|--------|--------|
| **Used** | `ABCDEFGH......O.QRST.....Z` | `abcdefghi..lmn.p..s..v..yz` |
| **Free** | `........IJKLMN.P....UVWXY.` | `.........jk...o.qr.tu.wx..` |

main.c
flatcap: Yeah, done

main.c
roccoblues: 2 lines above, sorry 🤦🏼

PR #4557, opened 2025-01-23
config: drop ConfigSet from API functions

flatcap:
The types, such as number or string, only deal with a single config item.
They don't need access to the ConfigSet.


roccoblues (approved):
lgtm 👍🏼

PR #4558, opened 2025-01-23
drop obsolete pgp/smime menus

flatcap:
The key bindings for the PGP Key Selection dialog are configured by: `bind pgp c verify-key`
The key bindings for the SMIME Key Selection dialog are configured by: `bind smime c verify-key`

When GPGME is used, it uses its own menu names `key_select_pgp` and `key_select_smime`.
They were introduced into Mutt in 2004, but never documented.

Remove the undocumented menus and use the `pgp` and `smime` names instead.


rayfordshire (approved):
LGTM

We could change the description of the `verify-key` from

> verify a PGP public key

to

> verify a public key

i.e. omit the "PGP" which might be confusing in the S/MIME menu.
The line would be in

https://github.com/neomutt/neomutt/blob/210dd978b3f9a346e63fc836b704cdcc2bc6769d/gui/opcodes.h#L934-L937

rayfordshire:
I created a pull request (#4560) which changes "verify a PGP public key" to ""verify a public key" (without PGP). Feel free to close that PR if you do think it's not worth the hassle for the translators.

PR #4559, opened 2025-01-24
array: foreach_reverse()

flatcap:
Add new ARRAY functions to iterate backwards.

- `ARRAY_FOREACH_REVERSE(elem, head)`
- `ARRAY_FOREACH_REVERSE_FROM(elem, head, from)`
- `ARRAY_FOREACH_REVERSE_TO(elem, head, to)`
- `ARRAY_FOREACH_REVERSE_FROM_TO(elem, head, from, to)`

I'd like to add these for my (so-far-mythical) Simple Pager.

---

The first commit adds functions and tests.
Unfortunately, they trigger a build warning:
```
warning: comparison of unsigned expression in ‘>= 0’ is always true [-Wtype-limits]
```

This comes from a test: `(ARRAY_FOREACH_IDX_##elem >= (to))` where `to` is 0.
`-Wtype-limits` is enabled by `-Wextra`.

---

The second commit changes a couple of types from `size_t` to `int`.
This works around the build warning.

### DIFF
--- a/mutt/array.h
+++ b/mutt/array.h
@@ -256,6 +256,57 @@
        ARRAY_FOREACH_IDX_##elem++)
 
 /**
+ * ARRAY_FOREACH_REVERSE - Iterate backwards over all elements of the array
+ * @param elem Variable to be used as pointer to the element at each iteration
+ * @param head Pointer to a struct defined using ARRAY_HEAD()
+ * 
+ * @note Range: (ARRAY_SIZE(head)-1) .. 0
+ */
+#define ARRAY_FOREACH_REVERSE(elem, head)                                      \
+  ARRAY_FOREACH_REVERSE_FROM_TO(elem, (head), (head)->size, 0)
+
+/**
+ * ARRAY_FOREACH_REVERSE_FROM - Iterate from an index to the beginning
+ * @param elem Variable to be used as pointer to the element at each iteration
+ * @param head Pointer to a struct defined using ARRAY_HEAD()
+ * @param from Starting index (exclusive)
+ *
+ * @note Range: (from-1) .. 0
+ * @note 'from' must be between 0 and ARRAY_SIZE(head)
+ */
+#define ARRAY_FOREACH_REVERSE_FROM(elem, head, from)                           \
+  ARRAY_FOREACH_REVERSE_FROM_TO(elem, (head), (from), 0)
+
+/**
+ * ARRAY_FOREACH_REVERSE_TO - Iterate from the end to an index
+ * @param elem Variable to be used as pointer to the element at each iteration
+ * @param head Pointer to a struct defined using ARRAY_HEAD()
+ * @param to   Terminating index (inclusive)
+ *
+ * @note Range: (ARRAY_SIZE(head)-1) .. to
+ * @note 'to' must be between 0 and ARRAY_SIZE(head)
+ */
+#define ARRAY_FOREACH_REVERSE_TO(elem, head, to)                               \
+  ARRAY_FOREACH_REVERSE_FROM_TO(elem, (head), (head)->size, (to))
+
+/**
+ * ARRAY_FOREACH_REVERSE_FROM_TO - Iterate between two indexes
+ * @param elem Variable to be used as pointer to the element at each iteration
+ * @param head Pointer to a struct defined using ARRAY_HEAD()
+ * @param from Starting index (exclusive)
+ * @param to   Terminating index (inclusive)
+ *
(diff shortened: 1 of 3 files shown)


mutt/array.h
```diff
@@ -250,6 +250,52 @@
...
+ */
+#define ARRAY_FOREACH_REVERSE_TO(elem, head, to)                               \
+  ARRAY_FOREACH_REVERSE_FROM_TO(elem, (head), (head)->size, (to))
+
+/**
+ * ARRAY_FOREACH_REVERSE_FROM_TO - Iterate between two indexes
+ * @param elem Variable to be used as pointer to the element at each iteration
+ * @param head Pointer to a struct defined using ARRAY_HEAD()
+ * @param from Starting index (inclusive)
+ * @param to   Terminating index (exclusive)
```
rayfordshire, line 288: "inclusive"/"exclusive" -> see note below

mutt/array.h
```diff
@@ -250,6 +250,52 @@
...
+ *
+ * @note The from and to indexes must be between 0 and ARRAY_SIZE(head);
+ *       the from index must not be smaller than to index.
+ */
+#define ARRAY_FOREACH_REVERSE_FROM_TO(elem, head, from, to)                    \
+  for (size_t ARRAY_FOREACH_IDX_##elem = (from) - 1;                           \
+       (ARRAY_FOREACH_IDX_##elem >= (to)) &&                                   \
```
rayfordshire, line 295: Here `from` is not used only `from - 1`, so it is exclusive,
while `to` is used with `>=`, i.e. it is inclusive.

This is inconsistent with the documentation.
I'm unsure what I find better

* [to,from[  (inclusive to, exclusive from) which is similar to how normal/forward ranges look *from an absolute point of view*
* ]to, from] (exclusive to, inclusive from) which is similar to how ranges look *in direction of traversal*, i.e. the start point is inclusive, the destination point exclusive.

What to pick?  Depends on the use case and taste but certainly
```
ARRAY_FOREACH_REVERSE_FROM_TO(elem, array, array->size, 0)
```
looks slightly better than
```
ARRAY_FOREACH_REVERSE_FROM_TO(elem, array, array->size-1, -1)
```

mutt/array.h
```diff
@@ -250,6 +250,52 @@
...
+ * @param head Pointer to a struct defined using ARRAY_HEAD()
+ */
+#define ARRAY_FOREACH_REVERSE(elem, head)                                      \
+  ARRAY_FOREACH_REVERSE_FROM_TO(elem, (head), (head)->size, 0)
+
+/**
+ * ARRAY_FOREACH_REVERSE_FROM - Iterate from an index to the beginning
+ * @param elem Variable to be used as pointer to the element at each iteration
+ * @param head Pointer to a struct defined using ARRAY_HEAD()
+ * @param from Starting index (inclusive)
```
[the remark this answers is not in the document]
flatcap: well spotted, fixed

mutt/array.h
```diff
@@ -250,6 +250,52 @@
...
+ * @note The from index must be between 0 and ARRAY_SIZE(head)
+ */
+#define ARRAY_FOREACH_REVERSE_FROM(elem, head, from)                           \
+  ARRAY_FOREACH_REVERSE_FROM_TO(elem, (head), (from), 0)
+
+/**
+ * ARRAY_FOREACH_REVERSE_TO - Iterate from the end to an index
+ * @param elem Variable to be used as pointer to the element at each iteration
+ * @param head Pointer to a struct defined using ARRAY_HEAD()
+ * @param to   Terminating index (exclusive)
```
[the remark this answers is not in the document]
flatcap: fixed

mutt/array.h
flatcap: fixed

mutt/array.h
flatcap: > This is inconsistent with the documentation.

Thanks, I've fixed all the docs.
I've added a separate commit that explicitly documents the ranges, e.g.
```c
 * @note Range: (ARRAY_SIZE(head)-1) .. to
```

It also tidies the verbal description, 'quoting' the parameters for clarity.

> `ARRAY_FOREACH_REVERSE_FROM_TO(elem, array, array->size, 0)`
> looks slightly better than
> `ARRAY_FOREACH_REVERSE_FROM_TO(elem, array, array->size-1, -1)`

Yeah, that's what I found.
Everything is swapped: (beginning, end), (from, to), (inclusive, exclusive).

There was the additional problem that the index types were `size_t`, so `-1` wasn't even possible.

PR #4560, opened 2025-01-25
Change description of verify-key to be crypto-scheme agnostic

rayfordshire:
The GPGME backend provides the verify-key (OP_VERIFY_KEY) function in the PGP and the S/MIME key select menu.  Change the description of that function to be crypto-scheme agnostic, i.e. do not mention "PGP public key" but use "public key" instead.

Related #4558


flatcap:
Ah!  Thanks.
I hadn't considered that next step -- completely remove the difference between PGP and SMIME.

I'll have a think, but...

Given that few people will be using both PGP and SMIME
and fewer still will have changed keybinding,
I might create a single `key` menu, or possibly `public_key`.

flatcap (approved):
I had a look at merging the pgp and smime menus -- not a chance! :-)
I've updated your commit to tweak `po/*.po` to reduce the translation work a little.
Thanks!

PR #4569, opened 2025-02-12
Preparation for `main()` refactoring

flatcap:
A couple of small steps towards tidying `main()`

- bf650cc9c main: drop `-B` (batch mode) command line option
  This option never actually did anything.
  (Apart from occasionally being a handy place to put a breakpoint)

- c60ddd10a global invert `OptNoCurses` to `OptGui`
  Conceptually, `if (OptGui)` is simpler that `if (!OptNoCurses)`
  Ideally, we'd eliminate all these ugly globals, but in the mean time...


flatcap:
merged: ace867087 main: drop -B (batch mode) option

needs-more-thought: global invert `OptNoCurses` to `OptGui`

PR #4570, opened 2025-02-12
Simplify the management of NeoMutt Commands

flatcap:
Three simple commits to simplify NeoMutt commands.

- f59bb1be8 commands: simplify commands_register()
- f17d72b66 commands: use const pointers
- db9ef3b95 commands: move to NeoMutt

Now, the commands are held by the NeoMutt object rather than being global.
This will allow us to register commands more easily.
(See the [Modules API PR](#4572))


flatcap:
Merged in bec9bef96

PR #4571, opened 2025-02-12
libcli - parse the command line

flatcap:
**WIP** -- Create a new library solely to parse the command line.

- bfd89eb52 libcli - parse the command line
- ec6f130bc test: libcli - test the command line
- a34d5f0e7 main: refactor command line parsing

The **current parser** has problems,

- It's buried deep within `main()`
- It uses lots of variables, which are lost amongst the others of `main()`
- It mixes parsing and acting
- It doesn't do the entire job -- some processing happens later

The **new parser** is:
- self-contained
- testable -- full coverage
- only a parser

NeoMutt has six distinct modes of operation (see [comment below](https://github.com/neomutt/neomutt/pull/4571#issuecomment-2652541997)).
Each of these has a separate `struct` for the results.

The parser works well, but its integration into `main()` is naive.

Now that I understand how the command line is meant to work,
I can tease apart the remaining large block of `main()` that does the work.

### DIFF
--- a/.cirrus.yml
+++ b/.cirrus.yml
@@ -28,8 +28,12 @@ task:
         - (cd ${NEOMUTT_TEST_DIR} && ./setup.sh)
       configure_script: ./configure --autocrypt --disable-doc --disable-idn --idn2 --notmuch --testing
       build_script: make
-      version_script: ./neomutt -v
-      test_script: make test
+      version_script:
+        - ./neomutt -v
+        - ./neomutt -h all
+      test_script:
+        - make test/neomutt-test
+        - make test
 
     - name: FreeBSD / OpenSSL
       install_script:
@@ -37,7 +41,9 @@ task:
         - ${PKG_INSTALL} ${PKG_COMMON} openssl
       configure_script: ./configure ${CONFIGURE_COMMON} --ssl
       build_script: make
-      version_script: ./neomutt -v
+      version_script:
+        - ./neomutt -v
+        - ./neomutt -h all
 
     - name: FreeBSD / LibreSSL
       install_script:
@@ -45,7 +51,9 @@ task:
         - ${PKG_INSTALL} ${PKG_COMMON} libressl
       configure_script: ./configure ${CONFIGURE_COMMON} --ssl
       build_script: make
-      version_script: ./neomutt -v
+      version_script:
+        - ./neomutt -v
+        - ./neomutt -h all
 
     - name: FreeBSD / GnuTLS
       install_script:
@@ -53,4 +61,6 @@ task:
         - ${PKG_INSTALL} ${PKG_COMMON} gnutls
       configure_script: ./configure ${CONFIGURE_COMMON} --gnutls
       build_script: make
-      version_script: ./neomutt -v
+      version_script:
+        - ./neomutt -v
+        - ./neomutt -h all
--- a/.github/workflows/asan.yml
+++ b/.github/workflows/asan.yml
@@ -43,6 +43,14 @@ jobs:
     - name: Configure Neomutt
       run: ./configure $CONFIGURE_OPTIONS
 
+    - name: Build Neomutt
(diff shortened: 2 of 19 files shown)


flatcap:
# Parsing the CLI

## Help

Help options for info about Neomutt.
They are executed before any config is read.
(We want the help displayed, even if the config is bad)

They exit immediately.

    -h            Print this help message and exit
    -v            Print the NeoMutt version and compile-time definitions and exit
    -vv           Print the NeoMutt license and copyright information and exit

## Config

These options affect how config is loaded.

    -F <config>   Specify an alternative initialization file to read
    -n            Do not read the system-wide configuration file

## Logging

These options affect how and where NeoMutt logs.

    -d <level>    Log debugging output to a file (default is "~/.neomuttdebug0")
                  The level can range from 1-5 and affects verbosity
    -l <file>     Specify a file for debugging output (default "~/.neomuttdebug0")

## Info

These options query NeoMutt for configuration information.
They are executed after the config is read.
They exit after display.

    -A <alias>    Print an expanded version of the given alias to stdout and exit
    -D            Dump all config variables as 'name=value' pairs to stdout
    -D -D         (or -DD) Like -D, but only show changed config
    -D -O         Like -D, but show one-liner documentation
    -D -S         Like -D, but hide the value of sensitive variables
    -Q <variable> Query a configuration variable and print its value to stdout
                  (after the config has been read and any commands executed)
                  Add -O for one-liner documentation

## Send

These options allow the user to send email from the command line.
They **may** start the GUI to collect more info.

    -a <file>     Attach one or more files to a message (must be the last option)
                  Add any addresses after the '--' argument
    -b <address>  Specify a blind carbon copy (Bcc) recipient
    -c <address>  Specify a carbon copy (Cc) recipient
    -C            Enable Command-line Crypto (signing/encryption)
    -E            Edit draft (-H) or include (-i) file during message composition
    -H <draft>    Specify a draft file with header and body for message composing
    -i <include>  Specify an include file to be embedded in the body of a message
    -s <subject>  Specify a subject (must be enclosed in quotes if it has spaces)
    [non-option]  Additional arguments
    --            Special argument forces NeoMutt to stop option parsing and treat
                  remaining arguments as addresses even if they start with a dash

## GUI

These options affect the startup of the GUI.

    -e <command>  Specify a command to be run after reading the config files
    -f <mailbox>  Specify a mailbox (as defined with 'mailboxes' command) to load
    -G            Start NeoMutt with a listing of subscribed newsgroups
    -g <server>   Like -G, but start at specified news server
    -m <type>     Specify a default mailbox format type for newly created folders
                  The type is either MH, MMDF, Maildir or mbox (case-insensitive)
    -p            Resume a prior postponed message, if any
    -R            Open mailbox in read-only mode
    -y            Start NeoMutt with a listing of all defined mailboxes

Additionally, these queries **optionally** start the GUI.

    -Z            Open the first mailbox with new message or exit immediately with
                  exit code 1 if none is found in all defined mailboxes
    -z            Open the first or specified (-f) mailbox if it holds any message
                  or exit immediately with exit code 1 otherwise
## Priority

The six CLI option groups have to be considered in order:

- **Parse**: Parse the Command Line into structs, but don't perform any actions yet

- **Help**: If help options are given
    - Display help
    - Exit

- **Config**: Read the system and user config files
  (honouring the `-n` and `-F` options)

- **Logging**: Open the log file
  (honoring the config, `-d` and `-l` options)

- **Info**: If info options are given
    - Display info
    - Exit

- **Send**: Send an email from the Command Line
    - GUI: Gather more info (optional)
    - Send email
    - Exit

- **GUI**: Start the NeoMutt GUI
  The options affect where the user starts, e.g.
  - `-p` Postponed emails
  - `-y` Mailbox browser

flatcap:
**Updates**:
- Logging needed splitting into early/late
- Parsing now uses `ARRAY`s -- simpler and easier
- Notes updated
- Still testing all the possible ways to run NeoMutt

PR #4572, opened 2025-02-12
Create a Module API

flatcap:
**WIP** -- Create a Module API

This branch _works_, but it's may be fragile.

Any time something needs initialising, somebody throws a new line in `main()`.
Somewhere.  Anywhere.

`main()` had grown to **1025 lines**
With this PR, `main()` is down to **693 lines**
It'll get smaller still, soon.

The Module API allows libraries to do their own initialisation and cleanup.
NeoMutt still manages some key services: locale, logging, config, commands.
NeoMutt has (nearly) zero-knowledge of the Modules.

The Libraries / Modules can decide when they need to get called to initialise themselves.
e.g. After the config is setup, Before the GUI starts

- e4e322197 core: define Module API
- 571af30db core: create empty modules
- bfa0bdf00 core: call modules from neomutt
- 1bc787a10 core: use modules for initialisation

Current API
```c
// Initialise a Module
bool init(struct NeoMutt *n);

// Set up Config Types
bool config_define_types(struct NeoMutt *n, struct ConfigSet *cs);

// Define the Config Variables
bool config_define_variables(struct NeoMutt *n, struct ConfigSet *cs);

// Register NeoMutt Commands
bool commands_register(struct NeoMutt *n, struct CommandArray *ca);

// Initialise the GUI
bool gui_init(struct NeoMutt *n);

// Clean up the GUI
void gui_cleanup(struct NeoMutt *n);

// Clean up a Module
void cleanup(struct NeoMutt *n);
```

Managing the config and commands was fairly easy.
The next, more challenging, step is to manage the GUI startup.

Currently, NeoMutt performs some GUI startup "just in case".
Soon, the GUI startup will executed on demand.


Yedfree (approved):

flatcap:
Hi @Yedfree.
Thanks for the feedback / vote of confidence :-)

The changes are starting to look good, but I've hit some roadblocks:
- The huge mass of tangled code in `main()` (that does the actual work)
- `mutt_send_message()` -- the sending code that has tons of GUI dependencies

This PR might have to remain a proof-of-concept, until I have the time to analyse and untangle the problems.

flatcap:
`mutt_send_message()` -- Callers and children:

image

PR #4587, opened 2025-03-09
body_handler: enriched handler without fmemopen

### LINKED ISSUE #4585: Crash with fmemopen and multi-part email
Viewing a multi-part email with a `text/enriched` part crashes.
This only happens with `fmemopen` enabled `./configure --fmemopen`

A `git bisect` takes me back to 2016 when the patch was first merged.

**Steps**:
- Create test email, below
- Run: `neomutt -n -F /dev/null -f test.mbox -e "push <enter>"`

---

Sample email: **`test.mbox`**
```
From test@example.com Fri Mar  7 12:52:48 2025
	by flatcap.org (Postfix) with ESMTP id 0DAB2820
	for <flatcap@flatcap.org>; Fri, 07 Mar 2025 12:53:08 +0000 (GMT)
To: Richard Russon <rich@flatcap.org>
Subject: TEST
From: TEST <test@example.com>
MIME-Version: 1.0
Content-Type: multipart/alternative; boundary="----42a79dd669260e1125a3dd3a1c1fbfba"
Date: Fri, 07 Mar 2025 12:56:02 +0000 (GMT)

------42a79dd669260e1125a3dd3a1c1fbfba

PLAIN

------42a79dd669260e1125a3dd3a1c1fbfba
Content-Type: text/enriched; charset=us-ascii

ENRICHED

------42a79dd669260e1125a3dd3a1c1fbfba--
```

luchr:
# attempt to fix text enriched handler with `USE_FMEMOPEN`

This is an attempt to fix #4585 .

The `text_enriched_handler` uses `fgetwc` which is not supported by glibc's `fmemopen` files.

This patch adds the `no_fmemopen` parameter to `run_decode_and_handler` such that `fmemopen`-calls can be prohibited by the caller.

Adds a `no_fmemopen` flag in `mutt_body_handler` in order to set this flag if "text/enriched" is encountered.

### DIFF
--- a/handler.c
+++ b/handler.c
@@ -1316,17 +1316,26 @@ static int multipart_handler(struct Body *b_email, struct State *state)
   return rc;
 }
 
+struct Tempdata {            /* either temp. memory buffer or tempfile name    */
+  bool use_fmemopen;
+  struct Buffer *tempfile;   /* if fmemopen is not used: filename for tempfile */
+  char *temp;                /* if fmemopen is used:  temp-Buffer  with ...    */
+  size_t tempsize;           /*                       its size                 */
+};
+
 /**
  * run_decode_and_handler - Run an appropriate decoder for an email
  * @param b         Body of the email
  * @param state         State to work with
  * @param handler   Callback function to process the content - Implements ::handler_t
  * @param plaintext Is the content in plain text
+ * @param no_fmemopen Are fmemopen-calls disallowed
  * @retval 0 Success
  * @retval -1 Error
  */
 static int run_decode_and_handler(struct Body *b, struct State *state,
-                                  handler_t handler, bool plaintext)
+                                  handler_t handler, bool plaintext,
+                                  bool no_fmemopen)
 {
   const char *save_prefix = NULL;
   FILE *fp = NULL;
@@ -1334,19 +1343,16 @@ static int run_decode_and_handler(struct Body *b, struct State *state,
   LOFF_T tmpoffset = 0;
   int decode = 0;
   int rc = 0;
+  struct Tempdata tempdata = { 0 };
+
 #ifndef USE_FMEMOPEN
-  struct Buffer *tempfile = NULL;
+  no_fmemopen = true; /* never use fmemopen */
 #endif
 
   if (!mutt_file_seek(state->fp_in, b->offset, SEEK_SET))
   {
     return -1;
   }
-
-#ifdef USE_FMEMOPEN
-  char *temp = NULL;
-  size_t tempsize = 0;
-#endif
 
   /* see if we need to decode this part before processing it */
   if ((b->encoding == ENC_BASE64) || (b->encoding == ENC_QUOTED_PRINTABLE) ||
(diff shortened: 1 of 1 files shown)


gahr:
I don't particularly like this approach, it breaks separation of concern: the idea of passing a `handler` to `run_decode_and_handler` is so that `run_decode_and_handler` doesn't need to care which handler it is. We shouldn't further parameterize the function based on which particular handler it's called with.

> The `text_enriched_handler` uses `fgetwc`

Why does it need to do that, in the first place?

luchr:
> Ah yeah, that makes sense.. I wonder, however, how other handlers do that. I guess text/plain can be UTF-8 or anything else just fine, no?

   * `text_plain_handler` uses `mutt_file_read_line` and is encoding agnostic  (as long as '\n', '\r', ' ', '-' are encoded with their ASCII values)
   * `message_handler` (rfc822)  is a purely 7-bit ASCII standard
   * `external_body_handler`  calls many (sub-)handlers;some are either encoding-agnostic (showing length of attachments in banner, etc.) or some are byte-based crypt-handlers, etc.

So let's look at some more interesing handlers:

   * `rfc3676_handler` (the f=f handler, format=flowed) uses `mutt_file_read_line` and then (inside `print_flowed_line`)
     iterates over all words (separated by " ") and uses `mutt_strwidth` to measure the length in screen cells and
     `mutt_strwidth` uses `mbrtowc` and then `wcwidth`.

This is the only handler(!) I've found which needs the concept of (multibyte/unicode-)characters [of course in other neomutt parts, e.g. the pager, this concept is also needed and used]. Here a `mbrtowc`-loop is used to get single multibyte-characters.

So there is another alternative:
Rewrite the enriched parser to use (internally) lines and `mbrtowc`-loops on each line instead of the current (mb-)character LALR-1 parsing. Because of the "n-1" rule in [RFC1563](https://datatracker.ietf.org/doc/html/rfc1563) this is not trivial. And additionally it's typically more complicated to write an stateful ('state' are lines and already processed mb-characters here) parser than a more "simpler" LALR-1 parser.

luchr:
I try a short summary of possible ways to fix the problem:

   1. special handlers (like the `text_enriched_handler`) need a full-fledged file-handle (and `fmemopen` does **not** return such a full-fledged file-handle), so invent a new parameter to `run_decode_and_handler` for a flag if a full-fledged file-handle is needed (this PR). [The caller which sets the handler has to provide this additional flag.]
   2. keep the LALR-1 parser of the current `text_enriched_handler` but rewrite it to use a byte-level "easy" encoding (e.g. a fixed length encoding UTF-32/UCS-4). [The caller which sets the handler has to provide the text in the right encoding.]
   3.  give up on the LALR-1 idea of the current `text_enriched_handler` and use `mutt_file_read_line` together with `mbrtowc`-loops to handle the single multibyte-characters. [No special steps needed by the caller.]

I don't have any preferences (and I consider this a tough decision).

gahr:
Hi @luchr , thanks for your detailed analysis. I favour approach 3, I think it's the cleanest.

luchr:
Please see #4588 for the next approximation step.

PR #4588, opened 2025-03-12
enriched: handle not wide-char oriented streams

luchr:
The rich text handler uses wide characters, but some file streams do not support fgetwc/ungetwc (e.g. streams created with fmemopen in glibc).

The rich text handler now takes care of this streams with a very thin layer which either calls fgetwc/ungetwc (if supported) or otherwise uses a buffer, fread and mbrtowc to create wide characters.

------

I don't expect this PR to be merged but to be a basis/foundation of discussion (just like #4587) to come closer to a solution that make many persons happy.

### DIFF
--- a/enriched.c
+++ b/enriched.c
@@ -114,6 +114,175 @@ struct EnrichedState
   struct State *state;
 };
 
+
+#define WL_BUFSIZE 1024
+
+/**
+ * struct WLgetwc - Wide layer/proxy for fgetwc and ungetwc
+ *
+ * If FILE-handle supports wide-orientation then just use fgetwc and ungetwc.
+ * If FILE-handle does not support wide-orientation (e.g. created with fmemopen)
+ * then use fread to read bytes of the FILE-handle to the buffer and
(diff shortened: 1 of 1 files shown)


gahr:
I think the approach makes sense to me. I will try to take a closer look at the code further on. Thanks!

luchr:
Hello.

(a) "won't fix" approach (but without segfault)
No problem. But even in the "won't fix" approach segfaulting is not nice. What about a `if`-statement testing with `fwide` and returning/rendering a message "Sorry. text/enriched emails are not supported with the fmemopen compile time flag." instead of segfaulting (see issue #4585).

(b) "wait" approach
Or just leave the PR open for later (in a few month or so); perhaps you or somebody else find the time to look at it. I'll try to include all the feedback.

gahr:
Mh.. I like the approach of a handler saying it's not able to render a message.. perhaps we could always fallback to `text_plain_handler` as a worst-case scenario.. so you'll get the tags in your pager instead of a crash or a "not able to render" error.

PR #4595, opened 2025-03-26
notmuch: check for parse failure

flatcap:
Try to cope if a message disappears.

---

Fixes: #4594 

This fixes a problem introduced by f5d55b4b7f which was a fix for #4474 

The fix is crude but I don't understand the situation that causes the problem.


flokli:
I ran into the same crash reported in https://github.com/neomutt/neomutt/issues/4594, and applying that patch on top of `neomutt-20250113` did fix it for me. Thanks!

PR #4596, opened 2025-03-26
query: allow <>s around email addresses

flatcap:
External address queries should be of the form:

    name@example.com<tab>Real Name

Historically, (Neo)Mutt would accept email addresses that were wrapped in `<>`s.

This feature stopped working in 5d8a41679 alias: unify query parsing

---

**This PR restores the old (undocumented) behaviour.**

**Is this what we want?**


scottkosty:
I don't have a preference, but if it's not supported, would it be easy to at least give an error message so the user knows the issue? I don't think users would mind removing the angle brackets if they know they cause problems.

gahr (approved):
LGTM

PR #4605, opened 2025-04-09
help: fix leaks

flatcap:
Fix memory leaks from <help>.

**Steps**:
- Build NeoMutt with ASAN
- Open NeoMutt
- Hit <kbd>h</kbd> (`<help>`)
- Quit Help and NeoMutt

**Expected**:
- No leaks :-)


gahr:
I'm sure I'm missing something obvious, but why not `bia_unbound`?

flatcap:
For `bia_{bind,macro,gen}` have generated strings, e.g. key bindings.
`bia_unbound` just uses static strings, e.g. from `OpGeneric[]` and `OpStrings[]`.

gahr (approved):

PR #4618, opened 2025-04-30
Build and test on FreeBSD

gahr:
Use https://github.com/vmactions/freebsd-vm/ to build & test on FreeBSD

### DIFF
--- a/.github/workflows/build-and-test-freebsd.yml
+++ b/.github/workflows/build-and-test-freebsd.yml
@@ -8,6 +8,9 @@ on:
       - 'devel/**'
   workflow_dispatch:
 
+concurrency:
+  group: ${{ github.workflow }}
+
 jobs:
   build:
     if: github.event.pull_request.draft == false
--- a/.github/workflows/build-and-test-freebsd.yml
+++ b/.github/workflows/build-and-test-freebsd.yml
@@ -1,0 +1,52 @@
+name: Build and Test (FreeBSD)
+
+on:
+  pull_request:
+  push:
+    branches:
+      - 'main'
+      - 'devel/**'
+  workflow_dispatch:
+
+jobs:
+  build:
+    if: github.event.pull_request.draft == false
+    name: Build
+    runs-on: ubuntu-latest
+
+    steps:
+    - name: Get number of CPU cores
+      uses: SimenB/github-actions-cpu-cores@v2
+      id: cpu-cores
+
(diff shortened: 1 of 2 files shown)


.github/workflows/build-and-test-freebsd.yml
```diff
@@ -0,0 +1,52 @@
...
+      with:
+        usesh: true
+        prepare: |
+          pkg install -y perl5 pkgconf gettext libxslt docbook-xsl db5 gdbm gpgme gnutls kyotocabinet tokyocabinet lmdb lua54 pcre2 notmuch qdbm rocksdb cyrus-sasl tdb zstd
+        run: |
+          cd test-files
+          ./setup.sh
+          cd -
+          mkdir build
+          cd build
```
flatcap, line 42: Any reason for building out-of-tree?
[`build-and-test.yml`](https://github.com/neomutt/neomutt/blob/main/.github/workflows/build-and-test.yml) already checks this.

.github/workflows/build-and-test-freebsd.yml
```diff
@@ -0,0 +1,52 @@
...
+          ../configure --full-doc --autocrypt --bdb --fmemopen --gdbm --gnutls --gpgme --gss --kyotocabinet --lmdb --lua --lz4 --notmuch --pcre2 --qdbm --rocksdb --sasl --tdb --testing --tokyocabinet --with-lock=fcntl --zlib --zstd
+          make -j ${{steps.cpu-cores.outputs.count}} neomutt
+          ./neomutt -v
+          ./neomutt -h all
+          pwd
+          ls -C
```
flatcap, line 48: I was going to ask why these two lines are here, but I've just spotted that you've copy'n'pasted from `build-and-test.yml`.
Would you mind killing my leftover debugging code, [there](https://github.com/neomutt/neomutt/blob/main/.github/workflows/build-and-test.yml) too.
Thanks :-)

.github/workflows/build-and-test-freebsd.yml
gahr: no, I can skip that

gahr:
There seems to be an issue that makes us wait indefinitely for the VM to start: https://github.com/vmactions/freebsd-vm/issues/91. Let's wait a bit before merging this, in case anything comes out of that issue.

flatcap (approved):

PR #4628, opened 2025-05-27
Use the standard `countof` instead of our `mutt_array_size()`

alejandro-colomar:
* **What does this PR do?**

   Replace an custom macro by a standard one.

    Link: <https://www.open-std.org/jtc1/sc22/wg14/www/docs/n3550.pdf#subsubsection.0.6.5.4.5>
    Link: <https://www.open-std.org/jtc1/sc22/wg14/www/docs/n3550.pdf#section.0.7.21>
    Link: <https://gcc.gnu.org/bugzilla/show_bug.cgi?id=117025>
    Link: <https://github.com/llvm/llvm-project/issues/102836>

---

Revisions:

<details>
<summary>v2</summary>

-  Use `__has_include` only if it's supported.  [@gahr ]

```
$ git range-diff neomutt/main neomutt/alx/countof alx/countof 
1:  c1cd00643 ! 1:  747fd6cda Use the standard countof instead of our mutt_array_size()
    @@ mutt/memory.h
      #ifndef MUTT_MUTT_MEMORY_H
      #define MUTT_MUTT_MEMORY_H
      
    -+#if __has_include(<stdcountof.h>)
    -+# include <stdcountof.h>
    ++#if defined __has_include
    ++# if __has_include(<stdcountof.h>)
    ++#  include <stdcountof.h>
    ++# endif
     +#endif
      #include <stddef.h>
      
```
</details>

<details>
<summary>v2b</summary>

-  Reviewed-by: @gahr 

```
$ git range-diff neomutt/main neomutt/alx/countof alx/countof 
1:  747fd6cda ! 1:  85353b1ea Use the standard countof instead of our mutt_array_size()
    @@ Commit message
         Link: <https://www.open-std.org/jtc1/sc22/wg14/www/docs/n3550.pdf#section.0.7.21>
         Link: <https://gcc.gnu.org/bugzilla/show_bug.cgi?id=117025>
         Link: <https://github.com/llvm/llvm-project/issues/102836>
    +    Reviewed-by: Pietro Cerutti <gahr@gahr.ch>
     
      ## config/quad.c ##
```
</details>

### DIFF
--- a/mutt/memory.h
+++ b/mutt/memory.h
@@ -23,6 +23,11 @@
 #ifndef MUTT_MUTT_MEMORY_H
 #define MUTT_MUTT_MEMORY_H
 
+#if defined __has_include
+# if __has_include(<stdcountof.h>)
+#  include <stdcountof.h>
+# endif
+#endif
 #include <stddef.h>
 
 #undef MAX
@@ -35,7 +40,9 @@
 #undef ROUND_UP
 #define ROUND_UP(NUM, STEP) ((((NUM) + (STEP) -1) / (STEP)) * (STEP))
 
-#define mutt_array_size(x) (sizeof(x) / sizeof((x)[0]))
+#if !defined(countof)
+# define countof(x)  (sizeof(x) / sizeof((x)[0]))
+#endif
 
 #define MUTT_MEM_CALLOC(n, type)  ((type *) mutt_mem_calloc(n, sizeof(type)))
 #define MUTT_MEM_MALLOC(n, type)  ((type *) mutt_mem_mallocarray(n, sizeof(type)))
--- a/config/quad.c
+++ b/config/quad.c
@@ -118,7 +118,7 @@ static int quad_string_get(void *var, const struct ConfigDef *cdef, struct Buffe
   else
     value = (int) cdef->initial;
 
-  if (value >= (mutt_array_size(QuadValues) - 1))
+  if (value >= (countof(QuadValues) - 1))
   {
     mutt_debug(LL_DEBUG1, "Variable has an invalid value: %d\n", value); /* LCOV_EXCL_LINE */
     return CSR_ERR_INVALID | CSR_INV_TYPE; /* LCOV_EXCL_LINE */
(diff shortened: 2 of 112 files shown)


mutt/memory.h
```diff
@@ -23,6 +23,9 @@
 #ifndef MUTT_MUTT_MEMORY_H
 #define MUTT_MUTT_MEMORY_H
 
+#if __has_include(<stdcountof.h>)
```
gahr, line 26: I think we might need to guard the whole thing by `#if defined __has_include`.

alejandro-colomar: > I think we might need to guard the whole thing by `#if defined __has_include`.

I don't think that's valid.  :)

alejandro-colomar: > Or check for the header file in our configure script, like we do for other headers

Hmmm, okay, that's a valid thing to do.

On the other hand, which compilers do we support?  `__has_include` is standard ISO C23, and has been supported by compilers for a long-ish time:

<https://en.cppreference.com/w/c/compiler_support.html>

-  GCC 5 (2015)
-  Clang (forever?)
-  MSVC 19.11 (2017)
-  Intel C++ 18.0 (2017)

alejandro-colomar: > > > I think we might need to guard the whole thing by `#if defined __has_include`.
> > 
> > 
> > I don't think that's valid. :)
> 
> I think it is: https://gcc.gnu.org/onlinedocs/cpp/_005f_005fhas_005finclude.html

Hmmm, interesting.  I don't know if the other implementations behave like that, but it could work.

alejandro-colomar: Done, in v2.  Thanks!

mutt/memory.h
```diff
@@ -23,6 +23,11 @@
 #ifndef MUTT_MUTT_MEMORY_H
 #define MUTT_MUTT_MEMORY_H
 
+#if defined __has_include
+# if __has_include(<stdcountof.h>)
+#  include <stdcountof.h>
+# endif
+#endif
```
gahr, line 30: How about

```
#if defined(__has_include) && __has_include(<stdcountof.h>)
#  include <stdcountof.h>
#else
#  define countof(x) (sizeof(x) / sizeof((x)[0]))
#endif
```

ossilator: that's likely to cause a parse error if the function is not defined.

gahr (approved):

alejandro-colomar:
Thanks!  Will you merge, or should we wait for @flatcap ?

Issue #3519, opened 2022-09-14
issue with syncing on gmail account


euglevi:
I have a weird issue when syncing the mailbox in my gmail account. When I read a new email, sync the mailbox, quit and get back in again, the email is correctly shown as read. When I add to these steps deleting a few emails before or after reading the new email, then the read email is shown up as new again on entering again neomutt. What could it be?

If needed, I can provide my configuration files and the neomutt log.

gahr:
is this a local maildir sync'd from IMAP using something like mbsync?

brauner:
No, it's just using the built-in imap capabilities of neomutt. The config is mostly:

```
## Receive options.
set imap_user=foo@bar.baz
set folder=imaps://imap.gmail.com
set imap_pass=$my_pw
set spoolfile=imaps://imap.gmail.com/INBOX
set record="imaps://imap.gmail.com/[Gmail]/Sent Mail"
set postponed="imaps://imap.gmail.com/[Gmail]/Drafts"
set trash='+[Gmail]/Trash'

## Send options.
set realname='My Name'
set from='my@mail.com'
set use_from=yes
set smtp_url     = "smtp://brauner@mail.kernel.org:587/"
set smtp_pass    = $my_other_pass
set ssl_starttls = yes
```

brauner:
As far as I can tell it started when I updated my trash settings `set trash='+[Gmail]/Trash'`

euglevi:
I moved to using a combination of lieer and notmuch (which works very well), but I was not using mbsync. I was using IMAP with [oauth2 authentication](https://github.com/neomutt/neomutt/blob/main/contrib/oauth2/mutt_oauth2.py.README).  I had the same folders as @brauner.

gahr:
Interesting.. I cannot reproduce this. So you say

* open a GMail mailbox with new mail
* delete an (old) email
* read the new mail
* quit
* reopen
* -> the email you read is still marked as new?

gahr:
now I can repro it with some minimal config

gahr:
Indeed, I can see that in the bad case (delete + read), we're not sending a `UID STORE <uid> +FLAGS.SILENT (\Seen)`

brauner:
Did this ever get fixed? I'm still having issues with NeoMutt 20240425

gahr:
Hi @brauner , no, I fear nobody's looked at it and *I* forgot. Do you want to try to take a look? I don't think I'll be able to get to it soon(ish).

[...]

PR #4629, opened 2025-06-02
Ensure that read mails are synced when expunging deleted mails

brauner:
This issue has been plaguing me for years and I can't take it anymore. When marking mails as read and deleting mails and syncing the deleted mails to the server the mails marked as read will not be synced to the servers. So next time the the mailbox is opened again they appear as unread. It's really terrible especially when having gone through thousands of mail. Sync seen mails as well.

Fixes: #3519

* **What does this PR do?**

Ensure that mails that are seen are synced when syncing deleted mails.

     for that)

* 
#3519

### DIFF
--- a/imap/imap.c
+++ b/imap/imap.c
@@ -1504,7 +1504,10 @@ enum MxStatus imap_sync_mailbox(struct Mailbox *m, bool expunge, bool close)
     struct UidArray uida = ARRAY_HEAD_INITIALIZER;
     select_email_uids(m->emails, m->msg_count, MUTT_DELETED, true, false, &uida);
     ARRAY_SORT(&uida, imap_sort_uid, NULL);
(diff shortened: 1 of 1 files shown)


gahr:
I'm on a mobile, so I can't properly review, but I think I read your changes as making all deleted emails be marked as read.

Are you talking about two different sets of emails (read and deleted) and only the status of the second set being sync'd? If so, how does your patch fix that?

gahr:
Also: won't this also mark as read a new mail that gets deleted? I think that'd be equally wrong.

brauner:
Frankly, I have zero idea how this works and the only reason I even touched this code is because I've been suffering from this misbehavior for years now. So I'm willing to fix this but I'll need guidance.

PR #4635, opened 2025-06-21
Handle a non-existing message_cache_dir

### LINKED ISSUE #4583: Behvaviour of message_cachedir
Hi, I've a question about `message_cachedir` and `header_cache`

In my rc file, I've set

~~~~
set header_cache = ~/.cache/neomutt/headers
set message_cachedir = ~/.cache/neomutt/bodies
~~~~

Question 1, and also not that important: Instead of writing `~/.cache`, how can I point to `$XDG_CACHE_HOME/neomutt/...` if `XDG_CACHE_HOME` defined, and fallback to `~/.cache/neomutt/...`, as described in the xdg specification?

Question 2; the actual question I had:

If I open a message, I get following error `Cache disabled, $message_cache_dir isn't a directory: /home/user/.cache/neomutt/bodies`, and the output of `ls /home/user/.cache/neomutt` is `headers`.

Thus neomutt created the header file, but not the folder for the cache directory.
If I create the folder `/home/user/.cache/neomutt/bodies`, then everything works correctly.

Shouldn't neomutt create the `bodies` folder, since it had no issues creating the `~/.cache/neomutt` folder (I assume because of `header_cache`)?

gahr:
The code was previously requiring `message_cache_dir` to exist *and* be a directory. We now only require that it is a directory *if* it exists. If it doesn't exist, we already take care of creating it at `mutt_bcache_put` time.

The check was introduced in #2661. @louis77, do you have an issue with the fix I propose?

Fixes #4583

### DIFF
--- a/bcache/bcache.c
+++ b/bcache/bcache.c
@@ -65,7 +65,7 @@ static int bcache_path(struct ConnAccount *account, const char *mailbox, struct 
     return -1;
 
   struct stat st = { 0 };
-  if (!((stat(c_message_cache_dir, &st) == 0) && S_ISDIR(st.st_mode)))
+  if ((stat(c_message_cache_dir, &st) == 0) && !S_ISDIR(st.st_mode))
   {
     mutt_error(_("Cache disabled, $message_cache_dir isn't a directory: %s"), c_message_cache_dir);
     return -1;


louis77:
No issues here! Thanks @gahr.

roccoblues (approved):
Yes, makes sense. 👍🏼

PR #4642, opened 2025-07-01
 Decouple `$crypt_encryption_info` and `$crypt_protected_headers_weed` from `$devel_security`

### LINKED ISSUE #4640: Config `$crypt_encryption_info` has no effect
Setting or unsetting the option `$crypt_encryption_info` has no visible effect, i.e. the `[-- Begin encryption information --]` is missing when viewing an email.

## NeoMutt Version

```
20250113
```

rayfordshire:
* **What does this PR do?**

Decouple `$crypt_encryption_info` and `$crypt_protected_headers_weed` from `$devel_security`, i.e. `$devel_security` does not need to be set in order for the variables to take effect.

Note that with `$devel_security` unset`$crypt_protected_headers_weed` only applies to the header "Subject", which is the only protected header neomutt supports without `$devel_security` (At least that's how I read the code).

Fixes #4640

### DIFF
--- a/ncrypt/crypt.c
+++ b/ncrypt/crypt.c
@@ -1132,9 +1132,7 @@ int mutt_protected_headers_handler(struct Body *b_email, struct State *state)
   const int wraplen = display ? mutt_window_wrap_cols(state->wraplen, c_wrap) : 0;
   const CopyHeaderFlags chflags = display ? CH_DISPLAY : CH_NO_FLAGS;
   struct Buffer *buf = buf_pool_get();
-  bool weed = (display && c_weed);
-  if (c_devel_security)
-    weed &= c_crypt_protected_headers_weed;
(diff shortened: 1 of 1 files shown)


rayfordshire:
Note that the documentation of `$crypt_protected_headers_weed` also does not mention `$devel_security`, thus I also removed that requirement.

alejandro-colomar:
Hmmm, I think `$crypt_protected_headers_weed` was intentionally (and temporarily) hidden under the devel flag.  But I've been using it flawlessly for over a year, so I think we should decouple it from the devel flag.

LGTM.  Thanks!

alejandro-colomar:
A thing that we might want is to be able to use H a second time to jump to right after these protected headers.  This was mentioned by someone during the discussion of the security features.

PR #4644, opened 2025-07-22
Provide an option to confirm on an empty To

### LINKED ISSUE #4643: Sent mail only to CC, but empty TO field.
Recently I made quite a blunder in an important email.

This email was supposed to go to a specific recipient, and CC'ed to several other recipients for reference.
For some reason, I forgot to fill in the TO field.  All CC'ed recipients got that mail just fine.

Shouldn't there be a warning if we CC a mail with an empty TO field?  I was surprised to learn that at least our mail server here just processed the mail without any hesitation.

gahr:
Fixes #4643

### DIFF
--- a/docs/config.c
+++ b/docs/config.c
@@ -744,6 +744,13 @@
 ** mailbox which does not yet exist before creating it.
 */
 
+{ "confirm_empty_to", DT_BOOL, false },
+/*
+** .pp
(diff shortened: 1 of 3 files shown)


gahr:
> Worth noting that neomutt wouldn't let me start composing an email without specifying a To address. So it's not too easy to hit accidentally.

Yep, you need to be in the compose menu and explicitly kill the To line :)

PR #4647, opened 2025-08-05
Don't use wide chars inside text/enriched handler

### LINKED ISSUE #4585: Crash with fmemopen and multi-part email

gahr:
This is a proposal to drop the use of wide chars in the text/enriched handler. I haven't found a case where this breaks, but I don't have a huge amount of text/enriched emails in my mailbox.

This *could* break with text/enriched and multi-byte character sets where the open tag marker `<` appears as part of a multi-byte sequence.

Fixes #4585

### DIFF
--- a/enriched.c
+++ b/enriched.c
@@ -34,8 +34,6 @@
 #include <stdbool.h>
 #include <stddef.h>
 #include <stdio.h>
-#include <wchar.h>
-#include <wctype.h>
 #include "mutt/lib.h"
 #include "email/lib.h"
 #include "enriched.h"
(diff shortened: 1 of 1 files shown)


gahr:
@luchr how about this, instead?

luchr:
This makes all the "Fill/Justification/Indentation Commands" from [RFC 1896](https://www.rfc-editor.org/rfc/rfc1896.html) horrible (i.e. non working) for languages with unicode codepoints that have more than one byte. Here are example screenshots for **centered** text with different window widths (starting with many columns to less columns): [Open each image in an own browser tab to see the effect better; here in the comment section they are scaled to the max width; so it seems that the font is getting larger].

## Now/working:







## With this PR; breaking:







With this PR, the three letter "абв" (which have in UTF-8 the byte sequence `d0 b0 d0 b1 d0 b2`) are treated with width 6, that's why the text is not in the centre any more. 

Please, please don't kill this working(!) feature.

gahr:
Right, I hadn't thought about that! Thanks for pointing it out!

PR #4659, opened 2025-08-28
Include term.h and [n]curses.h consistently

### LINKED ISSUE #4658: Missing header for tigetstr
When compiling on NetBSD against the native curses, there is a compilation warning:
```
gui/terminal.c: In function ‘mutt_tigetstr’:
gui/terminal.c:59:15: warning: implicit declaration of function ‘tigetstr’; did you mean ‘wgetstr’? [-Wimplicit-function-declaration]
   59 |   char *cap = tigetstr(name);
      |               ^~~~~~~~
      |               wgetstr
gui/terminal.c:59:15: warning: initialization of ‘char *’ from ‘int’ makes pointer from integer without a cast [-Wint-conversion]
```
The [NetBSD man page for tigetstr](https://man.netbsd.org/tigetstr.3) says to include `term.h`.

The [linux man page for tigetstr](https://linux.die.net/man/3/tigetstr) also lists both `curses.h` and `term.h`.

The patch is trivial, adding:
```
#include <term.h>
```
but I'd like to have some advice in which place to put the include to avoid having to rework the patch multiple times.

gahr:
The logic to include `[n]curses.h` and `term.h` was not consistent.

Fixes #4658

### DIFF
--- a/gui/terminal.c
+++ b/gui/terminal.c
@@ -33,10 +33,21 @@
 #include "mutt/lib.h"
 #include "terminal.h"
 #include "mutt_curses.h"
+
+/*
+ * The inclusion logic must be kept in sync with that of gui/mutt_curses.h.
+ * We aren't including term.h there because it pollutes the global name space
+ * with a lot of common names that we use in our source code, e.g., `pad_char`,
+ * so we'd like to scope its inclusion to this source file.
+ */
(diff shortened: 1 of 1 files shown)


roccoblues (approved):
lgtm

PR #4660, opened 2025-09-02
convert the ctype wrappers to use bool

flatcap:
Change the return type of the ctype wrappers, to `bool`.

- `bool mutt_isalnum (int arg);`
- `bool mutt_isalpha (int arg);`
- `bool mutt_isdigit (int arg);`
- `bool mutt_ispunct (int arg);`
- `bool mutt_isspace (int arg);`
- `bool mutt_isxdigit(int arg);`

All the callers were using the functions in a `bool`-y way, so they didn't need to be changed.

### DIFF
--- a/mutt/ctype.c
+++ b/mutt/ctype.c
@@ -29,89 +29,90 @@
  */
 
 #include <ctype.h>
+#include <stdbool.h>
 
 /**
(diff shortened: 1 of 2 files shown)


roccoblues (approved):
lgtm

PR #4665, opened 2025-09-06
Revamp the S/MIME docs

gahr:
I've recently re-acquired a S/MIME certificate and I went through the process of setting it up in NeoMutt. I've re-organized a bit our S/MIME docs, reimported `smime.rc` from the neomutt-contrib repo because smime-notes.txt refers to it, and generally cleaned up a bit.

Long term, I think we should re-think our S/MIME machinery, do more in C, and eventually get rid of the `smime_keys` perl script. But that's for another time :)


flatcap (approved):
Nice one!

> ... I think we should re-think our S/MIME machinery ...

You'd have to explain it to me first :-)

Long ago, I acquired a free certificate and managed to get it set up,
but a year later it expired without my having used it.

gahr:
> but a year later it expired without my having used it.

.. nor anyone else, I would hope :D

Issue #4662, opened 2025-09-06
Neomutt saves more than $save_history entries


whitney-cumber:
Issue https://github.com/neomutt/neomutt/issues/3623 is not fixed and still reproducible with the producer given there on neomutt-20250510.

flatcap:
Yeah.  As Issues go, it's low impact, low priority.

(Even the OP took 8 months to respond the the Issue's closure ;-)

The History code isn't too much of a mess.
If you've got the time to take a look, it'd be appreciated.

If not, this Issue might get closed too.
(we have few developers and very limited time)

whitney-cumber:
> Yeah. As Issues go, it's low impact, low priority.
> 
> (Even the OP took 8 months to respond the the Issue's closure ;-)

Guilty as charged :-)

> The History code isn't too much of a mess. If you've got the time to take a look, it'd be appreciated.

I'm sorry but I don't speak C.

> If not, this Issue might get closed too. (we have few developers and very limited time)

My 2c on this:
I don't mind that this issue gets low priority and lingers unfixed for years or decades -- after all it's not particularly critical or annoying and also you, the developers, do it in your spare time.

At this point a **huge thank you to all** who participate here!

However, I find it kind of rude and intransparent to close bugs just because they weren't fixed in some time frame.  Closing a report of a bug doesn't magically make the underlying bug go away.

roccoblues:
@whitney-cumber please try: https://github.com/neomutt/neomutt/pull/4673

whitney-cumber:
I've tested [b9765878](https://github.com/neomutt/neomutt/commit/b976587882add0f602f6ce280341056d412d5005) and it behaves as expected.

Thank you!

PR #4673, opened 2025-09-18
shrink history file on every new entry (fixes #3623, #4662)

### LINKED ISSUE #3623: Neomutt saves more than `$save_history` entries
## Expected Behaviour

Save at most `$save_history` entries in the history file.

## Actual Behaviour

Saves up to `2 * $save_history` entries in the history file.

## Steps to Reproduce

0. Start Neomutt: `neomutt -F /dev/null`
1. Create /tmp/neomutt_history: `touch /tmp/neomutt_history`
2. Follow the file in another terminal window: `tail --follow=name /tmp/neomutt_history`
3. Execute Commands
```
:set history_file=/tmp/neomutt_history
:set save_history=5
```
2. Fill history, e.g. execute commands:
```
echo 1
echo 2
echo 3
echo 4
echo 5
echo 6  # here the file had 6 entries
echo 7  # this shrinks the file to 5 entries: echo 3 - echo 7
echo 8
echo 9
echo 10
echo 11
echo 12 # here the file contains 9 entries
echo 13 # file shrunk again to 5 entries
```

- Always

## NeoMutt Version

```
20220429
```

roccoblues:
The problem was that we only checked every `$save_history` times if the history file needed compacting. That only works for the first batch of entries. But once we reach `$save_history` number of entries we need to compact on every new entry.

I'm not particularly convinced that this is a good fix. Because compacting the history file is a quite expensive operation that we do now on every command. However I also don't have a better idea right yet.

Side-note: the behavior is actually like that since the beginning of the feature: https://github.com/neomutt/neomutt/commit/fe054524f870b89acec9d041bf77138e54e50b90

### DIFF
--- a/history/history.c
+++ b/history/history.c
@@ -317,8 +317,6 @@ cleanup:
  */
 static void save_history(enum HistoryClass hclass, const char *str)
 {
-  static int n = 0;
-
   if (!str || (*str == '\0')) // This shouldn't happen, but it's safer
     return;
 
@@ -342,12 +340,7 @@ static void save_history(enum HistoryClass hclass, const char *str)
   mutt_file_fclose(&fp);
   FREE(&tmp);
 
-  if (--n < 0)
-  {
-    const short c_save_history = cs_subset_number(NeoMutt->sub, "save_history");
-    n = c_save_history;
-    shrink_histfile();
-  }
+  shrink_histfile();
 }
 
 /**


flatcap:
> compacting the history file is a quite expensive operation

We probably don't need to worry about that.
The history is 10 entries for 7 categories (by default).
If the user sets `$history` and `$save_history` too large, it's their own problem :-)

flatcap:
We _could_ make it faster/more efficient by only saving the file on exit.

roccoblues:
> > compacting the history file is a quite expensive operation
> 
> We probably don't need to worry about that. The history is 10 entries for 7 categories (by default). If the user sets `$history` and `$save_history` too large, it's their own problem :-)

We still open the file, read it and write it out again. 

> We could make it faster/more efficient by only saving the file on exit.

Yeah, that would be the correct solution.

flatcap (approved):

Issue #4682, opened 2025-10-18
message hooks do not properly configure pager


jltobler:
## NeoMutt Version

```
NeoMutt 20250905-dirty
Copyright (C) 2015-2025 Richard Russon and friends
NeoMutt comes with ABSOLUTELY NO WARRANTY; for details type 'neomutt -vv'.
NeoMutt is free software, and you are welcome to redistribute it
under certain conditions; type 'neomutt -vv' for details.

System: Linux 6.17.2-arch1-1 (x86_64)
ncurses: ncurses 6.5.20240427 (compiled with 6.5.20240427)
libidn2: 2.3.7 (compiled with 2.3.7)
GPGME: 2.0.0
GnuTLS: 3.8.10
libnotmuch: 5.6.0
storage: kyotocabinet, gdbm, lmdb
compression: lz4, zlib, zstd

Configure options: --prefix=/usr --sysconfdir=/etc --libexecdir=/usr/lib --gpgme --sqlite --autocrypt --lua --notmuch --gss --gnutls --sasl --with-ui=ncurses --with-idn2=/usr --disable-idn --idn2 --lmdb --kyotocabinet --gdbm --ubsan --lz4 --zlib --zstd

Compilation CFLAGS: -march=x86-64 -mtune=generic -O2 -pipe -fno-plt -fexceptions -Wp,-D_FORTIFY_SOURCE=3 -Wformat -Werror=format-security -fstack-clash-protection -fcf-protection -fno-omit-frame-pointer -mno-omit-leaf-frame-pointer -g -ffile-prefix-map=/build/neomutt/src=/usr/src/debug/neomutt -flto=auto -std=c11 -fno-delete-null-pointer-checks -D_ALL_SOURCE=1 -D_GNU_SOURCE=1 -D__EXTENSIONS__ -D_XOPEN_SOURCE_EXTENDED -I/usr/include -DNCURSES_WIDECHAR -I/usr/include/p11-kit-1 -fsanitize=undefined -fno-common -O0

Compile options:
  +autocrypt +fcntl -flock -fmemopen +futimens +getaddrinfo +gnutls +gpgme 
  -gsasl +gss +hcache -homespool +idn +inotify -locales_hack +lua +nls +notmuch 
  -openssl +pgp +regex +sasl +smime +sqlite +truecolor 

Devel options:
  ubsan 

MAILPATH="/var/mail"
PKGDATADIR="/usr/share/neomutt"
[...]

PR #4683, opened 2025-10-19
document magic config value for builtin pager

jltobler:
* **What does this PR do?**

The default builtin pager is used when the pager variable is either empty or set to the magic value "builtin". Update the documentation to mention the magic value.

While discussing the a bug report in https://github.com/neomutt/neomutt/issues/4682, it was suggested that we also update the pager config variable docs to clarify that there is a magic value that can also be used.

### DIFF
--- a/docs/config.c
+++ b/docs/config.c
@@ -2983,9 +2983,9 @@
 /*
 ** .pp
 ** This variable specifies which pager you would like to use to view
-** messages. When empty, NeoMutt will use the built-in pager, otherwise this
-** variable should specify the pathname of the external pager you would
-** like to use.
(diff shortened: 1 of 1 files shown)


flatcap:
Thanks for the PR, @jltobler.

As I noted in #4682

The `builtin` magic word is something we inherited from Mutt.
A few years ago, we started accepting `pager = ""` to mean: use the internal Pager.
A year later, we dropped (nearly :-) all the references from the docs.

It's still silently accepted for backwards compatability. 

I've updated the docs, a1f679160, and pushed them to [the website](https://neomutt.org/guide/configuration.html#message-hook).

PR #4684, opened 2025-10-19
execute message-hook sooner

flatcap:
In the docs, the example `message-hook` sets the value of $pager.

A refactoring 4 years ago, e3a3b9983, meant that we checked the value of $pager *before* the hook fired.

Execute the hook before we use any config.

Fixes: #4682


jltobler:
Thanks @flatcap for raising this change! Invoking the message-hook earlier makes complete sense to me.

I also tested the patch locally and confirmed that message-hooks can now configure the pager variable as expected. This patch looks good to me. :+1:

PR #4687, opened 2025-11-03
fix: `:exec` auto-completion

flatcap:
Auto-completion on the `:exec` command has been broken for quite a while.

**Steps**:
- In the Index, type: `:exec a<Tab><Tab>`

**Expected Result**:
- `alias-dialog`
- `autocrypt-acct-menu`
- ...

This fix isn't brilliant, but it works.

---

The `exec` command takes a NeoMutt function as its argument.
To know which functions are possible, the auto-completion needs to know which Dialog is active.

Historically, this info was obtained by dark magic.

`alldialogs_get_current()` looked for the top-most dialog and tried to determine its type.
This was replaced by function that used the current focus.

Unfortunately, the moment the user hits `:` (`<enter-command>`), the focus moves to the Editor Window.

Extend `struct FileCompletionData` to take a `MuttWindow` that `exec` can use.

### DIFF
--- a/complete/helpers.c
+++ b/complete/helpers.c
@@ -108,7 +108,7 @@ bool candidate(struct CompletionData *cd, char *user, const char *src, char *des
  * @retval 1 Success, a match
  * @retval 0 Error, no match
  */
-int mutt_command_complete(struct CompletionData *cd, struct Buffer *buf, int pos, int numtabs)
+int mutt_command_complete(struct CompletionData *cd, struct Buffer *buf, int pos, int numtabs, void *cdata)
 {
   char *pt = buf->data;
   int spaces; /* keep track of the number of leading spaces on the line */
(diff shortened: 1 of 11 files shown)


flatcap:
This bug highlights a problem with the NeoMutt commands, because they work at a global level.
Commands, such as `:push` or `:exec` need context to work correctly.

Auto-completion for `:exec` needs to know the current Dialog to know which functions are valid.
Then `:exec` does it's own lookup to turn the function into an opcode.
The final insult, is that it can't actually **perform** the action, so it just pushes the opcode into the keyboard buffer.

Ideas / Improvements welcomed :-)

complete/helpers.c
```diff
@@ -232,7 +232,21 @@ int mutt_command_complete(struct CompletionData *cd, struct Buffer *buf, int pos
...
+      if (win && win->wdata)
+      {
+        struct Menu *menu = win->wdata;
+        mtype = menu->type;
+      }
+    }
+    else
+    {
+      mtype = menu_get_current_type();
+    }
```
roccoblues: I doesn't hurt but does this ever make sense?

roccoblues:
> Ideas / Improvements welcomed :-)

I think your summary already outlines the solution. Commands needs to be created hierarchically based on the dialogs and executed within the context of those. For getting all available commands you need to traverse up the hierarchy.

Of course this is just after a two minutes think and probably fails on the first enemy contact. 😆

complete/helpers.c
flatcap: I'm fairly certain it can't get reached, but it won't cause any harm.

PR #4690, opened 2025-11-04
Add virtual-mailboxes to neomuttrc man page

skoved:
* **What does this PR do?**
Adds virtual-mailboxes and unvirtual-mailboxes commands to the the neomuttrc man page along side the mailboxes and named-mailboxes commands that are already present. The descriptions used are the same ones found in the commands section of the [Notmuch Feature](https://neomutt.org/feature/notmuch#6-%C2%A0commands) on the neomutt website


flatcap:
Thanks for the PR.
Sorry for the slow reply.

Actually, it's the other way around.

We want to kill `virtual-mailboxes` from the main docs.
It's out-of-date and we should use `mailboxes -label` or `named-mailboxes` instead.

I've fixed the docs (and a few other things) in be03dcdb4 and pushed the web docs.

Thanks.

PR #4691, opened 2025-11-04
move content from neomutt.org into the repository

roccoblues:
I'd love to make our website look nicer and more useful. However there's so much stuff on it, and a lot of it is outdated, that it's very hard. So my idea would be to move a lot of the content into markdown files in this repository and basically only keep a nice landing page, the guide and the searchable reference.

In general I'd aim for less content that we actively keep up-to-date as opposed to what we currently have on the website.

This is first draft to show the idea. It
* updates the top-level readme
* adds the features documentation
* adds the build instructions
* adds the installation instructions (I didn't move the instructions for the individual distros because I'm pretty sure most of them are outdated and/or not necessary)
* fixes the contribution formatting

Again, this is just a draft, there are probably some broken links and other rought edges. The screenshot is also just a placeholder.

Happy to hear your thoughts.

AI disclaimer: an LLM helped with converting the docs from html to markdown. I haven't read through everything.


roccoblues:
As discussed on IRC: we want to keep the docs in a separate repo. Likely handled by https://about.readthedocs.com. When we moved those out of the main site the design overhaul should be easier.

Closing for now.

PR #4705, opened 2025-11-28
po/eo.po: Update translation

keithbowes:
* **What does this PR do?**
Fixes a couple translations.

n/a

n/a


flatcap (approved):
Lovely, thanks.

I've sent you an invitation to join the NeoMutt org
(we've a very inclusive club :-)

That way, in future, you won't need to create a PR, (unless you have questions)
you could commit directly to the `[translate]` branch.

Cheers!

PR #4707, opened 2025-11-29
Lua: Bigger and Better

flatcap:
This is a **breaking change** to the Lua Scripting.
Fortunately, nobody uses it because it's not even documented.

This is a **work-in-progress**.

It's hacked together, but it works.
Importantly, I understand how all the moving parts fit together.

If you have **questions about the PR**, ask them below.

I've put some docs for the new code in a **new discussion**.
- https://github.com/neomutt/neomutt/discussions/categories/lua

If you have **general questions about Lua** and what we should implement first,
please ask them, there.

---

- b9699b519 pattern: split up mutt_pattern_func()
  Split apart the 'search' and 'limit' code, so that Lua can search

- d6810a74f lua: build
  Move the defines into the build (they'll be needed in every source file)

- fb883fd9b lua: module
  Start to encapsulate the Lua state

- a3510258f lua: debug
  Add a `--debug-lua` build option (as yet unused)

- f8962f575 lua: helpers
  Create some helper functions

- b4630bca2 lua: logging
  Create dedicated Lua logging (see console below)

- e7e10a7c3 lua: global
  Define some global functions for Lua to use

- d6710b304 lua: config
  Define the `config` global object (exposing the Config System to Lua)

- cd5716c9e lua: iterator
  Create an Iterator for the `EmailArray`

- 269955f9b lua: name
  Define the `NeoMutt`, `Account`, `Mailbox` and `Email` objects

- eb81f6cf4 lua: index
  Define the `Index` object (exposing the Index to Lua)

- 03d37fc28 lua: gui
  Define the `gui` global object

- 38066ac58 lua: console
  Create a Lua Console window for live logging

- 5201b786d lua: expando
  Define 3 custom Lua Expandos for `$index_format`


PatrickMassot:
Could you also add functions to access and manipulate attachments? A lot of the things I dream to do with Lua scripting have to do with attachments. For instance I would love to look at the attachment list of a mail, tag a bunch of them, hit a key that:
* save those attachments to some temporary folder.
* opens those attachments in some software and wait until the corresponding files have changed on disk
* create a reply to the mail with the new versions of the files attached

It would be really great when I need to fill attach forms or sign attached documents. By the way, if the above scenario is already possible in neomutt then I would love to learn about it.

flatcap:
> Could you also add functions to access and manipulate attachments?

It'd be lovely to have, but that's an extra level of difficulty.

Attachments don't always exist (1).

For an IMAP connection, we only download the headers until the user tries to read the email.
(If message caching is enabled, then we could have access to any previous read emails).

Attachments don't always exist (2).

Even when we have the full email available, the attachments may still be opaque to us.
They're often encoded for transport.
We'd be able to see their names and types, but to check their contents means decoding them to a file.

The same is true for the body of the email.

All this is solvable, but a fair bit of work, so it's not going to be in v1.0 :-)

PatrickMassot:
I was thinking about manipulating attachments when we currently already see them, either from the pager or from the compose menu before sending an email.

flatcap:
> from the pager or from the compose menu

Ah, I'd only thought as far as the Index.
We'd definitely have all the info we'd need in Pager or Compose.

Scripting Compose would be useful.

PR #4726, opened 2025-12-09
imap: protect against crash on shutdown

flatcap:
If the server sends untagged messages when NeoMutt is shutting down, the `ImapMboxData` might not exist.

KISS: Check every `mdata` pointer.

Fixes: #4632


flatcap:
Slight bit of creep, here.
Now it checks all the `adata` and `mdata` (and the static analysers are happy :-)

PR #4728, opened 2025-12-10
ignore

flatcap:
ignore


ossilator:
regardless of what the idea here was ...

having increments in both the header and body can be nasty, indeed. however, fixing that doesn't imply not using for(), as having an empty update statement is perfectly legit.

flatcap:
I pressed a button on the "Security Page" that offered to suggest a fix to some code issues.
I was interested to see what it'd come up with.

The PR was only generated because it defaults to "PR" and I missed the "Do nothing" selection.

Next, an automatic static code checker kicked in.
Then, copilot tried to fix the fix.

You've replied to the result of several levels of computer-generated nonsense -- marked "ignore" ;-)

ossilator:
yeah, i figured it would be something along these lines, hence the "disclaimer". ;-)

PR #4755, opened 2026-01-04
mutt/array.h: ARRAY_RESERVE(): Fix benign off-by-one bug

alejandro-colomar:
mutt/array.h: ARRAY_RESERVE(): Fix benign off-by-one bug
    
    If (head)->capacity == (num), we have enough storage, and no allocation
    is needed.  While this didn't misbehave, it was unnecessarily
    calling allocation APIs, and was confusing the reader (at least me).
    
Fixes: f4fd43e406ce (2020-08-22; "Implement ARRAY API")
Closes: <https://github.com/neomutt/neomutt/issues/4754>

---

Revisions:

<details>
<summary>v2</summary>

-  Fix parentheses and semicolon.  [@flatcap ]
-  Rename num => n.

```
$ git range-diff --creation-factor=99 HEAD^ neomutt/alx/offby1 alx/offby1 
1:  7ef8ccd1e ! 1:  d81b583c3 mutt/array.h: ARRAY_RESERVE(): Rewrite using GNU statement expression
    @@ Commit message
     
                 0e7c2c38957b (2025-01-13; "Update mutt/queue.h")
     
    +    Also, rename num => n.  It's just simpler, and more idiomatic.
    +
         Link: <https://github.com/neomutt/neomutt/issues/4753>
         Signed-off-by: Alejandro Colomar <alx@kernel.org>
     
      ## mutt/array.h ##
     @@
    -  * @retval num New capacity of the array
    + /**
    +  * ARRAY_RESERVE - Reserve memory for the array
    +  * @param head Pointer to a struct defined using ARRAY_HEAD()
    +- * @param num Number of elements to make room for
    +- * @retval num New capacity of the array
    ++ * @param n    Number of elements to make room for
    ++ * @retval n   New capacity of the array
       */
    - #define ARRAY_RESERVE(head, num)                                               \
    +-#define ARRAY_RESERVE(head, num)                                               \
     -  (((head)->capacity >= (num)) ?                                               \
     -       (head)->capacity :                                                      \
     -       ((mutt_mem_reallocarray(                                                \
    @@ mutt/array.h
     -                ((num) + ARRAY_HEADROOM - (head)->capacity) *                  \
     -                ARRAY_ELEM_SIZE(head))),                                       \
     -        ((head)->capacity = (num) + ARRAY_HEADROOM)))
    -+({                                                                             \
    -+  if ((head)->capacity < (num))                                                \
    -+  {                                                                            \
[...]

PR #4757, opened 2026-01-04
mutt/array.h: Readability improvements

alejandro-colomar:
* mutt/array.h: ARRAY_REMOVE(): Use conditional to avoid complex logic
* mutt/array.h: Use do{}while(0) to implement macros that don't return anything
* mutt/array.h: Rewrite using GNU statement expressions
* mutt/array.h: Rename macro parameter num => n
* mutt/array.h: Remove redunant parentheses
* mutt/array.h: Put macro newline escapes in column 71

This is queued after the bugfix <https://github.com/neomutt/neomutt/pull/4755>.

Since I had a look at that file, I stopped to improve its readability.

### DIFF
--- a/mutt/array.h
+++ b/mutt/array.h
@@ -332,9 +332,12 @@
 #define ARRAY_REMOVE(head, elem)                                               \
   do                                                                           \
   {                                                                            \
-    memmove(elem, (elem) + 1,                                                  \
-            ARRAY_ELEM_SIZE(head) *                                            \
-                MAX(0, (ARRAY_SIZE(head) - ARRAY_IDX(head, elem) - 1)));       \
+    if (ARRAY_SIZE(head) > ARRAY_IDX(head, elem) + 1)                          \
+    {                                                                          \
+      memmove(elem, (elem) + 1,                                                \
+              ARRAY_ELEM_SIZE(head) *                                          \
+              (ARRAY_SIZE(head) - ARRAY_IDX(head, elem) - 1));                 \
+    }                                                                          \
     ARRAY_SHRINK(head, 1);                                                     \
   } while (0)
 


ossilator:
> "Readability" is subjective. I like the old code better.

the familiar always looks nicer at first sight.
but the new code is _objectively_ better, as reading it causes _a lot_ less cognitive load. take it from someone new to this code.

gahr:
> take it from someone new to this code.

I would take it from someone new to both versions of the code, cause of course your version is clearer to you.

> but the new code is objectively better

I'm not going to start a discussion on your definition of objective or why you think your version is objectively better.

In general, I don't like PRs that change the style for the sake of it. This PR fixes *no issue* other than your preferences. Having said that, I don't want to waste anybody's time (including yours an mine) on this, so I won't be in the way if @flatcap wants to merge this.

gahr:
Ok, since I'm here...

> but the new code is objectively better, as reading it causes a lot less cognitive load

of course "cognitive load" is subjective (familiar things require *less* load), so your statement about the new code being *objectively* better is *objectively* wrong.

flatcap:
**Changes** (nothing functional):

- Drop 'Put macro newline escapes in column 71'
  Sorry, that's just weird.
- Fix typo in commit message
- Lightly run clang-format over the results

Thanks!

alejandro-colomar:
> > take it from someone new to this code.
> 
> I would take it from someone new to both versions of the code, cause of course your version is clearer to you.
> 
> > but the new code is objectively better
> 
> I'm not going to start a discussion on your definition of objective or why you think your version is objectively better.
> 

Just to clarify, this PR is mine (@alejandro-colomar ), not from @ossilator .

gahr:
> > > take it from someone new to this code.
> > 
> > 
> > I would take it from someone new to both versions of the code, cause of course your version is clearer to you.
> > > but the new code is objectively better
> > 
> > 
> > I'm not going to start a discussion on your definition of objective or why you think your version is objectively better.
> 
> Just to clarify, this PR is mine (@alejandro-colomar ), not from @ossilator .

Fair point :) so you both prefer the new code. Now you have (part of) it.

PR #4772, opened 2026-01-15
Unify buf_expand_path() and buf_expand_path_regex()

roccoblues:
As NeoMutt's code has matured, some utility functions have been upgraded.
Especially the ones that assemble / manipulate strings.

Rather than using a `char buf[1024]` they use `struct Buffer`.

In some cases a both versions exist, one as the wrapper of the other.
Having two versions leads to confusion of which should be used.

We should settle on using the newer, `struct Buffer`, versions
and upgrade the callers accordingly.

Unify: `buf_expand_path()` and `buf_expand_path_regex()`

1. Change callers of `buf_expand_path()` to use `buf_expand_path_regex()` instead.
2. Remove `buf_expand_path()`
3. Rename `buf_expand_path_regex()` to `expand_path()`


roccoblues:
@copilot please tackle the task in the description.

PR #4790, opened 2026-01-28
Fix: adding a default socket timeout in mutt_oauth2.py to prevent cli…

mmmonk:
Fix: adding a default socket timeout in mutt_oauth2.py to prevent client from hanging forever/freezing 

* **What does this PR do?**

It adds a default timeout to all socket operations in the mutt_oauth2.py client.

     for that)

### DIFF
--- a/contrib/oauth2/mutt_oauth2.py
+++ b/contrib/oauth2/mutt_oauth2.py
@@ -40,6 +40,10 @@ import socket
 import http.server
 import subprocess
 
+# Sometime the communication will hang, freezing Mutt.
+# this timeout prevents this from happening
+socket.setdefaulttimeout(15)
+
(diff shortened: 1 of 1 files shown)


mschilli87:
Could you please add some additional information regarding what this is fixing? What symptoms does the current implementation have and what is changing with this PR? I am using the script and don't think I ever observed a freeze.

mmmonk:
so it actually doesn't. I thought that the neomutt freeze I was experiencing got resolved by this. Typically in the past neomutt froze on me in less then an hour. After bunch of testing I ended up adding that timeout into the script and it was running good for multiple hours ..... but I spoke to soon. It froze again, sorry to waste your time. I should have tested this more. My bad. I am closing this PR. It doesn't resolve anything it seems.

PR #4795, opened 2026-01-30
Bunch of updates to mutt_oauth2.py

cskeeters:
This PR includes many changes to `mutt_oauth2.py`.

* Adds `--tenant` so that users can specify "consumers" for personal Outlook accounts.  Unused for Google.
* Adds `--redirect-uri` so that users can specify the redirect uri that matches the app registration at Microsoft.
* In devicecode authflow, user_code is copied to the clipboard using osc52
* Uses `webbrowser` module to open the authorization endpoint in the default browser
* Improves the legibility and usefulness of pages served by the web server in `localhostauthcode`.
* Adds empty lines to improve legibility
* Imports readline, which enables inputting text longer than 241 characters (my limit with ghostty/macos)
* Adds text documenting how the script works and provides examples of use and common pitfalls.  
* Adds import of `urllib.error` to avoid LSP errors (for devs only)

I'm open to changes, just thought I'd take a shot.

Also, I don't really like that arguments aren't used when the tokenfile exists.  That is unintuitive to me, and I suspect other just trying to get their first access code.  Since it writes the tokenfile before the access code is successfully obtained, this is very likely to bite someone.  I considered changing the code so the token file wouldn't be used at all if `--client-id` was provided, but I wasn't sure that wouldn't mess up someone else's workflow.  Feedback welcome.

### DIFF
--- a/contrib/oauth2/mutt_oauth2.py
+++ b/contrib/oauth2/mutt_oauth2.py
@@ -109,6 +109,8 @@ ap.add_argument('--client-id', type=str, default='',
                 help='Provider id from registration')
 ap.add_argument('--client-secret', type=str, default='',
                 help='(optional) Provider secret from registration')
+ap.add_argument('--redirect-uri', type=str, default='',
+                help='Specify redirect-uri (May need to match match app registration with provider).')
 ap.add_argument('--provider', type=str, choices=registrations.keys(),
                 help='Specify provider to use.')
 ap.add_argument('--tenant', type=str, default='common',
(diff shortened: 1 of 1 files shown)


MaxG87:
This script helped me to regain access to Outlook. Thank you very much. I think it would be beneficial if it wouldn't be buried in some random PR.

I had the impression that the script struggles to accept `--client-secret ''`. I was prompted for a secret anyways. Maybe that could be improved.

cskeeters:
You're welcome.  I updated the script so that the user won't be prompted if `--client-secret ''` is specified.  If you try it and like it, let me know.

PR #4798, opened 2026-02-02
build: make docs on demand

flatcap:
Let the user build the docs when they choose.
This mirrors the change we made to the `--testing` option in afbd882591

- `configure` still checks for the doc build dependencies, but only warns if they are missing

- Deprecate the `--full-doc` and `--disable-doc` options

- Add new `make` targets:
  - `make docs-full` -- build the docs including all optional features
  - `make docs-light` -- build the docs with only the configured features

These targets can now be run on-demand.
You don't need to choose at configure time.

The target names are open to debate.

---

**Workflow**:

- `./configure [OPTIONS]`
- `make`
- coding...
- `make docs-light`
- coding...
- `make clean-docs`
- `make docs-full` 
- `make validate-docs`

### DIFF
--- a/auto.def
+++ b/auto.def
@@ -20,7 +20,7 @@ define PACKAGE_VERSION  "20251211"
 define BUGS_ADDRESS     "neomutt-devel@neomutt.org"
 
 # Subdirectories that contain additional Makefile.autosetup files
-set subdirs {po data docs contrib test smime}
+set subdirs {po data contrib test smime}
 ###############################################################################
 
 ###############################################################################
@@ -30,7 +30,6 @@ set valid_options {
   with-ncurses:path         => "Location of ncurses"
 # Features w/o 3rd party dependencies
   doc=1                     => "Disable building the documentation"
-  full-doc=0                => "Build the full documentation set"
   docdir:path               => "Documentation root"
   with-lock:=fcntl          => "Select fcntl() or flock() to lock files"
   fmemopen=0                => "Use fmemopen() for temporary in-memory files"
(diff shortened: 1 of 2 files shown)


gahr:
> This mirrors the change we made to the --testing option in https://github.com/neomutt/neomutt/commit/afbd882591480b058a3e4f34cb76563b91480308

The big difference is that to build docs we need to bring in dependencies.

From a downstream packager's perspective, I'd rather we kept the choice to be done at configure time.

We (FreeBSD) currently provide users an option to enable / disable docs, and based on that, we decide what the NeoMutt package depends on for building. With this change, we *could* still do that by invoking (or not) the new targets, but we'd get spurious warnings for missing dependencies even if the user doesn't want docs.

What problem is this intended to solve?

flatcap:
Below is some rambling.
But I may have already thought of a better mid-ground.

---

My main problem, is `make` is building more than I need.

99% of the builds I do, I don't need to build the docs.
When I _do_ need them, I don't want to **have to** run `configure` again.

`configure` **commits me** to having: full-docs, light-docs, or none.
I don't see why I need to make that choice.

If I `configure` the docs to build, then `make` builds them every time.
The docs don't check any deps -- probably non-trivial to do so -- it just keeps rebuilding.
Try:
```sh
./configure
make
make
make
```

This will also help NeoMutt users who build NeoMutt to help us debug.
The most common thing we have to tell them is 'try again with `--disable-doc`'.

flatcap:
Abandoned in favour of:
- 6d9c9de9f docs: make full-docs the default
- f39124c10 build: better docs deps checking
- e092528f9 build: Don't unnecessarily rebuild docs and po

ossilator:
fwiw, you could require both: configure with support for docs, and have a separate target for it.

PR #4799, opened 2026-02-02
IMAP: Add timeouts and Log much more

flatcap:
**Overview**:

- Ensure connections timeout if the link fails
- Log lots more, so if anything goes wrong, we know what was happening

**I've tested**:

- Interrupting the connection for a short while
- Killing the connection
- Hitting <kbd>Ctrl-C</kbd> during a download
- Putting my machine to sleep / waking it

NeoMutt re-connects more reliably now.

---

:heart: If you use IMAP, please give this a test and **let me know** how you get on.


flatcap:
@codex review

PR #4807, opened 2026-02-09
mbox_check: fix has_new flag inconsistencies across backends

roccoblues:
The `has_new` flag is critical for user notifications and sidebar display. When set, it triggers:
- User notification messages ("New mail in...")
- Sidebar highlighting with color_sidebar_new
- Sidebar 'N' indicator
- Mailbox count in mutt_mailbox_check()

Several backends were returning `MX_STATUS_NEW_MAIL` but failing to set the `has_new` flag, causing silent notification failures for background mailbox checking.

Backends fixed:
- IMAP: Set `has_new` when `IMAP_NEWMAIL_PENDING` is set
- POP: Set `has_new` when new messages are detected
- NNTP: Set `has_new` when new articles arrive
- MH: Set `has_new` in `mh_check()` when incorporating new messages

This ensures consistent behavior across all mailbox types for both currently-open mailboxes and background checking of other mailboxes.

Created with @copilot.

### DIFF
--- a/imap/imap.c
+++ b/imap/imap.c
@@ -1264,7 +1264,10 @@ enum MxStatus imap_check_mailbox(struct Mailbox *m, bool force)
   if (mdata->check_status & IMAP_EXPUNGE_PENDING)
     check = MX_STATUS_REOPENED;
   else if (mdata->check_status & IMAP_NEWMAIL_PENDING)
+  {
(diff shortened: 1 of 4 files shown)


roccoblues:
closing as I'm not convinced this is correct

PR #4808, opened 2026-02-09
docs: fix a word

laur89:
Actually `command action` might be a thing. Unsure.


flatcap:
Good spot, thanks!
I've piggy-backed a bunch of other doc fixes into this PR,
(but you're still credited :-)

flatcap (approved):

PR #4813, opened 2026-02-15
Key Refactoring -- State Machine

flatcap:
This PR is another large step forward in the handling of key bindings.
<sub>More than a PoC, less than finished ;-)</sub>

---

## `a`/`ab` Binding Hell

This PR removes the restriction where one binding was a prefix of another.
e.g.
```sh
bind index a  next-entry
bind index ab previous-entry
```

These bindings are now **both** valid and will be resolved using a timeout.

## Timeouts

There are two timeouts in the key code (currently hard-coded).
The first is for no input.
The second is when there is ambiguity.

Using the config above...

- User types `a` and waits
  The timeout is reached, so `<next-entry>` is returned

- User types `a` then `b` (within the timeout)
  `<previous-entry>` is returned

## Counter

Currently the digit keys are bound to `<jump>`.
If you hit one, then a separate function sits waiting for you to type more digits and hit <kbd>\<enter\></kbd>.
This means they can't be used for anything else.

This PR treats numbers as a prefix to a function.
(see State Machine, below)

**This will allow the user to type `5j` and move the cursor down five lines.**

The function dispatchers were all upgraded in anticipation of this, in 1c28f2ad6d.
This is still a lot of work to upgrade each function, but that can be done in stages.

## Notifications

When typing a large prefix number, or a long keybinding, having some feedback would be helpful.

There is a new notification event that's sent from `km_dokey()`.
When the use types `123`, it can be displayed on the status bar.

## Testing

This PR has two functions enabled in the Index:
(upper-case letters)

- `J` - `<next-entry>`
- `K` - `<previous-entry>`

Open a large mailbox and type `5J` or `10K`!

## State Machine

The State Machine behind `km_dokey()`
**Source:** [key-state.gv](https://github.com/neomutt/gfx/blob/main/arch/key-state.gv)


flatcap:
@codex review

PR #4825, opened 2026-03-14
feature: add `$tmp_draft_dir` for composing messages

flatcap:
Add a new config option `$tmp_draft_dir` that controls where NeoMutt places temporary files when composing messages.
This defaults to `/var/tmp` so that draft files survive an unanticipated reboot, unlike `/tmp` which is typically cleared on restart.

The existing `$tmp_dir` continues to be used for transient temporary files
(displaying messages, crypto operations, protocol handling, etc.)
that do not need to survive a reboot.

Draft temp files are created for:
- Composing new messages and replies
- Editing with `$edit_headers`
- Forwarding/replying from the attachment menu
- Postponed message recovery
- Edit/resend message
- Command-line `-H draft` and `-i include` files

If `$TMPDIR` is set in the environment, both `$tmp_dir` and `$tmp_draft_dir` use it as the initial value.

### DIFF
--- a/attach/cid.c
+++ b/attach/cid.c
@@ -203,9 +203,10 @@ void cid_to_filename(struct Buffer *filename, const struct CidMapList *cid_map_l
    * HTML file may be opened as plain text by the viewer */
   const char *suffix = buf_rfind(filename, ".");
   if (suffix && *(suffix++))
-    buf_mktemp_pfx_sfx(tempfile, "neomutt", suffix);
+    buf_mktemp_full(tempfile, "neomutt", suffix, __FILE__, __LINE__, "tmp_dir");
   else
     buf_mktemp(tempfile);
+
   fp_out = mutt_file_fopen(buf_string(tempfile), "w+");
   if (!fp_out)
     goto bail;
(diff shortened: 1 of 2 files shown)


flatcap:
Good call.
I've merged the two pairs of functions using the config name. 
As a bonus, I've killed a couple intermediate macros.

gahr:
I don't know. Do we really need two temp dir configs? How does NeoMutt recover from a reboot in the middle of a compose? If it needs the user to manually go find the right temp file and try to resume composing from it, then I think the chances are high that those temporaries will be left over indefinitely.

flatcap:
> Do we really need two temp dir configs?

Need?  No.
It's a new upstream feature, so I'd _like_ to keep it for compatibility. 

> that those temporaries will be left over indefinitely

Yeah, that's system-dependent.

FYI: systemd's default rules clear `/var/tmp` files older than 30 days.

gahr:
> > Do we really need two temp dir configs?
> 
> Need? No. It's a new upstream feature, so I'd _like_ to keep it for compatibility.

Ah, that's where it's coming from! Fair enough.

roccoblues (approved):

PR #4830, opened 2026-04-01
store: deprecate old backends

flatcap:
Debian has raised an issue to drop tokyocabinet as a header cache backend -- it's very old and unmaintained.
- https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=1132419

**Update 2:**

- Don't pick a default header cache backend
- Force the user to set `$header_cache_backend` before `$header_cache`
- Change the list order of the backends to emphasise lmdb

Drop:
- tokyocabinet
- kyotocabinet
- qdbm
- bdb

Additionally: fe923b83a puts the name of store in the cache filename,
e.g. `imap/neo/devel.lmdb.hcache`

---

[ChatGPT's opinion](https://gist.github.com/flatcap/b632f2885a4271e62b1ced086dc32d6b) of the eight backends is:
(I haven't checked the findings, but I don't have reason to doubt them)

| Backend         | Maintained       | Recommended   | Notes                 |
| :-------------- | :--------------- | :------------ | :-------------------- |
| LMDB            | ✅ Yes           | ⭐⭐⭐⭐⭐    | Best choice           |
| RocksDB         | ✅ Yes           | ⭐⭐⭐⭐      | Powerful but overkill |
| GDBM            | ✅ Yes           | ⭐⭐⭐        | Safe fallback         |
| TDB             | ✅ Yes           | ⭐⭐          | Niche                 |
| Tokyo Cabinet   | ❌ No            | ❌            | Dead                  |
| Kyoto Cabinet   | ⚠️ Weak          | ❌            | Stagnant              |
| QDBM            | ❌ No            | ❌            | Dead                  |
| Berkeley DB     | ⚠️ Complicated   | ❌            | Licensing issues      |

### DIFF
--- a/hcache/hcache.c
+++ b/hcache/hcache.c
@@ -379,7 +379,7 @@ static void hcache_per_folder(struct HeaderCache *hc, struct Buffer *hcpath,
   struct Buffer *hcfile = buf_pool_get();
   if (namer)
   {
-    namer(hc->folder, hcfile);
+    namer(hc->store_ops, hc->folder, hcfile);
     mutt_encode_path(hcfile, buf_string(hcfile));
     buf_concat_path(hcpath, path, buf_string(hcfile));
   }
--- a/hcache/lib.h
+++ b/hcache/lib.h
@@ -108,10 +108,15 @@ struct HCacheEntry
  *
  * Prototype for function to compose hcache file names
  *
- * @param path    Path of message
- * @param dest    Buffer for filename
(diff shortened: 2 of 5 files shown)


gahr:
1. we make `headaer_cache_backend` mandatory whenever `header_cache` is set
1.1 users will start to get warnings if they haven't picked a backend explicitly
2. we remove the old backends altogether
2.1 users with unsupported backends will get an error

PR #4831, opened 2026-04-02
config: warn user of valid values on invalid input

flatcap:
When a **user sets** a config option to an **invalid value** for types where the set of valid values is known,
**show a helpful warning** instead of a terse error.

Add a `CSR_INV_WARNING` flag to signal that an invalid-value message includes guidance.

e.g. the user sees:
> Invalid value for header_cache_backend
> Choose from: lmdb, gdbm


gahr (approved):
Very nice!

PR #4867, opened 2026-05-10
alias: hide mail prompt

flatcap:
The Message Window is actually a Container with a stack of Windows.
Only the top-most Window is visible.

The Alias Dialog can be triggered by auto-completion of the `<mail>` `To:` prompt.
When this happens, the `To:` prompt is still visible and it obscures any warnings or errors.

Temporarily hide (pop) the `To:` prompt from the stack, then restore it afterwards.

---

**Steps**:
- `<mail>` (<kbd>m</kbd>)
- Either:
   - `<complete>` (<kbd>\<Tab\></kbd> by default) -\> Alias Dialog
   - `<complete-query>` (<kbd>Ctrl-T</kbd> by default) -\> Query Dialog
- `<backspace>` (or any unbound key)

**Results**:
- Error message visible

### DIFF
--- a/alias/dlg_alias.c
+++ b/alias/dlg_alias.c
@@ -445,7 +445,15 @@ int alias_complete(struct Buffer *buf, struct ConfigSubset *sub)
     mutt_pattern_alias_func(NULL, &mdata, PAA_VISIBLE, NULL);
   }
 
-  if (!dlg_alias(&mdata))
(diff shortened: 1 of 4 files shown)


roccoblues (approved):
Nice! Works fine for me. 🙏🏼

gahr (approved):

PR #4873, opened 2026-05-17
Make the Message Window more visible

flatcap:
Update the MessageWindow to control its own visibility.

Most of the commits are tidying up.
The last commit, 4aebd7e23f, does the work.

https://github.com/user-attachments/assets/b628a7b5-1df5-487f-9f8e-1208ff967c74

The Message Container is a stack of MuttWindow.
The default window displays error messages.

When this space is needed, e.g. for user input, another window is pushed into the stack and it takes the focus.
If an error occurs when another window is focused, the message may not be seen.  

msgwin

---

**Scenario**:

- User prompt: "Enter filename:"
- User enters a bad filename
- `mutt_error("Bad filename: %s", filename);`
- MessageWindow realises that is isn't current visible, so:
  - make MessageWindow visible
  - reflow all the windows

### DIFF
--- a/alias/dlg_alias.c
+++ b/alias/dlg_alias.c
@@ -445,15 +445,7 @@ int alias_complete(struct Buffer *buf, struct ConfigSubset *sub)
     mutt_pattern_alias_func(NULL, &mdata, PAA_VISIBLE, NULL);
   }
 
-  struct MuttWindow *hidden_prompt = alias_dialog_hide_prompt();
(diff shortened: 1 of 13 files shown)


charles2910:
wow, +1 one on this one. I actually didn't realize it could fail "silently". Here is the current behaviour for comparison:

https://github.com/user-attachments/assets/65ae8d64-51be-45ca-b4f5-b35a93c01db4

PR #4874, opened 2026-05-18
main: restore -C flag

funsafemath:
Restore [the '-C' flag](https://neomutt.org/feature/cli-crypto), which was accidentally removed in [699657c61b15](https://github.com/neomutt/neomutt/commit/699657c61b15621994b39d45596755f527d43d24)

### DIFF
--- a/main.c
+++ b/main.c
@@ -1188,6 +1188,9 @@ int main(int argc, char *argv[], char *envp[])
     if (OptGui)
       mutt_flushinp();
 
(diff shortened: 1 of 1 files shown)


flatcap (approved):

flatcap:
Damn, that was careless.

Thanks for the fix, @funsafemath :heart:
btw, if you'd like your real name in the Release Notes, let me know what it is :-)

PR #4878, opened 2026-05-28
Update range in description of `--debug-level` and `$debug_level`

rayfordshire:
* **What does this PR do?**

Update range in description of `--debug-level` and `$debug_level`.
At some point Neomutt learned the log level `LL_NOTIFY` (with a value of 6) but did not update the documentation of `--debug-level` and `$debug_level` accordingly.

Note: This changes the documentation, so the homepage might need a rebuild.

* **You (the maintainer) need to make a choice**

To prevent at least parts of the documentation to go out of sync with the implementation, we can – instead of hardcoding the log level values – use `printf()` to interpolate the actual log level values (`LL_MESSAGE` and `LL_MAX - 1`) in the output of `--help`.  This is done in the last two commits.  Whether that's worth it or not, is your decision to make.

(Obviously this does not prevent the other documentations from bitrotting as they still hardcode the log level values.)

### DIFF
--- a/docs/config.c
+++ b/docs/config.c
@@ -1110,7 +1110,7 @@
 ** The debug level controls how much information is saved to the log file. If
 ** you have a problem with NeoMutt, then enabling logging may help find the
 ** cause. Levels 1-3 will usually provide enough information for writing a bug
-** report. Levels 4,5 will be extremely verbose.
+** report. Levels 4,5,6 will be extremely verbose.
 ** .pp
 ** Warning: Logging at high levels may save private information to the file.
 ** .pp
--- a/docs/manual.xml.head
+++ b/docs/manual.xml.head
@@ -20082,7 +20082,7 @@ neomutt -a image.jpg *.png -- address1 address2
               <entry>
                 Log debugging output to a file (default is
                 &quot;<literal>~/.neomuttdebug0</literal>&quot;).
-                The <literal>level</literal> can range from 1&ndash;5 and
+                The <literal>level</literal> can range from 0&ndash;6 and
                 affects verbosity (a value of 2 is recommended). Using this
                 option along with <emphasis role="bold">-l</emphasis> is useful
                 to log the early startup process (before reading any
--- a/po/bg.po
+++ b/po/bg.po
@@ -9512,8 +9512,8 @@ msgstr "        -m, --mbox-type <type>       Set default mailbox type"
 
 #: usage.c:75
 #, fuzzy
-msgid "        -d, --debug-level <level>    Set logging level (1..5)"
-msgstr "        -d, --debug-level <level>    Set logging level (1..5)"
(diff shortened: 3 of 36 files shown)


flatcap:
Hello stranger :-)

> At some point Neomutt learned the log level `LL_NOTIFY`

897d6eb28 debug: add notify level
That was back in 2019, wow.

> but did not update the documentation

Ah, that was by choice.

We inherited 5 debug levels from Mutt, but their usage is a bit of a mess.
These levels usually offer some information that might help diagnose a problem.

When I created the notification system, I wanted a unique level for the messages, so that I could easily filter for them.
There are typically a LOT of notifications, and while logging them is useful to **me**, I can't see many occasions when it'd be useful to anyone else.

---

> **You (the maintainer) need to make a choice**

OK :-)

Thank you for raising the issue of levels.
I really appreciate the effort -- not many would have bothered updating the translation files.

> documentations from bitrotting

Possible, but I think the levels are unlikely to change again.

I really don't like turning down contributions, but I think in this case, I'm going to say, **Thanks**, but no thanks.
I hope that's OK.

Thanks!

rayfordshire:
> There are typically a LOT of notifications, and while logging them is useful to **me**, I can't see many occasions when it'd be useful to anyone else.

**me** should be **us** ;-) As, obviously, I also used it otherwise I wouldn't have found that inconsistency.

> I really don't like turning down contributions, but I think in this case, I'm going to say, **Thanks**, but no thanks.
> I hope that's OK.

Don't worry, that's fine by me.

---

I might point out, however, that `neomutt -h` says "1..5" while `neomutt -h all` states "0 (off), 1 (low) .. 5 (high)". If you like, I can prepare a pull request updating the documentation and help message to the range 0..5.

flatcap:
> **me** should be **us** ;-)

Apologies...

... logging them is useful to **us**, ...

Life gets in the way and developers disappear without a word, sometimes I feel a little isolated.

> I might point out, however, that `neomutt -h` says "1..5" while `neomutt -h all` states "0 (off), 1 (low) .. 5 (high)".
> If you like, I can prepare a pull request updating the documentation and help message to the range 0..5.

Yes, please.
Keep it simple, just hard-code the values into the text.

If we (see what I did there :-) ever change the limits again, I'll personally fix up every version of documentation.

Thanks

rayfordshire:
> > I might point out, however, that `neomutt -h` says "1..5" while `neomutt -h all` states "0 (off), 1 (low) .. 5 (high)".
> > If you like, I can prepare a pull request updating the documentation and help message to the range 0..5.
> 
> Yes, please. Keep it simple, just hard-code the values into the text.

Done in https://github.com/neomutt/neomutt/pull/4881.

> If we (see what I did there :-) ever change the limits again, I'll personally fix up every version of documentation.

:-)

---

Closing this pull request in favour of https://github.com/neomutt/neomutt/pull/4881.

PR #4881, opened 2026-05-29
Update log level description to cover the value 0

rayfordshire:
* **What does this PR do?**

Successor of https://github.com/neomutt/neomutt/pull/4878.  Updates the documentation to mention that 0 is also a valid value for the log level, which disables logging.


flatcap (approved):
Lovely, thanks!

PR #4885, opened 2026-06-01
mutt/buffer.c: don't over-allocate if initialized

mokomull:
parse_address() uses buf_new() to copy addresses into a struct Buffer, but only addr_set_local() and addr_set_intl() would ever modify that buffer.  Rounding every allocation up to the next multiple of 128 bytes wastes a lot of memory.

Loading a mailbox of approximately 1.8 million emails consumed 16.2GB of memory and allocated 7,417,113,344 bytes for Address.mailbox buffers (the sum of mailbox->dsize), of which only 643,359,347 bytes were used (the sum of buf_len(mailbox)).

This commit brings that down to 7.3GB, with only 672,332,445 bytes allocated for those same Address.mailbox buffers.

Fixes: a3b368548f23 ("stuct Address: replace char * uses with struct Buffer")

### DIFF
--- a/mutt/buffer.c
+++ b/mutt/buffer.c
@@ -312,7 +312,10 @@ struct Buffer *buf_new(const char *str)
     buf_addstr_n(buf, str, slen);
   }
   else
+  {
     buf_alloc(buf, 1);
+  }
+
   return buf;
 }
 
@@ -363,8 +366,7 @@ void buf_alloc(struct Buffer *buf, size_t new_size)
 }
 
 /**
- * buf_alloc_exact - Make sure a buffer can store at least new_size bytes,
- * without rounding up.
+ * buf_alloc_exact - Make sure a buffer can store at least new_size bytes, without rounding up
  * @param buf      Buffer to change
  * @param new_size New size
  *
--- a/mutt/buffer.c
+++ b/mutt/buffer.c
@@ -375,6 +375,9 @@ void buf_alloc_exact(struct Buffer *buf, size_t new_size)
   if (!buf)
     return;
 
+  if (new_size == 0)
+    new_size = 1;
+
   const bool was_empty = (buf->dptr == NULL);
   const size_t offset = (buf->dptr && buf->data) ? (buf->dptr - buf->data) : 0;
 
--- a/mutt/buffer.c
+++ b/mutt/buffer.c
@@ -306,7 +306,11 @@ struct Buffer *buf_new(const char *str)
   struct Buffer *buf = MUTT_MEM_CALLOC(1, struct Buffer);
 
   if (str)
-    buf_addstr(buf, str);
+  {
+    const size_t slen = mutt_str_len(str);
+    buf_alloc_exact(buf, slen + 1);
+    buf_addstr_n(buf, str, slen);
+  }
   else
     buf_alloc(buf, 1);
   return buf;
@@ -355,10 +359,34 @@ void buf_alloc(struct Buffer *buf, size_t new_size)
     // LCOV_EXCL_STOP
   }
 
+  buf_alloc_exact(buf, ROUND_UP(new_size + 1, BufferStepSize));
+}
+
+/**
+ * buf_alloc_exact - Make sure a buffer can store at least new_size bytes,
+ * without rounding up.
+ * @param buf      Buffer to change
+ * @param new_size New size
+ *
+ * @sa buf_alloc
+ */
+void buf_alloc_exact(struct Buffer *buf, size_t new_size)
+{
+  if (!buf)
+    return;
+
   const bool was_empty = (buf->dptr == NULL);
   const size_t offset = (buf->dptr && buf->data) ? (buf->dptr - buf->data) : 0;
 
-  buf->dsize = ROUND_UP(new_size + 1, BufferStepSize);
+  if (buf->data && (new_size <= buf->dsize))
+  {
+    // Extra sanity-checking
+    if (!buf->dptr || (buf->dptr < buf->data) || (buf->dptr > (buf->data + buf->dsize)))
+      buf->dptr = buf->data; // LCOV_EXCL_LINE
+    return;
+  }
+
+  buf->dsize = new_size;
 
   MUTT_MEM_REALLOC(&buf->data, buf->dsize, char);
   buf->dptr = buf->data + offset;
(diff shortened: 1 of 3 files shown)


flatcap:
@mokomull Thanks! :heart:
That's a hell of an improvement.

We've done a lot of refactoring to use `struct Buffer`, over the years, to reduce the number of memory reallocations.
But, as you've spotted, it can be wasteful.

For a **long** time I've wanted to create a `:memory` command that quantifies "how much" and "where" -- some day :-)

flatcap (approved):
LGTM (but I just want to get the opinion of another dev)

gahr:
I guess it's a trade-off between allocating less memory and allocating less often. We currently round-up, which allows us not to reallocate often as you append. This means more wasted memory but less fragmentation.

This PR chooses to allocate less memory but fragment more. I don't see a good argument why we'd prefer more fragmentation over more memory usage.

I am also not convinced that  a 1.8 million emails mailbox is the use case we should be optimizing for.

mokomull:
Yeah, the 1.8-million-emails is certainly a "me-problem" :)

This patch is indeed a pessimism at `buf_new()`-time, but for any given Buffer, it only applies once.  The next time `buf_alloc()` is called for a Buffer (e.g. via `buf_inline_replace` or `buf_addstr{,_n}`), then the size will still get rounded-up to a multiple of 128 bytes as always.

When I get back to churning through this mailbox this evening, I'll add some tracing locally to buf_alloc (similar to how I measured the allocated/used counts in the PR description) so I can get firm numbers on how often that actually happens.

mokomull:
As a heuristic, it seems that it is indeed pretty rare to reallocate a Buffer whose contents were known at `buf_new()` time (i.e. by passing a non-null `char *`).

I added
```diff
diff --git a/mutt/buffer.c b/mutt/buffer.c
index 44638f3ea..f2cc3fed3 100644
--- a/mutt/buffer.c
+++ b/mutt/buffer.c
@@ -41,6 +41,7 @@
 #include <stdint.h>
 #include <stdio.h>
 #include <string.h>
+#include <sys/sdt.h>
 #include "buffer.h"
 #include "exit.h"
 #include "logging2.h"
@@ -359,6 +360,10 @@ void buf_alloc(struct Buffer *buf, size_t new_size)
     // LCOV_EXCL_STOP
   }
 
+  if (buf->dsize % BufferStepSize != 0) {
+    DTRACE_PROBE2(neomutt, expand_short_buffer, buf->dsize, new_size);
+  }
+
   buf_alloc_exact(buf, ROUND_UP(new_size + 1, BufferStepSize));
 }
 
```

to `buf_alloc()` and hammered it for a couple hours of my workflow (which is mostly save-message and delete-pattern and limit, no composing or replying), and reallocating a buffer that is affected by this PR happened in exactly two places, only once per mailbox open each (the extra stack is because I switched mailboxes and came back during the experiment just to make sure the probe was still set in the right place):
```
@stacks[
        buf_alloc+63
        buf_addstr_n+107
        buf_addstr+50
        buf_strcpy+49
        rfc2047_decode_addrlist+99
        rfc2047_decode_envelope+32
        mutt_rfc822_read_header+305
        imap_read_headers+2074
        imap_mbox_open+736
        mx_mbox_open+796
        main+13893
        __libc_start_call_main+117
        __libc_start_main@@GLIBC_2.34+136
        _start+37
]: 1
@stacks[
        buf_alloc+63
        buf_addstr_n+107
        buf_addstr+50
        buf_strcpy+49
        rfc2047_decode_addrlist+99
        rfc2047_decode_envelope+32
        mutt_rfc822_read_header+305
        imap_read_headers+2074
        imap_mbox_open+736
        mx_mbox_open+796
        change_folder_mailbox+550
        change_folder_string+146
        op_main_change_folder+518
        index_function_dispatcher+384
        dlg_index+1010
        main+14100
        __libc_start_call_main+117
        __libc_start_main@@GLIBC_2.34+136
        _start+37
]: 1
@stacks[
        buf_alloc+63
        buf_inline_replace+161
        mx_path_canon+144
        mx_mbox_find2+86
        op_main_change_folder+346
        index_function_dispatcher+384
        dlg_index+1010
        main+14100
        __libc_start_call_main+117
        __libc_start_main@@GLIBC_2.34+136
        _start+37
]: 2
```

The rfc2047_decode_addrlist is the only one that worried me, because perhaps it is a mere coincidence that I have only one message in this mailbox that *expands* the encoding of an Address during character set conversion.  In this case, looks like it's a name in Cyrillic in a `From:` header that was sent in KOI8-R, so it makes sense that it expands out to multiple bytes per character in UTF-8.

But even this expansion happens *once* — rfc2047_decode_addrlist calls buf_strcpy which reallocates the Buffer with its final length.  The buffers that are used during decoding come from buf_pool_get(), which creates buffers with `buf_new(NULL)`.

For the sake of completeness, the tiny bpftrace script that generated the above:
```let @stacks = hash(5000);

usdt:/home/mmullins/projects/neomutt/neomutt:neomutt:expand_short_buffer {
        @stacks[ustack] += 1;
}

interval:1m {
        print(strftime("%H:%M:%S.%f", nsecs));
        print(@stacks);
        print("");
}
```

gahr:
I and @flatcap were chatting on IRC (you're welcome to join on Libera/#neomutt) and we said half-joking that if you don't reallocate, perhaps you're not a buffer..
I also suggested that perhaps we need a separate API for things we know for almost-sure we are not going to reallocate.

How about this instead: we pick the round-up number based on the string length, e.g.,
* len < 32 -> pick 32
* len < 64 -> pick 64
* else do what we do now

That would help some with small edits but would avoid huge wastes of memory.

flatcap:
Interesting...
I had a play with heaptrack (for the first time, yesterday).
It had a graph of allocation sizes (for a 40,000 email mailbox)

image

**Key:**
- orange: `mutt_mem_calloc()`
- yellow: `mutt_mem_realloc_array()`

gahr:
> It had a graph of allocation sizes (for a 40,000 email mailbox)

What's the third axis (stacked bars)?

flatcap:
Key added

gahr:
@mokomull , would https://github.com/neomutt/neomutt/pull/4888 work for you?

mokomull:
I will give it a try today!  That's actually pretty similar to my first pass, which was just to blindly change BufferStepSize to 16 and got not-ideal results but within a gigabyte or so of this PR.

mokomull:
> if you don't reallocate, perhaps you're not a buffer..

heh, that's also similar to something I'd tried — replaced the mailbox and personal fields of Address with `ReadOnlyBuffer` that very much did not compile, but the compiler errors pointed me at the very small handful of places where the buffers are actually mutated.

flatcap:
Thanks!
Superseded by #4888

PR #4892, opened 2026-06-04
expando: Return `Expando` struct when parsing an empty string

### LINKED ISSUE #4639: Empty `index-format-hook` replacement string triggers config error
After updating, Neomutt failed to parse my old config.  I boiled it down to this neomuttrc config:

```
set index_format = "%@date@"
index-format-hook  date  "~d<1d"  ""
```
(the `set index_format` is not necessary but probably there when using `index-format-hook`)

With this config `neomutt -F neomuttrc` fails with a very uninformative:

```
.../neomutt.rc:2: 
source: errors in .../neomutt.rc
```

Uninformative because only a line is given but not *what* whet wrong or *what* is expected, a better error would have been "failed parsing replacement format string" or so.

The problem seems to be that the replacement string is empty, replacing `""` with `"foo"` works as expected.

## NeoMutt Version

```
20250113
```

## Expected Behaviour

No Error AND ADDITIONALLY a better error message instead of just a blank "error"

## Actual Behaviour

Config parsing error.

- Always

After updating to neomutt-20250113, probably related to the refactor of format strings.

rayfordshire:
@flatcap I have no clue whether changing the return value for the empty string breaks some (implicit) assumptions in other parts of the code.  @gahr at the time was also not familiar with it (https://github.com/neomutt/neomutt/issues/4639#issuecomment-3023295845). So we need some code-hero with expertise to verify this commit doesn't have unwanted side effects.

---

Let `expando_parse` return an `Expando` struct for the empty string "" instead of NULL.

Code paths using `expando_parse` often treat the return value of NULL as an error condition.  As a result parsing the empty string "" via the expando mechanism results in an error.  For example when the user configures the following index-format-hook:

  index-format-hook date "~d<1d" ""

History:

`expando_parse` returning NULL for an empty string was introduced in commit 58c167e6067b4b28e59c6b80789a28f2b01face3 "expando: use ARRAYs".

Fixes: #4639

---

Totally Offtopic:

@flatcap I wonder why you choose to treat the the empty string "" specially by representing it as `NULL` increasing the code complexity and introducing additional code paths.  The config system does the same:

https://github.com/neomutt/neomutt/blob/e756936d099a6cfdb9c3ebb5d08c200dc132f3e2/config/string.c#L63-L65

What is/was your motivation? Is it all about saving 1 byte?

### DIFF
--- a/expando/expando.c
+++ b/expando/expando.c
@@ -73,15 +73,16 @@ void expando_free(struct Expando **ptr)
 
 /**
  * expando_parse - Parse an Expando string
- * @param str  String to parse
- * @param defs Data defining Expando
+ * @param str  String to parse; must be non-NULL
+ * @param defs Data defining Expando; must be non-NULL
  * @param err  Buffer for error messages
  * @retval ptr New Expando
+ * @retval NULL Error
  */
 struct Expando *expando_parse(const char *str, const struct ExpandoDefinition *defs,
                               struct Buffer *err)
 {
-  if (!str || (*str == '\0') || !defs)
+  if (!str || !defs)
     return NULL;
 
   struct Expando *exp = expando_new(str);
--- a/test/command/parse_index_hook.c
+++ b/test/command/parse_index_hook.c
@@ -37,6 +37,7 @@ static const struct Command IndexFormatHook = { "index-format-hook",
 static const struct CommandTest Tests[] = {
   // clang-format off
   { MUTT_CMD_WARNING, "" },
+  { MUTT_CMD_SUCCESS, "date '~d<1d' ''" },
   { MUTT_CMD_SUCCESS, "date '~d<1d' '%[%H:%M]'" },
   { MUTT_CMD_SUCCESS, "date '~d<1m' '%[%a %d]'" },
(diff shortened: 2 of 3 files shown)


flatcap:
Superseded by #4893 

Please can you give branch [[devel/index-format-hook]](https://github.com/neomutt/neomutt/tree/devel/index-format-hook) a try.

Thanks

flatcap:
> I wonder why you choose to treat the the empty string "" specially by representing it as `NULL`

I didn't.
That was a design decision LONG ago in Mutt.

> Is it all about saving 1 byte?

It saves more than a byte, because the OS is unlikely to work with such small allocations,
but it also saves the cost of an allocation.
These are both these costs a _very_ small though.

> The config system does the same:

_That_ I did choose.

There are ~500 config options -- many of which start empty.
They're shown to the user as strings, but many are structured data behind the scenes.
This means allocating all those empty options on startup (and cleaning them up later).

It seems cleaner and quicker to store them as `NULL`.
If we need to render them to the user, then `NULL` becomes `""`.

PR #4893, opened 2026-06-04
fix index-format-hook parsing

### LINKED ISSUE #4639: Empty `index-format-hook` replacement string triggers config error

flatcap:
If the index-format-string format string is empty, there's no Expando.
Check for an error message instead.

Fixes: #4639

**Tests**:

**empty.rc** -- No error, empty string when hook matches
```
set index_format = "%@date@"
index-format-hook  date  "~d<1d"  ""
```

**bad.rc** -- Error message about `%w`
```
set index_format = "%@date@"
index-format-hook  date  "~d<1d"  "%w"
```

### DIFF
--- a/hooks/parse.c
+++ b/hooks/parse.c
@@ -1110,7 +1110,7 @@ enum CommandResult parse_index_hook(const struct Command *cmd, struct Buffer *li
   parse_extract_token(fmt, line, TOKEN_NONE);
 
   exp = expando_parse(buf_string(fmt), IndexFormatDef, err);
-  if (!exp)
+  if (!buf_is_empty(err))
     goto out;
 
   if (MoreArgs(line))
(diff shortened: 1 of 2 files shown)


rayfordshire:
Thanks. Tested and works as expected.

I've took the liberty and ported the `index-format-hook` test of https://github.com/neomutt/neomutt/pull/4892 to this branch.

flatcap:
Lovely, thanks!

PR #4897, opened 2026-06-06
notmuch: allow <entire-thread> from any mailbox type

### LINKED ISSUE #105: notmuch entire-thread outside a virtual folder
When using mutt in a regular folder, I would often like to locate all messages in the thread of a given message that are known to notmuch. However, hitting '+' to call <entire-thread> fails, displaying "No virtual folder, aborting."

I think it would be more helpful if this function (or some other function), when called outside of a virtual folder, retrieved the Message-ID of the current messages (or of all tagged messages), asked notmuch for all messages in their threads, and generated a moved to a new vfolder with this view. I don't think I can script this behavior, because of the need of extracting the Message-Id (barring awful hacks like generating a command in a file that I would then source...).

Do you think it would make sense to add the feature that I describe? Thanks!

flatcap:
Previously <entire-thread> only worked in notmuch virtual folders, or from Maildir/MH mailboxes (via an on-the-fly notmuch query).
Drop the restriction to Maildir/MH so any mailbox type (IMAP, POP, MBOX, NNTP, compressed, ...)
can also use <entire-thread>, as long as the selected message has a Message-ID and notmuch has indexed it.

Fixes: #105 (maybe)

Someone posted a workaround to Issue #105, but it vanished almost immediately.

I still don't use notmuch, myself, but I had a quick hack.
I'm not _really_ sure how `<entire-thread>` is _meant_ to work, but this may help.

### DIFF
--- a/index/functions.c
+++ b/index/functions.c
@@ -3686,8 +3686,7 @@ static int op_main_entire_thread(struct IndexFunctionData *fdata, const struct K
   struct IndexPrivateData *priv = fdata->priv;
   if (shared->mailbox->type != MUTT_NOTMUCH)
   {
-    if (((shared->mailbox->type != MUTT_MH) && (shared->mailbox->type != MUTT_MAILDIR)) ||
-        (!shared->email || !shared->email->env || !shared->email->env->message_id))
+    if (!shared->email || !shared->email->env || !shared->email->env->message_id)
     {
       mutt_message(_("No virtual folder and no Message-ID, aborting"));
       return FR_ERROR;


yourealwaysbe:
>Someone posted a workaround to Issue #105, but it vanished almost immediately.

That was me. I realised my workaround was overcomplicated compared to the "horrible hack" suggested by the OP (which i ended up using).

I'm trying this PR now. It's working OK so far. Thanks.

For more context: in my case i have a (read-only) copy of my mail held locally, indexed by notmuch. But i'm currently pointing mutt directly at the remote servers (to avoid sync issues). With this setup, entire-thread gets the message ID from the remote message, then queries over the local store for the thread containing the message ID, which works fine.

flatcap:
Great!  Thanks for the feedback.  Sorry it took forever.

PR #4900, opened 2026-06-10
bcache: sanitize mailbox names that escape the cache dir

jmestwa-coder:
* **What does this PR do?**

  the body cache builds its on-disk path from the server-supplied mailbox
  name, so a name containing `..` components escapes `$message_cache_dir`.

  - replace each `..` path component of the mailbox name with `__` before
    building the cache path, so it stays inside the cache directory and
    caching keeps working for such names
  - components that merely contain dots (`a..b`, `..foo`) are untouched
  - covers both the imap and nntp body caches (the group name is reused as
    the path)

  none

### DIFF
--- a/bcache/bcache.c
+++ b/bcache/bcache.c
@@ -87,7 +87,23 @@ static int bcache_path(struct ConnAccount *account, const char *mailbox, struct 
   }
   buf_fix_dptr(host);
 
-  buf_addstr(host, mailbox);
+  /* A server-supplied mailbox name with '..' components would let the cache
+   * path escape $message_cache_dir.  Replace them with '__' so the path stays
+   * inside the cache directory.  imap_hcache_open() rejects these names for
+   * the header cache. */
+  for (const char *p = mailbox; p && *p;)
+  {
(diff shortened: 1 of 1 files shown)


gahr:
How about replacing those chars instead of rejecting the path altogether?

jmestwa-coder:
makes sense, switched to replacing instead of rejecting:

- each `..` path component of the mailbox name becomes `__`, so the path stays under `$message_cache_dir` and caching keeps working for those names
- names that merely contain dots (`a..b`, `..foo`) are left alone

two server names differing only in `..` vs `__` would now share a cache dir, but for a cache that seemed harmless. updated the commit and PR text to match.

gahr (approved):
LGTM

flatcap:
Thanks, @jmestwa-coder.
If you'd like your real name in the Credits/Release Notes, let me know what it is :-)

jmestwa-coder:
@flatcap, Kindly list my name as **Javeed Khan** in the Credits/Release Notes.

PR #4904, opened 2026-06-19
hcache: bound deserialization reads against the cache entry length

jmestwa-coder:
* **What does this PR do?**

Bounds-checks the header cache deserialiser, which walks a fetched entry without ever knowing its length.

- `restore_email()` and the `serial_restore_*` helpers advance an offset and `memcpy()` fixed fields plus a length-prefixed string (up to `SERIAL_MAX_CHAR_SIZE`) with no check that the read stays inside the entry; only the 8-byte uidvalidity/crc header is range-checked.
- A truncated or corrupt entry then reads past the end of the heap allocation, the worst case being the attacker-sized `size` copied in `serial_restore_char()`.
- Threads the fetched length through the restore helpers and rejects any read that would leave the blob via a new `serial_in_bounds()` helper.
- `decompress()` now also reports the inflated length, so the same bound covers lz4/zlib/zstd entries.

   - Doxygen comments updated for the new parameter
   - Code follows the style guide
   - Builds and the `compress` unit tests pass

None.

### DIFF
--- a/compress/lib.h
+++ b/compress/lib.h
@@ -100,16 +100,17 @@ struct ComprOps
    * @ingroup compress_api
    *
    * decompress - Decompress header cache data
-   * @param[in] handle Compression handle
-   * @param[in] cbuf   Data to be decompressed
-   * @param[in] clen   Length of the compressed input data
+   * @param[in]  handle Compression handle
+   * @param[in]  cbuf   Data to be decompressed
+   * @param[in]  clen   Length of the compressed input data
+   * @param[out] dlen   Length of the decompressed output data
    * @retval ptr  Success, pointer to decompressed data
    * @retval NULL Otherwise
    *
    * @note This function returns a pointer to data, which will be freed by the
    *       close() function.
    */
-  void *(*decompress)(ComprHandle *handle, const char *cbuf, size_t clen);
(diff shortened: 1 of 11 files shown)


gahr:
Thanks @jmestwa-coder , I like the approach of making sure we don't deserialize garbage. However, I'd rather not deserialize default values either... deserializing a `0` for any ints is bound to create as many issues as deserializing garbage...

What about failing to deserialize altogether? All the `serial_restore_*` functions could be enhanced to return a boolean and on `false`, we'd stop trying to deserialize.

jmestwa-coder:
makes sense, agreed that a deserialized 0 is no better. switched to failing instead of defaulting:

- every serial_restore_* now returns a bool; a read that would leave the blob (or a list count over the cap) returns false
- the first false stops the restore rather than zero-filling, so a short or crafted entry is rejected as a whole instead of producing an email built from defaults
- restore_email frees the partial email and returns NULL, which the callers already treat as a cache miss and fall back to parsing from the source

build is green and the compress tests still pass.

gahr (approved):
I haven't reviewed every single line, but LGTM.

flatcap (approved):
Very nice, thanks!

PR #4908, opened 2026-06-25
OSS-Fuzz: Add new fuzzer targets body processing

arthurscchan:
This PR adds a new OSS-Fuzz fuzzer targets body processing, together with the dict and options addition for the new fuzzer in https://github.com/neomutt/corpus-address/pull/1 and the fixes of the build script from OSS-Fuzz (https://github.com/google/oss-fuzz/pull/15802).


flatcap:
Fantastic!  Thanks, @arthurscchan 
Now I need to remember how all the pieces of the  build fit together :-)

---

I set up the first fuzzer many years ago.
I always _meant_ to add more bits to it, but other priorities kept getting in the way.

flatcap (approved):

arthurscchan:
You are welcome. Do let me know if you want some comments on how to build it on OSS-Fuzz side to run them.

flatcap:
Thanks.
Remarkably, my old notes still worked :-)

PR #4911, opened 2026-07-01
auth_cram: fix out-of-bounds read of cram-md5 challenge

jmestwa-coder:
imap_auth_cram_md5() decodes the server cram-md5 challenge with the full obuf->dsize as the output limit:
- mutt_b64_decode() returns raw bytes and never terminates, so a challenge that decodes to obuf->dsize (1024) fills the buffer with no NUL
- adata->buf grows unbounded, so a hostile or MITM server controls that length
- the buffer is then read as a C string by buf_string()/hmac_md5() -> strlen(), over-reading past the allocation

Decode into obuf->dsize - 1 and terminate at the returned length, like pop/auth.c already does.

### DIFF
--- a/imap/auth_cram.c
+++ b/imap/auth_cram.c
@@ -131,11 +131,14 @@ enum ImapAuthRes imap_auth_cram_md5(struct ImapAccountData *adata, const char *m
     goto bail;
   }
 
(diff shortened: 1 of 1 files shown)


flatcap (approved):
Lovely, thanks!

PR #4912, opened 2026-07-02
date: bound tm_mon in mutt_date_make_time

jmestwa-coder:
mutt_date_make_time() range-checks every struct tm field except tm_mon:
- a negative tm_mon indexes the static AccumDaysPerMonth[] out of bounds (e.g. tm_mon == -1)
- a value >= 12 slips through the % 12 and returns a bogus time_t instead of the TIME_T_MIN every other out-of-range field gets

Reject out-of-range tm_mon like the sibling fields. Added the two cases to the existing test.

### DIFF
--- a/mutt/date.c
+++ b/mutt/date.c
@@ -257,6 +257,8 @@ time_t mutt_date_make_time(struct tm *t, bool local)
 
   if ((t->tm_mday < 1) || (t->tm_mday > 31))
     return TIME_T_MIN;
+  if ((t->tm_mon < 0) || (t->tm_mon > 11))
+    return TIME_T_MIN;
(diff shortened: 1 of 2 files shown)


flatcap (approved):
Thanks!

Hmm... I wonder why I didn't think to check `tm_mon` when I put in the other range checks.

flatcap:
P.S.
Your first commits didn't have a real name on them, so I asked you what it was.
These last two commits _do_ have a name on them and it's different!
Are you a spy?  A programming collective?  Suffering nominative confusion? :-)

jmestwa-coder:
ha, no conspiracy. jmestwa-coder is just my github handle, and Syed Mohammed Nayyar is the actual name behind it. the first commits went out before i'd set user.name locally, so git fell back to the handle. sorted the config now so they line up. thanks for the review.

PR #4917, opened 2026-07-03
parse: fix out-of-bounds read in parse_references

jmestwa-coder:
* **What does this PR do?**

Out-of-bounds read parsing `References:` / `In-Reply-To:` headers:

- `mutt_extract_message_id()` rfc2047-decodes a copy of the header and reports the message-id offset within that decoded copy
- `parse_references()` advances the raw header by that offset (`s += off`)
- an encoded-word in a charset that expands to UTF-8 (e.g. `=?iso-8859-1?B?...?=`) makes the decoded text longer than the raw header, so the offset overruns the buffer and the next read walks off the end (`In-Reply-To` uses a tight `strdup`'d value, so ASAN flags it directly)

`parse_references()` now decodes the header once and iterates over that single copy, so the scan and the advance use the same string. The `<...>` scan moves into a small `extract_message_id()` helper; `mutt_extract_message_id()` is unchanged for its other (NULL-len) callers.

Regression added to `test/email/mutt_rfc822_read_header.c`: it feeds an encoded `References` with two ids (both now recovered) and an expanding `In-Reply-To`; the latter fails under ASAN before the change.

None.

### DIFF
--- a/email/parse.c
+++ b/email/parse.c
@@ -280,6 +280,43 @@ static void parse_content_disposition(const char *s, struct Body *b)
 }
 
 /**
+ * extract_message_id - Find a message-id in a string
+ * @param[in]  s   String to parse
+ * @param[out] len Number of bytes of s parsed (optional)
+ * @retval ptr  Message id found
+ * @retval NULL No more message ids
+ *
+ * @a len is an offset into @a s, so the caller may advance @a s by it to find
+ * the next message-id.  The string is scanned verbatim; decoding, if any, is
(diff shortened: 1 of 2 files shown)


flatcap:
Sorry it's taken a while, I've been tied up with mega-discussion #4902

I've made a few minor tweaks to comments and rebased.

flatcap (approved):

jmestwa-coder:
no worries on the wait, thanks for merging this.

PR #4918, opened 2026-07-04
fix(imap): detect and abort on write failure in imap_read_literal

acts-1631:
imap_read_literal() writes each byte of a declared IMAP literal to a
temp file via fputc(), but never checked for a write failure.  A
malicious or misbehaving server can declare a very large literal; if
the temp filesystem fills up or a quota is hit partway through,
fputc() starts failing silently and the loop spins through the
remaining declared byte count doing nothing useful.

Check ferror(fp) periodically (same cadence as the existing progress
bar update) and once more after the loop, and treat a write failure
the same way a socket read failure is already treated: mark the
connection IMAP_FATAL and return -1.

### DIFF
--- a/imap/imap.c
+++ b/imap/imap.c
@@ -787,7 +787,23 @@ int imap_read_literal(FILE *fp, struct ImapAccountData *adata,
     fputc(c, fp);
 
     if ((pos % 1024) == 0)
+    {
       progress_update(progress, pos, -1);
+
+      /* Check for a write failure (eg disk or quota full) periodically.
+       * A server can declare an arbitrarily large literal; without this
+       * check a full disk causes fputc() to fail silently and this loop
+       * spins through the rest of the declared byte count doing nothing
+       * useful. */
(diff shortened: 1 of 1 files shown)


flatcap:
@acts-1631
I welcome fixes, including those assisted or generatedby AI.
However, **anonymous** AI contributions are not acceptable.

If a human is willing to take responsibility for these changes,
then I'm much more likely to review them and, if they're good, merge them.

And yes, I've been following your upstream Mutt patches with interest.

acts-1631:
Hi. Upstream closed the merge requests, so I wanted to give neomutt a chance to accept them if you believe they're worth fixing. I've compile tested all the fixes and (though I'm far from an expert) they seem correct to me. There are more security-ish bugs pending validation, but if you prefer, I won't send any more. Thanks for at least making your objections understandable.

flatcap:
I want a human to take ownership of these pull-requests.

PR #4919, opened 2026-07-04
lib: grow mutt_file_read_line() buffer geometrically

acts-1631:
mutt_file_read_line() is the central line-reading function used
throughout neomutt (MIME headers, IMAP/POP responses, mailcap, config
files, PGP keyrings).  For a line with no newline, it grows its buffer
by a fixed 256-byte increment per iteration.  Since each growth
reallocs and copies everything accumulated so far, this makes the
total copy cost O(n^2) in the length of the line.

Switch to doubling the buffer on each growth instead of adding a
fixed increment.


flatcap:

PR #4920, opened 2026-07-04
imap: grow imap_cmd_step() response buffer geometrically

acts-1631:
imap_cmd_step() grows adata->buf by a fixed IMAP_CMD_BUFSIZE (512
byte) increment per iteration while reading a response line.  Each
growth reallocs and copies everything read so far, so a fixed
increment makes the total copying cost O(n^2) in the length of the
line.

Switch to doubling the buffer once past the initial allocation,
amortizing the copy cost to O(n).


flatcap:

PR #4921, opened 2026-07-04
pop: grow pop_fetch_data() accumulation buffer geometrically

acts-1631:
pop_fetch_data() accumulates a long logical line (one that spans
multiple socket reads without a CRLF) into inbuf, but reallocs inbuf
to fit exactly on every single iteration via
MUTT_MEM_REALLOC(&inbuf, lenbuf + sizeof(buf), char).  Each such
realloc copies everything accumulated so far, so a long line's total
copying cost is O(n^2) in the line length.

Track inbuf's actual capacity separately and only grow it (by
doubling) when the next chunk won't fit, amortizing the copy cost to
O(n).  To avoid holding a large allocation for the rest of the fetch
after one long line, shrink back to the base size once a full line
has been handed off to callback(), mirroring the existing shrink-back
behavior in imap_cmd_step() for idata->buf.

### DIFF
--- a/pop/lib.c
+++ b/pop/lib.c
@@ -514,13 +514,15 @@ int pop_fetch_data(struct PopAccountData *adata, const char *query,
   char buf[1024] = { 0 };
   long pos = 0;
   size_t lenbuf = 0;
+  size_t inbuflen;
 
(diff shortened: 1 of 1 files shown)


flatcap:

PR #4930, opened 2026-07-10
Add mouse support

### LINKED ISSUE #309: Mouse support under terminals?
Is it possible to use mouse when running `mutt` under terminals, in my case under `urxvt`?

I'd be really helpful to point and click on the items in the inbox to select and open them, because most of the time in the window manager your hand holds mouse, so it's easier to interact with the command line applications as well.

I mean in the style of other similar CLI programs such as `cmus`, `htop`, and `ranger`.

Freed-Wu:
* **What does this PR do?**

1. add mouse_row, mouse_col to KeyEvent
2. handle mouse event in mutt_getch_timeout()
3. add mouse support for sidebar/index
4. add tests

     for that)

Fix #309

### DIFF
--- a/gui/module.c
+++ b/gui/module.c
@@ -101,6 +101,9 @@ static bool gui_gui_init(struct NeoMutt *n)
   struct GuiModuleData *mod_data = neomutt_get_module_data(n, MODULE_ID_GUI);
 
   /* check whether terminal status is supported (must follow curses init) */
+  mousemask(BUTTON1_CLICKED | BUTTON1_DOUBLE_CLICKED | BUTTON4_PRESSED |
+                BUTTON5_PRESSED,
+            NULL);
   mod_data->ts_supported = mutt_ts_capability();
   mutt_resize_screen();
   log_gui();
--- a/gui/module.c
+++ b/gui/module.c
@@ -101,6 +101,10 @@ static bool gui_gui_init(struct NeoMutt *n)
   struct GuiModuleData *mod_data = neomutt_get_module_data(n, MODULE_ID_GUI);
 
   /* check whether terminal status is supported (must follow curses init) */
+  mousemask(BUTTON1_CLICKED | BUTTON1_DOUBLE_CLICKED | BUTTON2_CLICKED |
+                BUTTON2_DOUBLE_CLICKED | BUTTON3_CLICKED |
+                BUTTON3_DOUBLE_CLICKED | BUTTON4_PRESSED | BUTTON5_PRESSED,
+            NULL);
   mod_data->ts_supported = mutt_ts_capability();
   mutt_resize_screen();
   log_gui();
--- a/gui/opcodes.h
+++ b/gui/opcodes.h
@@ -29,11 +29,11 @@
 #include "mutt/lib.h"
 
 const char *opcodes_get_description(int op);
-const char *opcodes_get_name       (int op);
+const char *opcodes_get_name(int op);
 
-#define OP_REPAINT     (-3) ///< Repaint is needed
-#define OP_TIMEOUT     (-2) ///< 1 second with no events
-#define OP_ABORT       (-1) ///< $abort_key pressed (Ctrl-G)
+#define OP_REPAINT (-3) ///< Repaint is needed
+#define OP_TIMEOUT (-2) ///< 1 second with no events
+#define OP_ABORT (-1)   ///< $abort_key pressed (Ctrl-G)
 
 // clang-format off
 #define OPS_ATTACH(_fmt) \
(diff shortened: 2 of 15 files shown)


flatcap:
Ooh! Exciting!
Thanks for diving in :heart:

I'll try to remind myself of all my mouse ideas over the weekend.
The event handling has changed quite a lot (for the better/simpler) since the original issue.

Let's finally get mouse 🐭 support merged.

flatcap:
I've just pushed a commit that should fix the out-of-tree test build.

Freed-Wu:
I build and run it, the mouse can work for me. CI/CD still failure which need to take a look.

flatcap:
The two new 'test' dirs needed an entry in `BUILD_DIRS` in `test/Makefile.autosetup`

Freed-Wu:
Great. CI/CD has been fixed.

flatcap:
OK, I've had a quick look -- good news, bad news :-)

Good news: It works!  (of course)

Bad news: The implementation is bad from an architecture point-of-view.
It adds lots of dependencies to the key-handling code.

## History - Mutt

Quick history lesson.

In Mutt, in the Index Dialog, the code calls `km_dokey(MENU_MAIN)` to get a keystroke from the user.

`km_dokey()` looks up the `MENU_MAIN` in a global table to get a list of valid key bindings.
Then it has lots of hard-coded rules for particular menus, `MENU_PAGER`, `MENU_EDITOR`, `MENU_GENERIC`.

This is \*\*\*\*.

## NeoMutt

It took me a lot of effort, but in NeoMutt, the hard-coded rules are gone.
(well the ones described above, at least :-)

In NeoMutt, each **Dialog** defines the keys it takes and any fallbacks required, e.g. `generic`.
NeoMutt's Index Dialog calls `km_dokey()` with a `struct MenuDefinition`.

The `MenuDefinition` encapsulates all the functions and all the keybindings.
The low-level key code doesn't need to know anything about the caller (Dialog).

`km_dokey()` returns a `struct KeyEvent`.
It contains the raw key pressed and an opcode, e.g. `OP_PAGE_DOWN` looked up in the `MenuDefinition`.

## Mouse Support

Ideally, `km_dokey()` should turn mouse events into some standard opcodes, e.g. `OP_MOUSE_LEFT_CLICK`.

The Dialog would then decide for itself how to behave.
Initially, these behaviours could be hard-coded: left-click == select.

We'd want a full set of mouse button numbers and click, double-click, scroll?

**This is where I'd like this PR to go.**

## Future Plans

I'm not expecting you to code this as well (but it would be great if you did :-)

The next stage would be parameterise these actions.

We'd define new key-bindings, e.g. `<Left-Click>` that would be available for certain menus.
This would allow the user to, e.g.
```
bind index <LeftClick> display-message
```

## What Now?

I really want this feature working -- lots of people request it.

This PR was generated by Copilot, so (no offense) I don't know how good your C skills are.
Even if you have lots of AI credits, it will really help if you understand the "what" and the "why" of code.

Please ask lots of questions, either here, or on IRC: `#neomutt` on `irc.libera.chat` ([web client](https://kiwiirc.com/nextclient/irc.libera.chat/#neomutt))

Cheers, FlatCap

flatcap:
Ping?

I sent you an email on Wednesday, did you get it?

Freed-Wu:
Please wait for a while, as there are quite a few things to handle at the moment.

flatcap:
Absolutely, there's no rush.
I was just worried that I'd scared you off :-)

flatcap:
Thanks, that looks a lot better.
I'll have a deep dive when I get a second.

---

**Notes:**
I've just force-pushed to your branch -- no functional changes

I undid some whitespace changes in the header files.
We tend to format the headers manually, as whitespace often improves readability.

flatcap:
Looking good!

The next step is to break up the 'Mouse event handling' in `dlg_index()`.

Each Dialog is divided into Windows.
In the Index Dialog, those are:
- Index (email list)
- Index Bar (status bar)
- Sidebar

The full picture [looks like this](https://github.com/neomutt/gfx/blob/main/screenshots/window/dlg-index-pager-sidebar.svg).

The Event loop, in the Index Dialog, calls `km_dokey()` to get an event.
Whether the event is a keypress or a mouse click, doesn't matter.

Each **interactive** Window has a function dispatcher that does the actual work.
We want to let the function dispatchers decide what to do.

---

I'll start with the Sidebar, because it's a simpler case.
We can change `sb_function_dispatcher()`:

- Is this a **mouse** event?
- If so, is it within the Sidebar Window
  (move the Window intersection test into the Window code)
- If so, perform a hard-coded (for now) translation:
  - `MOUSE_CLICK` == `OP_SELECT_ENTRY`
  - etc
- Pass the modifed event to the Sidebar functions

I _think_ that...
The modified Event should contain the OpCode
and the coordinates of the Mouse Event **relative** to the Sidebar Window. 

Perhaps we can have a helper function in the Sidebar.
Something like: `sb_get_mailbox(Event)`
If the Event is a mouse event, then use its coordinates, if not, then use the current highlight.
I hope that makes sense.
If not, pester me with questions :-)

---

The Index is slightly more complicated.
It has two function dispatchers to think about.

`index_function_dispatcher()` handles Index specific functions, like opening the Email in the pager.
`menu_function_dispatcher()` handles motion / scrolling.

Each will have the same: Mouse Event translation.
Is it our Event?  Yes/No?
Is it something we can handle? Yes/No?

---

By moving the Mouse handling out to the function dispatchers,
we allow Events to be delegated.

Index gets first go.
If it can't handle the Event, then the Menu gets to try.
Failing that the Sidebar.

Freed-Wu:
@flatcap

Issue #4943, opened 2026-07-18
Resource leak in mutt_decode_save_attachment() on repeated decode of an unparsed attachment (found while investigating Coverity CID 557729)


chrisdebian:
As part of reviewing NeoMutt's Coverity Scan results, I went through all 40 currently outstanding findings. 14 turned out to be false positives after full code tracing, which I've marked accordingly in Coverity (Classification: False Positive, Action: Ignore) with a reasoning comment on each, including CID 557729, which had originally been filed as a use-after-free. That specific claim doesn't hold up (full reasoning is in the Coverity comment on the CID), but tracing it turned up a different, genuine bug, which is what this issue covers.

`mutt_decode_save_attachment` (attach/mutt_attach.c) is called from the compose menu to decode-and-save an attachment that hasn't been parsed yet (`fp == NULL`). If it's called twice on the same still-unparsed multipart or message/external-body attachment, for example saving the same attachment twice without leaving compose, the second call's freshly re-parsed Body tree gets silently discarded: the function's restore logic (`b->parts = saved_parts;`) overwrites the pointer holding it without freeing it first. Unlike the message/rfc822 case, there's no owning Email object whose `email_free()` would recursively clean it up.

I confirmed this isn't just a theoretical trace. I built an out-of-tree ASan configuration and wrote a regression test that calls `mutt_decode_save_attachment()` twice on the same synthetic unparsed multipart attachment. Run under LeakSanitizer, it reports 247 bytes leaked across 5 allocations, with every stack trace terminating at the second call's `mutt_parse_part()` invocation, exactly the code path described above.

I have the regression test ready locally and a clear idea of the fix (free the previous re-parsed tree, or restructure to avoid the double-parse). I'd like to submit a PR with both.

AI disclosure: this issue was drafted with Claude Code's assistance (code investigation, ASan reproduction, and Coverity triage), reviewed and sent by me.

chrisdebian:
PR opened: #4944

Issue #4939, opened 2026-07-18
Coverity CID 535655: LOFF_T/long narrowing in text_enriched_handler


chrisdebian:
As part of reviewing NeoMutt's Coverity Scan results, I went through all 40 currently outstanding findings. 14 turned out to be false positives after full code tracing, which I've marked accordingly in Coverity (Classification: False Positive, Action: Ignore) with a reasoning comment on each. This issue covers one of the ones that's a genuine, if low-priority, bug.

CID 535655, in text_enriched_handler (email/enriched.c): `long bytes = b_email->length` narrows LOFF_T (off_t) into long. On every platform NeoMutt actually targets (Linux, macOS, BSD) both types are 64-bit, so this is harmless in practice, but it's not guaranteed by the C standard and would matter on a hypothetical platform with a 32-bit long. `bytes` is only ever used as a read-countdown in this function, never a buffer size or array index, so the worst case even there is a very large enriched-text attachment rendering truncated, not memory corruption.

Proposed fix: change the local variable's type from `long` to `LOFF_T` to match the field it's copied from. One line, no behavioural change on any currently-supported platform.

I'd like to submit a PR for this.

AI disclosure: this issue was drafted with Claude Code's assistance (code investigation and Coverity triage), reviewed and sent by me.

chrisdebian:
PR opened: #4947

PR #4945, opened 2026-07-19
fix: fall back to hil_index -1 when sb_prev() finds nothing on removal

chrisdebian:
* **What does this PR do?**

`sb_remove_mailbox()` keeps the sidebar highlight in place when possible after a mailbox is removed. When the highlighted entry was the last one in the array and got deleted, it calls `sb_prev()` to backtrack to the previous unhidden entry, but never checked whether that succeeded.

If every remaining entry is hidden, `sb_prev()` returns false and `hil_index` is left at its old value — now one past the end of the shrunk array, since the entry it used to point to no longer exists. The sibling `is_hidden` branch a few lines below already falls back to `hil_index = -1` in its own equivalent failure case; this branch didn't.

Not memory-unsafe on its own: every `hil_index` reader goes through `ARRAY_GET()`, which bounds-checks and returns `NULL` out of range. But it leaves the sidebar in a state the many `hil_index < 0` "nothing highlighted" checks elsewhere in the file won't detect.

Fixed by checking `sb_prev()`'s return value and falling back to `-1`, matching the sibling branch.

Found via Coverity CID 561613 (Unchecked return value). Per your note on #4939-#4943 that simple fixes can skip the Issues step, this went straight to a PR.

  - No new config option, so no `manual.xml.head` change needed
  - Builds clean under ASan, no compiler warnings on the changed file
  - New regression test (`test/sidebar/sb_remove_mailbox.c`) confirmed to fail without the fix (`hil_index` stuck at the stale out-of-bounds value) and pass with it
  - Full unit test suite passes (all tests, `make test`)

* **AI disclosure**

This PR was written primarily by Claude Code, based on the Coverity CID 561613 triage. I've reviewed the diff and reasoning and can explain it.

### DIFF
--- a/sidebar/sidebar.c
+++ b/sidebar/sidebar.c
@@ -163,7 +163,8 @@ void sb_remove_mailbox(struct SidebarWindowData *wdata, const struct Mailbox *m)
       if (!sbep_cur)
       {
         // The last entry was deleted, so backtrack
-        sb_prev(wdata);
+        if (!sb_prev(wdata))
+          wdata->hil_index = -1;
       }
       else if ((*sbep_cur)->is_hidden)
       {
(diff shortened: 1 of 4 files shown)


flatcap:
Very minor formatting fix

flatcap (approved):

PR #4953, opened 2026-07-24
docs: add Repology packaging status badge to README

chrisdebian:
No packaging status indicator existed in the README badge row, unlike the website (distro.md), which already has one.

Confirmed before opening: the live README (fetched fresh via the GitHub API) has no existing mention of Repology or packaging status, and no prior open or closed PR/issue on this repo ever proposed one. Also confirmed the `neomutt` slug on Repology resolves correctly.

Used the `tiny-repos` badge variant rather than `vertical-allrepos` (the one used on the website's distro.md) since this README's badge row is a single line of small shields.io-style badges -- the tall vertical variant would break that layout.


flatcap (approved):
Nice!  Thanks

PR #4955, opened 2026-07-24
ci: add Dependabot config for GitHub Actions

chrisdebian:
No automated dependency-update tooling exists for the github-actions ecosystem -- confirmed via `find .github -iname dependabot*`, nothing found. Every action reference (`actions/checkout@v7`, `codeql-action@v4`, etc.) is a static, hand-picked version that never gets an update PR.

Weekly schedule matches the existing CI cadence (most scheduled workflows already run Mondays). `github-actions` ecosystem only, since that's the specific gap found -- not proposing anything for the C toolchain itself.

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -1,0 +1,7 @@
+version: 2
+updates:
+  - package-ecosystem: "github-actions"
+    directory: "/"
+    schedule:
+      interval: "weekly"
+      day: "monday"


flatcap (approved):
Yeah, that's a bit smarter than my current scheme -- being subscribed to all the relevant repos.

flatcap (approved):
Yeah, that's a bit smarted than my current scheme -- being subscribed to all the relevant repos.

flatcap:
And so it begins! #4958 :-)

PR #4959, opened 2026-07-24
conn: fix two size_t/int truncation issues found while scoping #4940

### LINKED ISSUE #4940: Coverity CID 392753 / 392751: size_t truncation in Connection::write() return value
As part of reviewing NeoMutt's Coverity Scan results, I went through all 40 currently outstanding findings. 14 turned out to be false positives after full code tracing, which I've marked accordingly in Coverity (Classification: False Positive, Action: Ignore) with a reasoning comment on each. This issue covers two related ones that are a genuine, if low-priority, bug.

CID 392753 (conn/tunnel.c, `tunnel_socket_write`) and CID 392751 (conn/raw.c, `raw_socket_write`) both have the identical pattern: `return sent;` where `sent` is a `size_t` accumulator, narrowed into the function's `int` return type. That return type is dictated by the shared `Connection::write()` interface used by every connection backend, so this isn't specific to either file, it's a systemic interface design issue. A single `write()` call would need to request over 2GB before this actually mattered, and email protocol writes never come close in practice, so the real-world risk is very low.

Given it's an interface-wide issue rather than a per-file one, fixing it properly means widening `Connection::write()`'s return type (or the size_t/int boundary generally) across all implementing backends, not just patching these two. I'd like to scope that properly before proposing a change, rather than a narrow single-file patch that leaves the same pattern elsewhere.

I'd like to submit a PR for this once I've confirmed the full set of affected backends.

AI disclosure: this issue was drafted with Claude Code's assistance (code investigation and Coverity triage), reviewed and sent by me.

chrisdebian:
Two small, independent fixes flagged in the #4940 investigation (comment
from 2026-07-21) — separate from the larger `Connection::write()` interface
question, which that comment concluded isn't worth pursuing (the sole
caller already bounds `count` to `INT_MAX`).

- **`gnutls.c` (`tls_socket_write`)**: same size_t-into-int narrowing
  pattern as Coverity CID 392753/392751 (`raw.c`/`tunnel.c`), though not
  itself Coverity-flagged. Adds an explicit `(int)` cast with a comment
  documenting why the narrowing is safe.
- **`zstrm.c` (`zstrm_write`)**: a distinct, non-CID bug — the old
  `(rc <= 0) ? 1 : rc` clamp silently reported a false "1 byte written" if
  `count` ever overflowed `INT_MAX`, instead of surfacing the failure.
  Replaced with an explicit overflow check that returns -1.

Both build-verified clean (`make conn/gnutls.o conn/zstrm.o`, `-Wall`,
zero warnings).

AI disclosure: implemented with Claude Code's assistance, following up on
the investigation already posted to #4940.

### DIFF
--- a/conn/gnutls.c
+++ b/conn/gnutls.c
@@ -1127,7 +1127,14 @@ static int tls_socket_write(struct Connection *conn, const char *buf, size_t cou
     sent += rc;
   } while (sent < count);
 
-  return sent;
+  // sent is a size_t accumulator narrowed into this function's int return
+  // type, mandated by the shared Connection::write() interface -- the same
+  // pattern flagged by Coverity CID 392753/392751 in raw.c/tunnel.c, though
+  // not currently flagged here. The sole caller of conn->write(),
+  // mutt_socket_write_d(), already takes an int len, so count is bounded to
+  // INT_MAX before it ever reaches here -- see #4940 for the full
+  // call-graph trace. Explicit cast documents the narrowing is safe.
+  return (int) sent;
 }
 
 /**
--- a/conn/gnutls.c
+++ b/conn/gnutls.c
@@ -1127,13 +1127,7 @@ static int tls_socket_write(struct Connection *conn, const char *buf, size_t cou
     sent += rc;
   } while (sent < count);
 
-  // sent is a size_t accumulator narrowed into this function's int return
-  // type, mandated by the shared Connection::write() interface -- the same
-  // pattern flagged by Coverity CID 392753/392751 in raw.c/tunnel.c, though
-  // not currently flagged here. The sole caller of conn->write(),
-  // mutt_socket_write_d(), already takes an int len, so count is bounded to
-  // INT_MAX before it ever reaches here -- see #4940 for the full
-  // call-graph trace. Explicit cast documents the narrowing is safe.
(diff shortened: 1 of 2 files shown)


flatcap (changes requested):
Thanks.
The fixes are sound, but talk about verbose.

The commit message documents everything anyone would every care to know,
so I suggest reducing the comments to one-liners, mentioning `Connection::write()`.

flatcap:
btw, thought any more about joining NeoMutt?

chrisdebian:
Thanks for asking again, I appreciate it. Could I hold off on joining for now? I tend to invest fairly intensively in a project for a stretch, then need to step back for a while, and I'd rather not accept and then not be as consistently present as being part of the org would suggest. Happy to keep contributing via PRs in the meantime.

flatcap:
> Could I hold off on joining for now?

Sure thing.

> I tend to invest fairly intensively in a project for a stretch, then need to step back for a while

hehe, I understand

> I'd rather not accept and then not be as consistently present as being part of the org would suggest.

Oh, there's no pressure from being a member.
(plus, every 6 months I cull the quiet members)

> Happy to keep contributing via PRs in the meantime.

Brilliant! :heart:

flatcap (approved):

Issue #4956, opened 2026-07-24
translate.yml pins a deploy action to @master instead of a version


chrisdebian:
While reviewing the CI setup, I noticed `translate.yml` pushes directly to the live `neomutt.github.io` repo using `ad-m/github-push-action@master`, pinned to the action's mutable default branch rather than a released version. This is the only workflow doing this that also holds a real secret (`TRANSLATE_DEPLOY_KEY`).

The risk: if that action's `master` branch were ever compromised (a malicious commit pushed by someone who gains control of that third-party repo), the next run of `translate.yml` would execute it with push access to the live website, using a real deploy key. Two other workflows also pin actions to `@master` (`cifuzz.yml`, `macos.yml`), but neither of those holds a write-capable secret, so this one is the one that actually matters.

How I'd address it, if useful: pin to the exact commit SHA of the current `v1.3.0` release tag instead, with a comment noting the version:

```yaml
uses: ad-m/github-push-action@881a6320fdb16eb5318c5054f31c218aec2b324c  # v1.3.0
```

I checked, and `master` currently points at that exact same commit, so this would be a zero functional change today, purely closing off the future risk. Happy to open a PR with exactly this change if you'd like it, or if you'd rather handle it yourself (maybe alongside rotating the deploy key, or checking nothing else needs the same treatment), that's completely fine too -- just wanted to flag it clearly either way.

AI disclosure: found and written up with Claude Code's assistance while reviewing the CI setup, reviewed and sent by me.

chrisdebian:
Correction/addition: while scoping a follow-up CI item (adding explicit `permissions:` blocks), I found a second instance of this exact issue that I missed the first time around -- \`doxygen.yml\` also uses \`ad-m/github-push-action@master\`, pushing to \`neomutt/code\` with a different deploy secret (\`secrets.DOXYGEN_DEPLOY_KEY\`). Same risk, same fix: pin to the SHA of the current \`v1.3.0\` release tag (\`881a6320fdb16eb5318c5054f31c218aec2b324c\`) instead of \`@master\` -- both workflows use the exact same action, so it's the same one-line change in two files.

Sorry for not catching this the first time -- wanted to correct it rather than leave the issue understating the scope.

flatcap:
Ah, good call.

I think I'll go for a different approach.
A bit of chat on a [push-action issue](https://github.com/ad-m/github-push-action/issues/104) led me to try using the **checkout** action to deploy instead -- #4957

[...]

PR #4960, opened 2026-07-25
ci: pin two more actions off the @master anti-pattern

chrisdebian:
Same issue as #4954 fixed elsewhere: both actions were pinned to their mutable default branch instead of a fixed commit.

`google/oss-fuzz` has no tagged releases at all (confirmed via the GitHub API — zero tags, zero releases), so pinned to the current `master` SHA directly with a date comment rather than a version tag. `Homebrew/actions` does cut dated releases, pinned to the latest (`2026.07.20.1`) with its SHA.

Neither of these actions holds a write-capable secret (unlike `translate.yml`'s push action, flagged separately in #4956), so this is safe to send without prior discussion.

### DIFF
--- a/.github/workflows/cifuzz.yml
+++ b/.github/workflows/cifuzz.yml
@@ -17,7 +17,7 @@ jobs:
     steps:
       - name: Build Fuzzers (${{ matrix.sanitizer }})
         id: build
-        uses: google/oss-fuzz/infra/cifuzz/actions/build_fuzzers@master
+        uses: google/oss-fuzz/infra/cifuzz/actions/build_fuzzers@aa9b1bbbb35c68581047066ca1681ef258c313ae  # master as of 2026-07-23; oss-fuzz has no tagged releases
         with:
           oss-fuzz-project-name: 'neomutt'
           dry-run: false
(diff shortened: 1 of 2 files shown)


flatcap:
hmm...
Looking at the histories, the oss-fuzz action doesn't change often, but the Homebrew action does.

If we pin our action to a particular commit, how do we know when to update?
Is that something the Dependabot can handle?

chrisdebian:
Good question. Dependabot's `github-actions` ecosystem is built to track SHA-pinned actions — the `Homebrew/actions` pin has a version comment (`2026.07.20.1`), which is exactly the pattern it looks for, so that one should get automatic update PRs same as any tagged action. `google/oss-fuzz` is the harder case since it has no tags or releases at all — there's no version for Dependabot to diff against, so I'm honestly not certain it'll catch drift there the same way. Already covered by the existing `dependabot.yml` from #4955 either way, no new config needed.

flatcap:
OK, rather than waiting months/years for upstream changes...
How about we pin both to an old commit (just one back).

On Monday, we see if the Dependabot notices.
If it does, great.
If not, we stick with `@master` and the minor risks.

chrisdebian:
Done, pinned both one step further back:
- oss-fuzz: 9fe8e9a (one commit behind; no tags to go by there)
- Homebrew/actions: 18fcb8e (2026.07.13.1, one release behind)

Will check back Monday to see if Dependabot has opened anything.

flatcap:
Of course it'll never work if I forget to merge it.  D'oh!

Issue #4962, opened 2026-07-25
IWYU CI check — proposal, cross-referencing iwyu.txt


chrisdebian:
@flatcap — saw `iwyu.txt` in your [brain-dump repo](https://github.com/neomutt/flatcap-brain-dump/blob/main/iwyu.txt) and had a look at what's actually involved in wiring IWYU into CI.

Turns out most of the hard infrastructure is already there:
- `./configure --compile-commands` (needs clang) is native, working build-system support.
- `github.com/neomutt/iwyu` already has `bin/iwyu.sh`, which auto-selects the right per-module mapping file from 55 `.imp` files — one per source directory, comprehensive coverage.
- `iwyu` itself is a standard `apt install iwyu` package on Ubuntu 24.04, which matches what GitHub's `ubuntu-latest` runners use — no from-source build needed.

I built with clang + `--compile-commands` and ran `iwyu.sh` against a small sample (`conn/*.c`) to check real-world signal quality before proposing anything. Result was clean and low-noise: 5 of the files came back "correct #includes/fwd-decls" with zero findings, and the one genuine hit (`raw.c` using `struct timeval` without including `<sys/time.h>`, currently working only because `<sys/select.h>` pulls it in transitively) checked out as a real, if minor, issue.

**Proposed approach, if you're on board:**
- New GitHub Actions workflow, clang + `--compile-commands` + `iwyu.sh` across the source tree.
- Warn/report mode rather than a hard CI failure — IWYU's suggestions are judgment calls (that's presumably part of why this has sat as a brain-dump item rather than already being CI), so a non-blocking report seems like the safer starting point than gating merges on it.
- Happy to open a PR with this, and separately fix the `raw.c` finding as its own small PR, once you've had a chance to weigh in on the approach — didn't want to build the whole thing unprompted given the policy calls involved.
- Once implemented, happy to send a follow-up marking `iwyu.txt` as done in your brain-dump repo — noting it's implemented rather than deleting the entry, so the history stays intact.

AI disclosure: found and written up with Claude Code's assistance while reviewing the CI setup, reviewed and sent by me.

flatcap:
Yeah, let's go for it...

I _have_ been thinking about this for some time,
but upgrading from some hacky bash and vim scripts to automation always seemed a bit daunting.

## Good

For 98% of the source (measured), IWYU is amazing.
(The NeoMutt tests should scan cleanly, too).

I run it periodically when bored, or after refactoring work (last run 2 months ago).

[...]

PR #4964, opened 2026-07-26
ci: add IWYU check (report-only)

chrisdebian:
Implements the approach proposed and agreed in #4962.

- New `iwyu.yml` workflow: configures with `--compile-commands`, builds
  `compile_commands.json`, and runs IWYU via `neomutt/iwyu`'s `iwyu.sh`
  (auto-selects the right per-module `.imp` mapping file).
- Runs in `ghcr.io/neomutt/fedora:44` rather than Ubuntu, per your comment
  — that image already ships `clang` and `iwyu` at versions matching
  Fedora 44's own repos, so results should track what you see locally
  rather than whatever Ubuntu's `apt` package happens to carry.
- Excludes the 18 files you listed as ignored in your own local runs
  (`debug/*.[ch]`, `color/debug.[ch]`, `conn/openssl.c`, `gui/resize.c`,
  `gui/terminal.c`, `imap/auth_sasl.c`).
- Report-only: findings are uploaded as a workflow artifact, the job
  doesn't fail the build. Given IWYU's suggestions are judgement calls,
  this seemed like the safer starting point — happy to revisit going
  stricter later if it proves reliable.
- Triggers on PRs touching `.c`/`.h` files, on push to `main`, and via
  manual dispatch.

The one real finding from initial scoping (`raw.c` missing
`sys/time.h`) is already up as its own small PR: #4963.

AI disclosure: implemented with Claude Code's assistance, following up
on the proposal already discussed in #4962.

### DIFF
--- a/.github/workflows/iwyu.yml
+++ b/.github/workflows/iwyu.yml
@@ -1,0 +1,62 @@
+name: IWYU
+
+on:
+  pull_request:
+    paths:
+      - '**/*.c'
+      - '**/*.h'
+  push:
+    branches:
+      - 'main'
+  workflow_dispatch:
+
+concurrency:
+  group: ${{ github.workflow }}-${{ github.ref }}
+  cancel-in-progress: true
+
+jobs:
+  iwyu:
+    if: github.event.pull_request.draft == false
+    runs-on: ubuntu-latest
+    container: ghcr.io/neomutt/ubuntu
+    permissions:
+      contents: read
+
+    steps:
+      - name: Checkout Code
+        uses: actions/checkout@v7
+
+      - name: Checkout IWYU Mapping Files
+        uses: actions/checkout@v7
+        with:
+          repository: neomutt/iwyu
+          path: iwyu-tool
+
+      - name: Install IWYU
+        run: apt-get update && apt-get install -y iwyu
+
+      - name: Configure NeoMutt
+        run: |
+          CC=clang ./configure --autocrypt --fmemopen --gdbm --gnutls --gpgme --gss \
+            --lmdb --lua --lz4 --notmuch --pcre2 --rocksdb --sasl --tdb \
+            --with-lock=fcntl --zlib --zstd --compile-commands
+
+      - name: Build (generates compile_commands.json)
+        run: make -j "$(nproc)" compile_commands.json
+
+      - name: Run IWYU
+        run: |
+          chmod +x iwyu-tool/bin/iwyu.sh
+          # Drive the file list from compile_commands.json itself, not a blind
+          # find -- some source files (e.g. conn/gsasl.c) are excluded by the
+          # feature flags used to configure this build, and running IWYU on
+          # them anyway produces spurious compile errors unrelated to includes.
+          tac compile_commands.json | sed '2 s/.$//' | tac | jq -r '.[] | .file' \
+            | xargs iwyu-tool/bin/iwyu.sh 2>&1 | tee iwyu-report.txt || true
+
+      - name: Upload IWYU Report
+        uses: actions/upload-artifact@v7
+        if: always()
+        with:
+          name: iwyu-report
+          path: iwyu-report.txt
--- a/.github/workflows/iwyu.yml
+++ b/.github/workflows/iwyu.yml
@@ -49,11 +49,22 @@ jobs:
           # feature flags used to configure this build, and running IWYU on
           # them anyway produces spurious compile errors unrelated to includes.
           #
-          # Excludes files flatcap ignores in his own local IWYU runs (#4962):
-          # debug/*.[ch], color/debug.[ch], conn/openssl.c, gui/resize.c,
-          # gui/terminal.c, imap/auth_sasl.c.
+          # Excludes files flatcap ignores in his own local IWYU runs (#4962),
+          # plus two machine-generated files not meant for include-hygiene
+          # checks at all: debug/*.[ch], color/debug.[ch], conn/openssl.c,
+          # gui/resize.c, gui/terminal.c, imap/auth_sasl.c, git_ver.c,
+          # conststrings.c.
+          #
+          # `compile_commands.json`'s `file` field is repo-root-relative with
+          # no leading slash (e.g. "gui/resize.c", not "/gui/resize.c") --
+          # patterns below use `(^|/)` rather than a bare leading `/`, which
+          # would never match and silently exclude nothing. Confirmed via a
+          # local build: the bare-`/` version matched zero files against a
+          # real compile_commands.json, letting all six targeted files
+          # through unfiltered -- caught by flatcap running the Action
+          # locally and diffing against his own manual IWYU runs (#4964).
           tac compile_commands.json | sed '2 s/.$//' | tac | jq -r '.[] | .file' \
-            | grep -vE '/debug/[^/]+\.[ch]$|/color/debug\.[ch]$|/conn/openssl\.c$|/gui/resize\.c$|/gui/terminal\.c$|/imap/auth_sasl\.c$' \
+            | grep -vE '(^|/)debug/[^/]+\.[ch]$|(^|/)color/debug\.[ch]$|(^|/)conn/openssl\.c$|(^|/)gui/resize\.c$|(^|/)gui/terminal\.c$|(^|/)imap/auth_sasl\.c$|(^|/)git_ver\.c$|(^|/)conststrings\.c$' \
             | xargs iwyu-tool/bin/iwyu.sh 2>&1 | tee iwyu-report.txt || true
 
       - name: Upload IWYU Report
--- a/.github/workflows/iwyu.yml
+++ b/.github/workflows/iwyu.yml
@@ -18,7 +18,7 @@ jobs:
   iwyu:
     if: github.event.pull_request.draft == false
     runs-on: ubuntu-latest
-    container: ghcr.io/neomutt/ubuntu
+    container: ghcr.io/neomutt/fedora:44
     permissions:
       contents: read
 
@@ -31,9 +31,6 @@ jobs:
         with:
           repository: neomutt/iwyu
           path: iwyu-tool
-
-      - name: Install IWYU
-        run: apt-get update && apt-get install -y iwyu
 
       - name: Configure NeoMutt
         run: |
@@ -51,7 +48,12 @@ jobs:
           # find -- some source files (e.g. conn/gsasl.c) are excluded by the
           # feature flags used to configure this build, and running IWYU on
           # them anyway produces spurious compile errors unrelated to includes.
+          #
+          # Excludes files flatcap ignores in his own local IWYU runs (#4962):
+          # debug/*.[ch], color/debug.[ch], conn/openssl.c, gui/resize.c,
+          # gui/terminal.c, imap/auth_sasl.c.
           tac compile_commands.json | sed '2 s/.$//' | tac | jq -r '.[] | .file' \
+            | grep -vE '/debug/[^/]+\.[ch]$|/color/debug\.[ch]$|/conn/openssl\.c$|/gui/resize\.c$|/gui/terminal\.c$|/imap/auth_sasl\.c$' \
             | xargs iwyu-tool/bin/iwyu.sh 2>&1 | tee iwyu-report.txt || true
 
       - name: Upload IWYU Report


.github/workflows/iwyu.yml
```diff
@@ -0,0 +1,64 @@
...
+          path: iwyu-tool
+
+      - name: Configure NeoMutt
+        run: |
+          CC=clang ./configure --autocrypt --fmemopen --gdbm --gnutls --gpgme --gss \
+            --lmdb --lua --lz4 --notmuch --pcre2 --rocksdb --sasl --tdb \
+            --with-lock=fcntl --zlib --zstd --compile-commands
+
+      - name: Build (generates compile_commands.json)
+        run: make -j "$(nproc)" compile_commands.json
```
flatcap, line 42: Ah!  `$(nproc)` that's a bit neater.
We can probably drop _this_ from the Linux Actions:
```yml
- name: Get number of CPU cores
  uses: SimenB/github-actions-cpu-cores@v3
```

.github/workflows/iwyu.yml
```diff
@@ -0,0 +1,64 @@
...
+
+      - name: Run IWYU
+        run: |
+          chmod +x iwyu-tool/bin/iwyu.sh
```
flatcap, line 46: Do we need this?
The execute bit is set in the repo.

.github/workflows/iwyu.yml
```diff
@@ -0,0 +1,64 @@
...
+          # Drive the file list from compile_commands.json itself, not a blind
+          # find -- some source files (e.g. conn/gsasl.c) are excluded by the
+          # feature flags used to configure this build, and running IWYU on
+          # them anyway produces spurious compile errors unrelated to includes.
+          #
+          # Excludes files flatcap ignores in his own local IWYU runs (#4962):
+          # debug/*.[ch], color/debug.[ch], conn/openssl.c, gui/resize.c,
+          # gui/terminal.c, imap/auth_sasl.c.
+          tac compile_commands.json | sed '2 s/.$//' | tac | jq -r '.[] | .file' \
+            | grep -vE '/debug/[^/]+\.[ch]$|/color/debug\.[ch]$|/conn/openssl\.c$|/gui/resize\.c$|/gui/terminal\.c$|/imap/auth_sasl\.c$' \
```
flatcap: hmm... ok
This checks the files in the build, but leaves some gaps.

.github/workflows/iwyu.yml
```diff
@@ -0,0 +1,64 @@
...
+        run: |
+          chmod +x iwyu-tool/bin/iwyu.sh
+          # Drive the file list from compile_commands.json itself, not a blind
+          # find -- some source files (e.g. conn/gsasl.c) are excluded by the
+          # feature flags used to configure this build, and running IWYU on
+          # them anyway produces spurious compile errors unrelated to includes.
+          #
+          # Excludes files flatcap ignores in his own local IWYU runs (#4962):
+          # debug/*.[ch], color/debug.[ch], conn/openssl.c, gui/resize.c,
+          # gui/terminal.c, imap/auth_sasl.c.
```
flatcap: The debug source has no regard for compartmentalisation :-)

`openssl.c` and `auth_sasl.c` rely on a conditional build (and I use the other versions).
We should probably find a way to test them -- e.g. multiple builds.

.github/workflows/iwyu.yml
```diff
@@ -0,0 +1,64 @@
...
+          tac compile_commands.json | sed '2 s/.$//' | tac | jq -r '.[] | .file' \
+            | grep -vE '/debug/[^/]+\.[ch]$|/color/debug\.[ch]$|/conn/openssl\.c$|/gui/resize\.c$|/gui/terminal\.c$|/imap/auth_sasl\.c$' \
+            | xargs iwyu-tool/bin/iwyu.sh 2>&1 | tee iwyu-report.txt || true
+
+      - name: Upload IWYU Report
```
flatcap, line 70: Lovely. We just need some major filtering.

flatcap (commented):
Thanks, @chrisdebian!
That's a great start.
I'll list my 'file' findings below.

flatcap:
I ran the Action locally and compared the results to my manual runs.

Files Checked: `*.c` (504)

This pulled in quite a few header files, but not all.

<details>
<summary>Missing headers (206)</summary>

- `address/idna2.h`
- `address/lib.h`
- `address/module_data.h`
- `alias/lib.h`
- `alias/module_data.h`
- `attach/lib.h`
- `attach/module_data.h`
- `autocrypt/lib.h`
- `autocrypt/module_data.h`
- `autocrypt/private.h`
- `bcache/lib.h`
- `bcache/module_data.h`
- `browser/lib.h`
- `browser/module_data.h`
- `cli/lib.h`
- `color/curses2.h`
- `color/lib.h`
- `color/merged.h`
- `color/module_data.h`
- `color/notify2.h`
- `color/regex4.h`
- `color/simple2.h`
- `commands/commands.h`
- `commands/lib.h`
- `commands/module_data.h`
- `complete/compapi.h`
- `complete/lib.h`
- `complete/module_data.h`
- `compmbox/lib.h`
- `compmbox/module_data.h`
- `compose/lib.h`
- `compose/module_data.h`
- `compose/private.h`
- `compress/lib.h`
- `compress/module_data.h`
- `compress/private.h`
- `config/inheritance.h`
- `config/lib.h`
- `config/module_data.h`
- `config/quad.c`
- `config/regex2.h`
- `config/set.c`
- `config/types.h`
- `conn/connection.h`
- `conn/gnutls.c`
- `conn/gsasl.c`
- `conn/gsasl2.h`
- `conn/lib.h`
- `conn/module_data.h`
- `conn/private.h`
- `conn/raw.c`
- `conn/sasl.c`
- `conn/ssl.h`
- `convert/lib.h`
- `convert/module_data.h`
- `core/lib.h`
- `core/mailbox.c`
- `core/module_api.h`
- `core/module_data.h`
- `core/mxapi.h`
- `core/neomutt.c`
- `editor/lib.h`
- `editor/module_data.h`
- `editor/wdata.h`
- `email/commands.h`
- `email/content.h`
- `email/lib.h`
- `email/module_data.h`
- `envelope/lib.h`
- `envelope/module_data.h`
- `envelope/private.h`
- `expando/definition.h`
- `expando/domain.h`
- `expando/filter.h`
- `expando/lib.h`
- `expando/module_data.h`
- `expando/uid.h`
- `fuzz/address.c`
- `fuzz/body.c`
- `fuzz/cli.c`
- `fuzz/date.c`
- `fuzzy/benchmark.c`
- `fuzzy/lib.h`
- `fuzzy/module_data.h`
- `globals.h`
- `gui/functions.h`
- `gui/lib.h`
- `gui/module_data.h`
- `gui/mutt_curses.h`
- `gui/rootwin.h`
- `hcache/lib.h`
- `hcache/module_data.h`
- `helpbar/lib.h`
- `helpbar/module_data.h`
- `helpbar/private.h`
- `history/lib.h`
- `history/module_data.h`
- `hooks/dump.c`
- `hooks/dump.h`
- `hooks/exec.h`
- `hooks/lib.h`
- `hooks/module_data.h`
- `hooks/parse.h`
- `imap/auth.c`
- `imap/auth_anon.c`
- `imap/auth_cram.c`
- `imap/auth_gsasl.c`
- `imap/auth_gss.c`
- `imap/lib.h`
- `imap/module_data.h`
- `imap/private.h`
- `index/ibar.h`
- `index/index.c`
- `index/lib.h`
- `index/module_data.h`
- `index/private.h`
- `key/commands.c`
- `key/dump.c`
- `key/get.c`
- `key/init.c`
- `key/lib.h`
- `key/menu.c`
- `key/module_data.h`
- `key/notify.h`
- `lua/lib.h`
- `lua/module_data.h`
- `maildir/hcache.h`
- `maildir/lib.h`
- `maildir/module_data.h`
- `maildir/path.c`
- `maildir/shared.h`
- `main.c`
- `mbox/lib.h`
- `mbox/module_data.h`
- `menu/lib.h`
- `menu/module_data.h`
- `menu/private.h`
- `menu/type.h`
- `mh/lib.h`
- `mh/module_data.h`
- `mlist/functions.c`
- `mlist/lib.h`
- `mlist/module_data.h`
- `mutt.h`
- `mutt/array.h`
- `mutt/ctype2.h`
- `mutt/date.c`
- `mutt/eqi.h`
- `mutt/lib.h`
- `mutt/logging2.h`
- `mutt/message.h`
- `mutt/module_data.h`
- `mutt/notify_type.h`
- `mutt/observer.h`
- `mutt/regex3.h`
- `mutt/signal2.h`
- `mutt/string.c`
- `mutt/string2.h`
- `mutt/timegm.c`
- `mutt/wcscasecmp.c`
- `ncrypt/gpgme_functions.h`
- `ncrypt/lib.h`
- `ncrypt/module_data.h`
- `ncrypt/pgp_functions.h`
- `ncrypt/private.h`
- `ncrypt/smime_functions.h`
- `ncrypt/sort.h`
- `nntp/expando_newsrc.h`
- `nntp/lib.h`
- `nntp/module_data.h`
- `nntp/private.h`
- `notmuch/lib.h`
- `notmuch/module_data.h`
- `notmuch/private.h`
- `notmuch/query.c`
- `pager/lib.h`
- `pager/module_data.h`
- `parse/lib.h`
- `parse/module_data.h`
- `pattern/exec.c`
- `pattern/expando.h`
- `pattern/lib.h`
- `pattern/module_data.h`
- `pattern/private.h`
- `pop/auth.c`
- `pop/lib.h`
- `pop/module_data.h`
- `pop/private.h`
- `postpone/lib.h`
- `postpone/module_data.h`
- `progress/lib.h`
- `progress/module_data.h`
- `progress/window.h`
- `question/lib.h`
- `question/module_data.h`
- `send/header.c`
- `send/lib.h`
- `send/module_data.h`
- `send/send.c`
- `send/smtp.c`
- `sidebar/functions_fuzzy.h`
- `sidebar/lib.h`
- `sidebar/module_data.h`
- `sidebar/private.h`
- `store/lib.h`
- `store/module_data.h`

</details>

It checked some files that it shouldn't have:

- Machine generated:
  - `git_ver.c`
  - `conststrings.c`

- Meant to be excluded:
  - `debug/names_expando.c`
  - `gui/resize.c`
  - `gui/terminal.c`
  - `imap/auth_sasl.c`

It missed some source files.
We could run multiple builds to pick these up.

- Configurable:
  - `conn/gsasl.c`
  - `imap/auth_anon.c`
  - `imap/auth_cram.c`
  - `imap/auth_gsasl.c`

- Missing dependency files:
  - `mutt/timegm.c`
  - `mutt/wcscasecmp.c`

- Misc other files
  - `fuzz/address.c`
  - `fuzz/body.c`
  - `fuzz/cli.c`
  - `fuzz/date.c`
  - `fuzzy/benchmark.c`

flatcap:
After heavily filtering the results, I get:

```
main.c should remove these lines:
- #include "conn/lib.h"

conn/raw.c should remove these lines:
- #include <sys/time.h>

gui/terminal.c should remove these lines:
- #include <ncursesw/term.h>

imap/auth_gss.c should add these lines:
#include <stdint.h>

key/commands.c should add these lines:
#include "key/menudef.h"

key/dump.c should add these lines:
#include "key/menudef.h"

key/get.c should add these lines:
#include "key/menudef.h"

key/menu.c should add these lines:
#include "key/menudef.h"
```

I suggest we don't fix these, for now, so we have a benchmark.

flatcap:
[Rebased over `main`]

chrisdebian:
Thanks for running this locally and diffing against your manual runs — found the actual bug.

The exclusion pattern used a leading `/` (e.g. `/gui/resize\.c$`), assuming `compile_commands.json`'s `file` field is absolute or otherwise slash-prefixed. Checked with a real local build: it's actually repo-root-relative with no leading slash (`gui/resize.c`, not `/gui/resize.c`), so the pattern matched **zero files** — silently excluding nothing. That's exactly why `debug/names_expando.c`, `gui/resize.c`, `gui/terminal.c`, and `imap/auth_sasl.c` all leaked through.

Fixed by anchoring each pattern with `(^|/)` instead of a bare `/`, and added `git_ver.c`/`conststrings.c` (the two machine-generated files) to the exclusion list too — those weren't on your original list since they're not something you'd think to mention, but generated code shouldn't go through include-hygiene checks either way. Re-verified against a real `compile_commands.json`: exactly those 6 files excluded now, 505 remain (close to your 504 — small diff expected, different build environment). Pushed as a new commit on the same branch.

Left the other two findings as-is for now, since they're bigger scope than this PR:
- The ~206 missing headers are inherent to driving the file list from `compile_commands.json` (which only lists `.c` files actually compiled, never headers pulled in transitively) — would need a different mechanism to close.
- The conditionally-compiled files (`conn/gsasl.c` etc.) match what the workflow's own comment already flagged as a known limitation — your multiple-builds suggestion sounds right, but feels like a separate follow-up PR rather than something to fold into this one.

flatcap:
> ... follow-up PR ...

Yeah

Testing Actions is always a pain.
Until they hit `main`, GitHub seems unaware of them.

flatcap (approved):

flatcap:
For the future improvements..
If we need any external file lists or scripts, we can put them in `.github` (or a subdir if there are lots).

flatcap:
**Update:**
[The Action has run](https://github.com/neomutt/neomutt/actions/runs/30266302173) on `main`.

The iwyu-report contains the following wisdom:
```
iwyu command not found in PATH. You can specify an alternate location with IWYU=/path/to/iwyu
```

flatcap:
I recently read that upload-artifact can now save a **single unzipped** file.

https://github.com/actions/upload-artifact#upload-an-individual-file-unzipped

PR #4965, opened 2026-07-26
ci: add clang-format enforcement

chrisdebian:
Adds CI enforcement for clang-format, which .clang-format already
configures and .github/skills/format-and-lint.yml can run locally,
but nothing checked in CI.

First commit fixes 10 files that had accumulated purely mechanical
formatting drift (compound-literal brace spacing, comment indentation
before preprocessor directives) - version drift since they were last
formatted, no functional change, all rebuild clean.

One of those files (test/parse/parse_rc_line.c) hit a genuine
clang-format 18 quirk: a local function prototype declared inside a
function body gets its pointer misaligned even though
PointerAlignment: Right is set. Reproduced in isolation before
concluding it wasn't user error on my end. Guarded with clang-format
off/on rather than accepting output that contradicts the project's
own stated rule.

Second commit adds the workflow itself, mirroring the existing local
skill's file selection (*.c only, excludes autosetup/ and docs/ -
headers are intentionally hand-formatted per project convention).

AI assistance: Claude Code drafted both commits and this PR
description. I reviewed the diffs, reproduced the clang-format quirk
independently to confirm it before working around it, and verified
all touched files rebuild clean before pushing.

### DIFF
--- a/.github/workflows/clang-format.yml
+++ b/.github/workflows/clang-format.yml
@@ -1,0 +1,44 @@
+name: Clang Format
+
+on:
+  pull_request:
+  push:
+    branches:
+      - 'main'
+      - 'devel/**'
+      - 'copilot/**'
+  workflow_dispatch:
+
+concurrency:
+  group: ${{ github.workflow }}-${{ github.ref }}
+  cancel-in-progress: true
+
+permissions:
+  contents: read
+
+jobs:
+  check:
+    name: Check Formatting
+    runs-on: ubuntu-latest
+
+    steps:
+      - name: Checkout Code
+        uses: actions/checkout@v7
+
+      - name: Install clang-format
+        run: sudo apt-get install -y clang-format
+
+      - name: Check formatting
+        run: |
+          FAIL=0
+          while IFS= read -r file; do
+            if ! clang-format -style=file "$file" | cmp -s "$file" -; then
+              echo "::error file=$file::needs clang-format"
+              FAIL=1
+            fi
+          done < <(git ls-files -- ':!autosetup/' ':!docs/' '*.c')
+
+          if [ "$FAIL" -ne 0 ]; then
+            echo "One or more files need clang-format. Run: clang-format -style=file -i <file>"
+            exit 1
+          fi
--- a/.github/workflows/clang-format.yml
+++ b/.github/workflows/clang-format.yml
@@ -26,7 +26,11 @@ jobs:
         uses: actions/checkout@v7
 
       - name: Install clang-format
-        run: sudo apt-get install -y clang-format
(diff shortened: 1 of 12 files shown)


flatcap:
Clang-Format?  Great stuff...

However, I've just run `clang-format` (v22.1.8) locally and it dutifully undid all the changes you made in 251cd57cc26baa85af753ed19fd940448b31504f :-)

I think this is another case where we need a modern version, else we'll get into format wars.

Cheers

flatcap:
btw, what's '**(P5-2.3)'**?

chrisdebian:
Hi flatcap,

`(P5-2.3)` is just my own internal tracking tag for this piece of work, not a NeoMutt convention — sorry for the confusion, I'll drop it from PR titles here going forward.

On the clang-format version — you're right, and it's actually the same problem this PR was trying to close. The CI workflow installs clang-format via plain `apt-get install -y clang-format`, so it gets whatever version Ubuntu's runner image ships, unpinned. That's almost certainly older than your v22.1.8, which is why it approved formatting that your local run then undid.

Rather than leave that gap, I'd like to pin CI to a specific clang-format version so there's one source of truth everyone can match locally. Do you already have a preferred version in mind, or should I pin to the current stable release (looks like 22.x)? Happy to push a follow-up commit to this PR once you confirm.

flatcap:
> `(P5-2.3)` is just my own internal tracking tag for this piece of work

Damn, you're organised :-)

> should I pin to the current stable release (looks like 22.x)

Yeah, that'll be good.
Thanks

chrisdebian:
Pinned CI to 22.1.8 (via the \`clang-format\` PyPI package, which has an exact matching build) and reformatted the 10 files that drifted under that version - same two categories as before, no functional change. Should be consistent with your local run now.

flatcap:
We don't need the protection in [test/parse/parse_rc_line.c](https://github.com/neomutt/neomutt/pull/4965/changes#diff-f6becffa4e01a42a145e458bd6a73ac8a5dec2488631ff9ceb278fe439b52914) any more either.
(the new clang-format doesn't mangle it)

chrisdebian:
Good catch — confirmed, removed. Pushed.

flatcap:
Possible future upgrades:

- Limit clang-format to the files that the user actually changed
  (it seems unfair to criticise them for someone else's formatting :-)

- Run clang-format on the changed files and save the diff as an artifact

- Commit the changes to the PR branch
  (in the bright future where we trust the Action implicitly :-)

flatcap (approved):

chrisdebian:
Thanks! All three are good ideas — logging them as follow-ups. Appreciate you taking the time on this one.

PR #4969, opened 2026-07-27
ci: enable UBSan alongside ASan (with a real bug found along the way)

chrisdebian:
ASan already runs in CI (`asan.yml`), but UBSan never did, despite `auto.def` supporting it (`--ubsan`). Tested locally first rather than just flipping the flag on — that surfaced a real, if practically harmless, undefined-behaviour bug (first commit): `attachmatch_free()` was declared as `void(struct AttachMatch **)` but called everywhere through a `(list_free_t)` cast to `void(void **)`. Calling through an incompatible function-pointer type is UB per the C standard — works fine on this ABI (both are just pointer-to-pointer), which is exactly why nobody had noticed. Fixed by giving it the generic `void **` signature `list_free_t` actually expects, matching the pattern the codebase's own test (`test_list_free` in `test/list/mutt_list_free_type.c`) already uses, and casting internally instead of at each of the 8 call sites.

Second commit adds `--ubsan` to the existing configure invocation, `UBSAN_OPTIONS` with `halt_on_error=1` so a real violation fails the job rather than just logging, and a log artifact upload matching the existing ASan one.

**Being upfront about one thing I couldn't fully verify**: testing locally (a git worktree, ASan+UBSan both on) also hit 17 test failures unrelated to either sanitizer — `mkstemp`/`chmod`/`stat` calls failing in ways that look specific to my local sandbox's filesystem restrictions rather than real bugs (they don't reproduce without the sanitizers active, but they're not proper UBSan/ASan reports either — no violation logged for any of them, just quiet permission-shaped failures). I couldn't rule out whether the actual CI container behaves the same way. If this run fails for reasons unrelated to the `attachmatch_free` fix, that's most likely what's happening — happy to help dig further if so.

AI assistance: Claude Code investigated the UBSan finding, wrote both commits and this description. I reviewed the diff, the reasoning for the fix, and reproduced the before/after locally before submitting.

### DIFF
--- a/.github/workflows/asan.yml
+++ b/.github/workflows/asan.yml
@@ -23,7 +23,8 @@ jobs:
 
     env:
       ASAN_OPTIONS: detect_leaks=1:color=never:log_path=asan:print_legend=false:print_summary=false
-      CONFIGURE_OPTIONS: --asan --autocrypt --gdbm --gnutls --gpgme --gss --lmdb --lua --lz4 --notmuch --sasl --tdb --with-lock=fcntl --zlib --zstd
+      UBSAN_OPTIONS: print_stacktrace=1:halt_on_error=1:log_path=ubsan
+      CONFIGURE_OPTIONS: --asan --ubsan --autocrypt --gdbm --gnutls --gpgme --gss --lmdb --lua --lz4 --notmuch --sasl --tdb --with-lock=fcntl --zlib --zstd
       # Force Node 24 until github-actions-cpu-cores gets a new release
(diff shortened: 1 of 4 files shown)


flatcap:
> Second commit adds `--ubsan`

hmm... for some reason, I had it in my head that you couldn't enable both at the same time

> also hit 17 test failures unrelated to either sanitizer

Un-sandboxed, it all looks good to me.

flatcap (approved):
Thanks!

PR #4970, opened 2026-07-28
ci: add clang-tidy check (report-only)

chrisdebian:
Adds a clang-tidy job alongside the existing IWYU workflow, same shape: built via `--compile-commands`, run against every file in `compile_commands.json`, uploads a report artifact rather than failing the build. Report-only, per the roadmap's own framing for this — findings don't block CI, this is a starting check set expected to be refined as real signal emerges.

**Checks**: `bugprone-*` and `clang-analyzer-*`, minus two project-specific false-positive sources found while testing locally:

- `bugprone-multi-level-implicit-pointer-conversion` fires on every call to the codebase's own `FREE()`/`ARRAY_FREE()` macros, which take a `void *` via an intentional implicit conversion — a systemic false positive against an established, universal idiom here, not a real defect class.
- `clang-analyzer-security.insecureAPI.*` suggests C11 Annex K bounds-checked functions (`memcpy_s`, `snprintf_s`, etc.) that glibc doesn't implement and this project doesn't use — pure noise on every `memcpy`/`memmove`/`snprintf` call.

Also checked directly (not assumed) whether IWYU's own file-exclusion list would apply here — it doesn't: those files (`conststrings.c`, `git_ver.c`, `debug/names_expando.c`, `gui/resize.c`, `gui/terminal.c`, `imap/auth_sasl.c`) compile and analyse cleanly under clang-tidy with no errors. The headers that trip up IWYU's include-checking aren't relevant to `bugprone`/`clang-analyzer`'s checks, so no exclusions were needed.

**Tested locally**: full 510-file `compile_commands.json` set (same configure flags as `iwyu.yml`), no tool crashes, ~3 warnings/file average with the two checks above disabled. Also confirmed the Makefile's known trailing-comma quirk in `compile_commands.json` (already worked around in `iwyu.yml`) applies here too, same fix.

---

This PR was written primarily by Claude Code, under my direction and review.

### DIFF
--- a/.clang-tidy
+++ b/.clang-tidy
@@ -1,0 +1,27 @@
+---
+# Report-only static analysis, run via .github/workflows/clang-tidy.yml.
+# Findings don't fail CI (the workflow uploads a report artifact instead) --
+# this is a starting check set, expected to be refined over time.
+Checks: >
+  -*,
+  bugprone-*,
+  clang-analyzer-*,
+  -bugprone-multi-level-implicit-pointer-conversion,
+  -bugprone-easily-swappable-parameters,
+  -clang-analyzer-security.insecureAPI.*
+# bugprone-multi-level-implicit-pointer-conversion: fires on every call to
+# the codebase's own FREE()/ARRAY_FREE() macros, which take a `void *` via an
+# intentional implicit conversion -- a systemic false positive against an
+# established, universal idiom here, not a real defect class.
+#
+# bugprone-easily-swappable-parameters: fires on almost any function with two
+# adjacent same-typed pointer parameters (e.g. any (haystack, needle)-style
+# string function) -- too subjective/noisy to be useful as a default-on
+# check.
+#
+# clang-analyzer-security.insecureAPI.*: suggests C11 Annex K bounds-checked
+# functions (memcpy_s, snprintf_s, etc.) that glibc doesn't implement and
+# this project doesn't use -- pure noise on every memcpy/memmove/snprintf
+# call.
+WarningsAsErrors: ''
+HeaderFilterRegex: '.*'
--- a/.github/clang-tidy-summary.py
+++ b/.github/clang-tidy-summary.py
@@ -1,0 +1,111 @@
+#!/usr/bin/env python3
+"""
+Generate quickfix.txt and summary.md from clang-tidy-report.txt.
+
+Filters:
+- Only warnings (not notes, errors, etc.)
+- Ignores files with absolute paths (outside project)
+- Strips leading './' from filenames
+"""
+
+import re
+import sys
+from collections import defaultdict
+from pathlib import Path
+
+INPUT_FILE = "clang-tidy-report.txt"
+QUICKFIX_FILE = "quickfix.txt"
+SUMMARY_FILE = "summary.md"
+
+# Matches: file:line:col: warning: message [check-name]
+WARNING_RE = re.compile(
+    r'^(\.?/?[^/\s][^:]*):(\d+):(\d+):\s+warning:\s+(.*?)\s+\[([^\]]+)\]\s*$'
+)
+
+
+def parse_warnings(path):
+    """Parse warning lines, returning list of (check, file, line, col, message).
+
+    Deduplicates by (file, line, col, check); preserves first-occurrence order
+    of check groups for quickfix output.
+    """
+    warnings = []
+    seen = set()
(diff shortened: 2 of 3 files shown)


flatcap:
Thanks!

I had a play with `clang-tidy` last year, but quickly got overwhelmed by options (and warnings :-)

OK, first run...
```
42316 warnings generated.
```

tidied, de-duped to **1074 warnings**.

| Count | Warning                                                 |
|------:|:--------------------------------------------------------|
|   458 | `[bugprone-narrowing-conversions]`                      |
|   207 | `[bugprone-assignment-in-if-condition]`                 |
|    76 | `[bugprone-unsafe-functions]`                           |
|    65 | `[bugprone-macro-parentheses]`                          |
|    55 | `[bugprone-switch-missing-default-case]`                |
|    39 | `[bugprone-unchecked-string-to-number-conversion]`      |
|    31 | `[clang-analyzer-unix.Errno]`                           |
|    31 | `[clang-analyzer-core.NullDereference]`                 |
|    26 | `[bugprone-implicit-widening-of-multiplication-result]` |
|    26 | `[bugprone-branch-clone]`                               |
|    25 | `[clang-analyzer-deadcode.DeadStores]`                  |
|     9 | `[bugprone-not-null-terminated-result]`                 |
|     7 | `[clang-analyzer-security.ArrayBound]`                  |
|     6 | `[bugprone-inc-dec-in-conditions]`                      |
|     3 | `[bugprone-reserved-identifier]`                        |
|     2 | `[clang-analyzer-unix.Stream]`                          |
|     2 | `[clang-analyzer-optin.performance.Padding]`            |
|     2 | `[clang-analyzer-optin.core.EnumCastOutOfRange]`        |
|     2 | `[bugprone-signed-char-misuse]`                         |
|     1 | `[clang-analyzer-core.NullPointerArithm]`               |
|     1 | `[clang-analyzer-core.NonNullParamChecker]`             |

hmm...

My first thought was, "that's too many for a human to wade through".
But there are a few types that'll be easy for AI to fix en masse.

- `[bugprone-unsafe-functions]` -- mostly `rewind()`
- `[bugprone-macro-parentheses]`
- `[bugprone-switch-missing-default-case]`

These should lead to very small, self-contained, fixes.

chrisdebian:
Picked off the `bugprone-unsafe-functions` category (rewind()/setbuf(), 76 sites) as a first pass — opened as #4974. The other two you flagged (`bugprone-macro-parentheses`, `bugprone-switch-missing-default-case`) I'll look at separately, since macro-parentheses in particular touches some widely-used core macros (array.h, queue.h) that deserve a more careful look than a purely mechanical pass.

chrisdebian:
Second category done — `bugprone-switch-missing-default-case` (55 sites) opened as #4975. Last one remaining is `bugprone-macro-parentheses`, which I'll look at separately since a good number of those sites are in core array/queue macros used throughout the codebase and deserve more individual care than a mechanical pass.

chrisdebian:
Third and last category done — `bugprone-macro-parentheses` opened as #4976. Turned out to have some genuine false positives among the 35 flagged locations: a handful of sites use the macro argument as a type name or an `offsetof()` member-designator rather than a value expression, where parenthesizing either doesn't compile or isn't valid. Verified one of those empirically with a standalone test file rather than assuming — details in the PR description. That's all three categories from your comment now covered.

flatcap:
I've made some minor tweaks and added a post-processing script...

- The action now runs weekly, Monday 4am and only demand.
  When the output is less verbose, we can add it back to PRs / `main`.

```diff
 on:
-  pull_request:
-    paths:
-      - '**/*.c'
-      - '**/*.h'
-  push:
-    branches:
-      - 'main'
+  schedule:
+    - cron: '10 4 * * 1'
+      # Mondays at 04:10
   workflow_dispatch:
```

- I've added a small script `.github/clang-tidy-summary.py` to generate two files:
  - `quickfix.txt` is just the warnings in Vim Quickfix format
  - `summary.md` is a count of warnings (like my summary, above)
    This is dumped into the build log.

- The Action now saves these extra files in the artifacts

If you're happy with the changes, I'll merge this PR and create a separate Issue for discussing the warnings.

chrisdebian:
Happy with these changes -- moving to a weekly cron plus the quickfix/summary artifacts is a good balance between visibility and noise. Please go ahead and merge, and thanks for following up on the warning categories.

flatcap (approved):

flatcap:
First run: #5004

PR #4974, opened 2026-07-29
bugprone: replace rewind()/setbuf() with checked equivalents

chrisdebian:
One of the three categories flatcap flagged as good candidates on #4970 (the clang-tidy CI PR): `bugprone-unsafe-functions`, mostly `rewind()`.

`rewind()` and `setbuf()` give no way to detect failure. Replaced with their exact behavioural equivalents:
- `rewind(fp)` → `fseek(fp, 0, SEEK_SET); clearerr(fp);` — `fseek()` alone doesn't clear the error indicator the way `rewind()` does, so both calls are needed for a faithful replacement, not just a syntactic swap.
- the one `setbuf(fp, NULL)` → `setvbuf(fp, NULL, _IONBF, 0)`.

Three sites (`nntp/nntp.c`, `ncrypt/crypt_gpgme.c`, `ncrypt/smime.c`) had the original `rewind()` as an unbraced single-statement `if`-body. Braces added where the two-statement replacement needed them — one of these (nntp.c) had a trailing `else` and would have failed to compile without the brace fix; the other two would have compiled fine but silently run the second statement unconditionally, regardless of the `if` condition. Found by checking every one of the 76 call sites individually against the line before it, not just relying on a clean compile.

Deliberately doesn't add explicit `fseek()`/`setvbuf()` return-value checking beyond what's already done at nearby call sites in this codebase — that's a separate, larger error-handling design question, not something this PR takes a position on.

**Verification**: `clang-tidy -p . <file> --checks='-*,bugprone-unsafe-functions'` now reports 0 warnings across all 76 sites (was 76). Full clean build (`make -j`), `make test` passes (no failures), `clang-format --dry-run -Werror` clean on every changed file.

**AI assistance**: the mechanical transform and the full 76-site brace-hazard review were done by Claude Code; I reviewed the diff and reasoning (especially the `clearerr()` behavioural-parity point and the three brace fixes) before this went up.

### DIFF
--- a/convert/content_info.c
+++ b/convert/content_info.c
@@ -272,7 +272,8 @@ struct Content *mutt_get_content_info(const char *fname, struct Body *b,
     slist_free(&chs);
   }
 
-  rewind(fp);
+  fseek(fp, 0, SEEK_SET);
+  clearerr(fp);
   while ((r = fread(buf, 1, sizeof(buf), fp)))
     mutt_update_content_info(info, &cstate, buf, r);
   mutt_update_content_info(info, &cstate, 0, 0);
--- a/convert/convert.c
+++ b/convert/convert.c
@@ -96,7 +96,8 @@ size_t mutt_convert_file_to(FILE *fp, const char *fromcode, struct Slist const *
     ni += 1;
   }
 
-  rewind(fp);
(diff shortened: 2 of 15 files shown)


flatcap (approved):
All good, thanks

PR #4975, opened 2026-07-29
bugprone: add default cases for switch-missing-default-case

chrisdebian:
The second of the three categories flatcap flagged as good candidates on #4970 (the clang-tidy CI PR) — `bugprone-switch-missing-default-case`, 55 sites across 34 files, matching his count exactly.

Each switch was read in full to determine the semantically correct default, not just silenced mechanically:

- Most are op/command dispatchers or menu-choice switches (`event->op`, `mw_multi_choice()` results, single-letter responses) where the existing code already has an implicit fallback after the switch (a later `return`, or falling through to shared code) — these get `default: break;` to make that explicit without changing behaviour.
- A few (`external.c`, `muttlib.c`, `send/send.c`) explicitly mirror the existing "abort" case's behaviour, since leaving the loop/function state unchanged on an impossible value would otherwise risk an infinite loop or a silently-wrong return.
- `pattern/exec.c`'s two large op-table dispatchers already had an explicit "unknown op" error path after the switch; the new default falls through to it, matching the existing intent exactly.
- A handful (`pop/auth.c`, `pop/lib.c`, `pop/pop.c`) switch on documented function return codes; checked each callee's own `@retval` doc comment to confirm every case was genuinely covered, and matched the default to whichever existing case already represents "no specific outcome" for that function.

No case does more than reproduce behaviour the code already had implicitly — this is a defensiveness/documentation change, not a behaviour change.

**Verification**: `clang-tidy -p . <file> --checks='-*,bugprone-switch-missing-default-case'` now reports 0 warnings across all 55 sites (was 55). Full clean build, `make test` passes (all suites), `clang-format --dry-run -Werror` clean on every changed file (one pre-existing, unrelated violation in `autocrypt/gpgme.c` confirmed via `git stash` before and after — not introduced by this change).

**AI assistance**: the mechanical location of all 55 sites and the semantic analysis of each switch (tracing callee return contracts, existing fallback behaviour, and loop-safety implications) was done by Claude Code; I reviewed the reasoning and diff before this went up.

### DIFF
--- a/alias/functions.c
+++ b/alias/functions.c
@@ -535,6 +535,8 @@ static int op_search(struct AliasFunctionData *fdata, const struct KeyEvent *eve
     case OP_SEARCH_OPPOSITE:
       flags |= SEARCH_OPPOSITE;
       break;
+    default:
+      break;
   }
 
   struct Menu *menu = mdata->menu;
@@ -588,6 +590,10 @@ static int op_sort(struct AliasFunctionData *fdata, const struct KeyEvent *event
 
     case 4: /* (u)nsorted */
       sort = ALIAS_SORT_UNSORTED;
+      break;
+
+    default:
+      resort = false;
       break;
   }
 
--- a/autocrypt/gpgme.c
+++ b/autocrypt/gpgme.c
@@ -305,6 +305,10 @@ int mutt_autocrypt_gpgme_select_or_create_key(struct Address *addr, struct Buffe
 
     case 1: /* create new */
       rc = mutt_autocrypt_gpgme_create_key(addr, keyid, keydata);
+      break;
+
+    default:
+      break;
   }
 
   return rc;
--- a/browser/dlg_browser.c
+++ b/browser/dlg_browser.c
@@ -950,6 +950,9 @@ void dlg_browser(struct Buffer *file, SelectFileFlags flags, struct Mailbox *m,
         case BROWSER_SORT_UNSORTED:
           browser_track = true;
           break;
+
+        default:
+          break;
       }
 
       /* We use mutt_browser_select_dir to initialize the two
--- a/browser/functions.c
+++ b/browser/functions.c
@@ -1262,6 +1262,10 @@ static int op_sort(struct BrowserPrivateData *priv, const struct KeyEvent *event
     case 7: /* do(n)'t sort */
       sort = BROWSER_SORT_UNSORTED;
       break;
+
+    default:
+      resort = false;
+      break;
   }
 
   if (!resort)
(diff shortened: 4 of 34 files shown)


flatcap:
hmm...
I've just skimmed through a lot of the changes and some of these look a bit off.
Perhaps it's late and my judgement's impaired.

Some `default`s just did a `break` when I might have expected it to `return` a failure.
I'll take a closer look in the morning.

As the insertions are strongly expected to **not** be hit, should we log, or assert?

chrisdebian:
Thanks for taking a look, and sorry for the ones that read oddly at a glance.

I went back through every `default` added in this PR against the function it sits in. Splitting them into two groups:

**Sites where a real failure return made sense**: these already return an explicit failure, not a silent `break`.
- `muttlib.c` `mutt_check_overwrite()` (both switches): `default: return -1;`, matching the documented `@retval -1 Abort`
- `external.c` `mutt_select_sort()`: `default: return false;`
- `imap/search.c` `compile_search_self()`: `default: return false;`
- `autocrypt/gpgme.c` `mutt_autocrypt_gpgme_select_or_create_key()`: `rc` is initialised to `-1` before the switch, so its `default: break;` already returns the documented error value

**Everything else** is a dialog/menu-choice switch (`mw_multi_choice()` results, key-flag parsing, sort-order selection, etc.) where the surrounding code already treats an unmatched value as a no-op, and the new `default: break;` just makes that explicit for the compiler rather than changing behaviour.

I couldn't find a site where `break` is silently swallowing something that should have been a failure. If you spot one when you look again, could you point me at the file and line? Happy to fix it properly.

On logging vs asserting: I'd suggest not applying either uniformly. Most of these switches are over values from `mw_multi_choice()` or similar bounded, non-enum inputs, where an unmatched value isn't really "shouldn't happen" in the assertable sense, it's just outside the choices you offered, so a quiet no-op is correct and a log line would just be noise on every run of NeoMutt for a case that's aesthetically defensive rather than truly exceptional. Where the switch is over a genuinely closed enum, `pattern/exec.c`'s pattern-operator switch is the one case I saw that fits, but it already falls through to an existing `mutt_error(_("error: unknown op %d (report this error)")); return false;` right below the switch, so that path's covered too.

If there's a specific site you want a log or assert added to, let me know which and I'll add it.

flatcap:
Thanks for re-checking -- late night code review probably isn't a good idea :-)
I'll have to postpone my second look 'til later.  Sorry.

Currently there are two new, keen, NeoMutt contributors.
You and alonaoum, who I've neglected for a few days.
I've got to catch up on my "Function Renaming" tasks.

flatcap (approved):

PR #4976, opened 2026-07-29
bugprone: parenthesize macro arguments/replacement lists

chrisdebian:
The last of the three categories flatcap flagged as good candidates on #4970 — `bugprone-macro-parentheses`. 28 sites across 13 files fixed; 7 confirmed as genuine false positives and deliberately left untouched (see below).

Every site was checked against its actual call sites for a live bug, not just mechanically wrapped:

- Simple negative-constant macros (`COLOR_DEFAULT`, `MUTT_WIN_SIZE_UNLIMITED`, `OP_REPAINT`/`TIMEOUT`/`ABORT`, `IMAP_RES_NO`/`BAD`, `MUTT_MAXRANGE`) get their replacement list wrapped.
- Value-expression arguments used unparenthesized next to an operator (`config/types.h`'s `IS_MAILBOX`/`IS_COMMAND`, `mutt/array.h`'s `ARRAY_GET`, `mutt/mbyte.h`'s `IsWPrint`/`IsBOM`, `notmuch/private.h`'s `LIBNOTMUCH_CHECK_VERSION`) get the argument wrapped — no live caller currently passes a compound expression, but this is exactly the class of bug that bites the next caller who does.
- `mutt/string2.h`'s `SKIPWS` and `mutt/array.h`'s `ARRAY_FOREACH_(REVERSE_)FROM_TO` wrap the argument even though it's used as an assignment/increment target, since `(x)++`/`(x) = ...` are equivalent to the unparenthesized form for any valid lvalue — harmless, satisfies the check.

Three false-positive patterns confirmed and left alone, since parenthesizing would either not compile or change meaning:

- `mutt/atoi.h:39` and `mutt/memory.h:57`: the macro argument is a **type name** (a function parameter's type, and a `_Generic` type-association respectively), not a value expression — wrapping a type name in parens isn't valid there.
- `mutt/array.h:52`: same class — `T` is the array element's type name in a struct member declaration.
- `mutt/queue.h:437`/`632`/`887`/`896` (all four remaining sites in that file): `field` is passed as a **member-designator** into `__containerof()` (built on `offsetof()`), not a value expression. Verified empirically with a standalone test file: parenthesizing a member-designator argument to `offsetof()` is a genuine compile error ("expected identifier"), not just a style question.

**Verification**: `clang-tidy -p . <file> --checks='-*,bugprone-macro-parentheses' --header-filter='.*'` now reports only the 7 confirmed-false-positive sites above (was 28 fixable + 7 false positives = 35 total unique locations). Full clean build, `make test` passes (all suites), `clang-format --dry-run -Werror` shows no *new* violations in any changed file (checked differentially against each file's pre-existing state, since several of these headers already had unrelated formatting drift).

**AI assistance**: the mechanical location of all sites, the call-site tracing to distinguish live risk from defensive-only fixes, and the identification + empirical verification of the false positives was done by Claude Code; I reviewed the reasoning and diff, including the compile-error test for the `offsetof()` case, before this went up.

### DIFF
--- a/color/color.h
+++ b/color/color.h
@@ -99,7 +99,7 @@ enum ColorId
 
 extern const struct Mapping ColorFields[];
 
-#define COLOR_DEFAULT -1
+#define COLOR_DEFAULT (-1)
 
 struct ColorModuleData;
 
--- a/compress/private.h
+++ b/compress/private.h
@@ -26,8 +26,8 @@
 #define COMPRESS_OPS(_name, _min_level, _max_level) \
   const struct ComprOps compr_##_name##_ops = {     \
     .name       = #_name,                           \
-    .min_level  = _min_level,                       \
-    .max_level  = _max_level,                       \
+    .min_level  = (_min_level),                     \
+    .max_level  = (_max_level),                     \
     .open       = compr_##_name##_open,             \
     .compress   = compr_##_name##_compress,         \
     .decompress = compr_##_name##_decompress,       \
--- a/config/types.h
+++ b/config/types.h
@@ -119,7 +119,7 @@ enum ConfigTypeField
 #define D_SORT_LAST                D_CUSTOM_BIT_0                      ///< Sort flag for -last prefix
 #define D_SORT_REVERSE             D_CUSTOM_BIT_1                      ///< Sort flag for -reverse prefix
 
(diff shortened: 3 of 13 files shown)


flatcap (approved):
All good, I just tweaked the alignment of a couple of comments

PR #4985, opened 2026-08-03
gui: add mview_email_at/mview_email_count accessors

chrisdebian:
Small PR with the accessor + 2 migrated call sites as the initial prototype for step 1 of the Branch-by-Abstraction plan from #4977 -- happy to follow up with the rest of the `index/` call sites in separate PRs once you've had a look at the pattern.

## What's here

- `mview_email_count()` / `mview_email_at()` in `gui/mview.c`/`gui/mview.h`, mirroring `mutt_get_virt_email()`'s existing NULL/bounds-check style exactly.
- Two call sites migrated as the proof of the pattern: `subjectrx_clear_mods()` (a plain loop) and `op_jump()` (a single-index lookup) -- the simplest example of each shape, deliberately not a full migration.

## A build-system wrinkle found along the way

Migrating even these two sites forces the real `gui/mview.o` into `test/neomutt-test`'s link for the first time, since `test/common.c`/`test/pattern/dummy.c` stub every other public `mview.c` function specifically so the real object (which pulls in imap/ncrypt/pattern/mx) never needs linking. Added matching stubs for the two new functions to `test/common.c` -- without them, `make test` fails to link. No new unit tests for the two functions themselves in this PR, since exercising the real (non-stubbed) implementation would reintroduce the same conflict; flagging that as a separate decision about the stub/dummy test architecture rather than folding it in here.

## Verification

- Full clean build, zero new warnings.
- `test/neomutt-test` links cleanly (confirmed it doesn't without the `test/common.c` addition above).
- Manually exercised both migrated call sites against a scratch Maildir and confirmed byte-for-byte identical behaviour against an unmodified `main` build with the same setup.

AI assistance: implementation, codebase research, and verification done by Claude Code; reviewed by me before opening this PR.

### DIFF
--- a/gui/mview.c
+++ b/gui/mview.c
@@ -366,6 +366,46 @@ bool message_is_tagged(struct Email *e)
 }
 
 /**
+ * mview_email_count - Number of Emails in a MailboxView
+ * @param mv MailboxView
+ * @retval num Number of Emails
+ *
+ * This safely gets the result of the following:
+ * - `mv->mailbox->msg_count`
+ */
+int mview_email_count(const struct MailboxView *mv)
+{
+  if (!mv || !mv->mailbox)
+    return 0;
+
+  return mv->mailbox->msg_count;
+}
+
+/**
+ * mview_email_at - Get an Email by physical (insertion-order) index
+ * @param mv  MailboxView
+ * @param num Physical index number (0-based; matches Email::msgno)
+ * @retval ptr  Email
+ * @retval NULL No Email selected, or bad index values
+ *
+ * This safely gets the result of the following:
+ * - `mv->mailbox->emails[num]`
+ *
+ * @note This is physical/insertion order, NOT display/sort order.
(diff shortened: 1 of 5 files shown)


flatcap:
That's a nice start!
(I was a bit worried that your first PR would be a monster! :-)

I've just got to decide what to _do_ with your improvements.
I'm planning to make a Devel Release as soon as I vanquish alonaoum's mountain of Function Renames.

I've created a branch [chrisdebian/views] (currently at `main`).
We can use that for staging your changes.

You can set the **base** branch when you create a PR, or later by editing the PR's title. 

> happy to follow up with the rest of the `index/` call sites in separate PRs

Yes, please

flatcap:
> A build-system wrinkle found along the way

hehe, yeah linking the tests is a challenge.
I've chosen the lazy path, of ~70 function stubs in `test/common.c`
Each, is a reminder of a dependency problem that needs fixing.
Some day :-)

flatcap (approved):

PR #4998, opened 2026-08-09
Add Background Job Functionality

fishman:
- **feat: add background command execution**
- **fix(bg): re-run last command on empty input**
- **feat(background): add background-wait command**
- **feat(background): make bg_wait() non-blocking**
- **fix: give the menu ownership of the background dialog's mapping**
- **build: add test/background to BUILD_DIRS**

- **What does this PR do?**
  It adds functionality for background jobs. It also allows for a background-wait functionality that allows me to schedule a command asynchronously once the job has been completed.

image

Yes, but not sure about the style guide for the new files (https://neomutt.org/dev/code is a dead link)

### DIFF
--- a/background.c
+++ b/background.c
@@ -32,6 +32,7 @@
 #include <stdbool.h>
 #include <stdio.h>
 #include <stdlib.h>
+#include <string.h>
 #include <sys/stat.h>
 #include <sys/wait.h>
 #include <unistd.h>
@@ -44,10 +45,14 @@
 #include "history/lib.h"
 #include "key/lib.h"
 #include "menu/lib.h"
+#include "muttlib.h"
 #include "pager/lib.h"
 
-/// Maximum number of jobs kept at once; oldest finished slot is recycled
+/// Maximum number of jobs kept at once; Clear frees the finished ones
 #define MAX_JOBS 10
+
+/// Maximum number of job outputs kept in memory
+#define BG_OUTPUT_RING 10
 
 /**
  * struct BackgroundJob - A command running in the background
@@ -58,17 +63,76 @@ struct BackgroundJob
   bool running;        ///< true if the process is still running
   int exit_code;       ///< Exit code, -1 if still running or killed by a signal
   struct Buffer *cmd;  ///< Command line
-  struct Buffer *file; ///< Temp file with captured output
+  struct Buffer *file; ///< Temp file with captured output while running
+  struct Buffer *out;  ///< Output kept in memory after completion
 };
 
 static struct BackgroundJob Jobs[MAX_JOBS] = { 0 };
 static struct Buffer *LastCommand = NULL;
+
+/// Finished jobs with output in memory, oldest first; evicted when full
+ARRAY_HEAD(OutputRing, struct BackgroundJob *);
+static struct OutputRing OutputRing = ARRAY_HEAD_INITIALIZER;
 
 /// Events of a macro paused by bg_wait(), resumed when the jobs finish
 static struct KeyEventArray Parked = ARRAY_HEAD_INITIALIZER;
 static bool ParkedActive = false;
 
 // -----------------------------------------------------------------------------
+
+/**
+ * output_ring_remove - Drop one output, shifting the rest
+ * @param pos Position in the ring
+ */
+static void output_ring_remove(int pos)
+{
+  struct BackgroundJob **job = ARRAY_GET(&OutputRing, pos);
+  buf_free(&(*job)->out);
+  const int n = ARRAY_SIZE(&OutputRing) - pos - 1;
+  if (n > 0)
+    memmove(job, job + 1, n * sizeof(struct BackgroundJob *));
+  ARRAY_SHRINK(&OutputRing, 1);
+}
+
+/**
+ * output_ring_remove_job - Drop a job's output from the ring
+ * @param job Job to remove
+ */
+static void output_ring_remove_job(struct BackgroundJob *job)
+{
+  for (size_t i = 0; i < ARRAY_SIZE(&OutputRing); i++)
+  {
+    if (*ARRAY_GET(&OutputRing, i) == job)
+    {
+      output_ring_remove(i);
+      return;
+    }
+  }
+}
+
+/**
+ * output_ring_add - Keep a finished job's output, evicting the oldest if needed
+ * @param job Job whose output was just copied to memory
+ */
+static void output_ring_add(struct BackgroundJob *job)
+{
+  ARRAY_ADD(&OutputRing, job);
+  if (ARRAY_SIZE(&OutputRing) > BG_OUTPUT_RING)
+    output_ring_remove(0);
+}
+
+/**
+ * output_ring_clear - Drop all outputs
+ */
+static void output_ring_clear(void)
+{
+  struct BackgroundJob **job = NULL;
+  ARRAY_FOREACH(job, &OutputRing)
+  {
+    buf_free(&(*job)->out);
+  }
+  ARRAY_FREE(&OutputRing);
+}
 
 /**
  * job_slot_free - Release a job slot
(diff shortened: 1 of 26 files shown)


fishman:
@flatcap sorry I didn't want to start a discussion when the CI wasn't running, and then it wasn't possible to reopen the PR.

fishman:
I also went ahead to check what's necessary to extend this for implementing an async send comand. It seems doable, but i'll hold off until some variant of async jobs are added. I guess there are some quirks about how mutt works in general that need to be paid attention to.

flatcap:
> I didn't want to start a discussion when the CI wasn't running

As a first-time contributor, I had to OK the PR before the CI would start

> and then it wasn't possible to reopen the PR.

hehe, GitHub, eh :-)

---

I've only had a quick play around with it so far.
It starts tasks, I can list them and look at some output.
Once.

After I've viewed the output, I can't seem to see it again, even if the process is still running and outputting.

The code looks pretty tidy, thanks.

Did you use AI to assist you?
(few people think to add a full test suite ;-)
It's OK if you did, I'd just like to know.

Food.  More comments later

flatcap:
When I quit the "output" screen, it drops back to the Index.
I expected to go back to the Background Dialog.

The Background Dialog doesn't support global commands.
i.e. I can't <kbd>:</kbd> (`<enter-command>`)

`<background-wait>` doesn't seem to **wait** for anything.

I've still only skimmed through the code.
I'll try to find some time tomorrow.

---

Future ideas (no pressure, just stuff that occurred to me :-)

- `background` subdir:
  - Background code as a library
  - GUI code separate
- Configurable Display - config expando format string
- Allow the user to delete/clear old jobs
- Allow the user to save the output

fishman:
> Did you use AI to assist you?
> (few people think to add a full test suite ;-)

Yes, the style is part of my style that I tend to request, and then there are the usual issues where AI names inconsistently with the surrounding style like, output_dlg_open instead  of dlg_output. And then I double check for leaks, since that would be a nightmare for long running terminal program like mutt.

>  It starts tasks, I can list them and look at some output. Once.

I guess a ring buffer makes sense for it.

> <background-wait> doesn't seem to wait for anything.

This was a bit tedious. What it does is the following:

`macro index O "<background-command>mbsync -a <enter><background-wait><change-folder>inbox<enter>" "full sync"`

if it will wait for mbsync -a to finish and then execute the change-folder.

fishman:
> The Background Dialog doesn't support global commands.

simple dialogs use menu_function_dispatcher, I don't think we should change it.

flatcap:
Don't worry about the failed "Clang Format" check -- it's a new Action that we're experimenting with.

fishman:
> Don't worry about the failed "Clang Format" check -- it's a new Action that we're experimenting with.

I ran it at first. The problem is that some files do not conform to the clang format output. Maybe the rules have to be adjusted. It will create commit noise especially for those column aligned header files

flatcap:
> I ran it at first

thanks

> column aligned header files

Yes, we've added lots of whitespace in the headers to improve readability, so we don't (often) format them.

> some files do not conform to the clang format output

It's a great tool, but every version produces _slightly_ different results.

The canonical version is: clang-format **v22.1.8**
(that's the version on my laptop :-)

Once a month, or so, I'll get bored and clang-format the entire source tree to keep it tidy.

flatcap:
## Splitting `background.c`

I think now might be the time to split up `background.c`
Create a `background` dir with:
- `dlg_background.c` - Dialog
- `background.c` - Background-y stuff
- `functions.c` - Custom functions

`Makefile.autosetup` is fairly modular, so this should be a quick copy'n'paste.
The `mlist/` library is probably a good starting point.
It's a Simple Dialog displaying mailing list actions, with a few custom functions.

## Global (generic) Functions

> > The Background Dialog doesn't support global functions.
> 
> simple dialogs use menu_function_dispatcher, I don't think we should change it.

You've added your background functions to `OpGeneric` (the 'generic' map).
This is good for the **global** functions.

You've also added some new default bindings to `set_default_bindings()`.
That's not so good.  They're used by **every** dialog.

The 'generic' map defines the functions that are common, such as selection, scrolling and global actions.
Anything else, we want to layer on top.

## Layering Functions

At startup `main()` calls `mlist_init_keys()`.
It defines a new keymap, called 'list' and populates it with functions from `OpList`.
Next, it adds the 'generic` functions.
Finally, it adds some keybindings, from `ListDefaultBindings`.

This creates a `MenuDefinition`, which we'll use later.

## Running the Dialog

When `dlg_mlist()` is called, it sets up its data, then creates a Simple Dialog (`simple_dialog_new()`).

In the Event Loop, it calls `km_dokey()` using the `MenuDefintion` from earlier.
That contains all the possible functions and any keybindings to those functions.

Next it calls **function dispatchers** to do the actual work.
It calls then in order:
- `mlist_function_dispatcher()` - mlist's custom functions
- `menu_function_dispatcher()` - general motion functions, etc
- `global_function_dispatcher()` - global functions

This dialog also supports tagging, so it calls `menu_tagging_dispatcher()`.                    

Each set of functions has its own function dispatcher.

---

That's a lot to take in, so have a look around and ask lots of questions.

background.c
```diff
@@ -0,0 +1,720 @@
...
+  {
+    goto done;
+  }
+
+  if (buf_is_empty(cmd)) // re-run the last command
+  {
+    if (!LastCommand)
+      goto done;
+    buf_copy(cmd, LastCommand);
+  }
```
gahr: I think this is very error prone.

fishman: You're right, dropping it.

fishman: But basically the opdispatcher deletes the message. So you have to redisplay the message. If you don't save it there during bg_wait you don't know what command was run

gahr:
I have a few comments:

1. if it wasn't for bg_wait, I would say launching background jobs doesn't really belong to NeoMutt, so to me the only relevant value is to be able to run commands once the job has finished
2. because of that, I think *multiple* background jobs is a useless complication: you can't *handle* any but the last submitted job
3. also because of 1., perhaps we want to merge the spawn and wait commands, and provide that takes two arguments: 1) the shell command to spawn, and 2) the NeoMutt command to execute once the shell command is done

flatcap:
I've just has a worrying thought about the wait command -- scope.

Steps:
1. Launch slow command
2. Queue wait macro
3. Command finishes, macro executes

If the user changes screen between 2 & 3, we could be in trouble.
At best, the macro will fail.
At worst, it could cause damage.

_Thinking..._

gahr:
> I've just has a worrying thought about the wait command -- scope.

Oh wow I hadn't thought about that.. that's a major inconvenience...

flatcap:
## Limitations

It's most likely that the user wants to act in the Index Dialog,
so we might want to restrict the macro to **only** the Index.

If the user only wants to `<sync-mailbox>`, then there should be no problem.
Same for other global functions, `<redraw-screen>`, `<enter-command>`

## Questions

**Who should queue the macro?**

- Key code
- Index Dialog

I think it's **got**  to be the Index.
The key code shouldn't be aware of its callers.

**When should it fire?**

1. After the command finishes, obviously
   (observes some event)
2. When the user is idle
   (probably a timeout)

**What state should the Index be in (pre/post)?**

Should the Index try to save/restore its start before/after the macro?
HOW?

## More problems

- What if the user tags some emails, and the macro fires?
- What if the user is in a second Index, e.g. Compose -\> Attach Message?
  (this is an independent instance of the Index)

fishman:
I have a patch queued for async send based on this. As for the queued macro. To be honest I never thought about using this outside the index. But yes the documentation should reflect potential issues, but it's not different from running commands manually while doing other operations.

fishman:
> What if the user tags some emails, and the macro fires?

Why would this be an issue? How is it different from when I do this manually?

gahr:
If the only use case is to sync the mailbox after fetching emails, then perhaps we should support it first-class, i.e., a command that *internally* executes some external command and sync the mailbox when it's done... I don't think there's much more that you'd want to do, and that you could safely do.

gahr:
/me proposes:

```
set external_sync_command = mbsync     # run mbsync to, well, sync...
set external_sync_timeout = 100        # run that every 100 seconds
```

flatcap:
> > What if the user tags some emails, and the macro fires?
> 
> Why would this be an issue?

It's _unlikely_ be a problem, but if the macro is allowed to do anything,
then the tagged state _could_ affect the behaviour.

fishman:
> /me proposes:
> 
> 
> 
> ```
> 
> set external_sync_command = mbsync     # run mbsync to, well, sync...
> 
> set external_sync_timeout = 100        # run that every 100 seconds
> 
> ```

Because this isn't how people use it. Maybe that's how you use your mail. But it fundamentally does not fit the purpose of neomutt which is very much built on multi step workflows. You're trying to protect from something that you shouldn't be protecting from. Fundamentally the reason for background tasks is that you want to do something and get feedback for it.

fishman:
> If the only use case is to sync the mailbox after fetching emails, then perhaps we should support it first-class, i.e., a command that *internally* executes some external command and sync the mailbox when it's done... I don't think there's much more that you'd want to do, and that you could safely do.

What is this? Completely misunderstand the feature. Make up unrelated issues which exist which manual behavior already and then propose a solution that implements the same backend with way worse usability how  is that of any value?

flatcap:
> `set external_sync_command = mbsync`

hmm... ok

Ideally, we'd still want to do this asynchronously -- it could take time.
And check that it succeeded before running `<sync-mailbox>`.

gahr:
> > If the only use case is to sync the mailbox after fetching emails, then perhaps we should support it first-class, i.e., a command that _internally_ executes some external command and sync the mailbox when it's done... I don't think there's much more that you'd want to do, and that you could safely do.
> 
> What is this? Completely misunderstand the feature. Make up unrelated issues which exist which manual behavior already and then propose a solution that implements the same backend with way worse usability how is that of any value?

I'm sorry if I misunderstood the feature. Perhaps you could elaborate the problem it solves? The only example you provided was to run mbsync and sync-mailbox once it's done. What else would you want to be able to do?

gahr:
> Ideally, we'd still want to do this asynchronously -- it could take time.

Yes, I didn't write it, but the idea was that the new setting would run the command asynchronously.

gahr:
@fishman what's the use case for wanting multiple background jobs running, and only being able to react when the last posted one finishes? (If I read the code right, the trigger is when the last posted job finishes, even if there are still jobs running, is that right?)

flatcap:
Sorry, gahr and I got a bit distracted there.
We were focussing on the practical problems.

To help you move forward we need to know...

## Questions

What command is being run, each time?
- Same command
- Different command

When is the command run?
- Manually
- Timer

What do **you** **need** from the payload of the 'wait' macro?
- `<sync-mailbox>` - Standard refresh
- `<change-folder>^<enter>` - Forced refresh
- Any NeoMutt macro

What is the 'wait' command waiting for?
- Latest command to finish
- All commands to finish

fishman:
@gahr do you have any feedback about the code? You said you had concerns about the code execution, what are those?

gahr:
> @gahr do you have any feedback about the code? You said you had concerns about the code execution, what are those?

I don't remember mentioning concerns about the code execution, what are you referring to?

i haven't carefully read the code yet, I was focussing on the functionalities offered by the feature and was trying to understand (and failed at that, according to [your previous comment](https://github.com/neomutt/neomutt/pull/4998#issuecomment-5253137718)) the use cases.

It is sure a decent amount of non-trivial code. I'd like to understand whether the generality is needed to address *your* immediate problem.

flatcap:
^^^^ Questions ^^^^

@fishman 
Reza, you have a clear picture, in your head, of what you want.
We don't, yet.

fishman:
@flatcap i think you have two kinds of people in your developer community. One is people like you that are curious about fixing architectural shortcomings of mutt and notmuch interaction, like yourself. And others who want to keep everything the way it is, in which case i don't really know why they joined neomutt to begin with.

fishman:
@flatcap he was right, async doesn't belong into neomutt. I have built a new mail client with proper async support that addresses all the issues I had mentioned

https://github.com/fishman/notmutt

PR #5009, opened 2026-08-10
fix invalid-shift(UBSan) in `mutt_decode_base64()`

### LINKED ISSUE #5008: Left shift of negative base64 digit in `mutt_decode_base64()` when `=` appears in leading positions
## Summary

`mutt_decode_base64()` reads four base64 characters into `buf[0..3]` and decodes them. Its fill filter (`email/handler.c:1589`) admits the `=` padding character at **any** position via `(base64val(ch) != -1) || (ch == '=')`, but the decode logic only special-cases `=` at positions 2 (`handler.c:1617`) and 3 (`handler.c:1633`). At positions 0 and 1 the code calls `base64val('=')` directly — which returns `-1` (the `Index64[61]` sentinel for "not a base64 digit") — and then evaluates `ch = (c1 << 2) | (c2 >> 4)` at `handler.c:1604`, performing a left shift of the negative value `-1`. Left-shifting a negative signed integer is undefined behavior (C11 §6.5.7), which UBSan reports as `left shift of negative value`. The sibling standalone decoder in `mutt/base64.c` validates `base64val(d) == BAD` *before* any shift (`base64.c:190,193,200,203`), so the library is internally inconsistent — `mutt_decode_base64()` is simply missing the validation its counterpart performs.

## NeoMutt Version

```
NeoMutt 20251211
```

## Expected Behaviour

Decoding a base64 stream with = padding in a leading position (slots 0/1, which is malformed per RFC 4648) should stop decoding cleanly without invoking undefined behavior. The sibling decoder mutt_b64_decode() (mutt/base64.c) already rejects = at leading positions by validating base64val(d) == BAD before shifting; mutt_decode_base64() should do the same.

## Actual Behaviour

With UBSan enabled, mutt_decode_base64() aborts: email/handler.c:1604:14: runtime error: left shift of negative value -1. Without sanitizers, the shift wraps (implementation-defined) and the function silently emits incorrect decoded bytes from malformed padding.

No coredump under a non-sanitized build (it's UB-on-arithmetic, not a memory fault).

### Stack Trace
```
email/handler.c:1604:14: runtime error: left shift of negative value -1
    #0 0x5fd641978e22 in mutt_decode_base64 /tmp/neomutt_build/email/handler.c:1604:14
    #1 0x5fd641877a61 in main poc.cpp:29:5
    #2 0x770cca4d51c9 in __libc_start_call_main csu/../sysdeps/nptl/libc_start_call_main.h:58:16
    #3 0x770cca4d528a in __libc_start_main csu/../csu/libc-start.c:360:3
    #4 0x5fd64178cb04 in _start (poc+0x3bdb04) (BuildId: d0361a893f7ec07e0bae5fbcba9fa03c840e1b02)

SUMMARY: UndefinedBehaviorSanitizer: undefined-behavior email/handler.c:1604:14 
```

### PoC Code

```cpp
#include <cstdio>
#include <cstring>

extern "C" {
#include "config.h"
#include "mutt/lib.h"   /* struct State, MuttLogger, log_disp_null, ICONV_T_INVALID */
#include "config/lib.h" /* defines enums email/lib.h forward-references under C linkage */
#include "core/lib.h"
#include "email/lib.h"  /* mutt_decode_base64() */
}

bool StartupComplete = true;

int main()
{
    MuttLogger = log_disp_null; /* sink any mutt_debug() output */

    static const char data[] = "=utf"; /* buf[0]='=' -> c1=-1 -> (-1)<<2 UB */
    FILE *in = fmemopen((void *) data, sizeof(data) - 1, "r");
    FILE *out = fopen("/dev/null", "w");
    if (!in || !out) { if (in) fclose(in); if (out) fclose(out); return 1; }

    struct State state;
    memset(&state, 0, sizeof(state));
    state.fp_in = in;
    state.fp_out = out;

    /* istext=false: skip state_set_prefix; cd=invalid: skip the iconv path. */
    mutt_decode_base64(&state, sizeof(data) - 1, false, ICONV_T_INVALID);

    fclose(in);
    fclose(out);
    return 0;
}
```

### Reproduction Step

```bash
# 1. Checkout NeoMutt at the version above
git clone https://github.com/neomutt/neomutt.git
cd neomutt
git checkout fb8e54b69cff6c7257d348122d19459cba5539bc

# 2. Build the library objects with ASan+UBSan.
export CC=clang CXX=clang++
export CFLAGS="-fsanitize=address,undefined -g -O0 -fno-omit-frame-pointer -w -fPIC"
export CXXFLAGS="$CFLAGS" LDFLAGS="$CFLAGS"
./configure --disable-doc --disable-nls --disable-pgp --disable-smime \
            --disable-inotify --disable-idn2 CC="$CC" CXX="$CXX"
make -j"$(nproc)"   # build NeoMutt and the lib*.a archives used in the link step below

# 3. Build the standalone PoC (paste the PoC from the section above into poc.cpp)
clang++ -g -O0 -fsanitize=address,undefined -fno-omit-frame-pointer \
    -I. \
    poc.cpp -o poc \
    -Wl,--start-group lib*.a -Wl,--end-group -lncursesw -ltinfo

# 4. Run
./poc
```

- Always, the PoC reproduces on every run.

Not upgrade/config-related. This is a pre-existing library defect in email/handler.c; the = filter has admitted = at all slots for as long as the fill loop has had the || (ch == '=') clause, but the decode logic only ever guarded slots 2/3. Discovered via fuzzing, not from a behavior change.

## Submission Statement

This report was produced by FuzzAnything's AI-assisted library fuzzer and manually verified by a team member. We reviewed the PoC against the upstream API documentation — call order, parameters, and memory ownership — and found no API misuse.

> Signed-off-by: FuzzAnything <fuzzanything@gmail.com>

yvonnelxxxx:
This PR attempts to fix [#5008](https://github.com/neomutt/neomutt/issues/5008#issue-5106624064)

## Problem

`mutt_decode_base64()` (`email/handler.c`) decodes base64 in 4-character quads. Its fill loop admits the `=` padding character at **any** buffer slot via the clause `(base64val(ch) != -1) || (ch == '=')` (`handler.c:1589`), but the decode logic only special-cases `=` at slots 2 (`:1617`) and 3 (`:1633`). At slots 0 and 1 the code calls `base64val('=')` directly, which returns the sentinel `-1` (`Index64[61] == -1`, `mutt/base64.c`), and then evaluates:

```c
const int c1 = base64val(buf[0]);   // '=' -> -1
const int c2 = base64val(buf[1]);
ch = (c1 << 2) | (c2 >> 4);         // :1604 — (-1) << 2 == UB
```

Left-shifting a negative signed value is undefined behavior (C11 §6.5.7). Under `-fsanitize=undefined` this is reported as `left shift of negative value` and aborts; on a non-sanitized two's-complement build it wraps without trapping and the function silently emits wrong decoded bytes from malformed padding.

The trigger is any base64 body with `=` padding at an invalid (leading) position (e.g. the 4-byte input `"=utf"`), reachable through the core MIME base64 decode path with no privileges required. 
## Fix

Extend the existing `=` guard (already present at slots 2 and 3) to slots 0 and 1, so a `=` (or any non-base64 byte) at a leading position stops decoding instead of being shifted. This mirrors the guarded standalone decoder in `mutt/base64.c`, which validates `base64val(d) == BAD` before every shift.

```diff
--- a/email/handler.c
+++ b/email/handler.c
@@ -1601,7 +1601,11 @@ void mutt_decode_base64(struct State *state, size_t len, bool istext, iconv_t cd)

     const int c1 = base64val(buf[0]);
+    if (c1 == -1)          /* '=' / non-base64 at slot 0: stop */
+      break;
     const int c2 = base64val(buf[1]);
+    if (c2 == -1)          /* '=' at slot 1: invalid padding, stop */
+      break;

     /* first char */
     ch = (c1 << 2) | (c2 >> 4);
```

`=` at slot 0 or 1 is structurally malformed under RFC 4648 (a single leading data char carries only 6 bits and cannot yield an output byte, and interior padding is forbidden), so `break` — the same action the decoder already takes for valid tail padding at slots 2/3 and for a short final quad (`if (i != 4) break;`, `:1592`) — is the spec-correct response. No signature or return-value change is needed (`mutt_decode_base64` returns `void`), and the guards never fire on well-formed input, so `XXXX` / `XXX=` / `XX==` all decode unchanged.

## Verification

1. **Reproduce (pre-fix):** build NeoMutt with `-fsanitize=address,undefined`, compile `poc.cpp` against `lib*.a` (`-lncursesw -ltinfo`), run `./poc` → UBSan reports `left shift of negative value -1` at `handler.c:1604`.
2. **Fixed:** apply the patch, rebuild, recompile, re-run `./poc` → clean exit, no UBSan report (3/3 deterministic clean runs). The PoC now hits `if (c1 == -1) break;` on the first quad and returns without shifting.
3. **Regression — valid base64:** feed `SGVsbG8=` (`Hello`) and a full quad (`dGVzdA==`, `test`) through the same harness; decoded output is byte-identical to the pre-fix build (the new guards never fire on well-formed input).
4. **Regression — tail padding:** confirm `XX==` and `XXX=` still yield 1 and 2 output bytes respectively (the existing `:1617`/`:1633` breaks are untouched).

## Submission Statement

This fix was generated using Claude Code and manually reviewed, tested, and verified by a team member.

Signed-off-by: FuzzAnything <fuzzanything@gmail.com>

### DIFF
--- a/email/handler.c
+++ b/email/handler.c
@@ -1601,7 +1601,11 @@ void mutt_decode_base64(struct State *state, size_t len, bool istext, iconv_t cd
     }
 
     const int c1 = base64val(buf[0]);
+    if (c1 == -1) /* '=' (or any non-base64 char) at slot 0: stop */
+      break;
     const int c2 = base64val(buf[1]);
+    if (c2 == -1) /* '=' at slot 1: invalid padding, stop */
+      break;
 
     /* first char */
     ch = (c1 << 2) | (c2 >> 4);


gahr (approved):

flatcap:
(clang-format, no code change)

PR #5011, opened 2026-08-10
fix: null-pointer dereference in mutt_replacelist_match()

### LINKED ISSUE #5010: Null-pointer dereference in `mutt_replacelist_match()` when an empty template string is stored
## Summary

When `mutt_replacelist_add()` is given a non-NULL but empty template string (`""`), it stores the result of `mutt_str_dup("")` — which is `NULL` — into `np->templ`, while still inserting the node into the `ReplaceList`. A subsequent `mutt_replacelist_match()` that finds a regex match then dereferences `np->templ` at `mutt/regex.c:508` (`for (p = np->templ; *p && ...)`) with no NULL guard, producing a UBSan null-pointer load and an AddressSanitizer SEGV (read of address `0x0`). The sibling function `mutt_replacelist_apply()` guards the same substitution loop with `if (np->templ)` at `mutt/regex.c:400`, so the two functions handle a NULL `templ` pointer inconsistently.

## NeoMutt Version

```
NeoMutt 20251211
```

## Expected Behaviour

`mutt_replacelist_add()` should reject an empty replacement template `""` the same way it rejects a `NULL` one, and `mutt_replacelist_match()` should never dereference `np->templ` without a NULL guard. A `spam`/`subjectrx` rule with an empty template should either be accepted as a no-op or rejected with an error — not silently create a half-initialized list node that crashes on the next match.

## Actual Behaviour

NeoMutt crashes with a null-pointer dereference (SEGV, read of address 0x0) in `mutt_replacelist_match()` at 
```cpp
mutt/regex.c:508: 
for (p = np->templ; *p && (tlen < (buflen - 1));) 
```
`mutt_replacelist_add()` only NULL-checks templ, not `*templ == '\0'`. `mutt_str_dup("")` returns `NULL`, so an empty template leaves `np->templ == NULL` on a node that's still added to the list. The first matching input then dereferences `NULL`. UBSan reports load of null pointer of type 'char'; ASan reports SEGV on unknown address 0x000000000000.

The sibling `mutt_replacelist_apply()` already guards the identical loop with `if (np->templ)` at `mutt/regex.c:400`, `match` is missing the check `apply` has.

No coredump was collected. Reproduction is via a sanitizer build (ASan+UBSan), which aborts and prints the stack trace directly. 

### Stack Trace
``` bash
mutt/regex.c:508:27: runtime error: load of null pointer of type 'char'
SUMMARY: UndefinedBehaviorSanitizer: undefined-behavior mutt/regex.c:508:27 
AddressSanitizer:DEADLYSIGNAL
=================================================================
==103586==ERROR: AddressSanitizer: SEGV on unknown address 0x000000000000 (pc 0x5cc855c34872 bp 0x7fff5f4625f0 sp 0x7fff5f462500 T0)
==103586==The signal is caused by a READ memory access.
==103586==Hint: address points to the zero page.
    #0 0x5cc855c34872 in mutt_replacelist_match /tmp/neomutt_build/mutt/regex.c:508:27
    #1 0x5cc855c275a0 in main poc.cpp:30:5
    #2 0x7e0a5fccc1c9 in __libc_start_call_main csu/../sysdeps/nptl/libc_start_call_main.h:58:16
    #3 0x7e0a5fccc28a in __libc_start_main csu/../csu/libc-start.c:360:3
    #4 0x5cc855b3c644 in _start (poc+0x49644) (BuildId: 80dc2e8e4923fe5654c5e0fafa035f4bc297662d)

==103586==Register values:
rax = 0x0000000000000000  rbx = 0x00007fff5f462500  rcx = 0x0000000000000000  rdx = 0x0000000000000000  
rdi = 0x00007fff5f461858  rsi = 0x0000000000000000  rbp = 0x00007fff5f4625f0  rsp = 0x00007fff5f462500  
 r8 = 0x00000000000003ff   r9 = 0x0000000000000000  r10 = 0xafffff00000fffff  r11 = 0x0000000000000246  
r12 = 0x00005cc855c6a320  r13 = 0x0000000000000000  r14 = 0x0000000000000000  r15 = 0x00007a3a5efe0170  
AddressSanitizer can not provide additional info.
SUMMARY: AddressSanitizer: SEGV /tmp/neomutt_build/mutt/regex.c:508:27 in mutt_replacelist_match
```

### PoC Code
```cpp
/* Root cause: mutt_replacelist_add() stores mutt_str_dup(templ) into
 * np->templ. mutt_str_dup("") returns NULL, so an empty (but non-NULL)
 * template leaves np->templ == NULL while the node is still added.
 * mutt_replacelist_match() then runs `for (p = np->templ; *p && ...)`
 * and dereferences NULL.
 */
#include <cstring>

extern "C" {
#include "config.h"
#include "mutt/lib.h"
}

bool StartupComplete = true;

int main()
{
    MuttLogger = log_disp_null;

    struct Buffer *err = buf_new(NULL);
    struct ReplaceList repl;
    STAILQ_INIT(&repl);

    /* Empty template "" is non-NULL, so the `!templ` guard passes — but mutt_str_dup("") yields NULL, so np->templ = NULL (node still added). */
    mutt_replacelist_add(&repl, "a", "", err);

    /* "a" matches "a"; on match the loop dereferences np->templ (NULL) at mutt/regex.c:508 → UBSan null load / ASan SEGV. */
    char buf[1024];
    memset(buf, 0, sizeof(buf));
    mutt_replacelist_match(&repl, buf, sizeof(buf), "a");

    mutt_replacelist_free(&repl);
    buf_free(&err);
    return 0;
}
```

### Reproduction Step

```bash
# 1. Checkout NeoMutt at the version above
git clone https://github.com/neomutt/neomutt.git
cd neomutt
git checkout fb8e54b69cff6c7257d348122d19459cba5539bc

# 2. Build the library objects with ASan+UBSan.
export CC=clang CXX=clang++
export CFLAGS="-fsanitize=address,undefined -g -O0 -fno-omit-frame-pointer -w -fPIC"
export CXXFLAGS="$CFLAGS" LDFLAGS="$CFLAGS"
./configure --disable-doc --disable-nls --disable-pgp --disable-smime \
            --disable-inotify --disable-idn2 CC="$CC" CXX="$CXX"
make -j"$(nproc)" || true   # binary link fails; sub-libraries are built before that

# 3. Build the standalone PoC
clang++ -g -O0 -fsanitize=address,undefined -fno-omit-frame-pointer \
    -I. \
    poc.cpp -o poc \
    -Wl,--start-group lib*.a -Wl,--end-group -lncursesw -ltinfo

# 4. Run
./poc
```

- Always, the crash is deterministic.

Not tied to an upgrade or a particular email. It is a latent bug in the `ReplaceList` implementation in `mutt/regex.c`.

## Submission Statement

This report was produced by FuzzAnything's AI-assisted library fuzzer and manually verified by a team member. We reviewed the PoC against the upstream API documentation — call order, parameters, and memory ownership — and found no API misuse.

> Signed-off-by: FuzzAnything <fuzzanything@gmail.com>

yvonnelxxxx:
This PR attempts to fix [#5010](https://github.com/neomutt/neomutt/issues/5010#issue-5107254173)

## Problem

`mutt_replacelist_add()` only NULL-checks its `templ` argument, never testing for the empty string:

```c
if (!rl || !pat || (*pat == '\0') || !templ)
  return 0;
...
np->templ = mutt_str_dup(templ);
```

`mutt_str_dup()` is documented to return `NULL` for an empty string (`mutt/string.c`: *"str was NULL or empty"*). So a non-NULL but empty template `""` leaves `np->templ == NULL` on a node that is still inserted into the list. `mutt_replacelist_match()` then dereferences it with no guard as soon as the regex matches:

```c
for (p = np->templ; *p && (tlen < (buflen - 1));)   /* mutt/regex.c:508 — *p reads NULL */
```

This is a deterministic null-pointer dereference (UBSan null load + ASan SEGV). The sibling function `mutt_replacelist_apply()` already guards the identical loop with `if (np->templ)` at `mutt/regex.c:400` — `match` is simply missing the check `apply` has, so the two are inconsistent.

The `ReplaceList` mechanism backs NeoMutt's `subjectrx`/`spam` commands, so an empty configured replacement template is a realistic, user-reachable trigger rather than an artificial input.

## Fix

Two complementary, defense-in-depth changes to `mutt/regex.c`. Either alone prevents the crash; together they restore the invariant (`np->templ != NULL` for every list node) *and* make `match` as robust as `apply`.

### 1. Reject empty templates at construction (`mutt_replacelist_add`)

Before:

```c
if (!rl || !pat || (*pat == '\0') || !templ)
  return 0;
```

After:

```c
if (!rl || !pat || (*pat == '\0') || !templ || (*templ == '\0'))
  return 0;
```

### 2. Guard `np->templ` in the consumer (`mutt_replacelist_match`)

Before:

```c
/* Copy template into buf, with substitutions. */
for (p = np->templ; *p && (tlen < (buflen - 1));)
{
  /* ... substitution loop body ... */
}
```

After:

```c
/* Copy template into buf, with substitutions.
 * Guard np->templ like mutt_replacelist_apply() does. */
if (np->templ)
{
  for (p = np->templ; *p && (tlen < (buflen - 1));)
  {
    /* ... substitution loop body, unchanged ... */
  }
}
```

## Verification

Reproduced deterministically against commit `fb8e54b69cff6c7257d348122d19459cba5539bc ` with an ASan+UBSan static build of `libneomutt.a` and a standalone PoC that calls the public API:

```c
mutt_replacelist_add(&repl, "a", "", err);                 // empty template
mutt_replacelist_match(&repl, buf, sizeof(buf), "a");      // match → deref NULL
```

**Before the fix**

```
regex.c:508:27: runtime error: load of null pointer of type 'char'
SUMMARY: UndefinedBehaviorSanitizer: null-pointer-use mutt/regex.c:508:27
==NNN==ERROR: AddressSanitizer: SEGV on unknown address 0x000000000000
    #0 mutt_replacelist_match .../mutt/regex.c:508:27
AddressSanitizer: deadly signal (ABRT)            [ exit 6 ]
```

**After the fix**

Clean exit, no sanitizer output.

## Submission Statement

This fix was generated using Claude Code and manually reviewed, tested, and verified by a team member.

Signed-off-by: FuzzAnything <fuzzanything@gmail.com>

### DIFF
--- a/mutt/regex.c
+++ b/mutt/regex.c
@@ -273,7 +273,7 @@ int mutt_regexlist_remove(struct RegexList *rl, const char *str)
 int mutt_replacelist_add(struct ReplaceList *rl, const char *pat,
                          const char *templ, struct Buffer *err)
 {
-  if (!rl || !pat || (*pat == '\0') || !templ || (*templ == '\0'))
+  if (!rl || !pat || (*pat == '\0') || !templ)
     return 0;
 
   struct Regex *rx = mutt_regex_compile(pat, REG_ICASE);
--- a/mutt/regex.c
+++ b/mutt/regex.c
@@ -273,7 +273,7 @@ int mutt_regexlist_remove(struct RegexList *rl, const char *str)
 int mutt_replacelist_add(struct ReplaceList *rl, const char *pat,
                          const char *templ, struct Buffer *err)
 {
-  if (!rl || !pat || (*pat == '\0') || !templ)
+  if (!rl || !pat || (*pat == '\0') || !templ || (*templ == '\0'))
     return 0;
 
   struct Regex *rx = mutt_regex_compile(pat, REG_ICASE);
@@ -505,33 +505,36 @@ bool mutt_replacelist_match(struct ReplaceList *rl, char *buf, size_t buflen, co
       mutt_debug(LL_DEBUG5, "%d subs\n", (int) np->regex->regex->re_nsub);
 
       /* Copy template into buf, with substitutions. */
-      for (p = np->templ; *p && (tlen < (buflen - 1));)
+      if (np->templ)
       {
-        /* backreference to pattern match substring, eg. %1, %2, etc) */
-        if (*p == '%')
+        for (p = np->templ; *p && (tlen < (buflen - 1));)
         {
-          char *e = NULL; /* used as pointer to end of integer backreference in strtol() call */
-
-          p++; /* skip over % char */
-          long n = strtol(p, &e, 10);
-          /* Ensure that the integer conversion succeeded (e!=p) and bounds check.  The upper bound check
-           * should not strictly be necessary since add_to_spam_list() finds the largest value, and
-           * the static array above is always large enough based on that value. */
-          if ((e != p) && (n >= 0) && (n < np->nmatch) && (pmatch[n].rm_so != -1))
+          /* backreference to pattern match substring, eg. %1, %2, etc) */
+          if (*p == '%')
           {
-            /* copy as much of the substring match as will fit in the output buffer, saving space for
-             * the terminating nul char */
-            for (int idx = pmatch[n].rm_so;
-                 (idx < pmatch[n].rm_eo) && (tlen < (buflen - 1)); idx++)
+            char *e = NULL; /* used as pointer to end of integer backreference in strtol() call */
+
+            p++; /* skip over % char */
+            long n = strtol(p, &e, 10);
+            /* Ensure that the integer conversion succeeded (e!=p) and bounds check.  The upper bound check
+             * should not strictly be necessary since add_to_spam_list() finds the largest value, and
+             * the static array above is always large enough based on that value. */
+            if ((e != p) && (n >= 0) && (n < np->nmatch) && (pmatch[n].rm_so != -1))
             {
-              buf[tlen++] = str[idx];
+              /* copy as much of the substring match as will fit in the output buffer, saving space for
+               * the terminating nul char */
+              for (int idx = pmatch[n].rm_so;
+                   (idx < pmatch[n].rm_eo) && (tlen < (buflen - 1)); idx++)
+              {
+                buf[tlen++] = str[idx];
+              }
             }
+            p = e; /* skip over the parsed integer */
           }
-          p = e; /* skip over the parsed integer */
-        }
-        else
-        {
-          buf[tlen++] = *p++;
+          else
+          {
+            buf[tlen++] = *p++;
+          }
         }
       }
       /* tlen should always be less than buflen except when buflen<=0


gahr:
Thanks! Please let's go for option 2: I think an empty second argument in the [spam command](https://neomutt.org/guide/configuration#26-%C2%A0spam-detection) is legit.

gahr:
Smaller reproducer:

`:spam foo ""`

In any mailbox where `foo` matches.

yvonnelxxxx:
@gahr Thanks for the review! I've updated the PR to keep option 2 only.

gahr (approved):
Thanks a lot!

flatcap:
Thanks, @yvonnelxxxx 
Keep 'em coming!

PR #5021, opened 2026-08-15
Remove four unreferenced functions

chrisdebian:
## What

Four functions with no callers anywhere in the tree:

- `pgp_canencrypt()` / `pgp_cansign()` (`ncrypt/pgplib.c`) -- superseded by the `KEYFLAG_CANENCRYPT`/`KEYFLAG_CANSIGN` bitflag system used throughout `ncrypt/` (`gnupgparse.c`, `smime.c`, `pgp.c`, `crypt_gpgme.c`, and others).
- `progress_set_size()` (`progress/progress.c`) -- a one-line wrapper around `progress_window_set_size()` with no callers of its own.
- `regex_color_list_new()` (`color/regex.c`) -- constructor for `RegexColorList` with no callers anywhere.

## How this was found

`cppcheck --enable=unusedFunction` over the whole project, cross-checked against a `-ffunction-sections`/`--gc-sections` linker build (which sees through function-pointer dispatch tables correctly, unlike the static analyzer alone -- two early candidates, `alias_dialog()` and `query_index()`, were false positives from exactly that pattern and are excluded here).

The remaining candidates were then verified by hand against the whole source tree with a plain text search, not just what compiles in one configuration -- a couple of others turned out to be reachable only through optional subsystems this local build doesn't have enabled (SSL/TLS, autocrypt+GPGME) and are correctly *not* included in this PR. The four here have no call site under any configuration.

## Verification

Clean build (clang, zero new warnings), 641/641 tests passing (`test/neomutt-test`, using `NEOMUTT_TEST_DIR`).

## Unrelated aside, FYI only

While working in this area I noticed `.github/workflows/clang-format.yml` only checks `*.c` files (`git ls-files ... '*.c'`), never headers -- and the three headers touched here weren't clang-format-22.1.8-clean even before this PR (pristine `main` included). Left them exactly as they were beyond the one-line removal each, since two of them use hand-aligned column formatting that clang-format's automatic reflow doesn't reproduce, so this may well be deliberate rather than an oversight. Flagging in case it's useful, not proposing a change.

This PR was written primarily by Claude Code, based on findings I reviewed and independently verified before opening it.

### DIFF
--- a/color/regex.c
+++ b/color/regex.c
@@ -147,22 +147,6 @@ struct RegexColor *regex_color_new(void)
 }
 
 /**
- * regex_color_list_new - Create a new RegexColorList
- * @retval ptr New RegexColorList
- */
-struct RegexColorList *regex_color_list_new(void)
-{
-  struct RegexColorList *rcl = MUTT_MEM_CALLOC(1, struct RegexColorList);
-
-  STAILQ_INIT(rcl);
-
-  struct RegexColor *rcol = regex_color_new();
-  STAILQ_INSERT_TAIL(rcl, rcol, entries);
-
-  return rcl;
-}
-
-/**
  * regex_color_list_clear - Free the contents of a RegexColorList
  * @param rcl List to clear
  *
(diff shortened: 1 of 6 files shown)


gahr:
See https://github.com/neomutt/neomutt/blob/5238806174a7ba4caf5a26b7548f036994dec004/.github/unused.sh#L17-L19 for `progress_set_size` and `regex_color_list_new`. I'd leave them alone for now..

I agree with killing the PGP ones.

Issue #5005, opened 2026-08-09
OpenSSL handshake and write retries can exceed `socket_timeout`


toddagibson:
## NeoMutt Version

Runtime reproduction:

```text
NeoMutt 20260616
System: Darwin 25.6.0 (arm64)
OpenSSL: OpenSSL 3.6.2 7 Apr 2026
Compile options include +openssl
```

I also verified that the relevant `ssl_negotiate()` retry loop remains unchanged on `main` as of `f5ddd2d9`.

## Expected Behaviour

If a TCP connection is established but the peer does not make progress during the TLS handshake, NeoMutt should stop waiting when `socket_timeout` expires.

The same bound should apply when an established TLS peer stops accepting writes.

## Actual Behaviour

`SSL_connect()` can be retried indefinitely after the underlying socket timeout. With `socket_timeout=2`, NeoMutt was still blocked when an independent seven-second watchdog stopped it:

```text
rc=124 elapsed=7s outer_timeout=yes
```

The same behavior is independently reproducible in `SSL_write()`. A loopback TLS SMTP server completes the SMTP exchange through `DATA`, then deliberately stops reading the message body. NeoMutt again exceeds `socket_timeout=2` and reaches an eight-second outer watchdog:

```text
rc=124 elapsed=8s outer_timeout=yes
```

### Handshake

1. Start a loopback TCP listener that accepts a connection but sends no TLS handshake data.
2. Configure NeoMutt to open an IMAPS connection to that listener with `socket_timeout=2`.
3. Run NeoMutt under an independent seven-second watchdog.

### Stalled Write

1. Start a loopback TLS SMTP server and configure NeoMutt with `socket_timeout=2`.
2. Complete the SMTP exchange through `DATA` and return the `354` response.
3. Stop reading while NeoMutt sends a message large enough to create socket backpressure.
4. Run NeoMutt under an independent eight-second watchdog.

In both cases, the watchdog expires instead of NeoMutt returning after approximately two seconds. The tests use only loopback listeners and dummy account values.

Both cases reproduce consistently in their local harnesses. Each harness will be included with the corresponding fix, demonstrating that the current behavior fails and the corrected behavior passes deterministically.

## Cause

In `ssl_negotiate()`, a failed `SSL_connect()` is immediately retried whenever `BIO_should_retry(SSL_get_rbio(...))` is true:

```c
retry:
  err = SSL_connect(ssldata->ssl);
  if (err != 1)
  {
[...]

PR #5028, opened 2026-08-31
openssl: honor socket timeout during handshake

toddagibson:
* **What does this PR do?**

Depends on #5007. The first commit is that PR; the handshake-specific changes
begin with the second commit.

Make the OpenSSL TLS handshake honor `socket_timeout` when a peer accepts TCP
but does not send a TLS response.

`SSL_connect()` can return a retryable read condition after the underlying
socket timeout. The old loop immediately called `SSL_connect()` again, allowing
each retry to consume another full timeout indefinitely. The new handling:

- retries harmless `EINTR` interruptions without requiring immediate readability;
- reports cancellation when `SIGINT` interrupts negotiation;
- otherwise preserves existing non-read retry behavior and retries a read
  condition only when the socket has become readable;
- reports the existing translated timeout error when no data is available.

Add an OpenSSL-only loopback regression. A silent peer accepts TCP while
`mutt_socket_open()` runs in a child process with `socket_timeout=2`; the parent
enforces an independent seven-second deadline. The old loop fails at that
deadline, while corrected negotiation returns after about two seconds. The
test also verifies that the peer is still open.

Extend that fixture with a delayed successful handshake, successful negotiation
after `SIGWINCH`, a silent peer that still times out after `SIGWINCH`, and prompt
cancellation after `SIGINT`. The signal cases use the same negotiation function
through STARTTLS, with non-restarting handlers installed after TCP connection
setup. A one-byte acknowledgment keeps successful clients connected until the
server finishes its handshake.

Factor the existing OpenSSL timeout fixture into shared helpers for loopback
setup, ephemeral TLS credentials, an idle post-handshake peer, child result
transport, watchdog waiting, and cleanup. The existing read-timeout regression
now uses the same fixture without changing its scenario.

Not applicable.

- User documentation is unchanged; a handshake waiting for a TLS response now
  honors the existing `socket_timeout` setting.
- The new test and helpers include Doxygen documentation.
- Local OpenSSL focused and full test suites pass.
- The repository Ubuntu CI container passes default, OpenSSL, and
  all-features/GnuTLS matrices, including docs, full tests, and
  install/uninstall cleanup.
- OpenSSL-only tests are omitted from non-OpenSSL configurations.
- Changed C and header sources pass pinned `clang-format` 22.1.8 and
  `git diff --check`.

Second focused step in #5005.

Extends and factors the OpenSSL loopback regression introduced by #5007.

## Validation

Tested on arm64 Darwin 25.6 with OpenSSL 3.6.4 and on x86_64 Linux with
OpenSSL 3.0.13.

```text
unfixed handshake: 3/3 runs failed at the seven-second watchdog
corrected handshake and signal cases: 3/3 focused runs passed
interruption correction removed: new regression failed as expected
cancellation handling removed: SIGINT case failed as expected
existing read timeout regression: passed
local full OpenSSL suite: passed (644 tests)
Ubuntu CI image: default, OpenSSL, and all-features/GnuTLS matrices passed
```

## AI assistance disclosure

OpenAI Codex assisted with converting the reproducer into the loopback test
fixture, implementing and validating the focused retry correction and signal
regressions, running the build matrix, factoring the shared test fixture, and
drafting the commit and pull-request text. The author reviewed the original
OpenSSL and socket control flow and test implementation, and personally
validated the original passing and deliberately regressed behavior.

### DIFF
--- a/conn/openssl.c
+++ b/conn/openssl.c
@@ -1176,12 +1176,24 @@ static int ssl_negotiate(struct Connection *conn, struct SslSockData *ssldata)
   ERR_clear_error();
 
 retry:
+  errno = 0;
   err = SSL_connect(ssldata->ssl);
+  const int saved_errno = errno;
   if (err != 1)
   {
-    // Temporary failure, e.g. signal received
     if (BIO_should_retry(SSL_get_rbio(ssldata->ssl)))
     {
+      // An interrupted read need not have any data ready yet.
+      if (saved_errno == EINTR)
+      {
+        if (SigInt)
+        {
+          mutt_error(_("Connection to %s has been aborted"), conn->account.host);
+          return -1;
+        }
+        goto retry;
+      }
+
       if (!BIO_should_read(SSL_get_rbio(ssldata->ssl)))
         goto retry;
 
--- a/conn/openssl.c
+++ b/conn/openssl.c
@@ -1181,7 +1181,19 @@ retry:
   {
     // Temporary failure, e.g. signal received
     if (BIO_should_retry(SSL_get_rbio(ssldata->ssl)))
-      goto retry;
+    {
+      if (!BIO_should_read(SSL_get_rbio(ssldata->ssl)))
+        goto retry;
+
+      const int poll_rc = raw_socket_poll(conn, 0);
+      if (poll_rc > 0)
+        goto retry;
+      if (poll_rc == 0)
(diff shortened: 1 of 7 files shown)


gahr:
I wonder if it's sustainable to introduce 850 lines of tests for a 20 lines change...

PR #5034, opened 2026-09-13
b64_encode_common: Initialise out if inlen is zero

MaxHearnden:
* **What does this PR do?**

At the moment, out is not initialised if in is NULL. This affects the built in GSSAPI authentication as sometimes zero length responses are sent to the server.

This commit makes b64_encode_common initialise out if it is not NULL and inlen is zero even if in is NULL.

### DIFF
--- a/mutt/base64.c
+++ b/mutt/base64.c
@@ -101,6 +101,12 @@ const int Index64[128] = {
 static size_t b64_encode_common(const char *in, size_t inlen, char *out,
                                 size_t outlen, const char *alpha)
 {
+  if (inlen == 0 && out)
+  {
+    *out = '\0';
+    return 0;
+  }
+
   if (!in || !out)
     return 0;
 


gahr:
That makes sense.. I suggest a sightly different diff:

```diff
diff --git a/mutt/base64.c b/mutt/base64.c
index da12af5f4..a6310cb80 100644
--- a/mutt/base64.c
+++ b/mutt/base64.c
@@ -101,9 +101,15 @@ const int Index64[128] = {
 static size_t b64_encode_common(const char *in, size_t inlen, char *out,
                                 size_t outlen, const char *alpha)
 {
-  if (!in || !out)
+  if (!out)
     return 0;
 
+  if (!in || !*in)
+  {
+    *out = '\0';
+    return 0;
+  }
+
   unsigned char *begin = (unsigned char *) out;
   const unsigned char *inu = (const unsigned char *) in;
```

Are you good with that?

PR #5036, opened 2026-09-14
Make sure mutt_b64_encode null-terminates the output

gahr:
Alternative for #5034 

@MaxHearnden


mutt/base64.c
```diff
@@ -101,9 +101,15 @@ const int Index64[128] = {
 static size_t b64_encode_common(const char *in, size_t inlen, char *out,
                                 size_t outlen, const char *alpha)
 {
-  if (!in || !out)
+  if (!out)
     return 0;
 
+  if (!in || !*in)
```
MaxHearnden:
```suggestion
  if (!in || inlen == 0)
```
in is not necessarily a null terminated string and might have a null byte at the beginning followed by more bytes that need to be base64 encoded.

MaxHearnden (changes requested):

mutt/base64.c
gahr: That makes sense, thanks!

MaxHearnden (approved):
Looks good, in comparison with my PR this also null terminates the output if the input is NULL.

PR #5037, opened 2026-09-17
fix: don't truncate attachment when saving it onto itself

### LINKED ISSUE #4988: `<copy-file>` of compose menu deletes contents of attached message
## NeoMutt Version

```
20251211-1017-356f04 (close to main at time of writing)
```

## Expected Behaviour

Do not delete the contents of an attached message.

## Actual Behaviour

The attached message becomes empty.

## Steps to Reproduce

Prepare a draft message with NeoMutt which contains an attached file. Due to bug https://github.com/neomutt/neomutt/issues/4987 I did that with a different NeoMutt config and saved the draft message in `~/postponed` (default postponed folder location).

### With `$attach_save_without_prompting` set:

1. `neomutt -F /dev/null`
2. `<recall-message>`
3. select attached message
4. `:set attach_save_without_prompting`
5. `<copy-file>` if `$attach_save_without_prompting` is not set, then choose the default filename.
6. NeoMutt asks overwrite?
7. Say cancel
8. The attachment became empty (can be tested with `<copy-file>` or `<pipe-entry>`)

### With overwrite default path

1. `<recall-message>`
2. go to the attached message
3. `<copy-file>` use default path
4. NeoMutt asks if it should overwrite, say yes
5. Target file became empty

Probably, target file and source file are the same which might be the/a problem.

### Additional Info

The `<save-entry>` of the attach menu does not show this behaviour.

Always

ethanstoner:
* **What does this PR do?**

  Fixes `<copy-file>` in the compose menu emptying the attached file.

  - In compose, the default save path is the attachment's own file. `mutt_save_attachment()` (send mode) opened the target with `"w"` before copying from the source, so saving onto the same file truncated it to 0 bytes (and `(a)ppend` would append the file to itself). It now compares `st_dev`/`st_ino` of the source and target, and refuses with "Can't save an attachment onto itself". The normal prompt then asks for another file name.
  - `save_without_prompting()` (used when `$attach_save_without_prompting` is set) only aborted when `mutt_check_overwrite()` returned `-1`. Choosing `(c)ancel` returns `1`, so it saved anyway. Any non-zero result now aborts. `(a)ppend` and `(o)verwrite` both return `0`, so they're unaffected.

   - All builds and tests are passing. New unit test `test_mutt_save_attachment` covers a normal copy plus overwrite/append/hard-link onto the source file. It fails on `main` (source truncated) and passes with this change. `make test`: 642/642 pass (641/641 on `main`), macOS arm64.
   - Manual check in tmux (`neomutt -n -F rc -s test -a att.txt -- a@example.com`, select the attachment, `C`):
     - Default path + `(o)verwrite`: on `main`, "Attachment saved" and `att.txt` is 0 bytes. With the fix, it shows the error, reprompts, and `att.txt` is intact.
     - `$attach_save_without_prompting` + `(c)ancel`: on `main`, "Attachment saved" and `att.txt` is 0 bytes. With the fix, nothing is saved and `att.txt` is intact.
   - Code is clang-formatted. No docs change is needed.

  Fixes #4988

This PR was written primarily by Claude Code.


attach/mutt_attach.c
```diff
@@ -1002,6 +1002,17 @@ int mutt_save_attachment(FILE *fp, struct Body *b, const char *path,
...
 
+    /* Don't truncate (or endlessly append to) the attachment's own file */
+    struct stat st_old = { 0 };
+    struct stat st_new = { 0 };
+    if ((fstat(fileno(fp_old), &st_old) == 0) && (stat(path, &st_new) == 0) &&
+        (st_old.st_dev == st_new.st_dev) && (st_old.st_ino == st_new.st_ino))
+    {
+      mutt_error(_("Can't save an attachment onto itself"));
+      mutt_file_fclose(&fp_old);
+      return -1;
```
gahr, line 1013: I guess an alternative here would be to return success...

gahr:
> * and `(a)ppend` would append the file to itself

wow, I just witnessed NeoMutt trying to create an infinite-sized file because of this!

gahr (approved):

